PR #1156, opened 2023-11-05
feat(linter): support eslint(default-case-last)

PerfectPan:


crates/oxc_linter/src/rules/eslint/default_case_last.rs
```diff
@@ -0,0 +1,119 @@
...
+    ///     default:
+    ///         baz();
+    ///         break;
+    ///     case 2:
+    ///         qux();
+    ///         break;
+    /// }
+    /// ```
+    DefaultCaseLast,
+    correctness
```
Boshen, line 53: I wonder whether this should be a `style` rule?

crates/oxc_linter/src/snapshots/default_case_last.snap
```diff
@@ -0,0 +1,104 @@
+---
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 105
+expression: default_case_last
+---
+  ⚠ eslint(default-case-last): Enforce default clauses in switch statements to be last
+   ╭─[default_case_last.tsx:1:1]
+ 1 │ switch (foo) { default: bar(); break; case 1: baz(); break; }
```
Boshen, line 8: nit: span the `default` keyword only, the body is uninteresting regarding the error. You can use `Span::new(default_clause.span.start, default_clause.span.start + 7)`.

crates/oxc_linter/src/rules/eslint/default_case_last.rs
PerfectPan: Sorry, I didn't notice it. I will check it later.

crates/oxc_linter/src/snapshots/default_case_last.snap
PerfectPan: Get.

PR #1157, opened 2023-11-05
chore: 🤖 make FileSystem trait object safe

IWANABETHATGUY:
1. make FileSystem trait object safe

### DIFF
--- a/crates/oxc_resolver/src/file_system.rs
+++ b/crates/oxc_resolver/src/file_system.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 /// File System abstraction used for `ResolverGeneric`.
-pub trait FileSystem: Default + Send + Sync {
+pub trait FileSystem: Send + Sync {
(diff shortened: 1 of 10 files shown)


crates/oxc_resolver/src/file_system.rs
```diff
@@ -4,7 +4,7 @@ use std::{
 };
 
 /// File System abstraction used for `ResolverGeneric`.
-pub trait FileSystem: Default + Send + Sync {
```
IWANABETHATGUY, line 7: `Default` has a trait bound `Sized`, this trait bound make the triat isn't object safety any more,  https://doc.rust-lang.org/reference/items/traits.html#object-safety:~:text=Sized%20must%20not%20be%20a%20supertrait.%20In%20other%20words%2C%20it%20must%20not%20require%20Self%3A%20Sized.

Boshen (approved):

PR #1172, opened 2023-11-06
feat(linter): eslint-plugin-jest: no-hooks

eryue0220:


camc314:
do we want to port this one? 
Looks like it isn't in the recommended set?

crates/oxc_linter/src/rules/jest/no_hooks.rs
```diff
@@ -0,0 +1,147 @@
...
+
+use crate::{
+    context::LintContext,
+    rule::Rule,
+    utils::{is_type_of_jest_fn_call, JestFnKind, JestGeneralFnKind, JEST_HOOK_NAMES},
+    AstNode,
+};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint-plugin-jest(no-hooks): ")]
```
Boshen, line 17: This is missing error message and help texts.

Boshen:
> do we want to port this one? Looks like it isn't in the recommended set?

I want to ship recommended set for a more complete product, but merging in non-recommended rules is acceptable because the end goal is to as many rules as we can. Clippy has more than 600 rules, we can probably reach 1000 😂

crates/oxc_linter/src/rules/jest/no_hooks.rs
```diff
@@ -0,0 +1,147 @@
...
+            &[JestFnKind::General(JestGeneralFnKind::Hook)],
+        ) {
+            return;
+        }
+
+        if let Expression::Identifier(ident) = &call_expr.callee {
+            if JEST_HOOK_NAMES.contains(&ident.name.as_str())
+                && !self.allow.contains(&ident.name.to_string())
+            {
+                ctx.diagnostic(UnexpectedHookDiagonsitc(call_expr.span));
```
mysteryven, line 105: What about we report on `call_expr.callee`? Some people may write many line codes in the hooks function, in our codebase, we prefer it to focus on the error position more detailed, and do you think it will make our find the error easier?

mysteryven (changes requested):
I think we can ignore the benchmark, which is not related to this PR, just two nit picks above.😉

crates/oxc_linter/src/rules/jest/no_hooks.rs
eryue0220: Done.

mysteryven (approved):
Nice work! Thanks for your contribution, hope you enjoy it!

crates/oxc_linter/src/rules/jest/no_hooks.rs
```diff
@@ -0,0 +1,147 @@
...
+            call_expr,
+            node,
+            ctx,
+            &[JestFnKind::General(JestGeneralFnKind::Hook)],
+        ) {
+            return;
+        }
+
+        if let Expression::Identifier(ident) = &call_expr.callee {
+            if JEST_HOOK_NAMES.contains(&ident.name.as_str())
```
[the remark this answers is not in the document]
eryue0220: Oh, Sure, thanks, I will fix it.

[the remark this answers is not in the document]
eryue0220: see: https://github.com/web-infra-dev/oxc/pull/1185

PR #1173, opened 2023-11-06
feat(transform): TypeScript Enum

magic-akari:
In this implementation, we have avoided attempting to inline the values of enum members, nor have we endeavored to compute the values of enum members or even calculate their increments. Instead, we deferred all computations to runtime, resulting in a straightforward implementation of our solution.

This implementation took inspiration from [Sucrase](https://sucrase.io/#transforms=typescript&compareWithTypeScript=true&code=enum%20Foo%20%7B%0A%20%20a%2C%0A%20%20%22b%22%2C%0A%20%20c%20%3D%20%22c%22%2C%0A%20%20d%20%3D%20a%2C%0A%20%20%22e-e%22%2C%0A%20%20f%0A%7D).

Input:
```TypeScript
enum Foo {
    a,
    b,
    c = "c",
    d = a,
    "e-e",
    f,
}
```

output:
```JavaScript
var Foo;
(Foo => {
        const a = 0;
        Foo[Foo['a'] = a] = 'a';
        const b = 1 + a;
        Foo[Foo['b'] = b] = 'b';
        const c = 'c';
        Foo['c'] = c;
        const d = a;
        Foo[Foo['d'] = d] = 'd';
        Foo[Foo['e-e'] = 1 + d] = 'e-e';
        const f = 1 + Foo['e-e'];
        Foo[Foo['f'] = f] = 'f';
})(Foo ||= {});
```


Boshen:
Nice! We can start merging once the conformance snapshot is updated!

magic-akari:
In this implementation, we have avoided attempting to inline the values of enum members, nor have we endeavored to compute the values of enum members or even calculate their increments. Instead, we deferred all computations to runtime, resulting in a straightforward implementation of our solution.

This implementation took inspiration from [Sucrase](https://sucrase.io/#transforms=typescript&compareWithTypeScript=true&code=enum%20Foo%20%7B%0A%20%20a%2C%0A%20%20%22b%22%2C%0A%20%20c%20%3D%20%22c%22%2C%0A%20%20d%20%3D%20a%2C%0A%20%20%22e-e%22%2C%0A%20%20f%0A%7D).

Boshen:
This is interesting. I'm curious on how we are going to align the tests with babel, if it cannot be aligned, we should somehow port all the test cases from surcase 🤔

magic-akari:
> [!NOTE] 
It's possible to inline the variable in compress/minify phase.

https://play.swc.rs/?version=1.3.98-nightly-20231115.1&code=H4sIAAAAAAAAA22QsQrDMAxE93zFbU4IhWY27pifKB1kOx5rSEqXJP9eSQ0tdW2wEOd3J%2BEnzRhztk3LFe6CtcFxQr4vDxAczvYjMnWVa8jc%2BIWkcG8Ll2d5QA%2BqOL06vTr9nzOIHMyvjwXBQwlHWaEyIioedUQ0FWA6TYrIjm9MlDI%2BHcTXUolKGpQ0JHHE3ulXbpvDunf2Bb0TmjNhAQAA&config=H4sIAAAAAAAAA32UO3LjMAyG%2B5zCo3qLHRcpcoB0ewYOLYIysxShIUDHmozvHlCSH4khdRI%2B%2FAAJgPh62e2aD2qbt92XfMrPYDNBvv2LhcbE9iyWhscBqM1h4ObPlTJV5G0kmEyXmTRscwdcVUD7v%2Fv9omgiIsFVsdj6kIIfH3O22A8ZiB5sYpWQpYfE9FO%2FsIyfFXAuj%2FYDYgSbNoixZEJi6CBrgVuM0Q4E5mSzEqWe1OZAqKWosDA4M2QcVJ5c4IBJcj5TB9aZFh0oKGRoOZxAk0kukSWS6yn3mbCDQ%2Bm6qc%2B%2F1HCysVhWcsJ5aomcVol6xEBsfElaCWe4UoMZLsX9rQzeZOCS07PuA0Na6cl%2FAKlAtETJ9qDFnTy8zNOa2m8qQ%2FIysjwqXOZbu2WCTopqQvBKZWtlIHPQupnBlRZqZVvtOAteKR8FBwa8l1lRQtNn4PaoJa0PHb0CpL%2FWa1M1A3N7hSu8PogN%2FC63ZH3AFo%2Fe8nGd0tgfMG4k6IGP6DYcpBWM6zjLljgP67wkBzIa4FSXQhN4XgLyABhNnPbl02zI85CIpot4uK%2BJxeFy28O9Td39vc%2Br%2BGVxaHp0ZYLLkq%2F9nVfza3N3um7h28GbQP%2Buyinp5RvxWkerMAYAAA%3D%3D

tasks/transform_conformance/ts_fixutures/enums__awaitAndYield.ts
```diff
@@ -0,0 +1,9 @@
+async function* test(x) {
+	var E;
+	(E => {
+		const foo = await x;
```
magic-akari, line 4: tsc will report error with this case.
[playground](https://www.typescriptlang.org/play?target=6#code/PTAEAEBcEMCcHMCmkBcoCiBlATABgIwCcAUCBAHYD2AKgJ4AOiAzgILkAmmtAtgEaUAbJmkiwAromLQmtcgGNQAMzHzIAS0rkAVKEjNIACgAeaAAqxK3NU0QAecmL6JYAPgCUoAN7FQv0IgduDC8fPzDFSkpQAF5QaAB3aDVIUCMAGlCw315oAC8Y0Fo1RAF2UHwMsIBfYiqgA)

magic-akari:
@Boshen I have updated the test cases.

tasks/transform_conformance/ts_fixutures/enums__enumErrors.ts
Boshen, line 1: Empty files?

crates/oxc_transformer/src/typescript/mod.rs
Boshen, line 1: We need to document enum transformation aligns with typescript instead of babel here. We also need to document why we are picking typescript instead of babel.

tasks/transform_conformance/src/lib.rs
Boshen, line 1: I think we need to somehow exclude the tests from babel.

Boshen:
@IWANABETHATGUY is picking up the transformer, please take a look.

tasks/transform_conformance/ts_fixutures/enums__enumErrors.ts
magic-akari: Parser error.

```
  × Expected `,` but found `;`
    ╭─[47:1]
 47 │ 
 48 │     postSemicolon;
    ·                  ┬
    ·                  ╰── `,` expected
 49 │     postColonValueComma: 2,
    ╰────
```

crates/oxc_transformer/src/typescript/mod.rs
magic-akari: We only used the input values of TypeScript test cases, but we didn't compare the results.
Our transformation results are different from TypeScript and Babel.

Including Babel test cases would be feasible as well.
Since our transformation results differ from both TypeScript and Babel, we currently face the challenge of manually verifying the output.

Perhaps one approach could be to check each value of the enum and run the program to compare the output results. This approach would be viable for enums that exist at the top level of the scope, but it would increase the complexity of our test code.

magic-akari:
For the input value:

```TypeScript
enum Foo {
    A = 1,
    B = A,
}
```

A typical tsc transformation result:
```JavaScript
var Foo;
(function (Foo) {
    Foo[Foo["A"] = 1] = "A";
    Foo[Foo["B"] = 1] = "B";
})(Foo || (Foo = {}));
```

Babel transformation result:
```JavaScript
var Foo = function (Foo) {
  Foo[Foo["A"] = 1] = "A";
  Foo[Foo["B"] = 1] = "B";
  return Foo;
}(Foo || {});
```

Sucrase transformation result:
```JavaScript
var Foo; (function (Foo) {
    const A = 1; Foo[Foo["A"] = A] = "A";
    const B = A; Foo[Foo["B"] = B] = "B";
})(Foo || (Foo = {}));
```

Our transformation result:
```JavaScript
var Foo;
(Foo => {
    const A = 1;Foo[Foo['A'] = A] = 'A';
    const B = A;Foo[Foo['B'] = B] = 'B';
})(Foo ||= {});
```

All these transformation results utilize the property of variable merging.
tsc initializes `Foo` using `Foo || (Foo = {})` at the input, so `Foo` always has a value, and the internal `Foo` can directly assign the fields without a return statement.
Babel uses a return statement internally, so it can merge the initialization into the `var` declaration.
Our transformation result is more similar to Sucrase, except that we use arrow functions.

magic-akari:
After reevaluating the differences between these transformations, I agree that adding an additional return statement would be a good approach. It can help avoid introducing extra `var` declarations, ensuring that each enum strictly corresponds to a single transformation output. This way, we don't need to resize statements vector. I will give it a try.

tasks/transform_conformance/ts_fixutures/enums__enumErrors.ts
Boshen: We can dump the errors into the snapshot as well.

Boshen:
There's too many snapshot files here, can we gather them and dump then into a single file? Just how I did with all the other snapshots.

magic-akari:
I didn't encounter the same error locally as on the CI. It appears that the git submodules are out of sync, and the "just update" command didn't take effect. How can I synchronize the git submodules?

crates/oxc_transformer/src/typescript/mod.rs
magic-akari: Test cases updated. I use both of them.

Boshen:
@IWANABETHATGUY Please take a look, we can start merging if this is a in mergable state.

crates/oxc_transformer/src/typescript/mod.rs
```diff
@@ -15,6 +20,8 @@ pub struct TypeScript<'a> {
     ast: Rc<AstBuilder<'a>>,
     ctx: TransformerCtx<'a>,
     verbatim_module_syntax: bool,
+
+    export_name_set: HashSet<Atom>,
```
IWANABETHATGUY, line 24: nit: use `rustc_hash::FxHashSet` since it is faster.

tasks/transform_conformance/babel.snap.md
```diff
@@ -759,21 +759,6 @@ Passed: 269/1081
...
-* enum/inner-references/input.ts
-* enum/mix-references/input.ts
-* enum/non-foldable-constant/input.ts
-* enum/non-scoped/input.ts
-* enum/outer-references/input.ts
-* enum/scoped/input.ts
-* enum/string-value/input.ts
-* enum/string-value-template/input.ts
-* enum/string-values-computed/input.ts
-* enum/ts5.0-const-foldable/input.ts
```
IWANABETHATGUY, line 776: I saw you generated the snapshot, but excluded the test in test runner, any reason to do that?

magic-akari: We use test cases from TypeScript/Babel as our input source to generate snapshots.
However, our output format differs from TypeScript/Babel, so we have excluded it from the original testing process.

IWANABETHATGUY: ok, got it.

IWANABETHATGUY (approved):

PR #1192, opened 2023-11-08
feat(transformer/react-jsx): support the `sourceType` is a `script`

Dunqing:


Boshen (approved):
This is amazing, thank you so much!

PR #1196, opened 2023-11-08
feat(transformer/react-jsx): throw the `pragma and pragmaFrag cannot be set when runtime is automatic` error

### LINKED ISSUE #1194: feat(transformer): support throwing errors
cc @Dunqing It seems like we need to support some error printing, I'll try and add a basic infra for it later.

Dunqing:
close: #1194

Here's a rough implementation of my idea of throwing an error.


Boshen:
Oh nice to see you are already working on this! 

Can you use the same technique as how I'm doing it in the parser or semantic builder?

https://github.com/web-infra-dev/oxc/blob/7d85492a03e34fec11e27eae42db28a7c3f12e16/crates/oxc_semantic/src/builder.rs#L57C1-L58

Add this to ctx and push errors into it.

We don't want to change the visitor functions.

I think we don't need to abort early, just let it run and then abort the whole compilation pipeline when there is a error at then end, you can do this check inside the `build` function, and then make the build function return a `Result`.

Dunqing:
> Oh nice to see you are already working on this!
> 
> Can you use the same technique as how I'm doing it in the parser or semantic builder?
> 
> https://github.com/web-infra-dev/oxc/blob/7d85492a03e34fec11e27eae42db28a7c3f12e16/crates/oxc_semantic/src/builder.rs#L57C1-L58
> 
> Add this to ctx and push errors into it.
> 
> We don't want to change the visitor functions.
> 
> I think we don't need to abort early, just let it run and then abort the whole compilation pipeline when there is a error at then end, you can do this check inside the `build` function, and then make the build function return a `Result`.

OK, I will try soon

crates/oxc_transformer/src/lib.rs
```diff
@@ -44,6 +45,7 @@ pub use crate::{
 };
 
 pub struct Transformer<'a> {
+    ctx: Rc<RefCell<TransformerCtx<'a>>>,
```
Boshen, line 48: This doesn't look right, why is wrapped again when everything inside it has a `Rc` already?

Boshen (changes requested):

crates/oxc_transformer/src/lib.rs
Dunqing: If we need to place the ctx into a struct we have to do this. Because we cannot move ctx to struct(We need to pass the ctx reference to each plugin) or put the reference of ctx to struct(local reference cannot do it).

Boshen: > We need to pass the ctx reference

Everything inside it has a `Rc` so it's free to `.clone()` and pass it around.

Dunqing: > Everything inside it has a `Rc` so it's free to `.clone()` and pass it around.

Ah... Yes. Thanks for the reminder. 

After all my code changes, I forgot that I could use `.clone`. It took me a long time to come up with this solution to the problem...

Boshen (approved):

PR #1197, opened 2023-11-08
fix(tasks/transform_conformance): only the filtered test cases should be included when passing a `filter`

Dunqing:


Boshen (commented):
Then we need to skip generating the snapshot if a filter is applied, can you add this logic as well?

Dunqing:
> Then we need to skip generating the snapshot if a filter is applied, can you add this logic as well?

This logic already exists.
https://github.com/web-infra-dev/oxc/blob/main/tasks/transform_conformance/src/lib.rs#L198-L206

PR #1205, opened 2023-11-10
fix(transformer/react-jsx): no need to wrap the Array when there is only one correct child element

Dunqing:


Boshen (approved):

Boshen:
Nice work! Thank you!

PR #1207, opened 2023-11-10
feat(transformer/react-jsx): when the source type is a script, use require to import the react

Dunqing:


Boshen:
Need a quick rebase.

Boshen (approved):

PR #1214, opened 2023-11-10
fix(parser): ASI of async class member

magic-akari:


magic-akari:
Did none of the snapshots of the test cases change?

Boshen:
I added https://github.com/oxc-project/oxc/pull/1215 so we can come up with a test case that fails in oxc:

```
class Foo {
  async
  async
  foo() {}
  static
  static
  static
  bar() {}
}
```

PR #1219, opened 2023-11-11
fix(transformer/react-jsx): missing default options when plugin without config

Dunqing:


Boshen (approved):

Boshen:
It seems like you need to either you graphite with stacked PRs or I'll have to merge your PRs in the correct sequence, otherwise the snapshot will be broken on main 😅

Dunqing:
> It seems like you need to either you graphite with stacked PRs or I'll have to merge your PRs in the correct sequence, otherwise the snapshot will be broken on main 😅

Hmmm, OK, I'll look at graphite

PR #1221, opened 2023-11-11
feat(transformer/react-jsx): support throw valueless-key error

Dunqing:


Boshen (approved):
Nice work! It feels like the code is getting complicated, if you think it's time to refactor things a bit, please do so. A good reference is https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformers/jsx.ts

Issue #492, opened 2023-06-28
☂️ linter: eslint-plugin-jest recommended rules 


Boshen:

Boshen:
@mysteryven Can you setup a todo list in issue summary?

keita-hino:
I will start working on the task of `jest/require-to-throw-message`.

connorshea:
Should [jest/unbound-method](https://github.com/jest-community/eslint-plugin-jest/blob/v29.2.2/docs/rules/unbound-method.md) be marked as not-needed? 

This rule exists already via tsgolint: https://oxc.rs/docs/guide/usage/linter/rules/typescript/unbound-method.html#typescript-unbound-method

We may want/need to update the `typescript/unbound-method` rule to incorporate the additions added in the jest version of the rule, though.

MateoKruk:
going to work on `jest/valid-expect-in-promise`!

baevm:
`jest/no-unneeded-async-expect-function` was first implemented in vitest-plugin in #17494, should this rule be moved to the jest plugin and marked as compatible with vitest, or is there a way to mark it as jest compatible?

cc @connorshea @camc314

connorshea:
> `jest/no-unneeded-async-expect-function` was first implemented in vitest-plugin in [#17494](https://github.com/oxc-project/oxc/pull/17494), should this rule be moved to the jest plugin and marked as compatible with vitest, or is there a way to mark it as jest compatible?

🙃 I always knew this day would come. [The jest rule](https://github.com/jest-community/eslint-plugin-jest/commits/v29.12.1/docs/rules/no-unneeded-async-expect-function.md) was merged into the original plugin on December 14, 2025 and [the vitest rule](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-unneeded-async-expect-function.md) was merged into the original plugin December 19, 2025. The PR to oxlint was opened on December 30, and I guess we missed that this was a rule they both had.

I think it should work fine to basically do this: Open a PR to move the rule to Jest *and* add it as a vitest-jest compatible rule *and* add support for any Jest-specific details in the rule implementation.

As long as the PR does all three of those things, it should be fine.

baevm:
> 
> As long as the PR does all three of those things, it should be fine.

ok, ill make a PR then

christopher-buss:
[...]

PR #1239, opened 2023-11-12
feat(linter): eslint-plugin-jest/max_expects 

eryue0220:
Rule Detail: [max-expects](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/max-expects.md)

https://github.com/oxc-project/oxc/issues/492


crates/oxc_linter/src/rules/jest/max_expects.rs
```diff
@@ -0,0 +1,476 @@
...
+        for (name, reference_id_list) in references {
+            if name.as_str() == "expect" {
+                for reference_id in reference_id_list {
+                    let reference = symbol_table.get_reference(*reference_id);
+                    let Some(parent_node) = nodes.parent_node(reference.node_id()) else {
+                        continue;
+                    };
+
+                    let Some(grand_node) = nodes.parent_node(parent_node.id()) else {
+                        continue;
```
mysteryven, line 96: The `expect` Call Expression may have other different formats, e.g. `expect(Promise.resolve('lemon')).resolves.not.toBe('octopus')`, and we also need support test function imported from `jest/globals`. 

What about we use `collect_possible_jest_call_node` to collect all nodes, and use `is_type_of_jest_fn_call` to filter all `expect` nodes?

eryue0220: done.

mysteryven (approved):
Thank you very much!

PR #1258, opened 2023-11-13
feat(task/transformer-conformance): skip babel 8 unsupported test cases

Dunqing:
Prior to babel8, object merging would use `babelHelper.extend`, which caused most tests to fail


Boshen (approved):

Boshen:
You can add a github link to the source code next time.

PR #1295, opened 2023-11-14
fix(tasks/prettier_conformance): remove unneeded installation instructions

shannonrothe:
`tasks/prettier_conformance` is snake-case not kebab-case 😄


Boshen:
Oh it's actually no longer required ... hold on.

PR #1349, opened 2023-11-16
feat(playground): add prettier formatting + IR

shannonrothe:
- [x] add Prettier format button which takes source text and updates view with formatted source text
- [ ] add button to show Prettier IR

### DIFF
--- a/crates/oxc_wasm/src/lib.rs
+++ b/crates/oxc_wasm/src/lib.rs
@@ -14,6 +14,7 @@ use oxc::{
     transformer::{TransformOptions, TransformTarget, Transformer},
 };
 use oxc_linter::{LintContext, Linter};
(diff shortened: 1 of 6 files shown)


Boshen (approved):

crates/oxc_wasm/src/lib.rs
```diff
@@ -204,6 +212,13 @@ impl Oxc {
             self.formatted_text = printed;
         }
 
+        if run_options.prettier_format() {
+            let printed =
+                Prettier::new(&allocator, source_text, trivias, PrettierOptions::default())
```
shannonrothe, line 217: We can probably expand later and add more options to configure Prettier behaviour

Boshen: The playground needs so many knobs!

PR #1402, opened 2023-11-18
perf(regexp): remove unnecessary sort

magic-akari:


Boshen:
Nice catch! I think we can go further by storing the bits in order then print them out in sequence.

PR #1410, opened 2023-11-18
feat(linter): eslint-plugin-unicorn: no-hex-escape

RiESAEX:
[Rule Detail](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-hex-escape.md)
#684


Boshen (approved):
Thank you!

PR #1418, opened 2023-11-18
feat(linter): for-direction rule add check for condition in reverse o…

ild0tt0re:
Align eslint **for-direction** rule with the new default behaviour https://github.com/eslint/eslint/pull/17755


camc314 (approved):
Thanks!

Issue #1141, opened 2023-11-04
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y (recommended)


Boshen:

Boshen:
cc @manzt

manzt:
## Tasks:

- [ ] [accessible-emoji](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/accessible-emoji.md)
- [x] [alt-text](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/alt-text.md) https://github.com/oxc-project/oxc/pull/1126
- [ ] [anchor-ambiguous-text](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-ambiguous-text.md)
- [ ] [anchor-has-content](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-has-content.md)
- [ ] [anchor-is-valid](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-is-valid.md)
- [ ] [aria-activedescendant-has-tabindex](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-activedescendant-has-tabindex.md)
- [ ] [aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-props.md)
- [ ] [aria-proptypes](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-proptypes.md)
- [ ] [aria-role](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-role.md)
- [ ] [aria-unsupported-elements](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-unsupported-elements.md)
- [ ] [autocomplete-valid](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/autocomplete-valid.md)
- [ ] [click-events-have-key-events](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/click-events-have-key-events.md)
[...]

PR #1431, opened 2023-11-19
feat(linter): `anchor-has-content` for eslint-plugin-jsx-a11y

msdlisper:
try to implement `anchor-has-content` for #1141


crates/oxc_linter/src/snapshots/anchor_has_content.snap
```diff
@@ -0,0 +1,48 @@
+---
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 119
+expression: anchor_has_content
+---
+  ⚠ eslint-plugin-jsx-a11y(anchor-has-content): Missing accessible content when using `a` elements.
+   ╭─[anchor_has_content.tsx:1:1]
+ 1 │ <a />
+   · ─────
```
Boshen, line 9: In oxc we try to minimize the pointing span that shows the error, in this case it would be helpful to just point to `a` here.

crates/oxc_linter/src/snapshots/anchor_has_content.snap
```diff
@@ -0,0 +1,48 @@
...
+   ╭─[anchor_has_content.tsx:1:1]
+ 1 │ <a />
+   ·  ─
+   ╰────
+  help: Provide screen reader accessible content when using `a` elements.
+
+  ⚠ eslint-plugin-jsx-a11y(anchor-has-content): Missing accessible content when using `a` elements.
+   ╭─[anchor_has_content.tsx:1:1]
+ 1 │ <a><Bar aria-hidden /></a>
+   · ──────────────────────────
```
Boshen, line 16: This is still pointing to the whole element, not `a` 🤔

msdlisper: Here I consider that the code it modifiers is to remove `aria-hidden`, notice the help hint, whether it would be better to annotate the whole `a` element, or just annotate the `Bar `element🤔

Boshen: Ohh, I didn't notice the subtle difference. Thank you for the explanation!

Boshen:
Great work! Thank you!

PR #1441, opened 2023-11-20
feat(prettier): when calculating the remaining width, the docs of the remaining commands should be included

Dunqing:


Dunqing:
* main
  * **PR #1441** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1441?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1442** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1443** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1443?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1441?utm_source=stack-comment).

crates/oxc_prettier/src/printer/mod.rs
```diff
@@ -176,11 +176,14 @@ impl<'a> Printer<'a> {
 
         // TODO: these should be commands
         let mut queue: VecDeque<&Doc<'a>> = doc.iter().collect();
+        let mut cmds = self.cmds.iter().rev();
```
Boshen, line 179: I don't see similar lines in https://github.com/prettier/prettier/blob/e3fe2a1f012e5426e007510743454f436cbba9df/src/document/printer.js#L196, can you point me in the right direction?

Boshen:
Adding @camc314 for review because I didn't work on this part 😅

crates/oxc_prettier/src/printer/mod.rs
Dunqing: It should be similar to this one.

https://github.com/prettier/prettier/blob/e3fe2a1f012e5426e007510743454f436cbba9df/src/document/printer.js#L219

crates/oxc_prettier/src/printer/mod.rs
```diff
@@ -211,6 +214,37 @@ impl<'a> Printer<'a> {
             if remaining_width < 0 {
                 return false;
             }
+
+            if check_cmds && queue.is_empty() {
```
Boshen, line 218: What about this block? I can't find the relevant code in prettier 😅

crates/oxc_prettier/src/printer/mod.rs
```diff
@@ -176,11 +176,14 @@ impl<'a> Printer<'a> {
...
         // TODO: these should be commands
         let mut queue: VecDeque<&Doc<'a>> = doc.iter().collect();
+        let mut cmds = self.cmds.iter().rev();
+        let mut check_cmds = true;
 
         while let Some(next) = queue.pop_front() {
             match next {
                 Doc::Str(string) => {
                     remaining_width -= string.len() as isize;
+                    // println!("{}, {}", string, remaining_width);
```
Boshen, line 186: Some commented out code here.

crates/oxc_prettier/src/printer/mod.rs
Dunqing: I didn't look closely at prettier's code, which was written to solve this case

`* import/inline.js`

Boshen: I think it's best to port most of prettier's code, otherwise it is going to be really hard to debug things :-(

Boshen: The prettier's doc printer is based on a paper, I think it is best to port line by line 🤔 See https://prettier.io/docs/en/technical-details

Dunqing: Okay, I'll dive into prettier

Dunqing: I noticed that the original implementation doesn't match the prettier's implementation. This block aligns with prettier, but still needs to improve

Boshen: > original implementation
Which implementation?

What I hope to see is an alignment with https://github.com/prettier/prettier/blob/e3fe2a1f012e5426e007510743454f436cbba9df/src/document/printer.js#L196-L300

Dunqing: I refactored it and now the implementation is similar to prettier. The regression test seems to need to be implemented in a different way

Boshen (approved):

Boshen:
### Merge activity

* **Nov 20, 3:58 AM**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1441).

Issue #1403, opened 2023-11-18
bug(cli): need to print all the diagnostics and exit with 1 in CI


Boshen:
The current implementation can deny warnings by `oxlint --max-warnings 0`, but then it'll stop printing anything to the terminal.

ESLint has to manually configure them in a config file, which is against our zero-config philosophy.

Rust has the `-D warnings` flag for this use case in CIs. I think this should be the right approach for this problem.

So to print all the diagnostics and exit with 1 in CI environments, we should go for

```
oxlint --deny-warnings
```

hyf0:
Should also consider the priority.

For `oxlint --deny-warnings --allow-some-warning-rule`, should not throw warnings for `some-warning-rule`.

PR #1452, opened 2023-11-20
feat(cli): add `--deny-warnings` flag

shannonrothe:
See #1403

Terminates linting with an exit code of 1 if the `--deny-warnings` flag is provided _and_ there is 1 or more warnings produced from the linting.

### DIFF
--- a/crates/oxc_cli/src/command.rs
+++ b/crates/oxc_cli/src/command.rs
@@ -235,6 +235,10 @@ pub struct WarningOptions {
     #[bpaf(switch, hide_usage)]
     pub quiet: bool,
 
+    /// Ensure warnings produce a non-zero exit code
+    #[bpaf(switch, hide_usage)]
+    pub deny_warnings: bool,
+
     /// Specify a warning threshold,
     /// which can be used to force exit with an error status if there are too many warning-level rule violations in your project
     #[bpaf(argument("INT"), hide_usage)]
--- a/crates/oxc_cli/src/lint/mod.rs
+++ b/crates/oxc_cli/src/lint/mod.rs
@@ -72,7 +72,6 @@ impl Runner for LintRunner {
 
         let diagnostic_service = DiagnosticService::default()
             .with_quiet(warning_options.quiet)
-            .with_deny_warnings(warning_options.deny_warnings)
             .with_max_warnings(warning_options.max_warnings);
 
         // Spawn linting in another thread so diagnostics can be printed immediately from diagnostic_service.run.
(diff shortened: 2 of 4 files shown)


crates/oxc_cli/src/command.rs
```diff
@@ -235,6 +235,10 @@ pub struct WarningOptions {
     #[bpaf(switch, hide_usage)]
     pub quiet: bool,
 
+    /// Ensure warnings produce a non-zero exit code
+    #[bpaf(switch, hide_usage)]
+    pub deny_warnings: bool,
```
shannonrothe, line 240: Do we want `-D` as well? Might need a hand as I'm not super familiar w/ `bpaf`

crates/oxc_cli/src/lint/mod.rs
```diff
@@ -93,6 +94,7 @@ impl Runner for LintRunner {
             number_of_warnings: diagnostic_service.warnings_count(),
             number_of_errors: diagnostic_service.errors_count(),
             max_warnings_exceeded: diagnostic_service.max_warnings_exceeded(),
+            violated_deny_warnings: diagnostic_service.violated_deny_warnings(),
```
shannonrothe, line 97: Open to better names 😄

crates/oxc_diagnostics/src/service.rs
```diff
@@ -73,6 +83,10 @@ impl DiagnosticService {
         self.max_warnings.map_or(false, |max_warnings| self.warnings_count.get() > max_warnings)
     }
 
+    pub fn violated_deny_warnings(&self) -> bool {
+        self.deny_warnings && self.warnings_count.get() > 0
```
Boshen, line 87: Does diagnostic service care about this logic 🤔 Does it make sense to move this logic to `CliResult`? Given `CliResult` also has access to `number_of_warnings` and `deny_warwnings`.

crates/oxc_cli/src/command.rs
Boshen: `-D` is not required, that's a `rustc` globaly flag thingy. We only need a plain cli flag for this.

crates/oxc_diagnostics/src/service.rs
shannonrothe: Yeah good call – I was just following what we do with `quiet` but now I see that's useful within the service.

shannonrothe: Updated in https://github.com/oxc-project/oxc/pull/1452/commits/29134b9b53836b450d8f5320f473e8d080e7e3d2

Boshen (approved):

crates/oxc_cli/src/command.rs
pacak: > Might need a hand as I'm not super familiar w/ bpaf

To add `-D` you can write something like  `#[bpaf(long, short('D'), switch, hide_usage)]`. Any named item (flags and arguments) takes a combination of `short` and `long`, with name being long by default.

PR #1458, opened 2023-11-20
feat(transformer): add transform property-literal plugin

IWANABETHATGUY:
1. Add `transform-property-literal-plugin`
2. Passing 2 testcases.

### DIFF
--- a/crates/oxc_transformer/src/utils.rs
+++ b/crates/oxc_transformer/src/utils.rs
@@ -52,7 +52,7 @@ pub trait CreateVars<'a> {
     }
 }
 
-pub const RESERVED_WORDS_ES3_ONLY: [&str; 24] = [
+pub const RESERVED_WORDS_ES3_ONLY: phf::Set<&str> = phf::phf_set![
     "abstract",
     "boolean",
     "byte",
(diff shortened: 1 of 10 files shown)


mysteryven:
You can add this plugin name to [CASES](https://github.com/web-infra-dev/oxc/blob/d8864c15b03d1c7819876868862308da32d1f528/tasks/transform_conformance/src/lib.rs#L43). 😀

crates/oxc_transformer/src/utils.rs
```diff
@@ -50,3 +51,120 @@ pub trait CreateVars<'a> {
...
+pub fn is_valid_identifier(name: &Atom, reserved: bool) -> bool {
+    if reserved && (KEYWORDS.contains(&name.as_str()) || is_strict_reserved_word(name, true)) {
+        return false;
+    }
+    is_identifier_name(name)
+}
+
+pub fn is_strict_reserved_word(name: &Atom, in_module: bool) -> bool {
+    is_reserved_word(name, in_module) || RESERVED_WORD_STRICT.contains(&name.as_str())
+}
```
mysteryven, line 161: Since it will check every identifier in the object key, is it will be better to use `Set` here?

crates/oxc_transformer/src/utils.rs
```diff
@@ -50,3 +51,120 @@ pub trait CreateVars<'a> {
...
+    "let",
+    "package",
+    "private",
+    "protected",
+    "public",
+    "static",
+    "yield",
+];
+
+pub const KEYWORKDS: phf::Set<&str> = phf::phf_set![
```
mysteryven, line 94: Just a small typo
```suggestion
pub const KEYWORDS: phf::Set<&str> = phf::phf_set![
```

Boshen (approved):

PR #1460, opened 2023-11-20
feat(prettier): init command of `Fill`

mysteryven:


mysteryven:
* main
  * **PR #1460** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1460?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1460?utm_source=stack-comment).

crates/oxc_prettier/src/printer/mod.rs
```diff
@@ -60,6 +60,8 @@ impl<'a> Printer<'a> {
                 Doc::Softline => self.handle_softline(indent, mode),
                 Doc::Hardline => self.handle_hardline(indent),
                 Doc::IfBreak(doc) => self.handle_if_break(doc.unbox(), indent, mode),
+                // TODO: need refactor `fits` to accept `cmd` instead of `doc`.
+                Doc::Fill(_) => {}
```
mysteryven, line 64: Seems we need to accept `cmd` by parameters in `fits`🤔. @Dunqing 

https://github.com/prettier/prettier/blob/eebf0e4b5ec8ac24393c56ced4b4819d4c551f31/src/document/printer.js#L470-L477

Dunqing: In prettier's fits implementation, only mode and doc are used. so we don't need cmd. i think that's enough!

Boshen:
I don't understand part yet so its up to @Dunqing to decided whether we merge or not.

Boshen:
I don't understand this part yet so I'll let @Dunqing decided whether we want to merge or not.

Boshen:
Which test are you targeting this PR with? This PR is getting big 😅

mysteryven:
> Which test are you targeting this PR with? This PR is getting big 😅

I will add test tonight, 😀 Hope everything works!

mysteryven:
The last Doc is `ifBreak(",", "", { groupId: Symbol.for("array") })`, I need to port more `if_break`'s logic first for pass `arrays/numbers1.js`, but I think the fit command looks good now.

CleanShot 2023-11-25 at 10 39 38@2x

crates/oxc_prettier/src/printer/mod.rs
mysteryven: I changed to accept `Command` finally for `handle_fill`. 😥

PR #1468, opened 2023-11-20
Update to ezno-checker 0.0.12

kaleidawave:
Work in progress.

Updates the ezno checker integration from the June release to now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -694,9 +694,9 @@ dependencies = [
 
 [[package]]
 name = "ezno-checker"
-version = "0.0.4"
+version = "0.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ce89c36282d2e7e84de91cca6c2c32ba11f4839ae8c73fb7cc0024411ab28521"
+checksum = "8f248ba818d385204ff072d04f3c9c70b2f617550da17d0fb5772a569f34abef"
 dependencies = [
  "bimap",
  "binary-serialize-derive",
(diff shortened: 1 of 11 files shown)


kaleidawave:
This branch is still is a way to go. I thought I could do it this evening but having some problems with lifetimes here:

https://github.com/oxc-project/oxc/blob/499922e6818aa03ee38f7bf94dbfcaf19b4b95b1/crates/oxc_type_synthesis/src/lib.rs#L24-L32

Will work on creating a MRE. The current synthesis API is built for the non-lifetime based AST of my own parser. As all Oxc AST has a lifetime I need to figure out how introduce the lifetimes into the existing APIs.

---

Once that has been fixed, I still need to fix up all the current `todo!()`s and add use the new API stuff

Boshen:
> This branch is still is a way to go. I thought I could do it this evening but having some problems with lifetimes here:
> 
> https://github.com/oxc-project/oxc/blob/499922e6818aa03ee38f7bf94dbfcaf19b4b95b1/crates/oxc_type_synthesis/src/lib.rs#L24-L32
> 
> Will work on creating a MRE. The current synthesis API is built for the non-lifetime based AST of my own parser. As all Oxc AST has a lifetime I need to figure out how introduce the lifetimes into the existing APIs.
> 
> Once that has been fixed, I still need to fix up all the current `todo!()`s and add use the new API stuff

Are there any unsafe Rust hackery that could ease the pain?

kaleidawave:
Don't think so, it is a issue with Rust traits, impls and types. Looks like a change needs to be made to  [ezno-checker](https://github.com/kaleidawave/ezno), I think all the AST associated types (apart from `OwnedModule`) need a lifetime parameter. Should compile after then, but means that existing items have to include explicit lifetimes (unfortunately doesn't look like Rust lifetime elision works here)

Here is a example of the fix: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=47a251281f532f35cd231070a548bfdb

Boshen:
Closing due to outdated.

PR #1477, opened 2023-11-21
feat(linter): `anchor-is-valid` for eslint-plugin-jsx-a11y

msdlisper:
try to implement `anchor-is-valid` for #1141


msdlisper:
Currently, I have implemented the base; the remaining work will be done when polymorphic components are supported.

crates/oxc_linter/src/rules/jsx_a11y/anchor_is_valid.rs
```diff
@@ -0,0 +1,682 @@
...
+                    if str_lit.value.is_empty() {
+                        return true;
+                    }
+                    if str_lit.value == "#" {
+                        return true;
+                    }
+                    if str_lit.value == "javascript:void(0)" {
+                        return true;
+                    }
+                    return false;
```
IWANABETHATGUY, line 142:
```suggestion
                  return str_lit.value.is_empty() || str_lit.value == "#" || str_lit.value == "javascript:void(0)"
```

crates/oxc_linter/src/snapshots/anchor_is_valid.snap
```diff
@@ -0,0 +1,89 @@
...
+   ╭─[anchor_is_valid.tsx:1:1]
+ 1 │ <a href={null} />
+   ·  ─
+   ╰────
+  help: Provide a correct href for the `a` element.
+
+  × Unterminated string
+   ╭─[anchor_is_valid.tsx:1:1]
+ 1 │ <a href=' />;
+   ·         ─────
```
Boshen, line 30: Is this a broken test or indented?

msdlisper: I had noticed this,  I found that OXC generates this AST in this case. So, I thought it should not hit this rule🧐
```
{
  "type": "Program",
  "start": 0,
  "end": 0,
  "sourceType": {
    "language": {
      "typeScript": {
        "isDefinitionFile": false
      }
    },
    "moduleKind": "module",
    "variant": "jsx",
    "alwaysStrict": false
  },
  "directives": [],
  "hashbang": null,
  "body": []
}
```

Boshen: So is this intended, we need a comment explaining it if this is the case.

msdlisper: This is not intentional, so have we commented out this case and made remarks?

Boshen: I just checked the source, this is actually intended to be parse error.

PR #1490, opened 2023-11-21
feat(linter): eslint-plugin-unicorn numeric-separators-style (style)

sirreal:
Part of #684.

Add eslint unicorn's [numeric-separators-style](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/numeric-separators-style.md) rule.

Todo:
- [x] Implement decimal exponent number linting/formatting
- [x] Add more test cases
- [x] Add test cases for configuration
- [x] Add configurationa
- [x] Address performace issues


crates/oxc_linter/src/snapshots/numeric_separators_style.snap.new
Boshen, line 1: Don't forget to `cargo insta accept`

Boshen:
Good work!

It seems like there are some serious performance regressions here, please let us know if you need a little help.

sirreal:
I think this PR is ready from a _correctness_ perspective. I believe the rule is behaving as expected. However [as noted](https://github.com/oxc-project/oxc/pull/1490#issuecomment-1822014131) and flagged by the benchmark CI, there are some unacceptable performance regressions.

I'm not an expert Rust programmer and am new to this project, so I could use some help. I'm not sure how best to profile the benchmarks so that I can identify the problems and optimize.

My intuition is that building the formatted numeric strings for every numeric or bigint literal is probably the source of the regression, but I'd like to confirm that before attempting to fix it. From a simplicity perspective, it's appealing to build the formatted string and compare with the original to know if its valid. I have some ideas for improvement there, but if someone with more experience has advice, again it would be appreciated.

Thanks in advance 🙇

sirreal:
I've addressed the performance regressions 🎉  This is now ready for review. 

For posterity, I installed [cargo-flamegraph](https://github.com/flamegraph-rs/flamegraph) and ran (roughly) the following:

```sh
cargo flamegraph --dev --bench linter -p oxc_benchmark
```

This helped to identify that the vast majority of the time was spent in `format_decimal` and specifically creating a `Regex`:

Screenshot 2023-11-29 at 21 41 39

From there, it was a quick fix to use `lazy_static` to create the regex once and reuse it.

crates/oxc_linter/src/rules/unicorn/numeric_separators_style.rs
```diff
@@ -0,0 +1,667 @@
...
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use regex::Regex;
+
+use crate::{context::LintContext, fixer::Fix, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint-plugin-unicorn(numeric-separators-style): Enforce the style of numeric separators by correctly grouping digits")]
+#[diagnostic(severity(error), help("Invalid group length in numeric value."))]
```
Boshen, line 18: Should be `severity(warning)`, all lints are warnings.

Boshen: It feels like the error message should be "Invalid group length in numeric value." and the help message be something like "Long numbers can become really hard to read, so cutting it into groups of digits, separated with a _, " (from their README)

crates/oxc_linter/src/rules/unicorn/numeric_separators_style.rs
```diff
@@ -0,0 +1,667 @@
...
+        }
+        SeparatorDir::Left => {
+            let mut pos = config.group_length;
+            while pos < s.len() {
+                s.insert(pos, '_');
+                pos += config.group_length + 1;
+            }
+        }
+    }
+}
```
Boshen, line 83: Shall we move all these below `impl Rule for NumericSeparatorsStyle {` so that `declare_oxc_lint` is at the top?

crates/oxc_linter/src/rules/unicorn/numeric_separators_style.rs
```diff
@@ -0,0 +1,667 @@
...
+
+impl NumericSeparatorsStyle {
+    //
+    // Numbers
+    //
+    fn format_number(&self, number: &NumberLiteral) -> String {
+        match number.base {
+            oxc_syntax::NumberBase::Binary => self.format_binary(number.raw),
```
Boshen, line 185: nit: you can add `use oxc_syntax::NumberBase`; to the top of this function.

crates/oxc_linter/src/rules/unicorn/numeric_separators_style.rs
sirreal: Severity is now `warning`.

I updated the error and help messages, let me know if you'd prefer something else. I tried to find a concise, clear, and descriptive help text.

Error (this is from unicorn's error message):

> Invalid group length in numeric value.

Help:

> Group digits with numeric separators (_) so longer numbers are easier to read.

Boshen:
Thank you for working on this!

PR #1492, opened 2023-11-22
feat(task/prettier-confromance): ignore tests with non-standard syntax

Dunqing:

PR #1493, opened 2023-11-22
feat(task/prettier-confromance): escape template literal

Dunqing:


Dunqing:
* main
  * **PR #1491** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1491?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1492** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1493** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1493?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1493?utm_source=stack-comment).

tasks/prettier_conformance/src/lib.rs
```diff
@@ -238,7 +238,8 @@ impl TestRunner {
             if snapshot_line.is_empty() { String::new() } else { title_snapshot_options }
         );
 
-        let output = Self::prettier(path, input, prettier_options);
+        let output = Self::prettier(path, input, prettier_options).replace('`', "\\`");
```
Boshen, line 241: This is confusing, I'll take a look at the root cause.

Dunqing: The root cause is that the test snapshot is wrapped by `. So the template literal needs to escape
```
exports[`call.js format 1`] = ` //  you can see end of this line 
xxxxxxx
xxxxxx
xxxxxx
` // and this
```

Boshen: Does it make sense to extract these from the AST instead of doing a catch all `file.contains(output)`? Although this will require a lot of work :-(

Boshen: It should be easy to extract all the exported strings and compared them. 

We can then eventually do a `exports[file]` inside our test runner get all the strings.

Dunqing: > Does it make sense to extract these from the AST instead of doing a catch all file.contains(output)? Although this will require a lot of work :-(

But that's something we need to escape anyway

Boshen: I see, let's merge. We can always improve this part later.

Boshen:
### Merge activity

* **Nov 22, 2:31 AM**: @Boshen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1493).

PR #1496, opened 2023-11-22
refactor(parser) Updated comments mentioning the ecma specification section 12.x

JonaAnders:
The ECMA specification seems to added the "Tokens" section to the specification as 12.6. This pushed all the other sections down, resulting in e.g. former 12.6 now being 12.7. Comments in the parser mention this part of the specification. All the mentions of section 12.6+ therefor are outdated now. This pull request tries to fix that by updating all the comments.


IWANABETHATGUY (approved):
Thanks

Issue #1164, opened 2023-11-06
feat(linter): regex parser


Ubugeeei:
Implement a regular expression parser equivalent to [@eslint-community/regexpp](https://github.com/eslint-community/regexpp).

Todo:
- [ ] Consider how to handle [Maneren's branch](https://github.com/Maneren/oxc/tree/regex)
  - [ ] Check if any part can be integrated
  - [ ] See if the remaining issues can be shared in this repository
- [ ] Document the rules requiring the implementation of the parser as pending

Ref: https://github.com/web-infra-dev/oxc/issues/611

Boshen:
Let's do this incrementally with proper testing, we can cherry-pick some of the code from [Maneren's branch](https://github.com/Maneren/oxc/tree/regex)

My requirements: 

1. 100% test coverage
2. use the same infra from `oxc_parser`, i.e. allocate the AST on `oxc_allocator` for maximum performance

Ubugeeei:
## Roadmap (Draft) (For Contributors)

- [ ] Definition of AST
  - [ ] Understanding of oxc infrastructure
    - [ ] Understanding of oxc_allocator
    - [ ] Understanding of oxc_span
  - [ ] Writing in Rust
    Refer to https://github.com/eslint-community/regexpp/blob/2e8f1af992fb12eae46a446253e8fa3f6cede92a/src/ast.ts
- [ ] Lexer (?)
  - [ ] Reading and understanding points to note (mainly performance requirements) from https://oxc-project.github.io/javascript-parser-in-rust/en/docs/lexer
  - [ ] Enumeration of Tokens
  - [ ] Implementation of Lexer (If possible, I would like to submit small PRs by dividing)
    - [ ] Writing tests (?)
- [ ] Parser, Validator
  - [ ] Implementing incrementally by narrowing down the target AST
    ※ Tasks will be added as needed
    - [ ] Understanding test262 https://github.com/tc39/test262/tree/main/test/language/literals/regexp
    - [ ] Implementation of parser and test
- [ ] Visitor

Boshen:
@Ubugeeei Nice work!

IWANABETHATGUY:
Hello, I saw it has been 3 weeks since your last pr, are you still interested in this feature, or can I continue your work?

Boshen:
@Ubugeeei is too busy with other stuff (vapor mode I guess?). Feel free to assign this to yourself @IWANABETHATGUY

Ubugeeei:
[...]

PR #1500, opened 2023-11-22
Initialize JS Regex crates and def AST.

ubugeeei:
ref: https://github.com/oxc-project/oxc/issues/1164

I have initialized a crate for handling JavaScript Regexp and defined the AST.
I implemented the AST while referring to [eslint-community/regexpp](https://github.com/eslint-community/regexpp/blob/2e8f1af992fb12eae46a446253e8fa3f6cede92a/src/ast.ts).

Review Points:

- Is the crate name appropriate?
- Are the crate settings appropriate? (I'm not very sure what should be done.)
- Is the method of defining the AST correct? (Usage of Vex and Box)
- Is the AST definition accurate? (It might be hard to tell if the information is sufficient without implementing a parser or visitor.)
  - Is the documentation appropriate?

---

This is my first PR for Rust code to OXC! Please check it out!
Thank you always for your kindness. It's really helpful. 🙏🏻


crates/oxc_js_regex/src/ast.rs
```diff
@@ -0,0 +1,387 @@
...
+#[derive(Debug)]
+pub struct Backreference<'a> {
+    pub span: Span,
+    pub reference: BackreferenceRef,
+    pub resolved: CapturingGroup<'a>,
+}
+
+/// The flags.
+#[derive(Debug)]
+pub struct Flags {
```
IWANABETHATGUY, line 377: `bitflags` is a better choice here, since it could save more space, you could refer https://github.com/oxc-project/oxc/blob/c934d93713b5b5fd577d704a9e5063518e546719/crates/oxc_semantic/src/node.rs#L13-L21 for more details.

Ubugeeei: Thank you! I didn't know that (I should have looked at other implementations a bit more).

---

While I was exploring how to use bit flags, I accidentally stumbled upon a similar implementation for flags in the JS AST. 

So, it's okay to disregard this and go ahead with a unique implementation in js_regex, right? 
(Seems like the only option?)

The crate I've created is intended to provide a more detailed Regex AST and parser.

https://github.com/oxc-project/oxc/blob/c934d93713b5b5fd577d704a9e5063518e546719/crates/oxc_ast/src/ast/literal.rs#L142-L155

crates/oxc_js_regex/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+categories.workspace   = true
+description.workspace  = true
+edition.workspace      = true
+homepage.workspace     = true
+keywords.workspace     = true
+license.workspace      = true
+repository.workspace   = true
+rust-version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Boshen, line 15: Let's remove this comment and add:

```
[lints]
workspace = true

[lib]
doctest = false
```

crates/oxc_js_regex/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name                   = "oxc_js_regex"
+version                = "0.0.0"
+publish                = false
+authors.workspace      = true
```
Boshen, line 5: You may add your name here 😁

Boshen (commented):
Re review points: yes to all questions!

Let's start merging after [lints] get added.

crates/oxc_js_regex/src/ast.rs
Ubugeeei: @Boshen

Boshen: We can change this later.

Boshen:
Nice work! You hands are getting Rusty ;-)

PR #1502, opened 2023-11-22
feat(transformer): 🐛 react jsx invalid runtime

IWANABETHATGUY:
1. Passed another `react_jsx` babel transformer test case

### DIFF
--- a/crates/oxc_transformer/src/react_jsx/mod.rs
+++ b/crates/oxc_transformer/src/react_jsx/mod.rs
@@ -60,6 +60,7 @@ pub struct ReactJsx<'a> {
     require_jsx_runtime: bool,
     jsx_runtime_importer: Atom,
     pub babel_8_breaking: Option<bool>,
+    default_runtime: ReactJsxRuntime,
 }
 
 enum JSXElementOrFragment<'a, 'b> {
@@ -120,8 +121,26 @@ impl<'a> ReactJsx<'a> {
             }
         }
 
+        let default_runtime = match jsx_options.runtime.as_ref() {
+            Some(runtime) if runtime == "classic" => ReactJsxRuntime::Classic,
+            Some(runtime) if runtime == "automatic" => ReactJsxRuntime::Automatic,
+            None => {
+                if options.babel_8_breaking == Some(true) {
+                    ReactJsxRuntime::Automatic
+                } else {
+                    ReactJsxRuntime::Classic
+                }
+            }
+            Some(_) => {
+                ctx.error(miette::Error::msg(
+                    "Runtime must be either \"classic\" or \"automatic\".",
+                ));
+                return None;
+            }
+        };
+
         let jsx_runtime_importer =
-            if jsx_options.import_source == "react" || jsx_options.runtime.is_classic() {
+            if jsx_options.import_source == "react" || default_runtime.is_classic() {
                 Atom::new_inline("react/jsx-runtime")
             } else {
                 Atom::from(format!("{}/jsx-runtime", jsx_options.import_source))
@@ -138,6 +157,7 @@ impl<'a> ReactJsx<'a> {
             import_fragment: false,
             import_create_element: false,
             babel_8_breaking: options.babel_8_breaking,
+            default_runtime,
         })
     }
 
@@ -154,7 +174,7 @@ impl<'a> ReactJsx<'a> {
     }
 
     pub fn add_react_jsx_runtime_imports(&mut self, program: &mut Program<'a>) {
-        if self.options.runtime.is_classic() {
+        if self.default_runtime.is_classic() {
             if self.options.import_source != "react" {
                 self.ctx.error(ImportSourceCannotBeSet);
             }
@@ -187,7 +207,7 @@ impl<'a> ReactJsx<'a> {
         has_key_after_props_spread: bool,
         need_jsxs: bool,
     ) {
-        if self.options.runtime.is_classic() {
+        if self.default_runtime.is_classic() {
             return;
         }
         match e {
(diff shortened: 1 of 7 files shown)


crates/oxc_transformer/src/react_jsx/options.rs
```diff
@@ -7,8 +7,7 @@ use serde::Deserialize;
 #[serde(rename_all = "camelCase")]
 pub struct ReactJsxOptions {
     /// Decides which runtime to use.
-    #[serde(default)]
-    pub runtime: ReactJsxRuntime,
+    pub runtime: Option<String>,
```
IWANABETHATGUY, line 10: We can't use a specific enum type here, because if deserialized fails, `serde` will fallback to the default value, the runtime/invalid-runtime/input.js test case requires emitting error when runtime neither `automatic` nor `classic`

Dunqing:
It looks like we don't need to support this error, we can mention in the documentation that if the runtime is invalid, it will be forced to change to the default runtime (automatic). What do you think?

crates/oxc_transformer/src/react_jsx/options.rs
Boshen: `ReactJsxOptions` is user facing, we shouldn't provide vague values such as `Option<String>`, we should implement the `std::str::FromStr` trait instead.

IWANABETHATGUY:
> It looks like we don't need to support this error, we can mention in the documentation that if the runtime is invalid, it will be forced to change to the default runtime (automatic). What do you think?

Not sure if this is a good idea, what if the user wants to use `classic` runtime, but uses the typo `classica` accidentaly, then the output would be problematic.

Dunqing:
> Not sure if this is a good idea, what if the user wants to use `classic` runtime, but uses the typo `classica` accidentaly, then the output would be problematic.

The problem is obvious, so I don't think we need to worry about it

Boshen:
> > Not sure if this is a good idea, what if the user wants to use `classic` runtime, but uses the typo `classica` accidentaly, then the output would be problematic.
> 
> The problem is obvious, so I don't think we need to worry about it

For maintainability point of view, it's probably better to align with babel and also produce meaningful errors.

crates/oxc_transformer/src/react_jsx/options.rs
```diff
@@ -56,7 +55,7 @@ fn default_pragma_frag() -> Cow<'static, str> {
 impl Default for ReactJsxOptions {
     fn default() -> Self {
         Self {
-            runtime: ReactJsxRuntime::Automatic,
+            runtime: None,
```
IWANABETHATGUY, line 58: Use `Option` instead of `ReactJsxRuntime`, it is used to fallback default runtime according to the related scenario.

crates/oxc_transformer/src/react_jsx/options.rs
```diff
@@ -67,13 +66,21 @@ impl Default for ReactJsxOptions {
     }
 }
 
-#[derive(Debug, Default, Clone, Copy, Deserialize)]
+#[derive(Deserialize, Debug, Clone)]
```
IWANABETHATGUY, line 69: The idea was stolen from https://github.com/serde-rs/serde/issues/912#issuecomment-1707349708, we want to catch four runtime input scenario, 
1. a valid input `automatic`
2. a valid input `classic`
3. None runtime.
4. wrong runtime that is neither `automatic` nor `classic`

Dunqing:
Is it possible to get the original value during serde parsing and warn if it's invalid so we don't have to change it too much? It sounds like what's being said [here](https://github.com/oxc-project/oxc/pull/1502#discussion_r1402870717)

IWANABETHATGUY:
> Is it possible to get the original value during serde parsing and warn if it's invalid so we don't have to change it too much? It sounds like what's being said [here](https://github.com/oxc-project/oxc/pull/1502#discussion_r1402870717)

Yeah, I don't change the **runtime** if it is not valid, just report a error, just like babel.

crates/oxc_transformer/src/react_jsx/options.rs
```diff
@@ -67,13 +66,21 @@ impl Default for ReactJsxOptions {
...
+#[serde(untagged)]
+pub enum ReactJsxRuntimeOption {
+    Valid(ReactJsxRuntime),
+    // The order matters. The most permissive variant (i.e. the catch-all)
+    // should be tried last during deserialization.
+    Unknown(String),
+}
```
Boshen, line 76: What about adding a `Default` tag here instead of a confusing `Option<ReactJsxRuntimeOption>`?

crates/oxc_transformer/src/react_jsx/mod.rs
```diff
@@ -120,8 +121,26 @@ impl<'a> ReactJsx<'a> {
...
+                    ReactJsxRuntime::Classic
+                }
+            }
+            Some(_) => {
+                ctx.error(miette::Error::msg(
+                    "Runtime must be either \"classic\" or \"automatic\".",
+                ));
+                return None;
+            }
+        };
```
Boshen, line 140: This block is big, can we refactor (move this out) in a follow up PR?

Boshen (changes requested):
Shall we address the two issues above and start merging?

crates/oxc_transformer/src/react_jsx/options.rs
IWANABETHATGUY: Do you mean 
```rs
pub struct ReactJsxOptions {
  ...
   runtime: Option<ReactJsxRuntime>
}
```
This data structure did work, cause any invalid or empty input of runtime will be deserialized to `None`, 
we need to figure out if the user passes an invalid runtime, and give some waring to pass a babel snapshot, and fallback to a default runtime if `runtime` is empty.

Boshen (approved):

PR #1510, opened 2023-11-23
Start on `function_name` transform.

milesj:


.gitmodules
```diff
@@ -17,4 +17,4 @@
 	path = tasks/prettier_conformance/prettier
 	url = git@github.com:prettier/prettier.git
 	shallow = true
-	branch = e8ac9f8d2a1c7083ac4f8a9b960521b62d60c94f #v3.1.0
```
milesj, line 20: this hash wasnt working

mysteryven: It has updated to 'main' now.

crates/oxc_transformer/src/es2015/function_name.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // Internal only
+    fn transform_expression<'b>(&mut self, expr: &mut Expression<'a>, mut id: BindingIdentifier) {
+        // function () {} -> function name() {}
+        if let Expression::FunctionExpression(func) = expr {
+            let scopes = self.ctx.scopes();
+            let mut count = 0;
+
+            // let mut finder = IdentFinder { id, found: 0 };
```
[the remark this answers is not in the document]
Boshen: Took me a while to figure out what's going on 😅 

For your particular problem, we cannot pass in `mut id: BindingIdentifier` - if you need to change the identifier, you need to replace the expression entirely. 

e.g. 

https://github.com/oxc-project/oxc/blob/6ec257ba2d9aad94930b3e8c44e17f2a6d0400a7/crates/oxc_transformer/src/es2021/logical_assignment_operators.rs#L94C23-L94C23

In your case, have a function returns a new `BindingIdentifier` and replace `VariableDeclarator`.id with it.

----

As for scope descendants, It seems like `scopes.descendants()` is for walking the entire scope tree, its not walking from the current scope. We may need to add more features to the scope tree in order to make this work. The scope tree is currently a parent pointing tree, it doesn't have any child node information yet.

Boshen:
It would be nice if we can merge most of the boilerplate first, so we don't hang onto a huge PR.

milesj:
@Boshen Agreed, let me polish this PR up, I'll "disable it from transforming", and we can land it and figure out next steps.

The big problem I keep running into, is how to traverse parent/children relationships, _and mutate them_. I've been looking at both babel and swc, and they do some crazy traversal stuff for some of these transforms. As a simple case, the `arrow_functions` transform mutates the parent by inserting `let _this = this` declarations, so that scoping within the arrow function isn't broken. There are far crazier patterns than that.

The only way I've gotten even close to supporting this, is by starting at the parent and mutating down, but this runs this risk of not transforming all possible use cases.

I think before we can move forward on a lot of these transforms, we need to figure out how we want to approach this. I've dabble a bit, but I still don't have the context into oxc like you do.

Issue #1506, opened 2023-11-22
feat(oxc_semantic): jsdoc for other AstTypes


etkachev:
Hello! Was wondering if the jsdoc builder can also register jsdocs for docs that are included on properties and methods on classes for example.

Looking through the [code](https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/src/jsdoc/builder.rs#L26), it looks like it only registers those that are declarations only. Would be nice to be able to extract jsdoc for things like this:

```typescript
@Component({
   selector: 'app-card-section',
})
export class CardSectionComponent {
  /**
   * text to display in top header
   */
  @Input() headerText: string;
  /**
   * secondary text
   */
  @Input() subText: string;
  /**
   * whether to show divider between sections
   */
  @Input() showDivider = true;

  /**
   * expand the card accordion section
   */
  expand() {
    // TODO
  }
}
```

shannonrothe:
These would be considered decorators not jsdoc annotations right @Boshen?

Boshen:
The current jsdoc build is incomplete, we welcome PRs that can expand its functionalities.

@etkachev Are you looking into oxc for building something new in Rust?

Boshen:
> These would be considered decorators not jsdoc annotations right @Boshen?

I don't get your questions 🤔 

But I looked up comment attachment in AST explorer and have confirmed that

`@Input() headerText: string;` should have the jsdoc comment `/**\n* text to display in top header\n*/`

shannonrothe:
Ah sorry, I thought they were referring to parsing the decorators on those class fields as jsdoc annotations!

I can take a look at handling more cases with jsdoc comments soon 👍🏻

etkachev:
@Boshen @shannonrothe  I saw that you guys merged some updates to the crates. Is there a timeline on how often you guys push out updates for these?

Boshen:
I've published the latest changes as v0.4.0 to crates.io

PR #1517, opened 2023-11-23
feat(ast/semantic): parse jsdoc on `PropertyDefinition`

shannonrothe:
This should be enough to handle jsdoc comments on class properties/fields.

See #1506

### DIFF
--- a/crates/oxc_ast/src/ast_kind.rs
+++ b/crates/oxc_ast/src/ast_kind.rs
@@ -183,7 +183,7 @@ impl<'a> AstKind<'a> {
         || matches!(self, Self::Class(class) if class.is_declaration())
         || matches!(self, Self::ModuleDeclaration(_) | Self::TSEnumDeclaration(_) | Self::TSModuleDeclaration(_)
             | Self::VariableDeclaration(_) | Self::TSInterfaceDeclaration(_)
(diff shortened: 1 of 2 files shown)


crates/oxc_ast/src/ast_kind.rs
```diff
@@ -183,7 +183,7 @@ impl<'a> AstKind<'a> {
         || matches!(self, Self::Class(class) if class.is_declaration())
         || matches!(self, Self::ModuleDeclaration(_) | Self::TSEnumDeclaration(_) | Self::TSModuleDeclaration(_)
             | Self::VariableDeclaration(_) | Self::TSInterfaceDeclaration(_)
-            | Self::TSTypeAliasDeclaration(_) | Self::TSImportEqualsDeclaration(_)
+            | Self::TSTypeAliasDeclaration(_) | Self::TSImportEqualsDeclaration(_) | Self::PropertyDefinition(_)
```
shannonrothe, line 186: Not 100% sure if this should be more specific, where else can `PropertyDefinition` exist? 😅

Boshen: You can quickly search it through `crates/oxc_ast/src/ast/js.rs` 😅

PR #1525, opened 2023-11-23
fix(oxc_vscode): vscode extension - check on file change (not on file save) 

### LINKED ISSUE #1518: vscode extension - check on file change (not on file save)
Currently, the extensions run a check each time a file is saved. Instead, it would be useful if it will each on file change (like ESLint does)

IWANABETHATGUY:
1. Closed https://github.com/oxc-project/oxc/issues/1518

[录屏 2023年11月24日 12时08分14秒.webm](https://github.com/oxc-project/oxc/assets/17974631/6ff42edf-b837-466a-a9fa-a1d1244dfd45)

### DIFF
--- a/editors/vscode/pnpm-lock.yaml
+++ b/editors/vscode/pnpm-lock.yaml
@@ -1,7 +1,7 @@
 lockfileVersion: '6.0'
 
 settings:
-  autoInstallPeers: true
+  autoInstallPeers: false
   excludeLinksFromLockfile: false
 
 dependencies:
--- a/editors/vscode/pnpm-lock.yaml
+++ b/editors/vscode/pnpm-lock.yaml
@@ -1,7 +1,7 @@
 lockfileVersion: '6.0'
 
 settings:
-  autoInstallPeers: false
+  autoInstallPeers: true
   excludeLinksFromLockfile: false
 
 dependencies:
--- a/editors/vscode/server/src/linter.rs
+++ b/editors/vscode/server/src/linter.rs
@@ -155,7 +155,11 @@ impl IsolatedLintHandler {
         Self::process_diagnostics(&rx_error)
     }
 
-    pub fn run_single(&self, path: &Path, content: Option<String>) -> Option<Vec<DiagnosticReport>> {
+    pub fn run_single(
+        &self,
+        path: &Path,
+        content: Option<String>,
+    ) -> Option<Vec<DiagnosticReport>> {
         if Self::is_wanted_ext(path) {
             Some(
                 Self::lint_path(&self.linter, path, Arc::clone(&self.plugin), content).map_or(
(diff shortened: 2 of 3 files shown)


.vscode/launch.json
```diff
@@ -10,7 +10,6 @@
 			"args": ["--extensionDevelopmentPath=${workspaceFolder}/editors/vscode" ],
 			"sourceMaps": true,
 			"outFiles": ["${workspaceFolder}/editors/vscode/dist/*.js"],
-			"preLaunchTask": "watch",
```
IWANABETHATGUY, line 13: https://github.com/microsoft/vscode/issues/180641, vscode can't prelaunch a esbuild-watch task

editors/vscode/package.json
```diff
@@ -50,12 +50,35 @@
         "title": "Show Trace Output Channel",
         "category": "Oxc"
       }
-    ]
+    ],
+    "configuration": {
```
IWANABETHATGUY, line 54: this configuration to get traces from lsp, e.g: 
image

editors/vscode/client/extension.ts
```diff
@@ -70,7 +69,8 @@ export async function activate(context: ExtensionContext) {
   const traceOutputChannel = window.createOutputChannel(traceOutputChannelName);
 
   const ext = process.platform === "win32" ? ".exe" : "";
-  const command = join(context.extensionPath, `./target/release/oxc_vscode${ext}`);
+  let serverPath = process.env.SERVER_PATH_DEV ?? `oxc_vscode${ext}`;
```
IWANABETHATGUY, line 72: We should have a language server path config to let user customize, this could be improve in next pr.

editors/vscode/pnpm-lock.yaml
```diff
@@ -1,7 +1,7 @@
 lockfileVersion: '6.0'
 
 settings:
-  autoInstallPeers: true
+  autoInstallPeers: false
```
Boshen, line 4: How is this being changed? I'm not seeing any other config changes related to this.

Boshen (commented):
This PR is mixing too many different changes making it hard to review. Which part is code fix which part is infra enhancement? Can you split the PR?

editors/vscode/server/src/linter.rs
```diff
@@ -228,9 +232,12 @@ impl IsolatedLintHandler {
         linter: &Linter,
         path: &Path,
         plugin: Plugin,
+        source_text: Option<String>,
     ) -> Option<(PathBuf, Vec<ErrorWithPosition>)> {
-        let source_text =
-            fs::read_to_string(path).unwrap_or_else(|_| panic!("Failed to read {path:?}"));
+        let source_text = source_text.unwrap_or_else(|| {
```
Boshen, line 237: I don't get this. What's the source of truth when it comes to reading source code in vscode? Is it from the disk or some other vscode API?

IWANABETHATGUY: It comes from the language server protocol, you can get the up-to-date source code from it, current problem is that we would get the outdated source code if the user doesn't trigger the save action. 
https://github.com/oxc-project/oxc/blob/023e6eabcd5121ff1a231b5e1fcf8e9b2cf607d4/editors/vscode/server/src/linter.rs#L232C8-L233C92

IWANABETHATGUY: For more details, you could refer https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentContentChangeEvent.

PR #1526, opened 2023-11-23
feat(query): Add is_getter, is_setter, is_constructor to all Function implementors

u9g:
---

<details open="true"><summary>Generated summary (powered by <a href="https://app.graphite.dev">Graphite</a>)</summary>

> # TL;DR
> This pull request adds new properties to the `ArrowFunction`, `FnDeclaration`, and `Function` interfaces in the Trustfall schema. It also adds implementation for resolving these properties in the `adapter.rs` and `properties.rs` files. Additionally, it adds a new method `function_scope_flag` to the `Vertex` struct in the `vertex.rs` file.
> 
> # What changed
> - Added new properties (`is_getter`, `is_setter`, `is_constructor`) to the `ArrowFunction`, `FnDeclaration`, and `Function` interfaces in the Trustfall schema.
> - Implemented the resolution of these properties in the `resolve_arrow_function_property`, `resolve_fn_declaration_property`, and `resolve_function_property` functions in the `properties.rs` file.
> - Added a new method `function_scope_flag` to the `Vertex` struct in the `vertex.rs` file.
> 
> # How to test
> 1. Run the test suite to ensure that all existing tests pass.
> 2. Add new tests to cover the newly added properties and the `function_scope_flag` method.
> 3. Run the test suite again and ensure that all tests pass.
> 
> # Why make this change
> - The new properties (`is_getter`, `is_setter`, `is_constructor`) provide additional information about functions in the Trustfall schema, allowing clients to query for these properties.
> - The `function_scope_flag` method in the `Vertex` struct provides a convenient way to access the scope flags of a function node, which can be useful for various analysis and processing tasks.
</details>

### DIFF
--- a/crates/oxc_query/src/vertex.rs
+++ b/crates/oxc_query/src/vertex.rs
@@ -4,7 +4,7 @@ use enum_as_inner::EnumAsInner;
 use oxc_ast::ast::ArrayExpressionElement;
 #[allow(clippy::wildcard_imports)]
 use oxc_ast::{ast::*, AstKind};
-use oxc_semantic::{AstNode, AstNodeId};
+use oxc_semantic::{AstNode, AstNodeId, ScopeFlags};
 use oxc_span::{GetSpan, Span};
 use trustfall::provider::{Typename, VertexIterator};
 use url::Url;
@@ -276,6 +276,29 @@ impl<'a> Vertex<'a> {
                 "'function_is_generator' function should only ever be called with an ArrowFunction or FnDeclaration"
             ),
         }
+    }
+
+    pub fn function_scope_flag(&self, adapter: &Adapter<'_>) -> ScopeFlags {
+        let wanted_node_hash = match &self {
+            Vertex::ArrowFunction(data) => calculate_hash(data.arrow_expression),
+            Vertex::FnDeclaration(data) => calculate_hash(data.function),
+            _ => unreachable!(
+                "'function_scope_flag' function should only ever be called with an ArrowFunction or FnDeclaration"
+            ),
+        };
+
+        let found = adapter.semantic.nodes().iter().find(|x| {
+            let hash_of_node = match x.kind() {
+                AstKind::ArrowExpression(ae) => calculate_hash(ae),
+                AstKind::Function(fn_) => calculate_hash(fn_),
+                _ => return false,
+            };
+            hash_of_node == wanted_node_hash
+        });
+
+        adapter.semantic.scopes().get_flags(
+            found.expect("an arrowexpression or function should always be in the ast if we can have the struct of it").scope_id(),
+        )
     }
 
     pub fn function_parameter(&self) -> VertexIterator<'a, Vertex<'a>> {
--- a/crates/oxc_query/src/vertex.rs
+++ b/crates/oxc_query/src/vertex.rs
@@ -279,15 +279,21 @@ impl<'a> Vertex<'a> {
     }
 
     pub fn function_scope_flag(&self, adapter: &Adapter<'_>) -> ScopeFlags {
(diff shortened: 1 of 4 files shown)


u9g:
* main
  * **PR #1524** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1526** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1526?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1526?utm_source=stack-comment).

crates/oxc_query/src/vertex.rs
```diff
@@ -278,6 +278,29 @@ impl<'a> Vertex<'a> {
...
+    pub fn function_scope_flag<'c, 'b>(&self, adapter: &'c Adapter<'b>) -> ScopeFlags {
+        let wanted_node_hash = match &self {
+            Vertex::ArrowFunction(data) => calculate_hash(data.arrow_expression),
+            Vertex::FnDeclaration(data) => calculate_hash(data.function),
+            _ => unreachable!(
+                "'function_scope_flag' function should only ever be called with an ArrowFunction or FnDeclaration"
+            ),
+        };
+
+        let found = adapter.semantic.nodes().iter().find(|x| {
```
u9g, line 290: it's not great to iterate the entire node stack each time, is it worth adding a hashmap here to just store the calculated hash to the scope flags so if we do multiple lookups we don't need to iterate all the nodes multiple times?

crates/oxc_query/src/vertex.rs
```diff
@@ -278,6 +278,29 @@ impl<'a> Vertex<'a> {
...
+            let hash_of_node = match x.kind() {
+                AstKind::ArrowExpression(ae) => calculate_hash(ae),
+                AstKind::Function(fn_) => calculate_hash(fn_),
+                _ => return false,
+            };
+            hash_of_node == wanted_node_hash
```
u9g, line 296: This fails to find the correct function on line 2, since both functions have the same hash. Maybe augmenting by also filtering for the same start and end span would work?

```
{ class X { set foo(v) { return String(v) } } }
{ x = function (v) { return String(v); } }
```

u9g (commented):
https://github.com/oxc-project/oxc/pull/1527 should be merged before this.

crates/oxc_query/src/vertex.rs
u9g: this is fixed in #1527 to my knowledge.

Boshen:
> #1527 should be merged before this.

Since you're using graphite, you need to `gt move` and reorder the PRs. Otherwise they'll be merged in the current order.

crates/oxc_query/src/vertex.rs
Boshen: Yes, hash + span should work.

crates/oxc_query/src/vertex.rs
Boshen: Let's just iterate it until performance regression is noticeable.

u9g:
> > #1527 should be merged before this.
> 
> Since you're using graphite, you need to `gt move` and reorder the PRs. Otherwise they'll be merged in the current order.

Is this not the correct order, as in that depends on this being merged before it can be merged?

The idea is that the fix pr should be merged into this pr, which I think is what the stack expresses now, maybe I'm not understanding correctly.

Boshen:
You are free to merge these PRs yourself, I don't know what's going on 😅

PR #1551, opened 2023-11-26
feat(prettier): add id on `IfBreak` and `Group`

mysteryven:


crates/oxc_prettier/src/doc.rs
```diff
@@ -37,22 +40,36 @@ pub enum Doc<'a> {
...
 }
 
+static ID_COUNTER: AtomicUsize = AtomicUsize::new(0);
+pub type DocId = u32;
+
+impl<'a> Doc<'a> {
+    pub fn next_id() -> DocId {
+        DocId::try_from(ID_COUNTER.fetch_add(1, Ordering::SeqCst)).unwrap_or_default()
+    }
+}
```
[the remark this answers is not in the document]
Boshen: I agree, we don't need atomics for tracking state.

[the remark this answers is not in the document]
mysteryven: Done, only use a `u32` without inner mutability, it's not needed current.

crates/oxc_prettier/src/lib.rs
```diff
@@ -23,6 +23,19 @@ use crate::{doc::Doc, format::Format, printer::Printer};
 
 pub use crate::options::{ArrowParens, EndOfLine, PrettierOptions, QuoteProps, TrailingComma};
 
+type GroupId = u32;
+#[derive(Default)]
+struct GroupIdBuilder {
```
Boshen, line 28: Let's use the newtype idiom here https://doc.rust-lang.org/rust-by-example/generics/new_types.html

Just call it `GroupId` and expose the API `next_id`.

crates/oxc_prettier/src/printer/mod.rs
```diff
@@ -220,9 +228,18 @@ impl<'a> Printer<'a> {
         self.line_suffix.push(Command { indent, mode, doc: Doc::Array(docs) });
     }
 
-    fn handle_if_break(&mut self, doc: Doc<'a>, indent: Indent, mode: Mode) {
-        if mode == Mode::Break {
-            self.cmds.push(Command::new(indent, Mode::Break, doc));
+    fn handle_if_break(&mut self, if_break: IfBreak<'a>, indent: Indent, mode: Mode) {
+        let IfBreak { break_contents, flat_content, group_id } = if_break;
+        let group_mode = group_id.map_or(Some(mode), |id| self.group_mode_map.get(&id).copied());
```
camc314, line 233:
```suggestion
        let group_mode = group_id.map_or(mode, |id| self.group_mode_map.get(&id).cloned().unwrap_or(Mode::Flat));
```
i don't think this should return an option - we should always print either the contents of `flat` or `break` regardless of whether the groupid was found in the map.

prehaps we should `dbg_assert!` that it exists?

camc314: prettier for ref -> https://github.com/prettier/prettier/blob/eebf0e4b5ec8ac24393c56ced4b4819d4c551f31/src/document/printer.js#L269-L271

camc314: oh my bad, was looking at `fits`

crates/oxc_prettier/src/lib.rs
mysteryven: Done!

Boshen:
### Merge activity

* **Nov 27, 9:53 AM**: @Boshen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1551).

PR #1553, opened 2023-11-26
refactor(prettier): using enter marco instead of wrap marco

Dunqing:
I'm not so sure this is a good change. But this way we can do some generic logic and use macros in `AstKind` that we can't do in `wrap!`.


Dunqing:
* main
  * **PR #1553** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1553?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1553?utm_source=stack-comment).

Boshen:
This seems to be confusing and introducing indirections too much. Your method will also expand to lots of code which hurts the final binary size code and compilation speed. I wrote some the rules here: https://oxc-project.github.io/docs/contribute/introduction.html#development-policy

FYI we are using macros to speed up the development process, they should be simple one liners, all other logic should go into normal functions.

PR #1608, opened 2023-12-02
feat(linter): eslint-plugin-unicorn require-array-join-separator(style)

Ken-HH24:
Try to implement require-array-join-separator for #684


camc314 (approved):
thanks!

PR #1611, opened 2023-12-02
feat(linter): Implement unicorn/require-array-join-separator

sirreal:
Part of #684 

Implement the [unicorn/require-array-join-separator](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/require-array-join-separator.md) rule


sirreal:
Whoops, duplicated #1608

PR #1615, opened 2023-12-03
fix(tasks): escape more special characters

mysteryven:
We also need escape `$` and `\`: https://github.com/vitest-dev/vitest/blob/9527e65a382e0775f171f88be0ca37c465317e25/packages/snapshot/src/port/inlineSnapshot.ts#L78-L94


Boshen:
### Merge activity

* **Dec 3, 2:29 AM**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1615).

PR #1617, opened 2023-12-03
feat(linter): eslint-plugin-unicorn: explicit-length-check

RiESAEX:
[Rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/explicit-length-check.md)
#684 
It's harder than I thought.
Test cases that require a vue parser have been skipped
Some test case needs [getStaticValue](https://github.com/eslint-community/eslint-utils/blob/main/src/get-static-value.mjs#L672) to pass.

### DIFF
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -151,6 +151,7 @@ mod unicorn {
     pub mod empty_brace_spaces;
     pub mod error_message;
     pub mod escape_case;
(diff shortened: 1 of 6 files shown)


Boshen (approved):
We don't have a [getStaticValue](https://github.com/eslint-community/eslint-utils/blob/main/src/get-static-value.mjs#L672), feel free to port it ;-)

PR #1629, opened 2023-12-04
feat(semantic): support scope descendents starting from a certain scope.

milesj:
@Boshen 

The `ScopeTree.descendants` function would return all scopes starting from the root, and wasn't truly descendants from a specific scope. To improve this, I've renamed this function to `descendants_from_root` and have introduced a new `descendants` function that does support from a specific scope.

Furthermore, I've introduced helper functions to `SymbolTree` to make reading symbols/scopes easier.

To verify this functionality, I enabled the `function_name` transformer (and fixed it), and ran some example transforms. Here's the input:

```js
let fn;

fn = function (a, b, c) {
  const d = "";
};

const func = function (arg) {
  {
    const value = "";
  }
};

const f = function (f) {};
```

And the output using _the old implementation_. Note that all function names are suffixed with a number, this is incorrect, since it was inheriting far too many scopes.

```js
let fn;
fn = function fn1(a, b, c) {
	const d = '';
};
const func = function func1(arg) {
	{
		const value = '';
	}
};
const f = function f2(f) {
};
```

And here's the output with the new implementation. Note that only `f` is suffixed with a number. That's because it has a shadowed argument of the same name.

```js
let fn;
fn = function fn(a, b, c) {
	const d = '';
};
const func = function func(arg) {
	{
		const value = '';
	}
};
const f = function f1(f) {
};
```

### DIFF
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -23,7 +23,7 @@ pub struct ScopeTree {
     parent_ids: IndexVec<ScopeId, Option<ScopeId>>,
 
     /// Maps a scope to direct children scopes
-    childs_ids: IndexMap<ScopeId, Vec<ScopeId>>,
+    child_ids: IndexMap<ScopeId, Vec<ScopeId>>,
 
     flags: IndexVec<ScopeId, ScopeFlags>,
     bindings: IndexVec<ScopeId, Bindings>,
@@ -61,7 +61,7 @@ impl ScopeTree {
             }
         }
 
-        add_to_list(&scope_id, &self.childs_ids, &mut list);
+        add_to_list(&scope_id, &self.child_ids, &mut list);
 
         list.into_iter()
     }
@@ -128,10 +128,10 @@ impl ScopeTree {
         _ = self.unresolved_references.push(UnresolvedReferences::default());
 
         if let Some(parent_id) = parent_id {
-            if let Some(children) = self.childs_ids.get_mut(&parent_id) {
+            if let Some(children) = self.child_ids.get_mut(&parent_id) {
                 children.push(scope_id);
             } else {
-                self.childs_ids.insert(parent_id, vec![scope_id]);
+                self.child_ids.insert(parent_id, vec![scope_id]);
             }
         }
 
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -128,14 +128,7 @@ impl ScopeTree {
         _ = self.unresolved_references.push(UnresolvedReferences::default());
 
         if let Some(parent_id) = parent_id {
-            let list = self.child_ids.entry(parent_id).or_default();
-            list.push(scope_id);
-
-            // if let Some(children) = self.child_ids.get_mut(&parent_id) {
-            //     children.push(scope_id);
-            // } else {
-            //     self.child_ids.insert(parent_id, vec![scope_id]);
-            // }
+            self.child_ids.entry(parent_id).or_default().push(scope_id);
         }
 
         scope_id
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::hash::BuildHasherDefault;
 
 use indexmap::IndexMap;
@@ -38,7 +39,23 @@ impl ScopeTree {
         std::iter::successors(Some(scope_id), |scope_id| self.parent_ids[*scope_id])
     }
 
-    pub fn descendants(&self) -> impl Iterator<Item = ScopeId> + '_ {
+    pub fn descendants(&self, scope_id: ScopeId) -> impl Iterator<Item = ScopeId> + '_ {
+        let mut list = vec![];
+        let mut queue = VecDeque::from_iter([scope_id]);
+
+        while let Some(current_id) = queue.pop_front() {
+            for (scope, parent_scope) in self.parent_ids.iter_enumerated() {
+                if parent_scope.as_ref().is_some_and(|parent_id| parent_id == &current_id) {
+                    list.push(scope);
+                    queue.push_back(scope);
+                }
+            }
+        }
+
+        list.into_iter()
+    }
+
+    pub fn descendants_from_root(&self) -> impl Iterator<Item = ScopeId> + '_ {
         self.parent_ids.iter_enumerated().map(|(scope_id, _)| scope_id)
     }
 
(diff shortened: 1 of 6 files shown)


crates/oxc_transformer/examples/transformer.rs
```diff
@@ -31,10 +32,8 @@ fn main() {
         return;
     }
 
-    let codegen_options = CodegenOptions;
-    let printed = Codegen::<false>::new(source_text.len(), codegen_options).build(&ret.program);
```
milesj, line 35: this was transforming the input, so just removed it

Boshen (approved):
Looks good to me from my end, @IWANABETHATGUY is more familiar with this kind of stuff than I do, please take a look 😁

crates/oxc_semantic/src/scope.rs
```diff
@@ -38,7 +39,23 @@ impl ScopeTree {
         std::iter::successors(Some(scope_id), |scope_id| self.parent_ids[*scope_id])
     }
 
-    pub fn descendants(&self) -> impl Iterator<Item = ScopeId> + '_ {
+    pub fn descendants(&self, scope_id: ScopeId) -> impl Iterator<Item = ScopeId> + '_ {
+        let mut list = vec![];
+        let mut queue = VecDeque::from_iter([scope_id]);
+
+        while let Some(current_id) = queue.pop_front() {
```
IWANABETHATGUY, line 46: This algo's time complexity could be `O(n^2)`

IWANABETHATGUY: We could create a new field called `child_ids` like `parent_ids` , it store each direct child scopeids, then `descendants` could just be a linear scan.

milesj: Let me give that a shot.

milesj:
@Boshen @IWANABETHATGUY Updated with a new `child_ids` field.

Here's an example of the tree output:

```
[crates/oxc_transformer/src/es2015/function_name.rs:130] &scopes = ScopeTree {
    parent_ids: [
        None,
        Some(
            0,
        ),
        Some(
            0,
        ),
        Some(
            2,
        ),
        Some(
            0,
        ),
    ],
    childs_ids: {
        0: [
            1,
            2,
            4,
        ],
        2: [
            3,
        ],
    },
    flags: [
        ScopeFlags(
            StrictMode | Top,
        ),
        ScopeFlags(
            StrictMode | Function,
        ),
        ScopeFlags(
            StrictMode | Function,
        ),
        ScopeFlags(
            StrictMode,
        ),
        ScopeFlags(
            StrictMode | Function,
        ),
    ],
    bindings: [
        {
            "fn": 0,
            "func": 5,
            "f": 8,
        },
        {
            "a": 1,
            "b": 2,
            "c": 3,
            "d": 4,
        },
        {
            "arg": 6,
        },
        {
            "value": 7,
        },
        {
            "f": 9,
        },
    ],
    unresolved_references: [
        {},
        {},
        {},
        {},
        {},
    ],
}
```

crates/oxc_semantic/src/scope.rs
```diff
@@ -19,7 +19,12 @@ type UnresolvedReferences = FxHashMap<Atom, Vec<ReferenceId>>;
 /// `SoA` (Struct of Arrays) for memory efficiency.
 #[derive(Debug, Default)]
 pub struct ScopeTree {
+    /// Maps a scope to the parent scope it belongs in
     parent_ids: IndexVec<ScopeId, Option<ScopeId>>,
+
+    /// Maps a scope to direct children scopes
+    child_ids: IndexMap<ScopeId, Vec<ScopeId>>,
```
Boshen, line 26: Should be a bit more performant with fxhasher, https://docs.rs/indexmap/latest/indexmap/#alternate-hashers

crates/oxc_semantic/src/scope.rs
```diff
@@ -38,7 +43,30 @@ impl ScopeTree {
...
+        // Has to be a `fn` and pass arguments because we can't
+        // have recursive closures
+        fn add_to_list(
+            parent_id: &ScopeId,
+            child_ids: &IndexMap<ScopeId, Vec<ScopeId>>,
+            items: &mut Vec<ScopeId>,
+        ) {
+            if let Some(children) = child_ids.get(parent_id) {
+                for child_id in children {
+                    items.push(*child_id);
```
Boshen, line 58: Would it make a little bit faster with a `with_capacity` before the loop?

crates/oxc_semantic/src/scope.rs
```diff
@@ -20,7 +19,12 @@ type UnresolvedReferences = FxHashMap<Atom, Vec<ReferenceId>>;
 /// `SoA` (Struct of Arrays) for memory efficiency.
 #[derive(Debug, Default)]
 pub struct ScopeTree {
+    /// Maps a scope to the parent scope it belongs in
     parent_ids: IndexVec<ScopeId, Option<ScopeId>>,
+
+    /// Maps a scope to direct children scopes
+    childs_ids: IndexMap<ScopeId, Vec<ScopeId>>,
```
Boshen: Maybe `child_ids`?

crates/oxc_semantic/src/scope.rs
```diff
@@ -98,6 +126,18 @@ impl ScopeTree {
...
+
+        if let Some(parent_id) = parent_id {
+            let list = self.child_ids.entry(parent_id).or_default();
+            list.push(scope_id);
+
+            // if let Some(children) = self.child_ids.get_mut(&parent_id) {
+            //     children.push(scope_id);
+            // } else {
+            //     self.child_ids.insert(parent_id, vec![scope_id]);
+            // }
```
Boshen, line 138: Some left over commented out code.

crates/oxc_transformer/src/es2015/function_name.rs
```diff
@@ -24,18 +26,18 @@ pub struct FunctionName<'a> {
...
+        options: &TransformOptions,
     ) -> Option<Self> {
         // Disabled for now
-        None
-        // (options.target < TransformTarget::ES2015 || options.function_name).then(|| Self {
-        //     ast,
-        //     ctx,
-        //     // TODO hook up the plugin
-        //     unicode_escapes: true,
-        // })
```
Boshen, line 38: Commented out code.

crates/oxc_transformer/src/es2015/function_name.rs
```diff
@@ -89,26 +99,39 @@ impl<'a> FunctionName<'a> {
...
+        mut id: BindingIdentifier,
+        scope_id: Option<ScopeId>,
+    ) {
         // function () {} -> function name() {}
         if let Expression::FunctionExpression(func) = expr {
-            let scopes = self.ctx.scopes();
             let mut count = 0;
 
             // let mut finder = IdentFinder { id, found: 0 };
             // finder.visit_expression(expr);
```
Boshen, line 124: Commented out code.

Boshen (commented):
Should be good to go after cleaning up some of the commented out code.

crates/oxc_semantic/src/scope.rs
milesj: Don't have a length before hand.

Issue #974, opened 2023-10-10
☂️ Transpiler / Transformer


Boshen:
## Contribution Instructions

The implementation is in `crates/oxc_transformer`.

To get started, take a look at the [previous implementations](https://github.com/web-infra-dev/oxc/pulls?q=is%3Apr+is%3Aclosed+label%3AA-transformer).

The integrations tests are located in `tasks/transform_conformance/src/lib.rs`,
Run it with `just watch 'run -p oxc_transform_conformance'` or with a filter `just watch "run -p oxc_transform_conformance -- --filter type-arguments"`.

---
## Tasks

- [ ] https://github.com/web-infra-dev/oxc/issues/979
- [x] https://github.com/web-infra-dev/oxc/pull/1001
- [x] https://github.com/web-infra-dev/oxc/issues/999
- [x] Fix transformer benchmark because it is transforming the already transformed AST https://github.com/web-infra-dev/oxc/pull/998

## Third Party Plugins

- [ ] https://www.npmjs.com/package/babel-plugin-import

## Feature list

Legend:
* [Syntax] means this is a syntax plugin, no code is required because it is supported by the parser.
* [Regex] means this is a regex transform, which is not supported.
* [Codegen] means the code generator is responsible for the feature.

## @babel/preset-react

- [ ] @babel/plugin-transform-react-jsx

## [@babel/preset-typescript](https://babel.dev/docs/babel-preset-typescript)

- [ ] @babel/plugin-transform-typescript

### ES2024

- [ ] [Regex] unicode-sets-regex

### ES2023

Does not have any features.

### ES2022

- [ ] class-properties
- [x] class-static-block https://github.com/web-infra-dev/oxc/pull/962
- [ ] private-methods
- [ ] private-property-in-object
- [x] [Syntax] syntax-top-level-await

### ES2021

- [ ] logical-assignment-operators https://github.com/web-infra-dev/oxc/pull/923
- [x] [Syntax] numeric-separator

### ES2020

- [ ] dynamic-import
- [ ] export-namespace-from
- [ ] nullish-coalescing-operator https://github.com/web-infra-dev/oxc/pull/1004
- [ ] optional-chaining
- [x] [Syntax] syntax-bigint
- [x] [Syntax] syntax-dynamic-import
- [x] [Syntax] syntax-import-meta

### ES2019

- [x] optional-catch-binding
- [x] [Codegen] json-strings

### ES2018

- [ ] async-generator-functions
- [ ] object-rest-spread
[...]

PR #1649, opened 2023-12-10
feat(transformer): duplicate keys

Ken-HH24:
Try to implement [duplicate-keys](https://github.com/babel/babel/tree/main/packages/babel-plugin-transform-duplicate-keys) for #974.
Left a failed test case which requires [computed-properties](https://babeljs.io/docs/babel-plugin-transform-computed-properties) to be implemented.


mysteryven (approved):
Nice work,Thank you!!

PR #1657, opened 2023-12-12
perf(linter/react): find class node by symbols in get_parent_es6_component

Dunqing:
This way we can get the class node faster. But I don't know if this is a good way. In `eslint-plugin-react`, they get class node by scope. But oxc cannot do the same way

### DIFF
--- a/crates/oxc_semantic/src/symbol.rs
+++ b/crates/oxc_semantic/src/symbol.rs
@@ -39,6 +39,10 @@ impl SymbolTable {
         self.spans.iter_enumerated().map(|(symbol_id, _)| symbol_id)
     }
 
+    pub fn iter_rev(&self) -> impl Iterator<Item = SymbolId> + '_ {
+        self.spans.iter_enumerated().rev().map(|(symbol_id, _)| symbol_id)
+    }
+
     pub fn get_symbol_id_from_span(&self, span: &Span) -> Option<SymbolId> {
         self.spans
             .iter_enumerated()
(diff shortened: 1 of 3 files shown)


Dunqing:
* main
  * **PR #1657** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1657?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1657?utm_source=stack-comment).

crates/oxc_semantic/src/symbol.rs
```diff
@@ -39,6 +39,10 @@ impl SymbolTable {
         self.spans.iter_enumerated().map(|(symbol_id, _)| symbol_id)
     }
 
+    pub fn iter_rev(&self) -> impl Iterator<Item = SymbolId> + '_ {
```
Boshen, line 42: I'm unsure whether this is a good public API. Are we able to directly call `iter().rev()` on call site?

Boshen (changes requested):
How did you measure the performance of this change?

crates/oxc_semantic/src/symbol.rs
Dunqing: I initially thought the same, but the `iter` cannot call the `rev()`. Here is the  implementation of `iter`

https://github.com/oxc-project/oxc/blob/341081fbfacece1986e852acacfddfac806bfbd7/crates/oxc_semantic/src/symbol.rs#L38-L40

Boshen (approved):

PR #1660, opened 2023-12-12
feat(linter): eslint-plugin-unicorn/prefer-prototype-methods

haocheng6:
This PR implements the [eslint-plugin-unicorn/prefer-prototype-methods](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-prototype-methods.md) rule.

Tests taken from https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/prefer-prototype-methods.mjs.

Related issue: #684

I added a "TODO" comment in the code because it looks like we don't have a utility function similar to `getPropertyName` (which uses `getStaticValue` under the hood) in @eslint-community/eslint-utils yet. I will try to implement the function sometime if no one else is working on it.


Dunqing:
Thank you for your contribution!

It looks like you are creating new rules manually. If you want to create a new rule you can run `just new-unicorn-rule prefer-prototype-methods`. The rule file will automatically created

crates/oxc_linter/src/snapshots/prefer_prototype_methods.snap
```diff
@@ -0,0 +1,143 @@
...
+   ╰────
+
+  ⚠ eslint-plugin-unicorn(prefer-prototype-methods): Prefer using `Array.prototype.push`.
+   ╭─[prefer_prototype_methods.tsx:1:1]
+ 1 │ const push = [].push.bind(foo)
+   ·              ───────
+   ╰────
+
+  ⚠ eslint-plugin-unicorn(prefer-prototype-methods): Prefer using method from `Array.prototype`.
+   ╭─[prefer_prototype_methods.tsx:1:1]
```
Dunqing, line 48: It looks like there should be `Array.prototype.method`

Dunqing (changes requested):

crates/oxc_linter/src/snapshots/prefer_prototype_methods.snap
haocheng6: For this test case, the error message generated by eslint-plugin-unicorn is also "Prefer using method from `Array.prototype`." (See [the snapshot file](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/82d3de639ef3041034c4e8c5973ac7ff39b5aaed/test/snapshots/prefer-prototype-methods.mjs.md?plain=1#L128-L133) in the eslint-plugin-unicorn repo.)

I compared my snapshot file with theirs. There are only two differences in the generated error message. (The fixed code for each case below is correct. Only the error message is different from that of eslint-plugin-unicorn.)

## Case 1

Code: `const method = "realMethodName";const foo = [][method].call(foo)`
Error message generated by eslint-plugin-unicorn: ``Prefer using method from `Array.prototype`.``
Error message generated by my code: ``Prefer using method from `Array.prototype`.``
Analysis: When we have a utility function similar to `getPropertyName` in @eslint-community/eslint-utils, the error message for this case can be easily fixed.

## Case 2

Code: `[][Symbol.iterator].call(foo)`
Error message generated by eslint-plugin-unicorn: ``Prefer using `Array.prototype.Symbol(Symbol.iterator)`.``
Error message generated by my code: ``Prefer using method from `Array.prototype`.``
Analysis: The error message generated by eslint-plugin-unicorn does not look correct to me because `Array.prototype.Symbol(Symbol.iterator)` is not valid JavaScript code, so I am unsure about the expected error message for this case.

What do you think about these two cases?

(Btw, to reply to your comment above, I actually generated the rule file using the `just` command you mentioned, but I did some manual editing to the generated test cases like removing unnecessary `r#` prefixes, so it might look like I manually created the rule file, but thanks for the pointer.)

Dunqing: Thanks for taking the time to investigate here.

I think we can keep the current error message. Looks better than unicorn

Dunqing (approved):

PR #1668, opened 2023-12-13
feat(linter): add  jsx-a11y settings

msdlisper:
When we developed linter  for #1141 , we needed to configure some settings for `jsx-a11y`, which was not supported before, but I am trying to support it now.
like this:
```
fn config() -> serde_json::Value {
    serde_json::json!([2,{
        "ignoreNonDOM": true
    }])
}

fn settings() -> serde_json::Value {
    serde_json::json!({
        "jsx-a11y": {
            "components": {
                "Button": "button",
            }
        }
    })
}

let pass = vec![
    ("<Button />", Some(config()), Some(settings())),
];
```

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -17,7 +17,7 @@ mod rules;
 mod service;
 mod utils;
 
-use std::{self, fs, io::Write, rc::Rc, time::Duration};
+use std::{self, collections::HashMap, fs, io::Write, rc::Rc, time::Duration};
 
 use oxc_diagnostics::Report;
 pub(crate) use oxc_semantic::AstNode;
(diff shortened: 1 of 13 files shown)


Boshen:
Thank you for the PR!

We now support reading from eslint config json! This means we can read values off 

https://www.npmjs.com/package/eslint-plugin-jsx-a11y#configurations

I think you can read these values into a struct (deserialized) and then add it to `LintContext`.

msdlisper:
@Boshen I check out the new code just now,  let me read the settings from config json.

msdlisper:
@Boshen Done :)

camc314 (approved):
Looks great to me! Thanks!

crates/oxc_linter/src/lib.rs
```diff
@@ -33,10 +33,38 @@ pub use crate::{
...
+impl Default for LintSettings {
+    fn default() -> Self {
+        Self { jsx_a11y: JsxA11y { polymorphic_prop_name: None, components: HashMap::new() } }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct JsxA11y {
+    polymorphic_prop_name: Option<String>,
+    components: HashMap<String, String>,
```
Boshen, line 50: Let's use `use rustc_hash::FxHashMap` here.

Boshen (approved):
The hashmap can be done with a follow up PR.

PR #1681, opened 2023-12-15
fix(vscode): report problem more accurately 

### LINKED ISSUE #1674: vscode: report problem more accurately
image
I'd like it to report only the problematic code, not lines between them.

IWANABETHATGUY:
Closed https://github.com/oxc-project/oxc/issues/1674

### DIFF
--- a/editors/vscode/server/src/linter.rs
+++ b/editors/vscode/server/src/linter.rs
@@ -61,7 +61,6 @@ impl ErrorWithPosition {
     fn to_lsp_diagnostic(&self, path: &PathBuf) -> lsp_types::Diagnostic {
         let severity = match self.miette_err.severity() {
             Some(Severity::Error) => Some(lsp_types::DiagnosticSeverity::ERROR),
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
I found the interesting thing is our marking convention is not same as `biome`, they use trigger span as primary span, the original as `secondary` span, we mark trigger span as secondary span,  referenced original span as primary span.


here is a case: 
```js
const a = 3;
const a = 3;
```
image

IWANABETHATGUY:
Noted that, the vscode diagnostic render engine only renders the original span of the diagnostic, so If we use the same render logic as biome, it will not work, here is an example: 

image
you could see multiple diagnostic reuse the same **original span**, so only the first line got marked as warning.
That is not what we want

IWANABETHATGUY:
So we use the same strategy as rust analyzer, they extend the diagnostics list with an inverted diagnostic list, for each related info or original diagnostic, we convert them into a diagnostic and append original diagnostic into their related info list, and mark the severity as `hints`, here is the screenshot: 

## Before
image
## After
image

Boshen (approved):

PR #1728, opened 2023-12-18
refactor(ast): introduce `ThisParameter`

magic-akari:
Most TypeScript types can be eliminated during the code generation phase by not printing the corresponding AST nodes. 
The changes in this PR enable applying a similar technique to the `this` parameter.


crates/oxc_parser/src/js/function.rs
```diff
@@ -74,6 +74,18 @@ impl<'a> Parser<'a> {
         Ok(self.ast.formal_parameters(self.end_span(span), params_kind, list.elements, list.rest))
     }
 
+    pub(crate) fn parse_formal_parameters_with_ts_this(
+        &mut self,
+        params_kind: FormalParameterKind,
+    ) -> Result<(Option<TSThisParameter<'a>>, Box<'a, FormalParameters<'a>>)> {
+        let span = self.start_span();
```
Boshen, line 81: I think we need a "if source is typescript" check here?

Boshen:
This'll definitely help downstream tools. +1 from me.

crates/oxc_parser/src/js/function.rs
magic-akari: If the source code is not TypeScript, then `this_param` will always be `None`, and the function will have the exact same effect as `parse_formal_parameters`.

The purpose of this function is to extract and separate `this_param` from the list and return it. This function's name may be misleading, perhaps it can be changed to a more suitable one?

crates/oxc_parser/src/ts/statement.rs
```diff
@@ -422,19 +422,11 @@ impl<'a> Parser<'a> {
         ))
     }
 
-    pub(crate) fn parse_ts_this_parameter(&mut self) -> Result<FormalParameter<'a>> {
+    pub(crate) fn parse_ts_this_parameter(&mut self) -> Result<TSThisParameter<'a>> {
         let span = self.start_span();
-        let (ident_span, name) = self.parse_identifier_kind(Kind::This);
+        let _ = self.parse_identifier_kind(Kind::This);
```
magic-akari, line 427: Does it make sense to keep the `this_ident` in AST?

Boshen: yeah it's good to add the `this` identifier to the AST.

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1530,6 +1530,7 @@ pub struct Function<'a> {
     pub expression: bool,
     pub generator: bool,
     pub r#async: bool,
+    pub this_param: Option<TSThisParameter<'a>>,
```
Boshen, line 1533: let's add a documentation for this, a reference to the typescript manual + an example in code.

Boshen (approved):

PR #1753, opened 2023-12-20
perf(linter): tighter main loop

Boshen:


Dunqing:
Wow, this is a huge improvement

PR #1765, opened 2023-12-21
fix: don't lint files in .gitignore and .eslintignore

### LINKED ISSUE #1752: VSCode: don't lint files in .gitignore
i do not need my auto-generated build files to be linted.
if somebody wants to lint files in git-ignored files, they can use the CLI.

IWANABETHATGUY:
Closed https://github.com/oxc-project/oxc/issues/1752

[录屏 2023年12月21日 18时14分38秒.webm](https://github.com/oxc-project/oxc/assets/17974631/8bbdddce-1d2b-4b37-9fbd-c36a4b626a13)

### DIFF
--- a/editors/vscode/server/src/main.rs
+++ b/editors/vscode/server/src/main.rs
@@ -4,11 +4,13 @@ mod options;
 mod walk;
 
 use crate::linter::{DiagnosticReport, ServerLinter};
+use globset::Glob;
+use ignore::gitignore::Gitignore;
 use log::{debug, error};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt::Debug;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use dashmap::DashMap;
@@ -33,6 +35,7 @@ struct Backend {
     server_linter: ServerLinter,
     diagnostics_report_map: DashMap<String, Vec<DiagnosticReport>>,
     options: Mutex<Options>,
+    gitignore_glob: Mutex<Option<Gitignore>>,
 }
 #[derive(Debug, Serialize, Deserialize, Default, PartialEq, PartialOrd, Clone, Copy)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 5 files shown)


editors/vscode/server/src/main.rs
```diff
@@ -236,20 +249,45 @@ impl LanguageServer for Backend {
...
+        let walk = ignore::WalkBuilder::new(uri.path()).ignore(true).hidden(false).git_global(false).build();
+        for item in walk.into_iter() {
+            if let Ok(entry) = item {
+                if ignore_file_glob_set.is_match(entry.path()) {
+                    gitignore_builder.add(entry.path());
+                }
+            }
+        }
+
+        *self.gitignore_glob.lock().await = gitignore_builder.build().ok();
```
Boshen, line 284: Let's move these new code to a function `init_file_ignore` into a function instead of dumping them here.

editors/vscode/server/src/main.rs
```diff
@@ -236,20 +249,45 @@ impl LanguageServer for Backend {
...
+
+        let mut gitignore_builder = ignore::gitignore::GitignoreBuilder::new(uri.path());
+        let walk = ignore::WalkBuilder::new(uri.path()).ignore(true).hidden(false).git_global(false).build();
+        for item in walk.into_iter() {
+            if let Ok(entry) = item {
+                if ignore_file_glob_set.is_match(entry.path()) {
+                    gitignore_builder.add(entry.path());
+                }
+            }
+        }
```
Boshen, line 282: Take a look at https://github.com/oxc-project/oxc/blob/82dc4035ebc15fcbd898381392423197bbcd00ea/crates/oxc_cli/src/walk.rs#L71-L89

It feels like they are the same thing? 

Can you double check whether your code can be simplified to what's written in `walk.rs`?

Boshen (changes requested):

editors/vscode/server/src/main.rs
```diff
@@ -276,6 +314,16 @@ impl Backend {
             }
         }
     }
+
+    async fn is_ignored(&self, uri: &Url) -> bool {
+        // TODO: cache the result, for now it only cost a few us to calculate if the file is ignored by gitignore or .eslintignore
```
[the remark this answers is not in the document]
IWANABETHATGUY: the copilot adds it, I don't find the `match` is slow actually.

editors/vscode/server/src/main.rs
IWANABETHATGUY: No,  two function usages are different, the `walk` function will traverse the whole file tree for every lint action, we can't get all ignore patterns from `Walk` which provided by `ignore` crate.  `init_ignore_blob` wants to collect all the `ignore` patterns only once when the server is initialized,

Boshen (approved):

PR #1771, opened 2023-12-21
fix(linter): support read the third item in config file

### LINKED ISSUE #1757: Linter: configuration for specific rules (e.g. eqeqeq)
Sorry if this information is available somewhere, I've searched around the site and GitHub quite extensively...

I'm looking at adding Oxlint to our workflow, and hopefully replace ESLint in the long run. I found it a bit hard to understad which rules are currently supported but I have now understood that `crates/oxc_linter/src/rules/eslint` has files corresponding to the ESLint rule names 🎉 

Trying to enable the same rules as we currently use in ESLint led to a hiccup though. My approach was something like: run `eslint --print-config rules` and keep the supported rules in a file called `oxlint.json`. Then run Oxlint with this config, `oxlint --config oxlint.json`.

However, this quickly led to a number of failed lints with the `eqeqeq` rule, since we are using it with the config `"null": "ignore"`.

This makes it so that triple equals is enforced for everything but `null` checks. I believe that this is a very good way to handle `===`, and when I see `=== undefined` or `=== null` I always look extra closely at the code as I've seen this introduce bugs numerous times.

For those not familiar, `x == null` is the equivalent of `x === null || x === undefined`. With how null is used in the wider JavaScript ecosystem, I've found that these cases are very common:

1. Libraries with functions that are documented to return `undefined`, but actually return `null`
2. Libraries with functions that are documented to return `null`, but actually return `undefined`
3. Libraries with functions that are documented to return `null`, but in some cases return `undefined` (e.g. due to early return without `null`)
4. Libraries with functions that change between `null`/`undefined` in a non-major versioned release

Because of this, I've taken to be liberal when receiving `null` or `undefined`, and always use `== null` unless there's supposed to be a clear distinction of `null` vs `undefined` (e.g. in a patch `null` could be remove this field, and `undefined` would be keep as is).

-----

So, that was a lot of background to ask: Is configuration of rules something that is on the radar? (or perhaps already possible, just missing in eqeqeq?)

mysteryven:
closes: #1757

### DIFF
--- a/crates/oxc_linter/fixtures/eslint_config.json
+++ b/crates/oxc_linter/fixtures/eslint_config.json
@@ -1,0 +1,20 @@
+{
+    "rules": {
+        "no-console": "off",
+        "no-bitwise": [
+            "error",
+            {
+                "allow": [
+                    "~"
+                ]
+            }
+        ],
+        "eqeqeq": [
+            "error",
+            "always",
+            {
+                "null": "ignore"
+            }
+        ]
+    }
+}
--- a/crates/oxc_linter/src/config/mod.rs
+++ b/crates/oxc_linter/src/config/mod.rs
@@ -80,7 +80,7 @@ impl ESLintConfig {
             // The rule is included if it's in the extends set and not explicitly disabled,
             // or if it's explicitly enabled
             if (in_extends && !is_explicitly_handled) || policy.is_enabled() {
-                Some(rule.read_json(config.cloned()))
+                Some(rule.read_json(config.clone()))
             } else {
                 None
             }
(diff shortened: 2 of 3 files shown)


mysteryven:
* `main`
  * **PR #1771** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1771?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1771?utm_source=stack-comment).

camc314 (approved):

PR #1802, opened 2023-12-24
WIP: feat(transformer): add support for babel-plugin-transform-instanceof

qinyuhang:
Hi, I have tried to implement the plugin https://github.com/babel/babel/tree/main/packages/babel-plugin-transform-instanceof.

I have made a progress of transform code from `a instanceOf b` to `babelHelpers.instanceof(a, b);`

But I am quiet confused about the original code of try to find if the binary expression is already under a babelHelpers function part. Does OXC has utils can performance the same function of those codes?
```
const isUnderHelper = path.findParent(path => {
            return (
              (path.isVariableDeclarator() && path.node.id === helper) ||
              (path.isFunctionDeclaration() &&
                path.node.id &&
                path.node.id.name === helper.name)
            );
          });
```
https://github.com/babel/babel/blob/d2928229f6a09e6213475bf8a183d8d9115227af/packages/babel-plugin-transform-instanceof/src/index.ts#L19

### DIFF
--- a/crates/oxc_transformer/src/es2015/instanceof.rs
+++ b/crates/oxc_transformer/src/es2015/instanceof.rs
@@ -1,4 +1,5 @@
-use oxc_ast::{ast::*, AstBuilder};
+use oxc_ast::{ast::*, AstBuilder, AstKind};
+use oxc_semantic::AstNodeId;
 use oxc_span::SPAN;
 use std::rc::Rc;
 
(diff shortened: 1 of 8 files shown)


Boshen:
> But I am quiet confused about the original code of try to find if the binary expression is already under a babelHelpers function part. Does OXC has utils can performance the same function of those codes?

We are still figuring out how to support these kind feature in Rust. Feel free to explore ways to achieve this without sacrificing performance.

As for this PR, can use clean up the errors and get CI to pass so I can start merging?

qinyuhang:
> > But I am quiet confused about the original code of try to find if the binary expression is already under a babelHelpers function part. Does OXC has utils can performance the same function of those codes?
> 
> We are still figuring out how to support these kind feature in Rust. Feel free to explore ways to achieve this without sacrificing performance.
> 
> As for this PR, can use clean up the errors and get CI to pass so I can start merging?

Got it. I'll mark that as TODO and clean the errors first.

qinyuhang:
I tried using those way to access the parent of the current visiting expression. But it look like we have AstKind -> Expression, but we do not have reference from Expression -> AstKind, thus, it is unable to visit the parent of current Expression?
```
let is_under_helper = self
                    .ctx
                    .semantic()
                    .nodes()
                    // FIXME: how to get the AstNodeId of the expression?
                    .iter_parents(AstNodeId::new(1)) // expr.get_id or sth
                    .any(|parent| {
                        if let AstKind::CallExpression(CallExpression { callee, .. }) =
                            parent.kind()
                        {
                            return callee.is_specific_member_access("babelHelpers", "instanceof");
                        };
                        false
                    });
```

Boshen (approved):

qinyuhang:
A little thoughts. In order to acquire the ability of trace back AstNode from an expression, I think maybe we can maintain a HashMap of AstNodeId and the Expression?

PR #1810, opened 2023-12-25
feat(linter): eslint-plugin-jsx-a11y aria-proptypes

poteboy:
This PR is a work in progress for implementing the `aria-proptypes` rule as defined in the [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-proptypes.md), associated with issue #1141. 

The current implementation checks JSX attributes against ARIA property definitions but struggles with static analysis when unary operators are used in prop values.


Boshen:
I rebased this on top of main, but it seems pretty hard to get the tests passing. Closing as stale. Feel free to pick this up again.

PR #1823, opened 2023-12-26
fix(lsp): make the server available in nvim-lspconfig

IWANABETHATGUY:
1. It seems that nvim-lspconfig diagnostics handler require **version** of **document**, now we can use 
`oxc_vscode` language server in neomvin with [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) 

image

### DIFF
--- a/editors/vscode/server/src/main.rs
+++ b/editors/vscode/server/src/main.rs
@@ -113,6 +113,7 @@ impl LanguageServer for Backend {
     }
 
     async fn did_change_configuration(&self, params: DidChangeConfigurationParams) {
+        debug!("{:?}", &params.settings);
         let changed_options = match serde_json::from_value::<Options>(params.settings) {
             Ok(option) => option,
             Err(err) => {
@@ -173,7 +174,7 @@ impl LanguageServer for Backend {
         if self.is_ignored(&params.text_document.uri).await {
             return;
         }
-        self.handle_file_update(params.text_document.uri, None).await;
+        self.handle_file_update(params.text_document.uri, None, None).await;
     }
 
     /// When the document changed, it may not be written to disk, so we should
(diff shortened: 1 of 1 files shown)


editors/vscode/server/src/main.rs
```diff
@@ -113,6 +113,7 @@ impl LanguageServer for Backend {
     }
 
     async fn did_change_configuration(&self, params: DidChangeConfigurationParams) {
+        debug!("{:?}", &params.settings);
```
Boshen, line 116: Do we want to add a lint for this? I've seen this twice already :-(

IWANABETHATGUY: removed

IWANABETHATGUY: This is different with `dbg! ()`, something we want to explicitly add some log to debug language server in production.

IWANABETHATGUY: But this `debug!` is unused indeed.

emko:
do you mind sharing how you got oxlint to work with nvim-lspconfig?

ahmedelgabri:
@emko 

> do you mind sharing how you got oxlint to work with nvim-lspconfig?

I was looking into this myself and here is how it works:

- `oxc_language_server` is already included with the `oxlint` package https://github.com/oxc-project/oxc/discussions/3272#discussioncomment-10592147, make sure it's available in your path. (_although it would be better if provided separately_)
- Then you will need to configure `nvim-lspconfig` like this, I found it in @IWANABETHATGUY neovim config https://github.com/IWANABETHATGUY/nvim-config/blob/cd8c56545b0a57877e1f6ddfb63f9e18ec742a34/lua/user/lsp/settings/oxc_ls.lua

That's it.

Issue #612, opened 2023-07-25
feat(linter): eslint/no-irregular-whitespace


DonIsaac:
Implement ESLint's [`no-irregular-whitespace`](https://eslint.org/docs/latest/rules/no-irregular-whitespace#rule-details) rule.

To get started, run
```sh
just new-rule no-irregular-whitespace
```

_note: this is not a sale error_

koddsson:
Yo! I don't really Rust but I'm messing around trying to implement this rule. Is it correct of me to think that the whitespaces don't show up as a AST node? Should I get the source code from the context instead and grep that for a irregular whitespace and add a diagnostic?

koddsson:
[Looks like that's how ESlint does it :)](https://github.com/eslint/eslint/blob/main/lib/rules/no-irregular-whitespace.js)

Boshen:
We have the parser, which means we can store the irregular whitespaces into a data structure (e.g. [Trivias](https://github.com/web-infra-dev/oxc/blob/main/crates/oxc_parser/src/lexer/trivia_builder.rs) in our case)

And then print them out from the linter.

DeividAlmeida:
it's done!!

PR #1835, opened 2023-12-27
feat(linter): no-irregular-whitespace rule

DeividAlmeida:
Parser, trivias and trivias_builder were edited to get all whitespaces. Now Trivias struct store comments and whitespaces Vec. After that, i will implement the no-irregular-whitespace rule.

P.S.: There isn't a way to implement this feature without lose a little bit of performance, comparing with my last PR #1819 to minimax this trouble instead of store the irregular whitespace as Span it was stored as u32, i removed a map iterator and removed too a unused function. If you have a suggestion about it pls give me a feedback.

### DIFF
--- a/crates/oxc_ast/src/trivia.rs
+++ b/crates/oxc_ast/src/trivia.rs
@@ -1,13 +1,13 @@
 //! Trivias such as comments
 
-use std::collections::BTreeMap;
 use oxc_span::Span;
+use std::collections::BTreeMap;
 
 /// A vec of trivias from the lexer, tupled by (span.start, span.end).
 #[derive(Debug, Default)]
 pub struct Trivias {
     pub comments: Vec<(u32, u32, CommentKind)>,
-    pub whitespaces: Vec<Span>,
+    pub irregular_whitespaces: Vec<Span>,
 }
 
 /// Trivias such as comments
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY:
Can we just implement the rule in eslint way, https://github.com/eslint/eslint/blob/main/lib/rules/no-irregular-whitespace.js, they just use source code to detect irregular whitespace rule, https://github.com/eslint/eslint/blob/main/lib/rules/no-irregular-whitespace.js#L203-L205

IWANABETHATGUY:
Because lexer is the infra of the Javascript tools, not only linter use it, transformer, formatter, minifier would use it too.

DeividAlmeida:
Hi @IWANABETHATGUY thanks for your feedback, I did this way because in the insure of these feature @Boshen suggest to edit the parser and store it in a vec

DeividAlmeida:
image

https://github.com/oxc-project/oxc/issues/612

DeividAlmeida:
in discord channel
image

DeividAlmeida:
but i am just wait you decide what way should i proceed, both way are well.

IWANABETHATGUY:
Let's just hold this pull request, and wait @Boshen to take a look.

Boshen:
To minimize the performance impact, we should only store the irregular whitespaces

https://github.com/oxc-project/oxc/blob/1f3f21d722d22f97f20d9d2b47050dc86e185660/crates/oxc_parser/src/lexer/mod.rs#L360

DeividAlmeida:
done!

Dunqing (approved):

PR #1850, opened 2023-12-29
Remove redundant code from lexer

overlookmotel:
Just removes a couple of lines of redundant code from the lexer.

A note on the 2nd one:

```rs
let mut builder = AutoCow::new(lexer);
let c = lexer.consume_char();
builder.push_matching(c);
```

`push_matching()` is a no-op unless `force_allocation_without_current_ascii_char()` has already been called. Here the `AutoCow` has just been freshly created, so we know it hasn't.


Boshen (approved):
Thank you! Very nice catch! Any improvements to the lexer will give us a little performance improvement. Love this.

overlookmotel:
Compiler would have probably been smart enough to chop them out anyway, but was just aiming for tidier code.

PR #1858, opened 2023-12-29
fix(linter): ignore anonymous functional components in arrays for eslint-plugin-react(jsx-key)

### LINKED ISSUE #1857: False positive - eslint-plugin-react(jsx-key): Missing "key" prop for element in array.
Defining a [decorator for a Ladle story](https://ladle.dev/docs/decorators/) like so...

```
MyStory.decorators = [
  (Component) => (
    <div style={{ margin: "3em" }}>
      <Component />
    </div>
  ),
];
```

reports that the `<div>` should have a key. I believe this is incorrect because this is actually an anonymous React FC that happens itself to be within an array, but the `<div>` is not like this

```
⚠ eslint-plugin-react(jsx-key): Missing "key" prop for element in array.
   ╭─[jsx_key.tsx:2:1]
 2 │         MyStory.decorators = [
 3 │           (Component) => <div><Component /></div>
   ·                           ───
 4 │         ];
   ╰────
```

I would love to contribute a fix for this myself... please let me try to provide a PR...

maurice:
Attempt to fix #1857

I'm new to the project so any and all feedback welcome - thanks :)

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_key.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_key.rs
@@ -83,6 +83,7 @@ fn is_in_array_or_iter<'a, 'b>(
 
     let mut is_outside_containing_function = false;
     let mut is_explicit_return = false;
+    let mut is_arrow_expr_statement = false;
 
     loop {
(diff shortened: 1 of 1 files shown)


camc314 (approved):
Thank you!

PR #1859, opened 2023-12-30
refactor(linter) remove duplicated code in media-has-caption

msdlisper:
I noticed that this function has been implemented in the utils, which is named `get_element_type`.


camc314 (approved):
Thank you!

PR #1860, opened 2023-12-30
fix(parser): error on source larger than 4 GiB

overlookmotel:
`Token` and `Span` both represent `start` and `end` as `u32`.

This limits size of source which can be parsed to `u32::MAX`.

https://github.com/oxc-project/oxc/blob/19577709dbcfc9761cfcb1b39ec7f3ebdc810285/crates/oxc_span/src/span.rs#L14-L20

However, this constraint is currently not enforced.

In a release build, code will not panic on arithmetic overflow, so `start`/`end` could wrap around back to zero if source is 4 GiB or more.

That'd produce nonsense spans. But worse, the lexer relies in some places on `self.current.token.start` being correct, so if the value wrapped around, possibly it'd keep rewinding to the start of the source and lexing it again, causing an infinite loop.

In worst case, if for some reason an application's public API used OXC's parser with user-supplied source code (parser-as-a-service!), this could be exploited for denial of service.

This PR adds an assertion to catch this at the start of parsing instead.

This does add an extra instruction, but I imagine the effect will be negligible compared to the work required to parse the code.

### DIFF
--- a/crates/oxc_parser/src/diagnostics.rs
+++ b/crates/oxc_parser/src/diagnostics.rs
@@ -7,7 +7,7 @@ use oxc_span::Span;
 #[derive(Debug, Error, Diagnostic)]
 #[error("Source length exceeds 4 GiB limit")]
 #[diagnostic()]
-pub struct OverlongSource(#[label] pub Span);
+pub struct OverlongSource;
 
 #[derive(Debug, Error, Diagnostic)]
 #[error("Flow is not supported")]
--- a/crates/oxc_parser/src/diagnostics.rs
+++ b/crates/oxc_parser/src/diagnostics.rs
@@ -5,6 +5,11 @@ use oxc_diagnostics::{
 use oxc_span::Span;
 
 #[derive(Debug, Error, Diagnostic)]
+#[error("Source length exceeds 4 GiB limit")]
+#[diagnostic()]
+pub struct OverlongSource(#[label] pub Span);
+
+#[derive(Debug, Error, Diagnostic)]
 #[error("Flow is not supported")]
 #[diagnostic()]
 pub struct Flow(#[label] pub Span);
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -32,7 +32,7 @@ use self::{
     string_builder::AutoCow,
     trivia_builder::TriviaBuilder,
 };
-use crate::diagnostics;
+use crate::{diagnostics, MAX_LEN};
 
 #[derive(Debug, Clone)]
 pub struct LexerCheckpoint<'a> {
@@ -72,6 +72,11 @@ pub struct Lexer<'a> {
 #[allow(clippy::unused_self)]
 impl<'a> Lexer<'a> {
     pub fn new(allocator: &'a Allocator, source: &'a str, source_type: SourceType) -> Self {
+        // Token's start and end are u32s, so limit for length of source is u32::MAX bytes.
+        // Only a debug assertion is required, as parser checks length of source before calling
+        // this method.
+        debug_assert!(source.len() <= MAX_LEN);
+
         let token = Token {
             // the first token is at the start of file, so is allows on a new line
             is_on_new_line: true,
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -75,7 +75,7 @@ impl<'a> Lexer<'a> {
         // Token's start and end are u32s, so limit for length of source is u32::MAX bytes.
         // Only a debug assertion is required, as parser checks length of source before calling
         // this method.
-        debug_assert!(source.len() <= MAX_LEN);
+        debug_assert!(source.len() <= MAX_LEN, "Source length exceeds MAX_LEN");
 
         let token = Token {
             // the first token is at the start of file, so is allows on a new line
--- a/crates/oxc_parser/src/lib.rs
+++ b/crates/oxc_parser/src/lib.rs
@@ -84,6 +84,10 @@ use crate::{
     state::ParserState,
 };
 
+/// Maximum length of source in bytes which can be parsed (~4 GiB).
+// Span's start and end are u32s, so size limit is u32::MAX bytes.
+pub const MAX_LEN: usize = u32::MAX as usize;
+
 /// Return value of parser consisting of AST, errors and comments
 ///
 /// The parser always return a valid AST.
@@ -135,8 +139,12 @@ pub struct Parser<'a> {
 impl<'a> Parser<'a> {
     /// Create a new parser
     pub fn new(allocator: &'a Allocator, source_text: &'a str, source_type: SourceType) -> Self {
+        // If source exceeds size limit, substitute a short source which will fail to parse.
+        // `parse()` will convert error to `diagnostics::OverlongSource`.
+        let source_text_for_lexer = if source_text.len() > MAX_LEN { "\0" } else { source_text };
+
         Self {
-            lexer: Lexer::new(allocator, source_text, source_type),
+            lexer: Lexer::new(allocator, source_text_for_lexer, source_type),
             source_type,
             source_text,
             errors: vec![],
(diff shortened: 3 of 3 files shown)


overlookmotel:
Actually maybe panicking isn't ideal. Should parser instead return an empty AST and an error (i.e. treat too-long code as a syntax error)?

overlookmotel:
Last commit changes behavior to return an error instead of panic if source size exceeds limit. I assume that's better.

2 problems remain:

1. I think it's ideal to have a `debug_assert!()` for source length in `Lexer::new` so that constraint is encapsulated in the lexer. But that makes the `overlong_source` test fail in a debug build (passes in release mode). That's the reason CI is failing. Any idea how to tackle that?
2. `legal_length_source` test takes about 5 secs to run in release mode, and 1 minute in debug mode. Is 5s too long for a test?

Could those tests be set to only run in release mode? Does CI runs tests against a release build?

Boshen:
I would just let it crash and change the public API

https://github.com/oxc-project/oxc/blob/0feeac540d2852bc158c8d0f0668788d50f144aa/crates/oxc_parser/src/lib.rs#L136-L137

to something like

```rust
/// # Panics
/// * When source length is 4GB (u32 max)
pub fn new(...) {
  assert!(source_text.len() < 4GB, "source text is too long ...");
}
```

Or play nice and return an empty AST and a nice error.

As for the tests, I'll just omit them because it's not that interesting to unit test a 4GB program.

overlookmotel:
I found solution to the above problems. Now squashed to a single commit and tests pass.

I went with returning empty AST and error approach. Currently I don't think parser has any code paths which lead to panic, and I worried introducing one might hurt performance.

Feel free to remove the 2nd test. That's the one that takes 5 secs to run, but it won't run on CI anyway.

However, I think it'd be ideal to keep the 1st test. `Atom` is going to rely on the invariant that a string which references source text cannot exceed 4 GiB, and if that invariant was violated, it could produce UB.

Note: To allow the `debug_assert!` for size in `Lexer::new`, there are 2 places in parser where the length is checked. However, the 2nd check only occurs on code path where parsing has failed, so won't affect performance of successful parsing, which I imagine is what matters.

Sorry for all the noise on this PR!

crates/oxc_parser/src/diagnostics.rs
```diff
@@ -4,6 +4,11 @@ use oxc_diagnostics::{
 };
 use oxc_span::Span;
 
+#[derive(Debug, Error, Diagnostic)]
+#[error("Source length exceeds 4 GiB limit")]
+#[diagnostic()]
+pub struct OverlongSource(#[label] pub Span);
```
Boshen, line 10: We may need to omit the span here, it'll choke miette, especially its code frame.

Also, imagine if oxc is run on the server, and someone sends in a `.js` file with gibberish in it, it'll definitely DOS the server.

crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -72,6 +72,11 @@ pub struct Lexer<'a> {
 #[allow(clippy::unused_self)]
 impl<'a> Lexer<'a> {
     pub fn new(allocator: &'a Allocator, source: &'a str, source_type: SourceType) -> Self {
+        // Token's start and end are u32s, so limit for length of source is u32::MAX bytes.
+        // Only a debug assertion is required, as parser checks length of source before calling
+        // this method.
+        debug_assert!(source.len() <= MAX_LEN);
```
Boshen, line 78: `debug_assert` has a second parameter where you can add some nice panic message.

overlookmotel: Added a commit with an error message.

crates/oxc_parser/src/diagnostics.rs
overlookmotel: Hmm. I'd not considered that.

I'm not familiar with miette. But if miette's code frame can't handle long sources, isn't that a problem that can be triggered by any long source containing a syntax error? e.g. a `.js` file which is 3.9 GiB (so "legal" length) with a syntax error in it somewhere?

IWANABETHATGUY: I can confirm that `mitte` will hang for about 40s, when the source code is 4GB, 
image
```rs
/*
You can derive a `Diagnostic` from any `std::error::Error` type.

`thiserror` is a great way to define them, and plays nicely with `miette`!
*/
use miette::{Diagnostic, SourceSpan};
use thiserror::Error;

#[derive(Error, Debug, Diagnostic)]
#[error("oops!")]
#[diagnostic(
    code(oops::my::bad),
    url(docsrs),
    help("try doing it better next time?")
)]
struct MyBad {
    // The Source that we're gonna be printing snippets out of.
    // This can be a String if you don't have or care about file names.
    #[source_code]
    src: NamedSource,
    // Snippets and highlights can be included in the diagnostic!
    #[label("This bit here")]
    bad_bit: SourceSpan,
}

/*
Now let's define a function!

Use this `Result` type (or its expanded version) as the return type
throughout your app (but NOT your libraries! Those should always return
concrete types!).
*/
use miette::{NamedSource, Result};
fn this_fails() -> Result<()> {
    // You can use plain strings as a `Source`, or anything that implements
    // the one-method `Source` trait.
    let src = "a".repeat(4294967296);
    let len = src.len();
    dbg!(&len);

    Err(MyBad {
        src: NamedSource::new("bad_file.rs", src),
        bad_bit: (9, 4).into(),
    })?;

    Ok(())
}

/*
Now to get everything printed nicely, just return a `Result<()>`
and you're all set!

Note: You can swap out the default reporter for a custom one using
`miette::set_hook()`
*/
fn pretend_this_is_main() -> Result<()> {
    // kaboom~
    this_fails()?;

    Ok(())
}
fn main() {
    pretend_this_is_main().unwrap();
}

```

IWANABETHATGUY: So, omit  the span should be a better idea indeed.

IWANABETHATGUY: After omitting the `SourceSpan` 
image

overlookmotel: Weird. I can't replicate that.

This branch adds 2 commits on top of this PR.
https://github.com/overlookmotel/oxc/tree/test-miette

* Commit 1: Add `test_miette` crate
* Commit 2: Remove the span from `OverlongSource` diagnostic

With the span present:

```sh
> cargo build --release -p test_miette
> time cargo run --release -p test_miette
created source: 408.9ms
parsed: 173.583µs
  × Source length exceeds 4 GiB limit
printed errors: 502.469416ms
cargo run --release -p test_miette  0.33s user 0.74s system 92% cpu 1.149 total
```

Without the span present:

```sh
> git reset --hard HEAD~1
> cargo build --release -p test_miette
> time cargo run --release -p test_miette
created source: 402.504209ms
parsed: 173.75µs
  × Source length exceeds 4 GiB limit
   ╭─[1:1]
 1 │ var x = 123456;
   · ▲
 2 │ var x = 123456;
   ╰────
printed errors: 504.958459ms
cargo run --release -p test_miette  0.33s user 0.74s system 92% cpu 1.148 total
```

No significant difference. This is on an M1 MacBook Pro.

NB: In this PR, the span used for the error is `Span::default()` i.e. start 0, end 0. So miette doesn't need to explore the whole source to find the line to print.

I'm happy to remove the span. Probably it's not helpful to print the first line of the file anyway.

But I'd interested to understand why @IWANABETHATGUY and I are seeing such different results. Would you have time to run this `test_miette` case on your machine and see if you see something very different from me?

IWANABETHATGUY: try use sourcecode without line feed

overlookmotel: If I change the source text from repetitions of `var x = 123456;\n` to `var x = 1234567;` i.e. whole file is single line of code rather than multiple lines, then miette dumps the entire file to stdout which yes, takes ages.

That's probably the reason for the difference between my observations and yours @IWANABETHATGUY.

I've pushed a commit to remove the span, since it's probably not helpful anyway.

However, please note that overwhelming miette is still possible by other means e.g. with this source text which is below u32::MAX:

```rs
let mut source_text = "a".repeat(4294967294); // 2 bytes less than 4 GiB
source_text.push('\0');
```

IWANABETHATGUY: I think a `SourceSpan` for  `Source length exceeds 4 GiB limit` error could not help too much, remove it should be a better .

IWANABETHATGUY (approved):

crates/oxc_parser/src/diagnostics.rs
overlookmotel: OK great. That's done, so marking this as resolved.

overlookmotel:
Shall I squash commits and rebase on main? Or will that happen automatically when merging?

Boshen:
Thank you for making Oxc safe!

PR #1874, opened 2024-01-02
fix(parser): terminate parsing if an EmptyParenthesizedExpression error occurs

### LINKED ISSUE #1870: EmptyParenthesizedExpression should generate non-recoverable error
EmptyParenthesizedExpression should generate a non-recoverable error since when it happens, the ast is no longer valid.

Dunqing:
close: https://github.com/oxc-project/oxc/issues/1870#issue-2061901976

### DIFF
--- a/crates/oxc_parser/src/js/expression.rs
+++ b/crates/oxc_parser/src/js/expression.rs
@@ -202,13 +202,14 @@ impl<'a> Parser<'a> {
         let mut expressions = list.elements;
         let paren_span = self.end_span(span);
 
+        if expressions.is_empty() {
+            return Err(diagnostics::EmptyParenthesizedExpression(paren_span).into());
+        }
+
         // ParenthesizedExpression is from acorn --preserveParens
         let expression = if expressions.len() == 1 {
             expressions.remove(0)
         } else {
-            if expressions.is_empty() {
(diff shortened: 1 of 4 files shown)


Dunqing:
* `main`
  * **PR #1874** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1874?utm_source=stack-comment).

crates/oxc_parser/src/js/expression.rs
```diff
@@ -202,13 +202,14 @@ impl<'a> Parser<'a> {
         let mut expressions = list.elements;
         let paren_span = self.end_span(span);
 
+        if expressions.is_empty() {
```
u9g, line 205: why did you move this diagnostic up?

Dunqing: Because if the expression is empty we need to return the diagnostic. So we don't neet to check `expression.len() == 1`

Boshen (approved):
This is producing better parser diagnostics 👍

PR #1877, opened 2024-01-02
feat(linter): no irregular whitespace

DeividAlmeida:
it was builded a loop that iterates in trivias.irregular-whitespace and return the warning if contain some specific sequence of chars.


DeividAlmeida:
Do you have any suggestion to improve the performance?

DeividAlmeida:
do you have any suggestion to improve the performance?

crates/oxc_linter/src/rules/eslint/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,436 @@
...
+impl Rule for NoIrregularWhitespace {
+    fn run_once(&self, ctx: &LintContext) {
+        let irregular_whitespaces = ctx.semantic().trivias().irregular_whitespaces().to_owned();
+        let source = ctx.source_text();
+
+        for irregular_whitespace in irregular_whitespaces {
+            let char = ctx
+                .source_text()
+                .get(irregular_whitespace.start as usize..irregular_whitespace.end as usize)
+                .unwrap();
```
Dunqing, line 45:
```suggestion
            let char = irregular_whitespace.source_text(ctx.source_text())
```

Dunqing (changes requested):
I don't think we need to check irregular_whitespaces again. we can loop over `irregular_whitespaces` and just call `ctx.diagnostic(NoIrregularWhitespaceDiagnostic(irregular_ whitespace))` and that's it!

crates/oxc_linter/src/rules/eslint/no_irregular_whitespace.rs
```diff
@@ -0,0 +1,436 @@
...
+    ///     return  42;
+    /// }
+    /// ```
+    NoIrregularWhitespace,
+    correctness
+);
+
+impl Rule for NoIrregularWhitespace {
+    fn run_once(&self, ctx: &LintContext) {
+        let irregular_whitespaces = ctx.semantic().trivias().irregular_whitespaces().to_owned();
```
Boshen, line 38: `to_owned` is a clone here I think, which is unecessary.

DeividAlmeida:
> I don't think we need to check irregular_whitespaces again. we can loop over `irregular_whitespaces` and just call `ctx.diagnostic(NoIrregularWhitespaceDiagnostic(irregular_ whitespace))` and that's it!

if we don't check it'll not pass and all tests, but i have some doubts about the tests relevance, can i proceed without checks?

Dunqing:
> > I don't think we need to check irregular_whitespaces again. we can loop over `irregular_whitespaces` and just call `ctx.diagnostic(NoIrregularWhitespaceDiagnostic(irregular_ whitespace))` and that's it!
> 
> if we don't check it'll not pass and all tests, but i have some doubts about the tests relevance, can i proceed without checks?

I think we can comment out the failed test first. And then figure out how to fix it.

Boshen:
> I think we can comment out the failed test first. And then figure out how to fix it.

I Agree. It seems like most of the tests can be fixed by fixing the lexer, it's not saving all the irregular whitespaces.

Dunqing (approved):

Boshen:
Thank you!

(I removed the redundant `.clone()`.)

Issue #1124, opened 2023-10-31
Running oxlint in pre-commit hook with lint-staged


JoseBuendiaDigio:
When I use `npx oxlint --fix` in list-staged and oxlint find errors it doesn't throws it. And seems that all is going well.

Maybe is due to in oxlint is a warning and I want that be an error, can I change this?

Boshen:
It seems like there is a problem with lint-staged passing in absolute paths.

A workaround right now is to call "oxlint ." from a function callback, e.g. https://github.com/web-infra-dev/rspack/blob/3c0c6ca9d9cc5f99919520288006e33f14867e57/lint-staged.config.mjs#L7C2-L7C39

I know this is troublesome, but I'll try and fix this when I get the time.

Boshen:
Oh I remember now! There is a rabbit hole with eslint + lint-staged + .eslintignore

* https://github.com/lint-staged/lint-staged#example-use-relative-paths-for-commands
* https://github.com/lint-staged/lint-staged#how-can-i-ignore-files-from-eslintignore
* `https://github.com/eslint/eslint/issues/12206`

tldr when you modify a `.eslintignor`ed file, we will still lint it because the the directory walker's default assumption is that you want to lint this file because the path is passed in explicitly.

clarkf:
So it seems like [lint-staged](https://github.com/lint-staged/lint-staged/) doesn't have much to do with this. It just passes the changed files explicitly to the linting command (e.g. `eslint`, `oxlint`). eslint/eslint#12206's author [was confused about](https://github.com/eslint/rfcs/pull/38#discussion_r334270836) the default behavior, which I totally get.

eslint v8.56.0 (latest version at time of writing) will emit a warning when an `.eslintignore`d file is explicitly passed as a CLI argument, but *will not* lint the file:

```sh
$ ./node_modules/.bin/eslint test.js

./oxlint-test/test.js
  0:0  warning  File ignored because of a matching ignore pattern. Use "--no-ignore" to override

✖ 1 problem (0 errors, 1 warning)

$ echo $?
0
```

oxlint (main, as of 2ac26303e0a4f5382dfe51039e8924e0ff32b58a) lets provided paths take precedence over ignore directives and *will lint* the files:

```sh
$ ~/src/oxc/target/debug/oxlint test.js

  ⚠ eslint(no-self-assign): this expression is assigned to itself
   ╭─[test.js:4:1]
 4 │
 5 │ foo = foo;
   ·       ───
[...]

PR #1878, opened 2024-01-03
Prioritize ignored paths when linting

clarkf:
As discussed in #1124

A common workflow is running `oxlint` automatically via something like lint-staged, which passes all changed file paths explicitly as CLI arguments. The ignore crate auto-whitelists explicit paths, overriding anything defined in an ignore file, leading to files that should be ignored causing CI and VC hooks to fail.

eslint avoids this issue by giving precedence to .eslintignore unless you pass `--no-ignore`.

Since `ignore` doesn't really give much control over this, the most effective solution seems to be filtering. ~~Once we've got a list of files to be linted, we can compare each against the .eslintignore and see if it really truly should be linted.~~

~~Unfortunately, this is sort of a naive solution. Ignored directories will still be walked, then filtered out after the fact. Changing this behavior at the walker level would affect the formatter, which probably isn't desirable.~~ (edit: The performance impact has been mostly resolved.)


Boshen (commented):
This is going to cause performance regression I think, given that `no_ignore = false` by default and thousands of paths collected from `Walk`.

Should we only trigger this code path when there are file paths passed from the command line? 

For a more granular control, should we introduce a `--lint-staged` flag to explicitly handle all the cases for lint staged?

Boshen:
I like the overall direction of this PR, it's just that eslint + lint-staged caused so much configuration confusion and wasted time for people, I have a feeling first-class support (via a cli flag) will be better than some random combinations of configuration options.

What do you think?

clarkf:
You're right that this introduces a performance regression, it definitely does. In most cases, the `paths` will be iterated a second time, filtered, and re-`Vec`ed. I think the pathological case might be overstated, though. For the overhead to be truly significant, the user would have to have passed a parent path to a significant number of ignored paths. Think something like `oxlint node_modules/`. This could be avoided by pushing the ignore filtering up into `Walk`, probably.

Adding the check for explicit paths is an easy win, though. I'll definitely do that, which should avoid the overhead for the common case.

>  ...eslint + lint-staged caused so much configuration confusion and wasted time for people, I have a feeling first-class support (via a cli flag) will be better than some random combinations of configuration options.

The confusion is definitely real, and ultimately I'd really defer to your judgement. My preference would be to stick with this direction and not implement git-specific logic. For what it's worth, this behavior (`.{tool}ignore` file taking precedence over provided paths) is how most of the common tools in this corner of the world (git, eslint, prettier) work.

Boshen:
Shall we try the following logic:

for each provided path, 
  if the path is a file (not a directory)
    check against ignore logic

---

FYI oxlint is designed to lint huge monorepos, with 10,000+ files being the norm.

clarkf:
I've approximated the requested logic, I think. Notably, it is now possible to pre-filter all paths, skipping linting altogether.

There's an outstanding question of user feedback. At no point during this process is the user informed that paths or filtered. eslint yields linter warnings, which doesn't seem perfect, but better than silence.

Boshen:
Thank you!

> There's an outstanding question of user feedback. At no point during this process is the user informed that paths or filtered. eslint yields linter warnings, which doesn't seem perfect, but better than silence.

We can have another issue / PR for this.

PR #1885, opened 2024-01-03
fix(parser): fix incorrectly identified directives

overlookmotel:
Parser incorrectly identifies string literals as directives if they follow after `import`s, `export`s, or decorators.

In all of these cases, `'use strict'` produces a directive in the AST, where it should be parsed as an `ExpressionStatement` containing a `StringLiteral`:

```js
import x from 'foo';
'use strict';
```

```js
export {x};
'use strict';
```

```js
@foo
'use strict';
```

[Playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAIC0gICAgICAgIC0G8rnONK89ITJ3zrK%2FUP7OmSZPgHQzStr3yMtwFTU%2BD1WPt09JgqZJLoYooydbGsM5vGcf34BnIA%3D)

This PR should fix that.

I'm not sure about the decorator case, though. I assume it's not a directive. But is prefixing a string literal with a decorator even legal syntax anyway?

And a side nit: If I'm reading it right, I don't think the `continue` statement in the decorator arm of the match does anything. Do I have that right?

Last question: Where does one go about putting a test? I guess these silly cases aren't covered by Babel etc's tests.

### DIFF
--- a/crates/oxc_parser/src/js/statement.rs
+++ b/crates/oxc_parser/src/js/statement.rs
@@ -44,13 +44,16 @@ impl<'a> Parser<'a> {
                 Kind::Import if !matches!(self.peek_kind(), Kind::Dot | Kind::LParen) => {
                     let stmt = self.parse_import_declaration()?;
                     statements.push(stmt);
+                    expecting_directives = false;
                 }
                 Kind::Export => {
                     let stmt = self.parse_export_declaration()?;
                     statements.push(stmt);
+                    expecting_directives = false;
                 }
                 Kind::At => {
                     self.eat_decorators()?;
+                    expecting_directives = false;
                     continue;
                 }
                 _ => {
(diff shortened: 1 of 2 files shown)


Boshen:
I couldn't think of any clever tricks so I added a plain unit test 😅

Boshen:
How did you catch this problem in the first place?

overlookmotel:
Thanks for merging.

Ah, so OXC is relying on Babel's tests (and other external sources of tests) for conformance testing, and doesn't have a large set of its own tests?

I was exploring the parser codebase, just trying to get familiar with it, and noticed that there seemed to be a mistake in the logic.

Boshen:
> Ah, so OXC is relying on Babel's tests (and other external sources of tests) for conformance testing, and doesn't have a large set of its own tests?

Yes, everything should come together when codegen is finished, and then we somehow get the test262 runner working in node.js (i.e. parse and print and run everything from test262).

Issue #503, opened 2023-06-30
☂️ typescript-eslint


nulladdict:
> [!WARNING]
> This comment is maintained by CI. Do not edit this comment directly.
> To update comment template, see https://github.com/oxc-project/oxc/tree/main/tasks/lint_rules

This is tracking issue for `@typescript-eslint/eslint-plugin`.

## Rules

- Kind: 🍀 = recommended | ⚠️ = deprecated
- Status: ✨ = implemented | 🚫 = not supported

| Name | Kind | Status | Docs |
| :--- | :--: | :----: | :--- |
| typescript/adjacent-overload-signatures | 🍀 | ✨ | https://typescript-eslint.io/rules/adjacent-overload-signatures |
| typescript/array-type | 🍀 |  | https://typescript-eslint.io/rules/array-type |
| typescript/ban-ts-comment | 🍀 | ✨ | https://typescript-eslint.io/rules/ban-ts-comment |
| typescript/ban-tslint-comment | 🍀 |  | https://typescript-eslint.io/rules/ban-tslint-comment |
| typescript/ban-types | 🍀 | ✨ | https://typescript-eslint.io/rules/ban-types |
| typescript/block-spacing | ⚠️ |  | https://typescript-eslint.io/rules/block-spacing |
| typescript/brace-style | ⚠️ |  | https://typescript-eslint.io/rules/brace-style |
| typescript/class-literal-property-style | 🍀 |  | https://typescript-eslint.io/rules/class-literal-property-style |
| typescript/class-methods-use-this |  |  | https://typescript-eslint.io/rules/class-methods-use-this |
| typescript/comma-dangle | ⚠️ |  | https://typescript-eslint.io/rules/comma-dangle |
| typescript/comma-spacing | ⚠️ |  | https://typescript-eslint.io/rules/comma-spacing |
| typescript/consistent-generic-constructors | 🍀 |  | https://typescript-eslint.io/rules/consistent-generic-constructors |
| typescript/consistent-indexed-object-style | 🍀 |  | https://typescript-eslint.io/rules/consistent-indexed-object-style |
| typescript/consistent-type-assertions | 🍀 |  | https://typescript-eslint.io/rules/consistent-type-assertions |
| typescript/consistent-type-definitions | 🍀 |  | https://typescript-eslint.io/rules/consistent-type-definitions |
| typescript/consistent-type-imports |  |  | https://typescript-eslint.io/rules/consistent-type-imports |
[...]

PR #1903, opened 2024-01-05
feat(linter): implement @typescript-eslint/triple-slash-reference

kaykdm:
implement @typescript-eslint/triple-slash-reference
Related issue: https://github.com/oxc-project/oxc/issues/503

original
- doc: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/triple-slash-reference.md
- code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/triple-slash-reference.ts


crates/oxc_linter/src/rules/typescript/triple_slash_reference.rs
```diff
@@ -0,0 +1,275 @@
...
+                    TSModuleReference::ExternalModuleReference(ref mod_ref) => {
+                        import_source_set.insert(mod_ref.expression.value.as_str());
+                    }
+                    TSModuleReference::TypeName(_) => {}
+                },
+                _ => {}
+            }
+        }
+
+        let reference_regexp =
```
IWANABETHATGUY, line 123: Should use `once_cell` wrap it.

kaykdm: added it!

crates/oxc_linter/src/rules/typescript/triple_slash_reference.rs
```diff
@@ -0,0 +1,277 @@
...
+                    "always" => TypesOption::Always,
+                    "never" => TypesOption::Never,
+                    "prefer-import" => TypesOption::PreferImport,
+                    _ => TypesOption::default(),
+                }),
+        }))
+    }
+    fn run_once(&self, ctx: &LintContext) {
+        let mut import_source_set = HashSet::new();
+        for node in ctx.nodes().iter() {
```
kaykdm, line 113: iterating nodes because we need to access `TSImportEqualsDeclaration`.
From my understanding, ast node is the only way to get  `TSImportEqualsDeclaration` while `ImportDeclaration` can also be accessed from `ctx.semantic().module_record().import_entries`

Boshen:
I think we can do this, given TSImportEqualsDeclaration and ImportDeclaration only appear at top level:

```
for each statement i in program.body
  if the statement is an import
    get all comments from i and i-1
      check for triple slash
```

kaykdm:
> I think we can do this, given TSImportEqualsDeclaration and ImportDeclaration only appear at top level:
> 
> ```
> for each statement i in program.body
>   if the statement is an import
>     get all comments from i and i-1
>       check for triple slash
> ```
> 
> For getting comments, use the range API from `BTreeMap`: `comments.range(span.start..span.end)`

@Boshen 
Thank you! 
I will fix that to use program.body instead of ctx.nodes().
However, even we use program.body to find import statement and iterate comment with range, we still need to check other comments because triple-slash directives are not always with import statement. 

Also, it made me realize that I don't need to iterate all comments since Triple-slash directives are only valid at the top of their containing file.
I should find the first program node to get start value and use it to iterate comments with range.
It seems like typescript-eslint uses program node's range to get comments. 
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/triple-slash-reference.ts#L100

kaykdm:
@Boshen 
I have done the following
- use program.body instead of ctx.nodes() to find TSImportEqualsDeclaration and ImportDeclaration
- find the start value of the first statement and use it for comments iteration range
- iterate comments first and then program.body to find import statement if necessary

I am not sure why performance regression is happening 🤔 
I think not that much processing is going on in both checker.ts and pdf.mjs

crates/oxc_linter/src/rules/typescript/triple_slash_reference.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        // We don't need to iterate over all comments since Triple-slash directives are only valid at the top of their containing file.
+        // We are trying to get the first statement start potioin, falling back to the program end if statement does not exist
+        let comments_range_end = program.body.first().map_or(program.span.end, |v| v.span().start);
+        let comments = ctx.semantic().trivias().comments();
+        let mut refs_for_import = HashMap::new();
+
+        for (start, comment) in comments.range(0..comments_range_end) {
+            let raw = &ctx.semantic().source_text()[*start as usize..comment.end() as usize];
+            if let Some(captures) = REFERENCE_REGEX.captures(raw) {
```
Dunqing, line 123: Can you try to figure it out without regex. Regex it will hurt performance I think

kaykdm: I fixed it!
Indeed, it got better
Screenshot 2024-01-08 at 22 44 35

crates/oxc_linter/src/rules/typescript/triple_slash_reference.rs
```diff
@@ -0,0 +1,327 @@
...
+                    }
+                    _ => {}
+                }
+            }
+        }
+    }
+}
+
+fn get_attr_key_and_value(raw: &str) -> Option<(String, String)> {
+    if !raw.starts_with('/') {
```
Dunqing, line 170:
```suggestion
    if !raw.starts_with('///') {
```
I think we can check if the beginning of raw contains `///`.

Dunqing (changes requested):

crates/oxc_linter/src/rules/typescript/triple_slash_reference.rs
kaykdm: all single line comment from trivias will be without `//`,
 so `/// <reference... />` will be `/ <reference... />`

Dunqing: I got it!

Dunqing (approved):

Issue #1880, opened 2024-01-03
Reduce size of `Token`?


overlookmotel:
`Token` struct used in the parser is currently 48 bytes. This could be reduced to 40 bytes with a couple of reasonably small changes, or reduced to 32 bytes with more invasive changes.

### Anatomy of a `Token`

`Token` is currently structured as:

```rs
// 48 bytes
pub struct Token<'a> {
  pub kind: Kind,             // 1 byte
  pub start: u32,             // 4 bytes
  pub end: u32,               // 4 bytes
  pub is_on_new_line: bool,   // 1 byte
  pub escaped: bool,          // 1 byte
  pub value: TokenValue<'a>,  // 32 bytes
  // padding for align 8      // 5 bytes
}

// 32 bytes
pub enum TokenValue<'a> {
  None,                       // 0 bytes
  Number(f64),                // 8 bytes
  BigInt(num_bigint::BigInt), // 32 bytes
  String(&'a str),            // 16 bytes
  RegExp(RegExp<'a>),         // 24 bytes
}

// 24 bytes
pub struct RegExp<'a> {
  pub pattern: &'a str,       // 16 bytes
  pub flags: RegExpFlags,     // 1 byte
}
```

### Reducing the size

To reduce `TokenValue` to 24 bytes:

#### 1. Box `BigInt`

`BigInt` is the largest variant of `TokenValue` enum. Replacing `BigInt` with `Box<BigInt>` would reduce that variant from 32 bytes to 8 bytes.

This would impose a cost of indirection when reading the value of `TokenValue::BigInt`, but BigInts are fairly rare in JS Code, so for most code this cost would be insignificant.

#### 2. Remove `RegExp` variant

Move `RegExpFlags` into `Token` (using one of the 5 bytes currently used for padding).

Then all that remains in `RegExp` is a `&str`, which is identical to the `String` variant. So those 2 variants can be merged.

Result is:

```rs
// 40 bytes
pub struct Token<'a> {
  pub kind: Kind,                  // 1 byte
  pub start: u32,                  // 4 bytes
  pub end: u32,                    // 4 bytes
  pub is_on_new_line: bool,        // 1 byte
  pub escaped: bool,               // 1 byte
  pub regexp_flags: RegExpFlags,   // 1 byte
  pub value: TokenValue<'a>,       // 24 bytes
  // padding for align 8           // 4 bytes
}

// 24 bytes
pub enum TokenValue<'a> {
  // <discriminant>                // 8 bytes
[...]

Issue #1800, opened 2023-12-24
Is there any documentation related to using vscode-extension oxc?


jjj-guo:

Boshen:
What kind of documentation are you looking for? I thought the extension is pretty easy to use, as there are no configurations. Feel free to reopen and discuss what we need to add.

PR #1924, opened 2024-01-07
refactor(parser): remove `TokenValue::BigInt` from `Token`

### LINKED ISSUE #1803: Memory leak in arena with `Atom`s
OXC appears to leak memory when source includes any strings over 24 bytes in length.

A demonstration here: https://github.com/overlookmotel/oxc/tree/leak-test

Checkout that branch, and `cargo run -p test_leak`.

### The problem

I believe the problem is this:

* Strings are stored as `Atom`s.
* `Atom` wraps `CompactString` from [compact_str](https://docs.rs/compact_str/).
* `CompactString` allocates to the heap if string is longer than 24 bytes.
* `Atom`s may appear anywhere in the AST, likely within a `Box`.
* Implementation of `Box` used for the arena has no `Drop` impl.
* So when parsed `Program` is dropped, `Atom`s don't get dropped, and the heap allocations `CompactString` made are not freed.

In the demo above, I've added `impl Drop for Atom` which logs when an `Atom` is dropped. It never gets called, demonstrating that the heap memory is leaked.

#### 1. Add `impl Drop` to `Box`

This would be like `bumpalo::collections::Box`.

Problem is that this would make dropping the AST more expensive, negating a major benefit of using an arena.

#### 2. Track and drop `Atom`s manually

Store pointers to all non-inline `Atoms` as they're created. A `Drop` impl on the arena could then drop them when the arena is dropped.

#### 3. Replace `CompactString`

Replace `CompactString` with a similar "small string" implementation which stores long strings in the bumpalo arena, instead of the general heap.

As well as solving the leak, I believe this would also be more performant:

1. Less heap allocations.
2. Strings would be stored close to the AST nodes containing them, producing better cache locality.

`CompactString` is quite complicated, but OXC only uses a small fraction of `CompactString`'s capabilities, so creating a custom small string implementation in OXC I don't think would be too tricky.

In addition:

`CompactString` is built as a replacement for std's `String`, and includes a `capacity` field to allow resizing. But OXC's `Atom` is immutable, so the `capacity` field is not required. All that's needed is a `Box<str>` equivalent.

So a new "small string" type could lose the `capacity` field and be reduced in size to 16 bytes, rather than the current 24. Obviously, only strings up to 16 bytes could then be stored inline, but still most JS identifiers are under 16 characters, and if string content is stored in the arena, storing longer strings "out of line" has less performance penalty.

### Further optimizations

For identifers without escapes (pretty much all identifiers) and string literals without escape sequences (most strings), `Atom` could just be a pointer to the string in source code, rather than copying the string into the arena.

Secondly, when lexing strings and identifiers containing escape sequences, `AutoCow` creates a string in the arena. That string can be pointed to by `Atom` rather than copied to another location in the arena.

In fact, escaped strings for some reason currently are stored in the arena *three* times, though I'm not sure why. Dump of arena contents after parsing `x = 'ABCDE\n';`:

```
|0c000000 00000000 486d602e 01000000| ........Hm`..... 00000000
|00000000 0e000000 00000000 00000000| ................ 00000010
|00000000 00000000 b06d602e 01000000| .........m`..... 00000020
|05000000 00000000 806d602e 01000000| .........m`..... 00000030
|00000000 0d000000 0098aa6f 01000000| ...........o.... 00000040
|04000000 0d000000 41424344 450a0000| ........ABCDE... 00000050
|00000000 00000000 00000000 000000c6| ................ 00000060
|f0010000 00004142 4344450a 42434445| ......ABCDE.BCDE 00000070
|00000000 01000000 00000000 01000000| ................ 00000080
|78000000 00000000 00000000 00000000| x............... 00000090
|00000000 000000c1 0090aa6f 01000000| ...........o.... 000000a0
```

Boshen:
This PR partially fixes #1803 and is part of #1880.

BigInt is removed from the `Token` value, so that the token size can be
reduced once we removed all the variants.

`Token` is now also `Copy`, which removes all the `clone` and `drop` calls.

This yields 5% performance improvement for the parser.

### DIFF
--- a/crates/oxc_parser/src/lexer/token.rs
+++ b/crates/oxc_parser/src/lexer/token.rs
@@ -5,7 +5,7 @@ use oxc_span::Span;
 
 use super::kind::Kind;
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct Token<'a> {
     /// Token Kind
     pub kind: Kind,
@@ -38,16 +38,15 @@ impl<'a> Token<'a> {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub enum TokenValue<'a> {
     None,
     Number(f64),
-    BigInt(num_bigint::BigInt),
     String(&'a str),
     RegExp(RegExp<'a>),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct RegExp<'a> {
     pub pattern: &'a str,
     pub flags: RegExpFlags,
@@ -64,13 +63,6 @@ impl<'a> TokenValue<'a> {
         match self {
             Self::Number(s) => *s,
             _ => unreachable!("expected number!"),
-        }
-    }
-
-    pub fn as_bigint(&self) -> num_bigint::BigInt {
-        match self {
-            Self::BigInt(s) => s.clone(),
-            _ => unreachable!("expected bigint!"),
         }
     }
 
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -236,7 +236,7 @@ impl<'a> Parser<'a> {
     pub(crate) fn checkpoint(&self) -> ParserCheckpoint<'a> {
         ParserCheckpoint {
             lexer: self.lexer.checkpoint(),
-            cur_token: self.token.clone(),
+            cur_token: self.token,
             prev_span_end: self.prev_token_end,
             errors_pos: self.errors.len(),
         }
--- a/crates/oxc_parser/src/js/expression.rs
+++ b/crates/oxc_parser/src/js/expression.rs
@@ -17,6 +17,7 @@ use super::{
 };
 use crate::{
     diagnostics,
+    lexer::parse_big_int,
     lexer::{Kind, TokenValue},
     list::SeparatedList,
     Context, Parser,
@@ -306,12 +307,12 @@ impl<'a> Parser<'a> {
             Kind::Hex => BigintBase::Hex,
             _ => return Err(self.unexpected()),
         };
-        let value = match self.cur_kind() {
-            kind if kind.is_number() => self.cur_token().value.as_bigint(),
-            _ => return Err(self.unexpected()),
-        };
+        let token = self.cur_token();
+        let src = self.cur_src().strip_suffix('n').unwrap();
+        let value = parse_big_int(src, token.kind)
+            .map_err(|err| diagnostics::InvalidNumber(err, token.span()))?;
         self.bump_any();
-        Ok(BigintLiteral { span: self.end_span(span), value, base })
+        Ok(self.ast.bigint_literal(self.end_span(span), value, base))
     }
 
     pub(crate) fn parse_literal_regexp(&mut self) -> Result<RegExpLiteral> {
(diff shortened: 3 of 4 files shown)


Boshen:
* `main`
  * **PR #1924** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1924?utm_source=stack-comment).

Dunqing:
Doesn't look like it has anything to do with #1800.

crates/oxc_parser/src/lexer/token.rs
```diff
@@ -5,7 +5,7 @@ use oxc_span::Span;
 
 use super::kind::Kind;
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Copy, Default)]
```
Boshen, line 8: ❗

Boshen:
cc @overlookmotel Checkout https://codspeed.io/oxc-project/oxc/branches/01-07-refactor_parser_parse_BigInt_lazily

Dunqing (approved):

Boshen:
image
image

overlookmotel:
Very nice performance gain! I assume that's due to reduced size of `Token` (and possibly also `Token` now being `Copy` and non-drop).

However...

I don't think `BigintLiteral` now containing a `Box<BigInt>` rather than just `BigInt` solves the memory leak.

To explain my thinking...

`oxc_allocator::Box` only contains a reference `&'alloc mut T`, not an owned `T`. So (if I understand right) it has no implicit `Drop` impl. And `oxc_allocator::Box` does not have an explicit `Drop` impl either.

This is different to Bumpalo's `Box` which ensures the content of the `Box` is dropped when the `Box` is with an explicit `Drop` impl:
https://docs.rs/bumpalo/latest/src/bumpalo/boxed.rs.html#337-344

So you could fix the memory leak by adding a `Drop` impl to `oxc_allocator::Box`. But I don't think you want to do that! If `Box` is drop, then `Expression` and `Statement` become drop, which means `Program` is drop too. So it would have the sizeable negative consequence that dropping an AST then involves traversing the entire AST to hunt for any `Box`es which contain drop types.

The fact that it *doesn't* do this is I think a significant factor in why OXC is so fast, and I had assumed `oxc_allocator::Box` was designed this way intentionally to achieve this.

The only viable solution I can see to the memory leak, which preserves OXC's speed, is to replace `BigInt` with an `ArenaBigInt<'a>` which stores its internal content in the arena - so then there's no data stored external to the arena that needs to be dropped.

Therefore, I don't think the change from `BigInt` to `Box<BigInt>` gains anything, and is in fact marginally less performant than before (not that it matters hugely because bigints are so rare in JS code).

All the (very significant) gain in this PR is due to the other change - removing `BigInt` from `Token`.

Or am I misunderstanding?

overlookmotel:
By the way, if just removing `Bigint` from `TokenValue` gets a 5% speedup, that's a very positive sign that removing `TokenValue` entirely is going to be MASSIVE.

Boshen:
I copied the arena implementation from jsparagus https://github.com/mozilla-spidermonkey/jsparagus/blob/master/crates/ast/src/arena.rs

We'll have to experiment with the drop problem. oxc is being integrated into Rolldown (an incremental compiler), we are going to get into real trouble if memory leaks 💥 

---

You're right about `Box<BigInt>`, let me drop the box from this PR for now, as it adds no functionality.

PR #1925, opened 2024-01-07
fix: use import plugin with relative path

### LINKED ISSUE #1887: Oxlint reports incorrect errors when it receives a relative path starting with `..`
```bash
test_repo % npx oxlint@latest --import-plugin ../test_repo 
× Failed to open file "test_repo/foo/bar/index.js" with error "No such file or directory (os error 2)"
help: Failed to open file "test_repo/foo/bar/index.js" with error "No such file or directory (os error 2)"
```
The error reported by `oxlint` appears to be incorrect. In fact, the file `test_repo/foo/bar/index.js` exists. It seems to be an issue related to `oxc_resolver`.

```js
// enhanced resolver
resolve.sync("../test_repo/foo/bar", "../main") // returns "../test_repo/foo/main.js"

// oxc_resolver
resolver.resolve("../test_repo/foo/bar", "../main").unwrap() // returns "test_repo/foo/main.js"
```

IWANABETHATGUY:
1.
Closed https://github.com/oxc-project/oxc/issues/1887
**Before**
image

**After**
image


crates/oxc_diagnostics/src/lib.rs
```diff
@@ -28,5 +28,4 @@ pub struct MinifiedFileError(pub PathBuf);
 
 #[derive(Debug, Error, Diagnostic)]
 #[error("Failed to open file {0:?} with error \"{1}\"")]
-#[diagnostic(help("Failed to open file {0:?} with error \"{1}\""))]
```
IWANABETHATGUY, line 31: The help message seems redundant

crates/oxc_linter/Cargo.toml
```diff
@@ -44,6 +44,7 @@ dashmap       = { workspace = true }
 convert_case  = { workspace = true }
 language-tags = { workspace = true }
 mime_guess    = { workspace = true }
+dunce = "1.0.4" # Normalize Windows paths to the most compatible format, avoiding UNC where possible
```
IWANABETHATGUY, line 47: Add this new dependency, but looks it will not increase the binary size, since it has been indirect imported before.

Boshen: This is because `oxc_resolver` already imports it.

crates/oxc_linter/src/service.rs
```diff
@@ -236,13 +236,15 @@ impl Runtime {
             self.update_cache_state(path);
 
             // Retrieve all dependency modules from this module.
-            let dir = path.parent().unwrap();
+            // Since the path maybe a relative path, we need to canonicalize it before pass into
+            // resolver, related to #1887
+            let dir = dunce::canonicalize(path.parent().unwrap()).unwrap();
```
Boshen, line 241: I don't think is the correct way, we shouldn't hit the disk again after getting paths from the walker.

IWANABETHATGUY: if path is `../test_dir/something` and specifier is `something.js`, the resolver will return `test_dir/something.js`, but the correct path should be `../test_dir/something/something.js`. you could see https://github.com/oxc-project/oxc/issues/1887#issue-2065115177 as a example.

IWANABETHATGUY: Do you mean we should `canonicalize` the path at the beginning?

IWANABETHATGUY: Or fixed the issue in `oxc_resolver` ?

PR #1942, opened 2024-01-08
fix(parser): unexpected ts type annotation in get/set

### LINKED ISSUE #1939: Syntax error in firefox, but not in oxc
```js
class A {
  get x: () => {
    return 5;
  }
}
```

Dunqing:
fix: https://github.com/oxc-project/oxc/issues/1939

### DIFF
--- a/crates/oxc_parser/src/js/class.rs
+++ b/crates/oxc_parser/src/js/class.rs
@@ -294,6 +294,10 @@ impl<'a> Parser<'a> {
             }
             Ok(definition)
         } else {
+            // getter and setter has no ts type annotation
+            if !kind.is_method() {
+                return Err(self.unexpected());
+            }
             let definition = self.parse_class_property_definition(
                 span,
                 key,
--- a/tasks/coverage/misc/fail/oxc-1942-1.ts
+++ b/tasks/coverage/misc/fail/oxc-1942-1.ts
@@ -1,0 +1,5 @@
+class Foo {
+  get x: () => {
+      return 5;
+  }
+}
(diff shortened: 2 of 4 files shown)


Dunqing:
* `main`
  * **PR #1942** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/1942?utm_source=stack-comment).

crates/oxc_parser/src/js/class.rs
```diff
@@ -294,6 +294,10 @@ impl<'a> Parser<'a> {
             }
             Ok(definition)
         } else {
+            // get / set no ts type annotation
```
Boshen, line 297:
```suggestion
            // getter and setter has no ts type annotation
```

crates/oxc_parser/src/lib.rs
```diff
@@ -330,6 +330,34 @@ mod test {
         assert_eq!(ret.errors.first().unwrap().to_string(), "Source length exceeds 4 GiB limit");
     }
 
+    #[test]
```
Boshen, line 333: Add then here: https://github.com/oxc-project/oxc/tree/main/tasks/coverage/misc/pass

We don't need to bloat this file with syntax tests

Dunqing: That's what I want to know

Boshen (approved):

PR #1946, opened 2024-01-08
feat(linter): implement react/require-render-return

kaykdm:
implement react/require-render-return
Related issue: https://github.com/oxc-project/oxc/issues/1022

original
- doc: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/require-render-return.md
- code: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/require-render-return.js

### DIFF
--- a/crates/oxc_linter/src/rules/react/require_render_return.rs
+++ b/crates/oxc_linter/src/rules/react/require_render_return.rs
@@ -1,0 +1,243 @@
+use oxc_ast::{
+    ast::{Argument, ClassElement, Expression, FunctionBody, ObjectPropertyKind},
+    AstKind,
+};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{
+    context::LintContext,
+    rule::Rule,
+    rules::eslint::array_callback_return::return_checker::{
+        check_statement, StatementReturnStatus,
+    },
+    utils::{is_es5_component, is_es6_component},
+    AstNode,
+};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error(
+    "eslint-plugin-react(require-render-return): Your render method should have a return statement"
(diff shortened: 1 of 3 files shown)


XantreGodlike:
I think you should add help info

crates/oxc_linter/src/rules/react/require_render_return.rs
```diff
@@ -0,0 +1,243 @@
...
+                                    None
+                                }
+                            }
+                            _ => None,
+                        })
+                        .find(|fn_body| !has_return_in_fn_body(fn_body))
+                    {
+                        ctx.diagnostic(RequireRenderReturnDiagnostic(fn_body.span));
+                    }
+                }
```
camc314, line 114: am reviewing fairly quickly, so correct me if i am wrong. But, won't this match

```ts
foo({ render: render: function() {} })
```
?

This is what the plugin reports for me (js not oxc)

Screenshot 2024-01-09 at 22 43 56


can we add a passing test case for the example above (unless i'm wrong lol)

camc314: lol i was wrong - should've read the code more carefully.

above:
```rs
if !is_es5_component(node) && !is_es6_component(node) {
    return;
}
```

`react.rs`
```rs
const PRAGMA: &str = "React";
const CREATE_CLASS: &str = "createReactClass";

pub fn is_es5_component(node: &AstNode) -> bool {
    let AstKind::CallExpression(call_expr) = node.kind() else { return false };

    if let Expression::MemberExpression(member_expr) = &call_expr.callee {
        if let Expression::Identifier(ident) = member_expr.object() {
            return ident.name == PRAGMA
                && member_expr.static_property_name() == Some(CREATE_CLASS);
        }
    }

    if let Some(ident_reference) = call_expr.callee.get_identifier_reference() {
        return ident_reference.name == CREATE_CLASS;
    }

    false
}
```

camc314 (approved):
Thank you!

PR #1950, opened 2024-01-08
ensure numbers without a raw are allocated during codegen

bmeck:
This was incorrectly using raw for dynamically generated numbers like in the minifier ( https://github.com/oxc-project/oxc/blob/6e0bd52af13e2add37a2bf444eedc02814f7082b/crates/oxc_minifier/src/compressor/fold.rs#L280 ).

This ensures they are dynamically generated during codegen.

This does not investigate why `just example minifier` does not take the `if MINIFY` branch.

### DIFF
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -947,7 +947,7 @@ fn need_space_before_dot<const MINIFY: bool>(bytes: &[u8], p: &mut Codegen<{ MIN
 impl<'a, const MINIFY: bool> Gen<MINIFY> for NumberLiteral<'a> {
     #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
     fn gen(&self, p: &mut Codegen<{ MINIFY }>, _ctx: Context) {
-        if MINIFY {
+        if MINIFY || self.raw == "" {
             p.print_space_before_identifier();
             let abs_value = self.value.abs();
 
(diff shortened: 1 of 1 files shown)


Boshen:
Oh hey a familiar face and twitter friend! Thank you for review my horrible code. Please let me know if you wish to work on oxc a bit more 😄 The project needs great experts like you.

> This does not investigate why just example minifier does not take the if MINIFY branch.

You'll need the `--whitespace` flag 

https://github.com/oxc-project/oxc/blob/6e0bd52af13e2add37a2bf444eedc02814f7082b/crates/oxc_minifier/examples/minifier.rs#L28-L30

crates/oxc_codegen/src/gen.rs
```diff
@@ -947,7 +947,7 @@ fn need_space_before_dot<const MINIFY: bool>(bytes: &[u8], p: &mut Codegen<{ MIN
 impl<'a, const MINIFY: bool> Gen<MINIFY> for NumberLiteral<'a> {
     #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
     fn gen(&self, p: &mut Codegen<{ MINIFY }>, _ctx: Context) {
-        if MINIFY {
+        if MINIFY || self.raw == "" {
```
Boshen, line 950:
```suggestion
        if MINIFY || self.raw.is_empty() {
```

bmeck:
@Boshen I would like to do a few things but might need a bit of guidance, I have a branch trying to do commutative `x === null` from `typeof x === "object" && !x` but am struggling a bit with how to pass data up the tree from the `visit_mut`. I'm a bit new to rust, but if this seems sane, you should be able to email me or DM me on Twitter.

PR #1961, opened 2024-01-09
feat(linter): eslint-plugin-jsx-a11y role-support-aria-props

re-taro:
## Summary

partof: #1141 

I re-implemented role-support-aria-props rule for jsx_a11y.

originals:

- docs: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-supports-aria-props.md
- code: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/role-supports-aria-props.js
- test: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/role-supports-aria-props-test.js


Boshen:
Thank you!

PR #1962, opened 2024-01-09
refactor(parser): reduce `Token` size from 32 to 16 bytes

Boshen:
Part of #1880

`Token` size is reduced from 32 to 16 bytes by changing the previous
token value `Option<&'a str>` to a u32 index handle.

It would be nice if this handle is eliminated entirely because
the normal case for a string is always `&source_text[token.span.start.token.span.end]`

Unfortunately, JavaScript allows escaped characters to appear in
identifiers, strings and templates. These strings need to be unescaped
for equality checks, i.e. `"\a"  === "a"`.

This leads us to adding a `escaped_strings[]` vec for storing these unescaped and allocated
strings.

Performance regression for adding this vec should be minimal because escaped strings are rare.

Background Reading:

* https://floooh.github.io/2018/06/17/handles-vs-pointers.html

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -24,13 +24,13 @@ use oxc_syntax::{
     },
     unicode_id_start::is_id_start_unicode,
 };
-pub use token::{Token, TokenValue};
 
 pub use self::{
     kind::Kind,
     number::{parse_big_int, parse_float, parse_int},
+    token::Token,
 };
-use self::{string_builder::AutoCow, trivia_builder::TriviaBuilder};
+use self::{string_builder::AutoCow, token::EscapedStringId, trivia_builder::TriviaBuilder};
 use crate::{diagnostics, MAX_LEN};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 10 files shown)


Boshen:
* `main`
  * **#1962** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/1962?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -850,16 +888,14 @@ impl<'a> Lexer<'a> {
...
+                        self.save_string(true, builder.finish_without_push(self));
                     }
                     self.current.chars.next();
                     return substitute;
                 }
                 '`' => {
                     if is_valid_escape_sequence {
-                        self.current.token.value =
-                            TokenValue::String(builder.finish_without_push(self));
+                        self.save_string(true, builder.finish_without_push(self));
```
Boshen, line 898: Note to self: all templates strings are stored because I can't figure out the correct way to handle the `cooked` value for normal template string and tagged template.

tagged`\x` is valid with a cooked value.
`\x` is a syntax error "Bad escape sequence in untagged template literal"

Boshen:
image

@overlookmotel I think this is the furthest I can go 😅

Issue #1969, opened 2024-01-09
Using an ESLint config file does not seem to affect oxlint rule count


jgeurts:
Hi! I'm trying to use an eslint config file to explicitly enable/disable oxlint rules. However, each time I run oxlint, it shows that it is using 150 rules even though ~180 rules are enabled in the config file.

fwiw, I think this file can be reduced substantially after #1966 is supported.

[oxlintrc.json](https://github.com/oxc-project/oxc/files/13874828/oxlintrc.json)

Boshen:
Does v0.2.0 fixes this problem?

jgeurts:
I admittedly had troubles with 0.2.0 loading any rules from the config file. It seemed to ignore the config file completely and ran with what appeared to be the default rule config that ships with oxlint. Unfortunately, I didn't have enough time to look into it further and reverted back to 0.1.2. 

I can try again and look into it a bit more in detail maybe tomorrow or Monday.

iSuslov:
Can confirm that something is off with respecting config. 

For instance running oxc like this `npx oxlint@latest` shows me `console.log` errors.

Then I created a config:
```json
{
  "rules": {
    "no-console": "off"
  }
}
``` 
Ran with `npx oxlint@latest -c .eslint.json` is still showing me console related linting errors. 

Setting `"no-console": "blablalba"` gives an error `Failed to parse rule severity, expected one of "allow", "off", "deny", "error" or "warn"` hence oxc reads the config.

Boshen:
> I admittedly had troubles with 0.2.0 loading any rules from the config file. It seemed to ignore the config file completely and ran with what appeared to be the default rule config that ships with oxlint. Unfortunately, I didn't have enough time to look into it further and reverted back to 0.1.2.
> 
> I can try again and look into it a bit more in detail maybe tomorrow or Monday.

@jgeurts Do you mind pasting your config? I'm lacking examples to test on 😅

Boshen:
v0.2.1 made a correction for this issue.

PR #1972, opened 2024-01-09
fix(linter): fix plugin name parsing when reading config file

haocheng6:
This PR fixes the plugin parsing logic when reading a config file.

Specifically, the plugin names defined in `RuleEnum`s use snake-case ("jsx_a11y") while in the config file they are written as "jsx-a11y". This inconsistency causes some rules to be filtered out.

I tested the change with the config json file provided in #1969.
- Before the change: `Finished in 21ms on 1 file with 157 rules using 8 threads.`
- After the change: `Finished in 23ms on 1 file with 178 rules using 8 threads.`

Related issue: #1969

### DIFF
--- a/crates/oxc_linter/fixtures/eslint_config.json
+++ b/crates/oxc_linter/fixtures/eslint_config.json
@@ -15,6 +15,8 @@
             {
                 "null": "ignore"
             }
-        ]
+        ],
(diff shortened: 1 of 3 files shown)


Boshen:
Thank you!

PR #1979, opened 2024-01-10
feat(linter): eslint-plugin-jest: no-test-return-statement

eryue0220:
Rule Detail: [link](https://github.com/eryue0220/eslint-plugin-jest/blob/main/src/rules/no-test-return-statement.ts)


mysteryven:
Welcome back!

crates/oxc_linter/src/rules/jest/no_test_return_statement.rs
```diff
@@ -0,0 +1,240 @@
...
+        return;
+    };
+
+    if ident.name != "expect" {
+        return;
+    }
+
+    ctx.diagnostic(NoTestReturnStatementDiagnostic(Span {
+        start: return_stmt.span().start,
+        end: call_expr.span.start - 1,
```
mysteryven, line 125: Only a small nitpick, this feels like will report position like:

```js
return      expect(1).toBe(1);
^^^^^^^^^^^
```

eryue0220: Yeah, I thought the rest of statement is valid and no need to report.

mysteryven (approved):
Thank you!

PR #1966, opened 2024-01-09
feat(linter): support overriding oxlint rules by eslint config

Boshen:
Previously if .eslintrc.json contains 

```
{
  "rules": {
    "no-empty": "off"
  }
}
```

Then no rules will be enabled.

---

This PR changes how we configure oxlint's rules.

The rules will start with the categories we apply, and then merge all the configurations stated in the `rules` field.

For example, if we begin with `-D correctness` with 80 rules, then

* `"no-empty-file": "off"` will remove the rule, yielding 79 rules
* `"no-empty": "error"` (restriction) will add the rule, yield 81 rules
* ""no-empty": ["error", { "allowEmptyCatch": true }]` add the rule's configuration

PR #1984, opened 2024-01-10
fix(linter): allow eslintrc to add rule when overriding

fi3ework:
Fix a bug introduced in https://github.com/oxc-project/oxc/pull/1966. Ideally, the rules in eslintrc should be merged into final rules as described:

> The rules will start with the categories we apply, and then merge all the configurations stated in the rules field.
>
> For example, if we begin with -D correctness with 80 rules, then
>
> "no-empty-file": "off" will remove the rule, yielding 79 rules
> "no-empty": "error" (restriction) will add the rule, yield 81 rules
> ""no-empty": ["error", { "allowEmptyCatch": true }]` add the rule's configuration

However, the implementation did not include the newly added rules in the eslintrc. As a test case and example, I added a new fixture to `﻿crates/oxc_cli/fixtures/no_undef`. No warn or deny will be found without this PR.

This is my first Rust PR ever. Any nitpicking suggestions are welcome. Thanks! 😊

### DIFF
--- a/crates/oxc_cli/fixtures/no_undef/eslintrc.json
+++ b/crates/oxc_cli/fixtures/no_undef/eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-undef": "warn"
+  }
+}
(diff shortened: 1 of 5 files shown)


Boshen:
Thank you!

PR #1988, opened 2024-01-11
feat(linter): eslint-plugin-import no-named-as-default-member rule

XantreDev:
- Docs: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default-member.md
- Code: https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-named-as-default-member.js
- Tests: https://github.com/import-js/eslint-plugin-import/blob/main/tests/src/rules/no-named-as-default-member.js

### DIFF
--- a/crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
+++ b/crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
@@ -1,0 +1,175 @@
+#![allow(clippy::significant_drop_tightening)]
+use std::collections::HashMap;
+
+use dashmap::mapref::one::Ref;
+use oxc_ast::{
+    ast::{BindingPatternKind, Expression, MemberExpression},
+    AstKind,
+};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{Atom, Span};
+use oxc_syntax::module_record::ImportImportName;
+
+use crate::{context::LintContext, rule::Rule};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint-plugin-import(no-named-as-default-member): {1:?} also has a named export {2:?}")]
+#[diagnostic(severity(warning), help("Check if you meant to write `import {{{2:}}} from {3:?}`"))]
+struct NoNamedAsDefaultMemberDignostic(#[label] pub Span, String, String, String);
+
+/// <https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default-member.md>
+#[derive(Debug, Default, Clone)]
+pub struct NoNamedAsDefaultMember;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Reports use of an exported name as a property on the default export.
+    ///
+    /// ### Example
+    ///
+    /// ```javascript
+    /// // ./bar.js
+    /// export function bar() { return null }
+    /// export default () => { return 1 }
+    ///
+    /// // ./foo.js
+    /// import bar from './bar'
+    /// const bar = foo.bar // trying to access named export via default
+    /// ```
+    NoNamedAsDefaultMember,
+    nursery
+);
+
+impl Rule for NoNamedAsDefaultMember {
+    fn run_once(&self, ctx: &LintContext<'_>) {
+        let module_record = ctx.semantic().module_record();
+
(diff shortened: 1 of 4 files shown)


XantreGodlike:
#1117

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
```diff
@@ -0,0 +1,166 @@
...
+                );
+            }
+        }
+
+        if has_members_map.is_empty() {
+            return;
+        };
+        let get_external_module_name_if_has_entry = |module_name: &Atom, entry_name: &Atom| {
+            has_members_map.get(&module_name).and_then(|it| {
+                if it.0.exported_bindings.contains_key(entry_name) {
```
mysteryven, line 76: Seem like will have a problem when encounter the same key in local scope.
```js
import foo from 'foo.js' 

{
   const foo = { bar: 'Hi'}
   console.log(foo.bar)
}
```

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
```diff
@@ -0,0 +1,166 @@
...
+
+            let specifier = import_entry.module_request.name();
+            let Some(remote_module_record_ref) = module_record.loaded_modules.get(specifier) else {
+                continue;
+            };
+
+            if !remote_module_record_ref.exported_bindings.is_empty() {
+                has_members_map.insert(
+                    import_entry.local_name.name(),
+                    (remote_module_record_ref, import_entry.module_request.name().to_string()),
```
mysteryven, line 66: Is the allocation of `to_string` need to be avoided?

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
XantreGodlike: Do you have an idea how to avoid it?

mysteryven: What about finding import statement in `ctx.symbols().resolved_references`?

https://github.com/oxc-project/oxc/blob/0d77e1e788440ca4d5fe047ca4bdea7a01ed6210/crates/oxc_linter/src/utils/jest.rs#L198-L210

XantreGodlike: I've added this test case to tests and there is no false positive

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
XantreGodlike: Started to use Atom, which is wrapper around `CompactString`, seems to be it will be more memory efficient.

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
mysteryven: I am a bit curious about why `foo.bar` not be visited when I print all the nodes in `ctx.semantic().nodes().iter()`. I'll check it after back to home.

XantreGodlike: I thought it because of block, but I've added test that checks for error with block and it's working fine.
```rs
        r"import baz from './named-exports'; 
        { 
            const a = baz.a;
        }",
```

mysteryven (approved):

XantreGodlike:
Will you merge it?)

XantreGodlike:
@mysteryven @Boshen

Boshen:
Thank you!

crates/oxc_linter/src/rules/import/no_named_as_default_member.rs
```diff
@@ -0,0 +1,175 @@
...
+            };
+        };
+
+        for item in ctx.semantic().nodes().iter() {
+            match item.kind() {
+                AstKind::MemberExpression(member_expr) => process_member_expr(member_expr),
+                AstKind::VariableDeclarator(decl) => {
+                    if let Some(Expression::MemberExpression(member_expr)) = &decl.init {
+                        process_member_expr(member_expr);
+                        return;
```
mysteryven, line 113: @XantreGodlike I know why, these `return` should be `continue`? 🙃 Change this will let the below case fail. 

```js
import baz from './named-exports';
  {
      const baz = {};
      const a = baz.a;
  }
```

XantreGodlike: Agree, will fix it in this week

XantreGodlike: #2071

PR #1999, opened 2024-01-11
refactor(parser): reduce work parsing regexps

overlookmotel:
#1926 produced a small performance regression because when parsing a regexp, some work is repeated.

Hopefully this PR may reverse it.

Perhaps implementation is a little messy, but posting it now as want to see what Codspeed says.

### DIFF
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -1,5 +1,6 @@
 //! Code related to navigating `Token`s from the lexer
 
+use oxc_ast::ast::RegExpFlags;
 use oxc_diagnostics::Result;
 use oxc_span::Span;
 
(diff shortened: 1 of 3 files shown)


overlookmotel:
Damn it! This makes no difference whatsoever!

Any idea why performance dropped slightly on #1926? Or was it just noise in the benchmarks?

Boshen:
> Damn it! This makes no difference whatsoever!

Any idea why performance dropped slightly on https://github.com/oxc-project/oxc/pull/1926? Or was it just noise in the benchmarks?

Just noise.

---

I forgot that regex has it's own standalone code path for getting the token, this PR is still good because it removes the weird re-parsing code from `parse_regex`.

There is no perf different because this is such a minor code path, and probably there's only a few regexes in the benchmark.

overlookmotel:
Thanks. One question: You said on #1926:

> In order to make things work nicely, the lexer will no longer recover
from a invalid regex.

Was that a desired outcome, or a compromise you had to make to get it working?

If the latter, I think the old behavior could now be restored after this PR by changing the `break`s back into `continue`s in the 2nd loop in `read_regex`:

https://github.com/oxc-project/oxc/blob/aa91fde1d956c0bff21036d834a1901597248fc5/crates/oxc_parser/src/lexer/mod.rs#L871-L888

Boshen:
> Was that a desired outcome, or a compromise you had to make to get it working?

A compromise. 

But leaving it as a hard error is also fine, I never had any rules for recoverability 😅

PR #2003, opened 2024-01-12
Commutative undefined minify

bmeck:
this should probably be squashed, i'm not familiar with code commit norms here. we can expand a bunch of the commutative compressions later.

### DIFF
--- a/crates/oxc_minifier/src/compressor/mod.rs
+++ b/crates/oxc_minifier/src/compressor/mod.rs
@@ -206,7 +206,6 @@ impl<'a> Compressor<'a> {
         if !self.options.typeofs {
             return
         }
-        // return None;
         match expr.operator {
             BinaryOperator::Equality | BinaryOperator::StrictEquality => {
                 let pair = self.commutative_pair((&expr.left, &expr.right),|a| {
--- a/crates/oxc_minifier/src/compressor/mod.rs
+++ b/crates/oxc_minifier/src/compressor/mod.rs
@@ -203,21 +203,6 @@ impl<'a> Compressor<'a> {
     /// Transforms `typeof foo == "undefined"` into `foo === void 0`
     /// Enabled by `compress.typeofs`
     fn compress_typeof_undefined(&self, expr: &mut BinaryExpression<'a>) {
-        // if expr.operator.is_equality() && self.options.typeofs {
-        //     if let Expression::UnaryExpression(unary_expr) = &expr.left {
-        //         if unary_expr.operator == UnaryOperator::Typeof {
-        //             if let Expression::Identifier(ident) = &unary_expr.argument {
-        //                 if expr.right.is_specific_string_literal("undefined") {
-        //                     let left = self.ast.identifier_reference_expression((*ident).clone());
-        //                     let right = self.ast.void_0();
-        //                     let operator = BinaryOperator::StrictEquality;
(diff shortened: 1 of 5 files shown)


crates/oxc_minifier/src/compressor/mod.rs
```diff
@@ -203,20 +203,77 @@ impl<'a> Compressor<'a> {
...
+        //                     let left = self.ast.identifier_reference_expression((*ident).clone());
+        //                     let right = self.ast.void_0();
+        //                     let operator = BinaryOperator::StrictEquality;
+        //                     *expr = BinaryExpression { span: SPAN, left, operator, right };
+        //                 }
+        //             }
+        //         }
+        //     }
+        // }
+        // return
```
Boshen, line 220: Let's remove the commented out code 😅

crates/oxc_minifier/src/compressor/mod.rs
```diff
@@ -203,20 +203,77 @@ impl<'a> Compressor<'a> {
...
+                    };
+                    *expr = cmp;
                 }
+
+            }
+            _ => {}
+        };
+    }
+
+    pub fn commutative_pair<A, F, G, RetF: 'a, RetG: 'a>(
```
Boshen, line 261:
```suggestion
   fn commutative_pair<A, F, G, RetF: 'a, RetG: 'a>(
```

Boshen (changes requested):

crates/oxc_minifier/src/compressor/mod.rs
Boshen, line 220:
```suggestion
```

Boshen (approved):

crates/oxc_minifier/src/compressor/mod.rs
```diff
@@ -203,20 +203,62 @@ impl<'a> Compressor<'a> {
...
-                    if let Expression::Identifier(ident) = &unary_expr.argument {
-                        if expr.right.is_specific_string_literal("undefined") {
-                            let left = self.ast.identifier_reference_expression((*ident).clone());
-                            let right = self.ast.void_0();
-                            let operator = BinaryOperator::StrictEquality;
-                            *expr = BinaryExpression { span: SPAN, left, operator, right };
+        if !self.options.typeofs {
+            return
+        }
+        // return None;
```
Boshen, line 209:
```suggestion
```

Boshen:
Thank you! A great start on learning Rust!

Boshen (commented):
Need a quick `cargo clippy` and `cargo format` fix.

crates/oxc_minifier/tests/oxc/folding.rs
```diff
@@ -8,6 +8,12 @@ fn addition_folding() {
     test("x+''", "x+'';");
 }
 
+#[test]
+fn typeof_folding() {
+    test("typeof x === 'undefined'", "void 0===x;");
```
magic-akari, line 13: This is dangerous. People use `typeof x` to determine if `x` exists.
For examples:

```JavaScript
if(typeof process === 'undefined') {
// not in node.js
}
```

If the definition of `x` is found in the context, then such an operation is feasible.

```JavaScript
function foo(x) {
    typeof x === 'undefined'; // We know that the variable x definitely exists.
}
```

bmeck:
@Boshen I've run those cmds and uploaded the new commit

PR #2010, opened 2024-01-13
perf(parser): reduce `Token` size from 16 to 12 bytes

Boshen:
I also had to change how the string for private identifiers are built, otherwise they will always be allocated.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -11,6 +11,7 @@ mod string_builder;
 mod token;
 mod trivia_builder;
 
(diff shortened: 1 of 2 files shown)


Boshen:
* `main`
  * **PR #2010** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2010?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2010?utm_source=stack-comment).

PR #2026, opened 2024-01-14
perf(parser): hint that end of file only happens once

overlookmotel:
Hint to branch predictor that lexer only gets to end of the file once.

I doubt this will actually make any difference. Just posting to see what Codspeed says.


overlookmotel:
Benchmark timed out downloading `antd.js`. Happened 3 times, but appears unrelated to content of this PR.

overlookmotel:
Ah no sorry, it was my fault. There was in infinite loop in my code.

overlookmotel:
Well after all that, the change makes no difference whatsoever!

PR #2027, opened 2024-01-14
fix(linter): jsx no undef match scope should check with ancestors

xxleyi:
when check undef in jsx-no-undef rule, the match one should be scope of current jsx or the ancestors scope of current jsx. cause I just begin to learn rust, I am not sure the code style and performance can be better, if there is any, I would change

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_no_undef.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_no_undef.rs
@@ -63,14 +63,13 @@ impl Rule for JsxNoUndef {
                 if ident.name.as_str() == "this" {
                     return;
                 }
-                let has_binding = ctx
-                    .symbols()
-                    .get_scope_id_from_name(&ident.name)
-                    .map_or(false, |scope_id| ctx.scopes().has_binding(scope_id, &ident.name));
-
-                if !has_binding {
-                    ctx.diagnostic(JsxNoUndefDiagnostic(ident.name.clone(), ident.span));
+                let jsx_scope_id = node.scope_id();
+                for scope_id in ctx.scopes().ancestors(jsx_scope_id) {
+                    if ctx.scopes().has_binding(scope_id, &ident.name) {
+                        return;
+                    }
                 }
+                ctx.diagnostic(JsxNoUndefDiagnostic(ident.name.clone(), ident.span));
             }
         }
     }
@@ -110,6 +109,8 @@ fn test() {
         "#,
             None,
         ),
+        ("var App; var React; enum A { App };  React.render(<App />);", None),
+        ("var React; enum A { App }; var App; React.render(<App />);", None),
     ];
 
     let fail = vec![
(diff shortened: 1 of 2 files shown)


Boshen:
I'm not sure whether enumerating the symbol table is a good idea, @Dunqing can you double check and find a better way?

Dunqing:
If too many symbols with the same name in the symbol table, enumerating the symbol table will affect performance. Instead, we should need to access the ancestor of the node's `scope_id`. 

Look like this
```rs
for scope_id in ctx.scopes().ancestors(jsx_scope_id) {
    if ctx.scopes().has_binding(scope_id, &ident.name) {
        return;
    }
}
ctx.diagnostic(JsxNoUndefDiagnostic(ident.name.clone(), ident.span));
```

Usually, the Component is defined or imported from other files in the root scope, So we can do a small optimization by checking if the binding is present in the root scope first.

Also, your test case is incorrect, the `App` variable is defined in terms of `var`, so the `App` variable can be accessed after `{ var App; }`.

```ts
var React; { var App; }; React.render(<App />);
```

You should change the `var App;` to `const App;`

xxleyi:
> If too many symbols with the same name in the symbol table, enumerating the symbol table will affect performance. Instead, we should need to access the ancestor of the node's `scope_id`.
> 
> Look like this
> 
> ```rust
> for scope_id in ctx.scopes().ancestors(jsx_scope_id) {
>     if ctx.scopes().has_binding(scope_id, &ident.name) {
>         return;
>     }
> }
> ctx.diagnostic(JsxNoUndefDiagnostic(ident.name.clone(), ident.span));
> ```
> 
> Usually, the Component is defined or imported from other files in the root scope, So we can do a small optimization by checking if the binding is present in the root scope first.
> 
> Also, your test case is incorrect, the `App` variable is defined in terms of `var`, so the `App` variable can be accessed after `{ var App; }`.
> 
> ```ts
> var React; { var App; }; React.render(<App />);
> ```
> 
> You should change the `var App;` to `let App;`

has_binding checked all scopes, not just current scope, which maybe also a bug.

as for test case, var should be alike let in strict mode, but I will double check this.

xxleyi:
> If too many symbols with the same name in the symbol table, enumerating the symbol table will affect performance. Instead, we should need to access the ancestor of the node's `scope_id`.
> 
> Look like this
> 
> ```rust
> for scope_id in ctx.scopes().ancestors(jsx_scope_id) {
>     if ctx.scopes().has_binding(scope_id, &ident.name) {
>         return;
>     }
> }
> ctx.diagnostic(JsxNoUndefDiagnostic(ident.name.clone(), ident.span));
> ```
> 
> Usually, the Component is defined or imported from other files in the root scope, So we can do a small optimization by checking if the binding is present in the root scope first.
> 
> Also, your test case is incorrect, the `App` variable is defined in terms of `var`, so the `App` variable can be accessed after `{ var App; }`.
> 
> ```ts
> var React; { var App; }; React.render(<App />);
> ```
> 
> You should change the `var App;` to `let App;`

Oh, I got your explanation, will change my PR.

Dunqing (approved):
Thank you!

PR #2029, opened 2024-01-14
perf(parser): lexer skip single space before tokens

overlookmotel:
It's very common for JS code to contain a single space between tokens (e.g. `x = y + 1`).

This PR makes `read_next_token()` handle this common case, by skipping a single space before tokenizing. The implementation is branchless.

Produces +1% speed up on parser benchmarks.

However, I'm not sure if this is a good idea! (hence marking this PR as draft)

`std::str::Chars` doesn't have an API for doing this branchlessly, so implementation uses some fairly gnarly unsafe code which relies on details of the internal implementation of `Chars`.

This may well not be worth the +1%. But opening this PR as a basis for discussion.

The same optimization could be achieved safely by replacing `Chars` iterator with an iterator over string bytes, and possibly that'd make other optimizations possible too.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -370,6 +370,55 @@ impl<'a> Lexer<'a> {
     /// Read each char and set the current token
     /// Whitespace and line terminators are skipped
     fn read_next_token(&mut self) -> Kind {
+        // Return `Kind::Eof` if end of file
+        let remaining = self.current.chars.as_str();
+        if remaining.is_empty() {
+            self.current.token.start = self.offset();
+            return Kind::Eof;
+        }
+
+        // Skip a single space.
+        // It's very common for JS code to contain a single space between tokens (e.g. `x = y + 1`),
+        // so handle this common case without branching.
+        // We skip a space without branching by manually mutating the `Chars` iterator's pointer
+        // - add 1 if next char is a space, or add 0 if it's not.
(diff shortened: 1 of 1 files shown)


Boshen:
May I refuse this PR due to the hard to understand code  + unsafe+ minor improvement?

Boshen:
If we really want go down this route ... getting rid of the `Chars` interface and work with raw bytes may be a better idea, but that'll be a huge change which will make the code look like C 😅

overlookmotel:
Yes you certainly may refuse it! As I mentioned, I also wasn't sure it was a good idea. But do you mind leaving this PR open for now? I may return to it and see if I can come up with a cleaner way of doing the same thing.

Boshen:
Sure :-)

overlookmotel:
I came up with another version which is also branchless but easier to follow and less unsafe:

```rs
let next_byte = remaining.as_bytes()[0];
let skip = usize::from(next_byte == b' ');
// SAFETY: `skip` is 1 only if next byte is an ASCII space, so can safely slice off `skip` bytes.
// Otherwise, `skip` is 0 in which case this is a no-op.
unsafe { self.current.chars = remaining.get_unchecked(skip..).chars() };
```

Unfortunately, this is a few more instructions than my initial quite unsafe version, and so does not produce a benefit.

In any case, the benefit of the original version has now been lessened due to other optimizations to `read_next_token()` made since I originally opened this.

So I'm going to close this.

PR #2030, opened 2024-01-14
feat: oxc regex parser

IWANABETHATGUY:


Boshen:
Looking good! Can you verify that this parser will handle all cases for the regex based eslint rules?

For these rules, we can make some of them built-in, and then emit the diagnostics in the linter rule. Or have a nice visitor to check these rules.

https://github.com/search?q=repo%3Aeslint%2Feslint%20RegExpValidator&type=code

Boshen:
continuing in https://github.com/oxc-project/oxc/pull/3824

PR #2039, opened 2024-01-15
refactor(parser): lexer handle unicode without branch

overlookmotel:
As suggested by @strager in https://github.com/oxc-project/oxc/pull/2025#pullrequestreview-1820273832, this PR adds `BYTE_HANDLERS` for first bytes of unicode characters.

This removes a branch from `read_next_token()` and produces a +1% speed-up on parser benchmarks.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -395,11 +395,7 @@ impl<'a> Lexer<'a> {
             }
 
             let byte = remaining.as_bytes()[0];
-            let kind = if byte < 128 {
-                BYTE_HANDLERS[byte as usize](self)
-            } else {
-                self.match_unicode_char()
-            };
+            let kind = BYTE_HANDLERS[byte as usize](self);
 
             if !matches!(
                 kind,
@@ -410,9 +406,7 @@ impl<'a> Lexer<'a> {
         }
     }
 
-    // `#[cold]` to hint to branch predictor that unicode identifiers and irregular whitespace are rare
-    #[cold]
-    fn match_unicode_char(&mut self) -> Kind {
+    fn unicode_char_handler(&mut self) -> Kind {
         let c = self.current.chars.clone().next().unwrap();
         match c {
             c if is_id_start_unicode(c) => {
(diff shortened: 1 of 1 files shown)


overlookmotel:
A couple of notes:

I've not investigated the performance of the other alternatives strager suggested. But at a guess, this will be the best for the fast path for ASCII characters (which I'd guess is 99%+ of code) as it removes a branch and adds no further instructions.

2 further optimizations are possible for handling unicode characters:

1. Separate handlers for the few bytes which can be the start of an irregular whitespace / line break unicode sequence (e.g. `c2`). The general handler for all other unicode bytes then can skip testing for whitespace.

2. Separate handlers for bytes which start a 2-byte, 3-byte, or 4-byte unicode sequence, to avoid a little work in `chars.next()`.

Both of these would involve quite a bit of code, and would only benefit the uncommon case of code which includes irregular whitespace or unicode identifiers, so I've left them for now.

Boshen:
I hate the irregular whitespaces so much, we would go so fast with SIMD without their precence.

Boshen:
I will take another look when I wake up, 1% isn't really justified for the change. 

The table was copied from the ratel parser, it had the unicode handlers, but I removed them without measure.

overlookmotel:
> 1% isn't really justified for the change

Do you mean it's not worth making the change for a 1% gain? Or that it's surprising this change produces a measurable gain?

Irregular whitespace is a massive pain. But I do still think there is scope for eating whitespace with SIMD. Could:

1. Test a block of 16 bytes to find the first byte that doesn't match space or tab.
2. If that byte is `\n` or `\r`, set `is_on_new_line = true`, and go again, this time skipping space, tab, `\n` and `\r`.
3. If byte is irregular whitespace (or 1st byte thereof for multi-byte sequences), handle that with `#[cold]` to minimize the cost of the branch, and loop back to the start.

Possibly could also do something similar for finding the end of identifiers, strings, and comments too.

I was planning to look into that.

aapoalas (approved):
Looks good idea and simple enough to me.

Boshen:
@overlookmotel FYI I had SIMD but removed due to usage of unsafe portable SIMD https://github.com/oxc-project/oxc/pull/645

Feel free to experiment with any safe API without nightly.

overlookmotel:
Thanks for merging.

And thanks for the note on SIMD. I'll take a look at your previous implementation and see if I can figure anything out.

PR #2040, opened 2024-01-15
build(linter): add new-react-perf rulegen task

DonIsaac:
Adds a new linter rule generation task, `just new-react-perf-rule`, for incorporating rules from [eslint-plugin-react-perf](https://github.com/cvazac/eslint-plugin-react-perf) into oxc.

Since this library has its own testing utilities and only 4 rules, I didn't bother writing code to port over test cases. If we deem this requisite I'll add this to the rulegen task.

### DIFF
--- a/justfile
+++ b/justfile
@@ -105,6 +105,9 @@ new-nextjs-rule name:
 new-jsdoc-rule name:
   cargo run -p rulegen {{name}} jsdoc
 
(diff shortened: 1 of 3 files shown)


Boshen:
> Since this library has its own testing utilities and only 4 rules

lol

PR #2046, opened 2024-01-16
perf(parser): lexer byte handlers consume ASCII chars faster

overlookmotel:
In the lexer, most `BYTE_HANDLER`s immediately consume the current char with `lexer.consume_char()`.

Byte handlers are only called if there's a certain value (or range of values) for the next char. This is their entire purpose. So in all cases we know for sure that we're not at EOF, and that the next char is a single-byte ASCII character.

The compiler, however, doesn't seem to be able to "see through" the `BYTE_HANDLERS[byte](self)` call and understand these invariants. So it produces very verbose ASM for `lexer.consume_char()`.

This PR replaces `lexer.consume_char()` in the byte handlers with an unsafe `lexer.consume_ascii_char()` which skips on to next char with a single `inc` instruction.

The difference in codegen can be seen here: https://godbolt.org/z/1ha3cr9W5 (compare the 2 x `core::ops::function::FnOnce::call_once` handlers).

Downside is that this does introduce a lot of unsafe blocks, but in my opinion they're all pretty trivial to validate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
+name = "assert-unchecked"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7330592adf847ee2e3513587b4db2db410a0d751378654e7e993d9adcbe5c795"
+
+[[package]]
 name = "async-trait"
 version = "0.1.77"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1735,6 +1741,7 @@ dependencies = [
 name = "oxc_parser"
 version = "0.5.0"
 dependencies = [
+ "assert-unchecked",
  "bitflags 2.4.1",
  "miette",
  "num-bigint",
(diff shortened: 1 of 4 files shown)


Boshen:
image

Boshen:
I shrank the `unsafe` surface area because all calls are self-contained (within the same file).

overlookmotel:
Interesting that it benefits TypeScript more than plain JS.

I see from your new commit that you're not keen on all the `unsafe`!

I totally get that. However, in my opinion, `consume_ascii_char()` should be an unsafe function. Calling it at EOF or if the next char isn't ASCII would be UB. If it's not an unsafe function, it'd be easy for someone to call it from somewhere else in lexer's codebase without considering the constraints they must uphold.

I did try to come up with another way to do it where the `unsafe` block is only within `read_next_token()`, rather than in every handler, but it's hard to find a good solution.

Boshen:
Perhaps we should move away from the `Chars` interface and code it like C 😅 

---

I guess this is why `unsafe` is frowned upon because now we are nitpicking where to put the `unsafe`. 

Both of our points are valid. 

The rule of thumb is to make the code more approachable for future readers, so feel free to disagree with my changes and make adjustments, I'm happy to merge.

Boshen:
> Out of interest, how do you feel about the kind of type system-based alternative I outlined above?

Feels like over abstracting with concepts, which isn't beginner friendly. The trying to keep it simple idea came from reading the Rome / Biome codebase (and partly swc as well).

overlookmotel:
I think I found a solution which squares the circle. No unsafe, but without sacrificing the performance gain. Will test it out and make a PR.

Issue #168, opened 2023-03-12
feat(ast,parser): parse jsdoc


Boshen:
I see there are some demand for reading jsdoc information, let's do this in two steps:

1. parse jsdoc into structured data, this should be lazy.
2. save them into trivias and have them accessible. I haven't figured out the best way to do this, but see https://github.com/Boshen/oxc/discussions/6 for disucssion.

---

Note: This is for parsing jsdoc comments to structured data. This is not about wiring up all the jsdoc comments into a document tree (which is the main purpose for jsdoc). 

To parse jsdoc comments, we need to:

1. add a new routine to parse jsdoc in oxc_parser, this should be a separate routine so we can parse jsdoc lazily
2. add the jsdoc AST to oxc_ast
3. setup jsdoc tests (I'm unsure on how to do this properly yet).

References:
* https://jsdoc.app/
* tests
  * https://github.com/jsdoc/jsdoc/tree/main/packages/jsdoc/test
  * typescript conformance tests: https://github.com/microsoft/TypeScript/tree/main/tests/cases/conformance/jsdoc

Boshen:
Prioritizing because @thepassle need this. See https://twitter.com/passle_/status/1637446645104164865

Boshen:
Update: 

The original intent was to come up with a generic comment attachment algorithm, where we attach comments to some AST node driven by a set of rules. After some research and reading about it in the ESLint and Babel codebase, I conclude that it is incomprehensible for me right now.

So instead, let's reduce the problem space down to "attach jsdoc comments to specific AST node". This will be much more approachable.

We will build the jsdoc data structure inside the semantic analyzer. When a jsdoc targeting AST node such as a Statement or Declaration is visited, we will ask `Trivias` to get us a leading comment. The jsdoc data structure will save the comment span and mark the Statement to indicate it has a jsdoc. A getter will be provided to parse the jsdoc lazily in a `OnceCell`.

ematipico:
I want to help somehow. What do you need or what can I do? :)

Boshen:
> I want to help somehow. What do you need or what can I do? :)

Hi Ema! I've forgotten everything we wrote about jsdocs, but would you be interested in getting a https://github.com/gajus/eslint-plugin-jsdoc rule to work? I can guide you on the missing pieces in the discord channel. I think we had some of the infra working, targeting a lint rule would be the easiest to fill in the gaps.

leaysgur:
Note about problems I found are 

[...]

PR #2064, opened 2024-01-17
fix(semantic): find the nearest valid jsdoc comment

leaysgur:
Current implementation fails to find jsdoc comment in some cases.

Like,

```js
/** @typedef {{ x: 1 }} X */
/** @type {X} */
let x;
```

For `let x;` it attempts to match `@typedef` comments, but nothing but whitespace can exist between them, so nothing matches as a result.

- - -

BTW, in this case, how should we get the first `@typedef` comment? It does not belong to any node but valid usage.


crates/oxc_semantic/src/jsdoc/mod.rs
```diff
@@ -17,7 +17,7 @@ pub struct JSDoc<'a> {
     docs: BTreeMap<Span, JSDocComment<'a>>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
leaysgur, line 20: This was added only for test, is it OK?

leaysgur:
It seems that tests containing comments in their test case are failing..., I'll check it later.

crates/oxc_semantic/src/jsdoc/mod.rs
Boshen: Yes this is ok.

Boshen:
> Besides Program, is there a larger AstKind that encompasses individuals, which may need to be excluded...?

It's easiest to work with a whitelist with unit tests added incrementally, otherwise you'll end up with unmanaged situations 😅 

Only excluding `Program` here may hurt performance a little bit because it'll check all AST nodes, which is a lot.

leaysgur:
@Boshen Thank you for your comment!

I tried a bit and realized that I should avoid whitelists, as they are still redundant to manage. 😓 

What I want to do here is to extend this original code,

https://github.com/leaysgur/oxc/blob/7a6261403db9163cad6638365c1f1053608f1061/crates/oxc_semantic/src/jsdoc/builder.rs#L57-L62

to allow normal comments as well as white-space.

I think I'm stuck on this now, do you have any good ideas?

leaysgur:
Closing once, I think I need more research and learning to cope with this. 🍃 

Current problems I found are https://github.com/oxc-project/oxc/issues/168#issuecomment-1900022366

PR #2071, opened 2024-01-17
fix(linter): eslint-plugin-import no-named-as-default-member rule

XantreDev:
Fixed false positives and early return @mysteryven 
#1988
https://github.com/oxc-project/oxc/pull/1988/files/64de6e9265d97d623d1f3a531b8c12a1a39dce02#r1453429430


mysteryven (approved):
Thank you!

Issue #2020, opened 2024-01-13
Automate the linter rules list


Boshen:
The process of tracking implemented and unimplemented rules should be automated.

A new task should be added to `tasks/`. It will pull all our supported plugins from their source, compare them against our implemented rules, and generate a markdown file with a tracking list.

We can initiate the task by only pulling from https://raw.githubusercontent.com/eslint/eslint/main/packages/js/src/configs/eslint-all.js and then expand to other plugins with follow up PRs.

eslint-all.js is a JS file but we can use oxc's parser to parse and extract all the rules.

Boshen:
Initial set of code has been merged: https://github.com/oxc-project/oxc/pull/2068

Pasting my reply here as well.

I think the next PR should pull in all the other plugins.

> If it is better to generate a file in Rust code, where should I output to?

What I have in mind is trigger the task in CI and update GitHub issues.

> Which one do you expect?

If we are going to update GitHub issues, it'll be easier with "for 1 plugin by specifying name
".

> Do we need a blacklist?(and how do I know the list?)

I think so, for example we won't be able to implement type checking lint rules. We'll need to hard code it I guess.

Boshen:
@leaysgur is coordinating the tasks.

leaysgur:
OK, I will add rough design docs and a roadmap like todo list later.
(For anyone who is interested in helping. 🙌)

leaysgur:
## What to make

For all rules of all plugins supported by `oxlint`, visualize their implementation progress with automatic workflows.

- A task that retrieves all rules for each plugin, compares them with the implementation status (file exists or not) and outputs the diff results.
- A task reflects its results to the umbrella issues of each plugin.
- GitHub Actions that are triggered by a merge to `main` of the PR and execute these tasks.

## Details
### Task to retrieve progress
- `cargo run -p lint_rules {PLUGIN_NAME}`
- Get a list of rules for each plugin(A)
  - Some may be all list, some need only recommended
  - The rules are in a JS configuration file, so parse with AST and enumerate
  - Exclude rules that are not supported by `oxlint`
- Compare with the list of implemented rules(= files or directories)(B) 
  - B in A ? ✅
  - (Some may be in B but not in A...?)
- Formatted into a Markdown TODO list

### Task to reflect progress
- `GITHUB_TOKEN=XXX cargo run -p lint_rules {PLUGIN_NAME} --apply`
  - If `GITHUB_TOKEN` is not enough, need to prepare a PAT?
[...]

PR #2081, opened 2024-01-19
(feat/tasks): refactor lint_rules for subsequent todos

leaysgur:
Part of #2020 

- Handle args and flags
- Split files
- Prettify CLI outputs

### DIFF
--- a/tasks/lint_rules/src/generate_list/eslint.rs
+++ b/tasks/lint_rules/src/generate_list/eslint.rs
@@ -1,0 +1,60 @@
+use std::collections::HashSet;
+use oxc_allocator::Allocator;
+use oxc_ast::{ast::ObjectPropertyKind, syntax_directed_operations::PropName, AstKind};
+use oxc_parser::Parser;
+use oxc_semantic::SemanticBuilder;
+use oxc_span::SourceType;
+
+pub const ORIGINAL_JS_SOURCE_URL: &str =
+    "https://raw.githubusercontent.com/eslint/eslint/main/packages/js/src/configs/eslint-all.js";
+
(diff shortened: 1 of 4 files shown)


leaysgur:
```sh
$ cargo run -p lint_rules eslint

⚠️ Rule: no-new-symbol is implemented but not in their lists.
⚠️ Rule: no-return-await is implemented but not in their lists.
⚠️ Rule: no-mixed-operators is implemented but not in their lists.
- [ ] accessor-pairs
- [x] array-callback-return
- [ ] arrow-body-style
- [ ] block-scoped-var

... ✂
```

FYI: These rules appear to be deprecated for now.

Boshen:
> FYI: These rules appear to be deprecated for now.

Nice catch! I'll look into their deprecation reasons.

Boshen:
For `list_implemented_rules`, you may use `RULES` instead of reading the rules.rs file

https://github.com/oxc-project/oxc/blob/b5b2ef34afaf3c81cd5ded8e52ab53bc0d24db9e/crates/oxc_linter/src/lib.rs#L36

Make `RULES` public and import it. Each rule has a `.name()` and `.plugin_name()` method.

https://github.com/oxc-project/oxc/blob/b5b2ef34afaf3c81cd5ded8e52ab53bc0d24db9e/crates/oxc_macros/src/declare_all_lint_rules/mod.rs#L79-L101

Feel free to follow this up with another PR.

---

Sorry I was in a rush yesterday so the message was unclear.

leaysgur:
> Sorry I was in a rush yesterday so

Never mind! I'll fix it later~ 🔜

PR #2082, opened 2024-01-19
perf(semantic): remove redeclare_variables

Dunqing:
I think we should remove the property used by a single rule

Now just a test

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_redeclare.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_redeclare.rs
@@ -7,7 +7,7 @@ use oxc_diagnostics::{
     thiserror::{self, Error},
 };
 use oxc_macros::declare_oxc_lint;
-use oxc_semantic::VariableInfo;
+use oxc_semantic::SymbolId;
 use oxc_span::{Atom, Span};
 
 use crate::{context::LintContext, globals::BUILTINS, rule::Rule};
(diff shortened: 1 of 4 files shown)


Dunqing:
* `main`
  * **PR #2082** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2082?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2082?utm_source=stack-comment).

Boshen:
Performance should be negligible because there aren't many re-declared variables.

I told the implementor to add this state because it's the easiest implementation. Happy to improve it.

Dunqing:
> Performance should be negligible because there aren't many re-declared variables.
> 
> I told the implementor to add this state because it's the easiest implementation. Happy to improve it.

I have figured out how to implement this rule and be compatible with more cases, but I don't want to add anything to `semantic`, is it possible for us to make the `run` method support `&mut self`.

Dunqing:
image

I am not certain if a 4% performance improvement can be achieved on every run

Boshen:
> is it possible for us to make the run method support &mut self

Impossible. The linter assumes all states are stabilized, making it `mut` will be chaotic 😰

PR #2090, opened 2024-01-19
feat/linter: (eslint-plugin-jest): no-restricted-matchers

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-restricted-matchers.ts)


mysteryven (approved):
Thank you! 

And sorry for the delay, I think I missed this notice :(

PR #2091, opened 2024-01-19
feat(linter): eslint-plugin-jest: no-restricted-jest-methods

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-restricted-jest-methods.ts)


mysteryven (approved):
Great work, Thank you!

PR #2097, opened 2024-01-20
chore(parser): explain the reason for omitting "}" and ">" in jsx text lexer

### LINKED ISSUE #2094: Bug in lexer's `read_jsx_child` function
Sorry not to be more specific in the issue description, but I don't really understand what I'm looking at, except to know it doesn't look right.

From `read_jsx_child`:

https://github.com/oxc-project/oxc/blob/a36813405b386815c1b4076885a0df07e605a6d9/crates/oxc_parser/src/lexer/mod.rs#L955-L971

I can see 3 problems here:

1. `push_matching()` is called with the 1st char twice.
2. `push_matching()` is a no-op as `force_allocation_without_current_ascii_char()` is never called.
3. `builder` is discarded without using it anyway.

I'm unclear whether the problem is that there's some code missing to handle escapes, or whether all it needs to do is consume the JSX Text, so `builder` serves no purpose and can be removed.

I'm further confused because the comment here doesn't match the code it accompanies. Are the characters in question `>` and `}` (as in the comment) or `<` and `{` (as in the code)?

https://github.com/oxc-project/oxc/blob/a36813405b386815c1b4076885a0df07e605a6d9/crates/oxc_parser/src/lexer/mod.rs#L959-L963

Boshen:
closes #2094

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -954,9 +954,11 @@ impl<'a> Lexer<'a> {
             }
             Some(_) => {
                 loop {
-                    // `>` and `}` are errors in TypeScript but not Babel
-                    // let's make this less strict so we can parse more code
-                    if matches!(self.peek(), Some('{' | '<')) {
+                    // The tokens `{`, `<`, `>` and `}` cannot appear in a jsx text.
+                    // The TypeScript compiler raises the error "Unexpected token. Did you mean `{'>'}` or `&gt;`?".
+                    // Where as the Babel compiler does not raise any errors.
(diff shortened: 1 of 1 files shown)


Boshen:
* `main`
  * **PR #2097** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2097?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2097?utm_source=stack-comment).

PR #2101, opened 2024-01-20
Benchmarks for lexer

overlookmotel:
This PR adds benchmarks for the lexer. I'm doing some work on optimizing the lexer and I thought it'd be useful to see the effects of changes in isolation, separate from the parser.

These benchmarks may not be ideal to keep long-term, but for now it'd be useful.

In order to do so, it's necessary for `oxc_parser` crate to expose the lexer, but have done that without adding it to the docs, and using an alias `__lexer`.

The results of the benchmarks are quite revealing. The flame graph shows that 49% of the entire run time lexing `checker.ts` is spent in `bumpalo::collections::string::String::push`. Not sure what's going on there!

### DIFF
--- a/crates/oxc_parser/src/lexer/token.rs
+++ b/crates/oxc_parser/src/lexer/token.rs
@@ -20,7 +20,10 @@ pub struct Token {
 
     /// True if the identifier / string / template kinds has escaped strings.
     /// The escaped strings are saved in [Lexer::escaped_strings] and [Lexer::escaped_templates] by
-    /// [Token::start]
+    /// [Token::start].
+    ///
+    /// [Lexer::escaped_strings]: [super::Lexer::escaped_strings]
+    /// [Lexer::escaped_templates]: [super::Lexer::escaped_templates]
     pub escaped: bool,
 }
 
--- a/crates/oxc_parser/src/lib.rs
+++ b/crates/oxc_parser/src/lib.rs
@@ -84,6 +84,12 @@ use crate::{
     state::ParserState,
 };
 
+// Expose lexer for benchmarks
+#[doc(hidden)]
+pub mod __lexer {
+    pub use super::lexer::{Kind, Lexer, Token};
+}
+
 /// Maximum length of source in bytes which can be parsed (~4 GiB).
 // Span's start and end are u32s, so size limit is u32::MAX bytes.
 pub const MAX_LEN: usize = u32::MAX as usize;
--- a/tasks/benchmark/Cargo.toml
+++ b/tasks/benchmark/Cargo.toml
@@ -43,6 +43,10 @@ harness = false
 name    = "minifier"
 harness = false
 
+[[bench]]
+name    = "lexer"
+harness = false
+
 [dependencies]
 oxc_allocator    = { workspace = true }
 oxc_linter       = { workspace = true }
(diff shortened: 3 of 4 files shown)


overlookmotel:
2 other notes:

1. If you're willing to merge this, please rename the commit. I wasn't sure how commits unrelated to the actual code should be named to fit the convention.

2. This benchmark isn't perfect. In the case of regexes, the lexer would initially return a `/` token, and then the parser would request it to be re-lexed, whereas this benchmark is just calling `next_token()` repeatedly, so will produce a different stream of tokens. But I think this is probably still a close enough approximation to be useful.

overlookmotel:
I don't know if I'm reading the flame graph correctly, but as well as the chunk of time taken by `String::push`, it looks like one of the `BYTE_HANDLERS` is taking up 85-90% of total lexing time. But:

1. Can't figure out which, as they're all called `core::ops::function::FnOnce::call_once` in the flamegraph.
2. I *think* that big block is a single handler, as there are other (thinner) handlers next to it in the graph too. But maybe I'm reading it wrong.

flame

Can anyone advise how to interpret this, and maybe suggest a "proper" flamegraph tool which will reveal which handler is which?

overlookmotel:
Ah, it's a mirage. I believe the problem is that this benchmark is not exercising the lexer properly, and it's probably interpreting `\` characters in comments etc as the start of identifiers, so throwing way more into `AutoCow` than it would do when the parser is running the lexer.

Am converting this PR to a draft until I can find a solution.

Boshen:
JavaScript lexical grammar has multiple goal symbols, which makes it impossible to lex all the tokens before hand. source: https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar

> There are several situations where the identification of lexical input elements is sensitive to the syntactic grammar context that is consuming the input elements. This requires multiple [goal symbols](https://tc39.es/ecma262/#sec-context-free-grammars) for the lexical grammar.

Hence the parser is the driver for the lexer, where as other programming languages is the reverse.

---

You may create micro-benchmarks for individual tokens. For example create a long string with some escape sequences, maybe using https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.combinations?

If this is temporary, you may just create a good old `cargo bench` benchmark and use it locally, instead of using criterion with codspeed.

Boshen:
Proper flame graph tools:

* mac instruments: https://oxc-project.github.io/docs/contribute/performance.html#profile
* linux: https://github.com/flamegraph-rs/flamegraph

overlookmotel:
Thanks Boshen. Yes, I was aware this benchmark wouldn't be perfect as it'd be "stupid" about JS grammar in a few places, but I hoped it'd be close enough to still be a useful measure.

I've pushed another commit which replaces the parts of the inputs which the lexer can't understand without the parser's help - escapes in regexes, and template strings. The results now look more sane, and I'd hope a decent approximation to the lexer's performance.

Anyway, I've learned a few things along the way: 1. AutoCow is pretty inefficient. 2. Bumpalo's `String::push_str` is very slow (because it pushes each byte of the string to the underlying `Vec` one by one rather than doing a mass copy).

Boshen (approved):
LGTM. Feel free to merge.

overlookmotel:
Thank you. I've merged it.

PR #2108, opened 2024-01-21
Added info about name to ModuleRecord export default

XantreDev:
https://github.com/oxc-project/oxc/issues/1117#issuecomment-1899575149


XantreGodlike:
Unnecessary for implementation

PR #2109, opened 2024-01-21
feat(linter): linter-eslint-plugin-import/no-named-as-default

XantreDev:
#1117 

Code: https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-named-as-default.js 
Tests: https://github.com/import-js/eslint-plugin-import/blob/main/tests/src/rules/no-named-as-default.js
Docs: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default.md

### DIFF
--- a/crates/oxc_linter/src/snapshots/no_named_as_default.snap
+++ b/crates/oxc_linter/src/snapshots/no_named_as_default.snap
@@ -1,0 +1,34 @@
+---
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 143
+expression: no_named_as_default
+---
+  ⚠ eslint-plugin-import(no-named-as-default): Using exported name "foo" as identifier for default export.
+   ╭─[index.js:1:1]
+ 1 │ import foo from "./bar";
+   ·        ───
+   ╰────
+  help: Use another name for default import to avoid confusion
+
+  ⚠ eslint-plugin-import(no-named-as-default): Using exported name "foo" as identifier for default export.
+   ╭─[index.js:1:1]
+ 1 │ import foo, { foo as bar } from "./bar";
+   ·        ───
+   ╰────
+  help: Use another name for default import to avoid confusion
+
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/no_named_as_default.snap
```diff
@@ -0,0 +1,34 @@
+---
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 143
+expression: no_named_as_default
+---
+  ⚠ eslint-plugin-import(no-named-as-default): Using exported name "foo" as identifier for default export.
```
Boshen, line 6: Do we need to say what the export name is from the other side? The current messages doesn't say about why this is bad.

Boshen, line 6: Do we need to say what the export name is from the other side? The current messages doesn't say about why this is bad.

XantreGodlike: I've used [same](https://github.com/import-js/eslint-plugin-import/blob/7a21f7e10f18c04473faadca94928af6b8e28009/tests/src/rules/no-named-as-default.js#L71C1-L72C1) error message as in eslint plugin. What do you want to add in it?

XantreGodlike: Maybe something like: `Module './bar' has named export 'foo'. Naming default import as 'foo' can be confusing`

Boshen: @XantreGodlike sounds good to me.

XantreGodlike: @Boshen fixed

Boshen:
Thank you! What a confusing rule 😅

Issue #1045, opened 2023-10-24
feat(codegen): sourcemap support


Boshen:
We need to generate sourcemap during codegen.

Boshen:
Mostly done, just need to polish things a little bit.

Issue #959, opened 2023-10-06
feat(parser): utf16 spans


Boshen:
💡 Utf-8 to Utf-16 conversion is costly for downstream tools, is it doable to offer utf-16 spans from our parser?

overlookmotel:
I don't think this would be too difficult to implement in the lexer, while it's running through the source character by character anyway.

But is the idea that spans always specify positions as UTF16? Or spans specify both UTF8 and UTF16 positions? Or one or the other depending on user choice?

As far as I understand, the source map "standard" states positions must be UTF16. But I have no idea what actual tools in the real world do.

By the way, if you operate on source as a slice of `u8` bytes, rather than `char`s, calculating UTF16 length can be quite fast. Here's some WIP code I wrote a while back while hacking on SWC, which uses SIMD:

```rs
/// Get length of string as number of UTF16 characters.
///
/// The implementation below is equivalent to
/// `s.chars().map(|c| c.len_utf16()).sum::<usize>()`
/// but hopefully faster.
///
/// It relies on translation of UTF8 coding to UTF16:
///
/// - 1-byte UTF8 sequence
///   = 1st byte `0xxxxxxx`
///   -> 1 x UTF16 char
///   UTF16 len = UTF8 len
/// - 2-byte UTF8 sequence
///   = 1st byte `110xxxxx`, remaining bytes `10xxxxxx`
///   -> 1 x UTF16
///   UTF16 len = UTF8 len - 1
/// - 3-byte UTF8 sequence
///   = 1st byte `1110xxxx`, remaining bytes `10xxxxxx`
///   -> 1 x UTF16
///   UTF16 len = UTF8 len - 2
/// - 4-byte UTF8 sequence
///   = 1st byte `1111xxxx`, remaining bytes `10xxxxxx`
///   -> 2 x UTF16
///   UTF16 len = UTF8 len - 2
///
/// So UTF16 len =
///   UTF8 len
///   minus count of UTF8 bytes indicating start of sequence 2 bytes or longer
///   minus count of UTF8 bytes indicating start of sequence 3 bytes or longer
///
/// See: https://stackoverflow.com/questions/5728045/c-most-efficient-way-to-determine-how-many-bytes-will-be-needed-for-a-utf-16-st
fn utf16_len(s: &str) -> usize {
  s.len()
    - s.bytes()
      .map(|c| (c >= 0xC0) as usize + (c >= 0xE0) as usize)
      .sum::<usize>()
}
```

[...]

PR #2110, opened 2024-01-21
WIP: feat(codegen): source map support using `source-map` crate from ezno

egasimus:
Related:
* https://github.com/oxc-project/oxc/issues/1045
* https://github.com/oxc-project/oxc/discussions/2105

This is far from complete, correct, or usable - but I'm opening the PR anyway to gather feedback and insights on how to implement it.

Work done so far:

* Created [a dev fork of `oxc`](https://github.com/egasimus/oxc/) and established dev workflow / test harness / inner loop [in my downstream repo](https://github.com/hackbg/ganesha/tree/feat/sourcemap).
* Added dependency on `source-map` crate (https://crates.io/crates/source-map, https://github.com/kaleidawave/source-map) and plugged it into `print_directives_and_statements_with_semicolon_order`.
* Replaced `Vec<u8>` in `struct Codegen` with `source_map::StringWithOptionalSourceMap`. (**I know using `String` in `oxc` is bad because allocations**; the `source_map` code would have to be adapted to use `Vec<u8>` instead.)
* Added `base64` to output the encoded sourcemap.
* Tested some (e.g. [test](https://github.com/hackbg/ganesha/blob/1d04135151b0694e0ce5d70501dd8e1cfba2ee59/src/test.rs#L10) -> [result](https://evanw.github.io/source-map-visualization/#NDM3AGNvbnN0IGJheiA9IHsKICAgICAgICBmb286JzEyMycKfTsKY29uc29sZS5sb2coewogICAgICAgIGJhejpiYXoKfSk7CgovLyMgc291cmNlTWFwcGluZ1VSTD1kYXRhOmFwcGxpY2F0aW9uL2pzb247YmFzZTY0LGV5SjJaWEp6YVc5dUlqb3pMQ0p6YjNWeVkyVlNiMjkwSWpvaUlpd2ljMjkxY21ObGN5STZXeUlvZFc1cmJtOTNiaUJ6YjNWeVkyVXBJbDBzSW5OdmRYSmpaWE5EYjI1MFpXNTBJanBiSW5SNWNHVWdSbTl2SUQwZ2MzUnlhVzVuWEc1Y2JtbHVkR1Z5Wm1GalpTQkNZWElnZXlCbWIyODZJRVp2YnlCOVhHNWNibU52Ym5OMElHSmhlam9nUW1GeUlEMGdleUJtYjI4NklGd2lNVEl6WENJZ2ZWeHVYRzVqYjI1emIyeGxMbXh2WnloN1ltRjZmU2xjYmlKZExDSnVZVzFsY3lJNlcxMHNJbTFoY0hCcGJtZHpJam9pUVVGQlFTeEJRVVZCT3p0QlFVVkJPenRCUVVWQkluMD0KMjM2AHsidmVyc2lvbiI6Mywic291cmNlUm9vdCI6IiIsInNvdXJjZXMiOlsiKHVua25vd24gc291cmNlKSJdLCJzb3VyY2VzQ29udGVudCI6WyJ0eXBlIEZvbyA9IHN0cmluZ1xuXG5pbnRlcmZhY2UgQmFyIHsgZm9vOiBGb28gfVxuXG5jb25zdCBiYXo6IEJhciA9IHsgZm9vOiBcIjEyM1wiIH1cblxuY29uc29sZS5sb2coe2Jhen0pXG4iXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IkFBQUEsQUFFQTs7QUFFQTs7QUFFQSJ9)) and saw how it doesn't generate the source map correctly at all :sweat_smile: (`source_map` expects you to pass newlines manually, while span start/end in Oxc are byte offsets?)

Next steps:
* Adapt `source_map` code to use `Vec<u8>` instead of string with manual newlines.
* I need to familiarize myself with how the source map format actually works - so far I've only used them.
* Possibly use a const generic to toggle source maps, like with `MINIFY`?
* **Alternate proposal:** I use source maps mostly to get correct line numbers in Node.js stack traces. I know this sounds weird (in the sense that nobody has done it yet?), but what about a codegen mode where TypeScript annotations are replaced with *whitespace* and the compiled code otherwise matches the input 1:1? We already have the span data :thinking:

Anyone got any pointers / opinions / denouncements? :grin:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,9 +105,9 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.21.5"
+version = "0.21.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
+checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
 name = "bitflags"
@@ -1397,6 +1397,7 @@ dependencies = [
 name = "oxc_codegen"
 version = "0.5.0"
 dependencies = [
+ "base64",
  "bitflags 2.4.1",
  "num-bigint",
  "oxc_allocator",
@@ -1404,6 +1405,7 @@ dependencies = [
  "oxc_parser",
  "oxc_span",
  "oxc_syntax",
+ "source-map",
 ]
 
 [[package]]
@@ -2386,6 +2388,12 @@ dependencies = [
  "libc",
  "windows-sys 0.48.0",
 ]
+
+[[package]]
+name = "source-map"
+version = "0.14.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f1c87aed86fe881b9427c79b2278f2e70b81dac06996dc789de58f2c1d79f7ec"
 
 [[package]]
 name = "spin"
--- a/crates/oxc_codegen/Cargo.toml
+++ b/crates/oxc_codegen/Cargo.toml
@@ -27,5 +27,8 @@ oxc_syntax    = { workspace = true }
 bitflags   = { workspace = true }
 num-bigint = { workspace = true }
 
+source-map = "0.14.9"
+base64     = "0.21.7"
+
 [dev-dependencies]
 oxc_parser = { workspace = true }
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -8,7 +8,7 @@ use oxc_syntax::{
     NumberBase,
 };
 use oxc_span::GetSpan;
-use source_map::SpanWithSource;
+use source_map::{SpanWithSource, ToString};
 
 use super::{Codegen, Context, Operator, Separator};
 
@@ -83,11 +83,11 @@ fn print_directives_and_statements_with_semicolon_order<const MINIFY: bool>(
             p.print_semicolon_if_needed();
         }
         let span = stmt.span();
-        p.sourcemap.add_mapping(&SpanWithSource {
+        p.code.add_mapping(&SpanWithSource {
             start: span.start,
             end: span.end,
             source: p.source_id
-        }, 0)
+        })
     }
 }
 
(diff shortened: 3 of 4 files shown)


Boshen:
Thank you so much for helping out! I always wanted to get sourcemap working but didn't find the time.

I think the overall direction is to get sourcemap working first, and then refactor it later.

> Possibly use a const generic to toggle source maps, like with MINIFY?

This isn't necessary yet.

> Alternate proposal: I use source maps mostly to get correct line numbers in Node.js stack traces. I know this sounds weird (in the sense that nobody has done it yet?), but what about a codegen mode where TypeScript annotations are replaced with whitespace and the compiled code otherwise matches the input 1:1? We already have the span data

There are features that move code around - e.g. import / exports, namespaces, enums.

---

It seems like the hardest part is to get line / column numbers working. If I understand correctly the exact problem is to find line / column number from a span?

Given the goal is to get it working first, we can start off by:

1. scan the original source
2. store all the new lines in a vec / BtreeMap
3. query the line number by span from the store data strucutre (binary search from vec or `range` from Btreemap

This is inefficient but correct, we need to get the confirm the sourcemap is correct.

Once this is in place and working, we'll then look at the overall infrastructure and find a better solution.

Boshen:
cc @overlookmotel this may interest you as well.

overlookmotel:
It *does* interest me, but I don't have time to delve into it right now.

NB: There is one further complication besides just identifying line breaks. Sourcemaps use UTF-16 indexes for column numbers, whereas OXC processes source code as UTF-8. The two are the same for ASCII characters, but differ for Unicode.

Thankfully non-ASCII characters are pretty rare in JavaScript/Typescript code, but they do need to be handled to guarantee sourcemap accuracy in all cases.

Some previous discussion of this on https://github.com/oxc-project/oxc/issues/959

Boshen:
esbuild uses another method for computing line column:

https://github.com/evanw/esbuild/blob/2af5ccf478812d2d7226ad4435d46fbbb3419a8c/internal/js_printer/js_printer.go#L4777-L4779

```
	// If we're writing out a source map, this table of line start indices lets
	// us do binary search on to figure out what line a given AST node came from
	LineOffsetTables []sourcemap.LineOffsetTable
```

Boshen:
The work has been picked up in #2524.

Thank you for initiating the effort.

PR #2111, opened 2024-01-21
refactor(syntax): reformat identifier byte tables

overlookmotel:
Not a substantive change, just reformatting the ASCII byte tables in `oxc_syntax`.

In my opinion, this makes them easier to read.


Boshen:
Nice touch! Thank you!

PR #2118, opened 2024-01-22
feat(tasks): Enable tasks/lint_rules CI 

leaysgur:
Part of #2020

👉🏻  See generated comment on #2117

### DIFF
--- a/.github/workflows/lint-rules.yml
+++ b/.github/workflows/lint-rules.yml
@@ -1,17 +1,17 @@
 name: Update lint_rules list
 
 on: push
-# on:
-#   pull_request:
-#     branches:
-#       - main
-#     types:
-#       - closed
+  branches:
+    - main
+  paths:
+    - 'crates/oxc_linter/**'
+    - 'tasks/lint_rules/**'
 
 jobs:
   lint_rules:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-rules.yml
```diff
@@ -0,0 +1,30 @@
+name: Update lint_rules list
+
+on: push
```
Boshen, line 3: Need to change to trigger on "push to main" only, when oxc_linter has a change.

leaysgur: 👍🏻 

👇🏻  was my original intention(currently commented out for debug)

```yml
on:
  pull_request:
    branches:
      - main
    types:
      - closed

jobs:
  lint_rules:
    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'A-linter')
```

or you prefer like this?

```yml
  push:
    branches:
      - main
    paths:
      - 'crates/oxc_linter/**'
```

leaysgur: (This one looks simpler. 🤔 This one looks simpler. I'll take this one.)

PR #2120, opened 2024-01-22
perf(parser): use vec instead of hashmap for storing escaped strings

### LINKED ISSUE #2016: Parser avoid hashing on writing/reading escaped strings
#2010 uses a clever wheeze to avoid storing an ID in `Token` for escaped strings.

It just occurred to me that maybe there's a further optimization:

When `get_string()` or `get_template_string()` is called, isn't the relevant `&str` always the last one to have been inserted into `escaped_strings` / `escaped_templates`?

If so, `escaped_strings` and `escaped_templates` could be `Vec`s, and `get_string()` would just do:

```rs
fn get_string(&self, token: Token) -> &'a str {
  if token.escaped {
    return self.escaped_strings.last().unwrap();
  }
  // ...
}
```

That would avoid hashing `start` on insertion and lookup.

Boshen:
closes #2016

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -69,11 +69,11 @@ pub struct Lexer<'a> {
     pub(crate) trivia_builder: TriviaBuilder,
 
     /// Data store for escaped strings, indexed by [Token::start] when [Token::escaped] is true
-    pub escaped_strings: FxHashMap<u32, &'a str>,
+    pub escaped_strings: Vec<&'a str>,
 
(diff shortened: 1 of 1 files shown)


Boshen:
* `main`
  * **PR #2120** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2120?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2120?utm_source=stack-comment).

Boshen:
Close due to lack of time.

PR #2128, opened 2024-01-22
refactor(syntax): ASCII tables static not const

overlookmotel:
A small nit: `ASCII_CONTINUE` is used in multiple places, so it's preferable that it's a `static` not a `const` so it only gets included in the binary once, rather than once for each usage.

Only makes a tiny difference on benchmarks (~0.1 ms), but change seems worthwhile anyway just to be more "good practice".


Boshen:
re graphite: did you setup the access token?

overlookmotel:
graphite

Argh! Maybe I didn't grant the right permissions on the access token. I'm not familiar with this stuff. I'll have to come back to this...

PR #2132, opened 2024-01-22
Refactor: Tidy functions used to read identifier characters

overlookmotel:
Just refactoring, related to functions in `oxc_syntax::identifier`, aiming to make the code a bit clearer. I had found it a bit confusing when trying to familiarise myself with the codebase.

Aims:

1. Consistent naming convention for `is_identifier_start`, `is_identifier_part`, and their variants.
2. Do not re-export `unicode_id_start` crate from `oxc_syntax`. It's confusing having 2 exports which do the same (or *almost* the same) from the same crate.
3. Use these functions with standardized names everywhere.

I've split into 4 commits as the changes affect different crates, but they're all on a similar theme. I'm not sure if you'd prefer this as 4 separate PRs, or if that'd be overkill.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -18,12 +18,10 @@ use oxc_allocator::{Allocator, String};
 use oxc_ast::ast::RegExpFlags;
 use oxc_diagnostics::Error;
 use oxc_span::{SourceType, Span};
-use oxc_syntax::{
-    identifier::{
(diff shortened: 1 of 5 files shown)


overlookmotel:
As with the other nits I've submitted recently, this is pre-amble for a much larger PR. Making these smaller changes separately aims reduce the diff of the large PR so it'll be easier to review.

Boshen:
Love this kind of refactoring where it improves consistency and removes a few loc! Keep'em coming!

PR #2134, opened 2024-01-23
feat(tasks): Add @typescript-eslint to lint_rules CI task

leaysgur:
Part of #2020

- Add `@typescript-eslint` plugin rules
- w/ refactoring
  - Fix compile errors to add other plugins
  - Remove not intended unsupported rule in `eslint`

(Use #2117 for updating for a while?)

### DIFF
--- a/tasks/lint_rules/src/generate_list/typescript.rs
+++ b/tasks/lint_rules/src/generate_list/typescript.rs
@@ -6,15 +6,15 @@ use oxc_span::SourceType;
 use std::collections::HashSet;
 
 const ORIGINAL_JS_SOURCE_URL: &str =
-    "https://raw.githubusercontent.com/eslint/eslint/main/packages/js/src/configs/eslint-all.js";
+    "https://raw.githubusercontent.com/typescript-eslint/typescript-eslint/main/packages/eslint-plugin/src/configs/recommended.ts";
 
-const UNSUPPORTED_RULES: &[&str] = &["yoda"];
+const UNSUPPORTED_RULES: &[&str] = &[];
 
 pub fn find_to_be_implemented_rules() -> Result<Vec<String>, String> {
     let source_text = super::fetch_plugin_rules_js_string(ORIGINAL_JS_SOURCE_URL)?;
 
     let allocator = Allocator::default();
-    let source_type = SourceType::default();
+    let source_type = SourceType::default().with_typescript(true);
     let ret = Parser::new(&allocator, &source_text, source_type).parse();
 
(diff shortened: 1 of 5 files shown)


leaysgur:
### Current generated outputs

```
- [x] ban-ts-comment
- [x] ban-types
- [ ] no-array-constructor
- [x] no-duplicate-enum-values
- [x] no-explicit-any
- [x] no-extra-non-null-assertion
- [ ] no-loss-of-precision
- [x] no-misused-new
- [x] no-namespace
- [x] no-non-null-asserted-optional-chain
- [x] no-this-alias
- [x] no-unnecessary-type-constraint
- [x] no-unsafe-declaration-merging
- [ ] no-unused-vars
- [x] no-var-requires
- [x] prefer-as-const
- [x] triple-slash-reference
```

### Warnings
```
⚠️ Rule: no-empty-interface is implemented but not in their lists.
⚠️ Rule: adjacent-overload-signatures is implemented but not in their lists.
⚠️ Rule: consistent-type-exports is implemented but not in their lists.
```

tasks/lint_rules/src/generate_list/typescript.rs
```diff
@@ -0,0 +1,67 @@
+use oxc_allocator::Allocator;
+use oxc_ast::{ast::ObjectPropertyKind, syntax_directed_operations::PropName, AstKind};
+use oxc_parser::Parser;
+use oxc_semantic::SemanticBuilder;
+use oxc_span::SourceType;
+use std::collections::HashSet;
+
+const ORIGINAL_JS_SOURCE_URL: &str =
```
Boshen, line 8: I wonder if there is another venue for getting the "all" list marking which ones are recommended. Otherwise we'll need to parse two files 😅

Boshen:
Ahh ... `no-array-constructor` is implemented by eslint/no-array-constructor, I'll need to fix this somehow 😢

Boshen (approved):

tasks/lint_rules/src/generate_list/typescript.rs
leaysgur: > need to parse two files

How about using JS runtime(instead of Rust + AST parsing)?
Actually clone plugin repo and `import`(or `require`) all rule files and check its `meta.recommended` prop.

At first I thought this was a wild idea, but then I found out that `jest` does not provide a static rule list, so maybe...?

> https://github.com/jest-community/eslint-plugin-jest/blob/main/src/index.ts

(Here, as with `eslint`, does this mean support for `all` as well as `recommended`? 👀 )

PR #2149, opened 2024-01-23
feat(linter) eslint-plugin-react/jsx-no-target-blank

hjiog:
#1022 try to implement `jsx-no-target-blank`

* code: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/jsx-no-target-blank.js
* doc: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md
* test: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/jsx-no-target-blank.js

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
@@ -58,9 +58,9 @@ impl Default for JsxNoTargetBlank {
 impl JsxNoTargetBlank {
     fn diagnostic(&self, span: Span, ctx: &LintContext) {
         if self.allow_referrer {
-            ctx.diagnostic(JsxNoTargetBlankDiagnostic::NoTargetBlankWithoutNoopener(span))
+            ctx.diagnostic(JsxNoTargetBlankDiagnostic::NoTargetBlankWithoutNoopener(span));
         } else {
-            ctx.diagnostic(JsxNoTargetBlankDiagnostic::NoTargetBlankWithoutNoreferrer(span))
+            ctx.diagnostic(JsxNoTargetBlankDiagnostic::NoTargetBlankWithoutNoreferrer(span));
         }
     }
 }
@@ -128,14 +128,13 @@ fn check_href(
         matches!(enforce_dynamic_links, EnforceDynamicLinksEnum::Never);
     match attribute_value {
         JSXAttributeValue::StringLiteral(str) => {
-            is_external_link = check_is_external_link(&str.value)
+            is_external_link = check_is_external_link(&str.value);
         }
-        JSXAttributeValue::ExpressionContainer(expr) => match &expr.expression {
-            JSXExpression::Expression(expr) => {
-                match_href_expression(expr, &mut is_external_link, &mut is_dynamic_link)
+        JSXAttributeValue::ExpressionContainer(expr) => {
+            if let JSXExpression::Expression(expr) = &expr.expression {
+                match_href_expression(expr, &mut is_external_link, &mut is_dynamic_link);
             }
-            _ => {}
-        },
+        }
         _ => {}
     };
     if is_enforce_dynamic_links_never {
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/jsx_no_target_blank.snap
```diff
@@ -0,0 +1,405 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: jsx_no_target_blank
+---
+  ⚠ eslint-plugin-react(jsx-no-target-blank): Using target=`_blank` without rel=`noreferrer` (which implies rel=`noopener`) is a security risk in older browsers: see https://mathiasbynens.github.io/
+  │ rel-noopener/#recommendations
+   ╭─[jsx_no_target_blank.tsx:1:1]
+ 1 │ <a target="_blank" href="https://example.com/1"></a>
```
mysteryven, line 8: Shall we report on `target="_blank"` ? Seems it's where the actual error occur.

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
```diff
@@ -0,0 +1,658 @@
...
+                match_target_expression(expr)
+            } else {
+                default
+            }
+        }
+        _ => default,
+    }
+}
+
+impl Rule for JsxNoTargetBlank {
```
mysteryven, line 253: What about moving this part below the `declare_oxc_lint`, we can see the main part of this rule at first glance.

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// /// incorrect
+    /// var Hello = <a target='_blank' href="https://example.com/"></a>
+    /// var Hello = <a target='_blank' href={dynamicLink}></a>
+    /// ```
+    JsxNoTargetBlank,
+    correctness
+);
+
+fn check_is_external_link(link: &Atom) -> bool {
+    lazy_static! {
```
mysteryven, line 99: Would it be possible to replace`Regex` with something else? It may hurt performance. :(

crates/oxc_linter/src/snapshots/jsx_no_target_blank.snap
hjiog: image
I tested the rule in VSCode, and it seems to also report on open JSX elements.

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
hjiog: here doesn't seem to be a solid implementation for this at the moment, I'll make a note and add a TODO.

crates/oxc_linter/src/snapshots/jsx_no_target_blank.snap
mysteryven: We don't need to keep same with it, do you think it will be easier to understand the error when focus on it?

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
```diff
@@ -0,0 +1,660 @@
...
+    /// var Hello = <a target='_blank' href={dynamicLink}></a>
+    /// ```
+    JsxNoTargetBlank,
+    correctness
+);
+
+impl Rule for JsxNoTargetBlank {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::JSXOpeningElement(jsx_ele) = node.kind() {
+            // TODO need to read from configuration: `settings: { linkComponents: ['Link'] }`
```
camc314, line 101: use `get_element_type` or is that just for jsx_a11y rules?

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
mysteryven: What about checking whether a link starts with `"\\"` or a word?

crates/oxc_linter/src/rules/react/jsx_no_target_blank.rs
hjiog: yes, I think get_element_type is just for jsx_a11y rules，the rule setting is like this:
image

mysteryven (approved):
Thank you!

PR #2154, opened 2024-01-23
up(cli): support --version

DonIsaac:
Adds support for `--version` to all commands in `oxc_cli`. I've set `oxc_cli`'s version to match the other packages in the monorepo (0.5.0). Let me know if you'd like the version set to some other value.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1386,7 +1386,7 @@ dependencies = [
 
 [[package]]
 name = "oxc_cli"
-version = "0.0.0"
+version = "0.5.0"
 dependencies = [
  "bpaf",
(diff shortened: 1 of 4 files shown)


Boshen:
The version should match https://github.com/oxc-project/oxc/blob/1ee6d8cea93827351aaab925d4290f57639c5bf9/tasks/coverage/src/test262.rs#L173-L174

I think we need to compile it into the binary by supplying it in CI

https://github.com/oxc-project/oxc/blob/7f89bfea0d808f701c5328decedb42b068db4db7/.github/workflows/release_oxlint.yml#L41

PR #2155, opened 2024-01-23
ci(oxlint): run smoke test against both glibc and musl

xc2:
fixes https://twitter.com/boshen_c/status/1749755460750836174

demo: https://github.com/xc2/ci-demo/actions/runs/7629534563/job/20783146502


Boshen:
Nice! Thank you!

PR #2157, opened 2024-01-24
fix(codegen): print `Directive` original string

underfin:
>  A Use Strict Directive may not contain an EscapeSequence or LineContinuation.

It is `Use Strict Directive` spec, but the `expression` of `Directive` isn't original string value, it has error if using it to codegen, so here using `directive` of `Directive` to codegen and not to escape it. Here is crashed test cases. 

``` js
'use str\
ict';
```
The babel will print the original string, I follow it and avoid using `print_str` because it will escape string.

I also changed some code using the `expression` of `Directive` to check  `Use Strict Directive` .

### DIFF
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -97,7 +97,7 @@ impl<const MINIFY: bool> Gen<MINIFY> for Directive {
         // See https://github.com/babel/babel/blob/main/packages/babel-generator/src/generators/base.ts#L64
         let quote = choose_quote(self.directive.as_str());
         p.print(quote as u8);
-        p.print_str(self.directive.as_str().to_string().as_bytes());
+        p.print_str(self.directive.as_bytes());
         p.print(quote as u8);
         p.print_semicolon();
     }
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -95,7 +95,7 @@ impl<const MINIFY: bool> Gen<MINIFY> for Directive {
         // A Use Strict Directive may not contain an EscapeSequence or LineContinuation.
         // So here should print original `directive` value, the `expression` value is escaped str.
         // See https://github.com/babel/babel/blob/main/packages/babel-generator/src/generators/base.ts#L64
-        print_origin_str(self.directive.as_str(), p);
+        p.print_str(self.directive.as_str().to_string().as_bytes());
         p.print_semicolon();
     }
 }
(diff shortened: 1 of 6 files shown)


Boshen (approved):

crates/oxc_codegen/src/gen.rs
```diff
@@ -1093,6 +1095,24 @@ fn print_str<const MINIFY: bool>(s: &str, p: &mut Codegen<{ MINIFY }>) {
     p.print(b'\'');
 }
 
+fn print_origin_str<const MINIFY: bool>(s: &str, p: &mut Codegen<{ MINIFY }>) {
```
underfin, line 1098: @Boshen I notice the `print_str` is changed, add the function maybe has messed, Maybe we can escape str and also it isn't valid?

Boshen: Instead of print_str, maybe we can use the original string slice:

directive.span.source_text(self.source_text())

(We'll need to pass in the original source text to codegen).

tasks/transform_conformance/babel.snap.md
```diff
@@ -608,6 +607,10 @@ Passed: 327/1369
 * transparent-expr-wrappers/ts-as-member-expression/input.ts
 * transparent-expr-wrappers/ts-parenthesized-expression-member-call/input.ts
 
+# babel-plugin-transform-json-strings (2/4)
```
underfin, line 610: The test case crashed because we need to escape `\u2028\u2029` at `plugin-transform-json-strings`. I can implement it at other pr.

Dunqing: I think we can merge this, but we will need to create an issue to track it

crates/oxc_codegen/src/gen.rs
```diff
@@ -92,8 +92,13 @@ impl<const MINIFY: bool> Gen<MINIFY> for Hashbang {
...
 impl<const MINIFY: bool> Gen<MINIFY> for Directive {
     fn gen(&self, p: &mut Codegen<{ MINIFY }>, _ctx: Context) {
-        // Use the string value instead of the raw self.directive because it can cannot escaped values.
-        print_str(self.expression.value.as_str(), p);
+        // A Use Strict Directive may not contain an EscapeSequence or LineContinuation.
+        // So here should print original `directive` value, the `expression` value is escaped str.
+        // See https://github.com/babel/babel/blob/main/packages/babel-generator/src/generators/base.ts#L64
+        let quote = choose_quote(self.directive.as_str());
+        p.print(quote as u8);
+        p.print_str(self.directive.as_str().to_string().as_bytes());
```
Boshen, line 100: `to_string` allocates, can we remove it somehow?

underfin: Updated.

tasks/transform_conformance/babel.snap.md
underfin: I already implemen it, i will open a new pr.

PR #2158, opened 2024-01-24
feat(ast): TypeScript definition for wasm target

### LINKED ISSUE #2151: AST Type generation for wasm targets
Ref: #301 for motivation.

More or less this boils down to deriving Tsify::tsify on _**all**_ the AST nodes in oxc_ast, some in oxc_syntax, oxc_span, oxc_semantic (bitflags, which semantic seems to have more of than the others, seem to be unreachable by the macro). Minimal changes to oxc_wasm itself.

I'm thinking the `cfg_attr(target_arch="wasm32", ...)` gate is probably evenly matched with a feature flag (`wasm`, naturally, as seen in zstd-rs, reqwest, chrono, etc), but I'm happy to go with the maintainers' preference.

Given the number of nodes (several hundred), the close association with serde and the low variation in the `cfg_attr(feature = "serde", ...)` macros (e.g. 80 identical ones in oxc_ast/src/js.rs), a macro_rules shorthand would be worthwhile, along the lines of:
```rust
#[ast_serde_macro(tag = "type")]
// expanding to:
#[cfg_attr(feature = "serde", derive(Serialize), serde(tag = "type"))]
#[cfg_attr(all(feature = "serde", feature = "wasm"), derive(Tsify)]
```
(as far as I can tell, it'll never be the intention to conditionally derive `serde::Deserialize`, so the only non-static part is the serde macro).

Let me know what you think is a reasonable compromise, I should have a PR up in a day or two.

As for the AST walker, I gather that'd be based on oxc_semantic's AstNodes impl, and wouldn't need to deal with closures crossing the JS-RS boundary (this is a particularly hairy part of wasm-bindgen :disappointed_relieved: ). Provided that's the case, these types should be sufficient.

H-Plus-Time:
Closes #2151.

See https://github.com/H-Plus-Time/oxc/pull/1 for the equivalent with a helper macro (just in the oxc_ast crate) and near-full coverage of the oxc_ast crate's structs.

Remaining semi-problematic types:

- [x]  IndexVec
- [x]  SymbolId
- [x]  ScopeId
- [x]  AstNodeId
- [x]  ReferenceId
- [x]  TSTypeOperatorType - not quite aliasing properly (serde rename changes it to TSTypeOperator, but dependent types still refer to the old name)
- [x]  TSAbstractMethodDefinition - incompatible extends of MethodDefinition (type tag)
- [x]  TSAbstractPropertyDefinition - incompatible extends of PropertyDefinition (type tag)
- [x]  Internal - generated by bitflags! macro
- [x]  CompactString - Atom is a NewType of it.


Boshen:
This is amazing. We should be able to ship a wasm version soon!

It would be awesome if you can show me the generated ts types.

Boshen:
I'll think about the tidied up version from https://github.com/H-Plus-Time/oxc/pull/1

H-Plus-Time:
> This is amazing. We should be able to ship a wasm version soon!
> 
> It would be awesome if you can show me the generated ts types.

In lieu of a gigantic comment, here's the gist: https://gist.github.com/H-Plus-Time/bdc8e102bfcf0ccd5c2c13322cc79937

NB: the getters for main Oxc class in oxc_wasm prevent the types from being overriden via tsify, there's a separate set of changes required to get that to line up with the generated types (basically it ends up with an exported class with the naive wasm-bindgen types (any and string mostly), and an interface with the specified types).

Boshen:
The `oxc_wasm` crate is currently being used as a proxy crate for the website, I'll probably make a better wasm target for general availability later. 

---

This is turning out great! Please let me know once this PR is ready to merge!

The nexts stage would be to package everything for a npm publish.

H-Plus-Time:
Alright, a few questions/notes on this one:

* Members of `TSTypeOperator` and `TSMappedTypeModifierOperator` all map to null (because they're C-style enums with implicit integer values) when serialized in the main branch. I've corrected this to serialize them as their string equivalents (e.g. Keyof serializes as `"keyof"`). I figure this was what you actually wanted, but just in case - let me know if the intent was to serialize as integers.
* SymbolFlags - this is the one and only type marked as unknown due to it's size - ideally, we'd figure something out for the bitflags! macro, use it for this and strip out the other (tolerably small) manual typedefs.

As for the export/publish path, at the moment, those types are generated via:
```bash
pushd crates/oxc_wasm
wasm-pack build --target=web
# output types file at pkg/oxc_wasm.d.ts, alongside the usual wasm and bindgen js
```

They should also show up in any wasm-bindgen'd project that bypasses oxc_wasm, so even something very basic like this:
```rust
use oxc::ast::ast::Statement;

pub fn foobar(input: Statement) {
    println!("{:?}", input);
}
```
(note the total lack of meaningful code), run through:
```bash
wasm-pack build --target=web
```
reproduces all annotated types touched by the pub exports of the module.

Boshen:
> Members of TSTypeOperator and TSMappedTypeModifierOperator all map to null

It was a bug

> SymbolFlags

Should be fine for now

--- 

Let me quickly test this locally.

H-Plus-Time:
Yeah, tests-wise, the only thing I can think of that makes sense to automate is:
```bash
npx -p typescript tsc --lib es2020,dom crates/oxc_wasm/pkg/oxc_wasm.d.ts
```
(would at least pick up on references to missing types (more or less that was my feedback loop when putting in the annotations)).

Happy to add this to the Check / Wasm GH flow if you like.

Cargo.toml
```diff
@@ -133,6 +133,10 @@ codspeed-criterion-compat = { version = "2.3.3", default-features = false }
 glob                      = { version = "0.3.1" }
 mime_guess                = { version = "2.0.4" }
 language-tags             = { version = "0.3.2" }
+# tsify                     = { version = "0.4.5" }
+tsify                     = { git = "https://github.com/siefkenj/tsify.git", rev = "36f42be" }
```
Boshen, line 137: What's missing from tsify? Do we need to publish another crate fork this? We don't want a git dependency for many reasons ...

Boshen:
Local test is working, I'll take another look tomorrow and prepare for a merge and release.

Cargo.toml
H-Plus-Time: Yep, fortunately that fork only addresses ident != name (as in, the name of the serde entity does not match the ident in rust), which is used in exactly two places - TSMappedTypeModifierOperator and TSTypeOperator.

I'll remove the original serde renames, that avoids the mismatching references (and the renames weren't that drastic anyway, I can't imagine all that many crate consumers breaking because of this), and the need for the fork.

PR #2162, opened 2024-01-24
feat(linter): eslint: no-void

eryue0220:
Rule Detail: [link](https://github.com/eslint/eslint/blob/main/lib/rules/no-void.js)

### DIFF
--- a/crates/oxc_linter/src/snapshots/no_void.snap
+++ b/crates/oxc_linter/src/snapshots/no_void.snap
@@ -6,42 +6,42 @@ expression: no_void
   ⚠ eslint(no-void): Disallow `void` operators
    ╭─[no_void.tsx:1:1]
  1 │ void 0
-   · ──────
+   · ────
    ╰────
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/no_void.snap
```diff
@@ -0,0 +1,48 @@
+---
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 150
+expression: no_void
+---
+  ⚠ eslint(no-void): Disallow `void` operators
+   ╭─[no_void.tsx:1:1]
+ 1 │ void 0
```
Dunqing, line 8: Shall we point the span to `void`?

Dunqing (changes requested):
Thank you!

crates/oxc_linter/src/snapshots/no_void.snap
eryue0220: Done~

Dunqing (approved):

PR #2163, opened 2024-01-24
feat(linter): eslint-plugin-jest: prefer-called-with

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-called-with.ts)

### DIFF
--- a/crates/oxc_linter/src/rules/jest/prefer_called_with.rs
+++ b/crates/oxc_linter/src/rules/jest/prefer_called_with.rs
@@ -1,10 +1,7 @@
 use crate::{
     context::LintContext,
     rule::Rule,
-    utils::{
-        collect_possible_jest_call_node, is_type_of_jest_fn_call, parse_expect_jest_fn_call,
-        JestFnKind, PossibleJestNode,
-    },
+    utils::{collect_possible_jest_call_node, parse_expect_jest_fn_call, PossibleJestNode},
 };
 
 use oxc_ast::AstKind;
@@ -60,10 +57,6 @@ impl PreferCalledWith {
         let AstKind::CallExpression(call_expr) = node.kind() else {
             return;
         };
-
-        if !is_type_of_jest_fn_call(call_expr, possible_jest_node, ctx, &[JestFnKind::Expect]) {
-            return;
-        }
 
         let Some(jest_fn_call) = parse_expect_jest_fn_call(call_expr, possible_jest_node, ctx)
         else {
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/jest/prefer_called_with.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl PreferCalledWith {
+    pub fn run<'a>(possible_jest_node: &PossibleJestNode<'a, '_>, ctx: &LintContext<'a>) {
+        let node = possible_jest_node.node;
+        let AstKind::CallExpression(call_expr) = node.kind() else {
+            return;
+        };
+
+        if !is_type_of_jest_fn_call(call_expr, possible_jest_node, ctx, &[JestFnKind::Expect]) {
```
mysteryven, line 64: Feel like we can remove this check, `parse_expect_jest_fn_call` will make sure it.

crates/oxc_linter/src/snapshots/prefer_called_with.snap
```diff
@@ -0,0 +1,27 @@
...
+source: crates/oxc_linter/src/tester.rs
+assertion_line: 150
+expression: prefer_called_with
+---
+  ⚠ eslint-plugin-jest(prefer-called-with): Suggest using `toBeCalledWith()` or `toHaveBeenCalledWith()`.
+   ╭─[prefer_called_with.tsx:1:1]
+ 1 │ expect(fn).toBeCalled();
+   ·            ──────────
+   ╰────
+  help: Prefer "toBeCalled"With(/* expected args */)
```
mysteryven, line 11: Is there missing a space between `"toBeCalled"` and `with`?

mysteryven (changes requested):

crates/oxc_linter/src/rules/jest/prefer_called_with.rs
eryue0220: Done~

crates/oxc_linter/src/snapshots/prefer_called_with.snap
eryue0220: Update. The problem is double quote shouldn't be here. It should be displayed with `toBeCallWith(/* expected args*/)` or `toHaveBeenCallWith(/* expected args*/)`.

mysteryven (approved):
Thank you!!

Issue #2013, opened 2024-01-13
Incorrect AstKind for ImportBinding Symbols


Dunqing:
https://oxc-project.github.io/oxc/playground/?code=3YCAAIDZgICAgICAgICFm8uTSqrl7ZqfoPUKV4Xs7evL4yxJGZbkMLzyZ8mg56yBMWVAxZNTVbs7SJ1mSSMtRRVbGt%2FF%2FS1i6V%2BIAa9qMVtTM%2BcpdiwXf39D0DCA

I am currently working on a case where I need to locate the AST node corresponding to `baz` in `export { baz }`. If `baz` is an `ImportSpecifier`, `ImportDefaultSpecifier`, or `ImportNamespaceSpecifier`, I need to check if its `import_kind` is type. If it is, I need to remove it.

However, I am encountering an issue. When I retrieve the kind from `self.ctx.nodes().kind(self.ctx.symbols().get_declaration(symbol_id))`, I am getting `AstKind::ModuleDeclaration` when I am expecting `AstKind::ImportSpecifier`.

To resolve this issue, it appears we need to modify the Binder implementation.

Dunqing:
I found another way to resolve this case in #2015. But this issue still needs to fix

PR #2164, opened 2024-01-24
feat(linter) no-import-assign: additional conditions

DeividAlmeida:


DeividAlmeida:
any suggestion to improve the performance?

Dunqing:
I'm not sure if we can change the reference flag to include read and write.

Boshen:
> I'm not sure if we can change the reference flag to include read and write.

I think we should improve all these references, the current method for getting the reference is not ideal, an O(1) solution is preferred.

But the current method is passing tests, so I propose we merge and create another issue for it.

Dunqing:
> I think we should improve all these references, the current method for getting the reference is not ideal, an O(1) solution is preferred.
> 

Yes, I'll try to figure out how to improve this

> But the current method is passing tests, so I propose we merge and create another issue for it.

It's too slow. I don't think we can merge

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
```diff
@@ -47,14 +49,47 @@ impl Rule for NoImportAssign {
...
+                | AstKind::AssignmentTarget(_)
+                | AstKind::ForOfStatement(_)
+                | AstKind::ForInStatement(_)
+                | AstKind::MemberExpression(_),
+            ) => {
+                return false;
+            }
+            (_, AstKind::MemberExpression(expr)) | (AstKind::MemberExpression(expr), _) => {
+                return expr.static_property_name() != Some("prop");
+            }
```
Dunqing, line 84: I don't understand why we need to check `static_property_name`. Can you explain it?

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
```diff
@@ -154,19 +189,18 @@ fn test() {
...
+        ("import * as mod3 from 'mod'; mod3.named++", None),
+        ("import * as mod4 from 'mod'; for (mod4.named in foo);", None),
+        ("import * as mod5 from 'mod'; for (mod5.named of foo);", None),
+        ("import * as mod6 from 'mod'; [mod6.named] = foo", None),
+        ("import * as mod7 from 'mod'; [mod7.named = 0] = foo", None),
+        ("import * as mod8 from 'mod'; [...mod8.named] = foo", None),
+        ("import * as mod9 from 'mod'; ({ bar: mod9.named } = foo)", None),
+        ("import * as mod10 from 'mod'; ({ bar: mod10.named = 0 } = foo)", None),
+        ("import * as mod11 from 'mod'; ({ ...mod11.named } = foo)", None),
+        ("import * as mod12 from 'mod'; delete mod12.named", None),
```
Dunqing, line 203: The import of these cases all belong to the `ImportNamespaceSpecifier`, so could we try to limit the check?

DeividAlmeida:
Thanks for your attention, at the weekend I'll work on it!!

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
DeividAlmeida: i did this check to pass these conditions 

image

and fails over these

image

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
DeividAlmeida: AstKind don't implement `ImportNamespaceSpecifier` could you help me how can i use it?

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
Dunqing: > i did this check to pass these conditions
> 
> image
> 

No, You shouldn't check this with `prop` name. We should forbid all property changes. Not just `prop`

crates/oxc_linter/src/rules/eslint/no_import_assign.rs
Dunqing: Yes. We need to finish #2013

DeividAlmeida: ok, i'll follow this PR, i think with these changes i can solve the property check problem, alright?

Dunqing: Yes

DeividAlmeida: image

i can't solve this because these case is `ImportNamespaceSpecifier` also and any approach over in this case will decrease significatively the linter performance.

PR #2165, opened 2024-01-24
fix(linter): explicit-length-check inside ternary

### LINKED ISSUE #2156: bug(linter): explicit-length-check inside ternary
I've got a false positive with `explicit-length-check` when size is inside a ternary.

```ts
  const totalCount = tests.reduce(
    (count, test) => count + (test.enabled ? test.maxSize : test.size),
    0,
  )
```

```
  ⚠ eslint-plugin-unicorn(explicit-length-check): Use `.size > 0` when checking size is not zero.
     ╭─[src/...]
 110 │   const totalCount = tests.reduce(
 111 │     (count, test) => count + (test.enabled ? test.maxSize : test.size),
     ·                                                               
 112 │     0,
     ╰────
```

maurice:
Fixes #2156 by checking that the length/size check is in the "test" part of a conditional (aka ternary).

I stole the technique from here https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/no_extra_boolean_cast.rs#L172-L175 - there might be potential for re-use as this one looks a bit more sophisticated? I didn't bother trying though, sorry!

I see there is already [an example in `pass` for conditional with the length/size check in the test part](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/explicit_length_check.rs#L331) but not in `fail` so I added one.

Hope it's all OK but shout if anything needs changing 😄

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/explicit_length_check.rs
+++ b/crates/oxc_linter/src/rules/unicorn/explicit_length_check.rs
@@ -348,6 +348,8 @@ fn test() {
         // need getStaticValue
         // ("const A_NUMBER = 2; const x = foo.length || A_NUMBER", None),
         ("class A { a(){ if(this.length); while(!this.size || foo);}}", None),
+        // Use of .size but not in conditional "test" position
+        ("const totalCount = tests.reduce((count, test) => count + (test.enabled ? test.maxSize : test.size), 0)", None),
     ];
 
     let fail = vec![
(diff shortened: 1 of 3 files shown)


camc314 (approved):

maurice:
Apparently my WSL clock is off by 7 days (according to the commit timestamp) I'll fix it for the next contribution if you don't mind?

PR #2167, opened 2024-01-25
feat(tasks): Add more plugins for tasks/lint_rules2

leaysgur:
Part pf #2020 

- [x] jsx-a11y
- [x] react, react-hooks, react-perf
- [x] nextjs

Now, we can track all target plugins(AFAIK)? 🕵🏻


Boshen:
> Now, we can track all target plugins(AFAIK)? 🕵🏻

Amazing!

PR #2168, opened 2024-01-25
feat(transfrom): transform-json-strings

underfin:
The pr intends to implement the plugin `babel-plugin-transform-json-strings`. But here is only mutate `Directive`, the `StringLiteral` is not implement. It need to changed the `StringLiteral` printer.

I'm intend to add the raw of `StringLiteral`, it will be mutate at plugin, and using the `raw` to print `StringLiteral`. If you other ideas, please let me know.

### DIFF
--- a/crates/oxc_transformer/src/es2019/json_strings.rs
+++ b/crates/oxc_transformer/src/es2019/json_strings.rs
@@ -37,8 +37,8 @@ impl JsonStrings {
     }
 
     #[allow(clippy::unused_self)]
-    // TODO oxc_codegen currently prints json strings correctly, 
-    // but we need a way to turn off this behaviour from codegen 
+    // TODO oxc_codegen currently prints json strings correctly,
+    // but we need a way to turn off this behaviour from codegen
     // and do the transformation here.
     pub fn transform_string_literal(&mut self, _literal: &mut StringLiteral) {
         // let str = &self.ctx.semantic().source_text()[literal.span.start as usize + 1..literal.span.end as usize - 1];
--- a/crates/oxc_transformer/src/es2019/json_strings.rs
+++ b/crates/oxc_transformer/src/es2019/json_strings.rs
@@ -1,0 +1,54 @@
+use oxc_ast::ast::*;
+use oxc_span::Atom;
+use oxc_syntax::identifier::{LS, PS};
+
+use crate::options::{TransformOptions, TransformTarget};
+
+/// ES2019: Json Strings
+///
+/// References:
+/// * <https://babeljs.io/docs/babel-plugin-transform-json-strings>
+/// * <https://github.com/babel/babel/tree/main/packages/babel-plugin-transform-json-strings>
+pub struct JsonStrings;
+
+impl JsonStrings {
+    pub fn new(options: &TransformOptions) -> Option<Self> {
+        (options.target < TransformTarget::ES2019 || options.json_strings).then(|| Self {})
+    }
+
+    // Allow `U+2028` and `U+2029` in string literals
+    // <https://tc39.es/proposal-json-superset>
+    // <https://github.com/tc39/proposal-json-superset>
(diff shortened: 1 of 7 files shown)


crates/oxc_transformer/src/es2019/json_strings.rs
```diff
@@ -0,0 +1,54 @@
...
+                (false, PS) => buf.push_str("\\u2029"),
+                _ => buf.push(c),
+            }
+            is_escaped = !is_escaped && matches!(c, '\\');
+        }
+        Some(buf.into())
+    }
+
+    #[allow(clippy::unused_self)]
+    // TODO here should mutate origin source string and print using origin source.
```
Boshen, line 40: I don't get this TODO, where is the failing test case?

underfin: The plugin feature is not completely implement, the print is always enable the `json-strings` transform, but actually it only enabled by added plugin. You can see https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=OQVwzgpgBGAuBOgKAgDoCgCWBjWwDcqAicZGBLWA_QgIwgDMB7eCRAQztgnktQHpeogwQD0oAFQAWrHGCisoAVQDUAJgAMKgBxQAMgEkAcgFEoAZSMAFAIIAlK2IDyNxWIBiAWgCMANiiYp8FAAFLSwAO4QEAB2UMC0jMzAclEAJrHsnPDAAJSEiJRQ_FAAUuCwcoqqGtr6xmaWtvZOhMj5uIUCRmCYrAAOEGnK6lq6hibm1naONi0gw5oFRV09_YNVI7XjDVNOQA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.23.8&externalPlugins=&assumptions=%7B%7D and https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=OQVwzgpgBGAuBOgKAgDoCgCWBjWwDcqAicZGBLWA_QgIwgDMB7eCRAQztgnktQHpeogwQD0oAFQAWrHGCisoAVQDUAJgAMKgBxQAMgEkAcgFEoAZSMAFAIIAlK2IDyNxWIBiAWgCMANiiYp8FAAFLSwAO4QEAB2UMC0jMzAclEAJrHsnPDAAJSEiJRQ_FAAUuCwcoqqGtr6xmaWtvZOhMj5uIUCRmCYrAAOEGnK6lq6hibm1naONi0gw5oFRV09_YNVI7XjDVNOQA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.23.8&externalPlugins=%40babel%2Fplugin-syntax-json-strings%407.8.3&assumptions=%7B%7D.

Boshen, line 40:
```suggestion
    // TODO oxc_codegen currently prints json strings correctly, 
    // but we need a way to turn off this behaviour from codegen 
    // and do the transformation here.
```

Boshen (approved):

crates/oxc_transformer/src/es2019/json_strings.rs
Boshen: I updated the comment. A todo should provide enough context for the next person to make actionable choices.

PR #2169, opened 2024-01-25
fix(linter): use correct rule name

leaysgur:
Found by accident but it should be.

- [x] react/jsx-no-comment-text-nodes
  - `jsx-no-comment-textnodes` is correct
  - https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/jsx-no-comment-textnodes.js
- [x] react/no-dangerously-set-inner-html
  - `no-danger` is correct
  - https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-danger.js
- [x] react-perf/no-jsx-as-prop
  - Correct name has `jsx-` prefix
  - https://github.com/cvazac/eslint-plugin-react-perf/tree/master/lib/rules
- [x] react-perf/no-new-array-as-prop
  - The same as above
- [x] react-perf/no-new-function-as-props
  - The same as above
- [x] react-perf/no-new-object-as-prop
  - The same as above
- [x] jsx-a11y/tab-index-no-positive
  - `tabindex-no-positive` is correct
  - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/tabindex-no-positive.js
- [x] jsx-a11y/role-support-aria-props
  - `role-suports-aria-props` is correct
  - https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/role-supports-aria-props.js

- - - 

+1, 

- typescript/consistent-type-exports
  - Listed in `disable-type-checked` config == This is type checking rule but implemented
  - https://github.com/typescript-eslint/typescript-eslint/blob/8622286207b3e04ddd260eaa8ac71927d47edd7d/packages/eslint-plugin/src/configs/disable-type-checked.ts#L12
  - Should remove it..?


Boshen:
Nice find!

> Should remove it..?

Yes, I'll remove it.

PR #2172, opened 2024-01-25
feat(vscode): allow config path configuration

### LINKED ISSUE #1944: support configuring the config file path from vscode settings.

Djiit:
Fixes #1944

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -47,11 +47,12 @@ enum Run {
 struct Options {
     run: Run,
     enable: bool,
+    config_path: String,
 }
 
 impl Default for Options {
     fn default() -> Self {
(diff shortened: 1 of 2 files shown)


shevernitskiy:
Could it handle all of the possible config names?
```
.eslintrc
.eslintrc.js
.eslintrc.cjs
.eslintrc.yaml
.eslintrc.yml
.eslintrc.json
package.json
```

Djiit:
Yes, but not in this PR, because in the end oxlint still expect a JSON format. Happy to revisit this when the core linter is able to parse more versions of it (including maybe the new config version).

Djiit:
@Boshen first try here. I could even keep the old code as a fallback, and remove the configuartion default. e.g. "if vscode configuration is empty, try `.eslintrc` and `.eslintrc.json`" wdyt?

PR #2185, opened 2024-01-26
inline constants

### LINKED ISSUE #2129: Minifier should constant fold bindings
Given

```mjs
var OK = 200;
// ...
if (res.status === OK) {
//  ...
}
```

The minifier should replace read access to `OK` with the constant value of `200` if the total number of replacements would save space (in this case 1 additional char per `200` vs `OK` vs savings of removing `var OK = 200;`).

```mjs
if (res.status === 200) {
//  ...
}
```

Doing so should work even when the constant occurs after closures using it (with potential need for TDZ checks):

```mjs
function onResult() {
  if (res.status === OK) // ...
}
let OK = 200;
```

The minifier should apply this to any binding with a known constant value including `var`, `let`, `const`, etc.

bmeck:
fixes: #2129

## TODO

- [ ] CFG checks for:
```
function foo() {
  x // needs tdz check in additon to inlining
}
x // should not inline
let x = 123;
```

### DIFF
--- a/crates/oxc_minifier/examples/minifier.rs
+++ b/crates/oxc_minifier/examples/minifier.rs
@@ -4,6 +4,7 @@ use oxc_allocator::Allocator;
 use oxc_codegen::{Codegen, CodegenOptions};
 use oxc_minifier::{Minifier, MinifierOptions};
 use oxc_parser::Parser;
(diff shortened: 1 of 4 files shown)


Boshen:
Closing due to inactivity.

PR #2187, opened 2024-01-27
chore(tasks/lint_rules): Arrange UI

leaysgur:
Part of #2020

- [x] rule counters
- [x] show two tables
- [x] omit all the deprecated rules

Eventually I decided to split them into 3 tables.

In order to be aware of the newly deprecated rules in the future, the deprecated rules are hidden by default, rather than omitted altogether.


Boshen:
Can't wait! Amazing work!

Boshen:
This is beautiful! Time to advertise!

leaysgur:
That's good to hear~! 😄

PR #2201, opened 2024-01-29
feat(allocator,ast): add `BoxWithDrop<BigInt>`

Boshen:
relates #1803

I wanna make things work first, get miri to pass, and then work on better workarounds. cc @overlookmotel 

Miri run: https://github.com/oxc-project/oxc/actions/runs/7693186484

### DIFF
--- a/crates/oxc_allocator/src/arena.rs
+++ b/crates/oxc_allocator/src/arena.rs
@@ -14,7 +14,7 @@ use serde::{ser::SerializeSeq, Serialize, Serializer};
 
 use crate::Allocator;
 
-/// Bumpalo Box
+/// A Box without `Drop`
 pub struct Box<'alloc, T: ?Sized>(pub &'alloc mut T);
 
 impl<'alloc, T> Box<'alloc, T> {
--- a/crates/oxc_allocator/Cargo.toml
+++ b/crates/oxc_allocator/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 doctest = false
 
 [dependencies]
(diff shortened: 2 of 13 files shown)


Boshen:
Ohhh... this won't work, the `drop` calls need to be chained all the way to the top :-) I can't just stick a `BoxWithDrop` in the middle of no where and expect it's drop to be called.

crates/oxc_allocator/src/arena.rs
```diff
@@ -14,7 +14,7 @@ use serde::{ser::SerializeSeq, Serialize, Serializer};
 
 use crate::Allocator;
 
-/// Bumpalo Box
+/// A Box without `Drop`
```
Boshen, line 17: Need to update to the following:

A Box without `Drop`. This is used for over coming self-reference structs.

It is a memory leak if the boxed value has a `Drop` implementation.

overlookmotel:
The best solution I think is to prevent `BigInt` from making allocations in the first place. Unfortunately that'd require reimplementing `BigInt` (or at least the parts of its API which OXC uses).

To do a swifter fix in the meantime, the only way I can see to fix the memory leak without making the entire AST drop is:

1. Store pointers to all `BigInt`s in a vec in `Allocator`.
2. Add `impl Drop` to `Allocator` which converts those pointers back to `BigInt`s and drops them.

That would solve the leak, though I'm not sure it'd be sound - the pointers would need to be `*mut`, so it might well be violating Rust's aliasing rules.

PR #2204, opened 2024-01-29
fix(parser): correct MAX_LEN for 32-bit systems

overlookmotel:
Maximum length of source parser can accept is limited on 32-bit systems to `isize::MAX` (i.e. `i32::MAX` not `u32::MAX`) because Rust [limits the size of allocations](https://doc.rust-lang.org/std/alloc/struct.Layout.html#method.from_size_align) to `isize::MAX`.

This PR takes that constraint into account when calculating `Parser::MAX_LEN`.

It also speeds up the `overlong_source` test so it runs in under 500ms (previously it took ~4 secs on a M1 Macbook Pro).

### DIFF
--- a/crates/oxc_parser/src/lib.rs
+++ b/crates/oxc_parser/src/lib.rs
@@ -90,9 +90,19 @@ pub mod __lexer {
     pub use super::lexer::{Kind, Lexer, Token};
 }
 
-/// Maximum length of source in bytes which can be parsed (~4 GiB).
-// Span's start and end are u32s, so size limit is u32::MAX bytes.
-pub const MAX_LEN: usize = u32::MAX as usize;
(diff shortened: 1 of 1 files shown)


Boshen (approved):

overlookmotel:
Graphite question: I tried to make this PR using Graphite as I have a follow-on that builds on this. But Graphite made the PR on my fork instead of upstream repo.

Can't find any solution in Graphite's docs. Do I need to be making my branches on upstream, rather than my fork, in order to use Graphite?

Boshen:
> Do I need to be making my branches on upstream, rather than my fork, in order to use Graphite?

Yes, stop using a fork lol, push to origin is the way.

overlookmotel:
OK cool. Thanks.

PR #2206, opened 2024-01-29
refactor(parser): move source length check into lexer

overlookmotel:
This change makes little difference in itself, but moving the check into the lexer will allow some optimizations in lexer using unsafe code which depend on this invariant.


Boshen:
Let me guess you are going to remove a lot of bound checks... I can't wait!

overlookmotel:
> Let me guess you are going to remove a lot of bound checks... I can't wait!

I have been investigating that, yes!

PR #2213, opened 2024-01-29
Better source line and column numbers

maurice:
Attempt to improve the graphical reporter source file **line number** and **column number** .

Currently `oxlint` renders the **line number** as the first line in the snippet **column number** is always `1` AFAICT. Example of current version on VS Code repo:

image

This means that when CTRL-clicking on the source-file:line:col "link" it takes me to the wrong place.

This PR is an attempt to improve things by getting closer to the source of the issue. Here is my version again on VS Code:

image

I'm a bit worried that none of the tests failed though - would that be expected?

Also the approach feels naïve and there may be several cases it won't work for, or won't work quite so well. Specifically it's just finding the first labelled-span, but I'm guessing in some cases that could be providing context, rather than the thing the user would want to change.

As before I'm no expert and any feedback is very much appreciated.

Also in my local testing, VS Code is still a bit unpredictable - even with the correct line + col it will always go to the correct line number but doesn't always go to the correct column 🤷 But that's not our issue 😄

### DIFF
--- a/crates/oxc_diagnostics/src/graphic_reporter.rs
+++ b/crates/oxc_diagnostics/src/graphic_reporter.rs
@@ -400,13 +400,23 @@ impl GraphicalReportHandler {
             self.theme.characters.hbar,
         )?;
 
+        let (source_line, source_col) = labels
+            .iter()
+            .find_map(|hl| {
+                lines
+                    .iter()
+                    .find(|line| line.span_applies(hl) && line.span_line_only(hl))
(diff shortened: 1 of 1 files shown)


Boshen:
> I'm a bit worried that none of the tests failed though - would that be expected?

This is the forked graphical reporter, the testers are using the once from miette. Let me clean this up a little bit.

maurice:
Ah that explains it, thanks @Boshen 🙏

PR #2214, opened 2024-01-29
feat(cli/linter): support .eslintrc.js

DonIsaac:
Adds support for JavaScript eslint configs. Configs are loaded by running `node` in a subprocess. This is significantly slower than simply parsing a JSON file, but increases compatibility with existing configs and should make migrating to oxc easier.

## Security
Since loading JS configs requires running `node`, this opens the door to RCE attacks. To mitigate threats, node's experimental [Permissions API](https://nodejs.org/docs/latest-v20.x/api/permissions.html) is used when available.

### DIFF
--- a/crates/oxc_linter/fixtures/eslint_config.js
+++ b/crates/oxc_linter/fixtures/eslint_config.js
@@ -1,0 +1,21 @@
+module.exports = {
+    rules: {
+        // requires jsonc support
+        'no-console': 'off',
+        'no-bitwise': [
+            'error',
+            {
+                allow: ['~'],
+            },
+        ],
+        eqeqeq: [
+            'error',
+            'always',
+            {
+                null: 'ignore',
+            },
+        ],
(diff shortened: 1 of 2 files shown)


Boshen:
I'll need to test this locally.

Boshen:
I'm going to reject this feature request for a couple of reasons after sitting on this for a while:

* This is a maintenance burden for me. It is a feature creep where we are making ourselves available to support more node specific features. To be dramatic, people may ask us to support this: https://github.com/elastic/kibana/blob/main/.eslintrc.js
* If you look at the eslint help questions on stackoverflow, its github issues, its discord - you'll find 90% of the issues are configuration issues
* I'm currently in the camp of static configs https://twitter.com/buildsghost/status/1742290248628146224
* oxlint don't require node.js nor node_modules, and I want to keep it this way for a long time.

Thank you for the PR. You made me aware that this is possible, and also inspired me to experiment with this https://github.com/oxc-project/oxc/issues/2218

PR #2215, opened 2024-01-30
perf(parser): faster offset calculation

overlookmotel:
A faster way to calculate offset in the lexer.

This only moves the needle because it's on the hottest path in the lexer - `Lexer::offset` is called for every token in `Lexer::read_next_token`.


Boshen:
image

Boshen:
@overlookmotel This is hard to understand without context, can you add some inline comments regarding why this is better than the previous version? 

Something like:

// This is changed from `self.source.len() - self.current.chars.as_str().len()`,
// where ... (bunch of assembly instructions)

overlookmotel:
Yes sure. Have added a new commit with a comment.

PR #2229, opened 2024-01-31
refactor(parser): byte handler for illegal bytes

overlookmotel:
This adds a separate byte handler to the lexer for byte values which should never be encountered:

1. UTF-8 continuation bytes (i.e. middle of a multi-byte UTF-8 byte sequence).
2. Bytes values which are illegal in valid UTF-8 strings.

At present, this function is impossible to reach, because `std::str::Chars` ensures the next byte is always the *start* of a valid UTF-8 byte sequence. But later changes I intend introducing unsafe code will make it possible (but highly undesirable!). In the meantime, I don't think it does any harm to handle this case.


overlookmotel:
Built on top of #2228. It'd be better if I was using Graphite, but I have a lot of WIP branches on my fork, so would like to get them wrapped up before migrating to Graphite.

overlookmotel:
Now rebased on main and ready for merge if you're happy with the change.

overlookmotel:
Who needs Graphite when you're this fast to review and merge? :)

PR #2238, opened 2024-01-31
feat(transformer/decorators): support method decorator and is not static

Dunqing:


Dunqing:
* `main`
  * **PR #2238** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2238?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2238?utm_source=stack-comment).

crates/oxc_ast/src/ast_builder.rs
```diff
@@ -879,6 +879,23 @@ impl<'a> AstBuilder<'a> {
         }))
     }
 
+    pub fn class_constructor(&self, span: Span, value: Box<'a, Function<'a>>) -> ClassElement<'a> {
+        ClassElement::MethodDefinition(self.alloc(MethodDefinition {
+            span,
+            key: self.property_key_expression(self.identifier_reference_expression(
+                IdentifierReference::new(SPAN, "constructor".into()),
```
Dunqing, line 886: I'm not sure we can use `SPAN` here. If we can use `SPAN` we can add more useful helper functions, such as `require_call`

Boshen: I'm still unsure about empty spans and their implications, but we'll use it until problems come up.

PR #2245, opened 2024-02-01
fix(lexer): correct the span for irregular whitespaces

### LINKED ISSUE #2236: Irregular whitespace in Lexer
Is it intentional that `self.consume_char()` is *after* `self.trivia_builder.add_irregular_whitespace()` here:

https://github.com/oxc-project/oxc/blob/a79988d5e2cd0cf616b56ae1236e5c29d3d950bd/crates/oxc_parser/src/lexer/unicode.rs#L30-L35

As it is at present, `self.current.token.start` and `self.offset()` are the same, so the "trivia"'s span is just a pointer to the start of the irregular whitespace. If `self.consume_char()` was moved to before, span would cover the characters in question (start to end).

I'm not clear what the purpose of recording irregular whitespace is, so not sure if this is intentional or not.

Boshen:
closes #2236

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_irregular_whitespace.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_irregular_whitespace.rs
@@ -42,6 +42,7 @@ impl Rule for NoIrregularWhitespace {
     }
 }
 
+#[allow(clippy::unicode_not_nfc, clippy::invisible_characters)]
 #[test]
 fn test() {
     use crate::tester::Tester;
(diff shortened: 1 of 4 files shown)


Boshen:
* `main`
  * **PR #2245** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2245?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2245?utm_source=stack-comment).

PR #2250, opened 2024-02-01
fix(oxc_semantic): proper traversal of try statements

### LINKED ISSUE #2227: [oxc_semantics] improper CFG for try-finally
According to https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-try-statement-runtime-semantics-evaluation, the case of a try-finally is quite strange and we are not handling it correctly. Essentially, there are 2 parallel universes created:

## 1 - The try completes successfully:

Here, this is the flow:

1. input node
2. the flow of the try block as a single node (no early catch bail-outs).
3. the flow of the finally
4. output node

## 2 - the try completes with an error:

Here, the flow is very different:

1. Input node
2. the flow of try block with a node for errors before and after each statement.
3. all the error nodes point to the start of the finally block.
4. The finally block runs.
5. The equivalent of a throw happens to wherever it would have gone in the input node.

## The problem

If we try to handle these cases with a single finally node or set thereof, we end up with seemingly valid paths to the output node where the entire try block wasn't executed. Consider the following:

```js
class A extends B {
    constructor() {
        try {
            super();
        } finally {}
        this.a();
    }
}
```

The current CFG implementation allows for the following path:

```js
class A extends B {
    constructor() {
        // try {
            // super(); throws an error
        // } finally {}
        this.a();
    }
}
```

Which would flag a diagnostic for the `no-this-before-super` eslint rule.

To fix this, we need to somehow fork the CFG construction in the `oxc_semantic` `builder.rs` or run through 2 separate flows. This could get a lot more complicated with multiple levels of try / catch / finally, so I think it's best to run the `visit_finally` twice in this case, potentially creating other try contexts therein.

I will try (no pun intended) to put up a PR for this in the next day or so, so will assign myself for now.

TzviPM:
Closes #2227


TzviPM:
This changes the snap files in the linter, because the proper try handling now catches some additional errors, but the PR is not really a linter PR.

Boshen (approved):
This is nice! It never occurred to me that the finally block is missed.

PR #2254, opened 2024-02-01
feat(linter): Implement no_this_before_super with cfg

### LINKED ISSUE #2279: eslint/no-this-before-super
Part of #479

https://eslint.org/docs/latest/rules/no-this-before-super

TzviPM:
Implements `eslint/no-this-before-super` in #479.

Closes #2279

### DIFF
--- a/crates/oxc_linter/src/snapshots/no_this_before_super.snap
+++ b/crates/oxc_linter/src/snapshots/no_this_before_super.snap
@@ -1,0 +1,200 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: no_this_before_super
+---
+
+  ⚠ eslint(no-this-before-super): Expected to always call super() before this/super property access.
+   ╭─[no_this_before_super.tsx:1:21]
+ 1 │ class A extends B { constructor() { this.c = 0; } }
+   ·                     ─────────────────────────────
+   ╰────
+  help: Call super() before this/super property access.
+
+  ⚠ eslint(no-this-before-super): Expected to always call super() before this/super property access.
+   ╭─[no_this_before_super.tsx:1:21]
+ 1 │ class A extends B { constructor() { this.c(); } }
+   ·                     ───────────────────────────
+   ╰────
+  help: Call super() before this/super property access.
+
+  ⚠ eslint(no-this-before-super): Expected to always call super() before this/super property access.
+   ╭─[no_this_before_super.tsx:1:21]
+ 1 │ class A extends B { constructor() { super.c(); } }
+   ·                     ────────────────────────────
+   ╰────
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/no_this_before_super.snap
```diff
@@ -0,0 +1,124 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: no_this_before_super
+---
+
+  ⚠ eslint(no-this-before-super): Expected to always call super() before this/super property access.
+   ╭─[no_this_before_super.tsx:1:32]
+ 1 │ class A extends B { constructor() { this.c = 0; } }
+   ·                                ──────────────────
```
Boshen, line 9: This start span is slightly off, let me check the AST node.

Boshen (commented):
I checked the error span biome and eslint emits, and combining with the general advice on error spans https://oxc-project.github.io/docs/contribute/linter.html#general-advice

> ## Pin point the error message to the shortest code span
> We want the user to focus on the problematic code rather than deciphering the error message to identify which part of the code is erroneous.

The best span would be 1 span pointing to the `constructor` keyword, and the other error span pointing to the `this` keyword.

```
 1 │ class A extends B { constructor() { this.c = 0; } }
   ·                     ___________     ____
```
The second best is to point to the whole method

```
 1 │ class A extends B { constructor() { this.c = 0; } }
   ·                     _____________________________
```

TzviPM:
> The best span would be 1 span pointing to the constructor keyword, and the other error span pointing to the this keyword.

Do we have a mechanism for this? My concern is that, say there are 2 uses of `this` before `super()`, if we highlight the first usage only, then the user fixes that one and then we show the error again with the second usage, it could be confusing. If we are able to have a diagnostic with multiple spans, we can do that here. Or else should we have multiple diagnostics? I'm happy to start by making sure to use the span for the entire method definition node.

Boshen:
Here's an example for multiple spans:

https://github.com/oxc-project/oxc/blob/0c225a49aa36f5d9c9738cdf2cb6d6c7d1810443/crates/oxc_linter/src/rules/eslint/no_sparse_arrays.rs#L58-L63

Feel free to to pick either one depending on your time availableness.

TzviPM:
I'm going to submit with the option highlighting the whole method for now and open an issue to make it better. I'd like to explore a bit more what the right approach would be; I have a few ideas.

TzviPM:
> I'm going to submit with the option highlighting the whole method for now and open an issue to make it better. I'd like to explore a bit more what the right approach would be; I have a few ideas.

Opened #2290

Boshen (approved):

PR #2255, opened 2024-02-01
fix(parser): Binding is not allowed to be optional in VariableDeclarator

### LINKED ISSUE #2253: BindingPattern shouldn't allow optional
Please check the playground, These cases should report a syntax error

https://oxc-project.github.io/oxc/playground/?code=3YCAAID6gICAgICAgICXYPzir2mHELJNP5ai%2BRRJpeC%2FNQpaxAyIIQDW3aGxwOFp8XLQlq5uBOSOyMfgKq6ev2ENr1%2FjDLFCrJTzzc96PycEillagwONy03dP7z%2BdcIPd9VMDu1dLhMAJ5Cu2XSP7d9%2FLb8AgA%3D%3D

In typescript, I get the syntax error

https://www.typescriptlang.org/play?target=99#code/MYewdgzgLgBAhgfhgXhgIgIJoFCkrAbQCMBdJVAtAITRN3GhgG8ZgYBfc51gLnQGE0HIA

Dunqing:
fix: #2253

I'm not sure what the best way to fix this bug is. I try to add a `Context::Require` to fix it. This is the first time I've changed the parser's code.

I can always close this PR if there is a better solution.

### DIFF
--- a/crates/oxc_parser/src/context.rs
+++ b/crates/oxc_parser/src/context.rs
@@ -44,6 +44,9 @@ bitflags! {
         ///   * ambient variable declaration => `declare var $: any`
         ///   * ambient class declaration => `declare class C { foo(); } , etc..`
         const Ambient = 1 << 5;
+
+        /// Binding is not allowed as optional
+        const Required = 1 << 6;
     }
 }
 
@@ -82,6 +85,11 @@ impl Context {
     #[inline]
     pub(crate) fn has_ambient(self) -> bool {
         self.contains(Self::Ambient)
+    }
+
+    #[inline]
+    pub(crate) fn has_required(self) -> bool {
+        self.contains(Self::Required)
     }
 
     #[inline]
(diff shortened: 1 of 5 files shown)


Dunqing:
* `main`
  * **PR #2255** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2255?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2255?utm_source=stack-comment).

Boshen:
It's complicated to check this in the parser, let's try with the checker instead.

Dunqing:
> It's complicated to check this in the parser, let's try with the checker instead.

In this case,  Should we report the error and break the parser? If we do this in `checker` the parse is done. I'm not sure this will have any other effect.

Boshen:
Let me give this a try

PR #2269, opened 2024-02-02
ci(conformance): run `just coverage` instead of a bunch of commands

Dunqing:


.github/workflows/conformance.yml
```diff
@@ -57,10 +57,7 @@ jobs:
           node-version: 18
           registry-url: 'https://registry.npmjs.org'
 
-      - run: cargo coverage
-      - run: cargo run --release -p oxc_transform_conformance -- --exec
-      - run: cargo run --release -p oxc_prettier_conformance
-      # - run: cargo minsize
+      - run: just coverage
```
Boshen, line 60: just is not installed, need that binstall action.

Dunqing:
Install `binstall` is too slow

PR #2275, opened 2024-02-02
chore(website): support deploy to netlify

Dunqing:


Boshen:
Do we need to change all the urls within our documentations to the netlify url? We don't have a domain though 😅

Dunqing:
> Do we need to change all the urls within our documentations to the netlify url? We don't have a domain though 😅

I think we need a domain, otherwise, we may have multiple netlify urls

Dunqing:
Or we can use `oxc.netlify.app` link to documentation, and add `oxc-playground.netlify.io` to link the playground.

PR #2283, opened 2024-02-03
chore: codegen test skip v8 failed test

underfin:
Because v8 isn't pass some test262 tests, we also need to skip them, you can find it at https://chromium.googlesource.com/v8/v8/+/refs/heads/main/test/test262/test262.status.


tasks/coverage/src/runtime/mod.rs
```diff
@@ -182,7 +201,7 @@ async fn request_run_code(json: impl serde::Serialize + Send + 'static) -> Resul
     tokio::spawn(async move {
         agent()
             .post("http://localhost:32055/run")
-            .timeout(Duration::from_secs(10))
+            .timeout(Duration::from_secs(50))
```
Boshen, line 204: This seems like an unrelated change? Why do we need to timeout for so long?

This goes against the philosophy of this project, we should strive to keep all the waiting times down.

tasks/coverage/src/lib.rs
```diff
@@ -80,6 +81,26 @@ impl AppArgs {
...
+    pub fn run_sync_v8_test262_status(&self) {
+        let res = agent()
+            .get("http://raw.githubusercontent.com/v8/v8/main/test/test262/test262.status")
+            .timeout(Duration::from_secs(10))
+            .call()
+            .expect("Get v8 test262 status failed")
+            .into_string()
+            .expect("Get v8 test262 status failed");
+        let mut content = String::new();
+        regex::Regex::new(r"'(.+)': \[FAIL\]").unwrap().captures_iter(&res).for_each(|caps| {
```
Boshen, line 96: There are also [SKIP] Markers

tasks/coverage/src/lib.rs
```diff
@@ -80,6 +81,26 @@ impl AppArgs {
...
+            content.push_str(caps.get(1).unwrap().as_str());
+            content.push('\n');
+        });
```
Boshen, line 99: Can we also sort the lines? So we don't get a big diff if the file ever changes.

tasks/coverage/src/runtime/mod.rs
underfin: Here has many timeout at my local environment, it is valid to avoid timeout.

Boshen: We need to fix the timeout if this is the case. cc @Dunqing if you happen to know why it's timing out.

underfin: Could you also help me update the test snapshot file? Thanks..

Boshen (approved):

Boshen:
Thank you!

tasks/coverage/src/runtime/mod.rs
Boshen: Try `just submodules`. You may need to delete their directories.

PR #2288, opened 2024-02-04
perf(parser): lexer replace `Chars` with `Source`

overlookmotel:
This PR replaces the `Chars` iterator in the lexer with a new structure `Source`.

## What it does

`Source` holds the source text, and allows:

* Iterating through source text char-by-char (same as `Chars` did).
* Iterating byte-by-byte.
* Getting a `SourcePosition` for current position, which can be used later to rewind to that position, without having to clone the entire `Source` struct.

`Source` has the same invariants as `Chars` - cursor must always be positioned on a UTF-8 character boundary (i.e. not in the middle of a multi-byte Unicode character).

However, unsafe APIs are provided to allow a caller to temporarily break that invariant, as long as they satisfy it again before they pass control back to safe code. This will be useful for processing batches of bytes.

## Why

I envisage most of the Lexer migrating to byte-by-byte iteration, and I believe it'll make a significant impact on performance.

It will allow efficiently processing batches of bytes (e.g. consuming identifiers or whitespace) without the overhead of calculating code points for every character. It should also make all the many `peek()`, `next_char()` and `next_eq()` calls faster.

`Source` is also more performant than `Chars` in itself. This wasn't my intent, but seems to be a pleasant side-effect of it being less opaque to the compiler than `Chars`, so it can apply more optimizations.

In addition, because checkpoints don't need to store the entire `Source` struct, but only a `SourcePosition` (8 bytes), was able to reduce the size of `LexerCheckpoint` and `ParserCheckpoint`, and make them both `Copy`.

## Notes on implementation

`Source` is heavily based on Rust's `std::str::Chars` and `std::slice::Iter` iterators and I've copied the code/concepts from them as much as possible.

As it's a low-level primitive, it uses raw pointers and contains a *lot* of unsafe code. I *think* I've crossed the T's and dotted the I's, and I've commented the code extensively, but I'd appreciate a close review if anyone has time.

I've split it into 2 commits.

* First commit is all the substantive changes.
* 2nd commit just does away with `lexer.current` which is no longer needed, and replaces `lexer.current.token` with `lexer.token` everywhere.

Hopefully looking just at the 1st commit will reduce the noise and make it easier to review.

### `SourcePosition`

There is one annoyance with the API which I haven't been able solve:

`SourcePosition` is a wrapper around a pointer, which can only be created from the current position of `Source`. Due to the invariant mentioned above, therefore `SourcePosition` is always in bounds of the source text, and points to a UTF-8 character boundary. So `Source` can be rewound to a `SourcePosition` cheaply, without any checks. I had originally envisaged `Source::set_position` being a safe function, as `SourcePosition` enforces the necessary invariants itself.

The fly in the ointment is that a `SourcePosition` could theoretically have been created from *another* `Source`. If that was the case, it would be out of bounds, and it would be instant UB. Consequently, `Source::set_position` has to be an unsafe function.

This feels rather ridiculous. *Of course* the parser won't create 2 Lexers at the same time. But still it's *possible*, so I think better to take the strict approach and make it unsafe until can find a way to statically prove the safety by some other means. Any ideas?

## Oddity in the benchmarks

There's something really odd going on with the semantic benchmark for `pdf.mjs`.

While I was developing this, small and seemingly irrelevant changes would flip that benchmark from +0.5% or so to -4%, and then another small change would flip it back.

What I don't understand is that parsing happens outside of the measurement loop in the semantic benchmark, so the parser shouldn't have *any* effect either way on semantic's benchmarks.

If CodSpeed's flame graph is to be believed, most of the negative effect appears to be a large Vec reallocation happening somewhere in semantic.

I've ruled out a few things: The AST produced by the parser for `pdf.mjs` after this PR is identical to what it was before. And semantic's `nodes` and `scopes` Vecs are same length as they were before. Nothing seems to have changed!

I really am at a loss to explain it. Have you seen anything like this before?

One possibility is a fault in my unsafe code which is manifesting only with `pdf.mjs`, and it's triggering UB, which I guess could explain the weird effects. I'm running the parser on `pdf.mjs` in Miri now and will see if it finds anything (Miri doesn't find any problem running the tests). It's been running for over an hour now. Hopefully it'll be done by morning!

I feel like this shouldn't merged until that question is resolved, so marking this as draft in the meantime.

### DIFF
--- a/crates/oxc_parser/src/lexer/source.rs
+++ b/crates/oxc_parser/src/lexer/source.rs
@@ -1,0 +1,410 @@
+#![allow(clippy::unnecessary_safety_comment)]
+
+use crate::MAX_LEN;
+
+use std::{marker::PhantomData, slice, str};
+
+/// `Source` holds the source text for the lexer, and provides APIs to read it.
+///
+/// It provides a cursor which allows consuming source text either as `char`s, or as bytes.
+/// It replaces `std::str::Chars` iterator which performed the same function previously,
+/// but was less flexible as only allowed consuming source char by char.
+///
+/// Consuming source text byte-by-byte is often more performant than char-by-char.
+///
+/// `Source` provides:
+///
+/// * Safe API for consuming source char-by-char (`Source::next_char`, `Source::peek_char`).
+/// * Safe API for peeking next source byte (`Source::peek_byte`).
+/// * Unsafe API for consuming source byte-by-byte (`Source::next_byte`).
+/// * Mostly-safe API for rewinding to a previous position in source
+///   (`Source::position`, `Source::set_position`).
+///
+/// # Composition of `Source`
+///
+/// * `start` is pointer to start of source text.
+/// * `end` is pointer to end of source text.
+/// * `ptr` is cursor for current position in source text.
+///
+/// # Invariants of `Source`
+///
+/// 1. `start` <= `end`
+/// 2. The region of memory bounded between `start` and `end` must be initialized,
+///    a single allocation, and contain the bytes of a valid UTF-8 string.
+/// 3. `ptr` must always be >= `start` and <= `end`.
+///    i.e. cursor always within bounds of source text `&str`, or 1 byte after last byte
+///    of source text (positioned on EOF).
+/// 4. `ptr` must always point to a UTF-8 character boundary, or EOF.
+///    i.e. pointing to *1st* byte of a UTF-8 character.
+///
+/// These invariants are the same as `std::str::Chars`, except `Source` allows temporarily
+/// breaking invariant (4) to step through source text byte-by-byte.
+///
+/// Invariants (1), (2) and (3) must be upheld at all times.
+/// Invariant (4) can be temporarily broken, as long as caller ensures it's satisfied again.
+///
+/// Invariants (1) and (2) are enforced by initializing `start` and `end` from a valid `&str`,
+/// and they are never modified after initialization.
+///
+/// Safe methods of `Source` enforce invariant (3) i.e. they do not allow reading past EOF.
+/// Unsafe methods e.g. `Source::next_byte_unchecked` and `Source::peek_byte_unchecked`
+/// require caller to uphold this invariant.
+///
+/// Invariant (4) is the most difficult to satisfy.
+/// `Source::next_char` relies on source text being valid UTF-8 to provide a safe API which
+/// upholds this invariant.
+/// `Source::next_byte` requires very careful use as it may violate invariant (4).
+/// That is fine temporarily, but caller *must* ensure the safety conditions of `Source::next_byte`
+/// are satisfied, to restore this invariant before passing control back to other code.
+/// It will often be preferable to instead use `Source::peek_byte`, followed by `Source::next_char`,
+/// which are safe methods, and compiler will often reduce to equally efficient code.
+#[derive(Clone)]
+pub(super) struct Source<'a> {
+    /// Pointer to start of source string. Never altered after initialization.
+    start: *const u8,
+    /// Pointer to end of source string. Never altered after initialization.
+    end: *const u8,
+    /// Pointer to current position in source string
+    ptr: *const u8,
+    /// Marker for immutable borrow of source string
+    _marker: PhantomData<&'a str>,
+}
+
+impl<'a> Source<'a> {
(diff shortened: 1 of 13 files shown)


crates/oxc_parser/src/lexer/source.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Move current position back by `n` bytes.
+    ///
+    /// # Panic
+    /// Panics if:
+    /// * `n` is 0.
+    /// * `n` is greater than current offset in source.
+    /// * Moving back `n` bytes would not place current position on a UTF-8 character boundary.
+    #[inline]
+    pub(super) fn back(&mut self, n: usize) {
+        assert!(n > 0, "Cannot call `Source::back` with 0");
```
Boshen, line 178: Possibly best to comment these `assert`s are intentional for compiler hints.

Boshen:
image

Interesting to see most of the speed ups come from `identifier_tail`, which is the only function that shows up from linter performance profiles.

Boshen (approved):

Boshen:
The semantic performance change is flaky, we can safely ignore it.

crates/oxc_parser/src/lexer/source.rs
overlookmotel: That assert is actually essential to make the function safe, rather than a compiler hint. But yes you're right, it wasn't clear why it was there, and needed a comment. Have added one.

Boshen: > to make the function safe

All the other functions use `debug_assert`, so I couldn't tell whether all the `asserts` need to be here in the release build.

Boshen:
Don't worry about the codspeed report, feel free to merge once you're satisfied with code since this is still marked as a draft.

overlookmotel:
> Interesting to see most of the speed ups come from `identifier_tail`, which is the only function that shows up from linter performance profiles.

I'm a bit dubious about CodSpeed's flame graph diffs. I think the `BYTE_HANDLERS` jump table makes it difficult for it to "match up" the functions before and after, perhaps because they're all called `core::ops::function::FnOnce::call_once`. In this case, it shows one byte handler function as being 6 x faster (the one in green) but then another one is 99.76% slower (the one in red). I don't think either of those results is plausible, so likely it's comparing one byte handler before to a different one after in both cases.

> The semantic performance change is flaky, we can safely ignore it.

I'm still a bit concerned it may possibly indicate UB. Miri run completed last night but the output was unreadable because it found 1150 memory leaks! (the ones we already know about) Am running it again with memory leak detection disabled. Will report back.

Can I ask: Have you ever seen flaky results with such a large margin (4%) before?

Also, now I've pushed the extra commit which added a comment, it's got much worse! (though that may be in part due to unrelated changes on `main` since last night)

Boshen:
I rebased and the report is correct now.

codspeed sometimes get flaky when there is a large memory allocation, don't worry about it.

overlookmotel:
Have rebased on top of #2301 to see if the lexer flame graph with named functions sheds any more light on the weird semantic benchmark.

Am also running Miri again (4 hours and counting!). If it doesn't find anything, I think we're all good. But I am nervous of merging this until Miri gives the all clear. There's so much unsafe raw pointer action in this PR it'd be easy for me to have made a mistake which causes UB.

overlookmotel:
Miri gives it a pass. Seems I was worrying too much.

So I think now ready to merge. But @boshen could you please merge #2301 and #2304 first? (if you're happy with them)

Boshen:
I'll run all the fuzzers when I wake up.

Boshen:
random bytes fuzzer: the parser timed out on `$g<  g<g<(b<b<(r=b<(r=g<(i<b<(r=b=g<(i<b<(r=b<(gb=g<(i<b<(r=b<(g<(g<g<(i<b<fh?.<[h<(/(/`, which is a know issue around rewind.

AST fuzzer passed.

I think we are good to go after the merge conflict is resolved.

overlookmotel:
Thanks very much for running the fuzzer. It feels like we can be more confident now that no problem with the changes introduced in this PR.

Have rebased on main, and now ready for merge once CI has passed.

overlookmotel:
CI passes. Have merged. Hooray! Thanks loads for all your help on this @boshen.

Issue #610, opened 2023-07-25
feat(linter): eslint/no-import-assign (partially done, in nursery category)


DonIsaac:
Implement ESLint's [`no-import-assign`](https://eslint.org/docs/latest/rules/no-import-assign) rule.

To get started, run
```sh
just new-rule no-import-assign
```

DeividAlmeida:
hi, i want to continue this rule, but i saw that it's implemented, what can i do to conclude ?

mysteryven:
> hi, i want to continue this rule, but i saw that it's implemented, what can i do to conclude ?

There are [some cases](https://github.com/oxc-project/oxc/blob/c6eb519417d4d6328a6a78de861f706541f56a39/crates/oxc_linter/src/rules/eslint/no_import_assign.rs#L158-L182) that are not supported yet, we can try to pass more!

DeividAlmeida:
should i open a else block to deal with a Read end ReadWrite flag in no_import_assign.rs?

DeividAlmeida:
what approach do you suggest in this case?

Boshen:
> should i open a else block to deal with a Read end ReadWrite flag in no_import_assign.rs?

You may give it a try 😁

DeividAlmeida:
To improve this should improve all its references like @Boshen said here [#2164 ](https://github.com/oxc-project/oxc/pull/2164)

Boshen:
I think we can port this exactly https://github.com/eslint/eslint/blob/972ef155a94ad2cc85db7d209ad869869222c14c/lib/rules/no-import-assign.js#L204-L206

PR #2298, opened 2024-02-04
feat(semantic): add ReferenceFlag::MemberModified for MemberExpression assign

Dunqing:
Looks like we need to add a new `ReferenceFlag` to represent this change.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1811,6 +1811,7 @@ impl<'a> SemanticBuilder<'a> {
             AstKind::IdentifierReference(_)
         ));
 
+        let mut is_member_expression = false;
         for (curr, parent) in self
             .nodes
             .iter_parents(self.current_node_id)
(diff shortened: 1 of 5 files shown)


Dunqing:
* `main`
  * **PR #2298** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2298?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2298?utm_source=stack-comment).

Dunqing:
Still cannot fully support #610 

In this case `import * as mod from 'mod'; mod.named.prop = 0` is expected to pass, but the reference flag is `ReferenceFlag::MemberModified`. But this case allows us to modify `named.prop`, Because the `prop` is not an `ImportBinding` . So I think we cannot fully support all cases by `ReferenceFlag`

Dunqing:
So far, we don't need to use this reference flag.

PR #2301, opened 2024-02-04
refactor(parser): name byte handler functions

overlookmotel:
This PR solves the problem of lexer byte handlers all being called `core::ops::function::FnOnce::call_once` in the flame graphs on CodSpeed, by defining them as named functions instead of closures.

Pure refactor, no substantive changes.


overlookmotel:
It works! Lexer flame graphs are now much more revealing. Identifiers take the majority of the time, with strings (`QOT` handler) the runner-up.

PR #2304, opened 2024-02-04
chore(benchmark): add radix-ui.jsx as a real world jsx example

overlookmotel:
Following #2297, this adds another benchmark.

This one is from radix-ui website. I've chosen this particular file because it differs from the other benchmark sources in 3 ways:

1. JSX not TSX (despite the file extension).
2. Contains no logic, only JSX component hierarchy, and content text.
3. Very small (60 LOC).

The last is particularly important, I think. Often developers will be working on small files (single component per file convention). And some possible directions for the parser (SIMD etc) involve optimizing chewing through chunks of text, with a de-opt at the end to process the final batch of bytes. If that imposes a penalty on short files, this benchmark will surface it.

### DIFF
--- a/tasks/common/src/test_file.rs
+++ b/tasks/common/src/test_file.rs
@@ -45,8 +45,10 @@ impl TestFiles {
         let files = [
             // TypeScript syntax (2.81MB)
             "https://raw.githubusercontent.com/microsoft/TypeScript/v5.3.3/src/compiler/checker.ts",
-            // Realword tsx (1.0M)
+            // Real world app tsx (1.0M)
             "https://raw.githubusercontent.com/oxc-project/benchmark-files/main/cal.com.tsx",
+            // Real world content-heavy app jsx (3K)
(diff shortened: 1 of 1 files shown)


Boshen:
Let me copy this file over to oxc-project/benchmark-files, since this file can change overtime.

Boshen (approved):

overlookmotel:
Ah of course, I didn't consider that the files could change. Doh! I should have used a permalink.

If you prefer to avoid hosting the files in an OXC repo, here's the permalink for the one I added:

https://raw.githubusercontent.com/radix-ui/website/6cf13bab6e56e8814f8e5cac156587c0fefe994d/components/marketing/AdoptionSection.tsx

Not sure which file you took from cal.com, but could take the same approach for that.

If you want to, let me know which file and I'll make a PR.

Boshen:
It's already done ;-)

overlookmotel:
OK cool!

PR #2324, opened 2024-02-05
fix(linter): false positive with require-yield where yield is anywhere but function scope

### LINKED ISSUE #2323: False positive in require-yield where yield is anywhere but function scope
*Working*

```js
function* foo() { yield 0; }
```

*Not working (simplified test case)*

```js
function* foo() { while (true) { yield 0; } }
```

*False positive result*

```
  ⚠ eslint(require-yield): This generator function does not have 'yield'
   ╭─[require_yield.tsx:1:11]
 1 │ function* foo() { while (true) { yield 0; } }
   ·           ───
   ╰────
```

maurice:
Fixes #2323

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -226,16 +226,23 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     pub fn set_function_node_flag(&mut self, flag: NodeFlags) {
+        if let Some(node_id) = self.get_function_node_id() {
+            *self.nodes.get_node_mut(node_id).flags_mut() |= flag;
+        }
+    }
+
+    fn get_function_node_id(&self) -> Option<AstNodeId> {
         if self.current_scope_flags().is_function() {
-            *self.nodes.get_node_mut(self.scope.get_node_id(self.current_scope_id)).flags_mut() |=
-                flag;
-        } else {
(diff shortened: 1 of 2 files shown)


Boshen:
I'll need to check the best place to fix this.

crates/oxc_semantic/src/builder.rs
```diff
@@ -229,6 +229,12 @@ impl<'a> SemanticBuilder<'a> {
         if self.current_scope_flags().is_function() {
             *self.nodes.get_node_mut(self.scope.get_node_id(self.current_scope_id)).flags_mut() |=
                 flag;
+        } else {
+            for scope_id in self.scope.ancestors(self.current_scope_id) {
+                if self.scope.get_flags(scope_id).is_function() {
+                    *self.nodes.get_node_mut(self.scope.get_node_id(scope_id)).flags_mut() |= flag;
+                }
+            }
```
maurice, line 237: hmm, sorry I pushed a bit too eagerly. I probably need a `break` here after finding the parent function scope.

I could also refactor it a bit to avoid the slight duplication between branches, ie, first find the function scope (either current or ancestor), **then** set the flag

maurice: Done

maurice:
> I'll need to check the best place to fix this.

Np, let me know if I can help

PR #2325, opened 2024-02-05
refactor(linter/config): Use serde::Deserialize for config parsing

### LINKED ISSUE #2258: Linter: try serde::Deserialize for some of the eslint config fields
https://github.com/oxc-project/oxc/blob/a95a16c2aea61fefcb3435d54acf52d456864369/crates/oxc_linter/src/config/mod.rs#L150-L364

All of the fields are parsed manually, we may be able to clean it up by using serde::Deserialize.

I have some examples here:

* https://github.com/oxc-project/oxc-resolver/blob/main/src/package_json.rs
* https://github.com/oxc-project/oxc-resolver/blob/main/src/tsconfig.rs

leaysgur:
Fixes #2258 

### Overview

- Re-implemented the config parser to use `serde::Deserialize`
- In order to benefit from it as much as possible, avoided implementing custom deserializers and tried to use attributes as much as possible
  - This required some changes to the caller signatures...

➕ 

- Fixed a bug that did not support for abbreviations like `"rule-name": 1`
- Fixed settings that should have been located in `settings.react` but were not

### TODOs

- [x] rules
  - [x] bug fix: `"rule-name": 0 | 1 | 2` does not work
- [x] settings
  - [x] move `react` plugin specific settings to under its name
- [x] env
- [x] misc...


leaysgur:
Progress 🏃🏻 

- Mostly todos are done
- [ ] Need to map error string for `ESLintConfig::deserialize()`
  - https://github.com/oxc-project/oxc/pull/2325/files#diff-f6a4fa4352c9a511864c80950203cd8419fbc3813cf4520d4039f2324fbd5780R62
  - but how...? I'm a bit stacked 😿 
- [ ] Consider to implement custom deserializer for settings and env too?
  - want to keep them as simple as possible

Boshen (approved):
Thank you so much! This is so much cleaner than before!

Issue #2296, opened 2024-02-04
SIMD in the lexer


Boshen:
As discussed in #2285, we'd like to embark on the journey of SIMD.

The current state of SIMD in Rust is functional, successful projects include https://github.com/rusticstuff/simdutf8 and https://github.com/BurntSushi/memchr. 

`portable-simd` is not an option due to its instability.

---

In this task, you will implement one of the tokenization methods for the following tokens in the lexer.

* Identifiers https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/identifier.rs#L8
* Strings https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/string.rs#L6
* Whitespace https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/byte_handlers.rs#L101
* Comments https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/comment.rs#L9 https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/comment.rs#L25
* Integers https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_parser/src/lexer/numeric.rs#L30

Specification of these tokens can be found at https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar

To get things working, you only need to implement for the architecture of your development machine.

For example
* aarch64_neon: https://github.com/rusticstuff/simdutf8/blob/main/src/implementation/aarch64/mod.rs
* avx2: https://github.com/rusticstuff/simdutf8/blob/main/src/implementation/x86/mod.rs

---

Some of these are easy, @overlookmotel drafted some algorithms:

Multi-line comments:
* At first search for `*/`, `\r`, `\n` or `0xE2` (1st byte of either `LS` or `PS`).
* Once a line break is found, just search for `*/` or `0xE2` (as it's not relevant whether there's further line breaks now).
* Put handling `0xE2` on a `#[cold]` branch as irregular line breaks (and other Unicode characters starting with `0xE2`) are rare.
[...]

PR #90, opened 2023-03-02
refactor: pad raw content to parse with 16 empty bytes

YoniFei:
Addresses #43 

Initial crude attempt.
Unlike in @strager's [padded-string](https://github.com/quick-lint/quick-lint-js/blob/master/src/quick-lint-js/container/padded-string.cpp) which actually manages the underlying memory, I just wrapped a `String`.

Issues/thoughts:

1. I had to alter `SINGLE_CHAR_TOKENS` so that an empty byte char (null terminator) is considered EOF. Without this modification we get an error in `asi()->can_insert_semicolon()` since our token is `Kind::Undetermined`. I'm not sure if this is a hack that can cause issues down the road, or completely fine.
2. The `Lexer`'s constructor receiving a `PaddedStringView` is what enforces we'll have the 16 byte padding. but perhaps the internals can be refactored so it's clearer that it's what allows making assumptions and/or so we can only work with it where needed (in the SIMD parsing).
3. File load performance shouldn't take a hit, but I haven't verified this. Need to validate `read_to_string` and `push_str` aren't a regression.
4. Is there an easy way to benchmark E2E before/after?

Issue #43, opened 2023-02-24
Refactor: try aligned string for simd in lexer


Boshen:
https://github.com/Boshen/oxc/pull/26#discussion_r1116408111

> You can guarantee that remainder is always empty or unimportant (thus you don't need to check for it) if you add enough zeros at the end of bytes. I do this in quick-lint-js when I load the file from the filesystem, so everything can safely assume that the extra bytes are present. (I have a dedicated type for this called padded_string_view which doesn't allow breaking this invariant.)

YoniFeng:
Interesting.
This is ~purely for ergonomics, right?
i.e. I'd expect perf gains to be non-existent or negligible, even if implemented in as close to a "zero-cost abstraction" as possible (but of course, only measuring will tell).

As @stager mentioned, this has to be done on the filesystem boundary/right when loading the file's contents.
That way only a single reallocation+copy is needed. (I wonder if there's a way to avoid even that, since file contents can be large. I'm guessing using `fs::metadata.len()` isn't worth it though).

Do you have any thoughts on where this shared "padded_string" should be located, filesystem structure-wise?
Anything accessing the Parser will need to have access to it as well.
Should it be in a new crate?

Boshen:
By looking at the PR, I think this over complicated things without significant performance increase.
 
If we are ever going to release oxc as crates, exposing this `PaddedStringView` is not really user friendly. 

So I suggest we close this issue and the PR. I'm glad you tried you this. Perhaps you can pop into our Discord channel for quicker feedback, you'll be able see whether I'm online or not.

YoniFeng:
If there's potential for future SIMD optimizations, I think it could be ~handy overall.

The external API can always have an overload that takes a `&str`, it just means that using it will mean allocation+copy.
I'll try to work on it a little more, and if you still don't like it we can throw it away?

Boshen:
I'm not seeing any perf improvements and I don't want to waste your time on this for such small gain. 

A more interesting task would be the allocator one.

YoniFeng:
@Boshen can close this issue

PR #2327, opened 2024-02-05
perf(parser): faster line comment

sno2:
Utilizes memchr to find the end of the line comment which _might_ yield performance improvements. Also, adds a new `Source::eat_until_byte3` along with `Source::set_eof` that allows us to naturally integrate memchr-based searches
into the lexer.

Thankfully, memchr does not have any dependencies so this change does not add that much bloat to the binary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1593,6 +1593,7 @@ version = "0.6.0"
 dependencies = [
  "assert-unchecked",
  "bitflags 2.4.2",
+ "memchr",
  "miette",
  "num-bigint",
  "ouroboros",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ ignore                    = { version = "0.4.22" }
 itertools                 = { version = "0.12.1" }
 jemallocator              = { version = "0.5.4" }
 lazy_static               = { version = "1.4.0" }
+memchr                    = { version = "2.7.1" }
 miette                    = { version = "5.10.0", features = ["fancy-no-backtrace"] }
 mimalloc                  = { version = "0.1.39" }
 num-bigint                = { version = "0.4.4" }
--- a/crates/oxc_parser/Cargo.toml
+++ b/crates/oxc_parser/Cargo.toml
@@ -28,6 +28,7 @@ oxc_index       = { workspace = true }
 assert-unchecked = { workspace = true }
 bitflags         = { workspace = true }
 rustc-hash       = { workspace = true }
+memchr           = { workspace = true }
 num-bigint       = { workspace = true }
 
 [dev-dependencies]
--- a/crates/oxc_parser/src/lexer/comment.rs
+++ b/crates/oxc_parser/src/lexer/comment.rs
@@ -8,15 +8,25 @@ impl<'a> Lexer<'a> {
     #[allow(clippy::cast_possible_truncation)]
     pub(super) fn skip_single_line_comment(&mut self) -> Kind {
         let start = self.token.start;
-        while let Some(c) = self.next_char() {
-            if is_line_terminator(c) {
-                self.token.is_on_new_line = true;
-                self.trivia_builder
-                    .add_single_line_comment(start, self.offset() - c.len_utf8() as u32);
-                return Kind::Skip;
+
+        // The first byte of the UTF-8 encoding of U+2028 and U+2029.
+        let ps_ls_start_byte = 0xE2;
+
+        while let Some(byte) = self.source.eat_until_byte3(b'\n', b'\r', ps_ls_start_byte) {
+            // Handle ambiguity between PS/LS and some other UTF-8 character.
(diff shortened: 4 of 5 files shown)


overlookmotel:
@sno2 Thanks for this. Not had time to review yet, but can I ask one question: Is this intended as an effort towards #2296?

sno2:
> @sno2 Thanks for this. Not had time to review yet, but can I ask one question: Is this intended as an effort towards https://github.com/oxc-project/oxc/issues/2296?

This was more of an experiment to see how well `memchr` would fare for this use-case. It is not that useful, though, because most of the other SIMD-oriented parts require more than 3 bytes for searching. To be frank, this may not be worth merging at all...

Boshen:
> To be frank, this may not be worth merging at all given how little it affected performance

This sped up `checker.ts` which has lots of comments.

image

overlookmotel:
Thanks very much for coming back. Yes, the performance gain is a little underwhelming. I would have hoped for more, especially given that on CI, the SIMD lane size is probably 32 (I assume Github's runners support AVX2, but I don't think memchr supports AVX-512 to do 64-byte batches).

Nonetheless, it would be worth doing, as it does provide *some* speedup in heavily-commented code. And many incremental gains add up.

But... I'm afraid I think you're missing one edge case. It also needs to handle `0x0B` and `0x0C` (irregular whitespace) as these need "trivia"s created for them if they're found.

That means there are 5 x possible bytes to search for, which rules out memchr, as it only supports up to 3.

In my opinion (and please note I am no expert on this stuff, so could be wrong), we need to use a different approach from memchr's, and use "pshufd"-based table lookups to support searching for any number of possible bytes. If you're interested, I linked to an article about  this on #2296.

So ultimately, I'm afraid I don't think this specific PR is viable. But that doesn't mean it was wasted time. It is always valuable to test and rule out different possibilities. The result is surprising (at least to me), so thank you for exploring this avenue.

If you're willing and keen to push further down this path towards SIMD, but don't want to dig into writing SIMD assembler code line by line (what sane person does?), there is one massive thing you could help with: I find it hard to believe there is not already a crate available which does "phufd"-based byte searching, as it seems like a common need. But I can't find one! if you were able to track one down, that would basically solve how to do SIMD in OXC *entirely*.

Boshen:
Previous attempt on padding the source:
* https://github.com/oxc-project/oxc/issues/43
* https://github.com/oxc-project/oxc/pull/90

sno2:
> But... I'm afraid I think you're missing one edge case. It also needs to handle 0x0B and 0x0C (irregular whitespace) as these need "trivia"s created for them if they're found.

I don't believe that the `0x0B` and `0x0C` were handled properly in the first place:

https://github.com/oxc-project/oxc/blob/0f28dc8894723cfc4a23063a3429f486833fbc23/crates/oxc_syntax/src/identifier.rs#L42-L64

https://github.com/oxc-project/oxc/blob/0f28dc8894723cfc4a23063a3429f486833fbc23/crates/oxc_parser/src/lexer/comment.rs#L11-L18

> In my opinion (and please note I am no expert on this stuff, so could be wrong), we need to use a different approach from memchr's, and use "pshufd"-based table lookups to support searching for any number of possible bytes. If you're interested, I linked to an article about this on https://github.com/oxc-project/oxc/issues/2296.

I am looking into how we could try to implement this.

> So ultimately, I'm afraid I don't think this specific PR is viable. But that doesn't mean it was wasted time. It is always valuable to test and rule out different possibilities. The result is surprising (at least to me), so thank you for exploring this avenue.

Yep, I agree.

> If you're willing and keen to push further down this path towards SIMD, but don't want to dig into writing SIMD assembler code line by line (what sane person does?), there is one massive thing you could help with: I find it hard to believe there is not already a crate available which does "phufd"-based byte searching, as it seems like a common need. But I can't find one! if you were able to track one down, that would basically solve how to do SIMD in OXC entirely.

I am not sure if it will be this simple. In many cases, we need to defer to a scalar version whenever we see non-ASCII characters which would not be possible in byte-based searches.

overlookmotel:
> I am not sure if it will be this simple. In many cases, we need to defer to a scalar version whenever we see non-ASCII characters which would not be possible in byte-based searches.

I'm currently working on a a re-write of lexing identifiers. It doesn't use SIMD, but is designed to be easily convertable to SIMD later: https://github.com/overlookmotel/oxc/blob/lexer-string5/crates/oxc_parser/src/lexer/identifier.rs
(please excuse the messy code, still very much a work in progress)

You're right about having to bust out to scalar if unicode chars are found, but approach I've taken is to speed through ASCII as fast as possible and then Unicode is a de-opt (and on a `#[cold]` path so compiler assumes it won't happen in most cases). This approach seems to work - produces about 60% speed-up on the Lexer benchmarks. Obviously using SIMD would be a further speed boost.

But, yeah, I think this is one approach to the problem you've identified.

PR #2335, opened 2024-02-06
fix(semantic): add export symbol flag to identifiers in export declarations

DonIsaac:
# What This PR Does

Consider the following code:
```js
export function foo() {}
function bar () {}
class Baz {}

export { bar }
export default Baz
```

Since symbol flags are added at the time of declaration, `bar` and `Baz` never get flagged with `SymbolFlags::Export`. This PR checks identifiers referenced inside `ExportNamedDeclaration`s and `ExportDefaultDeclaration`s, adding export flags to all resolvable symbols.

## Other Changes
- test(semantic): improvements to class and symbol testers (including better diagnostics)
- feat(semantic): implement `Debug` for `ModuleRecord` (in a non-crashing way)

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -644,6 +644,51 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         self.leave_node(kind);
     }
 
+    fn visit_export_named_declaration(&mut self, decl: &ExportNamedDeclaration<'a>) {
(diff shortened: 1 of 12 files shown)


Boshen:
I copied over `impl Debug for ModuleRecord`.

The rest of the code is hard to merge at the current state, so I'm going to close this PR for house keeping.

Dunqing:
I would love to have this, I will re-implement it

PR #2339, opened 2024-02-07
refactor(parser): wrapper type for parser

overlookmotel:
Split parser into public interface `Parser` and internal implementation `ParserImpl`.

This involves no changes to public API.

This change is a bit annoying, but justification is that it's required for #2341, which I believe to be very worthwhile.

The `ParserOptions` type also makes it a bit clearer what the defaults for `allow_return_outside_function` and `preserve_parens` are. It came as a surprise to me that `preserve_parens` defaults to `true`, and this refactor makes that a bit more obvious when reading the code.

All the real changes are in [oxc_parser/src/lib.rs](https://github.com/oxc-project/oxc/pull/2339/files#diff-8e59dfd35fc50b6ac9a9ccd991e25c8b5d30826e006d565a2e01f3d15dc5f7cb). The rest of the diff is basically replacing `Parser` with `ParserImpl` everywhere else.

PR #2340, opened 2024-02-07
parser(refactor): promise only one `Source` on a thread at a time

overlookmotel:
Introduce invariant that only a single `lexer::Source` can exist on a thread at one time.

This is a preparatory step for #2341.

2 notes:

Restriction is only 1 x `ParserImpl` / `Lexer` / `Source` on 1 *thread* at a time, not globally. So this does not prevent parsing multiple files simultaneously on different threads.

Restriction does not apply to public type `Parser`, only `ParserImpl`. `ParserImpl`s are not created in created in `Parser::new`, but instead in `Parser::parse`, where they're created and then immediately consumed. So the end user is also free to create multiple `Parser` instances (if they want to for some reason) on the same thread.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -44,7 +44,7 @@ pub use self::{
     number::{parse_big_int, parse_float, parse_int},
     token::Token,
 };
-use crate::diagnostics;
+use crate::{diagnostics, UniquePromise};
 
 #[derive(Debug, Clone, Copy)]
 pub struct LexerCheckpoint<'a> {
@@ -97,8 +97,17 @@ pub struct Lexer<'a> {
 
 #[allow(clippy::unused_self)]
 impl<'a> Lexer<'a> {
-    pub fn new(allocator: &'a Allocator, source_text: &'a str, source_type: SourceType) -> Self {
-        let source = Source::new(source_text);
+    /// Create new `Lexer`.
+    ///
+    /// Requiring a `UniquePromise` to be provided guarantees only 1 `Lexer` can exist
+    /// on a single thread at one time.
+    pub(super) fn new(
+        allocator: &'a Allocator,
+        source_text: &'a str,
+        source_type: SourceType,
+        unique: UniquePromise,
+    ) -> Self {
+        let source = Source::new(source_text, unique);
 
         // The first token is at the start of file, so is allows on a new line
         let token = Token::new_on_new_line();
@@ -114,6 +123,18 @@ impl<'a> Lexer<'a> {
             escaped_strings: FxHashMap::default(),
             escaped_templates: FxHashMap::default(),
         }
+    }
+
+    /// Backdoor to create a `Lexer` without holding a `UniquePromise`, for benchmarks.
+    /// This function must NOT be exposed in public API as it breaks safety invariants.
+    #[cfg(feature = "benchmarking")]
+    pub fn new_for_benchmarks(
+        allocator: &'a Allocator,
+        source_text: &'a str,
+        source_type: SourceType,
+    ) -> Self {
+        let unique = UniquePromise::new_for_tests();
+        Self::new(allocator, source_text, source_type, unique)
     }
 
     /// Remaining string from `Chars`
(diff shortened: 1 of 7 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2340?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `main`
  * **PR #2339** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #2340** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2340?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2340?utm_source=stack-comment).

overlookmotel:
PS: I initially wanted to use the type system to enforce this invariant, rather than making `ParserImpl::new`, `Lexer::new` and `Source::new` unsafe.

This would be possible using a "uniqueness token" type which only `Parser::parse` can create, and the `new` methods would require this token as a function parameter.

However, this would (by design) make it impossible to create a `ParserImpl` except in `Parser::parse`, which is incompatible with some of the unit tests, which need to create a `ParserImpl` and manipulate it manually.

overlookmotel:
Actually, yes, this is possible. Am going to refactor to use types to enforce the invariant, rather than `unsafe`.

Changed to draft until I've done that.

crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -97,7 +97,20 @@ pub struct Lexer<'a> {
...
 #[allow(clippy::unused_self)]
 impl<'a> Lexer<'a> {
-    pub fn new(allocator: &'a Allocator, source_text: &'a str, source_type: SourceType) -> Self {
+    /// Create new `Lexer`.
+    ///
+    /// # SAFETY
+    /// Caller must not allow more than one `Lexer` to exist at any one time on a thread.
+    /// In practice, this is easy to satisfy. Only means that, apart from `ParserImpl::new`,
+    /// no method of `ParserImpl` or `Lexer` can call this and create another `Lexer`.
+    #[allow(clippy::missing_safety_doc)] // Clippy is wrong!
```
Boshen, line 106: I thought the comment need to start with `SAFETY:`

crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -97,7 +97,20 @@ pub struct Lexer<'a> {
...
+    pub unsafe fn new(
+        allocator: &'a Allocator,
+        source_text: &'a str,
+        source_type: SourceType,
+    ) -> Self {
+        // SAFETY: We only create a single `Source`
+        #[allow(clippy::unnecessary_safety_comment)]
```
Boshen, line 113: This is an annoying rule it seems, I'll remove it tomorrow.

overlookmotel:
I've refactored to enforce this invariant through the type system instead of unsafe.

But have to stop for today now. I'm neglecting my day job!

crates/oxc_parser/src/lexer/mod.rs
overlookmotel: I think it's good to leave it enabled. It's helpful when you convert an unsafe function to safe, as clippy tells you everywhere you can now remove the unsafe blocks and comments.

It can be disabled on a module-by-module basis where it comes up frequently.

overlookmotel:
> Yeah the `unsafe`s are annoying.
> 
> Besides, how do we even create a test case for this one?

My new type system-based approach removes the unsafe.

I don't think with this new approach it's possible to write tests for it, as there's just no API to do what is meant to be illegal. But that's good, right?

crates/oxc_parser/src/lexer/mod.rs
overlookmotel: I've been able to remove the unsafe, and hence remove these annoying clippy attrs in refactored version.

Clippy is happy with `# Safety` but refuses to recognise `# SAFETY`. Clippy is an anti-capitalist!

Boshen:
### Merge activity

* **Feb 8, 1:51 AM EST**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2340).

PR #2342, opened 2024-02-07
fix(linter): add missing typescript-eslint(_) prefix for some errors

maurice:
Running latest on one of my projects these warnings jumped out at me because they were "anonymous" vs the others.

This PR just adds the usual rule-name prefix to the errors where it was missing


Boshen:
Thank you!

PR #2358, opened 2024-02-09
feat(linter): eslint-plugin-jest: prefer-equality-matcher

eryue0220:
Rule detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-equality-matcher.ts)


mysteryven (approved):
Thank you!

mysteryven:
Do you think we can also add the Fixer support?

eryue0220:
> Do you think we can also add the Fixer support?

Yeah, sure I've a thought before, but the rule only provided suggestion. So I removed those code. 😅

PR #2353, opened 2024-02-08
perf(parser): eat whitespace after line break

overlookmotel:
Uses the `byte_search!` macro introduced in #2352 to consume whitespace after a line break.

PR #2366, opened 2024-02-09
perf(parser): optimize lexing strings

overlookmotel:
Optimize lexing strings a bit.

### DIFF
--- a/crates/oxc_parser/src/lexer/search.rs
+++ b/crates/oxc_parser/src/lexer/search.rs
@@ -394,6 +394,7 @@ macro_rules! byte_search {
         $table.use_table();
 
         let mut pos = $start;
+        #[allow(unused_unsafe)] // Silence warnings if macro called in unsafe code
         loop {
             if pos.addr() <= $lexer.source.end_for_batch_search_addr() {
                 // Search a batch of `SEARCH_BATCH_SIZE` bytes.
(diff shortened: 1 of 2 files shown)


overlookmotel:
* `main`
  * **PR #2366** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2366?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2366?utm_source=stack-comment).

overlookmotel:
This is as good as I can do for now. Still a bit disappointing.

[As discussed](https://github.com/oxc-project/oxc/pull/2353#issuecomment-1935300679) on #2353, the increasing use of macros may be a concern. But could we go with it for now? If we introduce SIMD, or implement #2368, the macro will change a lot, and it'd be easier to keep those changes in one place (the macro) until it's fairly locked. At that point, as you said, if the macros slow down compilation we could expand them all manually.

Boshen:
> But could we go with it for now?

Of course! Improve then refactor.

overlookmotel:
> Of course! Improve then refactor.

Thanks Boshen!

PR #2376, opened 2024-02-10
perf(lint/no_var_requires): quicker way to check if the `IdentifierReference` point to a global variable

hyfdev:
The old code use this function to do the checking, which is too much for this simple checking

https://github.com/oxc-project/oxc/blob/ef336cb66b4746073e9d6a2f950d46185fec3cf6/crates/oxc_linter/src/ast_util.rs#L267-L276

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -384,10 +384,7 @@ pub fn get_new_expr_ident_name<'a>(new_expr: &'a NewExpression<'a>) -> Option<&'
 
 /// Check if the given [IdentifierReference] is a global reference.
 /// Such as `window`, `document`, `globalThis`, etc.
-pub fn is_global_reference(
-    ident: &IdentifierReference,
-    ctx: &LintContext,
-) -> bool {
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/ast_util.rs
```diff
@@ -381,3 +381,29 @@ pub fn get_new_expr_ident_name<'a>(new_expr: &'a NewExpression<'a>) -> Option<&'
...
+    reference.symbol_id().is_none()
+}
+
+pub fn is_global_require_call(call_expr: &CallExpression, ctx: &LintContext) -> bool {
+    if call_expr.arguments.len() != 1 {
+        return false;
+    }
+
+    if let Expression::Identifier(id_ref) = &call_expr.callee {
+        id_ref.name == "require" && is_global_reference(id_ref, ctx)
```
hyf0, line 405: I remove some of validations

https://github.com/oxc-project/oxc/blob/ef336cb66b4746073e9d6a2f950d46185fec3cf6/crates/oxc_ast/src/ast/js.rs#L728-L730

I guess even users use `require` in a wrong way, we probably still want to catch them in this rule.

Dunqing (approved):

PR #2377, opened 2024-02-10
perf(parser): consume multi-line comments faster

overlookmotel:
Consume multi-line comments faster.

* Initially search for `*/`, `\r`, `\n` or `0xE2` (first byte of irregular line breaks).
* Once a line break is found, switch to faster search which only looks for `*/`, as it's not relevant whether there are more line breaks or not.

Using `memchr` for the 2nd simpler search, as it's efficient for a search with only one "needle".

Initializing `memchr::memmem::Finder` is fairly expensive, and tried numerous ways to handle it. This is most performant way I could find. Any ideas how to avoid re-creating it for each Lexer pass? (it can't be a `static` as `Finder::new` is not a const function, and `lazy_static!` is too costly)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1593,6 +1593,7 @@ version = "0.7.0"
 dependencies = [
  "assert-unchecked",
  "bitflags 2.4.2",
+ "memchr",
  "miette",
  "num-bigint",
  "ouroboros",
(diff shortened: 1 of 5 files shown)


overlookmotel:
* `main`
  * **PR #2377** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2377?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2377?utm_source=stack-comment).

overlookmotel:
Ready for review now.

Boshen:
Nice! Lexer benchmark shows 3% improvement!

PR #2404, opened 2024-02-12
linter: implement from_configuration for jsx-a11y autocomplete-valid

### LINKED ISSUE #2362: linter: Implement `from_configuration()` for `jsx_a11y/autocomplete-valid`
- [ ] implement (using `utils::get_element_type`)
- [ ] test w/ config
- [ ] test w/ settings

tudorbarbu:
This closes #2362 

While fixing this issue I've found that the test data differed so I've regenerated it.
There seems to be no easy way to do this (Maybe create a test regen command in just?)
This way if there are new tests added, a CI hook or something could regenerate them on new plugin releases(changes in tests might mean changes that need to be addressed in the logic as well)
I've also added the tests that pass in the plugin rule, but seem to fail here. Since I was not able to find if the test values are really correct, I've left them commented out

### DIFF
--- a/crates/oxc_linter/src/rules/jsx_a11y/autocomplete_valid.rs
+++ b/crates/oxc_linter/src/rules/jsx_a11y/autocomplete_valid.rs
@@ -1,4 +1,9 @@
-use crate::{context::LintContext, rule::Rule, utils::has_jsx_prop_lowercase, AstNode};
+use crate::{
+    context::LintContext,
+    rule::Rule,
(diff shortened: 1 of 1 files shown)


Boshen (approved):

Boshen:
Thank you!

Boshen:
> There seems to be no easy way to do this (Maybe create a test regen command in just?)

Yeah ... syncing the tests is difficult at the moment.

PR #2413, opened 2024-02-15
feat(linter) eslint: no-new-wrappers

jepcd:
Part of: #479 

docs: https://eslint.org/docs/latest/rules/no-new-wrappers

I am unsure of what do do with 2 of the eslint tests that use `languageOptions: { globals: { String: "off" } }` and `/* global Boolean:off */`.

I've commented them out as of now.

### DIFF
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -72,6 +72,7 @@ mod eslint {
     pub mod no_loss_of_precision;
     pub mod no_mixed_operators;
     pub mod no_new_symbol;
+    pub mod no_new_wrappers;
     pub mod no_obj_calls;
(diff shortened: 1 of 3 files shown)


camc314 (commented):
is this the same as https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/new_for_builtins.rs

or very similar?

should we merge the two rules.

Thoughts @Boshen ?

alakhpc:
That one does seem a lot more comprehensive and should be preferred.

The only reason I can see to include this, is that this falls under `correctness` while `new_for_builtins` falls under `pedantic`

Boshen:
I changed this to `pedantic` to be consistent with eslint.

Boshen (approved):

PR #2416, opened 2024-02-15
perf(parser): inline identifier handler into byte handlers

overlookmotel:
Inlining `identifier_name_handler` into the byte handlers for `a-z`, `A-Z`, `_` and `$` produces a modest performance boost, as it's a hot path.

### DIFF
--- a/crates/oxc_parser/src/lexer/identifier.rs
+++ b/crates/oxc_parser/src/lexer/identifier.rs
@@ -47,7 +47,12 @@ impl<'a> Lexer<'a> {
     /// # SAFETY
     /// * `self.source` must not be exhausted (at least 1 char remaining).
     /// * Next char must be ASCII.
-    #[allow(clippy::missing_safety_doc)] // Clippy is wrong!
+    // Clippy is wrong!
+    #[allow(clippy::missing_safety_doc)]
+    // Inline into all identifier byte handlers, which avoids a function call on a hot path.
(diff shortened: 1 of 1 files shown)


overlookmotel:
* `main`
  * **PR #2416** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2416?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2416?utm_source=stack-comment).

overlookmotel:
Will have to check if this bloats binary size too much. It may not be worth it.

How do I go about that? I was expecting the `bloat` CI job to run, but just realised it's not triggered on PRs (and maybe isn't for that purpose anyway).

Boshen:
> but just realised it's not triggered on PRs (and maybe isn't for that purpose anyway).

You need to click the top right "run workflow" from the UI https://github.com/oxc-project/oxc/actions/workflows/bloat.yml

It'll display the result on the action page once it's done.

Boshen:
I think this optimization is a bit overkill if you look at the absolute change.

overlookmotel:
Fair enough!

It might be worth revisiting once SIMD is implemented, as then `identifier_name_handler` will shrink considerably and cost of inlining it would drop, but that's a way off...

Issue #2295, opened 2024-02-04
Make `Atom` safe and faster


Boshen:
`Atom` had a memory leak which is not desirable: https://github.com/oxc-project/oxc/issues/1803

I removed the memory and miri test to the parser: https://github.com/oxc-project/oxc/pull/2294

The change introduced some performance regressions and a unsafe

https://github.com/oxc-project/oxc/blob/6002560fa1398bc4bc68696e402e51c776f4ed53/crates/oxc_span/src/atom.rs#L103-L104

The next goal is to remove this unsafe.

overlookmotel:
Can I tackle this? I already have a WIP which steals a lot of the efficient code of `CompactString`. @Boshen I know you're aware of that, but obviously you ran out of patience waiting for me (I don't blame you). I seem to have got side-tracked by trying to optimize the lexer!

But now that the memory leak is solved, perhaps it's a little less urgent?

Boshen:
> That sound OK to you?

I can't wait! But take your time and enjoy the fun!

Boshen:
Instead of exposing a single `Atom` type, I'm thinking along the lines of:

* `ArenaStr<'a>` with the functionaries of inlined string + arena allocated str
* `CompactStr` with the functionaries of inlined string + heap allocated str

The `ArenaStr<'a>` will have `to_compact_str` and `to_string` for downstream usages, the user has the choice of choosing the most convenient one for them.

overlookmotel:
This is really nice - nice combination of efficiency and usability.

Just to check a couple of things:

`ArenaStr<'a>` is what will be used in the AST? `to_compact_str` / `to_string` are for downstream users, if they don't want to deal with lifetimes in their own code, or if they want to mutate the string?

I'm unclear how `ArenaStr<'a>` can be size of a `u32`. I was imagining `ArenaStr<'a>` being 16 bytes:

* Bytes 0-7: Pointer
* Bytes 8-11: Length (`u32`)
* Bytes 12-15: Unused (though could find uses for this e.g. index into Vec of escaped strings, or flags for "String is in source code" / "String is escaped" / "String is Unicode").

We could probably squeeze it down smaller using pointer compression, but:

1. That optimization could also apply to `Vec`s and `Box`s in the AST, so I was thinking of it as a broader issue.
2. 16 is probably a good size for making majority of strings be stored inline (most JS identifiers are under 16 bytes).

Or have I misunderstood, and you have something else in mind?

Boshen:
[...]

PR #2417, opened 2024-02-15
chore(deps): update `bumpalo` crate

overlookmotel:
Latest version of `bumpalo` includes a couple of performance fixes for `String` (e.g. https://github.com/fitzgen/bumpalo/pull/229) which may help the parser a little.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,9 +185,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.14.0"
+version = "3.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
+checksum = "d32a994c2b3ca201d9b263612a374263f05e7adde37c4707f693dcd375076d1f"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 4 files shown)


overlookmotel:
* `main`
  * **PR #2417** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2417?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2417?utm_source=stack-comment).

overlookmotel:
What the hell? It actually hurts the benchmarks (very slightly), rather than improving them. I'll have to look into this.

overlookmotel:
This is a weird one. It was *definitely* an improvement for bumpalo. It just replicates what std's `String` does, and produces an 80x speed-up for long strings. Question is why is it not having that positive effect in OXC? Probably it's that JS string literals are typically quite short, so it's not hitting the "sweet spot", but I'd like to bore into it a bit more and check that assumption.

Thanks for the tip about git forks. Yes, I'll do that next time.

Boshen:
It's probably the benchmark not hitting the sweat spots. Let's merge either way.

overlookmotel:
OK, thanks for merging.

We can probably do better in the parser as building strings now happens in batches of 32 bytes, so can write each batch with 2 x `xmm` fixed-length writes rather than calls to `copy_nonoverlapping`.

#2295 and #2409 may both require a bit of an overhaul of string handling, so can look at it then.

PR #2418, opened 2024-02-15
feat(cli): support plugin disable

### LINKED ISSUE #1970: Is it possible to disable a whole "plugin" ?
I'd like to be able to disable "eslint-unicorn-plugin" rules altogether. Is there any specific syntax for this? Thanks.

tudorbarbu:
This should fix #1970

There was also a bug where import-plugin was not taken into account and import rules were always enabled

### DIFF
--- a/crates/oxc_cli/src/command.rs
+++ b/crates/oxc_cli/src/command.rs
@@ -104,6 +104,34 @@ pub struct EnablePlugins {
     pub react_perf_plugin: bool,
 }
 
+/// Disable Default Plugins
+#[derive(Debug, Clone, Bpaf)]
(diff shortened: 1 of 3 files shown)


tudorbarbu:
I'm also not quite content with this approach since it would be hard to maintain when more plugins are added

Another approach would be to have them all as options in enable plugin, and create a `--recommended-plugins` flag/configuration that sets the current default ones.

This would be cleaner in my opinion, but could lead to a breaking change now if we choose to disable plugins if `--recommended-plugins` is not set

Boshen:
We'll need to find another way :-(

PR #2428, opened 2024-02-18
feat(linter): eslint no-nonoctal-decimal-escape

tudorbarbu:
Rule [source](https://github.com/eslint/eslint/blob/main/lib/rules/no-nonoctal-decimal-escape.js)
Rule [tests](https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-nonoctal-decimal-escape.js)

I created a basic sticky regex implementation based on what I could find about it, since rust does not have support for that.
There might be a better place to keep it instead of this rule file


crates/oxc_linter/src/rules/eslint/no_nonoctal_decimal_escape.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+        (None, 0)
+    }
+}
+
+#[allow(clippy::cast_possible_truncation)]
+fn check_string(ctx: &LintContext<'_>, string: &str) {
+    lazy_static! {
+        static ref QUICK_TEST_REGEX: Regex = Regex::new(r"\\[89]").unwrap();
```
Boshen, line 97: I think we can do a quick scan with string methods instead of using a regex, which is quite heavy.

Boshen (approved):

Boshen (approved):

PR #2437, opened 2024-02-19
fix(semantic): Refactor jsdoc finding

leaysgur:
Partial fix for #168 

- [x] Fix general finding behavior for leading comments
- [x] Accept multiple jsdoc comments per node
- [x] Provide `get_one` and also `get_all` 
- [x] Add `iter_all()` for non-node related usage
- [x] Limit AST node kinds to parse


crates/oxc_semantic/src/builder.rs
```diff
@@ -203,10 +206,23 @@ impl<'a> SemanticBuilder<'a> {
...
+        let span = kind.span();
+
+        let mut leading_comments = vec![];
+        // Should handle trailing comments as well?
+        for (start, comment) in self.trivias.comments().range(..span.start) {
+            if !self.comments_before_seen.contains(start) {
+                leading_comments.push((start, comment));
+            }
+            self.comments_before_seen.insert(*start);
+        }
```
Boshen, line 218: I think these logic belong to the jsdoc builder, having implementation details here not clean.

crates/oxc_semantic/src/builder.rs
```diff
@@ -203,10 +206,23 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     fn create_ast_node(&mut self, kind: AstKind<'a>) {
+        let span = kind.span();
+
+        let mut leading_comments = vec![];
+        // Should handle trailing comments as well?
+        for (start, comment) in self.trivias.comments().range(..span.start) {
```
Boshen, line 213: This will hurt performance I think, because it's run on every ast node :-(

Boshen:
Thank you so much for trying to implement one of the hardest part of this project ❤️ 

You are free to rewrite everything if you think it's necessary. What I wrote is just a prototype to initiate eslint-plugin-jsdoc.

You have write permission, you may try graphite with stacked PRs.

crates/oxc_semantic/src/builder.rs
leaysgur: The reason why I keep this here was,
- there are various usecases for comments
  - such as `@ts-ignore`, `eslint-ignore` and `prettier-ignore`, etc..
- but finding can be done at once in advance

How do you think?

crates/oxc_semantic/src/builder.rs
leaysgur: Is it OK if I limit the target kinds like original code does? (although need to find out what to target from TS codebase ⛰️  )

Boshen: Yes it's ok

crates/oxc_semantic/src/builder.rs
Boshen: As long as things are well encapsulated, it's going to be hard to navigate the code since this file is huge with so many builders ...

leaysgur: OK, I'll move these to jsdoc builder.

(BTW, I wish the AST included comments... 😴 )

leaysgur:
> one of the hardest part of this project

This is so true! 😂 The more I looked into it, the more I got confused...

Although there are many uses for JSDoc comments and also implementations(I learned JSDoc TS, jsdoc.app, eslint-plugin-jsdoc are all differently implemented 🫠), if we set primary goal to implement `eslint-plugin-jsdoc`, I'm going to try to identify again what kind of API are needed.

Boshen:
Yeah the first goal is to get `eslint-plugin-jsdoc` tests passing. `jsdoc` doesn't have a formal spec so it's a bit hard to figure out which implementation to go for.

By looking at the dependencies of `eslint-plugin-jsdoc`, I think we can split the task into:

* jsdoc comment parsing - https://www.npmjs.com/package/comment-parser
* find the rules of jsdoc comment attachment

leaysgur:
Thank you so much~! I'm just now deciphering that code base... 🕵🏻‍♂️ 
(Sorry the PR is still half done.)

Boshen:
Feels like I can merge this as is after looking at the code.

leaysgur:
There are a few things I want to fix here.

- Decide how to handle `()`: ParenthesizedExpression
  - `jsdoccomment` does not attach JSDoc to this, but TS does(and this is intuitive tho)
- Return the nearest comment only or split API for this
  - Still keeping them in array
- Note about `min, maxLines` settings does not supported
  - `jsdoccomment` and `eslint-plugin-jsdoc` has options for this
  - Blank lines are configurable between a node and its attached JSDoc
    - e.g. if 2 lines are open, do not attach.
    - e.g. Ignore it if on the same line
    - (I'm not sure what this for...?)
- Correct all non-attached JSDoc, in particular, those located at the EOF and not attached to any node

(Of course, feel free to merge first. I'll send a PR for this later.)

Boshen:
Thank you so much for working on this, we'll get there eventually 😁 

It seems like we're a little bit closer to https://github.com/import-js/eslint-plugin-import/blob/v2.29.1/docs/rules/no-deprecated.md

Let me setup the code for it.

leaysgur:
Oh, JSDoc everywhere! 😮 

Now I can understand this kind of things easily...

- `eslint.sourceCode#getCommentsBefore()`
  - https://github.com/import-js/eslint-plugin-import/blob/ee5fadeffff68f2300bed7f67a310496cb969d61/src/ExportMap.js#L224
- `doctorine.parse()`
  - https://github.com/import-js/eslint-plugin-import/blob/ee5fadeffff68f2300bed7f67a310496cb969d61/src/ExportMap.js#L263

Wish you luck. 🚀

PR #2439, opened 2024-02-19
perf(parser): `byte_search` macro always unroll main loop

overlookmotel:
Refactor `byte_search!` macro to move logic out of the main loop. This ensures the compiler unrolls the loop.

This speeds up lexing single-line comments by 20%-25% on the benchmarks which contain enough comments for the change to register. Presumably the loop wasn't unrolled previously.

The code required to do this is a little odd. It adds an extra `loop {}` which always exits on the first turn (so not really a useful loop), but is required to be able to use `break` to exit that "loop", making 2 different paths for (1) matching byte found and (2) `for` loop completed without finding any match.

This is only way I could find to produce this behavior without using a macro. Is there a more "normal" way to get the same logic?

### DIFF
--- a/crates/oxc_parser/src/lexer/search.rs
+++ b/crates/oxc_parser/src/lexer/search.rs
@@ -495,45 +495,56 @@ macro_rules! byte_search {
 
         let mut $pos = $start;
         #[allow(unused_unsafe)] // Silence warnings if macro called in unsafe code
-        loop {
+        'outer: loop {
+            #[allow(clippy::redundant_else)]
             if $pos.addr() <= $lexer.source.end_for_batch_search_addr() {
                 // Search a batch of `SEARCH_BATCH_SIZE` bytes.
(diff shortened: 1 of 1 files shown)


overlookmotel:
* `main`
  * **PR #2439** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2439?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2439?utm_source=stack-comment).

Boshen:
### Merge activity

* **Feb 19, 9:39 PM EST**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2439).

PR #2457, opened 2024-02-20
AST transfer WIP

overlookmotel:
Very rough POC for #2409. Nowhere near ready to merge.

Mostly works. JS AST for `checker.ts` via "raw" transfer matches AST produced via current JSON API exactly. ~~Some errors in AST for `pdf.mjs` and `antd.js` due to some of reasons listed below.~~

Edit 8/3/24: Output for `pdf.mjs` and `antd.js` is now correct.

Current deficiencies:

* ~~Incorrect output for files containing any non-ASCII characters.~~
* ~~Does not handle strings/identifiers/regexps containing `\` escapes.~~
* `Statement` and a few other types increase in size by 8 bytes (likely the cause of the performance regression seen in benchmarks).
* ~~Requires size of AST to be known in advance.~~
* Hacky method of creating allocations.
* ~~Memory handling may be unsound!~~


overlookmotel:
* **#2457** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2457?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2457?utm_source=stack-comment).

andymac4182:
Does this mean we could use oxc_parser for eslint as well as prettier?

overlookmotel:
Maybe... Just to say this is very early stage work, the concept is not proven, and there is a long way to go yet. Some significant things would need to change within OXC to convert this from an experiment to a solid, dependable reality.

Personally, I do believe there is potential in this approach, but I think it's premature at this stage to talk a great deal about applications.

Sorry if that sounds negative. I'm just trying to give a realistic answer.

andymac4182:
That's okay 😀 that's a perfectly reasonably answer

overlookmotel:
I don't believe the Codspeed benchmarks are accurate. Running benchmarks locally, I'm seeing more like x5 speed-up. Will investigate.

overlookmotel:
Yes, benchmarks on CodSpeed are completely wrong. Have raised an issue https://github.com/CodSpeedHQ/action/issues/96

overlookmotel:
Benchmarks are now correct (though with a lot of variance in them).

overlookmotel:
Closing as I'm working on a new version.

Issue #2038, opened 2024-01-15
Finish codegen


Boshen:
We are running the test262 conformance suite in Node.js.

It produced a snapshot https://github.com/oxc-project/oxc/blob/main/tasks/coverage/codegen_runtime_test262.snap

We need to fix all the errors :-)

Boshen:
We also need to align implementation with esbuild, since that's what we have been doing.

PR #2458, opened 2024-02-20
fix(Codegen): Improvement to statement output.

andymac4182:
Skip loop iterations where TS syntax would of been output.

Working towards https://github.com/oxc-project/oxc/issues/2038


andymac4182:
Replaced by https://github.com/oxc-project/oxc/pull/2460

PR #2474, opened 2024-02-23
fix(linter): Handle cases where createElement is an Identifier in is_create_element_call

keita-hino:
I fixed a false negative issue in the `is_create_element_call` function where createElement was an Identifier.
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/utils/react.rs#L13-L19

In the case of [eslint-plugin-react's button-has-type rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md), the following triggers the expected errors:
```ts
// error
React.createElement("button")
// error
createElement("button")
```

However, with Oxlint, it behaves differently:
```ts
// error
React.createElement("button")
// false negative
createElement("button")
```


Boshen:
Thank you!

PR #2477, opened 2024-02-23
feat(linter): eslint-plugin-react void-dom-elements-no-children

keita-hino:
partof: https://github.com/oxc-project/oxc/issues/1022

docs: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/void-dom-elements-no-children.md
code: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/void-dom-elements-no-children.js
test: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/void-dom-elements-no-children.js


Boshen (approved):

Boshen:
Thank you!

PR #2497, opened 2024-02-25
[BREAKING CHANGE] Change `Atom` to `Atom<'a>` to make it safe

Boshen:
Part of #2295

This PR splits the `Atom` type into `Atom<'a>` and `CompactString`.

All the AST node strings now use `Atom<'a>` instead of `Atom` to signify it belongs to the arena.

It is now up to the user to select which form of the string to use.

This PR essentially removes this really unsafe code 

https://github.com/oxc-project/oxc/blob/93742f89e91bc7c45491b6d2b49a134fa65f2b5c/crates/oxc_span/src/atom.rs#L98-L107

which can lead to 
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,6 +199,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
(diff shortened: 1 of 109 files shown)


Boshen:
@overlookmotel Here we go ... I still remember the week it took me to understand lifetimes and adding them to the AST, which is 2 years ago. It should be faster this time ... just follow compiler errors :-)

Issue #2463, opened 2024-02-21
feat(parser): serialize to estree


Boshen:
We should serialize our AST to estree, to create less chaos in the JavaScript ecosystem.

We need to:

- * align the smaller parts to estree by modifying the AST structs directly
- * write serde serializers for the bigger ones
- * Have an AST pass that rewrites all Unicode-8 span positions to Unicode 16, depending on https://github.com/oxc-project/oxc/issues/959

overlookmotel:
@ArnaudBarre I see from your Github repos that you mostly code in JS/TS, so I don't know if you're comfortable writing Rust. But if you are, would you consider assisting with this?

What I'm thinking is that rather than building out further your OXC prettier parser layer in JS, could you instead submit the changes upstream in OXC?

For the mismatches where all that's needed is changing field names, it's as simple as e.g.:

```diff
pub struct ComputedMemberExpression<'a> {
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub span: Span,
     pub object: Expression<'a>,
+    #[cfg_attr(feature = "serde", serde(rename = "property"))]
     pub expression: Expression<'a>,
     pub optional: bool, // for optional chaining
 }
```

Where difference is snake_case vs camelCase:

```diff
#[derive(Debug, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize), serde(tag = "type"))]
+#[cfg_attr(feature = "serde", derive(Serialize), serde(tag = "type", rename_all = "camelCase"))]
#[cfg_attr(all(feature = "serde", feature = "wasm"), derive(tsify::Tsify))]
pub struct NewExpression<'a> {
    #[cfg_attr(feature = "serde", serde(flatten))]
    pub span: Span,
    pub callee: Expression<'a>,
    pub arguments: Vec<'a, Argument<'a>>,
    pub type_parameters: Option<Box<'a, TSTypeParameterInstantiation<'a>>>,
}
```

To cover all the `node.importKind = node.importKind.type` transforms, I think all we need is:

```diff
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize), serde(tag = "type", rename_all = "camelCase"))]
+#[cfg_attr(feature = "serde", derive(Serialize), serde(rename_all = "camelCase"))]
#[cfg_attr(all(feature = "serde", feature = "wasm"), derive(tsify::Tsify))]
pub enum ImportOrExportKind {
[...]

PR #2506, opened 2024-02-26
Fix few serialization issues

ArnaudBarre:
Re: #2463


crates/oxc_ast/src/ast/ts.rs
```diff
@@ -41,6 +41,7 @@ pub struct TSEnumDeclaration<'a> {
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub span: Span,
     pub id: BindingIdentifier,
+    #[cfg_attr(feature = "serde", serde(flatten))]
```
Boshen, line 44: We may be able to inline this, let me take a look.

Boshen: I'll make another PR.

Boshen (approved):

PR #2508, opened 2024-02-26
fix(semantic): add export symbol flag to identifiers in export declarations

Dunqing:
Related PR: #2335

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -6,7 +6,10 @@ use std::{cell::RefCell, path::PathBuf, rc::Rc, sync::Arc};
 use oxc_ast::{ast::*, AstKind, Trivias, TriviasMap, Visit};
 use oxc_diagnostics::Error;
 use oxc_span::{Atom, SourceType, Span};
-use oxc_syntax::{module_record::ModuleRecord, operator::AssignmentOperator};
+use oxc_syntax::{
+    module_record::{ExportLocalName, ModuleRecord},
+    operator::AssignmentOperator,
+};
 
 use crate::{
     binder::Binder,
@@ -337,6 +340,22 @@ impl<'a> SemanticBuilder<'a> {
 
     pub fn add_redeclared_variables(&mut self, variable: VariableInfo) {
         self.redeclare_variables.variables.push(variable);
+    }
+
+    fn add_export_flag_for_export_identifier(&mut self) {
+        self.module_record.local_export_entries.iter().for_each(|entry| match &entry.local_name {
+            ExportLocalName::Name(name_span) => {
+                if let Some(symbol_id) = self.scope.get_root_binding(name_span.name()) {
+                    self.symbols.union_flag(symbol_id, SymbolFlags::Export);
+                }
+            }
+            ExportLocalName::Default(span) => {
+                if let Some(symbol_id) = self.symbols.get_symbol_id_from_span(span) {
+                    self.symbols.union_flag(symbol_id, SymbolFlags::Export);
+                }
+            }
+            ExportLocalName::Null => {}
+        });
     }
 }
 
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#2508** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2508?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2508?utm_source=stack-comment).

Dunqing:
@DonIsaac Thank you for your previous work! This is a problem I realized a long time ago but I don't know how to fix it, thanks for enlightening me

Boshen:
Do we have a specific use case for this? I want to understand what we are trying to achieve.

Dunqing:
> Do we have a specific use case for this? I want to understand what we are trying to achieve.

In this case, we cannot tell whether a is referenced or not, but if a's flag contains `SymbolFlags::Export`, we can tell.

https://oxc-project.github.io/oxc/playground/?
code=3YCAAICggICAgICAgICxG0qZRraXD3k8CFklCNW9ja59iHxNE9O%2F0tzgoLr6Nw5LHCC%2Ff3d%2BoIA%3D

We also need it in the transformer typescript plugin

Boshen:
I think we should reuse the data from `ModuleRecord`, and do the bind during the last stage of `visit_program`.

I'm always against adding more state to `SemanticBuild`, there should always be a single source of truth.

crates/oxc_semantic/src/module_record/builder.rs
```diff
@@ -241,17 +241,27 @@ impl ModuleRecordBuilder {
...
-            ExportDefaultDeclarationKind::Expression(_) => None,
-            ExportDefaultDeclarationKind::FunctionDeclaration(func) => func.id.as_ref(),
-            ExportDefaultDeclarationKind::ClassDeclaration(class) => class.id.as_ref(),
+            ExportDefaultDeclarationKind::Expression(expr) => {
+                if let Expression::Identifier(ident) = &expr {
+                    Some((ident.name.clone(), ident.span))
+                } else {
+                    None
+                }
+            }
```
Dunqing, line 250: To solve this.
```ts
const a = 1;
export default a;
```

Boshen: We need to take a look at the spec on this one. The algorithms of this module strictly follow the spec.

Dunqing: Ok, I will take a look

magic-akari: What is the purpose of this change?
It should be `default` exported instead of `a`.

Dunqing: > What is the purpose of this change? It should be `default` exported instead of `a`.

We need to add `SymbolFlags::Export` to all exports if the export is `Identifier` or `IdentifierReference `. If we want to solve this problem via `ModuleRecord` we must get all ExportEntries, But if the export entry is default we cannot get the `a` symbol currently, Because `ExportLocalName::Default` only have a span.

Dunqing: I will revert these changes. I found the best way. I can get the symbol id via `symbols.get_symbol_id_from_span`

magic-akari: The three exports below differ in their behavior: only 'a' and 'b' are live bindings exports, whereas 'c' is a value export.

```javascript
export function a() {}
export class b {}
export default c;
```

So, it is useful to track references to both `a` and `b`.
However, I don't see any *value* to track `c`

Dunqing: I've already mentioned why we need to add SymbolFlags::Export here. https://github.com/oxc-project/oxc/pull/2508#issuecomment-1963253987

Dunqing:
Looks like the benchmark regression is expected, `cal.com` has a lot of `exports`. But I'm not sure if it's worth it, normally there wouldn't be so many exports.

PR #2512, opened 2024-02-26
fix(linter): Correct configuration file parsing for jsx-no-useless-fragment

### LINKED ISSUE #2484: bug/feature request with jsx-no-useless-fragment
Hi, thanks for making oxlint. I'm using it for the first time in my personal coding project. I think I've discovered something that might be a bug. (Forgive me if I'm reporting this in the wrong way. Happy to adjust anything I'm doing here.)

Using oxlint 0.2.10

I'm getting a warning from the rule `eslint-plugin-react(jsx-no-useless-fragment)` which don't think I should be getting. The code in question is:

`{num && <>{num}</>}`

When I run eslint, I don't get a warning on that line. I don't have any specific configuration for that line in my eslintrc other than `"plugin:react/recommended", "plugin:react/jsx-runtime"`. I've checked out the documentation on this rule and it seems there is an additional rule option [allowExpressions](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md#allowexpressions) that could help here.

I've added this rule specifically to my eslintrc file:
* `"react/jsx-no-useless-fragment": ["warn", {"allowExpressions": false}]` eslint produces the warning
* `"react/jsx-no-useless-fragment": ["warn", {"allowExpressions": true}]` eslint does not produce the warning

Two issues:
* I get these eslint warnings even though I also have [eslint-plugin-oxlint](https://github.com/oxc-project/eslint-plugin-oxlint) configured
* I'm calling oxlint `oxlint -c .eslintrc.json` and oxlint produces the warning with either of those configurations in my eslintrc file so I don't know how to make that warning go away.

Is this intended behavior or something that should/could be adjusted in oxlint?

Thanks for your great work!

keita-hino:
fix: #2484 

I fixed the configuration parsing for jsx-no-useless-fragment.

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_no_useless_fragment.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_no_useless_fragment.rs
@@ -56,10 +56,13 @@ declare_oxc_lint!(
 
 impl Rule for JsxNoUselessFragment {
     fn from_configuration(value: serde_json::Value) -> Self {
-        let allow_expressions =
-            value.get("allowExpressions").and_then(serde_json::Value::as_bool).unwrap_or(false);
+        let value = value.as_array().and_then(|arr| arr.first()).and_then(|val| val.as_object());
 
-        Self { allow_expressions }
+        Self {
+            allow_expressions: value
+                .and_then(|val| val.get("allowExpressions").and_then(serde_json::Value::as_bool))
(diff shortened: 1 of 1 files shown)


Boshen (approved):

Boshen:
Thank you!

PR #2515, opened 2024-02-26
ci: use codspeed v2.4.0 for benchmark

Boshen:


Boshen:
This breaks CI cache :-(

PR #2520, opened 2024-02-26
Support file type in playground

ArnaudBarre:
Screenshot 2024-02-27 at 00 16 21

Not sure about the bindgen thing.

I made the filename optional but given the impact on parsing I think it could be great for external users to not depend on the tsx default in the future


Boshen:
Thank you!

PR #2524, opened 2024-02-27
feat(codegen): add sourcemap

underfin:
Add sourcemap support for codegen, the code port from esbuild, it will using ast position to search origin code and find line/column, and also generate  line/column of output code, it is sourcemap `token`. The `token` will added into `sourcemap_builder` using rust `sourcemap` crate.

The test come from babel codgen sourcemap test,it using snapshot test. Because the implement is different, the mapping has different, so here snapshot sourcemap at test.


tasks/coverage/codegen_sourcemap.snap
```diff
@@ -0,0 +1,57 @@
+- ClassBody-StaticBlock/input.js
+{"version":3,"sources":[""],"sourcesContent":["class Foo {\n  static {}\n}\n\nclass A1 {\n  static{\n    foo;\n  }\n}\n\nclass A2 {\n  static {\n    foo;bar;\n  }\n}\n"],"names":["Foo","A1","A2"],"mappings":"AAAA,MAAMA,IAAI;CACR;CAAS;AACV;AAED,MAAMC,GAAG;CACP;AACE;CACD;AACF;AAED,MAAMC,GAAG;CACP;AACE;AAAI;CACL;AACF"}
+https://evanw.github.io/source-map-visualization/#OTcAY2xhc3MgRm9vIHsKCXN0YXRpY3sKCX0KfQpjbGFzcyBBMSB7CglzdGF0aWN7CgkJZm9vOwoJfQp9CmNsYXNzIEEyIHsKCXN0YXRpY3sKCQlmb287CgkJYmFyOwoJfQp9CjMxOAB7InZlcnNpb24iOjMsInNvdXJjZXMiOlsiIl0sInNvdXJjZXNDb250ZW50IjpbImNsYXNzIEZvbyB7XG4gIHN0YXRpYyB7fVxufVxuXG5jbGFzcyBBMSB7XG4gIHN0YXRpY3tcbiAgICBmb287XG4gIH1cbn1cblxuY2xhc3MgQTIge1xuICBzdGF0aWMge1xuICAgIGZvbztiYXI7XG4gIH1cbn1cbiJdLCJuYW1lcyI6WyJGb28iLCJBMSIsIkEyIl0sIm1hcHBpbmdzIjoiQUFBQSxNQUFNQSxJQUFJO0NBQ1I7Q0FBUztBQUNWO0FBRUQsTUFBTUMsR0FBRztDQUNQO0FBQ0U7Q0FDRDtBQUNGO0FBRUQsTUFBTUMsR0FBRztDQUNQO0FBQ0U7QUFBSTtDQUNMO0FBQ0YifQo=
```
underfin, line 3: Here snapshot sourcemap visualizer url to easy review result.

crates/oxc_codegen/src/lib.rs
```diff
@@ -61,6 +69,16 @@ pub struct Codegen<const MINIFY: bool> {
 
     /// Track the current indentation level
     indentation: u8,
+
+    // sourcemap
+    enable_sourcemap: bool,
```
Boshen, line 74: It would be nice to have a `SourceMapBuilder` for encapsulating all these states and methods.

crates/oxc_codegen/src/lib.rs
```diff
@@ -105,6 +131,22 @@ impl<const MINIFY: bool> Codegen<MINIFY> {
         unsafe { String::from_utf8_unchecked(self.code) }
     }
 
+    pub fn build_with_sourcemap(
```
Boshen, line 134: We don't need a separate method.

I think the API should be a builder pattern `fn with_sourcemap(mut self, yes: bool) {}` with a `sourcemap: bool` in the codegen struct.

crates/oxc_codegen/src/lib.rs
```diff
@@ -377,6 +428,154 @@ impl<const MINIFY: bool> Codegen<MINIFY> {
...
+            }
+        }
+
+        tables.push(LineOffsetTable {
+            columns,
+            byte_offset_to_first,
+            byte_offset_to_start_of_line: line_byte_offset,
+        });
+
+        // dbg!(&tables);
```
Boshen, line 575:
```suggestion
```

tasks/coverage/src/sourcemap.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        let codegen_options = oxc_codegen::CodegenOptions::default();
+        let (content, map) = oxc_codegen::Codegen::<false>::new(source_text.len(), codegen_options)
+            .build_with_sourcemap(&ret.program, source_text, "");
+        let mut buff = vec![];
+        map.to_writer(&mut buff).unwrap();
+
+        let mut result = String::from_utf8(buff).unwrap();
+        result.push('\n');
+        result.push_str(Self::create_visualizer_url(&content, &result).as_str());
```
Boshen, line 132: I think we should discuss how to test this, this is just printing things out, there is no validation yet :-(

crates/oxc_codegen/src/lib.rs
```diff
@@ -377,6 +428,154 @@ impl<const MINIFY: bool> Codegen<MINIFY> {
...
+                    } else {
+                        self.generated_column += 2;
+                    }
+                }
+            }
+        }
+        self.last_generated_update = self.code.len();
+    }
+
+    fn generate_line_offset_tables(content: &str) -> Vec<LineOffsetTable> {
```
Boshen, line 503: I think we can optimize this with `memchr`, but we'll need some benchmark setup before this.

Boshen:
Let's figure out the testing strategy and benchmark setup before merging. Functionality wise this is mergable.

tasks/coverage/src/sourcemap.rs
```diff
@@ -0,0 +1,221 @@
...
+pub struct SourcemapSuite<T: Case> {
+    test_root: PathBuf,
+    test_cases: Vec<T>,
+}
+
+impl<T: Case> SourcemapSuite<T> {
+    pub fn new() -> Self {
+        Self {
+            test_root: project_root().join(FIXTURES_PATH),
+            test_cases: TestFiles::complicated()
```
underfin, line 22: Here the big file the snapshot is very big, maybe we should remove it and add some small test.

Boshen: Agree. Is "react" small enough? You can filter it from tasks/common/src/test_file.rs

tasks/benchmark/benches/codegen_sourcemap.rs
Boshen, line 1: I'll need to add relevant code to CI and Cargo.toml

tasks/coverage/src/lib.rs
```diff
@@ -64,6 +66,7 @@ impl AppArgs {
         BabelSuite::<CodegenBabelCase>::new().run("codegen_babel", self);
         TypeScriptSuite::<CodegenTypeScriptCase>::new().run("codegen_typescript", self);
         MiscSuite::<CodegenMiscCase>::new().run("codegen_misc", self);
+        SourcemapSuite::<SourcemapCase>::new().run("codegen_sourcemap", self);
```
Boshen, line 69: We are not running and test suites, let me move this to somewhere else.

tasks/coverage/src/sourcemap.rs
underfin: Updated.

tasks/coverage/src/sourcemap.rs
```diff
@@ -0,0 +1,222 @@
...
+                    line_byte_offset = i;
+                }
+                _ => {}
+            }
+        }
+        tables.push(content[line_byte_offset..].encode_utf16().collect::<Vec<_>>());
+        tables
+    }
+
+    fn create_visualizer_text(
```
Boshen, line 109: Nice, it seems the only thing left is to improve the snapshot format.

Maybe we can move this to the codegen crate so we can re-use this in the example and playground as well?

underfin: Not sure for this, it is a debugger format, the `https://evanw.github.io/source-map-visualization` is better to review, maybe the playground can link it if enable sourcemap.

Boshen: Oh ok, I thought we're going to copy the babel visualizer 😅

Boshen:
PR is broken ... I can't update this PR, I'll create another one https://github.com/oxc-project/oxc/pull/2565

PR #2528, opened 2024-02-27
perf(parser): lex JSX strings with `memchr`

overlookmotel:
Simplify lexing JSX string attributes. As the search is purely for 1 byte value (the closing quote), and so doesn't require a byte table, use `memchr`.

This change doesn't really register on benchmarks, but it's one step closer to removing `AutoCow`, and transitioning all the searches in the lexer to byte-by-byte.

### DIFF
--- a/crates/oxc_parser/src/lexer/jsx.rs
+++ b/crates/oxc_parser/src/lexer/jsx.rs
@@ -1,6 +1,7 @@
-use super::{AutoCow, Kind, Lexer, Token};
+use super::{Kind, Lexer, Token};
 use crate::diagnostics;
 
+use memchr::memchr;
 use oxc_syntax::identifier::{is_identifier_part, is_identifier_start};
 
 impl<'a> Lexer<'a> {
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#2528** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2528?utm_source=stack-comment).

overlookmotel:
Question: Is it legal for JSX string attributes to contain line breaks? The current implementation allows it, and this PR retains that behavior. But should this be allowed?

```jsx
<Foo x="
" />
```

ArnaudBarre:
> Is it legal for JSX string attributes to contain line breaks? 

I didn't check the spec, but I was curious so I dropped 
```tsx
<div x="
" />
```
into my source code and Prettier formatted it, TS was fine, Websotrm didn't have issue for syntax highlighting, SWC in dev was fine and esbuild for build too. So I guess this is a thing (even if it looks horrible).

Boshen:
https://github.com/oxc-project/oxc/blob/f7601080948bfeb07869d3d6b4f379e0c760809e/crates/oxc_parser/src/lexer/jsx.rs#L7-L16

https://facebook.github.io/jsx/#prod-JSXDoubleStringCharacters

overlookmotel:
Thanks both for clarifying that. Weird, but OK...

PR #2531, opened 2024-02-28
refactor(semantic/jsdoc): Misc fixes for JSDoc related things

leaysgur:
Sorry for the rather large size of PR. 😓 But essentially, not changed so much.

#### 1. Reorganize directories and namings

```
src/jsdoc
├── builder.rs 👈🏻 for SemanticBuilder
├── finder.rs  👈🏻 `semantic.jsdoc()`
├── mod.rs
└── parser
   ├── jsdoc.rs     👈🏻 `JSDoc` struct which has `comment` and `tags`
   ├── jsdoc_tag.rs 👈🏻 `JSDocTag` struct
   ├── mod.rs
   ├── parse.rs     👈🏻 parsing logic by `JSDocParser`
   └── utils.rs
```

Now `mod.rs` has only export things.

#### 2. Introduce `JSDocTagKind::Unknown(name)`

We need to keep their name as-is to check valid tag names are used.(e.g. `jsdoc/check-tag-names` rule)

#### 3. Support multiline description

- Comment for JSDoc
- Comment for each JSDocTag

```js
/**
 * @foo this comment continues
 * here but leading * should be ignored!!
 */
```

- - -

Please correct me if I am doing something wrong... 🐰


Boshen:
Thank you! I'm just going to trust you on this one :-)

leaysgur:
Oh, I'll do my best! 😲 (It would be perfect if I wasn't a Rust beginner...)

PR #2538, opened 2024-02-28
feat(linter/import): partial support namespace check

Dunqing:


Dunqing:
* **#2538** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2538?utm_source=stack-comment).

crates/oxc_linter/src/rules/import/namespace.rs
```diff
@@ -24,7 +26,41 @@ declare_oxc_lint!(
 );
 
 impl Rule for Namespace {
-    fn run_once(&self, _ctx: &LintContext<'_>) {}
+    fn run(&self, node: &AstNode<'_>, ctx: &LintContext<'_>) {
```
Boshen, line 29: Feels like we can test on each `MemberExpression` and see if the lhs identifier reference has the `import` flag?

Dunqing: No, Because It's not enough to have an `import` flag, we need to check if it's an `ImportNamespaceSpecifier`.

crates/oxc_linter/src/rules/import/namespace.rs
```diff
@@ -1,30 +1,104 @@
...
+    /// Also, will report for computed references (i.e. foo["bar"]()).
+    /// Reports on assignment to a member of an imported namespace.
     Namespace,
     nursery
 );
 
 impl Rule for Namespace {
-    fn run_once(&self, _ctx: &LintContext<'_>) {}
+    fn run(&self, node: &AstNode<'_>, ctx: &LintContext<'_>) {
+        if let AstKind::ImportNamespaceSpecifier(specifier) = node.kind() {
```
Boshen, line 41: Can we remove a level of indentation level by using `if let else? I prefer code with less indentations and less big `if` blocks.

crates/oxc_linter/src/rules/import/namespace.rs
```diff
@@ -1,30 +1,104 @@
...
+            };
+
+            let source = &decl.source.value;
+            let module_record = ctx.semantic().module_record();
+            let Some(module) = module_record.loaded_modules.get(source.as_ref()) else {
+                return;
+            };
+
+            let check_binding_exported = |name: &str, span| {
+                if module.exported_bindings.get(name).is_none() {
```
Boshen, line 53: I added a `ModuleRecord::no_esm` yesterday, we need to skip all commonjs modules.

Boshen: The early skip can be done after the check

```
            let Some(module) = module_record.loaded_modules.get(source.as_ref()) else {
                return;
            };
```

crates/oxc_linter/src/rules/import/namespace.rs
```diff
@@ -1,30 +1,104 @@
...
+                                    ctx.nodes().parent_kind(node.id()),
+                                    Some(AstKind::SimpleAssignmentTarget(_))
+                                ) {
+                                    ctx.diagnostic(NamespaceDiagnostic::Assignment(
+                                        member.span(),
+                                        name.clone(),
+                                    ));
+                                };
+
+                                if member.is_computed() {
```
Boshen, line 79: This is behind a configuration https://github.com/import-js/eslint-plugin-import/blob/ee5fadeffff68f2300bed7f67a310496cb969d61/src/rules/namespace.js#L127

the default is false, can you add a TODO somewhere?

crates/oxc_linter/src/rules/import/namespace.rs
Boshen, line 41: Would it be faster / easier to use `run_once` on `ModuleRecord::import_entries` with `ImportImportName::NamespaceObject`?

Dunqing: It will, so we don't need to find the parent kind

Boshen (approved):

PR #2543, opened 2024-02-29
Fix getter return rule

BlackSoulHub:


Boshen:
I think the better choice is to not check for TypeScript, to align with typescript-eslint 

https://github.com/typescript-eslint/typescript-eslint/blob/6954a4a463f9a2a1c20e41c91ee2c75c953dcc9f/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts#L24

BlackSoulHub:
> I think the better choice is to not check for TypeScript, to align with typescript-eslint
> 
> https://github.com/typescript-eslint/typescript-eslint/blob/6954a4a463f9a2a1c20e41c91ee2c75c953dcc9f/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts#L24

Something like that?:

```rust
if ctx.source_type().is_typescript() {
    return (DefinitelyReturnsOrThrowsOrUnreachable::Yes, false);
}
```

Boshen:
> > I think the better choice is to not check for TypeScript, to align with typescript-eslint
> > https://github.com/typescript-eslint/typescript-eslint/blob/6954a4a463f9a2a1c20e41c91ee2c75c953dcc9f/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts#L24
> 
> Something like that?:
> 
> ```rust
> if ctx.source_type().is_typescript() {
>     return (DefinitelyReturnsOrThrowsOrUnreachable::Yes, false);
> }
> ```

I'll do the whole batch from the list.

And since you're comfortable with code, you may try https://eslint.org/docs/latest/rules/no-unreachable which is the last unimplemented cfg rule.

Boshen:
Let's improve on this instead of skipping TypeScript.

BlackSoulHub:
I thought about what would happen to functions that return `void`? I check this later

BlackSoulHub:
> And since you're comfortable with code, you may try https://eslint.org/docs/latest/rules/no-unreachable which is the last unimplemented cfg rule.

I'll try to do it. But what about this PR?

Boshen:
> > And since you're comfortable with code, you may try https://eslint.org/docs/latest/rules/no-unreachable which is the last unimplemented cfg rule.
> 
> I'll try to do it. But what about this PR?

I'll get this merged once all tests pass.

Boshen (approved):

PR #2560, opened 2024-03-01
fix(ast): Update `is_statement()` and `is_declaration()` lists

leaysgur:
- Remove `VariableDeclaration` from `is_statement()`
  - Is this intended?(Sorry to bother you if so.)
- Add more declarations to `is_declaration()`


leaysgur:
I see. 👀 

The reason I opened this PR was to make this check more closer to TS's.

https://github.com/oxc-project/oxc/blob/18cff6aab85ff25bb808e06bb29bbf6facd12307/crates/oxc_semantic/src/jsdoc/builder.rs#L75-L83

So I will create another utils for this purpose using each `AstKind::` directly.

PR #2565, opened 2024-03-01
feat(codegen): add sourcemap

Boshen:


Boshen:
@underfin I'll clean up the APIs and infrastructure. You may continue working on this, if there are outstanding problems to fix.

PR #2566, opened 2024-03-02
fix(ast): rename serialized fields to camel case

overlookmotel:
Fixes a few more AST fields names which are currently snake case in serialized JSON.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -383,7 +383,7 @@ pub struct ThisExpression {
 
 /// <https://tc39.es/ecma262/#prod-ArrayLiteral>
 #[derive(Debug, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize), serde(tag = "type"))]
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#2566** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@ArnaudBarre Was there a reason why you didn't do these already in #2522?

ArnaudBarre:
I didn't include this one because I think tailing comma part of ESTree and I though about discussing skip vs camelCase later

overlookmotel:
Ah I see. Well maybe they should be skipped. But regardless of whether they're in ESTree or not, I think it's better all properties in JSON AST are camelCase - snake_case would be unexpected to JS users.

ArnaudBarre:
Yeah this could be added to all nodes for potential future field if the cost is 0, if there is a small cost I think this is fine adding it when needed

overlookmotel:
I don't think there's any runtime cost to field renaming, as the serializers are built at compile-time.

PR #2574, opened 2024-03-02
fix(parser): parse empty method declaration as TSEmptyBodyFunctionExpression

ArnaudBarre:
That a tricky one, because it's time to decide what does ESTree compliant means in the TS world (re #2463)

This code:

```ts
export declare class ByteBuffer {
  clear(): void;
     // ^^
}
```

- Is parsed by [Babel](https://astexplorer.net/#/gist/9f43cf0fe91760e11f1572934681c92f/d38530204eaa6e12dbd14f3ef2d7d2fcb7da7bc2) as `FunctionExpression` with an empty body
- By [TS-ESLint](https://astexplorer.net/#/gist/626dd346956a02c221d4e128450652af/4ea4e2feb5ae7bb8787f8c7e452d442c935c1f09) as [TSEmptyBodyFunctionExpression](https://github.com/typescript-eslint/typescript-eslint/pull/1289)
- By [OXC](https://oxc-project.github.io/oxc/playground/?code=3YCAAIC1gICAgICAgICyHorESipoTXPdvBaE9wxzlOraoWs19SUxDvdcwSVU0kbBO2b7ppX3x2P5IhQlpGHOYEHNCEfLf38HUICA) as `TSDeclareFunction`

I'm going the easy way to fix this to the Babel way, but I think following TS-ESLint would make sense. There is an [open babel issue](https://github.com/babel/babel/issues/13878) about that. 

Edit: Ok that not so easy and require updating some logic.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1770,6 +1770,8 @@ pub enum FunctionType {
     FunctionDeclaration,
     FunctionExpression,
     TSDeclareFunction,
+    /// <https://github.com/typescript-eslint/typescript-eslint/pull/1289>
+    TSEmptyBodyFunctionExpression,
 }
 
 /// <https://tc39.es/ecma262/#prod-FormalParameters>
--- a/crates/oxc_parser/src/js/function.rs
+++ b/crates/oxc_parser/src/js/function.rs
@@ -105,16 +105,22 @@ impl<'a> ParserImpl<'a> {
             return Err(self.unexpected());
         }
 
-        let function_type = if body.is_none() {
-            FunctionType::TSDeclareFunction
-        } else {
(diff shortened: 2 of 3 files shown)


Boshen:
Given babel raised it as issue, I'll try and align it with ts-eslint.

Boshen (approved):

ArnaudBarre:
I didn't succeed to the task locally, but from the CI logs I see that this case:

```ts
var b = {
    foo(x = 1), // error
    foo(x = 1) { }, // error
}
```

was failing for a bad reason, now is parsing, but in fact should failed because with the error: `A parameter initializer is only allowed in a function or constructor implementation.(2371)`.

What's the current philosophy of OXC parser, should I try to add this logic or is this fine?

Boshen (approved):

PR #2578, opened 2024-03-03
fix(ast)!: align TSImportType with ESTree

ArnaudBarre:
Implements https://github.com/typescript-eslint/typescript-eslint/issues/2998

The copy of props feels wrong, but could not get it working otherwise with the box and borrow things 😅

Also I found that TSImportType was missing some entries for visitors and codegen.

In the case of codegen I'm not really understand the need as all the types seems to be dismissed?


Boshen:
The codegen also prints TypeScript AST now https://github.com/oxc-project/oxc/blob/d9cc429d4a23c55330dbee0f7d353a3195737156/crates/oxc_codegen/src/lib.rs#L39

PR #2581, opened 2024-03-03
feat(linter): eslint-plugin-jest: prefer-strict-equal

eryue0220:
Rule detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-strict-equal.ts)


crates/oxc_linter/src/rules/jest/prefer_strict_equal.rs
```diff
@@ -0,0 +1,91 @@
...
+        };
+        let Some(matcher) = parse_jest_expect_fn_call.matcher() else {
+            return;
+        };
+        let Some(matcher_name) = matcher.name() else {
+            return;
+        };
+
+        if matcher_name.eq("toEqual") {
+            ctx.diagnostic(UseToStrictEqual(matcher.span));
```
mysteryven, line 69: Why not add Fixer?👀

eryue0220: Done~ Because this rule also only provided suggestion.

mysteryven (approved):
Thank you!

PR #2626, opened 2024-03-06
Test perf impact of adding utf16 offset to span

Boshen:


Boshen:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2626?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2626** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2626?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2625** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2625?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Boshen:
@overlookmotel 
image

What about adding a utf16 offset to `Span`?

overlookmotel:
Thanks loads for testing this. That's not nearly as bad as I expected!

I guess we'd need also need a field for UTF-16 span end. Have pushed another commit to add that, so we can see the perf impact.

To implement it, we'd need either:

1. Feature flag to add extra fields to AST (as you say).
2. Parameterize `Lexer` and `ParserImpl` with a generic i.e. `Lexer<SpanType>` which does not alter AST, but changes what `Span::start` + `Span::end` contain (UTF-8 / UTF-16 offsets).

NB: for (2) the public API for `Parser` would be largely unchanged. Consumer would provide an option for span type they want and `Parser::parse` would then construct a `ParserImpl<SpanType::Utf8>` or `ParserImpl<SpanType::Utf16>` internally. `Parser` would not need to become `Parser<SpanType>`.

Personally, I'm not a fan of features, because then the consumer is limited to *either* one or the other. Option (2) allows them to parse one thing with UTF8 offsets, and then another with UTF16. But maybe (1) is simpler for a first iteration.

Boshen:
It feels like we need both values in the whole life cycle of a bundler:

* it reports syntax errors on the Rust to the terminal, requiring utf8 spans
* it produces source map, requiring utf16 spans
* it gets passed to js plugins, requiring utf16 spans

Shall we accept the 3 - 4% performance regression to begin with and get these utf16 spans out the door?

Boshen:
> Hmm. Can you give me a bit of time to think about this?

Sure. It is at least acceptable to add two `u32`s to a `Span` from this experiment.

overlookmotel:
If you have time to give thought to my questions on #959, that'd be really helpful. I feel like I'm lashing around in the dark on this. Like I *somewhat* understand what the problem is, but not *really*.

PR #2642, opened 2024-03-08
feat(linter): Implement jsdoc/check-access

leaysgur:
Part of #1170, Finally... 🗻 

Some preparation PRs may be needed in advance.

- [x] settings.jsdoc #2706 
- [x] new struct design #2765
- [x] handle `Span` for diagnostics #2815

Implement plugin itself.

- https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/check-access.md
- https://github.com/gajus/eslint-plugin-jsdoc/blob/main/src/rules/checkAccess.js

I'll send a PR to make this plugin public after confirming that a few more rules can be implemented without any problems.


leaysgur:
@Boshen These are some concerns noticed when I tried it. What do you think?

### Is it OK to use `run_once`?

I got a review in a previous [PR](https://github.com/oxc-project/oxc/pull/2410#discussion_r1490702321): If `run_once` is implemented, but `run` is still executed. 

Is this unavoidable?

### `struct JSDocTag`

Currently, there is a single struct `JSDocTag` and distinguished by its `.kind`.

Should there be a separate struct for each tag?
Then, for example, something like `has_valid_level()` could be implemented in `JSDocAccessTag`.

But it seems that the current string based implementation is fine.
(There is no specification of how users actually describe JSDoc. 🙄 )

Depends on how deeply Oxc wants to support JSDoc?

### Failed CI

(I'm not sure I can fix this...!)

This is the failed comments from antd.mjs.

```js
/**
 * flat tree data on expanded state
 *
 * @export
 * @template T
 * @param {*} data : table data
 * @param {string} childrenColumnName : 指定树形结构的列名
 * @param {Set<Key>} expandedKeys : 展开的行对应的keys
 * @param {GetRowKey<T>} getRowKey  : 获取当前rowKey的方法
 * @returns flattened data
 */
```

```js
/**
 * @param {Set<Key>} expandedKeys : 展开的行对应的keys
 * @param {GetRowKey<T>} getRowKey  : 获取当前rowKey的方法
 */
```

This also fails. Something wrong in `@param` parsing.

Boshen:
> Is it OK to use run_once?

This is the work flow:

https://github.com/oxc-project/oxc/blob/53a8e7ffbebb19cdd92bd2bcef6c308e0897bf05/crates/oxc_linter/src/lib.rs#L107-L129

We try to avoid visiting the AST twice.

> Should there be a separate struct for each tag?

Do this instead:

```rust
impl<'a> JSDocTag<'a> {
    fn kind(&self) -> JSDocTagKind<'a> {
        self.kind
    }
}

impl<'a> JSDocTagKind<'a> {
    fn is_deprecated(&self) -> bool {
        matches!(self, Self::Deprecated)
    }

    fn parameter(&self) -> Option<&Param<'a>> {
        if let Self::Parameter(param) = self {
            Some(param)
        } else {
            None
        }
    }
}
```

leaysgur:
Many of `plugin-jsdoc` rules will use `run_once()` only.

In these cases,

```rs
for node in semantic.nodes().iter() { 
   for (rule_name, rule) in &self.rules { 
       ctx.with_rule_name(rule_name); 
       rule.run(node, &ctx); // this run() fn is almost empty 👀 
   } 
} 
```

Does this mean negligibly cheep?

leaysgur:
Write to dump thoughts...

- plugin-jsdoc has `tagNamePreference` settings
- By this settings, user can create alias for tag name
  - e.g. `@prm {string} x` instead of `@param {string} x`, etc..
- This means that we cannot determine parsing logic from the tag name(kind)
- More example:
  - `@foo x` as `@param x`(x = type name) and `@foo x` as `@access x`(x = comment) cannot be distinguished

How to solve this?

- Give up to support `tagNamePreference`?
- Parse tag body lazily on demand?
  - `tag.parse_as_type_comment()`, `tag.parse_as_comment()`, etc...
  - How many syntax patterns there??

Boshen (approved):

Boshen:
Thank you! Sorry I missed the PR.

PR #2114, opened 2024-01-22
refactor(semantic): add binder for FormalParameters and RestElement, replacing the binder for FormalParameters

Dunqing:
Similar with #2013.

This way we will be able to find the corresponding Ast accurately.

The snapshots look a little strange but the change is expected.

PR #2644, opened 2024-03-08
fix(semantic): incorrect reference in parameters

### LINKED ISSUE #2499: Wrong scope analysis for catch block scope
## input

```js
try {
	throw 'FAIL';
} catch (t) {
	var t = 'PASS';
	assert.strictEqual(t, 'PASS');
}

let a = 1;
let def = 'PASS2';
try {
	throw ['FAIL2', 'PASS1'];
} catch ({ [a]: b, 3: d = def }) {
	let a = 0,
		def = 'FAIL3';
	assert.strictEqual(b, 'PASS1');
	assert.strictEqual(d, 'PASS2');
}
```

`def` in `d = def` and  `a` in `[a]: b` resolve to a wrong binding.

visualized scope id by https://github.com/oxc-project/oxc/pull/2498
```js
try{
        throw 'FAIL';
}catch(t#2){
        var t#2 = 'PASS';
        assert.strictEqual(t#2, 'PASS');
}let a#0 = 1;
let def#0 = 'PASS2';
try{
        throw ['FAIL2', 'PASS1'];
}catch({[a#4]:b#4, 3:d#4=def#4}){
        let a#4 = 0, def#4 = 'FAIL3';
        assert.strictEqual(b#4, 'PASS1');
        assert.strictEqual(d#4, 'PASS2');
}
```

Dunqing:
close: #2499

We need to collect parameter bindings. If an IdentifierReference is used on a parameter, we need to check if the current parameter binding is already defined, if not then it can only access the parent scope's bindings.

Let's look at an example.

```ts
const foo = 0;
function a(b = foo, c = b) {
   const foo = 1;
   const c = 2;
}  
```

The parameter has a reference to `foo`, but there is no `foo` in the parameter binding, so it must access the parent scope. Let's look at `c = b`, which has a reference to `b`, and the first parameter already has a `b` binding defined, so in the current scope we can access `b`.

I apologize for adding two more properties to the SemanticBuilder, the current implementation can't avoid adding them. Please let me know if you have a better solution, I'm interested in a better solution!

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -10,6 +10,7 @@ use oxc_syntax::{
     module_record::{ExportLocalName, ModuleRecord},
     operator::AssignmentOperator,
 };
+use rustc_hash::FxHashSet;
 
 use crate::{
     binder::Binder,
@@ -54,6 +55,9 @@ pub struct SemanticBuilder<'a> {
     /// If true, the current node is in the type definition
     in_type_definition: bool,
     current_reference_flag: ReferenceFlag,
+
+    is_inside_parameters: bool,
+    parameters_bindings: Vec<FxHashSet<&'a str>>,
 
     // builders
     pub nodes: AstNodes<'a>,
@@ -100,6 +104,8 @@ impl<'a> SemanticBuilder<'a> {
             nodes: AstNodes::default(),
             scope,
             symbols: SymbolTable::default(),
+            is_inside_parameters: false,
+            parameters_bindings: vec![],
             module_record: Arc::new(ModuleRecord::default()),
             label_builder: LabelBuilder::default(),
             jsdoc: JSDocBuilder::new(source_text, &trivias),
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#2644** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2644?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
@hyf0 cc

hyf0:
Would it be easier to create a facade scope for the params part?

Boshen:
My questions will always be: how does TypeScript do it?

Another questions, can we do everything inside `FormalParameters` instead of 

https://github.com/oxc-project/oxc/blob/c72675e89eb544e164b008d11d6034a323b26fb7/crates/oxc_semantic/src/binder.rs#L183-L242

Dunqing:
> My questions will always be: how does TypeScript do it?

I've taken a cursory look at typescript, but I don't see a specific way to handle it yet

> Another questions, can we do everything inside `FormalParameters` instead of
> 
> https://github.com/oxc-project/oxc/blob/c72675e89eb544e164b008d11d6034a323b26fb7/crates/oxc_semantic/src/binder.rs#L183-L242

CatchClause no `FormalParameters`

https://github.com/oxc-project/oxc/blob/7550838b0b951f12ff699a9e773b0927b00bf3c5/crates/oxc_ast/src/visit.rs#L291-L301

Boshen:
Can we add some tests so we can experiment with this?

Boshen:
I think catch clause is another special case ...

Dunqing:
> Another questions, can we do everything inside `FormalParameters` instead of
> 
> https://github.com/oxc-project/oxc/blob/c72675e89eb544e164b008d11d6034a323b26fb7/crates/oxc_semantic/src/binder.rs#L183-L242


Initially, we did everything in FormalParameters, but that would result in the node query by node id not being specific enough
https://github.com/oxc-project/oxc/pull/2114

Boshen:
Ahh ... I'll take a look at this next week :-(

PR #2567, opened 2024-03-02
fix(ast): add `RestElement`s in serialized AST to elements array

overlookmotel:
A step towards #2463.

This PR adds `rest` onto end of `elements` / `properties` array in JSON AST for `ObjectPattern`, `ArrayPattern`, `ObjectAssignmentTarget`, `ArrayAssignmentTarget` and `FormalParameters`.

PR #2652, opened 2024-03-09
fix(ast): fix serializing rest elements

overlookmotel:
Fix a mistake I made in #2567. Length that `serialize_seq` is called with should only be `+1` if there is a rest element being added on the end.

Makes no difference for serializing to JSON, as JSON serializer doesn't use the `len` value, but still better to get it right.

### DIFF
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -227,13 +227,15 @@ impl<'a, 'b, E, R> ElementsAndRest<'a, 'b, E, R> {
 
 impl<'a, 'b, E: Serialize, R: Serialize> Serialize for ElementsAndRest<'a, 'b, E, R> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let mut seq = serializer.serialize_seq(Some(self.elements.len() + 1))?;
-        for element in self.elements {
-            seq.serialize_element(element)?;
-        }
         if let Some(rest) = self.rest {
+            let mut seq = serializer.serialize_seq(Some(self.elements.len() + 1))?;
+            for element in self.elements {
+                seq.serialize_element(element)?;
+            }
             seq.serialize_element(rest)?;
+            seq.end()
+        } else {
+            self.elements.serialize(serializer)
         }
-        seq.end()
     }
 }


overlookmotel:
* **#2652** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ArnaudBarre:
Just checked now the implication of your last PR. I'm still updating the wrapper, but the types are now a bit wrong because they are generated from the original struct and not the one used in the serializer. Do you think this is fixable?

overlookmotel:
> Also the manual `AssignmentTargetRest` should be updated to use `argument` field

Thanks for spotting this. Damn! I thought I'd got it all correct in the end (and it was a bit of a battle).

I struggled a bit because I wasn't sure what AST type was our target for what we're trying to match. I was comparing to Acorn's output as produced by https://ast.sxzz.moe/ but that doesn't include the TS-related fields. What parser are we comparing to, and is that parser's output available on any playground/AST explorer?

On this specific one, could you possibly provide some source code which produces the incorrect AST? From looking at the Serde impls, it should already be renaming the `target` field to `argument`.

https://github.com/oxc-project/oxc/blob/32303b20fb54aafa60138c7133e31429d2421df5/crates/oxc_ast/src/ast/js.rs#L1040-L1045

> Just checked now the implication of your last PR. I'm still updating the wrapper, but the types are now a bit wrong because they are generated from the original struct and not the one used in the serializer. Do you think this is fixable?

Argh! I found `tsify` difficult. The types being generated from the original *should* be fine as it has `tsify(type = "...")` and `serde` attributes on the field which `tsify` macro uses to create the type defs:

https://github.com/oxc-project/oxc/blob/32303b20fb54aafa60138c7133e31429d2421df5/crates/oxc_ast/src/ast/js.rs#L1720-L1730

But I guess not...

It's tricky as CI checks that the type defs are valid (i.e. internally consistent), but doesn't notice if they don't match the actual AST objects.

Are you able to list which types are wrong? I'm sure I can fix them, but it'd help to have pointers where it's wrong (and were you checking them with a tool, or verifying them by eye?). Sorry to ask - but if you have time to give more info, I'll be able to get them fixed faster.

overlookmotel:
By the way, none of the above is pertinent to this PR, as far as I'm aware. I don't think it blocks this being merged if you agree it's a good change.

ArnaudBarre:
Yeah nothing blocking, just was also related to the previous PR.

For now I'm trying to match `@typescript-eslint/parser` because it's the default prettier parser. It's almost identical to Babel with `typescript` & `estree` plugins, but I found some subtitle variations like for #2654 & #2574. In both cases, I think the TSESLint team have done a good job at making an ESTree version that follow the TS syntax updates. I'm using `https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/ast-spec` to look at type definitions.

I'm not using special tools to detect differences, but I notice when things go wrong when I'm unable to update my wrapper implementation because of missing fields

overlookmotel:
Thanks for moving your comments Arnaud. I've made a bit of a mess here  by replying first and then deciding to open separate issues! Will reply on those issues...

overlookmotel:
Tests pass and I think the change is uncontentious, so merging...

PR #2673, opened 2024-03-10
fix(parser): fix span for JSXEmptyExpression with comment

ArnaudBarre:
[playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAICVgICAgICAgICejwtjmCpbllbPawdM2eEFKwhGb62iFlQWu39yrLCA)


Boshen:
`prev_span_end` is a private value that shouldn't be used in parse methods, It seems like it's always a +1 / -1 from the container spans.

Boshen (approved):

PR #2678, opened 2024-03-11
feat(linter/import): support check reexport binding in namespace

Dunqing:


Dunqing:
* **#2678** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (commented):
I think we should add reexported bindings inside the linter service instead of doing a very expensive lookup here.

Dunqing:
> I think we should add reexported bindings inside the linter service instead of doing a very expensive lookup here.

Can you explain what I have to do? I don't quite understand `add reexported bindings inside the linter service`

PR #2680, opened 2024-03-11
refactor(span): change shape of `Language`

### LINKED ISSUE #2677: Change shape of `Language`
Could we change the shape of `oxc_span::source_type::Language`?

Currently it's:

https://github.com/oxc-project/oxc/blob/d47f0e247db24319995fbe4f26255379a931db6d/crates/oxc_span/src/source_type.rs#L29-L35

Could we change this to either:

```rs
pub enum Language {
    JavaScript,
    TypeScript,
    TypeScriptDefinition,
}
```

or:

```rs
pub enum Language {
    JavaScript,
    TypeScript(TypescriptVariant),
}
pub struct TypescriptVariant {
    is_definition_file: bool,
}
```

Reason I'm asking is that it's tricky the way it is now for #2409, because [memoffset::offset_of](https://docs.rs/memoffset/0.9.0/memoffset/macro.offset_of.html) does not allow getting the offset of a nested enum variant field, due to some weird compiler quirk. This is the only type in the AST which has this problem.

The first option seems cleaner to me, but is there a reason the JSON AST needs to be as it is now? Currently it's one of:

* `"language": "javaScript"`
* `"language": { "typeScript": { "isDefinitionFile": false } }`
* `"language": { "typeScript": { "isDefinitionFile": true } }`

overlookmotel:
Closes #2677.

Also changed the JSON output to be `javascript`, `typescript` or `typescriptDefinition`. The current value `javaScript` / `typeScript` is a bit weird.

### DIFF
--- a/crates/oxc_span/src/source_type.rs
+++ b/crates/oxc_span/src/source_type.rs
@@ -31,13 +31,12 @@ pub struct SourceType {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, SerAttrs)]
 #[cfg_attr(feature = "serde", derive(Serialize))]
 #[cfg_attr(feature = "wasm", derive(Tsify))]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "lowercase")]
 pub enum Language {
     JavaScript,
-    #[serde(rename_all = "camelCase")]
-    TypeScript {
-        is_definition_file: bool,
-    },
+    TypeScript,
+    #[serde(rename = "typescriptDefinition")]
+    TypeScriptDefinition,
 }
 
 /// Script or Module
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#2680** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_span/src/source_type.rs
```diff
@@ -182,14 +181,15 @@ impl SourceType {
...
 
-        let is_definition_file = file_name.ends_with(".d.ts")
-            || file_name.ends_with(".d.mts")
-            || file_name.ends_with(".d.cts");
-
         let language = match extension {
             "js" | "mjs" | "cjs" | "jsx" => Language::JavaScript,
-            "ts" | "mts" | "cts" | "tsx" => Language::TypeScript { is_definition_file },
-            _ => unreachable!(),
+            "ts" if file_name.ends_with(".d.ts") => Language::TypeScriptDefinition,
```
Boshen, line 186: This is 3 string matches, can be reduced to strip start + match?

overlookmotel: My intent with moving the `ends_with` checks was to avoid testing `ends_with(".d.cts")` when `extension` is `ts`, as it can never match.

Boshen: Oh, never mind, I thought I had something more clever 😅

overlookmotel: Yes it's not completely optimal. But as Arnaud said on #2677, maybe this should have further changes anyway.

PR #2702, opened 2024-03-13
feat(linter): eslint-plugin-jest: `prefer-to-be`

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-be.ts)


mysteryven:
I'll take a closer look at it tonight!

crates/oxc_linter/src/utils/jest.rs
```diff
@@ -289,6 +290,35 @@ pub fn get_node_name_vec<'a>(expr: &'a Expression<'a>) -> Vec<Cow<'a, str>> {
...
+    match expr {
+        Expression::TSAsExpression(ts_as_expr) => {
+            follow_type_assertion_chain(&ts_as_expr.expression)
+        }
+        Expression::TSTypeAssertion(ts_type) => follow_type_assertion_chain(&ts_type.expression),
+        _ => expr,
+    }
+}
+
+pub fn get_first_matcher_arg<'a>(
```
mysteryven, line 310: For a utils function, could it avoid using `unwrap`?

crates/oxc_linter/src/rules/jest/prefer_to_be.rs
```diff
@@ -0,0 +1,511 @@
...
+        let span = matcher.span;
+        let end = call_expr.span.end;
+        let Some(Expression::MemberExpression(mem_expr)) = matcher.parent else {
+            return;
+        };
+        let is_cmp_mem_expr = matches!(mem_expr.0, MemberExpression::ComputedMemberExpression(_));
+        let modifiers = jest_expect_fn_call.modifiers();
+        let maybe_not_modifier = modifiers.iter().find(|modifier| modifier.is_name_equal("not"));
+
+        if kind == "Undefined" {
```
mysteryven, line 182: Would you like to use `Enum` to replace this string flag? I think it will be more robust.

crates/oxc_linter/src/rules/jest/prefer_to_be.rs
```diff
@@ -0,0 +1,511 @@
...
+                Self::check_and_fix("Undefined", call_expr, matcher, &jest_expect_fn_call, ctx);
+                return;
+            }
+        }
+
+        if !is_equality_matcher(matcher) || jest_expect_fn_call.args.len() == 0 {
+            return;
+        }
+
+        let first_matcher_arg = get_first_matcher_arg(&jest_expect_fn_call);
```
mysteryven, line 119: Seems like `get_inner_expression` is enough.

crates/oxc_linter/src/utils/jest.rs
eryue0220: I removed this function, `get_inner_expression` is enough instead.

crates/oxc_linter/src/rules/jest/prefer_to_be.rs
eryue0220: Update.

crates/oxc_linter/src/rules/jest/prefer_to_be.rs
eryue0220: Yes. Update.

mysteryven (approved):
Thank you!

PR #2703, opened 2024-03-13
feat(linter): eslint-plugin-jest: `prefer-expect-resolves`

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-expect-resolves.ts)


crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
```diff
@@ -0,0 +1,267 @@
...
+        };
+        let Some(expect_fn_call) = parse_expect_jest_fn_call(call_expr, possible_jest_node, ctx)
+        else {
+            return;
+        };
+        let Some(parent) = expect_fn_call.head.parent else {
+            return;
+        };
+
+        if !parent.is_call_expression() {
```
mysteryven, line 98: Seems useless because of the following check?

crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        ctx.diagnostic_with_fix(ExpectResolves(await_expr.span), || {
+            let content = Self::build_code(call_expr, ctx);
+            Fix::new(content, call_expr.span)
+        });
+    }
+
+    fn build_code<'a>(call_expr: &CallExpression<'a>, ctx: &LintContext<'a>) -> String {
+        let mut formatter = ctx.codegen();
+        let ident = call_expr.callee.get_identifier_reference().unwrap();
```
mysteryven, line 120: Feels like using `unwrap` not 100% safety, even I can't show a case now.

crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
```diff
@@ -0,0 +1,267 @@
...
+        let offset = match &await_expr.argument {
+            Expression::CallExpression(call_expr) => call_expr.span.start - ident.span.end,
+            Expression::Identifier(promise_ident) => promise_ident.span.start - ident.span.end,
+            _ => 0,
+        };
+
+        let arg_span = Span::new(
+            call_expr.span.start + (ident.span.end - ident.span.start) + offset,
+            call_expr.span.end,
+        );
```
mysteryven, line 135: Do you mind adding some comment here to show the `offset` and `arg_span` :)

mysteryven (commented):
LGTM, just some nitpick.

crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
eryue0220: I opened a new [PR](https://github.com/oxc-project/oxc/pull/2715) for this, which I can get the identifier directly, and it has been parsed by `parse_expect_jest_fn`.

mysteryven: merged.

crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
eryue0220: Yes, I remove it.

crates/oxc_linter/src/rules/jest/prefer_expect_resolves.rs
eryue0220: There is no `span` in `await_expr.argument`, but it' not enough. because the keyword of `await` in expect need to be removed as well.

PR #2706, opened 2024-03-13
feat(linter): Add settings.jsdoc

leaysgur:
Base work for #2642 🏃🏻 

- [x] struct
- [x] bool flags
- [x] tagNamePreferences

### DIFF
--- a/crates/oxc_linter/src/config/settings/jsdoc.rs
+++ b/crates/oxc_linter/src/config/settings/jsdoc.rs
@@ -4,79 +4,114 @@ use serde::Deserialize;
 /// https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/settings.md
 #[derive(Debug, Deserialize, Default)]
 pub struct ESLintSettingsJSDoc {
-    /// Not apply to `check-access` and `empty-tags` rule
+    /// For all rules but NOT apply to `check-access` and `empty-tags` rule
     #[serde(default, rename = "ignorePrivate")]
     pub ignore_private: bool,
-    /// Not apply to `empty-tags` rule
+    /// For all rules but NOT apply to `empty-tags` rule
     #[serde(default, rename = "ignoreInternal")]
     pub ignore_internal: bool,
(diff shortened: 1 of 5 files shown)


crates/oxc_linter/src/config/settings/jsdoc.rs
```diff
@@ -4,79 +4,114 @@ use serde::Deserialize;
...
+    //   disallowName?: string;
+    //   allowName?: string;
+    //   context?: string;
+    //   comment?: string;
+    //   tags?: string[];
+    //   replacement?: string;
+    //   minimum?: number;
+    //   message?: string;
+    //   forceRequireReturn?: boolean;
+    // }[]
```
Boshen, line 71: I think the rule of thumb is to add more data when required, otherwise we may end up implementing useless stuff.

crates/oxc_linter/src/config/settings/jsdoc.rs
```diff
@@ -0,0 +1,207 @@
+use rustc_hash::FxHashMap;
+use serde::Deserialize;
+
+/// https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/settings.md
+#[derive(Debug, Deserialize, Default)]
+pub struct ESLintSettingsJSDoc {
```
Boshen, line 6: Maybe `JSDocPluginSettings`?

leaysgur: Renamed along with other structs. 👉🏻 https://github.com/oxc-project/oxc/pull/2706/commits/4267517d19826f3e91ca9313deb40ddbb713dac5

crates/oxc_linter/src/config/settings/jsdoc.rs
leaysgur: Yes, I agree! 
(But the most difficult part of this plugin is that everything is tightly coupled and completely unpredictable where any of them will be needed. 😂)

PR #2712, opened 2024-03-13
feat(prettier/ts): print `type` keyword inside import specifier braces, fixes #2708

### LINKED ISSUE #2708: Prettier drops `type` keyword inside import specifier braces
## Details

It's working as expected
```ts
import type { FC } from 'react'
```

But in this case, `type` is dropped
```ts
import { type FC } from 'react'
// Will be
import { FC } from 'react'
```

## Expected Behavior
If only one specifier is listed, `type` will be moved to the outside, otherwise nothing will change
```ts
// One specifier
import { type FC } from 'react'
// Will be
import type { FC } from 'react'

// Multiple specifiers
import { type FC, useState } from 'react'
// Nothing changed
import { type FC, useState } from 'react'
```

LIMPIX31:
- #2708

### DIFF
--- a/crates/oxc_prettier/src/format/mod.rs
+++ b/crates/oxc_prettier/src/format/mod.rs
@@ -1102,14 +1102,16 @@ impl<'a> Format<'a> for ImportDeclarationSpecifier<'a> {
 
 impl<'a> Format<'a> for ImportSpecifier<'a> {
     fn format(&self, p: &mut Prettier<'a>) -> Doc<'a> {
-        if self.imported.span() == self.local.span {
-            if self.import_kind.is_type() && p.import_inner_type {
-                array![p, ss!("type "), self.local.format(p)]
-            } else {
-                self.local.format(p)
-            }
+        let typed = if self.import_kind.is_type() && p.import_inner_type {
+            ss!("type ")
         } else {
-            array![p, self.imported.format(p), ss!(" as "), self.local.format(p)]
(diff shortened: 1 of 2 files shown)


Boshen:
Thank you. Formatting TypeScript is only 4% done: https://github.com/oxc-project/oxc/blob/main/tasks/prettier_conformance/prettier.ts.snap.md

There's currently no one working on this so any help is appreciated.

Boshen:
It seems like we are adding custom logic (`import_inner_type`) in this PR, I prefer not to introduce these options because this is not part of our initial goals.

I'm happy to merge if this PR only prints `type` without any custom logic.

LIMPIX31:
> It seems like we are adding custom logic (`import_inner_type`) in this PR, I prefer not to introduce these options because this is not part of our initial goals.
> 
> I'm happy to merge if this PR only prints `type` without any custom logic.

No problem. But I need to think of a way around the use of this state.

Is there now a way to store the state or some context when formatting?

Boshen:
`prettier` doesn't move the `type` keyword to the outside, I want to refrain from adding these kind of custom logic at this stage.

LIMPIX31:
> `prettier` doesn't move the `type` keyword to the outside, I want to refrain from adding these kind of custom logic at this stage.

Then that simplifies it)

mysteryven (approved):
Thank you for your patience and work. :)

PR #2716, opened 2024-03-14
feat: merge features `serde` and `wasm` to `serialize`

Boshen:
This PR merges the previous confusing features `serde` and `wasm` into a single `serialize` feature.

We'll eventually do serialize + type information for both wasm and napi targets.

`oxc_macros` is removed from `oxc_ast`'s dependency because it requires `syn` and friends, which goes against our policy ["Third-party dependencies should be minimal."](https://oxc-project.github.io/docs/contribute/rules.html#development-policy)


Boshen:
@overlookmotel please forgive me 🙇

Boshen:
Ahh ... so much pain from testing this locally due to feature unification 😡

Boshen:
This is blocking crate releases so I'm going to merge this.

overlookmotel:
Oh @Boshen! Why do you hurt me so???

Can I propose we look at this again? Agree it's good to combine "serde" and "wasm" features, and removing `oxc_ast`'s dependency on `syn` is good. But I do think the reappearance of all the `#[cfg_attr]`s is a major downside for readability.

How about this:

* Add a crate `oxc_ser_macros`.
* `oxc_ser_macros` contains just the `SerAttrs` macro.
* `oxc_ser_macros` crate does not need a dependency on `syn`, as the macro doesn't actually do anything. It's just a hack to enable using `#[serde]` alone without wrapping it in `#[cfg_attr]` everywhere.

Because the `SerAttrs` macro is a no-op, I believe its impact on compile times will be insignificant. This would solve the problem you wanted to solve, but without sacrificing readability.

What do you think?

overlookmotel:
By the way, I think we will be able to remove `Tsify` later on. The types schema which is created for AST transfer can also be used to create TS type defs - and they could be added to `oxc-parser` NPM package as well as WASM.

PR #2720, opened 2024-03-14
feat(linter/import) implement no_unused_modules rule

Dunqing:


Dunqing:
* **#2720** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
> > I'm fine with that.
> 
> Can we retrieve and save all the src files in the linter service in advance so that we don't have to repeat the query every time we run `run_once`?

Can you give an example? I haven't looked at what this rule does yet.

Dunqing:
> > > I'm fine with that.
> > 
> > 
> > Can we retrieve and save all the src files in the linter service in advance so that we don't have to repeat the query every time we run `run_once`?
> 
> Can you give an example? I haven't looked at what this rule does yet.

You can see https://github.com/import-js/eslint-plugin-import/blob/c77c1a6ace311f99ed25878768c2dbee49c0512a/docs/rules/no-unused-modules.md#example-for-unused-exports

Dunqing:
Support for the unused_exports option is not considered for the time being

PR #2724, opened 2024-03-15
chore(napi): add `oxc-parser` to benchmarks

### LINKED ISSUE #2616: Benchmark NodeJS parser
For #2457 and also the work around ESTree-izing the parser's JSON output, it'd be good to be benchmarking the NodeJS [oxc-parser](https://www.npmjs.com/package/oxc-parser) interface.

Any idea how we could go about this? [CodSpeed](https://codspeed.io/) says it supports NodeJS. But:

Should we create a separate set of benchmarks in CodSpeed for it, or try to get it added to the current list?

If NodeJS OXC parser is a primary performance target, it could be good to add it to main list, so we see any performance regressions/improvements on every commit, same as we do for the Rust interfaces.

A hacky way to do this would be:

* Add a benchmark implemented in Rust.
* That benchmark starts up NodeJS as a child process.
* On each turn of the benchmark loop in Rust, write a line to child process's stdin.
* NodeJS process waits to receive a line on stdin, runs parser, and then writes a line to stdout to signal that it's done.
* Rust benchmark code waits until it receives a line on child process's stdout, then completes the turn of the benchmark loop.

Purpose of co-ordinating over stdin/stdout is avoid creating and destroying a NodeJS child process on each turn of the benchmark loop, as that overhead would be high, and not representative of real-world usage where typically you'd parse many files in a single NodeJS process.

But is there a better way than this?

overlookmotel:
Closes #2616.

Adds benchmarks for NodeJS NAPI build. Measurement includes `JSON.parse` of the AST on JS side, since that's how it'll be used 99% of the time.

Benchmarks run against same files as Rust parser benchmarks, so we can see the overhead of transferring AST to JS.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,6 +38,12 @@ jobs:
           shared-key: 'benchmark'
           save-cache: ${{ github.ref_name == 'main' }}
 
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org'
+
       - name: Install codspeed
         uses: taiki-e/install-action@v2
         with:
@@ -59,9 +65,19 @@ jobs:
           mv target/release/deps/codegen_sourcemap-* target/codspeed/oxc_benchmark
           rm -rf target/codspeed/oxc_benchmark/*.d
 
+      - name: Build NAPI benchmark
+        working-directory: ./napi/parser
+        run: |
+          corepack enable
+          pnpm install
+          pnpm run build
+
       - name: Run benchmark
         uses: CodSpeedHQ/action@v2
         timeout-minutes: 30
         with:
-          run: cargo codspeed run
+          run: |
+            cargo codspeed run
+            cd napi/parser
+            pnpm run bench
           token: ${{ secrets.CODSPEED_TOKEN }}
--- a/napi/parser/package.json
+++ b/napi/parser/package.json
@@ -3,17 +3,20 @@
   "private": true,
   "scripts": {
     "build": "napi build --platform --release",
-    "test": "node test.mjs"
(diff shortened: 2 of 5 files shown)


overlookmotel:
* **#2724** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
NB: I initially tried making the JS benchmarks a separate task, but codspeed then only displays 1 set of benchmarks or the other - whichever results arrive latest clobber the other task's results. It seems all benchmarks have to be in a single call to their `CodSpeedHQ/action` action.

overlookmotel:
Had to update pnpm as was failing to install with `ERR_INVALID_THIS` errors.

Seems to be a [common problem with pnpm](https://github.com/pnpm/pnpm/issues/6424) and only apparent solution is to update to latest.

Boshen:
> NB: I initially tried making the JS benchmarks a separate task, but codspeed then only displays 1 set of benchmarks or the other - whichever results arrive latest clobber the other task's results. It seems all benchmarks have to be in a single call to their `CodSpeedHQ/action` action.

Let me call for help.

overlookmotel:
I think it's basically the same problem as sharding the benchmarks (https://github.com/CodSpeedHQ/action/issues/92). If we can figure out how to do that, we can also move NodeJS/NAPI benchmark into a separate task.

From looking at https://github.com/CodSpeedHQ/runner, I do think this should be possible to do, though it'll require some experiementation.

I'm really keen to get the NAPI benchmarks up and running, so I can get #2457 to a state where it can be merged. Would you consider merging this now, and then returning to the sharding problem later?

Boshen:
Ok let's see what happens.

PR #2736, opened 2024-03-15
implemented import/no-default-export

XantreDev:
Tests: https://github.com/import-js/eslint-plugin-import/blob/v2.29.1/tests/src/rules/no-default-export.js
Rule: https://github.com/import-js/eslint-plugin-import/blob/v2.29.1/docs/rules/no-default-export.md


Boshen (approved):

Boshen:
I updated the snapshot.

PR #2739, opened 2024-03-16
feat(linter): eslint/max-lines

andipabst:
Relates to #479 

Rule detail: https://eslint.org/docs/latest/rules/max-lines

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/max_lines.rs
+++ b/crates/oxc_linter/src/rules/eslint/max_lines.rs
@@ -51,7 +51,7 @@ declare_oxc_lint!(
     /// ```javascript
     /// ```
     MaxLines,
-    correctness
+    pedantic
 );
 
 impl Rule for MaxLines {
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/max_lines.rs
```diff
@@ -0,0 +1,320 @@
...
+    /// ### Why is this bad?
+    ///
+    /// Some people consider large files a code smell. Large files tend to do a lot of things and can make it hard following what’s going.
+    /// While there is not an objective maximum number of lines considered acceptable in a file, most people would agree it should not be in the thousands. Recommendations usually range from 100 to 500 lines.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// ```
+    MaxLines,
+    correctness
```
mysteryven, line 58: There are several categories in our linter, Clippy has a [similar rule](https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_lines), its category is `pedantic`, what about we keep the same with it?

https://github.com/oxc-project/oxc/blob/32504cadb745c4f81be89fa74fb317822daee1ed/crates/oxc_linter/src/rule.rs#L33-L54

crates/oxc_linter/src/snapshots/max_lines.snap
```diff
@@ -0,0 +1,237 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: max_lines
+---
+  ⚠ eslint(max-lines): "File has too many lines (3). Maximum allowed is 2."
+   ╭─[max_lines.tsx:1:1]
+ 1 │ var xyz;
```
mysteryven, line 7: Looks like the range of error is not really corresponding :)

mysteryven:
We may need to run `just ready` in local to check if there are some broken.

crates/oxc_linter/src/rules/eslint/max_lines.rs
andipabst: Thanks for the hint!

crates/oxc_linter/src/snapshots/max_lines.snap
andipabst: Thanks for the hint, my usage of Span was not correct! Fixed now.

mysteryven (approved):
Learned new things from your implementation, thank you!

PR #2740, opened 2024-03-17
feat(transformer/react-jsx): support development mode for jsx transform

fireairforce:
support: development mode https://github.com/babel/babel/blob/ff3481746a830e0e94626de4c4cb075ea5f2f5dc/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L77-L81


Boshen:
Thank you for the PR, we are continuing the work in the redo.

PR #2741, opened 2024-03-17
feat(linter): eslint/no-with

andipabst:
Relates to #479 

Rule detail: https://eslint.org/docs/latest/rules/no-with


mysteryven (approved):
Nice work!

PR #2744, opened 2024-03-17
feat(linter): eslint/no-ternary

JoSeBu1:
Rule detail: https://eslint.org/docs/latest/rules/no-ternary

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_ternary.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_ternary.rs
@@ -1,0 +1,67 @@
+use oxc_ast::{ast::Expression,AstKind};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::{self, Error},
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint(no-ternary): Unexpected use of ternary expression")]
+#[diagnostic(severity(warning), help("Do not use the ternary expression."))]
+struct NoTernaryDiagnostic(#[label] pub Span);
+
+#[derive(Debug, Default, Clone)]
+pub struct NoTernary;
+
+declare_oxc_lint!(
+    /// ### What it does
+    /// Disallow ternary operators
+    ///
+    /// ### Why is this bad?
+    /// The ternary operator is used to conditionally assign a value to a variable. Some believe that the use of ternary operators leads to unclear code.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// var foo = isBar ? baz : qux;
+    //
+    // function quux() {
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/src/rules/eslint/no_ternary.rs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+
+    let pass = vec![r#""x ? y";"#, "if (true) { thing() } else { stuff() };"];
```
JoSeBu1, line 58: @mysteryven I have a problem with this first test, because is as rulegen created it, but seems strange to my, and I don't know how to pass it, I tried adding this test to other places where ternaries are checked and it fails. The second test was added by me, let me know if there is some problem. Thanks! 
https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-ternary.js

mysteryven: Ops, maybe there are some broken in our rulegen, we can change it to `let pass = vec![r#""x ? y";"#];`. 

It's a bit strange, I run `just new-rule no-ternary` on my compute, the test case is normal:

```
#[test]
fn test() {
    use crate::tester::Tester;

    let pass = vec![r#""x ? y";"#];

    let fail = vec![
        "var foo = true ? thing : stuff;",
        "true ? thing() : stuff();",
        "function foo(bar) { return bar ? baz : qux; }",
    ];

    Tester::new(NoTernary::NAME, pass, fail).test_and_snapshot();
}
```

JoSeBu1: @mysteryven Maybe I haven't explained it correctly, the result of the rulegen is the same for you and me, as I said, I added the second test, the problem is that the first test that should pass fails.

JoSeBu1: Sorry, I'm reviewing it and I have confused the error, the fail comes from other test

mysteryven: Sorry to misunderstand, are you confused the by the [`r#` syntax](https://doc.rust-lang.org/rust-by-example/std/str.html#literals-and-escapes) 🤔, `r#""x ? y";"#` is same as `"x ? y"` in JavaScript.

JoSeBu1: Now should be working, but I don't understand why I haven't to check if the conditional is a binary operator 🤔

mysteryven: The `run` function will visit every AST node in the source text :)

https://github.com/oxc-project/oxc/blob/02c82c3f78faa8db832a519e953986fa430da0d6/crates/oxc_linter/src/lib.rs#L124-L130

mysteryven (approved):
Nice work!

PR #2746, opened 2024-03-17
feat(linter): eslint/guard-for-in

JoSeBu1:
Rule details: https://eslint.org/docs/latest/rules/guard-for-in

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/guard_for_in.rs
+++ b/crates/oxc_linter/src/rules/eslint/guard_for_in.rs
@@ -37,12 +37,16 @@ declare_oxc_lint!(
 impl Rule for GuardForIn {
     fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
         if let AstKind::ForInStatement(for_in_statement) = node.kind() {
-
             match &for_in_statement.body {
                 Statement::EmptyStatement(_) => return,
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/guard_for_in.rs
```diff
@@ -0,0 +1,100 @@
...
+                                return;
+                            }
+                        }
+                    }
+                }
+                _ => {}
+            }
+            ctx.diagnostic(GuardForInDiagnostic(Span::new(
+                for_in_statement.span.start,
+                for_in_statement.span.end,
```
mysteryven, line 71: Well done, just one nitpick: maybe the `for..in` loop has a long body, is this possible to make the span position smaller?

mysteryven: Do you think the first statement of `for..in`‘s body  is suitable?

JoSeBu1: I don't know what can be done, I will take a look at it reviewing another snapshots

mysteryven: Just advice, It's also good now!

mysteryven (approved):
Thank you!

PR #2754, opened 2024-03-18
feat(linter): eslint-plugin-react checked-requires-onchange-or-readonly

keita-hino:
partof: https://github.com/oxc-project/oxc/issues/1022

docs: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/checked-requires-onchange-or-readonly.md
code: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/checked-requires-onchange-or-readonly.js
test: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/checked-requires-onchange-or-readonly.js


overlookmotel:
@keita-hino In case you're wondering why I have rebased this on main, I made a mistake when converting the benchmarks to run sharded, which lead to benchmarks results not being submitted correctly on some PRs. Now fixed (#2759) so am re-running benchmarks on affected PRs by rebasing on top of the fix.

keita-hino:
@overlookmotel 
Thank you for the fix!

crates/oxc_linter/src/snapshots/checked_requires_onchange_or_readonly.snap
```diff
@@ -0,0 +1,101 @@
...
+   ╭─[checked_requires_onchange_or_readonly.tsx:1:1]
+ 1 │ <input type='checkbox' checked defaultChecked />
+   · ────────────────────────────────────────────────
+   ╰────
+  help: Remove either `checked` or `defaultChecked`.
+
+  ⚠ eslint-plugin-react(checked-requires-onchange-or-readonly): `checked` should be used with either `onChange` or `readOnly`.
+   ╭─[checked_requires_onchange_or_readonly.tsx:1:1]
+ 1 │ <input type='checkbox' checked defaultChecked />
+   · ────────────────────────────────────────────────
```
Dunqing, line 99: Let's point the span to the `checked` attribute

crates/oxc_linter/src/snapshots/checked_requires_onchange_or_readonly.snap
```diff
@@ -0,0 +1,101 @@
...
+   ╰────
+  help: Add either `onChange` or `readOnly`.
+
+  ⚠ eslint-plugin-react(checked-requires-onchange-or-readonly): `checked` should be used with either `onChange` or `readOnly`.
+   ╭─[checked_requires_onchange_or_readonly.tsx:1:1]
+ 1 │ React.createElement('input', { checked: false })
+   · ────────────────────────────────────────────────
```
Dunqing, line 57: Let's point the span to the `checked` attribute

Dunqing (changes requested):
Thank you for your work! Let's improve the error message to make it clearer.

crates/oxc_linter/src/snapshots/checked_requires_onchange_or_readonly.snap
keita-hino: Thank you!
I have corrected it to point to the `checked` attribute.
e8e9bb3509261b48ff243eb0c1f23fc5036d9964

keita-hino:
I'm reconsidering whether performance can be improved.

crates/oxc_linter/src/rules/react/checked_requires_onchange_or_readonly.rs
```diff
@@ -0,0 +1,270 @@
...
+        if !self.ignore_exclusive_checked_attribute && props.contains(&"defaultChecked".to_string())
+        {
+            ctx.diagnostic(CheckedRequiresOnchangeOrReadonlyDiagnostic::ExclusiveCheckedAttribute(
+                span,
+            ));
+        }
+
+        if !(self.ignore_missing_properties
+            || props.contains(&"onChange".to_string())
+            || props.contains(&"readOnly".to_string()))
```
Dunqing, line 189:
```suggestion
            || props.contains("onChange")
            || props.contains("readOnly"))
```

I think we don't need to use `to_string`

Dunqing (changes requested):

Dunqing:
> I'm reconsidering whether performance can be improved.

We should try to avoid making memory allocations such as calling `to_string`

crates/oxc_linter/src/rules/react/checked_requires_onchange_or_readonly.rs
keita-hino: Thank you.
When `to_string` was removed, the following error was displayed 🤔

> mismatched types expected reference `&std::string::String` found reference `&'static str`

Dunqing: Try to see if you can change the type of prop_names to Vec<&'a str>.

keita-hino: After attempting the fix, the error within `check_attributes_and_report` disappeared.
Instead, the following error started occurring in the caller of `check_attributes_and_report`:

> mismatched types expected struct std::vec::Vec<&str> found reference &std::vec::Vec<std::string::String>

Since the [prop_names in the caller](https://github.com/oxc-project/oxc/pull/2754/files#diff-948df61eb03dd7247d0b26e3583f7e3d4ded6b1606bf637b9f02821beabafef8R130-R142) is of type Vec\<String>, I attempted to modify it to Vec<&str>, which resulted in the following error:

> cannot return value referencing temporary value returns a value referencing data owned by the current function

```diff rust
- let prop_names: Vec<String> = obj_expr
+ let prop_names: Vec<&str> = obj_expr
    .properties
    .iter()
    .filter_map(|attr| {
        if let ObjectPropertyKind::ObjectProperty(obj_prop) = attr {
            let name = &obj_prop.key.name()?;
            if is_target_prop(name) {
-                 return Some(name.to_string());
+                 return Some(name.as_str());
            }
        }
        None
    })
    .collect();
```

Dunqing: Oh, yes, we can't do that. Thank you for your try!

Dunqing:
I suddenly realized that you don't need to collect `prop_names`, you just need to iterate through the properties and if you find `readOnly` or `onChange` that's the end of it, if not, emit a diagnostic message

keita-hino:
I have marked the PR as draft to address the performance degradation.

Dunqing:
The current implementation works well, but it seems we could still improve it by traversing once to find `checked`, `readOnly`, `onChange` and `defaultChecked`.

Dunqing (changes requested):

crates/oxc_linter/src/snapshots/checked_requires_onchange_or_readonly.snap
```diff
@@ -0,0 +1,101 @@
...
+   ╭─[checked_requires_onchange_or_readonly.tsx:1:24]
+ 1 │ <input type='checkbox' checked defaultChecked />
+   ·                        ───────
+   ╰────
+  help: Add either `onChange` or `readOnly`.
+
+  ⚠ eslint-plugin-react(checked-requires-onchange-or-readonly): Use either `checked` or `defaultChecked`, but not both.
+   ╭─[checked_requires_onchange_or_readonly.tsx:1:24]
+ 1 │ <input type='checkbox' checked defaultChecked />
+   ·                        ───────
```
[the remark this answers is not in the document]
keita-hino: Thank you!
I've corrected to include the `defaultChecked` attribute in pointing for ExclusiveCheckedAttribute.
4aa1f5a3fc6c73c442d3e65cbb2374faf6c8e141

crates/oxc_linter/src/rules/react/checked_requires_onchange_or_readonly.rs
```diff
@@ -0,0 +1,294 @@
...
+                                        checked_span
+                                    },
+                                    if default_checked_span.is_none() && name == "defaultChecked" {
+                                        Some(jsx_attr.span)
+                                    } else {
+                                        default_checked_span
+                                    },
+                                    is_missing_property
+                                        && !(name.contains("onChange")
+                                            || name.contains("readOnly")),
```
Dunqing, line 95: Looks we should use `==`, consider we have a attribute and it naming `onChange2` or `readOnlyXX`

Dunqing (changes requested):

crates/oxc_linter/src/rules/react/checked_requires_onchange_or_readonly.rs
keita-hino: Thank you!
I've corrected to use `==`.
5825a98feca9e1a84a05bfc4e2148f4822e4e692

Dunqing (approved):
LGTM, Thanks for the effort you put into improving it!

keita-hino:
@Dunqing Thank you for your thoughtful review and suggestions for improvement🙏  I've learned a lot!

PR #2755, opened 2024-03-18
fix(linter): fix guard_for_in span error

JoSeBu1:
As talked here https://github.com/oxc-project/oxc/pull/2746#discussion_r1528145418 errors in guard_for_in could be improved so I attempted to improve it.

If anyone have a suggestion for the span element positions say me!


mysteryven (approved):
It's a nice way!

PR #2757, opened 2024-03-18
feat(linter): no_eq_null

JoSeBu1:
Rule detail: https://eslint.org/docs/latest/rules/no-eq-null


Dunqing (approved):
Thank you for your awesome work!

PR #2761, opened 2024-03-18
feat(linter): no_script_url

JoSeBu1:
Rule detail: https://eslint.org/docs/latest/rules/no-script-url

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_script_url.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_script_url.rs
@@ -46,14 +46,15 @@ impl Rule for NoScriptUrl {
             AstKind::TemplateLiteral(literal)
                 if !is_tagged_template_expression(ctx, node, literal.span) =>
             {
-                if literal
-                    .quasis
-                    .first()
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/no_script_url.rs
```diff
@@ -0,0 +1,98 @@
...
+                if !is_tagged_template_expression(ctx, node, literal.span) =>
+            {
+                if literal
+                    .quasis
+                    .first()
+                    .unwrap()
+                    .value
+                    .raw
+                    .to_lowercase()
+                    .starts_with("javascript:")
```
mysteryven, line 56: Looks like [ESLint](https://github.com/eslint/eslint/blob/9aa4df3f4d85960eee72923f3b9bfc88e62f04fb/lib/rules/utils/ast-utils.js#L230-L232) also check `quasis`'s length only be 1.

JoSeBu1: done

mysteryven (approved):

PR #2767, opened 2024-03-19
feat(cli): wildcard expansion in paths for windows

### LINKED ISSUE #2695: doc(cli): You have to specify the directory, not wildcards.
Well, found the problem. You have to specify the directory, not wildcards.
```
$ npx oxlint src/components
Finished in 8ms on 7 files with 87 rules using 12 threads.
```
Seems a problem specific to Windows that does not expand wildcards on the command line.
Maybe adding this to the documentation could help...

_Originally posted by @crystalfp in https://github.com/oxc-project/oxc/issues/1915#issuecomment-1991639303_

andipabst:
Unlike on other OS, on Windows there is no wildcard expansion/globbing by the shell. Instead the application has to handle this. Therefore I used the `glob` package to handle wildcards on Windows.

I also had to make the parent directory check more strict due to the glob package resolving `..` in the middle of the path as well.

This closes #2695.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
 
 [[package]]
+name = "errno"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Boshen (commented):
Thank you!

I'm wondering, does all cli applications have this kind of logic for windows? Seems like a weird thing to do for every single cli application :-/

Boshen (approved):

Boshen:
Thank you for the explanation!

PR #2776, opened 2024-03-20
refactor(ast): add walk_mut functions

rzvxa:
In this PR I'm going to address the issue of our visitor abstraction using the walk functions similar to how it is implemented in the `rustc` visitors.

It is the continuation of #2764, and should get merged after it since I've continued on those commits.


rzvxa:
If you guys really want it, We can squash this PR instead of having 2 but I think since they are addressing different problems we should keep them seperated.

rzvxa:
Since this PR has already exceeded 5k lines of code moving around, I'm going to break it into 2 parts to make merging it less of a chore.
In this PR I've tried to implement walk functions for the VisitMut trait, these functions are contained in their own module called `walk_mut`. I also moved both visit and visit_mut modules to a super module called `visit` which shouldn't be really necessary but I find it more pleasant and have no strong preference if you want to revert it.

Dunqing:
Thank you for your efforts in accomplishing this amazing work. We are thinking about how VisitMut can be implemented better as it is crucial for us to implement the transformer. btw, @Boshen is on vacation, so he may not respond in time.

rzvxa:
> Thank you for your efforts in accomplishing this amazing work. We are thinking about how VisitMut can be implemented better as it is crucial for us to implement the transformer. btw, @Boshen is on vacation, so he may not respond in time.

Thanks for letting me know mate, I would also love to see your feedback if you have any on this and/or my other PR.

milesj:
Love this change.

I do have a suggestion though, maybe in a follow-up PR, but can we rename `AstKind` or `AstType`? They basically mean the same thing, and I'm constantly getting confused which is which.

rzvxa:
@milesj Hi, Thanks for the feedback.

I have to agree on the fact that having similar names may cause some confusion for users especially if they are new to this codebase. Do you have any suggestions for it? As it is right now they are both needed, AstKind is used for making the semantics in the Visit trait and AstType is a non-referencing version of it for VisitMut which has more limitations on passing references around. If @Boshen green lights it we may be able to do the same thing here for Visit but we are still going to need AstKind for our AstNode type.

I based this naming on something an old collage of mine taught me before, Basically, the conversion here is that `AstType` is an enum of all types of AST nodes, while `AstKind` is an enum that contains anything of the AST kind, so one is there to hold the Ast and other one is there to inform us only about the type.

rzvxa:
It is ready to merge again!

rzvxa:
@Boshen Sorry to ping you here, This PR moves visit traits to their own module. If it is something you are not found of just let me know to revert it before resolving the conflicts here.

Boshen:
I like how things are structured in `syn` https://docs.rs/syn/latest/syn/index.html#modules

It's all green light from me if the direction we are heading is "be like `syn`" 😅

rzvxa:
> I like how things are structured in syn https://docs.rs/syn/latest/syn/index.html#modules
>
> It's all green light from me if the direction we are heading is "be like syn" 😅 mod things so we can feature gate them to reduce a bit of compile time.

Well, I'm not familiar with how `syn` has structured its code but I gave it a cursory look and I get what you mean, We can definitely add features to selectively choose from toolchain modules.

I mostly moved visitors to their own module to have them in a separate directory in case we want to add things like `ScopeVisit` and other visitor traits. It shouldn't make any difference in how we want to feature gate them.

> Our largest consumer right now is Rolldown, I want to make it compile as fast as possible.

I'll take a look into how Rolldown is using this project to see what features can be feature-gated and which set of them can be grouped together.

> I'll give you write permission if you like this project, and intend to push things forward (especially regarding the AST and the transformer) 😁

Well, I'm grateful for the offer, I actually am interested in making the transformer feature complete(at least the `preset-env` part of it), and having the write permission for stuff that has tests; that won't need that much supervision can indeed smooth things out.
As for the AST and feature gating, it won't make that big of a difference since we almost always need to have you and/or other contributors review the changes.

-----------------

Let me know what's your decision about moving the visit traits.

Boshen:
> Let me know what's your decision about moving the visit traits.

👍

Boshen:
@rzvxa https://github.com/oxc-project/oxc/invitations

I gave you write permission so you get access to graphite. https://oxc-project.github.io/docs/contribute/rules.html

rzvxa:
> @rzvxa https://github.com/oxc-project/oxc/invitations
> 
> I gave you write permission so you get access to graphite. https://oxc-project.github.io/docs/contribute/rules.html

Well thanks, I haven't used graphite before so bear with me in my first few PRs till I get the hang of it.

PR #2779, opened 2024-03-21
refactor(sourcemap): enable_source_map using Arc<str> instead of String

underfin:
The `source_name` using `String` caused more memory allocate at rolldown, here using `Arc<str>` to avoid this.

### DIFF
--- a/crates/oxc_codegen/examples/codegen.rs
+++ b/crates/oxc_codegen/examples/codegen.rs
@@ -30,12 +30,12 @@ fn main() -> std::io::Result<()> {
 
     let options = CodegenOptions::default();
     let printed =
-        Codegen::<false>::new(&source_text, options.clone()).build(&ret.program).source_text;
+        Codegen::<false>::new("", &source_text, options.clone()).build(&ret.program).source_text;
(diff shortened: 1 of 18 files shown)


Boshen:
`Arc` feels so weird here, I wonder whether we can change the API and pass in a reference somewhere. `Codegen` doesn't really need to own the source map name ...

underfin:
I want to avoid `CodegenOptions` to `CodegenOptions<'a>`, as it will be caused `Codegen<'a>`.  What do you think about 
change `Codegen::new(source_text: &str, options: CodegenOptions)` to `Codegen::new(source_name &str, source_text: &str, options: CodegenOptions)`.

Boshen:
`Codegen::new(source_name &str, source_text: &str, options: CodegenOptions)` seems better, shall we try this one?

Should it be a filename or a file path though, feels like it should be a `&Path`?

underfin:
> Should it be a filename or a file path though, feels like it should be a &Path?

The `&str` is enough for it, the rolldown could cast it or other tools.

Dunqing (approved):

PR #2789, opened 2024-03-22
feat(linter): @typescript-eslint/prefer-for-of

charnog:


charnog:
The PR should have only the A-linter label. Others were added accidentally when I force-pushed the merged commit.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
```diff
@@ -0,0 +1,296 @@
...
+    ///   console.log(arr[i]);
+    /// }
+    ///
+    /// // Good
+    /// for (const item of arr) {
+    ///   console.log(item);
+    /// }
+    /// ```
+    PreferForOf,
+    correctness
```
mysteryven, line 110: Could you change it to [`style`](https://github.com/oxc-project/oxc/blob/4c5abb590ec214023c3ec5d43616afd8a85b5f6e/crates/oxc_cli/src/command/lint.rs#L83)?  It looks like the `prefer-*` rule more considered to be best practice.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
```diff
@@ -0,0 +1,296 @@
...
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("typescript-eslint(prefer-for-of): Expected a `for-of` loop instead of a `for` loop with this simple iteration.")]
+#[diagnostic(severity(warning), help("Consider using a for-of loop for this simple iteration."))]
+struct PreferForOfDiagnostic(#[label] pub Span);
+
+#[derive(Debug, Default, Clone)]
+pub struct PreferForOf;
+
+declare_oxc_lint!(
```
mysteryven, line 88: Can we put these main codes at the top of files?  It would provide a more clear overview of the rule's structure.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
```diff
@@ -0,0 +1,296 @@
...
+            match &com_mem_expr.object {
+                Expression::Identifier(id) => id.name.as_str() != array_name,
+                Expression::MemberExpression(mem_expr) => {
+                    matches!(&**mem_expr,
+                        MemberExpression::StaticMemberExpression(st_mem_expr)
+                        if st_mem_expr.property.name != array_name
+                    )
+                }
+                _ => true,
+            }
```
mysteryven, line 214: Can you explain why we need this part? Didn't find the corresponding part in [prefer-for-of](https://github.com/typescript-eslint/typescript-eslint/blob/04e32d6834c19160fb4f44f812a3e64671fca893/packages/eslint-plugin/src/rules/prefer-for-of.ts#L153-L154).🤔

And I didn't find our logic of checking [`ThisExpression`](https://github.com/typescript-eslint/typescript-eslint/blob/04e32d6834c19160fb4f44f812a3e64671fca893/packages/eslint-plugin/src/rules/prefer-for-of.ts#L117).

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
```diff
@@ -0,0 +1,296 @@
...
+            if test_and_update_span.contains(symbol_span) || !body_span.contains(symbol_span) {
+                return false;
+            }
+
+            let Some(ref_parent) = nodes.parent_node(ref_id) else { return true };
+            if matches!(nodes.parent_node(ref_parent.id()), Some(ref_grand_parent)
+                if is_assignee(ref_grand_parent))
+            {
+                return true;
+            }
```
mysteryven, line 197: Looks like we didn't do the same with the original [`isAssignee`](https://github.com/typescript-eslint/typescript-eslint/blob/04e32d6834c19160fb4f44f812a3e64671fca893/packages/eslint-plugin/src/util/isAssignee.ts#L13). (e.g. we didn't check it's on left side of `AssignmentExpression`). Do we need to keep the same with it?

mysteryven (changes requested):

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
```diff
@@ -0,0 +1,296 @@
...
+        let nodes = ctx.nodes();
+        let test_and_update_span = test_expr.span.merge(&update_expr.span());
+        let body_span = for_stmt.body.span();
+
+        let Some(var_symbol_id) = var_symbol_id else { return };
+        if ctx.semantic().symbol_references(var_symbol_id).any(|reference| {
+            let ref_id = reference.node_id();
+
+            let symbol_span = nodes.get_node(ref_id).kind().span();
+            if test_and_update_span.contains(symbol_span) || !body_span.contains(symbol_span) {
```
mysteryven, line 188: Why do we need to check `test_and_update_span`?

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
charnog: Done.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
charnog: Done.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
charnog: I can't. Probably something from the earlier versions. I deleted it.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
charnog: There is an `isIndexOnlyUsedWithArray` function. In the original, they are checking it as "if ALL of the references are used only with an array (from the `test` expression), then suggest 'prefer-for-of'".

The ported version uses reversed logic: "if ANY of the references are used not with the array (from the `test` expression), then skip the rule". In our case, returning `true` means that the reference "is bad".

```ts
const arrayText = context.sourceCode.getText(arrayExpression);
return indexVar.references.every(reference => {
  const id = reference.identifier;
  const node = id.parent;

  return (
    !contains(body, id) || (
        node.type === AST_NODE_TYPES.MemberExpression &&
        node.object.type !== AST_NODE_TYPES.ThisExpression &&
        node.property === id &&
        context.sourceCode.getText(node.object) === arrayText &&
        !isAssignee(node)
    )
  );
});

```

```rust
let parent_kind = ref_parent.kind();
let AstKind::MemberExpression(mem_expr) = parent_kind else { return true };
let MemberExpression::ComputedMemberExpression(com_mem_expr) = &mem_expr else {
    return true;
};

match &com_mem_expr.object {
    Expression::Identifier(id) => id.name.as_str() != array_name,
    Expression::MemberExpression(mem_expr) => {
        matches!(&**mem_expr,
            MemberExpression::StaticMemberExpression(st_mem_expr)
            if st_mem_expr.property.name != array_name
        )
    }

    // Probably a better approach...
    // Expression::MemberExpression(mem_expr) => match mem_expr.static_property_name() {
    //     Some(prop_name) => prop_name != array_name,
    //     None => true,
    // },

    // node.object.type !== AST_NODE_TYPES.ThisExpression from the original goes here
    _ => true,
}

```

However, I may be mistaken somewhere, so please don't take my words at face value.

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
charnog: Checked all the cases, and fortunately for us, it looks like the following:

| Code | Grand Parent Kind |
| - | - |
| `for (let i = 0; i < arr.length; i++) { delete arr[i]; }` | UnaryExpression |
| `for (let i = 0; i < arr.length; i++) { arr[i] = 1; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { arr[i] += 1; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { arr[i]++; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { --arr[i]; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { [arr[i]] = [0]; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { [...arr[i]] = [0]; }` | SimpleAssignmentTarget |
| `for (let i = 0; i < arr.length; i++) { ({ foo: arr[i] } = { foo: 0 }) }` | SimpleAssignmentTarget |

When we assign or update the value within the array, the grandparent of `i` is identified as a `SimpleAssignmentTarget`.

As an example, in the most common form:
- `arr[i] = 1;` is an `AssignmentExpression`.
- `arr[i]` in `assign_expr.left` is categorized under the `SimpleAssignmentTarget` kind, containing a `ComputedMemberExpression` with `object` being `"arr"` and `expression` being `"i"`.

Therefore, we do not need to check the `left` side since we are already positioned on that node (on the `left`). All that is required is to verify its type.

I'm gonna to update the code and completely remove the `is_assignee` function.

charnog:
Btw, we use `static_property_name()` while checking `MemberExpression`, that internally checks between `ComputedMemberExpression` and `StaticMemberExpression`. That means that Oxc version must work for `obj["arr"]` cases too.

Should we add cases for this?

mysteryven:
> Should we add cases for this?

It's OK, we don't need to add it :)

crates/oxc_linter/src/rules/typescript/prefer_for_of.rs
mysteryven: Nice approach!

mysteryven (approved):
Appreciate all your effort on this rule!

PR #2791, opened 2024-03-22
refactor(ast): add walk functions to Visit trait.

### LINKED ISSUE #2442: [request] Move `Visit` and `VisitMut` default implementations to another method
The `Visit` and `VisitMut` traits are great but there's 1 giant usability issue, and that is that each method defines its own implementation. 

Now why is this a problem? Well, if I implement my own visitor and override a method, I lost the default implementation, and there's _no easy way_ (besides dynamic dispatch below) to reuse the implementation, so I end up copying and pasting a bunch of code. For example, say I have my own visitor that needs to do something with `Program`, I would do this:

```rust
pub struct Example<'a>;

impl<'a> Visit<'a> for Example<'a> {
    fn visit_program(&mut self, program: &Program<'a>) {
        for stmt in &program.body {
            // Do something here
        }
    }
}
```

But I still want the visitor to work correctly, so then I have to copy the default code like so:

```rust
pub struct Example<'a>;

impl<'a> Visit<'a> for Example<'a> {
    fn visit_program(&mut self, program: &Program<'a>) {
        for stmt in &program.body {
            // Do something here
        }

        // Default
        let kind = AstKind::Program(self.alloc(program));
        self.enter_scope({
            let mut flags = ScopeFlags::Top;
            if program.is_strict() {
                flags |= ScopeFlags::StrictMode;
            }
            flags
        });
        self.enter_node(kind);
        for directive in &program.directives {
            self.visit_directive(directive);
        }
        self.visit_statements(&program.body);
        self.leave_node(kind);
        self.leave_scope();
    }
}
```

You can sort of work around this be using dynamic dispatch, like the following. 

```rust
pub struct Example<'a>;

impl<'a> Visit<'a> for Example<'a> {
    fn visit_program(&mut self, program: &Program<'a>) {
        for stmt in &program.body {
            // Do something here
        }

        // Default
        Visit::visit_program(self, program);
    }
}
```

This works for some methods, but causes warnings for others. A warning I see often is "function cannot return without recursing". I'm not sure if this actually causes problems though.

Regardless, I have a suggestion here. Instead of defining a default implementation for methods, why not define another method like `do_visit_program` that has the implementation, and then update `visit_program` to call `do_visit_program`. That way, when the method is overridden, we can still call it normally.

For example:

```rust
// Before
pub trait Visit<'a>: Sized {
    fn visit_program(&mut self, program: &Program<'a>) {
        let kind = AstKind::Program(self.alloc(program));
        self.enter_scope({
            let mut flags = ScopeFlags::Top;
            if program.is_strict() {
                flags |= ScopeFlags::StrictMode;
            }
            flags
        });
        self.enter_node(kind);
        for directive in &program.directives {
            self.visit_directive(directive);
        }
        self.visit_statements(&program.body);
        self.leave_node(kind);
        self.leave_scope();
    }
}

// After
pub trait Visit<'a>: Sized {
    fn do_visit_program(&mut self, program: &Program<'a>) {
        let kind = AstKind::Program(self.alloc(program));
        self.enter_scope({
            let mut flags = ScopeFlags::Top;
            if program.is_strict() {
                flags |= ScopeFlags::StrictMode;
            }
            flags
        });
        self.enter_node(kind);
        for directive in &program.directives {
            self.visit_directive(directive);
        }
        self.visit_statements(&program.body);
        self.leave_node(kind);
        self.leave_scope();
    }

    fn visit_program(&mut self, program: &Program<'a>) {
        self.do_visit_program(program);
    }
}
```

rzvxa:
closes #2442

### DIFF
--- a/crates/oxc_ast/src/lib.rs
+++ b/crates/oxc_ast/src/lib.rs
@@ -28,7 +28,7 @@ pub use crate::{
     ast_builder::AstBuilder,
     ast_kind::AstKind,
     trivia::{Comment, CommentKind, Trivias, TriviasMap},
-    visit::Visit,
+    visit::{walk, Visit},
     visit_mut::VisitMut,
 };
 
--- a/crates/oxc_ast/src/lib.rs
+++ b/crates/oxc_ast/src/lib.rs
@@ -28,7 +28,7 @@ pub use crate::{
     ast_builder::AstBuilder,
     ast_kind::{AstKind, AstType},
     trivia::{Comment, CommentKind, Trivias, TriviasMap},
-    visit::Visit,
+    visit::{walk, Visit},
     visit_mut::VisitMut,
 };
 
--- a/crates/oxc_ast/src/visit.rs
+++ b/crates/oxc_ast/src/visit.rs
@@ -1448,7 +1448,7 @@ pub mod walk {
         match arg {
             ArrayExpressionElement::SpreadElement(spread) => visitor.visit_spread_element(spread),
             ArrayExpressionElement::Expression(expr) => {
-                visitor.visit_expression_array_element(expr)
+                visitor.visit_expression_array_element(expr);
             }
             ArrayExpressionElement::Elision(span) => visitor.visit_elision(*span),
         }
--- a/crates/oxc_ast/src/visit.rs
+++ b/crates/oxc_ast/src/visit.rs
@@ -10,6 +10,8 @@ use oxc_syntax::scope::ScopeFlags;
 
 use crate::{ast::*, ast_kind::AstKind};
 
+use walk::*;
+
 /// Syntax tree traversal
 pub trait Visit<'a>: Sized {
     fn enter_node(&mut self, _kind: AstKind<'a>) {}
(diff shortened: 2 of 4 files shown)


rzvxa:
It should be ready to merge, I leave the `SemanticBuilder` to fully implement every visitor instead of using walk functions since it needs to do stuff in the middle of a walk operation.

I can also happily confirm that we can remove all 4 of our event functions(`enter_*`/`leave_*`) from the `Visit` and `VisitMut` traits, These are all trivial to implement with the help of walk abstraction and I would even help with third-party tools to migrate to this new pattern.

In case we want to keep the abstraction for scope events it can be done via a trait implementing `Visit` called `ScopeVisit` which could provide `enter_scope` and `leave_scope` methods.
As for the node events, We should really lean toward not passing any reference other than the ones given in the visit operation, We can achieve most of the things that we need these methods for via a simple type containing no reference.

I've already done this exact same approach for the `ModuleLexer`(which you can see in this PR), among others.

rzvxa:
@Boshen I didn't expect this to perform slower in any benchmark, maybe wait before merging it in? I can resolve conflicts in #2776 so we merge that one first. There might be a redundant walk in there, I want to check this one again.

rzvxa:
It is ready to merge again!

Boshen:
Thank you so much for the tremendous work. 

👍 for `ScopeVisit`.

I'm not sure about enter/leave node due to its convenience, but we can bikeshed a bit more.

PR #2792, opened 2024-03-23
fix(semantic): ModuleRecord's indirect_export_entires missing reexported imports

Dunqing:
local_name has `name` and `span`. We can't directly compare ExportEntry's local_name and ImportEntry's local_name. because the span is never equal.

### DIFF
--- a/crates/oxc_semantic/src/module_record/builder.rs
+++ b/crates/oxc_semantic/src/module_record/builder.rs
@@ -85,15 +85,14 @@ impl ModuleRecordBuilder {
         for ee in export_entries {
             // a. If ee.[[ModuleRequest]] is null, then
             if ee.module_request.is_none() {
-                let local_name = match &ee.local_name {
-                    ExportLocalName::Name(name) => Some(name),
+                let found_import_entry = match &ee.local_name {
+                    ExportLocalName::Name(name) => self
+                        .module_record
+                        .import_entries
+                        .iter()
+                        .find(|entry| entry.local_name.name() == name.name()),
                     _ => None,
                 };
-                let found_import_entry = self
-                    .module_record
-                    .import_entries
-                    .iter()
-                    .find(|import_entry| Some(&import_entry.local_name) == local_name);
                 match found_import_entry {
                     // i. If ee.[[LocalName]] is not an element of importedBoundNames, then
                     None => {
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#2792** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/module_record/mod.rs
```diff
@@ -244,4 +244,31 @@ mod module_record_tests {
...
+                import_name: ExportImportName::Name(NameSpan::new("x".into(), Span::new(9, 10))),
+                export_name: ExportExportName::Name(NameSpan::new("x".into(), Span::new(33, 34))),
+                local_name: ExportLocalName::Null,
+            }
+        );
+        assert_eq!(
+            module_record.indirect_export_entries[1],
+            ExportEntry {
+                module_request: Some(NameSpan::new("mod".into(), Span::new(57, 62))),
+                span: Span::new(0, 0),
```
Dunqing, line 267: I will send another PR to fix the incorrect span

crates/oxc_semantic/src/module_record/builder.rs
```diff
@@ -85,15 +85,14 @@ impl ModuleRecordBuilder {
...
+                        .iter()
+                        .find(|entry| entry.local_name.name() == name.name()),
                     _ => None,
                 };
-                let found_import_entry = self
-                    .module_record
-                    .import_entries
-                    .iter()
-                    .find(|import_entry| Some(&import_entry.local_name) == local_name);
                 match found_import_entry {
```
Dunqing, line 96: I think the performance improvement comes from here. ImportEntry is only looked for if `ee.local_name` is `ExportLocalName::Name`

Boshen:
Can you double check with the spec? 

https://tc39.es/ecma262/#sec-parsemodule step 10

Dunqing:
> Can you double check with the spec? I want to make sure that we are on the same page.
> 
> https://tc39.es/ecma262/#sec-parsemodule step 10

I have been checked, comments and behaviour consistent with the spec

PR #2794, opened 2024-03-23
refactor(transformer): pass options via context.

rzvxa:
With this PR all transformers would get their `AstBuilder` and options via `TransformerCtx`.


crates/oxc_transformer/src/typescript/mod.rs
```diff
@@ -30,19 +29,13 @@ pub struct TypeScript<'a> {
 }
 
 impl<'a> TypeScript<'a> {
-    pub fn new(
-        ast: Rc<AstBuilder<'a>>,
-        ctx: TransformerCtx<'a>,
-        verbatim_module_syntax: bool,
-        options: &TransformOptions,
-    ) -> Self {
+    pub fn new(ctx: TransformerCtx<'a>, verbatim_module_syntax: bool) -> Self {
```
rzvxa, line 32: Is there any real difference between `verbatim_module_syntax` and `only_remove_type_imports` in the `TypescriptOptions`? Either way, I think the `verbatim_module_syntax` parameter should be a field in `TypescriptOptions` not a parameter.

Dunqing: For now, `verbatim_module_syntax` doesn't do anything. I also don't see it documented in [@babel/plugin-transform-typescript](https://babeljs.io/docs/babel-plugin-transform-typescript#options). I agree that moving it into TypescriptOptions makes more sense

rzvxa: > I also don't see it documented in [@babel/plugin-transform-typescript](https://babeljs.io/docs/babel-plugin-transform-typescript#options).

It is not a babel option, It is from [typescript](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax). We can either pick a name and deserialize it from multiple names or have multiple fields and use a method to compute the value, Right now these two are used like this in the typescript transformer:

```rust
self.verbatim_module_syntax || self.options.only_remove_type_imports
```

But in some places, we are only checking one which makes me wonder if these are intentional or genuine mistakes.
Let me know what you think.

Dunqing: I think we can remove the `verbatim_module_syntax ` option.  https://github.com/babel/babel/issues/15493#issuecomment-1466453493. 

Since we are porting Babel, we should be as consistent as possible with Babel's plugin's options

rzvxa: Well I agree that we should aim for having similar options to Babel, But I can't find the `only_remove_type_impots` flag in Babel (it is the old name of verbatim in typescript 3.8, verbatim came in ts 5).

I think verbatim is too much of a big word for such a simple option But since it is the newer name for this option maybe we should use it?

For now, I move the flag to the typescript options since it is the part that we are sure about. As for the flag itself, We may need to discuss it with others to see which is the more popular one and Do we need to alias it or not.

Dunqing: > Well I agree that we should aim for having similar options to Babel, But I can't find the `only_remove_type_impots` flag in Babel (it is the old name of verbatim in typescript 3.8, verbatim came in ts 5).
> 

You can see https://babeljs.io/docs/babel-plugin-transform-typescript#onlyremovetypeimports

Dunqing (approved):
I like these changes.

PR #2795, opened 2024-03-23
feat(transformer): numeric separator plugin.

rzvxa:
[es2021 numeric separator](https://babeljs.io/docs/babel-plugin-transform-numeric-separator)


Dunqing:
You need to update the snapshot. Just run `just coverage`

rzvxa:
> You need to update the snapshot. Just run `just coverage`

Thanks, Yeah I'm aware of the snapshot being different. My issue with this one is that I'm only passing one of 2 tests and since it is such a small thing I can't put my finger on the thing that causes this issue, Have to do some tracing to figure it out.

```
  # All Passed:
-* babel-plugin-transform-numeric-separator
 * babel-plugin-transform-optional-catch-binding
 * babel-plugin-transform-json-strings
 * babel-plugin-transform-shorthand-properties
@@ -515,6 +514,9 @@ Passed: 352/1369
 # babel-plugin-transform-logical-assignment-operators (5/6)
 * logical-assignment/null-coalescing/input.js
 
+# babel-plugin-transform-numeric-separator (1/2)
+* used-with-transform-es2015-literals/input.js
+
 # babel-plugin-transform-export-namespace-from (0/4)
 * export-namespace/namespace-default/input.mjs
 * export-namespace/namespace-es6/input.mjs
 ```


It seems like the snapshot already has these tests passing, Since they were implemented in codegen, It is broken since I'm passing 1 of 2 instead of 2 of 2 so updating snapshots shouldn't be necessary when I get it working.

**Comment by rzvxa** [comment]
@Dunqing Do you have any idea about this? I'm generating the same output as Babel but the test fails. Is there anything that I'm missing?

I'm getting this:

```
Original:

1_000;
0xA0_B0_C0;
0b1010_0001_1000_0101;
0o0_6_6_6;
var o = 0o7; // octal integer literal
const u = "Hello\u{000A}\u{0009}!"; // unicode string literals, newline and tab


Transformed:

1000;
0xA0B0C0;
41349;
438;
```

and Babel produces this:

```
1000;
0xA0B0C0;
41349;
438;
```

Here is the config used for the babel:
```
{
  "plugins": [
    "@babel/plugin-transform-numeric-separator",
    "@babel/plugin-transform-literals"
  ]
}
```

Dunqing:
> @Dunqing Sorry to bother you, Do you have any idea about this? I'm generating the same output as Babel but the test fails. Is there anything that I'm missing?
> 
> I'm getting this:
> 
> ```
> Original:
> 
> 1_000;
> 0xA0_B0_C0;
> 0b1010_0001_1000_0101;
> 0o0_6_6_6;
> 
> 
> Transformed:
> 
> 1000;
> 0xA0B0C0;
> 41349;
> 438;
> ```
> 
> and Babel produces this:
> 
> ```
> 1000;
> 0xA0B0C0;
> 41349;
> 438;
> ```
> 
> Here is the config used for the babel:
> 
> ```
> {
>   "plugins": [
>     "@babel/plugin-transform-numeric-separator",
>     "@babel/plugin-transform-literals"
>   ]
> }
> ```

I will take a look soon

Dunqing:
There are two problems that need to be solved

1. The test lacks an output file, which is a babel problem https://github.com/babel/babel/pull/16381
2. Due to the lack of an output file, oxc transform_confromance will use the input file as the output file #2808

rzvxa:
@Dunqing Wow great work! It literally took me a day or so to look at the same code wondering why is it wrong, Never would've thought that the test case itself was incomplete.

I guess with this knowledge, We can think of this and #2797 as passing the tests since they are producing the same output as Babel.

crates/oxc_transformer/src/es2015/literals.rs
Dunqing, line 1: Looks like you need to revert the changes, You do the same thing in #2797

Dunqing:
I think we can merge this after you have cleaned up the code and fixed the conflict.

Dunqing (approved):

Dunqing:
Thank you!!

PR #2796, opened 2024-03-24
feat(transformer/typescript): remove `verbatim_module_syntax` option

Dunqing:
Remove `verbatim_module_syntax` option, Because Babel's [onlyRemoveTypeImports](https://babeljs.io/docs/babel-plugin-transform-typescript#onlyremovetypeimports) option same behavior with `verbatim_module_syntax` , You can see https://github.com/babel/babel/issues/15493#issuecomment-1466453493

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -104,8 +104,7 @@ impl<'a> Transformer<'a> {
         Self {
             ctx: ctx.clone(),
             decorators: Decorators::new(Rc::clone(&ast), ctx.clone(), &options),
-            // TODO: pass verbatim_module_syntax from user config
-            typescript: source_type.is_typescript().then(|| TypeScript::new(Rc::clone(&ast), ctx.clone(), false, &options)),
+            typescript: source_type.is_typescript().then(|| TypeScript::new(Rc::clone(&ast), ctx.clone(), &options)),
             regexp_flags: RegexpFlags::new(Rc::clone(&ast), &options),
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#2796** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
@rzvxa cc

rzvxa:
@Dunqing Well it is great, I think we should move on with this. It is always possible to add an alias to the cli so we can set this flag with either option.

rzvxa:
I'll resolve the discussion on #2794, I commented there before seeing this PR.

Dunqing:
> It is always possible to add an alias to the client so we can set this flag with either option.

Sorry, I don't get your mean. Can you explain this?

rzvxa:
> > It is always possible to add an alias to the client so we can set this flag with either option.
> 
> Sorry, I don't get your mean. Can you explain this?

Yeah sure, Sorry I didn't make my assumptions clear, Let me try to rephrase it.

Since `verbatim_module_syntax` and `only_remove_type_imports` are doing basically the same thing we can have an alias for them, So no matter which option is passed to the oxc we always set the `only_remove_type_imports` since it has an easier and more clear name in my opinion.

Remember that I'm not 100% sure these 2 are supposed to do the same thing since the `verbatim_module_syntax` can mean more things in the future.

PR #2807, opened 2024-03-25
feat(linter): eslint-plugin-jest/no-untyped-mock-factory

eryue0220:
Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-untyped-mock-factory.ts)


crates/oxc_linter/src/rules/jest/no_untyped_mock_factory.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// );
+    /// ```
+    ///
+    NoUntypedMockFactory,
+    correctness,
+);
+
+impl Rule for NoUntypedMockFactory {
+    fn run_once(&self, ctx: &LintContext<'_>) {
+        for possible_jest_node in &collect_possible_jest_call_node(ctx) {
```
mysteryven, line 99: Looks like we only run this rule on TS files.

```rust
if !ctx.source_type().is_typescript() {
      return;
}
```

mysteryven (changes requested):

crates/oxc_linter/src/rules/jest/no_untyped_mock_factory.rs
```diff
@@ -0,0 +1,410 @@
...
+            return;
+        };
+        let Argument::Expression(expr) = name_node else {
+            return;
+        };
+
+        if let Expression::StringLiteral(string_literal) = expr {
+            ctx.diagnostic_with_fix(
+                AddTypeParameterToModuleMockDiagnostic(
+                    string_literal.value.to_string(),
```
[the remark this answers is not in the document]
eryue0220: Done.

crates/oxc_linter/src/rules/jest/no_untyped_mock_factory.rs
eryue0220: Done.

mysteryven (approved):

PR #2815, opened 2024-03-26
feat(semantic/jsdoc): Add `Span` for JSDoc, JSDocTag

leaysgur:
- [x] Refactor comment trimming
- [x] Pass comment span to `JSDoc`
- [x] `BTreeMap<tag.span, JSDocTag>`
- [x] Consider `kind_span` => Use `tag.kind` span instead of `tag.span`
- [x] Wait upstream CI issue resolved?


leaysgur:
### Consider `kind_span`

In this case:

```js
/**
 * @kind1 bar
 * baz...
 * @kind2
 */
```

`tag.span` for `@kind1` is `@kind1 bar\n * baz...\n * `, for `@kind2` is `@kind2\n `.

It is too verbose and may not fit for diagnostic span.

So, 

- introduce `kind_span` as props?
- Via helper?
  - `kind_span.end = kind_span.start + 1 + u32::try_from(tag.kind.len()).unwrap_or_default();`
- Use `kind_span` instead of `tag.span` as default?

crates/oxc_semantic/src/jsdoc/parser/jsdoc.rs
```diff
@@ -1,29 +1,125 @@
...
+use oxc_span::Span;
 use std::cell::OnceCell;
+use std::collections::BTreeMap;
 
 #[derive(Debug, Clone)]
 pub struct JSDoc<'a> {
     raw: &'a str,
     /// Cached+parsed JSDoc comment and tags
-    cached: OnceCell<(String, Vec<JSDocTag<'a>>)>,
+    cached: OnceCell<(String, BTreeMap<Span, JSDocTag<'a>>)>,
```
Boshen, line 11: Do we need a `BTreeMap` here? If jsdocs are always parsed from top to bottom, then the `Vec` is already sorted.

crates/oxc_semantic/src/jsdoc/builder.rs
```diff
@@ -156,7 +156,8 @@ impl<'a> JSDocBuilder<'a> {
         }
 
         // Remove the very first `*`
-        Some(JSDoc::new(&comment_content[1..]))
+        let jsdoc_span = Span::new(comment_span.start + 1, comment_span.end);
+        Some(JSDoc::new(&comment_content[1..], jsdoc_span))
```
Boshen, line 160: pass in `comment_span` and compute `jsdoc_span` inside `JSDoc::new` instead? Feels weird to pass in a `jsdoc_span`.

crates/oxc_semantic/src/jsdoc/parser/jsdoc.rs
Boshen: `BtreeMap` is not very performant, we should try and avoid it.

leaysgur: Oh, I see.
My original motivation was just to manage `Span` outside of `JSDocTag`.

How about a `FxHashMap` instead?

crates/oxc_semantic/src/jsdoc/builder.rs
leaysgur: Since I was passing a string slice with the `*` removed to `JSDoc::new()`, I was also passing the `Span` with the `*` removed.

Do you mean both should be done on the `JSDoc` side? 👀

Boshen (approved):

PR #2814, opened 2024-03-26
ci: update babel version

rzvxa:
In my privius PR I changed this version in justfile It seems like the version used in CI is different from justfile.

PR #2813, opened 2024-03-25
ci(coverage): update babel version.

rzvxa:
This PR updates the babel submodule in the justfile to take advantage of [this PR](https://github.com/babel/babel/pull/16381). Related to #2795 and #2797.

PR #2817, opened 2024-03-26
chore: update snapshots

Dunqing:
Babel repo is updated in #2813 and #2814


rzvxa (approved):
Thank you, Sorry for the inconvenience. I really appreciate what you did both with figuring out the babel issue and resolving this.

Dunqing:
There is also another issue, when we update the commit hash of the submodule, we also need to manually execute `just submodule` to update our cloned repo to be consistent with the configuration in justfile. Otherwise, the snapshot will never be the same. Maybe we check if the commit hash is the same when testing execution

Dunqing:
### Merge activity

* **Mar 25, 11:52 PM EDT**: @Dunqing started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2817).

PR #2822, opened 2024-03-26
fix(transformer): optional-catch-binding unused variable side effect

rzvxa:
Before this PR for this given case:

```javascript
const _unused = "It's a lie, They gonna use me:(";
try {
    throw 0;
} catch {
  console.log(_unused);
}
```

We would've generated this:

```javascript
const _unused = "It's a lie, They gonna use me:(";
try {
    throw 0;
} catch (_unused) {
  console.log(_unused);
}
```

This is incorrect, This PR aims to use the `CreateVars` trait in order to ensure the variable uniqueness.

Now it would output this:

```javascript
const _unused = "It's a lie, They gonna use me:(";
try {
    throw 0;
} catch (_unused2) {
  console.log(_unused);
}
```

### DIFF
--- a/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
+++ b/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
@@ -223,7 +223,7 @@ impl<'a> ExponentiationOperator<'a> {
         expr: Expression<'a>,
         nodes: &mut Vec<'a, Expression<'a>>,
     ) -> Expression<'a> {
-        let ident = self.create_new_var(&expr);
+        let ident = self.create_new_var_with_expression(&expr);
         // Add new reference `_name = name` to nodes
         let target = self.ast.simple_assignment_target_identifier(ident.clone());
         let op = AssignmentOperator::Assign;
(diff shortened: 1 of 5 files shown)


rzvxa:
* **#2822** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2822?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
There are no snapshots updated?

rzvxa:
> There are no snapshots updated?

No the tests on eslint don't consider this so it was passing anyway.

Dunqing (approved):

rzvxa:
@Dunqing I can add the appropriate tests to the eslint if you find it relative to oxc.

Dunqing:
> @Dunqing I can add the appropriate tests to the eslint if you find it relative to oxc.

I prefer to add the test here. But this is babel repo 😅 https://github.com/babel/babel/blob/f856fc22d05bf886d1a58c5c91490e1af8f90c44/packages/babel-plugin-transform-optional-catch-binding/test/fixtures/optional-catch-bindings

rzvxa:
> I prefer to add the test here. But this is babel repo 😅 https://github.com/babel/babel/blob/f856fc22d05bf886d1a58c5c91490e1af8f90c44/packages/babel-plugin-transform-optional-catch-binding/test/fixtures/optional-catch-bindings

I agree, We may want to add our set of unit tests on top of esbuild compatibility checks. However, I would add a try-catch-shadow case there when I get the time. Even if we don't rely on it, It would be net positive.

PR #2825, opened 2024-03-26
feat: add oxc sourcemap crate

underfin:
The sourcemap implement port from [rust-sourcemap](https://github.com/getsentry/rust-sourcemap), but has some different with it.

- Encode sourcemap at parallel, including quote `sourceContent` and encode token to `vlq` mappings.
- Avoid `Sourcemap` some methods overhead, like `SourceMap::tokens()` caused extra overhead at common cases. Here using `SourceViewToken` to instead of it.


crates/oxc_sourcemap/src/decode.rs
```diff
@@ -0,0 +1,397 @@
...
+                if src == !0 { None } else { Some(src) },
+                if name == !0 { None } else { Some(name) },
+            ));
+        }
+    }
+
+    Ok(tokens)
+}
+
+const B64: [i8; 256] = [
```
Boshen, line 91: Can you format this by hand and add a rustfmt skip?

Boshen (commented):
This is really hard to review, can you comment in the places where we should take a look?

crates/oxc_sourcemap/src/decode.rs
underfin: Done

underfin:
> This is really hard to review, can you comment in the places where we should take a look?

Yeah. The code is larger to review, because the most code is port from `rust-sourcemap`. I only add some optimize to it. Maybe you can review `ConcatSourcemapBuilder.rs` or `encode.rs` at now, the changes of `rust-sourcemap` other files are small.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+    buf.push_str("\"names\":[");
+    buf.push_str(&sourcemap.names.iter().map(|x| format!("{x:?}")).collect::<Vec<_>>().join(","));
+    buf.push_str("],\"sources\":[");
+    buf.push_str(&sourcemap.sources.iter().map(|x| format!("{x:?}")).collect::<Vec<_>>().join(","));
+    // Quote `source_content` at parallel.
+    if let Some(source_contents) = &sourcemap.source_contents {
+        buf.push_str("],\"sourcesContent\":[");
+        buf.push_str(
+            &source_contents.par_iter().map(|x| format!("{x:?}")).collect::<Vec<_>>().join(","),
```
Boshen, line 24: This `par_iter` is probably redundant. Remember, it's costly to create threads.

Boshen (commented):
Need to run `just fmt`.

crates/oxc_sourcemap/src/encode.rs
underfin: It is expected for larger string, quote `source content` is very expensive.

underfin: image
I upload a profile. In a threejs10x build, its time percent is `23%`.

underfin: ref: esbuild sourcemap also do it. https://github.com/evanw/esbuild/blob/main/internal/bundler/bundler.go#L2997-L3045

Boshen (approved):

PR #2846, opened 2024-03-28
feat(transformer): implement @babel/plugin-transform-react-jsx-self

fireairforce:
implement: https://babeljs.io/docs/babel-plugin-transform-react-jsx-self

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -52,7 +52,7 @@ use crate::{
     es2021::{LogicalAssignmentOperators, NumericSeparator},
     es2022::ClassStaticBlock,
     es3::PropertyLiteral,
-    react_jsx::ReactJsx,
+    react_jsx::{JsxSelf, ReactJsx},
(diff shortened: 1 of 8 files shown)


crates/oxc_transformer/src/lib.rs
```diff
@@ -202,6 +201,10 @@ impl<'a> VisitMut<'a> for Transformer<'a> {
         self.decorators.as_mut().map(|t| t.transform_declaration(decl));
     }
 
+    fn visit_jsx_opening_element(&mut self, elem: &mut JSXOpeningElement<'a>) {
```
fireairforce, line 204: it seems that `visit_jsx_opening_element` will execute after `visit_expression`, so the `jsx_self` transformer will execute after `react_jsx` transformer, maybe i need to move  `self.jsx_self.as_mut().map(|t| t.transform_jsx_opening_element(elem));` into `visit_expression` ?

Boshen:
Thank you for working on this. I copied your implementation and merged them in https://github.com/oxc-project/oxc/pull/2946

Issue #2685, opened 2024-03-11
Combine `Statement` and `Declaration` types


overlookmotel:
Would it be a pain if we merged AST `Statement` and `Declaration` types into one?

Currently:

```rs
pub enum Statement<'a> {
    BlockStatement(Box<'a, BlockStatement<'a>>),
    /* ... */
    ModuleDeclaration(Box<'a, ModuleDeclaration<'a>>),

    Declaration(Declaration<'a>),
}

pub enum Declaration<'a> {
    VariableDeclaration(Box<'a, VariableDeclaration<'a>>),
    /* ... */
    TSImportEqualsDeclaration(Box<'a, TSImportEqualsDeclaration<'a>>),
}
```

Could we make it just 1 enum combining all the variants:

```rs
pub enum Statement<'a> {
    // From `Statement`
    BlockStatement(Box<'a, BlockStatement<'a>>),
    /* ... */
    ModuleDeclaration(Box<'a, ModuleDeclaration<'a>>),

    // From `Declaration`
    VariableDeclaration(Box<'a, VariableDeclaration<'a>>),
    /* ... */
    TSImportEqualsDeclaration(Box<'a, TSImportEqualsDeclaration<'a>>),
}
```

I'm asking because this would simplify #2409. It is still possible to make it work without this change, but it's a bit more brittle.

To explain:

For AST transfer, ideally we want all `enum`s to be `#[repr(u8)]` so the byte value for each option's discriminant can be specified explicitly. Then they're completely locked and you can read the bytes representing a `Statement` in memory, and reliably interpret them.

But in this case, the compiler is cleverly merging the discriminants for `Statement` and `Declaration` using niche optimization, which means `Statement` is 16 bytes. If we make `Statement` `#[repr(u8)]`, it disables the niche optimization, and `Statement` becomes 24 bytes. Not good!

There is a way around this without changing the types: Don't make `Statement` `#[repr(u8)]`, and the compiler in practice is predictable in what discriminants it assigns. But the compiler makes no promise not to change that at any minute without warning, so while I don't think likely that it *will* break, it *could* break at any time.

To get to the point... It'd be simpler to sidestep the issue entirely by folding `Declaration`'s variants into `Statement`. Question is how much of a pain would that be? So is it worth it?

[...]

PR #2847, opened 2024-03-28
refactor: add declaration variants to `Statement`

overlookmotel:
An alternative solution to #2685.

Declaration variants are added to `Statement`, but the `Declaration` type is **not** removed, so can still be used independently.

The discriminants for the 2 enums are aligned, so `Statement` and `Declaration` have identical memory representations, and one can be transmuted to the other at zero cost. Type sizes remain the same, and no effect on performance.

Obviously it'd be preferable to avoid this entirely, but I'm having a great deal of trouble handling the nested enums on #2457, and this would really help.

Not making this PR with intent of it being merged as is, but just as a demonstration of what I'm exploring. @Boshen would you consider something like this?

NB: This could be simplified, removing the unsafe code, if `Visit::visit_declaration` doesn't need to be its own function and could be broken up into `Visit::visit_var_declaration`, `Visit::visit_function_declaration` etc. That would also remove a branch and probably speed up the visitor a little.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1193,6 +1193,7 @@ pub struct ParenthesizedExpression<'a> {
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(untagged))]
+#[repr(u8)]
 pub enum Statement<'a> {
     // Statements
     BlockStatement(Box<'a, BlockStatement<'a>>),
@@ -1215,7 +1216,19 @@ pub enum Statement<'a> {
     WithStatement(Box<'a, WithStatement<'a>>),
 
     ModuleDeclaration(Box<'a, ModuleDeclaration<'a>>),
-    Declaration(Declaration<'a>),
+
+    // SAFETY: Discriminants and types here MUST match discriminants for same variants in `Declaration`,
+    // to allow transmuting between the two types
+    VariableDeclaration(Box<'a, VariableDeclaration<'a>>) = 32,
+    FunctionDeclaration(Box<'a, Function<'a>>) = 33,
+    ClassDeclaration(Box<'a, Class<'a>>) = 34,
+    UsingDeclaration(Box<'a, UsingDeclaration<'a>>) = 35,
+
+    TSTypeAliasDeclaration(Box<'a, TSTypeAliasDeclaration<'a>>) = 36,
+    TSInterfaceDeclaration(Box<'a, TSInterfaceDeclaration<'a>>) = 37,
+    TSEnumDeclaration(Box<'a, TSEnumDeclaration<'a>>) = 38,
+    TSModuleDeclaration(Box<'a, TSModuleDeclaration<'a>>) = 39,
+    TSImportEqualsDeclaration(Box<'a, TSImportEqualsDeclaration<'a>>) = 40,
 }
 
 impl<'a> Statement<'a> {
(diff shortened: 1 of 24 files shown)


overlookmotel:
* **#2847** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
Oh this is clever for an intermediate step. Last time I tried but gave up because there were too many code changes.

Let me make a release before merging this. And then we'll continue the refactor.

overlookmotel:
Thanks for coming back @Boshen.

Before we go ahead with this, can I just check you're really OK with it? The same nested enums problem also exists with other AST types, so this would be just the first of a series of PRs to do the same thing to each of them. Probably we shouldn't start down this path if you think it's going to get too much by the end.

I've started with the big one - `Statement`. All the other affected types are more rarely used, so changes will be lower impact (e.g. `ExportDefaultDeclarationKind`, `AssignmentTargetMaybeDefault`). But still, there'll be a good number of them - maybe 10 in total.

On the positive side, in a few cases we'll be able to reduce the size of the types by squashing 2 discriminants into 1.

2nd question: Does `Visit::visit_declaration` need to exist?

If we can get rid of it, we can make the changes in this PR much simpler, and remove the unsafe + transmuting.

Removing it wouldn't cause any problems within Oxc itself, but question is do external consumers need `visit_declaration` to exist in order to be able to override it in their own `Visit` impl? (they could still get whatever they need to do done by instead implementing `visit_var_declaration`, `visit_function_declaration` etc instead, but I don't know how `Visit` is used "in the wild", so I don't know if that'd be very unergonomic for users)

Boshen:
> Before we go ahead with this, can I just check you're really OK with it? 

It's ok.

> But still, there'll be a good number of them - maybe 10 in total.

I think it's a case by case analysis.

> Does Visit::visit_declaration need to exist?

Let's do the removal in another PR. Most of the usages are in the transformer, but we plan to nuke the current transformer, so we'll have less code to change.

overlookmotel:
> I think it's a case by case analysis.

It will only solve my problem on #2457 if we change them all, and make it a rule for AST types that there can be no nested enums. If we're not going to do them all, there isn't much point in doing any of them (at least for my purposes in #2457).

---

The complete list of nested enums at present:

`Statement` -> `Declaration`

`ArrayExpressionElement` -> `Expression`
`PropertyKey` -> `Expression`
`Argument` -> `Expression`
`JSXExpression` -> `Expression`
`ForStatementInit` -> `Expression`
`ExportDefaultDeclarationKind` -> `Expression`
`TSEnumMemberName` -> `Expression`

`ForStatementLeft` -> `AssignmentTarget`
`AssignmentTargetMaybeDefault` -> `AssignmentTarget`
`AssignmentTarget` -> `SimpleAssignmentTarget`
`AssignmentTarget` -> `AssignmentTargetPattern`

`TSTupleElement` -> `TSType`
`TSTypeQueryExprName` -> `TSTypeName`
`TSModuleReference` -> `TSTypeName`

It's worse than I thought. Inlining all `Expression`s variants into `ArrayExpressionElement` and 6 other types seems impractical - at least without using a macro, and I know how you feel about that!

Let me think about it a bit more...

---

Going in the opposite direction, these enum variants could be unboxed to reduce indirection:

`Statement::ModuleDeclaration`: `Box<ModuleDeclaration>`
`Expression::MemberExpression`: `Box<MemberExpression>`
`SimpleAssignmentTarget::MemberAssignmentTarget`: `Box<MemberExpression>`
`ChainElement::MemberExpression`: `Box<MemberExpression>`

(side note: one nice thing about the types schema which #2457 produces is that it's really easy to run these kind of queries on the type definitions)

Boshen:
This is going to be impractical if we are going to bend the AST just for the sake of AST transfer 😢

overlookmotel:
Well I had thought it's worth bending it *to a degree*, as in my view AST transfer is a pretty worthwhile feature (and as I said at the start, there would likely be some trade-offs involved to make it work). But, yes, probably this is too much.

Let me try again and see if I can find another solution to the problem of calculating niches for nested enums. Can we leave this PR open for now until I reach a conclusion?

It's annoying that Rust doesn't offer type introspection itself. [It was going to](https://soasis.org/posts/a-mirror-for-rust-a-plan-for-generic-compile-time-introspection-in-rust/) but that effort fell apart due to [a personal feud](https://soasis.org/posts/statement-on-rustconf-compile-time-introspection/) with/within the Rust Project.

Boshen:
> Can we leave this PR open for now until I reach a conclusion?

Sure

> It's annoying that Rust doesn't offer type introspection itself. [It was going to](https://soasis.org/posts/a-mirror-for-rust-a-plan-for-generic-compile-time-introspection-in-rust/) but that effort fell apart due to [a personal feud](https://soasis.org/posts/statement-on-rustconf-compile-time-introspection/) with/within the Rust Project.

I remember the drama ...

overlookmotel:
Superceded by #3115.

PR #2851, opened 2024-03-29
refactor: arena sourcemap

underfin:


crates/oxc_codegen/src/sourcemap_builder.rs
```diff
@@ -22,21 +23,21 @@ pub struct LineOffsetTable {
...
-    sourcemap_builder: oxc_sourcemap::SourceMapBuilder,
+    sourcemap_builder: oxc_sourcemap::SourceMapBuilder<'a>,
     generated_line: u32,
     generated_column: u32,
 }
 
-impl Default for SourcemapBuilder {
-    fn default() -> Self {
+impl<'a> SourcemapBuilder<'a> {
+    pub fn new (allocator: &'a Allocator) -> Self {
```
Boshen, line 40: You don't need to pass in the allocator, the data is self contained.

I think you can fix this by wrapping the data structures.

underfin:
Have a try at the local benchmark, it caused performance regression.

PR #2861, opened 2024-03-29
feat(ast): add `AstNode` trait and provide access to `AstNodeId`

### LINKED ISSUE #2818: Add `ast_node_id` field to all ast nodes
We should have a field similar to `IdentifierReference`'s `reference_id` that would keep the `AstNodeId` of each AST node.

Can be written manually, using a declarative `ast_node` macro for definition or a procedural derive macro.
I guess we can also remove `AstNode` as it can be changed into SOA in the `AstNodes` structure.
If we remove the `AstNode` struct we can add a trait with the same name which would implement a function called `ast_node_id` that would give us the information for getting things that `AstNode` would've provided before. With this trait, we can use any ast node with the help of generics since with the ast_node_id alone we can infer the node's kind, parents, scope, control flow, and anything that gets added down the road.

rzvxa:
closes #2818


rzvxa:
* **#2861** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2861?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
In my view `AstNodeId` should be a non-zero type e.g. `NonZeroUsize`. Then `AstNodeIdContainer` will be half the size.

```rs
std::mem::size_of::<Cell<Option<usize>>>() == 16
std::mem::size_of::<Cell<Option<NonZeroUsize>>>() == 8
```

This would likely reduce the performance reduction visible on this PR at present.

You'd just need to initialize `AstNodes::nodes` and `AstNodes::parent_ids` with dummy entries to occupy index 0, and enforce that that dummy entry can never be popped off again (maybe wrap both `Vec`s to make them append-only).

Also, in my view `AstNodeId` could be a `u32` (by which I mean `NonZeroU32`) for the reasons mentioned in https://github.com/oxc-project/oxc/pull/2876#issuecomment-2032987405. But that's a separate argument from what I'm saying here.

overlookmotel:
Let's keep this branch live. We may want to come back to this later.

PR #2866, opened 2024-03-30
feat(transformer): add `transform-typescript` boilerplate

Boshen:


Boshen:
* **#2866** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
A plugin can be turned off, we haven't decided which one is better:

```rust
pub struct Transformer<'a> {
    typescript: Option<TypeScript>,
}
```

or 

```rust
pub struct TransformOptions {
    pub typescript: Option<TypeScriptOptions>,
}
```

A `None` indicates the plugin is turned off, both methods have tradeoffs, we'll bike-shed this later.

rzvxa:
> A plugin can be turned off, we haven't decided which one is better:
> 
> ```rust
> pub struct Transformer<'a> {
>     typescript: Option<TypeScript>,
> }
> ```
> 
> or
> 
> ```rust
> #[derive(Debug, Default, Clone)]
> pub struct TransformOptions {
>     pub typescript: Option<TypeScriptOptions>,
> }
> ```
> 
> A `None` indicates the plugin is turned off, both methods have tradeoffs, we'll bike-shed this later.

Aren't they both needed? A `None` option means the plugin is not turned on and would result in a `None` transformer.

Boshen:
> Aren't they both needed? A `None` option means the plugin is not turned on and would result in a `None` transformer.

True. I was thinking about eliminating `if let Some(plugin) = &self.plugin` and hide this check inside each individual plugin.

I'll do some follow up PRs.

rzvxa:
Well, I guess we can have the plugin options set to default when it's disabled but It means that we need to have some other way to configure our list of plugins. What is your alternative for it? When fetching the options to initialize the transformer we need to return None anyway in order to notify the Transformer to not create this plugin.

rzvxa:
> True. I was thinking about eliminating if let Some(plugin) = &self.plugin and hide this check inside each individual plugin.

You mean when calling the transform methods inside our main Transformer?

Boshen:
The previous version had something like this all over the place

```rust
  self.es2021_logical_assignment_operators.as_mut().map(|t| t.add_vars_to_statements(stmts));
  self.es2020_nullish_coalescing_operators.as_mut().map(|t| t.add_vars_to_statements(stmts));
  self.es2016_exponentiation_operator.as_mut().map(|t| t.add_vars_to_statements(stmts));
  self.es2015_arrow_functions.as_mut().map(|t| t.transform_statements(stmts));
```

I was thinking about changing it to 

```
self.es2021_logical_assignment_operators.add_vars_to_statements(stmts);
```

And hide the `Option` check inside the `add_vars_to_statements` function.

This seems like a burden to plugin authors, not a really good idea.

rzvxa:
Yes, we have to wait to make the call on this one until the mutability is figured out, Since as we discussed in [here](https://github.com/oxc-project/oxc/discussions/2704#discussioncomment-8946163) we might need a macro for applying the transformation which can also be used to hide these checks.

Dunqing (approved):

Boshen:
### Merge activity

* **Mar 30, 8:48 AM EDT**: @Boshen merged this pull request with [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/2866).

Issue #2828, opened 2024-03-26
Bug: Can not use `@oxc-parser/wasm` without error


Saeid-Za:
Hello ! I've been trying to use `@oxc-parser/wasm` but there is an error.

```
Error: Cannot find package 'test/node_modules/@oxc-parser/wasm/package.json' imported from 'test/a.mjs'
```

To reproduce the error:

1. run `npm i @oxc-parser/wasm` in an empty folder
2. create a `test.mjs` file, having this content:

```js
import {parseSync} from "@oxc-parser/wasm"
console.log(parseSync)
```

3. Try to run the code.

Boshen:
The state of wasm build is really confusing.

As stated in the README

> This package is built with wasm-pack build --release --target web for bundler (webpack / vite) consumption. Checkout [oxc-parser](https://www.npmjs.com/package/oxc-parser) for usage in node.js.

The original issue is about making the napi version work on stackbliz, pinging @Brooooooklyn who may know more.

Boshen:
The original issue is about making it work on stackbliz, pinging @Brooooooklyn who may know more.

Boshen:
We may need to publish more builds (wasi, node.js target) in the near future, but I'll close this for now as I don't have the bandwidth.

Boshen:
Closed by #2916

Released v0.1.0 https://www.npmjs.com/package/@oxc-parser/wasm

PR #2871, opened 2024-03-30
Adding more `wasm-parser` builds

Saeid-Za:
I've added `bundler` and `node` target builds and added code samples to showcase the usage.

Related Issue: [Bug: Can not use @oxc-parser/wasm without error](https://github.com/oxc-project/oxc/issues/2828)


Saeid-Za:
We could further decrease the package size, by shipping only one wasm file, which i think could be shared between builds

Boshen:
This is amazing! I'll take a look at this tomorrow and tinker with the release.

Saeid-Za:
Is there anything I could help with?

Boshen:
Sorry it'll take a while to go through a huge PR like this, and get to this when I get the time.

Saeid-Za:
Oh, sure, please take your time
If there is anything I could help with, don't hesitate to mention me !

wasm/parser/scripts/build.ts
```diff
@@ -0,0 +1,43 @@
...
+const outDir = resolve(join(__dirname, "../../../npm/wasm-parser"))
+
+async function main() {
+    consola.start("Building WASM target ...");
+
+    await rm(outDir, { recursive: true, force: true })
+    await mkdir(outDir, { recursive: true })
+
+    await Promise.all([
+        buildProject("bundler", join(outDir, "bundler")),
```
Boshen, line 18: Why is this changed from `web` to `bundler`?

Boshen:
Closed by #2916

Released v0.1.0 https://www.npmjs.com/package/@oxc-parser/wasm

Saeid-Za:
Regarding why build targets `bundler` instead of `web`, most users use a bundler like vite or webpack as their primary tooling in front-end development.
The bundler target, supports both these tools. (Hence the vue/vite demo in the PR), and the `wasm-pack` itself, defaults to it because the majority use these toolings.
The web target, requires manual initiation and I think I couldn't make it work with vite (don't remember why though).

PR #2885, opened 2024-04-02
feat(linter): typescript-eslint/consistent-type-definitions

todor-a:


crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
```diff
@@ -0,0 +1,214 @@
...
+    ///
+    ///
+    /// ### Why is this bad?
+    ///
+    ///
+    /// ### Example
+    /// ```javascript
+    /// ```
+    ConsistentTypeDefinitions,
+    correctness
```
Boshen, line 46: This is a `style` rule from the docs https://typescript-eslint.io/rules/consistent-type-definitions/

todor-a: my bad, fixed

todor-a:
Not sure what causes the performance regression 🤔

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
```diff
@@ -0,0 +1,214 @@
...
+            "
+            export type W<T> = {
+              x: T;
+            };
+                  ",
+            Some(serde_json::json!(["type"])),
+        ),
+    ];
+
+    let fail = vec![
```
mysteryven, line 127: Our rulegen didn't generate case surrounded by [`noFormat`](https://github.com/typescript-eslint/typescript-eslint/blob/fde44275117b3a7bbdcaaf267c73986b6637c5f9/packages/eslint-plugin/tests/rules/consistent-type-definitions.test.ts#L66), it will be [fixed](https://github.com/oxc-project/oxc/pull/2903)! Could you update the test cases?

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
```diff
@@ -0,0 +1,214 @@
...
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    ///
+    /// ### Why is this bad?
+    ///
+    ///
+    /// ### Example
+    /// ```javascript
+    /// ```
```
mysteryven, line 44: Shall we fill this part like [other rules](https://github.com/oxc-project/oxc/blob/e898918aba3c1cab6ea209568da2b9d879c31b40/crates/oxc_linter/src/rules/typescript/adjacent_overload_signatures.rs#L31-L79)?

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
```diff
@@ -0,0 +1,214 @@
...
+        Self { config }
+    }
+
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::TSTypeAliasDeclaration(typ) => match &typ.type_annotation {
+                TSType::TSTypeLiteral(lit)
+                    if self.config == ConsistentTypeDefinitionsConfig::Interface =>
+                {
+                    ctx.diagnostic(ConsistentTypeDefinitionsDiagnostic("interface", "type", lit.span))
```
mysteryven, line 67: There is [fixer](https://github.com/typescript-eslint/typescript-eslint/blob/fde44275117b3a7bbdcaaf267c73986b6637c5f9/packages/eslint-plugin/src/rules/consistent-type-definitions.ts#L54-L82) for this rule, are you having interest to achieving it too :)

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
```diff
@@ -0,0 +1,214 @@
...
+                }
+                _ => {}
+            },
+
+            AstKind::ExportDefaultDeclaration(exp) => match &exp.declaration {
+                ExportDefaultDeclarationKind::TSInterfaceDeclaration(decl) => ctx.diagnostic(
```
mysteryven, line 73: Need to check `if self.config == ConsistentTypeDefinitionsConfig::Type` here?

crates/oxc_linter/src/snapshots/consistent_type_definitions.snap
```diff
@@ -0,0 +1,104 @@
...
+ 4 │ ├─▶         }
+   · ╰──── Use an `type` instead of a `interface`
+ 5 │                   
+   ╰────
+  help: Use an `type` instead of a `interface`
+
+  ⚠ typescript-eslint(consistent-type-definitions):
+   ╭─[consistent_type_definitions.tsx:3:11]
+ 2 │             namespace JSX {
+ 3 │ ╭─▶           interface Array<T> {
```
mysteryven, line 30: Shall we only point the keyword `interface` or `type` in to make the error more specific?

mysteryven (changes requested):

mysteryven:
You can run `just ready` to help you fix the CI in local.

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
todor-a: done, let me know if it looks okay

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
todor-a: done, thank you. I just copied the description from https://typescript-eslint.io/rules/consistent-type-definitions, hope that's okay

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
todor-a: it makes sense to need the check but the tests do not catch its absence. anyway, i've added it, thanks 🙌

crates/oxc_linter/src/snapshots/consistent_type_definitions.snap
todor-a: Some of the hints pointed to the entire expression, when the error is for a single line type, and to the stuff between the brackets, when multiline. I've updated it to always point to the entire expression in the name of consistency.

I have no strong preference where the error should point tho (and I couldn't find how to get the `span` of the `interface`/`type` keyword).

Interested to hear what you think :)

mysteryven: For `TSInterfaceDeclaration`, the keyword `interface`'s span is from `TSInterfaceDeclaration.start` to `TSInterfaceDeclaration.start + 9` (9 is the length of 'interface'), `type` in `TSTypeAliasDeclaration` is the same as it.

For `ExportDefaultDeclaration`, looks like we can iterate source text from `ExportDefaultDeclaration.start`, and find the target keyword? 

```rust
let slice = ctx.semantic().source_text()[(start.start as usize)..].chars();
// ...
```

crates/oxc_linter/src/snapshots/consistent_type_definitions.snap
```diff
@@ -0,0 +1,168 @@
...
+   ╰────
+  help: Use an `interface` instead of a `type`
+
+  ⚠ typescript-eslint(consistent-type-definitions):
+   ╭─[consistent_type_definitions.tsx:1:33]
+ 1 │ type T=                         { x: number; };
+   ·                                 ───────┬──────
+   ·                                        ╰── Use an `interface` instead of a `type`
+   ╰────
+  help: Use an `interface` instead of a `type`
```
mysteryven, line 27: It points to `typeAnnotation`, which may not helpful for this error.

crates/oxc_linter/src/rules/typescript/consistent_type_definitions.rs
mysteryven: Checked, it has been added.

crates/oxc_linter/src/snapshots/consistent_type_definitions.snap
todor-a: makes sense, will take a look how to fix it

crates/oxc_linter/src/snapshots/consistent_type_definitions.snap
todor-a: just learned something new, thanks!

mysteryven (approved):
Thank you! 

PS: I only fixed a bit Clippy's error.

todor-a:
> Thank you!
> 
> PS: I only fixed a bit Clippy's error.

I was planning to work on the auto fix today after work. Will try to submit it in another PR.

Thanks for your help with this.

PR #2889, opened 2024-04-02
fix(sourcemap): using serde_json::to_string to quote sourcemap string

underfin:


crates/oxc_sourcemap/src/sourcemap.rs
```diff
@@ -17,7 +17,7 @@ pub struct SourceMap {
     pub(crate) token_chunks: Option<Vec<TokenChunk>>,
 }
 
-#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::cast_possible_truncation, clippy::missing_errors_doc)]
```
Boshen, line 20: Let's document the error properly.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -14,20 +15,35 @@ pub fn encode(sourcemap: &SourceMap) -> String {
...
+        .map(|x| serde_json::to_string(x.as_ref()))
+        .collect::<std::result::Result<Vec<_>, serde_json::Error>>()
+        .map_err(Error::from)?;
+    buf.push_str(&names.join(","));
     buf.push_str("],\"sources\":[");
-    buf.push_str(&sourcemap.sources.iter().map(|x| format!("{x:?}")).collect::<Vec<_>>().join(","));
+    let sources = sourcemap
+        .sources
+        .iter()
+        .map(|x| serde_json::to_string(x.as_ref()))
```
Boshen, line 29: I think we should document why we need `serde_json::to_string`, and we should have a test case for this at least.

Boshen (commented):
What's the reason for the performance improvement?

crates/oxc_sourcemap/src/sourcemap.rs
underfin: Done

crates/oxc_sourcemap/src/encode.rs
underfin: Done

underfin:
> What's the reason for the performance improvement?

I'm not found other different, here look like has a fast lookup table, https://github.com/serde-rs/json/commit/fc436ebd73e5bcafa24b34b3a3aeb44eede212a4

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -125,14 +143,24 @@ fn encode_vlq(out: &mut String, num: i64) {
...
 
     for (tok1, tok2) in sm.get_tokens().zip(sm2.get_tokens()) {
         assert_eq!(tok1, tok2);
     }
 }
+
+#[test]
+fn test_encode_escape_string() {
+    // '\0' should be escaped.
+    let sm = SourceMap::new(None, vec!["\0a.js".into()], vec![], None, vec![], None);
```
Boshen, line 161: You mentioned "Here using `serde_json::to_string` to serialization `names/source_contents/sources`."

I think the test case should cover all 3.

Boshen (commented):
I wonder whether we should bubble up the error or consume the error by replacing it with a placeholder such as "UnableAbleToSerializeString" into it, it feels weird that the consumer have to handle this error.

underfin:
> I wonder whether we should bubble up the error or consume the error by replacing it with a placeholder such as "UnableAbleToSerializeString" into it, it feels weird that the consumer have to handle this error.

The `Error::BadJson(serde_json::Error)` already exported it.

Boshen:
> > I wonder whether we should bubble up the error or consume the error by replacing it with a placeholder such as "UnableAbleToSerializeString" into it, it feels weird that the consumer have to handle this error.
> 
> The `Error::BadJson(serde_json::Error)` already exported it.

I mean the user of this crate need to handle this error, but I think they shouldn't.

underfin:
Actually, the `serde_json` error does not help to resolve the issue, the message is not clear to json. I using the editor to see json problems.

underfin:
> I mean the user of this crate need to handle this error, but I think they shouldn't.

It's hard to say, but the rolldown is already to handle this correctly.

Boshen (approved):

PR #2892, opened 2024-04-02
[WIP] Add `Result` type to `VisitMut`

milesj:


crates/oxc_transformer/src/visit_utils.rs
```diff
@@ -0,0 +1,61 @@
+use oxc_ast::ast::{Expression, Statement};
+use oxc_ast::VisitResult;
+
+#[derive(Default)]
+pub struct TransformResult<'a> {
```
rzvxa, line 5: I think we should make this into an enum for example

```rust
enum TransformerResult<'a> {
    Pass,
    Append(Vec<'a, TransformReplaceKind<'a>>,
    Replace(Vec<'a, TransformReplaceKind<'a>>,
}
```

crates/oxc_ast/src/visit/visit_mut.rs
```diff
@@ -9,823 +9,948 @@ use crate::{ast::*, AstType};
 #[allow(clippy::wildcard_imports)]
 use self::walk_mut::*;
 
+pub trait VisitResult {
```
rzvxa, line 12: I'm not sure if this pattern would work, We can have much better control if we use default and many. Where we want to visit a list we need a way to construct a single result from all of those results in the list.

This way we can leave those untouched in implementations other than Transformer, It is the same for BasicVisitResult, if instead of using a concrete type we just implement `VisitResult` for unit (`()`) we can have methods that don't need to explicitly return.

crates/oxc_transformer/src/visit_utils.rs
rzvxa: Maybe append isn't needed at all, Since replacing something with a list containing that same node would cost us cloning we might consider adding something like append if we find a usecase for it

PR #2893, opened 2024-04-03
feat(linter): Implement jsdoc/empty-tags

leaysgur:
Part of #1170 

- https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/empty-tags.md
- https://github.com/gajus/eslint-plugin-jsdoc/blob/main/src/rules/emptyTags.js

### DIFF
--- a/crates/oxc_linter/src/rules/jsdoc/empty_tags.rs
+++ b/crates/oxc_linter/src/rules/jsdoc/empty_tags.rs
@@ -1,0 +1,351 @@
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use phf::phf_set;
+use serde::Deserialize;
+
+use crate::{context::LintContext, rule::Rule};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint-plugin-jsdoc(empty-tags): Expects the void tags to be empty of any content.")]
+#[diagnostic(severity(warning), help("`@{1}` tag should be empty."))]
+struct EmptyTagsDiagnostic(#[label] Span, String);
+
+#[derive(Debug, Default, Clone)]
+pub struct EmptyTags(Box<EmptyTagsConfig>);
+
+declare_oxc_lint!(
+    /// ### What it does
+    /// Expects the following tags to be empty of any content:
+    /// - `@abstract`
+    /// - `@async`
+    /// - `@generator`
+    /// - `@global`
+    /// - `@hideconstructor`
+    /// - `@ignore`
+    /// - `@inner`
+    /// - `@instance`
+    /// - `@override`
+    /// - `@readonly`
+    /// - `@inheritDoc`
+    /// - `@internal`
+    /// - `@overload`
+    /// - `@package`
+    /// - `@private`
+    /// - `@protected`
+    /// - `@public`
+    /// - `@static`
+    ///
+    /// ### Why is this bad?
+    /// The void tags should be empty.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// // Passing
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/src/rules/jsdoc/empty_tags.rs
```diff
@@ -0,0 +1,351 @@
...
+    ///
+    /// /** @private */
+    ///
+    /// // Failing
+    /// /** @async foo */
+    ///
+    /// /** @private bar */
+    /// ```
+    EmptyTags,
+    restriction
```
Boshen, line 58: I'd go for `correctness` because the text is useless.

But we need to add a `--jsdoc-plugin` flag cli first.

crates/oxc_linter/src/snapshots/empty_tags.snap
```diff
@@ -0,0 +1,75 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: empty_tags
+---
+  ⚠ eslint-plugin-jsdoc(empty-tags): Expects the void tags to be empty of any content.
+   ╭─[empty_tags.tsx:3:17]
+ 2 │                       /**
+ 3 │                        * @abstract extra text
+   ·                          ─────────
```
Boshen, line 9: The error message emphasizes "empty text" so I would put the span on the extra text.

crates/oxc_linter/src/snapshots/empty_tags.snap
```diff
@@ -0,0 +1,75 @@
...
+ 4 │                        */
+   ╰────
+  help: `@abstract` tag should be empty.
```
Boshen, line 12: "The tag should be empty" is confusing. Maybe something like "the description text will not show up in jsdoc?" to convey it's useless What the proper term for the text anyway?

crates/oxc_linter/src/rules/jsdoc/empty_tags.rs
leaysgur: > I'd go for correctness 

OK, I'll fix the category.

- - - 

> we need to add a --jsdoc-plugin flag cli first.

For this,

> I'll send a PR to make this plugin public after confirming that a few more rules can be implemented without any problems.

Please see also https://github.com/oxc-project/oxc/pull/2642

crates/oxc_linter/src/snapshots/empty_tags.snap
leaysgur: > https://github.com/oxc-project/oxc/pull/2815#issuecomment-2019347874

Let me reconsider this a bit...

crates/oxc_linter/src/snapshots/empty_tags.snap
leaysgur: > "the description text will not show up in jsdoc?"

This may be tricky topic. 😓 

- Original JSDoc.app does not expect THIS for `@async` and may just ignore in its implementation?
  - https://jsdoc.app/tags-async
  - https://github.com/jsdoc/jsdoc/blob/7565dfd53cb5d86e334a9a3fa1193c767a31a0d7/packages/jsdoc-tag/lib/definitions/core.js#L59
- But TypeScript renders THIS( and seems to accept THIS everywhere)
  - image
  - image

> What the proper term for the text anyway?

I call THIS as "comment part for tag" or just "tag comment" since TS defined THIS in `.comment` property.

Original JSDoc seems to use the term "value"?, `jsdoccomment`(and `comment-parser`) using the term "description".

Boshen: Tag comment sounds good to me.

If there are things can render this comment ... then we need to put this into the restriction or pedantic category.

Boshen (approved):
Let's merge and get the momentum going.

PR #2895, opened 2024-04-03
fix(codegen): avoid unsafe code search original name

### LINKED ISSUE #716: ci: crash on windows
Here is error.
```
process didn't exit successfully: `D:\a\rolldown\rolldown\target\debug\deps\esbuild-1ce2e539eae6e9f7.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

https://github.com/rolldown/rolldown/actions/runs/8505493287/job/23294065477

underfin:
close https://github.com/rolldown/rolldown/issues/716

### DIFF
--- a/crates/oxc_codegen/src/sourcemap_builder.rs
+++ b/crates/oxc_codegen/src/sourcemap_builder.rs
@@ -62,12 +62,12 @@ impl SourcemapBuilder {
     }
 
     pub fn add_source_mapping_for_name(&mut self, output: &[u8], span: Span, name: &str) {
-        // SAFETY: search original string by span.
-        let original_name =
-            unsafe { self.original_source.get_unchecked(span.start as usize..span.end as usize) };
+        // Here using `get_unchecked` happened crash at windows. See https://github.com/rolldown/rolldown/issues/716
+        let original_name = self.original_source.get(span.start as usize..span.end as usize);
         // The token name should be original name.
         // If it hasn't change, name should be `None` to reduce `SourceMap` size.
-        let token_name = if original_name == name { None } else { Some(original_name.into()) };
+        let token_name =
+            if original_name == Some(name) { None } else { original_name.map(Into::into) };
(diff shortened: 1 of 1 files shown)


crates/oxc_codegen/src/sourcemap_builder.rs
```diff
@@ -62,12 +62,12 @@ impl SourcemapBuilder {
     }
 
     pub fn add_source_mapping_for_name(&mut self, output: &[u8], span: Span, name: &str) {
-        // SAFETY: search original string by span.
-        let original_name =
-            unsafe { self.original_source.get_unchecked(span.start as usize..span.end as usize) };
+        // Here using `get_unchecked` happened crash at windows. See https://github.com/rolldown/rolldown/issues/716
```
Boshen, line 65:
```suggestion
```

Boshen (approved):

crates/oxc_codegen/src/sourcemap_builder.rs
hyf0: Maybe we could add a `debug_assert!(span.end < self.original_source.len())` to deny all possiable invalid span in debug build. And and silent the error in release build because nothing we could do in that time.

crates/oxc_codegen/src/sourcemap_builder.rs
```diff
@@ -62,12 +62,11 @@ impl SourcemapBuilder {
     }
 
     pub fn add_source_mapping_for_name(&mut self, output: &[u8], span: Span, name: &str) {
-        // SAFETY: search original string by span.
-        let original_name =
-            unsafe { self.original_source.get_unchecked(span.start as usize..span.end as usize) };
+        let original_name = self.original_source.get(span.start as usize..span.end as usize);
```
hyf0, line 65: Maybe we should add a `debug_assert` to deny all invalid span as many as possible in debug build. For release build, we just ignore it because nothing we could do there.

PR #2896, opened 2024-04-03
fix(semantic): symbols inside functions and classes incorrectly flagged as exported

DonIsaac:
# What This PR Does

Symbols declared inside exported functions and classes were being incorrectly flagged with `SymbolFlags::Export`.

For example,
```ts
export function foo<T>(a: T) {
  let b = String(a)
  return b
}
```
`T`, `a`, and `b` were all flagged as exported.

## Further Work
It doesn't seem like exported enums and interfaces are being included in `ModuleRecord`. Am I looking in the wrong place, or are they actually missing?


DonIsaac:
It looks like this is affecting more than just function parameters. Working on this now.

DonIsaac:
Should be good to go now.

Boshen (approved):

PR #2897, opened 2024-04-05
Add basic preset/plugin layer for transforms.

milesj:
This assumes we'll be using `VisitMut`.


crates/oxc_transformer/src/preset_plugin.rs
```diff
@@ -0,0 +1,20 @@
+use oxc_ast::ast::Program;
+
+pub trait Transformation {
+    fn transform(&mut self, _program: &mut Program<'_>) {}
+}
+
+pub type BoxedTransformation = Box<dyn Transformation>;
+
+#[macro_export]
+macro_rules! impl_preset_transformation {
```
milesj, line 10: Will leave this in for now until we figure out the API.

Boshen (commented):
I'll cherry-pick some of these for Milestone 1, I don't want to expand the code too quickly for April.

PR #2905, opened 2024-04-06
feat(linter/import): Add `ignoreTypes` option for the `import/no-cycle` rule

### LINKED ISSUE #2904: [Feature Request][Linter] Allow type-only imports to be ignored by `import/no-cycle`
The `import/no-cycle` rule in `oxlint` currently reports cycles that include type-only imports.

```ts
// moduleA.ts
import type { Bar } from './moduleB';

export type Foo = string;

// ----------------------------------

// moduleB.ts
import type { Foo } from './moduleA';

export type Bar = number;
```

Running `oxlint` on the above example would result in an error like:
```
  ⚠ eslint-plugin-import(no-cycle): Dependency cycle detected
   ╭─[cycles/moduleA.ts]
 1 │ import type { Bar } from "./moduleB";
   ·                          ───────────
   ╰────
  help: These paths form a cycle:
        -> ./moduleB - examples/moduleB.ts
        -> ./moduleA - examples/moduleA.ts
```

The `eslint-plugin-import` version of `import/no-cycle` ignores type-only imports, because they have no runtime effect. See: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md#importno-cycle

It would be great if the `import/no-cycle` rule in `oxlint` could be configured to ignore type-only imports when checking for cycles

JohnDaly:
Fixes: https://github.com/oxc-project/oxc/issues/2904

Adds an `ignoreTypes` option for the `import/no-cycle` rule, to ignore imports that don't have any runtime effect.

### DIFF
--- a/crates/oxc_linter/src/rules/import/no_cycle.rs
+++ b/crates/oxc_linter/src/rules/import/no_cycle.rs
@@ -26,6 +26,9 @@ struct NoCycleDiagnostic(#[label] Span, String);
 pub struct NoCycle {
     /// maximum dependency depth to traverse
     max_depth: u32,
+    /// ignore type only imports
+    #[allow(unused)]
+    ignore_types: bool,
     /// ignore external modules
     #[allow(unused)]
     ignore_external: bool,
@@ -38,6 +41,7 @@ impl Default for NoCycle {
     fn default() -> Self {
         Self {
             max_depth: u32::MAX,
+            ignore_types: false,
             ignore_external: false,
             allow_unsafe_dynamic_cyclic_dependency: false,
         }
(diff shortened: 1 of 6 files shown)


crates/oxc_linter/src/rules/import/no_cycle.rs
```diff
@@ -26,6 +26,9 @@ struct NoCycleDiagnostic(#[label] Span, String);
 pub struct NoCycle {
     /// maximum dependency depth to traverse
     max_depth: u32,
+    /// ignore type only imports
+    #[allow(unused)]
```
Boshen, line 30:
```suggestion
```

Boshen (approved):
Thank you!

Issue #2875, opened 2024-03-30
Svelte parser fails with generic components


notramo:
The Svelte parser fails when the component has TypeScript generics.
This is valid Svelte syntax:
```svelte
<script lang=ts generics="T extends Record<string, unknown>">
```
The TS generic type syntax in `Record<string, unknown>` is considered as closing `>` by the parser, disregarding that it's inside an attribute.

Boshen:
Similar to https://github.com/oxc-project/oxc/issues/2017, we need to improve the parser to balance out the angle brackets.

The parsing code is here https://github.com/oxc-project/oxc/tree/main/crates/oxc_linter/src/partial_loader

PRs are welcome.

kalvenschraut:
@Boshen  this issue can be closed, not sure why it didn't auto close on my PR being merged

PR #2906, opened 2024-04-07
fix(linter): svelte partial loader handle generics (#2875)

kalvenschraut:
While looking into this, I knew vue would have similar problems. Which then led me to realize the generic case was handled already so I tried abstract out the function so both the vue and svelte partial loaders could reuse the code. Finally added some svelte test cases to prove this resolved the issue.

I am a bit new to rust so if there was a better way to reuse the find_script_closing_angle function open to suggestions.


Boshen (approved):

Boshen (approved):
Thank you!

Issue #2017, opened 2024-01-13
Incorrect parsing of Astro component script tags


minht11:
Have astro component like this
```astro
---
const { props } = Astro

const attrs = {}
---

<script is:inline {...attrs} set:html={props.src ? '' : props.code} />

<script>
	interface Attrs {
		'data-attrs'?: string
	}
</script>

```
running OXC 0.2 produces 
```
 x Expected a semicolon or an implicit semicolon after a statement, but found none
         site:lint |    ,-[src/components/Script.astro:3:1]
         site:lint |  3 | <script>
         site:lint |  4 |     interface Attrs {
         site:lint |    :              ^
         site:lint |  5 |         'data-attrs'?: string
         site:lint |    `----
         site:lint |   help: Try insert a semicolon here
```

My guess is that oxc does not recognize script as typescript

minht11:
Commenting out code
```astro
---
const { props } = Astro

const attrs = {}
---

<script is:inline {...attrs} set:html={props.src ? '' : props.code} />

<!-- <script>
	interface Attrs {
		'data-attrs'?: string
	}
</script> -->
```
Produces
```
x Unexpected token
         site:lint |    ,-[src/components/Script.astro:2:1]
         site:lint |  2 | 
         site:lint |  3 | <!-- <script>
         site:lint |    :  ^
         site:lint |  4 |     interface Attrs {
         site:lint |    `----
```

Boshen:
This is harder than I thought 😅

IWANABETHATGUY:
Since it has error recovery, we don't need to struggle with the wired edge case, the only drawback is it may increase binary size.

IWANABETHATGUY:
> > Since it has error recovery, we don't need to struggle with the wired edge case, the only drawback is it may increase binary size.
> 
> Binary size + performance is not something I'd trade off for, I'll take a look at their compiler test suite.

They don't have too many test suites, https://github.com/ikatyang/tree-sitter-vue/blob/master/corpus/spec.txt, the most attractive feature of `tree-sitter` is the error recovery, they can always find the valid `script` tag, what every the source code is, this could solve our problem perfectly.

Boshen:
we need to improve the parser to balance out the angle brackets.

The parsing code is here https://github.com/oxc-project/oxc/tree/main/crates/oxc_linter/src/partial_loader

PRs are welcome.

kalvenschraut:
[...]

PR #2907, opened 2024-04-07
fix(linter): handle self closing script tags in astro partial loader (#2017)

kalvenschraut:
Added tests to figure out what the problem from the issue was and realize the self closing tag was not expected.  Added code to handle this case.

I am not that familiar with astro to know how common self closing script tags are, their docs seem to use an opening and closing one for inline scripts https://docs.astro.build/en/guides/client-side-scripts/#load-external-scripts


Boshen (approved):

Boshen:
Thank you!

PR #2912, opened 2024-04-07
feat(linter): typescript-eslint/no-floating-promises

valeneiko:
https://github.com/oxc-project/oxc/discussions/2855

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -316,6 +316,51 @@ impl<'a> Expression<'a> {
             _ => false,
         }
     }
+
+    pub fn get_span(&self) -> &Span {
+        match self {
+            Expression::BooleanLiteral(e) => &e.span,
+            Expression::NullLiteral(e) => &e.span,
+            Expression::NumericLiteral(e) => &e.span,
+            Expression::BigintLiteral(e) => &e.span,
+            Expression::RegExpLiteral(e) => &e.span,
+            Expression::StringLiteral(e) => &e.span,
+            Expression::TemplateLiteral(e) => &e.span,
+            Expression::Identifier(e) => &e.span,
+            Expression::MetaProperty(e) => &e.span,
+            Expression::Super(e) => &e.span,
+            Expression::ArrayExpression(e) => &e.span,
+            Expression::ArrowFunctionExpression(e) => &e.span,
+            Expression::AssignmentExpression(e) => &e.span,
+            Expression::AwaitExpression(e) => &e.span,
+            Expression::BinaryExpression(e) => &e.span,
+            Expression::CallExpression(e) => &e.span,
+            Expression::ChainExpression(e) => &e.span,
+            Expression::ClassExpression(e) => &e.span,
+            Expression::ConditionalExpression(e) => &e.span,
+            Expression::FunctionExpression(e) => &e.span,
+            Expression::ImportExpression(e) => &e.span,
+            Expression::LogicalExpression(e) => &e.span,
+            Expression::MemberExpression(e) => e.get_span(),
+            Expression::NewExpression(e) => &e.span,
+            Expression::ObjectExpression(e) => &e.span,
+            Expression::ParenthesizedExpression(e) => &e.span,
+            Expression::SequenceExpression(e) => &e.span,
+            Expression::TaggedTemplateExpression(e) => &e.span,
+            Expression::ThisExpression(e) => &e.span,
+            Expression::UnaryExpression(e) => &e.span,
+            Expression::UpdateExpression(e) => &e.span,
+            Expression::YieldExpression(e) => &e.span,
+            Expression::PrivateInExpression(e) => &e.span,
+            Expression::JSXElement(e) => &e.span,
+            Expression::JSXFragment(e) => &e.span,
+            Expression::TSAsExpression(e) => &e.span,
+            Expression::TSSatisfiesExpression(e) => &e.span,
+            Expression::TSTypeAssertion(e) => &e.span,
+            Expression::TSNonNullExpression(e) => &e.span,
+            Expression::TSInstantiationExpression(e) => &e.span,
+        }
+    }
 }
 
 /// Identifier Name
@@ -718,6 +763,14 @@ impl<'a> MemberExpression<'a> {
         self.object().is_specific_id(object)
             && self.static_property_name().is_some_and(|p| p == property)
     }
+
+    pub fn get_span(&self) -> &Span {
+        match self {
+            MemberExpression::ComputedMemberExpression(e) => &e.span,
+            MemberExpression::StaticMemberExpression(e) => &e.span,
+            MemberExpression::PrivateFieldExpression(e) => &e.span,
+        }
+    }
 }
 
 /// `MemberExpression[?Yield, ?Await] [ Expression[+In, ?Yield, ?Await] ]`
@@ -855,6 +908,13 @@ pub enum Argument<'a> {
 impl Argument<'_> {
     pub fn is_spread(&self) -> bool {
         matches!(self, Self::SpreadElement(_))
+    }
+
+    pub fn get_span(&self) -> &Span {
+        match self {
+            Argument::SpreadElement(e) => &e.span,
+            Argument::Expression(e) => e.get_span(),
+        }
     }
 }
 
@@ -1177,6 +1237,15 @@ pub struct ChainExpression<'a> {
 pub enum ChainElement<'a> {
     CallExpression(Box<'a, CallExpression<'a>>),
     MemberExpression(Box<'a, MemberExpression<'a>>),
+}
+
+impl<'a> ChainElement<'a> {
+    pub fn get_span(&self) -> &Span {
+        match self {
+            ChainElement::CallExpression(e) => &e.span,
+            ChainElement::MemberExpression(e) => e.get_span(),
+        }
+    }
 }
 
 /// Parenthesized Expression
--- a/crates/oxc_cli/src/command/lint.rs
+++ b/crates/oxc_cli/src/command/lint.rs
@@ -170,6 +170,10 @@ pub struct EnablePlugins {
     /// Enable the React performance plugin and detect rendering performance problems
     #[bpaf(switch, hide_usage)]
     pub react_perf_plugin: bool,
+
+    /// Enable the TypeCheck plugin and detect type-based problems
+    #[bpaf(switch, hide_usage)]
+    pub typecheck_plugin: bool,
 }
 
 #[cfg(test)]
--- a/crates/oxc_cli/src/lint/mod.rs
+++ b/crates/oxc_cli/src/lint/mod.rs
@@ -98,7 +98,8 @@ impl Runner for LintRunner {
             .with_jest_plugin(enable_plugins.jest_plugin)
             .with_jsx_a11y_plugin(enable_plugins.jsx_a11y_plugin)
             .with_nextjs_plugin(enable_plugins.nextjs_plugin)
-            .with_react_perf_plugin(enable_plugins.react_perf_plugin);
+            .with_react_perf_plugin(enable_plugins.react_perf_plugin)
+            .with_type_info(enable_plugins.typecheck_plugin);
 
         let linter = match Linter::from_options(lint_options) {
             Ok(lint_service) => lint_service,
--- a/crates/oxc_language_server/src/linter.rs
+++ b/crates/oxc_language_server/src/linter.rs
@@ -300,6 +300,7 @@ impl IsolatedLintHandler {
             };
 
             let lint_ctx = LintContext::new(
+                Path::new("./"),
                 path.to_path_buf().into_boxed_path(),
                 &Rc::new(semantic_ret.semantic),
                 // TODO: create type checker
--- a/crates/oxc_language_server/src/linter.rs
+++ b/crates/oxc_language_server/src/linter.rs
@@ -302,6 +302,8 @@ impl IsolatedLintHandler {
             let lint_ctx = LintContext::new(
                 path.to_path_buf().into_boxed_path(),
                 &Rc::new(semantic_ret.semantic),
+                // TODO: create type checker
+                None,
             );
 
             let result = linter.run(lint_ctx);
--- a/crates/oxc_linter/fixtures/typecheck/tsconfig.json
+++ b/crates/oxc_linter/fixtures/typecheck/tsconfig.json
@@ -1,0 +1,18 @@
+{
+  "compilerOptions": {
+    "jsx": "preserve",
+    "target": "es5",
+    "module": "commonjs",
+    "strict": true,
+    "esModuleInterop": true,
+    "lib": [
+      "es2015",
+      "es2017",
+      "esnext"
+    ],
+    "experimentalDecorators": true
+  },
+  "include": [
+    "file.ts"
+  ]
+}
--- a/crates/oxc_linter/src/context.rs
+++ b/crates/oxc_linter/src/context.rs
@@ -32,6 +32,8 @@ pub struct LintContext<'a> {
 
     file_path: Box<Path>,
 
+    absolute_path: Box<String>,
+
     settings: Arc<ESLintSettings>,
 
     env: Arc<ESLintEnv>,
@@ -41,12 +43,14 @@ pub struct LintContext<'a> {
 
 impl<'a> LintContext<'a> {
     pub fn new(
+        cwd: &Path,
         file_path: Box<Path>,
         semantic: &Rc<Semantic<'a>>,
         type_checker: Option<Arc<Mutex<TSServerClient>>>,
     ) -> Self {
         let disable_directives =
             DisableDirectivesBuilder::new(semantic.source_text(), semantic.trivias()).build();
+        let absolute_path = Box::new(cwd.join(file_path.as_ref()).to_string_lossy().into());
         Self {
             semantic: Rc::clone(semantic),
             diagnostics: RefCell::new(vec![]),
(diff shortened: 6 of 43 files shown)


Boshen:
OMG this is amazing, fingers crossed for communicating with the ts server, as discussed in https://github.com/oxc-project/oxc/discussions/2855

Boshen:
@valeneiko Such an amazing effort! Feel free to ping me with preliminary results!

valeneiko:
@Boshen Almost there!

It now successfully detects a floating Promise in a simple file:
```console
cargo run --bin oxlint -- --typecheck-plugin -A all -D no-floating-promises \
  --tsconfig ./crates/oxc_linter/fixtures/typecheck/tsconfig.json \
  ./crates/oxc_linter/fixtures/typecheck/sample.ts
```
```typescript
async function funcAsync() {
  return Promise.resolve();
}

function func() {
  return 7;
}

export async function test() {
  func();
  funcAsync();
}
```

first_result

Boshen:
image

I tested this on cal.com, and it's working on multiple files!

It panic with 

```
called `Result::unwrap()` on an `Err` value: CommandFailed("Error processing request. Cannot read properties of undefined (reading 'kind')\nTypeError: Cannot read properties of undefined (reading 'kind')\n    at isDeclarationNameOrImportPropertyName (/Users/bytedance/github/oxc/npm/oxc-typecheck/node_modules/.pnpm/typescript@5.4.4/node_modules/typescript/lib/typescript.js:87886:25)\n    at getTypeOfNode (/Users/bytedance/github/oxc/npm/oxc-typecheck/node_modules/.pnpm/typescript@5.4.4/node_modules/typescript/lib/typescript.js:85186:11)\n    at Object.getTypeAtLocation (/Users/bytedance/github/oxc/npm/oxc-typecheck/node_modules/.pnpm/typescript@5.4.4/node_modules/typescript/lib/typescript.js:46305:23)\n    at Module.isPromiseArray (file:///Users/bytedance/github/oxc/npm/oxc-typecheck/dist/rules/no-floating-promises.js:4:26)\n    at noFloatingPromises::isPromiseArray (file:///Users/bytedance/github/oxc/npm/oxc-typecheck/dist/handlers.js:46:43)\n    at executeCommand (file:///Users/bytedance/github/oxc/npm/oxc-typecheck/dist/server.js:41:26)\n    at onMessage (file:///Users/bytedance/github/oxc/npm/oxc-typecheck/dist/server.js:26:48)\n    at Interface.<anonymous> (file:///Users/bytedance/github/oxc/npm/oxc-typecheck/dist/server.js:16:9)\n    at Interface.emit (node:events:518:28)\n    at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)")
```

I think it's due to providing utf8 span positions instead if utf16, we don't need to worry about this as there is already a tracking issue.

Now onto performance: if performance of what we have right now is less or equal to @typescript/eslint, then we are all good!

Boshen:
@valeneiko Can you time our setup and compare to one or two repositories with no-floating-promises on? https://github.com/search?q=no-floating-promises&type=code

https://github.com/mermaid-js/mermaid seems like a good place to start: `Finished in 41ms on 397 files with 89 rules using 12 threads.`

valeneiko:
@Boshen I have tried to run lint on the repo you linked, but it fails with the exact same error. And it's cause is not just encoding differences.

Take a look at the AST of one of the files from the repo:
- oxc: [link](https://oxc-project.github.io/oxc/playground/?code=3YCAAIBghoCAgICAgIC0G8rnONK89IyprRfcKLUDvt6c47wuXWLAMhBkTB4yV2kmXJ106nG0aw6vDM5E3y%2BCbB%2FUClbZiGHjC42TB5idIDuM6dji96hinXeuaqqGiiJ60HbxFAGx71EnzFA5cVQWV6QtiEOvsqRonetP6AkmjHH39SrtIuHGqSRZ6V6lac%2FpDIc%2FS1tcGiDnf1QleYnCiR7RdfmtTVgXp4SxC6TvdL0BdtIiuFGHcQ64dytdQJFG1%2Bj1kjqSA7OxENoOQ87h%2FW%2BiI9wECFAc8rTxL89PU4S%2FRP%2F2L744LAGJFsVnu4CoPa7uA8%2B3mBYPa244E7sYpoXswPncufuoe1eXQ1Cw7pGYlmAyLtEN0ubNz6wYY%2FXnJi35aMIOK5ZLHveNaNa1y%2FC4iJthGmRCjX7afS09xSI2jUtbWR3xIT5KegEqrDtT25Qj08ZS64iitX4ZntfJnxmZQEGs1Zn6qivO%2Fm9KYztP%2B8LRdQWEqZ7qkKqDsXNbbkovB6ypmzi%2FxB1iqG1GHjppW5ieMkGkS8sqoS52EklQtKkuxLcp5BpgU3L5fGvmDsE9gsCnrROT3PjY2Hp9RjCRntCcoqsiY1vgYpx5nWCiln8ljpnaX51qC8y7HqpjduzKDe7Yp3fNNpc3mUlQDNeiPsySgGjuLmIjAmdhi3rbyH4uKWl5bLg06o8521BjREh%2BIFoeoAF3htsiCM25iCgo%2FtIiR0ywArMM1UYzF00FbMpv%2Fis5tY%2BONNC26xdClhHC1zs%2FbB6Bw1UE3fJRqJsMezx0gxy66Qd2yeFqRhd8AwIkAefr30PeGM8Zn%2FK%2BhRI6e86%2BKDvaFLtqIbLGJDRT4RoylWMd%2BpE6dmyYfn0DxhanQmBUeU0V5t8D%2BRqYdZ%2Fa83%2FCVpbcOAv0FtMX85JFXNX%2FS48GMIDEsMf8G4RKOGPgdHEO62Q5%2F%2FoO55tLVQVe5nvHmv8zhPAuJkFaM0swLx8%2F2DnTEMiOpGmnCwHtcFZc4C2pmiBiMTcpYm8MPijOENF4sSKC428GhZtNfFBPjw%3D%3D)
- typescript: [link](https://astexplorer.net/#/gist/55f532b8f881ef6decf34528be8f93a4/9c6b1d4d7fd555a0a9c35e927ac1ebd86746b407)

OXC makes a request for `CallExpression at (58, 1758)`, but typescript has this node at `(56, 1754)`. Looks like typescript includes leading newlines and spaces as part of the CallExpression span.


Do you have any suggestions for how we can better map between the two? I am thinking we could try a full path or just child indexes, but now quite sure how to get this from OXC.


On a side note, I finished the remaining parts of the lint rule.

Boshen:
> @Boshen I have tried to run lint on the repo you linked, but it fails with the exact same error. And it's cause is not just encoding differences.
> 
> Take a look at the AST of one of the files from the repo:
> 
> * oxc: [link](https://oxc-project.github.io/oxc/playground/?code=3YCAAIBghoCAgICAgIC0G8rnONK89IyprRfcKLUDvt6c47wuXWLAMhBkTB4yV2kmXJ106nG0aw6vDM5E3y%2BCbB%2FUClbZiGHjC42TB5idIDuM6dji96hinXeuaqqGiiJ60HbxFAGx71EnzFA5cVQWV6QtiEOvsqRonetP6AkmjHH39SrtIuHGqSRZ6V6lac%2FpDIc%2FS1tcGiDnf1QleYnCiR7RdfmtTVgXp4SxC6TvdL0BdtIiuFGHcQ64dytdQJFG1%2Bj1kjqSA7OxENoOQ87h%2FW%2BiI9wECFAc8rTxL89PU4S%2FRP%2F2L744LAGJFsVnu4CoPa7uA8%2B3mBYPa244E7sYpoXswPncufuoe1eXQ1Cw7pGYlmAyLtEN0ubNz6wYY%2FXnJi35aMIOK5ZLHveNaNa1y%2FC4iJthGmRCjX7afS09xSI2jUtbWR3xIT5KegEqrDtT25Qj08ZS64iitX4ZntfJnxmZQEGs1Zn6qivO%2Fm9KYztP%2B8LRdQWEqZ7qkKqDsXNbbkovB6ypmzi%2FxB1iqG1GHjppW5ieMkGkS8sqoS52EklQtKkuxLcp5BpgU3L5fGvmDsE9gsCnrROT3PjY2Hp9RjCRntCcoqsiY1vgYpx5nWCiln8ljpnaX51qC8y7HqpjduzKDe7Yp3fNNpc3mUlQDNeiPsySgGjuLmIjAmdhi3rbyH4uKWl5bLg06o8521BjREh%2BIFoeoAF3htsiCM25iCgo%2FtIiR0ywArMM1UYzF00FbMpv%2Fis5tY%2BONNC26xdClhHC1zs%2FbB6Bw1UE3fJRqJsMezx0gxy66Qd2yeFqRhd8AwIkAefr30PeGM8Zn%2FK%2BhRI6e86%2BKDvaFLtqIbLGJDRT4RoylWMd%2BpE6dmyYfn0DxhanQmBUeU0V5t8D%2BRqYdZ%2Fa83%2FCVpbcOAv0FtMX85JFXNX%2FS48GMIDEsMf8G4RKOGPgdHEO62Q5%2F%2FoO55tLVQVe5nvHmv8zhPAuJkFaM0swLx8%2F2DnTEMiOpGmnCwHtcFZc4C2pmiBiMTcpYm8MPijOENF4sSKC428GhZtNfFBPjw%3D%3D)
> * typescript: [link](https://astexplorer.net/#/gist/55f532b8f881ef6decf34528be8f93a4/9c6b1d4d7fd555a0a9c35e927ac1ebd86746b407)
> 
> OXC makes a request for `CallExpression at (58, 1758)`, but typescript has this node at `(56, 1754)`. Looks like typescript includes leading newlines and spaces as part of the CallExpression span.
> 
> Do you have any suggestions for how we can better map between the two? I am thinking we could try a full path or just child indexes, but now quite sure how to get this from OXC.
> 
> On a side note, I finished the remaining parts of the lint rule.

Ahh such a big blocker, where is typescript eslint do the conversion? It's probably in https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-estree/src/ast-converter.ts but I can't find it from a quick scan.

I need to measure the effort of going forward with this experiment with we want to do the conversion.

In the meantime, can you help out and just ignore the error / crash and let the whole thing complete so we can compare the total running time between typescript eslint and our approach?

It's gonna be a lot of balancing of tradeoffs 🥵

valeneiko:
> Ahh such a big blocker, where is typescript eslint do the conversion?

Here is the conversion logic in typescript-eslint: https://github.com/typescript-eslint/typescript-eslint/blob/9c94b81b2560558fcc2d85dd644679a5dbf4a277/packages/typescript-estree/src/convert.ts#L757

If I understand it correctly, typescript eslint build their own AST by visiting the AST typescript produced.


> In the meantime, can you help out and just ignore the error / crash and let the whole thing complete so we can compare the total running time between typescript eslint and our approach?

Ok. I'll find a way to ignore those errors for now.

Boshen:
> If I understand it correctly, typescript eslint build their own AST by visiting the AST typescript produced.

Yeah I saw that too, but I can't find where it's changing the span positions :-(

> Ok. I'll find a way to ignore those errors for now.

Thank you!

valeneiko:
> Yeah I saw that too, but I can't find where it's changing the span positions :-(

They don't need to. Instead of looking up nodes by spans, the just directly map the nodes, because they store which TS node they used to create each one.


> ...

I've pushed the commits to ignore the errors. And managed to run the rule against mermaid repo. I've modified their `.eslintrc.cjs` to only enable `@typescript-eslint/no-floating-promises` and ESLint runs in `3.2 seconds`. OXC with that single rule takes `90 seconds` (using 24 threads), or `220 seconds` (single threaded). There is obviously something wrong.

- It could be that mapping errors lead us to type check different code paths.
- Out lint rule implementeation is not 100% correct, so we may be calling TS more than necessary
  - I know at least 1 case: `() => new Promise()`. ESLint completely skips those, but we call typecheck and generate an error.
- But more likely there is something wrong in our JS side implementation. It should not be taking this long.

Boshen:
> But more likely there is something wrong in our JS side implementation. It should not be taking this long.

I agree, my guess is that typescript eslint is using some kind of caching technique or some API that we don't know about 😅

valeneiko:
Result of running OXC in debug mode against mermaid with all the measurements:
```yaml
                      #: average /   total
                  parse: 0.007ms /  0.225s
              stringify: 0.019ms /  0.624s
                   open: 7.939ms / 66.672s
                  close: 0.320ms /  0.127s
         isPromiseArray: 0.173ms /  2.703s
          isPromiseLike: 0.014ms /  0.226s
isValidRejectionHandler: 0.039ms /  0.000s
             getProgram: 0.015ms /  0.460s
         getTypechecker: 0.000ms /  0.008s
                getNode: 0.004ms /  0.114s
        channelOverhead: 0.070ms /  2.261s
                   idle: 0.071ms /  2.283s
```

A few observations:
- Actual work is basically the same as `typescript-eslint`
- We spend over 3 seconds on just communication overhead
  - We could lower it by not sending file content
  - We should pipeline the requests, so that TS is never idle
  - We could use something faster than JSON
- We spend most of our time opening files in TS
  - This seems like an unnecassary step - we should look into how ESLint does it

Boshen:
> A few observations:

I see these are optimizations that can be done later.

Performance wise we are on the correct path, which is really good news 👍 .

Now the last puzzle piece to solve is the mismatched AST nodes. Let's explore this for a few days before jumping to any conclusions.

valeneiko:
After looking at a few different examples, OXC and TS have both different structure and different spans. So mapping code is unavoidable.

One idea comes to mind:
- Build a visitor, that walks the tree in the same order on both sides. Then we can assign each node an id, and look it up by id on both sides. Visitor implementation in TypeScript is only about 600 lines ([here](https://github.com/microsoft/TypeScript/blob/f71d629e3f3a6665f2313d1d54ffb79fe12c1902/src/compiler/parser.ts#L499)). So not too diffucult to map. And we can probably exploit the fact that nodes are going to be visited in the order they appear in the source file. And only assign the Ids to nodes that map nicely (at least to start with), since we only need to be able to resolve a subset of them anyway.

valeneiko:
After exploring it a bit more, I don't think my previous idea would be enough. At this point the only solution I see is to fully map the AST. But given the effort required to maintain it I am not sure it would be worth it: we don't gain anything perf wise by having an ability to write these rules in rust.

Boshen:
> After exploring it a bit more, I don't think my previous idea would be enough. At this point the only solution I see is to fully map the AST. But given the effort required to maintain it I am not sure it would be worth it: we don't gain anything perf wise by having an ability to write these rules in rust.

😢 It's sad to see that you have spent so much time and energy on this but it's blocked by how typescript works. TypeScript doesn't offer linter plugins, and we are forced to do this whole round trip despite us willing to do all these hard work, but we'll also need to maintain this whole round trip.

I think at this point, we should just stop thinking about porting these rules from @typescript/eslint and start innovating our own set of rules that can help people catch bugs.

I mean, it was tried before `https://github.com/microsoft/TypeScript/pull/53146` and it wasn't prioritized ...

If anyone is going to ask again about typing rules, I'll just point them to this PR :-(

---

So back to the drawing board.

Since we have cross-file linting working with `--import-plugin`, should we start innovating with cross-file information, and maybe start implementing things like `#[must_use]`?

What do you think?

valeneiko:
> Since we have cross-file linting working with `--import-plugin` https://oxc-project.github.io/blog/2024-05-04-import-plugin-alpha.html, should we start innovating with cross-file information, and also look into things like `#[must_use]`?
> 
> What do you think?

Sound like a logical next step. It does feel like re-implementing TypeScript in Rust, but maybe we only need a very small subset to get most of the way. The complex part would be symbol resolution and I think we should start with that.

I am not so sure about adding extra annotation to code, I would prefer enabling this rule instead: [explicit-module-boundary-types](https://typescript-eslint.io/rules/explicit-module-boundary-types/). Then the cross-file step could easily mark if a function or method returns a promise. And even without the rule, most of those functions would be annotated with `async` already, and we could use that instead of the return type. But regarless of the approach, we would need to have an ability to resolve references to their corresponding symbols, so that we could look up annotations on them.

Boshen:
Let me organize this a bit tomorrow.

> The complex part would be symbol resolution and I think we should start with that.

If the goal is still no-floating-promises, I think we can get it working first by marking exported async functions of a module as `async` in `ModuleRecord`

https://github.com/oxc-project/oxc/blob/07076d9765cee6f4986c5a4e599f9edc61f60d8f/crates/oxc_syntax/src/module_record.rs#L67-L68

> I am not so sure about adding extra annotation to code, I would prefer enabling this rule instead: [explicit-module-boundary-types](https://typescript-eslint.io/rules/explicit-module-boundary-types/).

I meant marking things with extra annotations, there is a system in rust where you can mark things with different things in compile time, `#[must_use]` is one of them.

Now that you have mentioned it, maybe we can leverage isolated-declarations? And just do linting under isolated-declarations 🤔 We prepare for the future 😆 

https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#isolated-declarations

valeneiko:
> Now that you have mentioned it, maybe we can leverage isolated-declarations?

Yes, this is exactly what I meant. But since that feature is not even out in TS, `explicit-module-boundary-types` could still make sense, and achieves the same result.

> If the goal is still no-floating-promises, I think we can get it working first by marking exported async functions of a module as async in ModuleRecord

That could be a good start. I was mostly concerned about complexity of doing it for cases other than named import of a function:
```ts
// foo.ts
export function foo() {}
export const obj = { foo() {} };
export class Cls { foo() {} }

// -- default / namespace import
import mod from './foo';
mod.foo(); // <-- How do we resolve foo?

// -- object properties
import {obj} from './foo';
obj.foo(); // <-- How do we resolve foo?

// -- class fields
import {Cls} from './foo';
const x = new Cls();
x.foo(); // <-- How do we resolve foo?
```

Boshen:
I feel said about closing this PR :-(

Continue in https://github.com/oxc-project/oxc/issues/3105#issuecomment-2100859582

Boshen:
image

Source: https://speakerdeck.com/unvalley/exploring-type-informed-lint-rules-in-rust-based-linters

PR #2917, opened 2024-04-08
refactor(semantic/jsdoc): Rework JSDoc struct for better Span handling

leaysgur:
> The error message emphasizes "empty text" so I would put the span on the extra text.
> https://github.com/oxc-project/oxc/pull/2893#discussion_r1548843621

To address this, special `Span` handling should be implemented for comment part.

So, this PR introduces:

- `JSDocCommentPart` struct holds raw `.span` and special `.span_trimmed_first_line()`
- Add `JSDocKindPart`, `JSDocTypePart` and `JSDocTypeNamePart` in the same manner
  - `JSDocTag` uses these depending on the purpose


leaysgur:
I'm not sure what causes performance regression...
Since the boundary between Semantic and JSDoc has not changed as `OnceCell`.

@Boshen would you help me?

Boshen (approved):
Amazing work! Thank you!

The performance regression is unrelated.

PR #2919, opened 2024-04-08
feat(transformer): typescript and react presets

Boshen:


Boshen:
@milesj This still feels weird. The conformance suite needs to use each individual plugin and pass in the options. So plugins and preset needs to be orthogonal to each other, nesting them with boolean flags make things harder.

I'll explore a bit more and see how to glue these things together, without the babel complexity.

PR #2920, opened 2024-04-08
perf(npm): use linked binary package for startup boost

dalisoft:
This changes/PR should improve startup performance of CLI by up 8 times and very huge impact for lint-staged, pre-commit and pre-push commits

### Benchmark using `hyperfine`

| Command                                      |  Mean [ms] | Min [ms] | Max [ms] |    Relative |
| :------------------------------------------- | ---------: | -------: | -------: | ----------: |
| `oxlint`                                     | 51.7 ± 0.4 |     51.0 |     52.2 | 7.84 ± 0.33 |
| `./node_modules/@oxlint/darwin-arm64/oxlint` |  6.6 ± 0.3 |      6.3 |      7.1 |        1.00 |

<details>

<summary>Benchmark script</summary>

```sh
#!/bin/sh
set -eu

# Exports for binaries access
export PATH="./node_modules/.bin:$PATH"

hyperfine --warmup 3 --runs 10 'oxlint' './node_modules/@oxlint/darwin-arm64/oxlint' --export-markdown "oxlint.md"
```

</details>


Boshen:
How does the root package know which binary to call if the oxlint bin script is removed 🤔 ?

dalisoft:
@Boshen oxlint itself blazingly fast but nodejs overhead caused it to be slow for startup. 

Actually, biome, dprint and most of rust projects suffers from this bug/slowdown

Boshen:
This doesn't work

image

And I think you didn't test this locally by running the `generate-packages.mjs` script because it errors out.

dalisoft:
@Boshen it should work after upload/publish and install into project. After build, try link and install on some project, then test

dalisoft:
@Boshen i'm added my own repo to demonstrate the how fast is **oxlint** itself

https://github.com/dalisoft/oxlint-rs-npm/tree/master/benchmark

Boshen:
We need the bin script in the `oxlint` package, otherwise `npx oxlint@latest` will not work.

For your benchmark, it's ok to call the binary via `node_modules/.bin/oxlint`.

dalisoft:
I understood, thank you. So i'll close the this PR the

PR #2925, opened 2024-04-09
fix(cli): if format is json do not print summary information (#2899)

### LINKED ISSUE #2899: [Feature Request][Enhancement] json file output
Right now, there isn't a clean way to get JSON output into a file. In [this sample project](https://github.com/MIreland/oxc-example), I use:

`"oxlint:json": "mkdir -p artifacts && oxlint src --format json > artifacts/oxc-lint.json",`

But the output is invalid:

```
[
	{"message": "eslint-plugin-react(jsx-no-target-blank): Using target=`_blank` without rel=`noreferrer` (which implies rel=`noopener`) is a security risk in older browsers: see https://mathiasbynens.github.io/rel-noopener/#recommendations","severity": "warning","causes": [],"help": "add rel=`noreferrer` to the element","filename": "src/App.tsx","labels": [{"span": {"offset": 595,"length": 8}}],"related": []},
	{"message": "eslint-plugin-react(jsx-no-target-blank): Using target=`_blank` without rel=`noreferrer` (which implies rel=`noopener`) is a security risk in older browsers: see https://mathiasbynens.github.io/rel-noopener/#recommendations","severity": "warning","causes": [],"help": "add rel=`noreferrer` to the element","filename": "src/App.tsx","labels": [{"span": {"offset": 727,"length": 8}}],"related": []}
]

Finished in 10ms on 5 files with 89 rules using 10 threads.
Found 2 warnings and 0 errors.

```

On a similar note (I can make this a separate ticket if desired), we would _**REALLY**_ appreciate being able to specify multiple reporters- specifically, we want human-readable terminal output at the same time we generate machine-readable JSON output.

We currently use [this eslint wrapper](https://github.com/lwhiteley/eslint-output?tab=readme-ov-file#eslint-output) to achieve this with eslint. It would be great to offer it natively for oxlint.

kalvenschraut:
this allows the ability to redirect the stdout to so can create a valid json file

this also matches with eslint's format option

closes #2899


Boshen (approved):
Thank you!

PR #2926, opened 2024-04-09
feat(sourcemap): add sourceRoot

underfin:


crates/oxc_sourcemap/src/decode.rs
```diff
@@ -8,6 +8,7 @@ use crate::{SourceMap, Token};
 struct JSONSourceMap {
     file: Option<String>,
     mappings: Option<String>,
+    source_root: Option<String>,
     sources: Option<Vec<Option<String>>>,
     sources_content: Option<Vec<Option<String>>>,
     names: Option<Vec<String>>,
```
Boshen, line 14: I think it's time to document what these fields are ...

Boshen (approved):

PR #2935, opened 2024-04-11
feat(linter): Add --jsdoc-plugin flag

leaysgur:
Add flag to:

```sh
$ oxlint --jsdoc-plugin
```

image


leaysgur:
Should update here too...?

https://github.com/oxc-project/oxc/blob/6757dbab8a9473b2588a99e5a2b4edc7aeeee3d6/crates/oxc_linter/src/options.rs#L239-L255

Boshen:
Seems like we missed all the plugins here 

https://github.com/oxc-project/oxc/blob/c250b288ef0b9a3ca72ac01ff96e3e240b78a30d/tasks/benchmark/benches/linter.rs#L35-L38

Can you help me and add them, so we can catch performance regressions.

leaysgur:
OK, I added 👉🏻 https://github.com/oxc-project/oxc/pull/2935/commits/547417b533dd8f5eecbd8101aabb6eed318f4d85

By the way, `LintOptions` has `xxx_plugin` properties only for

- import
- jsdoc
- jest
- jsx_a11y
- nextjs
- react_perf

Should we define this flag also for react, tree_shaking, typescript, unicorn, etc...?

Boshen:
> Should we define this flag also for react, tree_shaking, typescript, unicorn, etc...?

react, typescript and unicorn are enabled by default, I don't know how to turn them off now lol

Boshen (approved):

PR #2937, opened 2024-04-11
feat(transformer): implement plugin-transform-react-display-name top-down

Boshen:
Missing case:

https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-react-display-name/test/fixtures/display-name/nested/input.js

```js
var foo = qux(createReactClass({}));
var bar = qux(React.createClass({}));
```

This requires recursing down.

Top-down implementation in swc:

https://github.com/swc-project/swc/blob/67ec5e09b945591d4efba7261248c0280c605aa4/crates/swc_ecma_transforms_react/src/display_name/mod.rs#L108-L132

Or bottom-up in babel:

https://github.com/babel/babel/blob/08b0472069cd207f043dd40a4d157addfdd36011/packages/babel-plugin-transform-react-display-name/src/index.ts#L87-L98

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -104,4 +104,19 @@ impl<'a> VisitMut<'a> for Transformer<'a> {
         self.x1_react.transform_expression(expr);
         walk_mut::walk_expression_mut(self, expr);
     }
+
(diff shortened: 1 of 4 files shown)


Boshen:
I'm going to merge this and moving onto finding the next challenge for y'all.

PR #2943, opened 2024-04-11
Make ast types covariant over the allocator lifetime.

wan9chi:
## Why

Due to the usage of `&'alloc mut T` in `oxc_allocator::Box`, and `bumpalo::collections::Vec` in `oxc_allocator::Vec`, ast types are currently invariant over their allocator lifetime `'a`. This prevents `ouroboros` from generating `borrow_*` on ast type fields, leading to the unfriendly `with_*` api: https://github.com/oxc-project/oxc/blob/c250b288ef0b9a3ca72ac01ff96e3e240b78a30d/crates/oxc_parser/examples/multi-thread.rs#L82-L84

## How

- For `oxc_allocator::Vec`, switch to `allocator_api2::vec::Vec`, which has a covariant relationship with the allocator lifetime.
- For `oxc_allocator::Box`, use `std::ptr::NonNull` which is specifically designed to be covariant. I don't use `allocator_api2::boxed::Box` because it holds the allocator for dropping, so the size is bigger.

## Downside

Now that `oxc_allocator::Box` uses the unsafe `NonNull`. It has to be a private field to be safe. This make it impossible to do `Box(....)` pattern matching.

### DIFF
--- a/crates/oxc_allocator/src/arena.rs
+++ b/crates/oxc_allocator/src/arena.rs
@@ -5,8 +5,9 @@ use std::{
     self,
     fmt::{self, Debug, Formatter},
     hash::{Hash, Hasher},
+    marker::PhantomData,
     ops::{self, Deref},
-    ptr,
+    ptr::{self, NonNull},
 };
 
 use allocator_api2::vec;
@@ -19,7 +20,7 @@ use crate::Allocator;
 /// A Box without Drop.
 /// This is used for over coming self-referential structs.
 /// It is a memory leak if the boxed value has a `Drop` implementation.
-pub struct Box<'alloc, T: ?Sized>(pub &'alloc mut T);
+pub struct Box<'alloc, T: ?Sized>(NonNull<T>, PhantomData<&'alloc T>);
 
 impl<'alloc, T> Box<'alloc, T> {
     #[allow(unsafe_code)]
@@ -30,21 +31,29 @@ impl<'alloc, T> Box<'alloc, T> {
         // borrowed from some other reference. This in turn is because we never
         // construct an alloc::Box with a borrowed reference, only with a fresh
         // one just allocated from a Bump.
-        unsafe { ptr::read(self.0 as *mut T) }
+        unsafe { ptr::read(self.0.as_ptr()) }
+    }
+}
+
+impl<'alloc, T> Box<'alloc, T> {
+    pub fn new_in(x: T, alloc: &Allocator) -> Self {
+        Self(alloc.alloc(x).into(), PhantomData)
     }
 }
 
 impl<'alloc, T: ?Sized> ops::Deref for Box<'alloc, T> {
     type Target = T;
 
+    #[allow(unsafe_code)]
     fn deref(&self) -> &T {
-        self.0
+        unsafe { self.0.as_ref() }
     }
 }
 
 impl<'alloc, T: ?Sized> ops::DerefMut for Box<'alloc, T> {
+    #[allow(unsafe_code)]
     fn deref_mut(&mut self) -> &mut T {
-        self.0
+        unsafe { self.0.as_mut() }
     }
 }
 
@@ -72,7 +81,7 @@ where
     where
         S: Serializer,
     {
-        self.0.serialize(s)
+        self.deref().serialize(s)
     }
 }
 
@@ -85,6 +94,12 @@ impl<'alloc, T: Hash> Hash for Box<'alloc, T> {
 /// Bumpalo Vec
 #[derive(Debug, PartialEq, Eq)]
 pub struct Vec<'alloc, T>(vec::Vec<T, &'alloc Bump>);
+
+fn _variance_asserts() {
+    fn test_fn<'a: 'b, 'b>(program: Vec<'a, ()>) -> Vec<'b, ()> {
+        program
+    }
+}
 
 impl<'alloc, T> Vec<'alloc, T> {
     #[inline]
@@ -191,7 +206,7 @@ mod test {
     #[test]
     fn box_deref_mut() {
         let allocator = Allocator::default();
-        let mut b = Box(allocator.alloc("x"));
+        let mut b = Box::new_in("x", &allocator);
         let b = &mut *b;
         *b = allocator.alloc("v");
         assert_eq!(*b, "v");
@@ -200,7 +215,7 @@ mod test {
     #[test]
     fn box_debug() {
         let allocator = Allocator::default();
-        let b = Box(allocator.alloc("x"));
+        let b = Box::new_in("x", &allocator);
         let b = format!("{b:?}");
         assert_eq!(b, "\"x\"");
     }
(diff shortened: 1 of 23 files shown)


Boshen:
Are you working on something that requires `borrow_*` from `ouroboros`, or is this a soundness change?

I'm asking because I don't understand this topic, and need to know more about the context of this change.

crates/oxc_parser/examples/multi-thread.rs
```diff
@@ -34,7 +34,7 @@ struct AST {
     allocator: Allocator,
     source_text: Arc<str>,
     #[borrows(allocator, source_text)]
-    #[not_covariant]
+    #[covariant]
```
Boshen, line 37: 👋 Here's the covariance change.

crates/oxc_linter/src/rules/unicorn/prefer_regexp_test.rs
```diff
@@ -82,7 +82,7 @@ impl Rule for PreferRegexpTest {
                 };
 
                 // Check if the `test` of the for statement is the same node as the call expression.
-                if call_expr2.0 as *const _ != call_expr as *const _ {
+                if std::ptr::addr_of!(**call_expr2) != call_expr as *const _ {
```
Boshen, line 85: We may need a method for this, this is not obvious.

branchseer:
> Are you working on something that requires `borrow_*` from `ouroboros`, or is this a soundness change?
> 
> I'm asking because I don't understand this topic, and need to know more about the context of this change.

@Boshen Thank you for your time reviewing.

I was working on a side project using oxc and bumped[^1] into this restriction. I just think `borrow_*` would be nice to have, although it's not essential for my case and I don't think this is a soundness change.

Also on other day I saw rolldown is also [creating self-referencing ast types](https://github.com/rolldown/rolldown/blob/37b14422d02627f217e2ad09778973ab09e30698/crates/rolldown_oxc_utils/src/oxc_ast/mod.rs#L14), so maybe this can make their life a little easier as well.

I am no expert to unsafe rust. But seeing both std's and allocator_api2's `Box` and `Vec` are lifetime covariant gives me some confidence about this change.

[^1]: pun intended

Boshen (approved):

Boshen:
Thank you for working on this.

Merging because tests, rustc and miri all passed.

zakarumych:
`&mut T` is invariant over `T` because it is borrowed and when borrow ends `T` would be observed in it original form.
If `T` could be coerced to shorter lifetime and overwritten, original value would contain value with shorter lifetime than type's.


Example.
```rust

fn break_it<'a, 'b: 'a>(a: &mut &'b Vec<i32>, b: &'a Vec<i32>) {
  let coerced_a: &mut &'a Vec<i32> = a; // Would be valid if `&mut T` covariant over `T`.
  *coerced_a = b;
}

let x = vec![1, 2, 3];
let mut y: &Vec<i32> = &x;
break_it(&mut y, &vec![]);
println!("{y:?}"); // use after free 💥
```

This is not an issue with `Box` because it is owned value, you can't observe its content when `Box` over type with shortened lifetime drops.



```rust

fn break_it<'a, 'b: 'a>(a: Box<&'b Vec<i32>>, b: &'a Vec<i32>) {
  let coerced_a: Box<&'a Vec<i32>> = a; // Would be valid if `&mut T` covariant over `T`.
  *coerced_a = b;
}

let x = vec![1, 2, 3];
let mut y: &Vec<i32> = &x;
break_it(Box::new(y), &vec![]);
println!("{y:?}"); // Doesn't compile, `y` is no more.
```

I'd like to replace `Allocator` copy in `allocator_api2::boxed::Box` with `Deallocator` that would be ZST for `bumpallo` case.
But it requires existence of `Deallocator` and my proposal for it left unattended :)

Boshen:
Weird, this is causing a stack over only in the code coverage run https://github.com/oxc-project/oxc/actions/runs/8660479790/job/23749372302

crates/oxc_allocator/src/arena.rs
```diff
@@ -5,19 +5,22 @@ use std::{
...
+use bumpalo::Bump;
 use serde::{ser::SerializeSeq, Serialize, Serializer};
 
 use crate::Allocator;
 
 /// A Box without Drop.
 /// This is used for over coming self-referential structs.
 /// It is a memory leak if the boxed value has a `Drop` implementation.
-pub struct Box<'alloc, T: ?Sized>(pub &'alloc mut T);
+pub struct Box<'alloc, T: ?Sized>(NonNull<T>, PhantomData<&'alloc T>);
```
zakarumych, line 23: It should be `PhantomData<(T, &'alloc ())>` or something like this.
It should convey that `Box` owns `T`

Boshen: Landing in https://github.com/oxc-project/oxc/pull/2952

overlookmotel:
@branchseer Sorry to resurrect a long-dead issue, but can you please help me with something?

What is it that makes `allocator_api2::vec::Vec` covariant, whereas `bumpalo::collections::Vec` isn't? Looking at the source code of both, the only difference I can see is that the `alloc` field of `RawVec` is `&'a Bump` in `bumpalo`, whereas it's a generic `A: Allocator` in `allocator_api2`.

But in our case, the `Allocator` we use with `allocator_api2::vec::Vec` is a `&'a Bump` anyway. So what's the difference?

https://docs.rs/allocator-api2/0.2.21/src/allocator_api2/stable/raw_vec.rs.html#116-120
https://docs.rs/bumpalo/latest/src/bumpalo/collections/raw_vec.rs.html#53-57

Are you able to shed any light?

branchseer:
@overlookmotel 

> What is it that makes allocator_api2::vec::Vec covariant, whereas bumpalo::collections::Vec isn't?

 In  https://github.com/branchseer/oxc/commit/8bebc893b4a912ca053c54b376cb7d2d091c619a I swapped  `allocator_api2::vec::Vec` with `bumpalo::collections::Vec`. The covariant test assertion passes. So I was wrong and `bumpalo::collections::Vec` is indeed covariant. Sorry for the confusion:)

> And weirdly, neither of them includes a PhantomData<T> [which std::vec::Vec does](https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#76-79).

`std::vet::Vec` doesn't use `T` in its pointer type `Unique<u8>`, so `PhantomData<T>` is mandatory. `allocator_api2` and `bumpalo` always uses `T` in their pointer type `NonNull<T>`.

overlookmotel:
> In [branchseer@8bebc89](https://github.com/branchseer/oxc/commit/8bebc893b4a912ca053c54b376cb7d2d091c619a) I swapped `allocator_api2::vec::Vec` with `bumpalo::collections::Vec`. The covariant test assertion passes. So I was wrong and `bumpalo::collections::Vec` is indeed covariant. Sorry for the confusion:)

OK great. Thanks for investigating. Maybe you were right at the time and `bumpalo` changed something since last year. 🤷

> `std::vet::Vec` doesn't use `T` in its pointer type `Unique<u8>`, so `PhantomData<T>` is mandatory. `allocator_api2` and `bumpalo`'s `Vec` already uses `T` in their pointer type `NonNull<T>`.

Ah ha. Yes, of course. I missed the difference `NonNull<T>` vs `Unique<u8>`.

Issue #2934, opened 2024-04-10
Port missed test cases for existing typescript-ESLint rules


mysteryven:
We ignored test cases that wrapped in `noFormat` previously, for example, [`// @ts-expect-error    : TS1234 because xyz`](https://github.com/typescript-eslint/typescript-eslint/blob/da006b1c9bb758b51c52c8b5af4ce43b44976e3f/packages/eslint-plugin/tests/rules/ban-ts-comment.test.ts#L410) hadn't been added. It may bring potential bugs.

These are the rules we need to check.(hope just adding test cases is enough)

- [x] ban-ts-comment.test.ts:14
- ~ban-types.test.ts:5~
- [x] no-empty-interface.test.ts:2
- ~no-extra-non-null-assertion.test.ts:5~
- ~no-non-null-asserted-optional-chain.test.ts:7~
- ~no-unnecessary-type-constraint.test.ts:5~
- ~prefer-function-type.test.ts:3~

The delete items means has been added already.

ban-types are also not added, I assume using [custom options](https://typescript-eslint.io/rules/ban-types/#types) in ban-type is a rare condition, the missed cases need [custom options](https://github.com/typescript-eslint/typescript-eslint/blob/a6ab2cb9be5a09498e953896e7fad04b2c53ea42/packages/eslint-plugin/tests/rules/ban-types.test.ts#L614-L620) which we don't support for now.

JoSeBu1:
Hi!, 

I have added missing tests for:
`no-empty-interface`
`no-extra-non-null-assertion`
`no-unnecessary-type-constraint`

Some rules have all test cases, maybe added manually:
`prefer-function-type`
`no-non-null-asserted-optional-chain`

And some others have tests no passing when new tests are added:
`ban-ts-comment`
`ban-types`

I attempted to fix it but it is not possible for me now.

Boshen:
Is this all done and can be closed @mysteryven ?

mysteryven:
[**@Boshen**](https://github.com/Boshen) commented on [Apr 19, 2024, 12:37 PM GMT+8](https://github.com/oxc-project/oxc/issues/2934#issuecomment-2065741558 "2024-04-19T04:37:02Z - Replied by Github Reply Comments"):
> Is this all done and can be closed @mysteryven ?
> 
> _Originally posted by @Boshen in https://github.com/oxc-project/oxc/issues/2934#issuecomment-2065741558_

[...]

PR #2975, opened 2024-04-14
feat(linter): add missing test cases to no-extra-non-null-assertion

JoSeBu1:
Based on this https://github.com/oxc-project/oxc/issues/2934 I added missing test cases


mysteryven:
It looks like the new version missed some cases(The old version has 8 items). it's my bad, this rule also doesn't need to port test cases.

PR #2976, opened 2024-04-14
feat(linter): add missing test cases to no-unnecessary-type-constraint

JoSeBu1:
Based on this https://github.com/oxc-project/oxc/issues/2934 I added missing test cases


JoSeBu1:
Reviewing it I saw that test cases are the same

mysteryven:
I'm sorry, I thought people may not consider it.

PR #2973, opened 2024-04-14
feat(linter): add missing test cases to no-empty-interface and add config

JoSeBu1:
Based on this https://github.com/oxc-project/oxc/issues/2934 I added missing test cases and rule configuration

PR #2979, opened 2024-04-14
fix(linter/no-empty-interface): add missing test

JoSeBu1:
As requested add new test https://github.com/oxc-project/oxc/pull/2973#discussion_r1564728188


JoSeBu1:
I think that maybe the error indicated in the snapshot is not correct

mysteryven (approved):

PR #2997, opened 2024-04-16
feat(transformer/typescript): support for transform enum

Dunqing:
The current implementation is copied from the previous implementation

### DIFF
--- a/crates/oxc_transformer/src/utils.rs
+++ b/crates/oxc_transformer/src/utils.rs
@@ -1,0 +1,12 @@
+use oxc_syntax::{identifier::is_identifier_name, keyword::is_keyword};
+
+pub fn is_valid_identifier(name: &str, reserved: bool) -> bool {
+    if reserved && (is_keyword(name) || is_reserved_word(name, true)) {
+        return false;
+    }
+    is_identifier_name(name)
+}
+
+pub fn is_reserved_word(name: &str, in_module: bool) -> bool {
+    (in_module && name == "await") || name == "enum"
+}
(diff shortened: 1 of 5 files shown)


Dunqing:
* **#2997** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/2997?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/typescript/annotations.rs
```diff
@@ -325,4 +327,236 @@ impl<'a> TypeScriptAnnotations<'a> {
...
+    ///   X
+    /// }
+    /// ```
+    /// ```JavaScript
+    /// var Foo = ((Foo) => {
+    ///   const X = 0; Foo[Foo["X"] = X] = "X";
+    ///   return Foo;
+    /// })(Foo || {});
+    /// ```
+    fn transform_ts_enum(
```
Boshen, line 342: Can we isolate this in a new file?

crates/oxc_transformer/src/utils.rs
```diff
@@ -0,0 +1,66 @@
...
+    "import",
+    "null",
+    "true",
+    "false",
+    "in",
+    "instanceof",
+    "typeof",
+    "void",
+    "delete",
+];
```
Boshen, line 51: These should be in `oxc_syntax`. crates/oxc_syntax/src/keyword.rs

crates/oxc_transformer/src/typescript/annotations.rs
milesj: Yeah I was gonna do this in an enums file.

Dunqing: Yes, i moved this to `enum.rs`

Boshen (approved):

PR #3005, opened 2024-04-16
fix(transformer/typescript): modifiers should not be removed

Dunqing:
`codegen` relies on the modifier to decide whether to print or not. For example, if declare is present, nothing will be printed if typescript is not enabled.

We may need to solve this problem in a way that doesn't rely on modifers.


Dunqing:
* **#3006** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3006?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3005** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):

PR #3009, opened 2024-04-16
feat(transformer/react): reports duplicate __self/__source prop error

Dunqing:


Dunqing:
* **#3009** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/transform_conformance/src/test_case.rs
```diff
@@ -17,6 +17,12 @@ use oxc_transformer::{ReactOptions, TransformOptions, Transformer, TypeScriptOpt
 
 use crate::{fixture_root, packages_root, TestRunnerEnv, PLUGINS_NOT_SUPPORTED_YET};
 
+static ERROR_MESSAGE_MAPPING: phf::Map<&'static str, &'static str> = phf::phf_map! {
```
Boshen, line 20: Do we need a phf? Just a simple match should work.

milesj (approved):

crates/oxc_transformer/src/react/jsx_self/mod.rs
```diff
@@ -29,6 +33,17 @@ impl<'a> ReactJsxSelf<'a> {
     }
 
     pub fn transform_jsx_opening_element(&self, elem: &mut JSXOpeningElement<'a>) {
+        for item in &elem.attributes {
```
Boshen, line 36: I think we should avoid this loop on every opening element, this should be checked while transforming.

And I wonder whether we should only emit the diagnostic only once per file, otherwise there's going to be a lot. Can you double check with babel?

tasks/transform_conformance/Cargo.toml
```diff
@@ -35,3 +35,4 @@ pico-args  = { workspace = true }
 indexmap   = { workspace = true }
 serde      = { workspace = true }
 serde_json = { workspace = true }
+phf        = { workspace = true }
```
Boshen, line 38:
```suggestion
```

crates/oxc_transformer/src/react/jsx_self/mod.rs
Dunqing: > And I wonder whether we should only emit the diagnostic only once per file, otherwise there's going to be a lot. Can you double check with babel?

I don't see the logic of babel only reporting once https://github.com/babel/babel/blob/acf3d17fdfe150a270c822581b709dddac4548ce/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L529-L531

Boshen (approved):

PR #3014, opened 2024-04-17
feat(transformer/react): correct import binding names

Dunqing:


Dunqing:
* **#3014** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3014?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/lib.rs
```diff
@@ -79,6 +82,7 @@ impl<'a> Transformer<'a> {
 
 impl<'a> VisitMut<'a> for Transformer<'a> {
     fn visit_program(&mut self, program: &mut Program<'a>) {
+        self.ctx.borrow_mut().naming.visit_program(program);
```
Dunqing, line 85: We need to collect all bindings first.

crates/oxc_transformer/src/context.rs
```diff
@@ -10,9 +10,9 @@ use oxc_ast::AstBuilder;
 use oxc_diagnostics::Error;
 use oxc_semantic::Semantic;
 
-use crate::{helpers::module_imports::ModuleImports, TransformOptions};
+use crate::{helpers::module_imports::ModuleImports, naming::TransformerNaming, TransformOptions};
 
-pub type Ctx<'a> = Rc<TransformCtx<'a>>;
```
Dunqing, line 15: `TransformerNaming` needs to mutation.

In a previous implementation. All the `TransformerCtx` fields that cannot be copied are wrapped in `Rc`
https://github.com/oxc-project/oxc/blob/31ed532b790691b843182768351477dd26428b94/crates/oxc_transformer/src/context.rs#L15-L22

Boshen:
Let me iterate on this tomorrow.

crates/oxc_transformer/src/context.rs
milesj: This is a big change.

It probably makes more sense to use interior mutability in `TransformerNaming`, instead of adding all these borrow calls everywhere.

Boshen:
I'll find sometime to figure out recreating scopes and bindings during visits, probably this weekend.

crates/oxc_transformer/src/context.rs
Dunqing: > It probably makes more sense to use interior mutability in `TransformerNaming`, instead of adding all these borrow calls everywhere.

Yes, I agree. When we finally consider this PR, I'll refactor it.

milesj:
> I'll find sometime to figure out recreating scopes and bindings during visits, probably this weekend.

The problem we had before is that the nodes don't have IDs, so there was no easy way to back reference. That's why @rzvxa was trying to add IDs to everything.

rzvxa:
> > I'll find sometime to figure out recreating scopes and bindings during visits, probably this weekend.
> 
> The problem we had before is that the nodes don't have IDs, so there was no easy way to back reference. That's why @rzvxa was trying to add IDs to everything.

We still can add this, I didn't receive any pushback while implementing it in #2876 and #2818, I just stopped to figure out the traverse first; Now that the traverse is almost sorted out we may want to revisit the ast_node_id.
Since now we have a parent field for all nodes because of the traverse, We can change the parent with an AstPath structure containing both the opaque parent's pointer and our own ID.

rzvxa:
I suspect that by adding node_ids to the nodes we can do a minimal refactor for linters which removes a bunch of useless checks and improves the performance by like 10% to 20%.

We can still provide a way to handle AstNodes so we don't have to refactor them all from day one, But we can offer additional traits to intercept the inner node without needing extra match on AstKind.

rzvxa:
I'm just thinking this one out loud;

Since now we have a parent pointer - even though it is opaque - we might be able to map between that and our own ast_node_id so we don't have to add an extra `usize` to our nodes.

For example if we always allocate all of our nodes in the same buffer and we have the root of our buffer we can find our relative position to that and use it to fetch our relative data in a different buffer similar to an indexing operation.

Dunqing:
Finally, we have a `generateUid` api. #3394

PR #3017, opened 2024-04-18
fix(semantic/jsdoc): Skip parsing `@` inside of backticks

leaysgur:
This PR aims to support these cases.

````js
/**
 * This is normal comment, `@xxx` should not parsed as tag.
 *
 * @example ```ts
    // @comment
    @decoratorInComment
    class Foo { }
   ```
 */
````

Only `@example` should be parsed as tag.

### DIFF
--- a/crates/oxc_semantic/src/jsdoc/parser/parse.rs
+++ b/crates/oxc_semantic/src/jsdoc/parser/parse.rs
@@ -24,6 +24,7 @@ pub fn parse_jsdoc(source_text: &str, jsdoc_span_start: u32) -> (JSDocCommentPar
     let mut comment_found = false;
     // Parser local offsets, not for global span
     let (mut start, mut end) = (0, 0);
+
     let mut chars = source_text.chars().peekable();
     while let Some(ch) = chars.next() {
         let can_parse = !(in_braces || in_backticks);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by Boshen — APPROVED** [review]

File: crates/oxc_semantic/src/jsdoc/parser/parse.rs [hunk 1]
```diff
@@ -19,14 +19,29 @@ pub fn parse_jsdoc(source_text: &str, jsdoc_span_start: u32) -> (JSDocCommentPar
...
+        let can_parse = !(in_braces || in_backticks);
         match ch {
+            // NOTE: For now, only odd backtick(s) are handled.
+            // - 1 backtick: inline code
+            // - 3, 5, ... backticks: code fence
+            // Not so common but technically, major markdown parser can handle 3 or more backticks as code fence.
+            // (for nested code blocks)
+            // But for now, 4, 6, ... backticks are not handled here to keep things simple...
+            '`' => {
+                if chars.peek().map_or(false, |&c| c != '`') {
```
Boshen, line 38: Here's a cool trick: https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and

Boshen: Or `!matches(chars.peek(), Some('`'))`

leaysgur: Thanks! 😉  https://github.com/oxc-project/oxc/pull/3017/commits/38dd7f5107681747c61c78ffbfae2b486a90cd61

Boshen (approved):

PR #3030, opened 2024-04-19
feat(linter): no barrel file.

### LINKED ISSUE #3004: feat(linter): noBarrelFile
https://biomejs.dev/linter/rules/no-barrel-file/

Biome only checks for `export *` in a file.

In oxlint, we can do better when `--import-plugin` is enabled, by displaying the total number of dependencies pulled into a given file by walking 

https://github.com/oxc-project/oxc/blob/67045467c7831f063084576124e830e97dc59123/crates/oxc_syntax/src/module_record.rs#L41

You can find similar plugins in https://github.com/oxc-project/oxc/tree/main/crates/oxc_linter/src/rules/import

rzvxa:
closes #3004 

I've based it on [this plugin](https://github.com/thepassle/eslint-plugin-barrel-files/blob/main/lib/rules/avoid-barrel-files.js) instead of [biome](https://github.com/biomejs/biome/blob/4130ff0e4c56c4944236b11f2b4f34ca6e8ca03a/crates/biome_js_analyze/src/lint/performance/no_barrel_file.rs#L70) Since the original plugin is less likely to detect a false positive.

I didn't understand your statement [here](https://github.com/oxc-project/oxc/issues/3004#issue-2245574895), Where you've mentioned:

> In oxlint, we can do better when --import-plugin is enabled, by displaying the total number of dependencies pulled into a given file by walking

I would appreciate it if you expand upon it.

------

#### Edit:

I've added it under `eslint-plugin-import` even though it is not; I wasn't sure If I should create a whole new category for this single rule, It is a different story if we would like to adopt the other rules in [here](https://github.com/thepassle/eslint-plugin-barrel-files/tree/main/lib/rules).

Also, check my diagnosis messages; It is my first contribution to the linters so I'm just not familiar enough with the conventions of messages, rules, etc.

### DIFF
--- a/crates/oxc_linter/src/rules/oxc/no_barrel_file.rs
+++ b/crates/oxc_linter/src/rules/oxc/no_barrel_file.rs
@@ -76,18 +76,29 @@ fn test() {
         r#"export type * from "foo";"#,
         r#"export type { foo } from "foo";"#,
         r#"export type * from "foo";
-           export type { bar } from bar;"#,
+           export type { bar } from "bar";"#,
+        r#"import { foo, bar, baz } from "../feature";
+           export { foo };
+           export { bar };"#,
     ];
 
     let fail = vec![
         r#"export * from "foo";
-           export * from "bar";"#,
+           export * from "bar";
+           export * from "baz";
+           export * from "qux";"#,
         r#"export { foo } from "foo";
-           export { bar } from "bar";"#,
-        r#"export { default as module1 } from "./module1";"#,
-        r#"export { foo, type Bar } from "foo";"#,
-        r#"import { foo, bar, baz } from "../feature";
-           export { foo, bar, baz };"#,
+           export { bar } from "bar";
+           export { baz } from "baz";
+           export { qux } from "qux";"#,
+        r#"export { default as module1 } from "./module1";
+           export { default as module2 } from "./module2";
+           export { default as module3 } from "./module3";
+           export { default as module4 } from "./module4";"#,
+        r#"export { foo, type Foo } from "foo";
+           export { bar, type Bar } from "bar";
+           export { baz, type Baz } from "baz";
+           export { qux, type Qux } from "qux";"#,
     ];
 
     Tester::new(NoBarrelFile::NAME, pass, fail)
--- a/crates/oxc_linter/src/rules/oxc/no_barrel_file.rs
+++ b/crates/oxc_linter/src/rules/oxc/no_barrel_file.rs
@@ -1,0 +1,97 @@
+use oxc_ast::{ast::Statement, AstKind};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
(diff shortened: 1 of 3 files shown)


rzvxa:
* **#3030** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/rules/import/no_barrel_file.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+
+    let pass = vec![
+        // r#"export type * from "foo";"#,
```
rzvxa, line 79: Why these don't pass? Are we lacking the correct syntax for the export type?

@Boshen

Boshen:
I forced pushed some changes. The remaining check is on `export type *`.

Boshen:
> I didn't understand your statement https://github.com/oxc-project/oxc/issues/3004#issue-2245574895, Where you've mentioned:

In oxlint, we can do better when --import-plugin is enabled, by displaying the total number of dependencies pulled into a given file by walking

This is just a bonus. I want people see how many modules a simple `export *` imports.

This can be calculated by recursively walking `loaded_modules` and tally the total number of modules.

https://github.com/oxc-project/oxc/blob/67045467c7831f063084576124e830e97dc59123/crates/oxc_syntax/src/module_record.rs#L41

rzvxa:
> I forced pushed some changes. The remaining check is on `export type *`.

Oh thanks, I didn't think of file extension since I had it copy pasted from a similar rule.

> This can be calculated by recursively walking loaded_modules and tally the total number of modules.

Can this be an opt-in thing? Since with large import graphs (which is the primary characteristic for the barrel files), it is going to make a small but noticeable dent in performance.

In either case, I'll add it in, Let me know if you like the idea of making it an optional thing or not.

crates/oxc_linter/src/rules/oxc/no_barrel_file.rs
```diff
@@ -0,0 +1,108 @@
...
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error(
+    "eslint-plugin-import(no-barrel-file): \
```
Boshen, line 13:
```suggestion
    "oxc(no-barrel-file): \
```

Boshen (approved):

Boshen:
### Merge activity

* **Apr 21, 10:09 PM EDT**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/3030).

PR #3033, opened 2024-04-19
new: Start on es2022 preset. Implement `class-static-block`.

milesj:


crates/oxc_transformer/src/es2022/class_static_block/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn transform_class_body(&mut self, body: &mut ClassBody<'a>) {
+        let mut private_names = FxHashSet::default();
+        let ast = &self.ctx.ast;
+
+        for stmt in &body.body {
+            if stmt.is_private() {
+                if let Some(name) = stmt.property_key().and_then(|k| k.private_name()) {
+                    private_names.insert(name.as_str().to_owned());
```
milesj, line 25: I tried to avoid the clones but kept running into borrow issues.

crates/oxc_transformer/src/es2022/class_static_block/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+            id
+        };
+
+        let stmts = ast.move_vec(&mut body.body).into_iter().map(|stmt| {
+            if let ClassElement::StaticBlock(mut block) = stmt {
+                let private_id = generate_id();
+
+                // We special-case the single expression case to avoid the iife,
+                // since it's common.
+                let replacement = if block.body.len() == 1
```
milesj, line 50: Is there a better way to handle this branch??

crates/oxc_transformer/src/lib.rs
```diff
@@ -43,9 +45,12 @@ use crate::{
 
 pub struct Transformer<'a> {
     ctx: Ctx<'a>,
+
     // NOTE: all callbacks must run in order.
     x0_typescript: TypeScript<'a>,
     x1_react: React<'a>,
+    // x2_decorators,
+    x3_es2022: Es2022<'a>,
```
milesj, line 53: Ordering here is weird. When we add es2024, es2025, etc, all the numbers will shift. We probably need a different ordering system.

Boshen:
Thank you for the PR! Are we settled on everything, or should we refactor the current a bit more before getting into other presets?

milesj:
@Boshen We're still waiting on that PR to see if it's viable.

magic-akari:
> I think we should support `babel-preset-env` instead of es20xx preset. See https://babeljs.io/docs/env

I agree that, compared to the es version, preset-env is more practical.

Moreover, I believe that the current Babel implementation completely separates env from the es version.
We could allow env to accept two types of inputs, treating the es version as a kind of env, accepting both `chrome 90` and `es2020`.

However, when we implement the code logic, it is OK to classify according to the ES version; this is merely a matter of code organization.

milesj:
Yes this is purely for organization. It's also required for tranformers to run in the correct order: typescript -> react -> es2024 -> es2022 -> es2021, etc.

And furthermore, transforms within a preset must also be ran in a specific order.

Dunqing:
> Yes this is purely for organization. It's also required for tranformers to run in the correct order: typescript -> react -> es2024 -> es2022 -> es2021, etc.
> 
> And furthermore, transforms within a preset must also be ran in a specific order.

I got it, Thanks!

PR #3051, opened 2024-04-21
perf(ast): reduce indirection in AST types

### LINKED ISSUE #3048: Reduce indirection in AST types
AST types in a few places have what seems like unnecessary indirection via `Vec<Box<T>>`. I believe `Vec<T>` should suffice.

The downsides of `Vec<Box<T>>` are:

1. An extra level of indirection.
2. The `T`s are scattered in memory rather than in a contiguous block.

Only advantage I can see is that if the `Vec` grows repeatedly, reallocation is cheaper when entries of the `Vec` are `Box<T>` (8 bytes) rather than `T` (generally larger). But if this is a genuine advantage, we could probably be using that optimization elsewhere e.g. `Vec<'a, TemplateElement<'a>>` in `TemplateLiteral`.

@Boshen Is there a rationale for `Vec<Box<T>>`? Or just an accidental "how it ended up" as the AST evolved?

Current uses of `Vec<Box<T>>` are:

```rs
pub struct Class<'a> {
    pub implements: Option<Vec<'a, Box<'a, TSClassImplements<'a>>>>,
}
```

```rs
pub struct TSTypeParameterDeclaration<'a> {
    pub params: Vec<'a, Box<'a, TSTypeParameter<'a>>>,
}
```

```rs
pub struct TSInterfaceDeclaration<'a> {
    pub extends: Option<Vec<'a, Box<'a, TSInterfaceHeritage<'a>>>>,
}
```

```rs
pub struct TSIndexSignature<'a> {
    pub parameters: Vec<'a, Box<'a, TSIndexSignatureName<'a>>>,
}
```

overlookmotel:
Fixes #3048.

No apparent change on benchmarks. Likely these TS features are not much used in the benchmark fixtures.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -2025,7 +2025,7 @@ pub struct Class<'a> {
     pub body: Box<'a, ClassBody<'a>>,
     pub type_parameters: Option<Box<'a, TSTypeParameterDeclaration<'a>>>,
     pub super_type_parameters: Option<Box<'a, TSTypeParameterInstantiation<'a>>>,
-    pub implements: Option<Vec<'a, Box<'a, TSClassImplements<'a>>>>,
+    pub implements: Option<Vec<'a, TSClassImplements<'a>>>,
     pub decorators: Vec<'a, Decorator<'a>>,
     /// Valid Modifiers: `export`, `abstract`
     pub modifiers: Modifiers<'a>,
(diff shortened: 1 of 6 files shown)


overlookmotel:
* **#3051** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3051?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):

PR #3053, opened 2024-04-21
refactor(bench): reuse allocator in parser + lexer benchmarks

overlookmotel:
Re-use allocator in parser + lexer benchmarks.

I believe this is the recommended usage when parsing a bunch of files - to re-use one allocator rather than create a fresh one for each run, so it makes sense to me that this is what the benchmark should measure.

Doesn't show much difference on CodSpeed because it only runs the benchmark once, and it treats allocations as free anyway. But I imagine the difference may show up a bit more in a standard criterion benchmark.

### DIFF
--- a/crates/oxc_allocator/src/lib.rs
+++ b/crates/oxc_allocator/src/lib.rs
@@ -1,4 +1,7 @@
-use std::{convert::From, ops::Deref};
+use std::{
+    convert::From,
+    ops::{Deref, DerefMut},
+};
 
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#3053** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3053?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Issue #3047, opened 2024-04-21
Reduce AST enum type sizes


overlookmotel:
In general, Oxc's AST types which are enums have all their variants boxed. e.g.:

```rs
pub enum Expression<'a> {
    BooleanLiteral(Box<'a, BooleanLiteral>),
    NullLiteral(Box<'a, NullLiteral>),
    /* ... */
}
```

The enum then only weighs 16 bytes, and is used unboxed in other types' fields:

```rs
pub struct UnaryExpression<'a> {
    pub argument: Expression<'a>,
    /* ... */
}
```

There are however some exceptions to this rule, where enums are larger than they need to be because not all variants are boxed, and some variants are larger than others. This results in less compact storage in the arena, due to excess padding.

Boxing the variants in these cases should save on memory. I imagine it *may* cost a little in performance in the parser due to extra calls to `alloc`, but in return traversing the AST should be cheaper, as the data is more compact, so less cache misses.

However, I am not 100% sure of that in all cases. What I think should be done for each type is listed below, but we should apply each change individually, and check benchmarks to ensure performance does not degrade.

The affected types are:

#### `ImportDeclarationSpecifier`
Variant sizes: 40, 40, 80
Unboxed variants: `ImportSpecifier` (struct), `ImportDefaultSpecifier` (struct), `ImportNamespaceSpecifier` (struct)
Dependents: `Vec<ImportDeclarationSpecifier>` in `ImportDeclaration` (struct)
=> Box all variants.

#### `MemberExpression`
Variant sizes: 48, 56, 56
Unboxed variants: `ComputedMemberExpression` (struct), `StaticMemberExpression` (struct), `PrivateFieldExpression` (struct)
Dependents: `Box<MemberExpression>` in `Expression` (enum), `SimpleAssignmentTarget` (enum), `ChainElement` (enum)
=> Not sure. Always used within box as another enum variant. Would be better to box variants and flatten this enum into parent enums, but that's tricky.

#### `ModuleExportName`
Variant sizes: 24, 24
Unboxed variants: `IdentifierName` (struct), `StringLiteral` (struct)
Dependents:
- `ImportSpecifier` (struct)
- `Option<ModuleExportName>` in `ExportAllDeclaration` (struct)
- `ExportDefaultDeclaration` (struct)
- `ExportSpecifier` (struct)

[...]

PR #3058, opened 2024-04-21
perf(ast): box enum variants

overlookmotel:
Box all enum variants for JSX types (`JSXAttributeName`, `JSXAttributeValue`, `JSXChild`, `JSXElementName`, `JSXMemberExpressionObject`). Part of #3047.

I'm not sure how to interpret the benchmark results. As I said on #3047:

> I imagine it may cost a little in performance in the parser due to extra calls to `alloc`, but in return traversing the AST should be cheaper, as the data is more compact, so less cache misses.

Sure enough, there is a small impact (1%) on the 2 parser benchmarks for JSX files. However, the other benchmarks have too much noise in them to see whether this is repaid in a speed up on transformer etc, especially as the transformer benchmarks also include parsing.

What do you think @Boshen?

### DIFF
--- a/crates/oxc_ast/src/ast/jsx.rs
+++ b/crates/oxc_ast/src/ast/jsx.rs
@@ -212,8 +212,8 @@ pub enum JSXAttributeName<'a> {
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(untagged))]
 pub enum JSXAttributeValue<'a> {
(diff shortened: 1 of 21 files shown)


overlookmotel:
* **#3058** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3058?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):

PR #3070, opened 2024-04-22
feat(linter): eslint/no-await-in-loop

woai3c:
[no-await-in-loop](https://eslint.org/docs/latest/rules/no-await-in-loop)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_await_in_loop.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_await_in_loop.rs
@@ -1,0 +1,233 @@
+use oxc_ast::{
+    ast::{Expression, Statement},
+    AstKind,
+};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint(no-await-in-loop): Unexpected `await` inside a loop.")]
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/no_await_in_loop.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///   const userRecord = await getUserRecord(user);
+    /// }
+    /// ```
+    ///
+    /// Good:
+    /// ```javascript
+    /// await Promise.all(users.map(user => getUserRecord(user)));
+    /// ```
+    NoAwaitInLoop,
+    correctness
```
mysteryven, line 45: Is [`perf`](https://github.com/oxc-project/oxc/blob/32504cadb745c4f81be89fa74fb317822daee1ed/crates/oxc_linter/src/rule.rs#L42-L43) much better?

crates/oxc_linter/src/snapshots/no_await_in_loop.snap
```diff
@@ -0,0 +1,87 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: no_await_in_loop
+---
+  ⚠ eslint(no-await-in-loop): Unexpected `await` inside a loop.
+   ╭─[no_await_in_loop.tsx:1:38]
+ 1 │ async function foo() { while (baz) { await bar; } }
```
mysteryven, line 7: AwaitExpression may have a long argument, Shall we just point to the keyword `await`?

mysteryven (commented):
Looks good to me.

mysteryven (approved):
Nice work!

PR #3095, opened 2024-04-25
chore: cleanup the dependencies on `static_assertions` and `oxc_index`.

rzvxa:
We used to export `static_assertions` as part of the `oxc_index`. It would've made sense back when it was only a vessel for exporting other crates - although even then it wouldn't make much sense other than being convenient - Now with it turning into a port of `index_vec` and potentially getting bigger as the result of specific needs of the project; It makes much more sense to stop exporting it from `oxc_index` and use the crate directly in places that used to use what `oxc_index` were exporting.

PS: we may want to follow up this with an `oxc_asset` crate containing our own set of assertion tools which would also export `static_assertions`.


rzvxa:
* **#3095** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
@Boshen Suggestion: rename `oxc_index` to `oxc_indexed`, In my opinion, `index vec` makes sense but `oxc index` doesn't, while `oxc indexed` is more meaning full.

Just a nitpick I wanted to throw out there.

Boshen (approved):

Boshen:
This was setup like rustc, but I can't find the source 😅 

`static_assertions` is only used in a few places (not like rustc it's everywhere), so I think we should be fine for now without an extra `oxc_xxx` crate

PR #3097, opened 2024-04-25
feat(linter): typescript-eslint/prefer-enum-initializers

todor-a:


crates/oxc_linter/src/rules/typescript/prefer_enum_initializers.rs
```diff
@@ -0,0 +1,104 @@
...
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("typescript-eslint(prefer-enum-initializers):")]
+#[diagnostic(severity(warning), help(""))]
```
mysteryven, line 13: Aha, we need to fill error message (line 12) and help message (line 13).

crates/oxc_linter/src/rules/typescript/prefer_enum_initializers.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// ### Example
+    /// ```typescript
+    /// // wrong, the value of `Close` is not constant
+    /// enum Status {
+    ///  Open = 1,
+    ///  Close,
+    /// }
+    /// ```
+    PreferEnumInitializers,
+    correctness
```
mysteryven, line 35: I prefer [`pedantic`](https://github.com/oxc-project/oxc/blob/0185eb2edcb3b01afb701cdab34643a26c2ad17e/crates/oxc_cli/src/command/lint.rs#L82-L83). As the document says, some care about enums having implicit values. 🤔

https://typescript-eslint.io/rules/prefer-enum-initializers/#when-not-to-use-it

crates/oxc_linter/src/rules/typescript/prefer_enum_initializers.rs
todor-a: ugh, I keep forgetting those. thanks, fixed!

crates/oxc_linter/src/rules/typescript/prefer_enum_initializers.rs
todor-a: i've used `pedantic`, although I do not necessarily agree. this rule will very much help avoid errors with enum members being move around.

mysteryven (approved):
Thank you!

PR #3102, opened 2024-04-25
refactor(transformer/typescript): reimplementation of Enum conversion based on Babel

Dunqing:
The remaining test cases will perform better with a scope implementation, and while we can implement them without the scope, it still requires us to do what the scope did.


Dunqing:
* **#3102** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3102?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -227,4 +280,262 @@ impl<'a> TypeScript<'a> {
...
+            }
+            BinaryOperator::ShiftRight | BinaryOperator::ShiftRightZeroFill => {
+                Some(ConstantValue::F64((left as i64 >> right as i64) as f64))
+            }
+            BinaryOperator::ShiftLeft => {
+                Some(ConstantValue::F64(((left as i64) << right as i64) as f64))
+            }
+            BinaryOperator::BitwiseXOR => {
+                Some(ConstantValue::F64((left as i64 ^ right as i64) as f64))
+            }
```
magic-akari, line 439: I think we should use `i32` instead of `i64` here, considering the bitwise operations in JavaScript.

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -202,21 +264,12 @@ impl<'a> TypeScript<'a> {
...
-                    value: 1.0,
-                    raw: "1",
-                    base: NumberBase::Decimal,
-                });
-
-                self.ctx.ast.binary_expression(SPAN, one, BinaryOperator::Addition, self_ref)
-            };
         }
 
+        self.enums.insert(enum_name.clone(), previous_enum_members.clone());
```
magic-akari, line 271: For the following expression:
```TypeScript
enum Foo {
    A = 1,
}
```
We need to store not only the correspondence of A -> 1, but also the correspondence of 1 -> A.
This way, `Foo[1]` can return `A`.

crates/oxc_transformer/src/typescript/enum.rs
Dunqing: Thank you, you're right! Now I know why I'm casting to `i32`, it's defined in the [spec](https://tc39.es/ecma262/#sec-numberbitwiseop).

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -227,4 +282,263 @@ impl<'a> TypeScript<'a> {
...
+            };
+
+            let right_string = match right {
+                ConstantValue::NumericLiteral((raw, _)) => raw.to_string(),
+                ConstantValue::String(str) => str,
+                ConstantValue::F64(v) => v.to_string(),
+                ConstantValue::Identifier(_) => unreachable!(),
+            };
+
+            return Some(ConstantValue::String(format!("{left_string}{right_string}")));
```
magic-akari, line 413: In JavaScript, numbers are converted to strings using a distinct method compared to Rust or C/C++. For instance, the number 1e21 is rendered as `"1e21"` by JavaScript's `toString` method, while Rust would format it as `"1000000000000000000000f"`. 
Similarly, the representation of Infinity varies: JavaScript outputs `"Infinity"` but Rust uses `"inf"`.

magic-akari: I mean this line.

https://github.com/oxc-project/oxc/blob/feaf394e4707fd5546bfa42f40e49246700db933/crates/oxc_transformer/src/typescript/enum.rs#L402

Dunqing: Thank you, I will add some tests to cover this.

magic-akari (approved):

PR #3107, opened 2024-04-26
feat(linter): add more "ban-ts-comment" test cases.

woai3c:
https://github.com/oxc-project/oxc/issues/2934

| name           | passed | failed |
| -------------- | ------ | ------ |
| ts-expect-error| 14     | 20     |
| ts-ignore      | 15     | 20     |
| ts-nocheck     | 11     | 9      |
| ts-check       | 11     | 8      |
| total          | 51     | 57

### DIFF
--- a/crates/oxc_linter/Cargo.toml
+++ b/crates/oxc_linter/Cargo.toml
@@ -43,6 +43,7 @@ convert_case  = { workspace = true }
 language-tags = { workspace = true }
 mime_guess    = { workspace = true }
 url           = { workspace = true }
+unicode-segmentation = "1.11.0"
 
 rust-lapper         = { workspace = true }
 once_cell           = { workspace = true }
--- a/crates/oxc_linter/Cargo.toml
+++ b/crates/oxc_linter/Cargo.toml
@@ -43,7 +43,6 @@ convert_case  = { workspace = true }
 language-tags = { workspace = true }
 mime_guess    = { workspace = true }
 url           = { workspace = true }
-unicode-segmentation = "1.11.0"
 
 rust-lapper         = { workspace = true }
 once_cell           = { workspace = true }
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/Cargo.toml
```diff
@@ -43,6 +43,7 @@ convert_case  = { workspace = true }
 language-tags = { workspace = true }
 mime_guess    = { workspace = true }
 url           = { workspace = true }
+unicode-segmentation = "1.11.0"
```
mysteryven, line 46: Need call @Boshen for help, this crate is added for compute [`minimumDescriptionLength`](https://typescript-eslint.io/rules/ban-ts-comment#minimumdescriptionlength), how we deal with this?

Boshen: What's the problem with a normal `.len()`? I don't think it's a problem if there is 1 or 2 characters in difference.

woai3c:
```ts
/**
 * @ts-expect-error 👨‍👩‍👧‍👦 */
```
If use the normal `.len()`, the length of `👨‍👩‍👧‍👦` is 27. And use `unicode_segmentation` to calcute the the length of `👨‍👩‍👧‍👦`, it would be 3.

image

So, If we use the normal `.len()` and description has emoji, These test cases about `minimumDescriptionLength` will be failed.

Boshen: Let's ignore this test for now, it's a rare case.

woai3c: > Let's ignore this test for now, it's a rare case.

ok

woai3c:
I have removed test cases related emoji. Please review the code again, Thank you. @Boshen @mysteryven

mysteryven (approved):

PR #3110, opened 2024-04-26
feat(linter): eslint-plugin-jest/require-hook

eryue0220:

Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/require-hook.ts)


mysteryven:
This rule is a bit strict. 😢

mysteryven (approved):
Nice work :)

eryue0220:
> This rule is a bit strict. 😢

Ok, I'll update the rule style in my next pr.

mysteryven:
> Ok, I'll update the rule style in my next pr.

It's good to keep it `style`, I wanted to express my though about this rule, I may write many codes that go against this rule:(

PR #3122, opened 2024-04-28
fix(semantic): connect `test` expression of `for` statements to the cfg.

rzvxa:
I don't know if it is correct or not, Fixes my issues with dangling cfg nodes created in for statements. #3071

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -785,6 +785,7 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         }
 
         /* cfg */
+        let after_test_graph_ix = self.cfg.current_node_ix;
         let update_graph_ix = self.cfg.new_basic_block();
         /* cfg */
 
@@ -803,10 +804,10 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         /* cfg */
         let after_for_stmt = self.cfg.new_basic_block();
         self.cfg.add_edge(before_for_graph_ix, test_graph_ix, EdgeType::Normal);
-        self.cfg.add_edge(test_graph_ix, body_graph_ix, EdgeType::Normal);
+        self.cfg.add_edge(after_test_graph_ix, body_graph_ix, EdgeType::Normal);
         self.cfg.add_edge(body_graph_ix, update_graph_ix, EdgeType::Backedge);
         self.cfg.add_edge(update_graph_ix, test_graph_ix, EdgeType::Backedge);
-        self.cfg.add_edge(test_graph_ix, after_for_stmt, EdgeType::Normal);
+        self.cfg.add_edge(after_test_graph_ix, after_for_stmt, EdgeType::Normal);
 
         self.cfg.after_statement(
             &statement_state,


rzvxa:
* **#3122** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
Given this block of code:

```js
            function t1() {
              const res = [];
              const additionalCond = true;
              for (let i = 0; i !== 10 && additionalCond; ++i ) {
                res.push(i);
              }
              React.useLayoutEffect(() => {});
            }
```

the cfg used to look like this:

```
                                  (0) Program--------------------------- (9) Dummy cfg
                                        |
                                        |
                                        V
                                  (1) Function
                                        |
                                        |
                                        V
   |------------------------->(2) Logical Expression
   |                                  /       \--------------------------------------------------------------------------------------------------
   |                                 /         \                                                                  \                              \
   |                                /           \                                                                  \                              \
   |                               /             \                                                                  \                              \
   |                              /               \                                                                  \                              \
   |                             /                 \                                                                  \                              \
   |                            V                   V                                                                  \                              \
   |      (3) Ident(`additionalCond`)------------>(4) Dummy cfg Created by the Logical Expression or others             \                              \
   |                                                                                                                     \                              \
   |                                                                                                                      \                              \
   |                                                                                                                       \                              \
   |                                                                                                                        \                              \
   |                                                                                                                         \                              \
   |                                                                                                                          \                              \
   |                                                                                                                          V                               V
   ---------------(5) UpdateExpression<------------------------------------------------------------------------------(6) BlockStatement            (7) ExpressionStatement
                                                                                                                                                               \
                                                                                                                                                                \
                                                                                                                                                                 \
                                                                                                                                                           (8) ArrowFunctionExpression
```

With this change, it would look like this:

```
                                  (0) Program--------------------------- (9) Dummy cfg
                                        |
                                        |
                                        V
                                  (1) Function
                                        |
                                        |
                                        V
   |------------------------->(2) Logical Expression
   |                                  /       \
   |                                 /         \
   |                                /           \
   |                               /             \
   |                              /               \
   |                             /                 \
   |                            V                   V
   |      (3) Ident(`additionalCond`)------------>(4) Dummy cfg Created by the Logical Expression or others
   |                                                            /              \
   |                                                           /                \
   |                                                          /                  \
   |                                                         /                    \
   |                                                        /                      \
   |                                                       /                        \
   |                                                      V                         V
   ---------------(5) UpdateExpression<----------(6) BlockStatement            (7) ExpressionStatement
                                                                                            \
                                                                                             \
                                                                                              \
                                                                                        (8) ArrowFunctionExpression

```

I'm unsure if this is illegal, Or would break anything in the type system, etc. But it would simplify the cfg traversal since every node had a follower node and we can check if any 2 nodes share a subtree by checking both are connected to the same parent, and If they have a connection they are also always on the same subtree.

@Boshen  Please give it a look I'm not really familiar with what a cfg should look like so it might be just wrong.

What I'm looking for is simplifying checks but I can always find another way to solve it. Right now I'm implementing #3071 with off-the-shelf algorithms so if this PR doesn't work I can just write a specialized one.

rzvxa:
Please don't merge it yet, Since if it gets approved there are other places that need the same treatment

Boshen:
There's also a graphical output from https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/examples/cfg.rs

Your change seems to be more correct than the previous one.

This can be verifed with the biome cfg,

image

There's a loop and a exit edge, which is your graph after the change.

Boshen (approved):

Boshen:
### Merge activity

* **May 10, 10:11 AM EDT**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/3122).

PR #3126, opened 2024-04-29
feat(linter): @typescript-eslint/consistent-indexed-object-style

todor-a:
implements https://typescript-eslint.io/rules/consistent-indexed-object-style/


crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
```diff
@@ -0,0 +1,583 @@
...
+                                            ),
+                                        }
+                                    }
+                                }
+                                TSType::TSUnknownKeyword(_) | TSType::TSAnyKeyword(_) => {
+                                    ctx.diagnostic(ConsistentIndexedObjectStyleDiagnostic(
+                                        "record",
+                                        "index signature",
+                                        idx.span,
+                                    ));
```
mysteryven, line 112: Why do we need to diagnostic here? 🤔

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
```diff
@@ -0,0 +1,583 @@
...
+                                }
+                                _ => {}
+                            }
+                        }
+                    }
+                }
+            }
+            AstKind::TSTypeAliasDeclaration(al) => match &al.type_annotation {
```
mysteryven, line 120: Does it work if we replace line 120 - line 355 with visit `AstKind::TSTypeLiteral` directly?

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
```diff
@@ -0,0 +1,583 @@
...
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("typescript-eslint(consistent-indexed-object-style):")]
```
mysteryven, line 15: Error message missed here :)

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
```diff
@@ -0,0 +1,583 @@
...
+        Self { is_index_signature: config == ConsistentIndexedObjectStyleConfig::IndexSignature }
+    }
+
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::TSInterfaceDeclaration(decl) => {
+                if decl.body.body.len() == 1 {
+                    for signature in &decl.body.body {
+                        if let TSSignature::TSIndexSignature(idx) = signature {
+                            match &idx.type_annotation.type_annotation {
```
mysteryven, line 67: Why we need to check `idx.type_annotation.type_annotation`, look like typescript-ESLint only check [`parameter`'s type](https://github.com/typescript-eslint/typescript-eslint/blob/584db29ec44ce4e9cb71afac35d48994889168e6/packages/eslint-plugin/src/rules/consistent-indexed-object-style.ts#L48-L61).

mysteryven (changes requested):

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: the `TSAnyKeyword` is for this test:
```ts
interface Foo {
  [key: string]: any;
}
```
you might ask, why not just error if there is an index signature, and the answer is that in some cases, even though the setting is `Record`, an index signature is allowed. For example this:
```ts
interface Foo<T> {
  [key: string]: Foo<T> | string;
}
```

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: not sure i got this one, can you elaborate?

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: for the following case:
```ts
interface Foo {
   [key: string]: Foo;
}
```

even with a setting of `Record`, the above is still valid

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: added!

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
mysteryven: Like this?

```rust
      match node.kind() {
          // --snip--
          AstKind::TSTypeLiteral(lit) => {
              if self.is_index_signature {
                  return;
              } 
              if lit.members.len() != 1 {
                  return;
              }
              let member = &lit.members[0];

              let TSSignature::TSIndexSignature(idx) = member else {
                  return;
              }

              // --snip--
          }
          AstKind::TSTypeReference(refercene) => {
              if !self.is_index_signature {
                  return;
              }
              if let TSTypeName::IdentifierReference(ident) = &refercene.type_name {
                  if ident.name == "Record" {
                      // snip
                  }
              } 
          }
          _ => {}
      }
  }
```

todor-a: I think I got it. Would you please take another look? :)

(sorry it took me so long)

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: I played around wit this but could not find a way to remove the TSUnknownKeyword & TSAnyKeyword checks and still have the tests pass 🤔

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
todor-a: not sure I got what you meant here

mysteryven: For case

```ts
interface Foo {
   [key: string]: Foo; // a
}
```

It is valid because it been treated as a circle, `Foo` in the second line refers to the interface name `Foo`.

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
```diff
@@ -0,0 +1,553 @@
...
+                    }
+                }
+                _ => (),
+            },
+            AstKind::Program(prog) => {
+                for body in &prog.body {
+                    if let Statement::FunctionDeclaration(func) = body {
+                        if let Some(return_type) = &func.return_type {
+                            if !self.is_index_signature {
+                                if let TSType::TSTypeLiteral(lit) = &return_type.type_annotation {
```
mysteryven, line 260: Perhaps we don't need to revisit `TSType::TSTypeLiteral`? Since all `TSType::TSTypeLiteral` has been visited.

```rust
      match node.kind() {
          AstKind::TSTypeLiteral(lit) => { } // here will visit all TSTypeLiteral in our file
      }
}
```

crates/oxc_linter/src/rules/typescript/consistent_indexed_object_style.rs
mysteryven: Maybe we need to add early bailout if we find a [circle](https://github.com/typescript-eslint/typescript-eslint/pull/4347/files)?

```ts
interface Foo {
	[key: string]: Foo | string;
}
```

mysteryven:
Thanks to your effort on this rule, take your time!

todor-a:
Hey, @mysteryven. After some more tinkering I think finally did it. Have another look when you get a chance and let me know what you think.

P.S. I managed to implement a fixer only for one of the settings. I will try to implement it for the other one as well but if not, we can log a separate task.

mysteryven (approved):

mysteryven:
>  I will try to implement it for the other one as well but if not, we can log a separate task.

I gonna to merge this first, let's iterate it later :)

And thanks again for all your work.

PR #3169, opened 2024-05-05
feat(transform): `oxc_traverse` crate

overlookmotel:
First part of #3152.

This adds a crate `oxc_traverse`, but doesn't connect it up to the transformer or anything else yet.

I think we could merge this now - as it doesn't affect any code that's in use - and then iterate on it to add scopes before using it in transformer. Please see https://github.com/oxc-project/oxc/pull/3152#issuecomment-2094965406 for the broader picture.


crates/oxc_traverse/src/lib.rs
```diff
@@ -0,0 +1,149 @@
+//! AST traversal with ability to read up the tree from visitors.
+//!
+//! Please see [`traverse_mut`] for an explanation of API.
```
mysteryven, line 3: These documents explain things really really good! I should read your PRs one by one, they are all treasures.

overlookmotel: That's incredibly kind of you to say. Thank you.

I do think it's important to write lengthy comments to explain how things work. But I'm a native English speaker, so I have an advantage in that respect - it's quite easy for me to turn thoughts into English words.

If you find anything where the comments don't make sense, or don't fully explain what the code is doing, or why, please do alert me and I'll do my best to improve them.

PR #3170, opened 2024-05-06
feat(traverse): add a comment for hardcode to retag code

Dunqing:
The hardcode number is hard to understand. Do you think we need to add a comment to it? @overlookmotel

### DIFF
--- a/crates/oxc_traverse/scripts/lib/ancestor.mjs
+++ b/crates/oxc_traverse/scripts/lib/ancestor.mjs
@@ -61,8 +61,10 @@ export default function generateAncestorsCode(types) {
 
             const variantName = `${type.name}${fieldNameCamel}`;
             variantNames.push(variantName);
-
-            enumVariants += `${variantName}(${structName}) = ${discriminant},\n`;
+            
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#3170** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Yes, I agree the hardcoded numbers are confusing.

In part, rationale was that this generates a ton of code and I was worried about compile time, so I tried to make it shorter where I could. I intended to further compact generated the code later on to squeeze it down further.

But, yes, that's at the cost of readability, and maybe that's a bad trade off.

If we do want to make it more comprehensible, instead of comments, how about this: Add another enum `AncestorType` with named variants and same discriminants as `Ancestor`, and then use that for `retag_stack` i.e. `ctx.retag_stack(AncestorType::ProgramBody)`?

overlookmotel:
@Dunqing Please see #3173. Does that work for you?

Issue #3213, opened 2024-05-09
Redo linter diagnostics


Boshen:
I discovered that compilation is slow - https://github.com/zkat/miette/issues/373

So it got me thinking why are we defining so many diagnostics, and how can we reduce them.

I think defining 1 `LinterDiagnostic` and then `LinterDiagnostic::new()` with different data should also work.

PR #3214, opened 2024-05-09
refactor(parser,diagnostic): one diagnostic struct to eliminate monomorphization of generic types

Boshen:
part of #3213

We should only have one diagnostic struct instead 353 copies of them, so we don't end up choking LLVM with 50k lines of the same code due to monomorphization.

If the proposed approach is good, then I'll start writing a codemod to turn all the existing structs to plain functions.

---

Background:

Using `--timings`, we see `oxc_linter` is slow on codegen (the purple part).

image

The crate currently contains 353 miette errors. [cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines) displays

```
cargo llvm-lines -p oxc_linter --lib --release

  Lines                 Copies               Function name
  -----                 ------               -------------
  830350                33438                (TOTAL)
   29252 (3.5%,  3.5%)    808 (2.4%,  2.4%)  <alloc::boxed::Box<T,A> as core::ops::drop::Drop>::drop
   23298 (2.8%,  6.3%)    353 (1.1%,  3.5%)  miette::eyreish::error::object_downcast
   19062 (2.3%,  8.6%)    706 (2.1%,  5.6%)  core::error::Error::type_id
   12610 (1.5%, 10.1%)     65 (0.2%,  5.8%)  alloc::raw_vec::RawVec<T,A>::grow_amortized
   12002 (1.4%, 11.6%)    706 (2.1%,  7.9%)  miette::eyreish::ptr::Own<T>::boxed
    9215 (1.1%, 12.7%)    115 (0.3%,  8.2%)  core::iter::traits::iterator::Iterator::try_fold
    9150 (1.1%, 13.8%)      1 (0.0%,  8.2%)  oxc_linter::rules::RuleEnum::read_json
    8825 (1.1%, 14.9%)    353 (1.1%,  9.3%)  <miette::eyreish::error::ErrorImpl<E> as core::error::Error>::source
    8822 (1.1%, 15.9%)    353 (1.1%, 10.3%)  miette::eyreish::error::<impl miette::eyreish::Report>::construct
    8119 (1.0%, 16.9%)    353 (1.1%, 11.4%)  miette::eyreish::error::object_ref
    8119 (1.0%, 17.9%)    353 (1.1%, 12.5%)  miette::eyreish::error::object_ref_stderr
    7413 (0.9%, 18.8%)    353 (1.1%, 13.5%)  <miette::eyreish::error::ErrorImpl<E> as core::fmt::Display>::fmt
    7413 (0.9%, 19.7%)    353 (1.1%, 14.6%)  miette::eyreish::ptr::Own<T>::new
    6669 (0.8%, 20.5%)     39 (0.1%, 14.7%)  alloc::raw_vec::RawVec<T,A>::try_allocate_in
    6173 (0.7%, 21.2%)    353 (1.1%, 15.7%)  miette::eyreish::error::<impl miette::eyreish::Report>::from_std
    6027 (0.7%, 21.9%)     70 (0.2%, 16.0%)  <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
    6001 (0.7%, 22.7%)    353 (1.1%, 17.0%)  miette::eyreish::error::object_drop
    6001 (0.7%, 23.4%)    353 (1.1%, 18.1%)  miette::eyreish::error::object_drop_front
    5648 (0.7%, 24.1%)    353 (1.1%, 19.1%)  <miette::eyreish::error::ErrorImpl<E> as core::fmt::Debug>::fmt
```

It's totalling more than 50k llvm lines, and is putting pressure on rustc codegen (the purple part on `oxc_linter` in the image above.

---

It's pretty obvious by looking at https://github.com/zkat/miette/blob/main/src/eyreish/error.rs, the generics can expand out to lots of code.

### DIFF
--- a/crates/oxc_diagnostics/src/lib.rs
+++ b/crates/oxc_diagnostics/src/lib.rs
@@ -6,7 +6,11 @@ mod graphical_theme;
 mod reporter;
 mod service;
 
-use std::path::PathBuf;
+use std::{
+    fmt::{self, Display},
+    ops::Deref,
+    path::PathBuf,
+};
 
 pub use miette;
 pub use thiserror;
@@ -21,9 +25,10 @@ pub type Error = miette::Error;
 pub type Severity = miette::Severity;
 pub type Report = miette::Report;
 
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T> = std::result::Result<T, OxcDiagnostic>;
 
-use miette::Diagnostic;
+pub use miette::LabeledSpan;
+use miette::{Diagnostic, SourceCode};
 use thiserror::Error;
 
 #[derive(Debug, Error, Diagnostic)]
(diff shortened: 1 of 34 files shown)


crates/oxc_diagnostics/src/lib.rs
```diff
@@ -35,3 +39,64 @@ pub struct MinifiedFileError(pub PathBuf);
...
+
+impl Display for OxcDiagnostic {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.message)
+    }
+}
+
+impl Diagnostic for OxcDiagnostic {
+    fn severity(&self) -> Option<Severity> {
+        Some(self.severity.unwrap_or(Severity::Warning))
```
Dunqing, line 90:
```suggestion
        self.severity
```

/// If `None`, reporters should treat this as [`Severity::Error`].

Dunqing (changes requested):

elenakrittik:
Would it be better to use miette's built-in [`MietteDiagnostic`](https://docs.rs/miette/latest/miette/struct.MietteDiagnostic.html) instead of a custom type? Or do you expect the cost of `String` allocations to be too high?

Boshen:
> Would it be better to use miette's built-in [`MietteDiagnostic`](https://docs.rs/miette/latest/miette/struct.MietteDiagnostic.html) instead of a custom type? Or do you expect the cost of `String` allocations to be too high?

How did I miss this after all these time (2 years) ...

Boshen:
> Would it be better to use miette's built-in [`MietteDiagnostic`](https://docs.rs/miette/latest/miette/struct.MietteDiagnostic.html) instead of a custom type? Or do you expect the cost of `String` allocations to be too high?

Haha, rust complained, back to creating our own struct

```
error: the `Err`-variant returned from this function is very large
   --> crates/oxc_parser/src/lib.rs:338:36
    |
338 |     fn parse_program(&mut self) -> Result<Program<'a>> {
    |                                    ^^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 128 bytes
```

Boshen:
> Would it be better to use miette's built-in [`MietteDiagnostic`](https://docs.rs/miette/latest/miette/struct.MietteDiagnostic.html) instead of a custom type? Or do you expect the cost of `String` allocations to be too high?

Haha, rust complained, back to creating our own struct

```
error: the `Err`-variant returned from this function is very large
   --> crates/oxc_parser/src/lib.rs:338:36
    |
338 |     fn parse_program(&mut self) -> Result<Program<'a>> {
    |                                    ^^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 128 bytes
```

Boshen:
Binary size for `cargo build --release --example parser` shrank from 1.0M to 958K. Big win!

Dunqing:
Very much looking forward to shrinking in the linter!

Dunqing (approved):

PR #3216, opened 2024-05-09
refactor(traverse): simplified way to get TraverseCtx

Dunqing:
When we want to use `TraverseCtx` in nested functions, we run into trouble. We need to keep passing the `TraverseCtx` around. I've put `TraverseCtx` inside `TransformerCtx` because `TransformerCtx` is everywhere.

### DIFF
--- a/crates/oxc_transformer/src/context.rs
+++ b/crates/oxc_transformer/src/context.rs
@@ -1,5 +1,5 @@
 use std::{
-    cell::RefCell,
+    cell::{Ref, RefCell, RefMut},
     mem,
     path::{Path, PathBuf},
     rc::Rc,
@@ -9,6 +9,7 @@ use oxc_allocator::Allocator;
 use oxc_ast::{AstBuilder, Trivias};
 use oxc_diagnostics::Error;
 use oxc_span::SourceType;
+use oxc_traverse::TraverseCtx;
 
 use crate::{helpers::module_imports::ModuleImports, TransformOptions};
 
@@ -18,6 +19,8 @@ pub struct TransformCtx<'a> {
     errors: RefCell<Vec<Error>>,
 
     pub trivias: &'a Trivias,
+
+    traverse: Rc<RefCell<TraverseCtx<'a>>>,
 
     pub ast: AstBuilder<'a>,
 
@@ -45,6 +48,8 @@ impl<'a> TransformCtx<'a> {
         trivias: &'a Trivias,
         options: &TransformOptions,
     ) -> Self {
+        let traverse = Rc::new(RefCell::new(TraverseCtx::new(allocator)));
+
         let filename = source_path
             .file_stem() // omit file extension
             .map_or_else(|| String::from("unknown"), |name| name.to_string_lossy().to_string());
@@ -61,6 +66,7 @@ impl<'a> TransformCtx<'a> {
             source_type,
             source_text,
             trivias,
+            traverse,
             module_imports: ModuleImports::new(allocator),
         }
     }
@@ -73,5 +79,13 @@ impl<'a> TransformCtx<'a> {
     #[allow(unused)]
     pub fn error<T: Into<Error>>(&self, error: T) {
         self.errors.borrow_mut().push(error.into());
+    }
+
+    pub fn traverse(&self) -> Ref<TraverseCtx<'a>> {
+        self.traverse.borrow()
+    }
+
+    pub fn traverse_mut(&self) -> RefMut<TraverseCtx<'a>> {
+        self.traverse.borrow_mut()
     }
 }
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -21,7 +21,7 @@ mod helpers {
     pub mod module_imports;
 }
 
-use std::{path::Path, rc::Rc};
+use std::{cell::RefMut, path::Path, rc::Rc};
 
 use es2015::ES2015;
 use oxc_allocator::{Allocator, Vec};
@@ -78,8 +78,7 @@ impl<'a> Transformer<'a> {
     ///
     /// Returns `Vec<Error>` if any errors were collected during the transformation.
     pub fn build(mut self, program: &mut Program<'a>) -> Result<(), std::vec::Vec<Error>> {
-        let allocator = self.ctx.ast.allocator;
-        traverse_mut(&mut self, program, allocator);
+        traverse_mut(&mut self, program);
 
         let errors = self.ctx.take_errors();
         if errors.is_empty() {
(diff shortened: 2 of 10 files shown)


Dunqing:
* **#3216** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
I don't know how to fix these `ancestor_lifetime` tests

overlookmotel:
I'm afraid I don't think this is possible.

### Why you can't do this

The safety of the scheme depends on making it impossible for an `Ancestor` to exist outside of/outlive an `enter_*` or `exit_*` method. If it was possible, then you could:

* Save an `Ancestor` into a property of `self`.
* Later on in an `exit_*` method e.g. `exit_program`, pull that `Ancestor` back out of `self`.
* Call one of the methods on the `Ancestor` to get a reference to an AST node.
* You now hold a `&mut Program` and simultaneously an immut `&` ref to one of its children.
* Boom! You've broken the aliasing rules. Undefined behavior!

The thing that prevents you "holding on" to an `Ancestor` for longer is:

* You have to borrow it from `&ctx`.
* `&ctx` only lives as long as the visitor function is executing.

So tying the lifetime of the `ctx` object is essential to the safety of the scheme. And only way to do that is have it as a function parameter so that `walk_*` controls the lifetime.

`Traverse` allows you to do something you usually can't do (go back up the tree), but *without breaking Rust's aliasing rules*. The API is quite carefully designed to accomplish this. I've made the API as pleasant as I can, but unfortunately there are limits to how far it can be pushed within Rust's rules, without breaking the whole thing.

Some of this is quite subtle. I have added [docs](https://github.com/oxc-project/oxc/blob/main/crates/oxc_traverse/src/lib.rs) a couple of days ago. But probably I should add more docs to explain it more fully.

### Performance impact

In this PR, you're also introducing an `Rc` where there wasn't one before.

CodSpeed is showing this as a -3% perf regression. 3% doesn't sound like much, but bear in mind that that benchmark time also includes the parser, and that's 80% of the time - only ~20% is the transformer itself. So if a change to transformer shows as 3% regression on this benchmark, it actually means the transformer itself has got 15% slower. That's quite bad.

### What we could do instead

I had actually been thinking we could go in the other direction - move more stuff that all transforms need *into* the `TraverseCtx` object. This would:

1. Remove repeated code from all the individual transformer impls.
2. Likely remove the need for some of the other `Rc`s which we currently use, which will give a perf boost.

I already put the allocator and AST builder in `TraverseCtx`. That was meant to be just the start.

Does this make any sense?

overlookmotel:
> When we want to use TraverseCtx in nested functions, we run into trouble

Can you give me an example of this?

Dunqing:
> > When we want to use TraverseCtx in nested functions, we run into trouble
> 
> Can you give me an example of this?

https://github.com/oxc-project/oxc/blob/83f50bc6ce9615dc9c55084f610e9775c87a2e42/crates/oxc_transformer/src/react/jsx/mod.rs#L262
I want to call `find_ancestor` here. This means that I need to pass TraverseCtx throughout the call stack.

overlookmotel:
Ah I see. And this isn't a very helpful reply, but... sorry, can't do anything about it. If you need to use `ctx` in a function, it has to be passed into that function, there's no way to make it global within `self`.

But... unlike `Visit` and `VisitMut`, `enter_*` and `exit_*` don't call `walk_*` themselves, so that does open the door to better composability.

If all individual transforms are `Traverse` impls:

```rs
struct MyTransform1;

impl<'a> Traverse<'a> for MyTransform1 {
  fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
    // Do stuff
  }
}

struct MyTransform2;

impl<'a> Traverse<'a> for MyTransform2 {
  fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
    // Do more stuff
  }
}
```

Then we can codegen a generic type `CombinedTraverse`:

```rs
struct CombinedTraverse<T1: Traverse, T2: Traverse> {
  t1: T1,
  t2: T2,
}

impl<T1: Traverse, T2: Traverse> CombinedTraverse<T1, T2> {
  pub fn new(t1: T1, t2, T2) -> Self {
    Self { t1, t2 }
  }
}

impl<T1: Traverse, T2: Traverse> CombinedTraverse<T1, T2> {
  fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
    self.t1.enter_program(program, ctx);
    self.t2.enter_program(program, ctx);
  }

  // etc - all the enter/exit methods
}
```

Then rather than writing by hand a `Traverse` which just forwards all methods to the individual transforms, you combine 2 transforms and run them both with e.g.:

```rs
let transform = CombinedTraverse::new(
  TypescriptTransform::new(),
  ReactTransform::new(),
);
traverse_mut(&mut transform, program, allocator);
```

or even:

```rs
let transform = CombinedTraverse::new(
  CombinedTraverse::new(
    TypescriptTransform::new(),
    CombinedTraverse::new(
      ReactTransform::new(),
      CombinedTraverse::new(
        ES6Transform::new(),
        ES5Transform::new(),
      ),
    ),
  ),
);
traverse_mut(&mut transform, program, allocator);
```

Or with a macro to sugar that:

```rs
let transform = combine_transforms!(
  TypescriptTransform::new(),
  ReactTransform::new(),
  ES6Transform::new(),
  ES5Transform::new(),
);
traverse_mut(&mut transform, program, allocator);
```

Can't do anything about needing `ctx` param in all your visitor functions, but it *would* remove the need to have intermediate types which pass down `ctx` everywhere.

Does this API appeal to you?

NB: What I don't know is whether this would blow up compile times. Maybe it's impractical for that reason.

Dunqing:
> ### Why you can't do this
> The safety of the scheme depends on making it impossible for an `Ancestor` to exist outside of/outlive an `enter_*` or `exit_*` method. If it was possible, then you could:
> 
> * Save an `Ancestor` into a property of `self`.
> * Later on in an `exit_*` method e.g. `exit_program`, pull that `Ancestor` back out of `self`.
> * Call one of the methods on the `Ancestor` to get a reference to an AST node.
> * You now hold a `&mut Program` and simultaneously an immut `&` ref to one of its children.
> * Boom! You've broken the aliasing rules. Undefined behavior!
> 
> The thing that prevents you "holding on" to an `Ancestor` for longer is:
> 
> * You have to borrow it from `&ctx`.
> * `&ctx` only lives as long as the visitor function is executing.
> 
> So tying the lifetime of the `ctx` object is essential to the safety of the scheme. And only way to do that is have it as a function parameter so that `walk_*` controls the lifetime.
> 
> `Traverse` allows you to do something you usually can't do (go back up the tree), but _without breaking Rust's aliasing rules_. The API is quite carefully designed to accomplish this. I've made the API as pleasant as I can, but unfortunately there are limits to how far it can be pushed within Rust's rules, without breaking the whole thing.
> 
> Some of this is quite subtle. I have added [docs](https://github.com/oxc-project/oxc/blob/main/crates/oxc_traverse/src/lib.rs) a couple of days ago. But probably I should add more docs to explain it more fully.
> 

Thank you for explaining this. Do we have a runtime check way to avoid a break? 

```rust
#[allow(unsafe_code)]
pub fn traverse_mut<'a, Tr: Traverse<'a>>(traverser: &mut Tr, program: &mut Program<'a>) {
    // SAFETY: Walk functions are constructed to avoid unsoundness
    // Save an `TraverseCtx` into property a of`self`
    unsafe { walk::walk_program(traverser, program as *mut Program) };
    debug_assert!(traverser.ctx().ancestors_depth() == 1);
    // Remove `TraverseCtx` from self 
}
```
Is this possible?

I think ease of use is something we have to consider. We have tons of plugins to write.

Dunqing:
> But... unlike `Visit` and `VisitMut`, `Traverse`'s `enter_*` and `exit_*` don't call `walk_*` themselves, so that does open the door to better composability.
> 
> If all individual transforms are `Traverse` impls:
> 
> ```rust
> struct MyTransform1;
> 
> impl<'a> Traverse<'a> for MyTransform1 {
>   fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
>     // Do stuff
>   }
> }
> 
> struct MyTransform2;
> 
> impl<'a> Traverse<'a> for MyTransform2 {
>   fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
>     // Do more stuff
>   }
> }
> ```
> 
> Then we can codegen a generic type `CombinedTraverse`:
> 
> ```rust
> struct CombinedTraverse<T1: Traverse, T2: Traverse> {
>   t1: T1,
>   t2: T2,
> }
> 
> impl<T1: Traverse, T2: Traverse> CombinedTraverse<T1, T2> {
>   pub fn new(t1: T1, t2, T2) -> Self {
>     Self { t1, t2 }
>   }
> }
> 
> impl<T1: Traverse, T2: Traverse> CombinedTraverse<T1, T2> {
>   fn enter_program(&mut self, program: &mut Program<'a>, ctx: &TraverseCtx<'a>) {
>     self.t1.enter_program(program, ctx);
>     self.t2.enter_program(program, ctx);
>   }
> 
>   // etc - all the enter/exit methods
> }
> ```
> 
> Then rather than writing by hand a `Traverse` which just forwards all methods to the individual transforms, you combine 2 transforms and run them both with e.g.:
> 
> ```rust
> let transform = CombinedTraverse::new(
>   TypescriptTransform::new(),
>   ReactTransform::new(),
> );
> traverse_mut(&mut transform, program, allocator);
> ```
> 
> or even:
> 
> ```rust
> let transform = CombinedTraverse::new(
>   TypescriptTransform::new(),
>   CombinedTraverse::new(
>     ReactTransform::new(),
>     CombinedTraverse::new(
>       ES6Transform::new(),
>       ES5Transform::new(),
>     ),
>   ),
> );
> traverse_mut(&mut transform, program, allocator);
> ```
> 
> Or with a macro to sugar that:
> 
> ```rust
> let transform = combine_transforms!(
>   TypescriptTransform::new(),
>   ReactTransform::new(),
>   ES6Transform::new(),
>   ES5Transform::new(),
> );
> traverse_mut(&mut transform, program, allocator);
> ```
> 
> Can't do anything about needing `ctx` param in all your visitor functions, but it _would_ remove the need to have intermediate types which pass down `ctx` everywhere.
> 
> Does this API appeal to you?

Yes, I think it's quite good. It will help us reduce a lot of boilerplate code. But for complex transformations, we still need to abstract into multiple functions (such as `transform_jsx`), and we still need to pass the ctx.

overlookmotel:
> Thank you for explaining this. Do we have a runtime check way to avoid a break?

If you try to make an `Ancestor` "escape" from `enter_*`, that's a compile time error. So we can't do a runtime check, as it would refuse to compile!

But this is what the [compile failure tests](https://github.com/oxc-project/oxc/tree/main/crates/oxc_traverse/tests) are for - those are the tests which are failing on this PR.

> Is this possible?
> I think ease of use is something we have to consider. We have tons of plugins to write.

I do see your point, but I don't think it's possible. Let me think about it. *Maybe* I can come up with something, but it's pretty tricky. The whole scheme is quite finely balanced to give us what we want (ability to access back up the tree) while avoiding the pitfalls of potential UB.

I also would like to leave the door open for *mutating up*, which I think can be possible with this scheme, and could be useful in future. So I wouldn't want to block that possibility while trying to remove `ctx`. And we also need to avoid the 15% perf cost of `Rc<RefCell>`, in my opinion.

So, yes please leave it with me and I will try to come up with something. But I'm afraid I'm unlikely to get anywhere for probably a week (it took me 3 weeks and 4 attempts to get this far).

> Yes, I think it's quite good. It will help us reduce a lot of boilerplate code. But for complex transformations, we still need to abstract into multiple functions (such as transform_jsx), and we still need to pass the ctx.

Composition can be many levels deep. i.e. any transform can compose multiple other transforms.

What I had in mind is that `React`, `ReactJsx`, `ReactJsxSelf`, `ReactJsxSource` are all implemented as `Traverse` impls. Then `React` transform is created as a combination of them:

```rs
let reactTransform = combine_transforms!(
  ReactJsx::new( /* options */ ),
  ReactJsxSelf::new( /* options */ ),
  ReactJsxSource::new( /* options */ ),
  // etc...
);
```

and then the top level of transformer:

```rs
let allTransform = combine_transforms!(
  Typescript::new( /* options */ ),
  React::new( /* options */ ),
  /* ... any more */
```

i.e. Each transform "plugin" is either a direct implementation of a transform *or* a composition of multiple other transforms. And each of those can be a composition of more transforms if you like. You can go as deep as you want to.

@Boshen Can you advise if this `CombinedTraverse` is going to be a compile-time killer? Or do we just need to try it to find out?

overlookmotel:
I've implemented `CombinedTraverse` in #3219 so we can try it out.

PR #3218, opened 2024-05-10
feat(linter/jsdoc): Implement require-returns rule

leaysgur:
Part of #1170

> https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-returns.md


Boshen (approved):
That's a lot of test cases! Thank you!

PR #3220, opened 2024-05-10
feat(linter): unicorn/no-anonymous-default-export

1zumii:
Refer to [eslint-plugin-unicorn/no-anonymous-default-export](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v52.0.0/docs/rules/no-anonymous-default-export.md)

link: #684

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_anonymous_default_export.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_anonymous_default_export.rs
@@ -1,0 +1,155 @@
+use std::fmt;
+
+use oxc_ast::{
+    ast::{AssignmentExpression, AssignmentTarget, ExportDefaultDeclarationKind, Expression},
+    AstKind,
+};
+use oxc_diagnostics::{
+    miette::{self, Diagnostic},
+    thiserror::Error,
+};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Error, Diagnostic)]
+#[error("eslint-plugin-unicorn(no-anonymous-default-export): Disallow anonymous functions and classes as the default export")]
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/unicorn/no_anonymous_default_export.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    let pass = vec![
+        r"export default class Foo {}",
+        r"export default function foo () {}",
+        r"const foo = () => {}; export default foo;",
+        r"module.exports = class Foo {};",
+        r"module.exports = function foo () {};",
+        r"const foo = () => {}; module.exports = foo;",
+        // TODO: need handle this situation?
+        // r"module['exports'] = function foo () {};",
```
rzvxa, line 142: First off thanks for your contribution🥳.

I'm not sure about this one so I let others pitch in, but there is another case that fails,

```js
        export default (async function * () {})
        export default (class extends class {} {})
```

You also have to check for nested items within parenthesized expressions(and in theory can be nested infinitely).

1zumii: 🤔 you mean that declaration of `export default` can be wrapped in any level parens?

rzvxa: Yes exactly, I suggest you take a look at these cases from the original plugin(if you already haven't).
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v52.0.0/test/no-anonymous-default-export.mjs

Porting the tests from the original rule can help to prevent any unintentional difference between them.

Dunqing:
You should use `just new-unicorn-rule no-anonymous-default-export` to create a new rule. The rule file will contain these tests. In general, you don't need to add these tests manually

Boshen:
I'll fix the conflict with the main branch.

Boshen (approved):

PR #3221, opened 2024-05-10
ci: build each benchmark only with deps it needs

overlookmotel:
This PR builds on #3201 to further speed up the benchmarks and reduce CI time.

* Build and run each benchmark as separate job (like before).
* But now each bench is only built with the dependencies it needs.
* For linter benchmarks, build benchmark in 1 job (like #3201 does).
* Run each linter fixture in a separate job as they're slow.

This reduces total time to complete benchmarks from between 6m-7m to ~4m40s.

All the individual jobs complete in under 1m30s, except for building linter benchmark which takes 2m30s. So there won't be the problem of blocking the CI queue that there was before.

NB: I did try this before, and didn't see a benefit. But I realized today what I was doing wrong - it only works once the caches are populated by a previous run on main branch.

So the CI times in this PR won't look good, but once it's merged to main, it will take effect. Here it is running on main branch of my fork:

https://github.com/overlookmotel/oxc/actions/runs/9030511348

I also added a step to delete the temp artefacts which aren't needed once the run has completed.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,3 +1,16 @@
+# Benchmarks are sharded.
+#
+# Each benchmark (parser, transformer, etc) runs in parallel in a separate job.
+# Linter benchmarks are much slower to build and run than the rest, so linter benchmark
+# is built in 1 job, and then run on each fixture in parallel in separate jobs.
+# When all jobs are complete, a final job uploads all the results to CodSpeed.
+#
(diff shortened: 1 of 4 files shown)


overlookmotel:
* **#3221** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):

PR #3172, opened 2024-05-06
ci: reduce the total number of linter benchmarks

Boshen:
closes #2981

PR #3222, opened 2024-05-10
ci: run all linter benchmarks

overlookmotel:
After #3221 we can re-enable the linter benchmarks which were disabled in #3172, as they won't hog so much CI time any more.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -160,6 +160,9 @@ jobs:
         fixture:
           - 0
           - 1
+          - 2
+          - 3
+          - 4
 
     steps:
       - name: Checkout Branch


overlookmotel:
I still don't understand why the linter benchmark takes so long to build. The linter itself is an enormous bunch of code, so you'd expect it, but the benchmark crate also takes ages to build.

`--timings` output on CI:

linter-bench-ci

Boshen:
Github action queues the jobs, adding too many linter benchmarks is pretty useless. Let me put this on hold.

overlookmotel:
OK, no problem. Actually we should probably drop the 5th benchmark (`antd.js`) anyway as it takes too long to run even all by itself.

My logic was that now that the jobs don't include building the benchmark and run in ~1 minute, they don't gum up the queue for long, so contention is less of a concern.

Boshen:
Feels like 2 linter benchmarks is enough for me, closing.

PR #3228, opened 2024-05-10
refactor(ast): order AST type fields in visitation order

overlookmotel:
Reorder the fields in a couple of AST types.

`walk_*` functions in `oxc_traverse` visit the fields in the order they appear in AST. So this change means that decorators are visited before classes and methods that they decorate.

PR #3226, opened 2024-05-10
refactor(traverse): do not expose `TraverseCtx::new`

overlookmotel:
Creating a `TraverseCtx` with `TraverseCtx::new` should be an internal API within `oxc_traverse`. Don't expose it outside the crate.

PR #3229, opened 2024-05-10
feat(traverse): add scope flags to `TraverseCtx`

### LINKED ISSUE #3189: Add scopes support to `Traverse` / transformer
The intent is to replace `VisitMut` in transformer with the new `oxc_traverse::Traverse`. #3182 is a starting point for that.

However, transformers require scope information, which is not yet implemented.

`VisitMut` provides only fairly minimal information - just `ScopeFlags` indicating:

* What kind of environment visitor is in (top level / function / method / class static block / etc).
* Whether strict mode or not.

This would be fairly easy to replicate in `Traverse`.

However, I assume transformers actually need more info than this, probably more like the info available from semantic. Opening this issue to discuss:

1. What scope info transformers need.
2. How to implement it.

overlookmotel:
Add scope flags to `TraverseCtx`.

Closes #3189.

`walk_*` functions build a stack of `ScopeFlags` as AST is traversed, and they can be queried from within visitors with `ctx.scope()`, `ctx.ancestor_scope()` and `ctx.find_scope()`.

The codegen which generates `walk_*` functions gets the info about which AST types have scopes, and how to check for strict mode from the `#[visited_node]` attrs on AST type definitions in `oxc_ast`.

A few notes:

Each scope inherits the strict mode flag from the level before it in the stack, so if you need to know "am I in strict mode context here?", `ctx.scope().is_strict_mode()` will tell you - no need to travel back up the stack to find out.

Scopes do *not* inherit any other flags from level before it. So `ctx.scope()` in a block nested in a function will return `ScopeFlags::empty()` not `ScopeFlags::Function`.

I had to add an extra flag `ScopeFlags::Method`. The reason for this is to deal with when a `Function` is actually a `MethodDefinition`, and to avoid creating 2 scopes in this case. The principle I'm trying to follow is to encode as little logic in the codegen as possible, as it's rather hidden away. Instead the codegen follows a standard logic for every node, guided by attributes which are visible next to the types in `oxc_ast`. This hopefully makes how `Traverse`'s visitors are generated less mysterious, and easier to change.

The case of `Function` within `MethodDefinition` is a weird one and would not be possible to implement without encoding a magic "special case" within the codegen without this extra `ScopeFlags::Method` variant. Its existence does not alter the operation of any other code in Oxc which uses `ScopeFlags`.

In my view `ScopeFlags` might benefit from a little bit of an overhaul anyway. I believe we could pack more information into the bits and make it more useful.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -14,6 +14,7 @@ use oxc_syntax::{
         AssignmentOperator, BinaryOperator, LogicalOperator, UnaryOperator, UpdateOperator,
     },
     reference::{ReferenceFlag, ReferenceId},
+    scope::ScopeFlags,
     symbol::SymbolId,
 };
 #[cfg(feature = "serialize")]
@@ -41,7 +42,10 @@ export interface FormalParameterRest extends Span {
 }
 "#;
 
-#[visited_node]
+#[visited_node(
+    scope(ScopeFlags::Top),
+    strict_if(self.source_type.is_strict() || self.directives.iter().any(Directive::is_use_strict))
+)]
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(tag = "type", rename_all = "camelCase"))]
@@ -1508,7 +1512,7 @@ pub struct Hashbang<'a> {
 }
 
 /// Block Statement
-#[visited_node]
+#[visited_node(scope(ScopeFlags::empty()))]
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(tag = "type"))]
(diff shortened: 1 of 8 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3229?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3229** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3228** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3227** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3226** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3226?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Boshen (approved):
I may have missed this, where is the logic for checking "use strict" inside a function?

Edit: oh it's in another PR.

Boshen:
### Merge activity

* **May 11, 12:16 AM EDT**: @Boshen started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/3229).

overlookmotel:
@Boshen FYI The logic for checking for `"use strict"` is in the `#[visited_node]` attr on `Function` (`strict_if` clause):

https://github.com/oxc-project/oxc/blob/ec41dba1971b67ff61449105f3892eb7d222e49f/crates/oxc_ast/src/ast/js.rs#L2134-L2143

The additional `is_use_strict` and `has_use_strict_directive` methods added in PR further down the stack were  just to make the code in `#[visited_node]` less verbose.

PR #3234, opened 2024-05-11
fix(traverse): create scope for function nested in class method

overlookmotel:
Fixes a bug in #3229.

The logic to prevent a duplicate scope being created for a `Function` which is a `MethodDefinition` would also stop a scope being created for inner function in:

```rs
class X {
  foo() {
    function bar() {}
  }
}
```

or

```rs
class X {
  foo( bar = function() {} ) {}
}
```

This PR fixes that. This change also allows removing `ScopeFlags::Method` which #3229 added.


overlookmotel:
* **#3234** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #3235, opened 2024-05-11
refactor(linter,diagnostic): one diagnostic struct to eliminate monomorphization of generic types

Boshen:
part of #3213


Boshen:
On my M2, running `just oxlint`:
Before: 5.4M 
After: 5.1M 

Let me keep going ...

PR #3238, opened 2024-05-11
feat(linter/eslint): add `no_unreachable` rule.

### LINKED ISSUE #621: feat(linter): eslint/no-unreachable
Implement ESLint's [`no-unreachable`](https://eslint.org/docs/latest/rules/no-unreachable#rule-details) rule.

We will probably need to implement a control-flow-graph for `Semantic` before implementing this rule.

rzvxa:
closes #621 
[no-unreachable](https://github.com/eslint/eslint/blob/069aa680c78b8516b9a1b568519f1d01e74fb2a2/lib/rules/no-unreachable.js#L196)

[oxlint-echosystem-ci result](https://github.com/rzvxa/oxlint-ecosystem-ci/actions/runs/9406195143/job/25909079029)

This rule is done but since it is running for every possible statement and does quite a bit of work on them to determine whether it is 100% reachable or not; The performance in my opinion is kind of abysmal.

I'll try to work it out, I know Biome does 2 types of checks to simplify the rule for some nodes, However, they have a lot more false negatives than our implementation.

##### Here is one example of those [false negatives](https://biomejs.dev/playground/?code=ZgB1AG4AYwB0AGkAbwBuACAAeAAoACkAIAB7ACAAZABvACAAewAgAGEAKAApADsAIAB9ACAAdwBoAGkAbABlACgAdAByAHUAZQApADsAIABiACgAKQA7ACAAfQA%3D)

-------------

### Update 1:

I've benchmarked this rule using only the simplified reachability checks and it was around 5% faster, To be honest, it isn't much improvement especially considering that we can only use this check for a small portion of nodes and even that is accompanied by newly introduced checks which would lessen the amount of performance gain further.

Most of the performance regression is because of allocations during our depth first search since we have to store both the visited and finished nodes which results in a bunch of rapid-fire allocations back to back. Currently, At the moment I don't have a great idea of how to improve it, We may have to implement our own graph to use arenas underneath.

Given that this rule is the most extensive use case of control flow (It doesn't come with a limited scope similar to property and constructor rules already implemented) this performance drop might be reasonable to some extent.

------------

### Update 2:

I reworked my approach in 2 senses, First I used @Boshen's suggestion inspired by TypeScript and kept some of the reachability information in the basic block structure instead of calculating it on the fly. It is done by propagating the `Unreachable` edge and `Unreachable` instruction throughout subgraphs.

This for sure helped with the performance but the next part is what never failed to amaze me, Going from something near `O(n!)` in the worst-case scenario to `O(n^2)` (in the worst-case scenario). By changing the approach instead of checking the reachability of each statement we do it in 3 paths; First, we do a path on the entire CFG and query all reachable but suspicious cases, and then we do another path on each of these suspicions subgraphs to determine the reachability with higher confidence. Finally, we iterate all of the appropriate nodes and check their reachability status according to the information collected in 2 previous paths.

With these 2 this rule went from `-24%` to `~-2%`.

This performance gain doesn't come for free though; It increases the likelihood of false positives/negatives, But as long as we are passing our `ecosystem-ci` it should be fine. We can always sacrifice some performance to check for edge cases if there are any.

[new oxlint-echosystem-ci result](https://github.com/rzvxa/oxlint-ecosystem-ci/actions/runs/9490791181)

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1690,6 +1690,10 @@ impl<'a> VariableDeclaration<'a> {
     pub fn is_typescript_syntax(&self) -> bool {
         self.modifiers.contains(ModifierKind::Declare)
     }
+
+    pub fn has_init(&self) -> bool {
+        self.declarations.iter().any(|decl| decl.init.is_some())
+    }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -95,6 +95,7 @@ mod eslint {
     pub mod no_ternary;
     pub mod no_this_before_super;
     pub mod no_undef;
+    pub mod no_unreachable;
     pub mod no_unsafe_finally;
     pub mod no_unsafe_negation;
     pub mod no_unsafe_optional_chaining;
@@ -483,6 +484,7 @@ oxc_macros::declare_all_lint_rules! {
     eslint::no_shadow_restricted_names,
     eslint::no_sparse_arrays,
     eslint::no_undef,
+    eslint::no_unreachable,
     eslint::no_unsafe_finally,
     eslint::no_unsafe_negation,
     eslint::no_unsafe_optional_chaining,
--- a/crates/oxc_linter/src/rules/eslint/no_unreachable.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_unreachable.rs
@@ -1,0 +1,307 @@
+use oxc_ast::{ast::VariableDeclarationKind, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_semantic::{
+    petgraph::{
+        visit::{depth_first_search, Control, DfsEvent, EdgeRef},
+        Direction,
+    },
+    EdgeType, InstructionKind,
+};
+use oxc_span::{GetSpan, Span};
+
+use crate::{context::LintContext, rule::Rule};
+
+fn no_unreachable_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::error("eslint(no-unreachable): Unreachable code.").with_labels([span.into()])
+}
+
+/// <https://github.com/eslint/eslint/blob/069aa680c78b8516b9a1b568519f1d01e74fb2a2/lib/rules/no-unreachable.js#L196>
+#[derive(Debug, Default, Clone)]
+pub struct NoUnreachable;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Disallow unreachable code after `return`, `throw`, `continue`, and `break` statements
+    ///
+    NoUnreachable,
+    correctness
+);
+
+impl Rule for NoUnreachable {
+    fn run_once(&self, ctx: &LintContext) {
+        let nodes = ctx.nodes();
+        let Some(root) = nodes.root_node() else { return };
+        let cfg = ctx.semantic().cfg();
+        let graph = &cfg.graph;
+
+        // A pre-allocated vector containing the reachability status of all the basic blocks.
+        // We initialize this vector with all nodes set to `unreachable` since if we don't visit a
+        // node in our paths then it should be unreachable by definition.
+        let mut unreachables = vec![true; cfg.basic_blocks.len()];
+
+        // All of the end points of infinite loops we encountered.
+        let mut infinite_loops = Vec::new();
+
+        // Set the root as reachable.
+        unreachables[root.cfg_id().index()] = false;
+
+        // In our first path we first check if each block is definitely unreachable, If it is then
+        // we set it as such, If we encounter an infinite loop we keep its end block since it can
+        // prevent other reachable blocks from ever getting executed.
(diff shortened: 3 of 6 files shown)


rzvxa:
* **#3238** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3238?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
@Boshen Can you read the first comment on this PR? I was wondering how much should I improve the performance to make it acceptable.

Boshen:
@rzvxa Instead of doing the check inside the rule, are you able to mark the statement as unreachable during the cfg building phase inside semantic builder? For example add any of the following:

```rust
pub struct Semantic<'a> {
    pub unreachable: Vec<Span>, // All the unreachable spans
}
```

```rust
pub struct SemanticNode<'a> {
  pub unreachable: bool
}
```

Think outside of the box, we have the whole compiler pipeline, we can precompute a lot of stuff!

rzvxa:
> @rzvxa Instead of doing the check inside the rule, are you able to mark the statement as unreachable during the cfg building phase inside semantic builder? For example add any of the following:
> 
> ```rust
> pub struct Semantic<'a> {
>     pub unreachable: Vec<Span>, // All the unreachable spans
> }
> ```
> 
> In the rule you just print out these spans in one go - a rule with only one line of code lol.
> 
> Or:
> 
> ```rust
> pub struct SemanticNode<'a> {
>   pub unreachable: bool
> }
> ```
> 
> Think outside of the box, we have the whole compiler pipeline, we can precompute a lot of stuff!

Doesn't adding this to the semantic mean we have to pay for this cost in places where it isn't necessary? Not to mention the amount of code we have to add to our semantic builder to compute and determine the reachability.

I'm sure with this much performance regression even I wouldn't like my current results (and if it was my project I wouldn't allow it to merge) But having it in semantic doesn't feel right to me either. Maybe we can achieve a better performance by rewriting the rule to use `run_once` and compare reachablity in relation to the last rechable node instead of checking from function(or root) down.

For example if we visit an infinite loop we know for sure everything after this loop is unreachable, But in my current implementation we can't remember that information between nodes so we get to do it all over again.

Boshen:
Yes, a `run_once` through the entire CFG can do the trick! Assuming you can query out the unreachable spans.

Boshen:
In TypeScript, it marks nodes as unreachable via `FlowFlags.Unreachable`, and is queried in checker.ts by the function `isReachableFlowNode`, maybe we can draw some inspirations there 😅

Boshen:
Maybe there's a bug somewhere? Or maybe you need to add some cache to the visited nodes?

image

`checker.ts` has some really complicated TypeScript code so maybe we hit an edge case somewhere.

rzvxa:
> In TypeScript, it marks nodes as unreachable via `FlowFlags.Unreachable`, and is queried in checker.ts by the function `isReachableFlowNode`, maybe we can draw some inspirations there 😅

We already have an unreachable instruction to mark the first unreachable block but it doesn't propagate and we use it to filter out the subgraph when walking through the flow, I assume they are doing all of our checks - if not more - to set these flags correctly. I'll give the typescript approach a read before making any major changes but I have a hunch that whatever approach we use to implement this rule it should get rid of our current redundant revisits either by running once or precomputing the information(which in of itself is another kind of run once happening in the core instead of linter crate).

It all brings me to the question I was going to ask before seeing this comment. Considering that this information can be reused but comes at the cost of more memory and overhead for everything using the semantics, Do you think having it there is going to be better overall?
Let's say we will have a 15% regression for the linter benchmark(which includes everything else so the number is higher than 15% if we pinpoint it to the linter itself) but if we do it in the semantic it would be faster let's say the time is cut in half and is 7% for the doing it in the semantic. It would make our linter benchmark to be 8% faster than the other approach but in return, it makes all of the other benchmarks to be about 7% slower; in reality, things like semantic benchmarks are going to get hit much greater than something like transformers - when they are all implemented - unless we heavily lean into feature-gating it.

rzvxa:
> Maybe there's a bug somewhere? Or maybe you need to add some cache to the visited nodes?
> image
> 
> `checker.ts` has some really complicated TypeScript code so maybe we hit an edge case somewhere.

I know we are revisiting nodes over and over since we can't memorize the information from our run on the previous nodes, That's why I hope we get a great boost after getting rid of them, it would be something like going from around `O(n^3)` or even `O(n!)`  to `O(n^2)`.

rzvxa:
The sad part is that all of our dfs paths are only 8% of the performance regression the other `12%` is for the allocations these visits do.

Boshen:
> Do you think having it there is going to be better overall

We'll eventually make cfg construction optional at some point.

Regarding performance regression, I think we can accept around 10 - 15% given that we are doing a lot of work for cfg.

rzvxa:
> Regarding performance regression, I think we can accept around 10 - 15% given that we are doing a lot of work for cfg.

Just to be clear do you mean it is acceptable if we keep it in the semantic or are you talking about this rule alone?

Boshen:
> > Regarding performance regression, I think we can accept around 10 - 15% given that we are doing a lot of work for cfg.
> 
> Just to be clear do you mean it is acceptable if we keep it in the semantics or are you talking about this rule alone?

Overall performance regression of 10 to 15%.

We can keep the unreachable query in the linter rule if that's what you prefer, I was only suggesting that maybe it's better to precompute in the semantic builder. I'll let you make the decision 😁

PR #3241, opened 2024-05-11
feat(linter/eslint): Implement max-classes-per-file

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/max-classes-per-file)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs
+++ b/crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs
@@ -1,0 +1,196 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+
+use oxc_syntax::class::ClassId;
+use serde_json::Value;
+
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule};
+
+fn max_classes_per_file_diagnostic(total: usize, max: usize, span1: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warning(format!(
+        "eslint(max-classes-per-file): File has too many classes ({total}). Maximum allowed is {max}",
+    ))
+    .with_help("Reduce the number of classes in this file")
+    .with_labels([span1.into()])
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct MaxClassesPerFile(Box<MaxClassesPerFileConfig>);
+
+#[derive(Debug, Clone)]
+pub struct MaxClassesPerFileConfig {
+    pub max: usize,
+    pub ignore_expressions: bool,
+}
+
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs
```diff
@@ -0,0 +1,194 @@
...
+            if class_count == self.max {
+                if let AstKind::Class(class) = kind {
+                    span = class.span;
+                }
+            }
+
+            if !kind.is_declaration() && self.ignore_expressions {
+            } else {
+                class_count += 1;
+            }
```
jelly, line 101: Maybe there is a more Rust-like way to write this?

jelly: I could change this into `filter().cnt()` and then look up `span` via `max` but I am not 100% sure how one does it as `ClassId` is not an integer but a Struct ClassId?

Dunqing: I think you can implement both `ignore_expression` is true or not logic separately

You can add the `len` method in classes and check `class_count <= self.max` first

```rust
let mut class_count = ctx.semantic.classes().len()
if class_count <= self.max {
   return;
}
```

If true
```rust
let ast_node_id = ctx.semantic().classes().get_node_id(self.max as ClassId)
let AstKind::Class(class) = ctx.nodes().kind(*node_id);
span = class.span;
``` 

If false, you can use `ctx.semantic().classes().iter_enumerated().filter()` to get `class_count` and first `span`. if `class_count <= self.max` we can return

anonrig (approved):

crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs
jelly: Thanks for the suggestions, I implemented something along the same lines.

crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        let ast_node_id = ctx.semantic().classes().get_node_id(ClassId::from(self.max));
+        let span = if let AstKind::Class(class) = ctx.nodes().kind(ast_node_id) {
+            class.span
+        } else {
+            Span::new(0, 0)
+        };
+
+        let error =
+            format!("File has too many classes ({}). Maximum allowed is {}", class_count, self.max);
```
Dunqing, line 111: Can you put this error message into the `max_classes_per_file_diagnostic`? That way we only have to call `format!` once

Dunqing (changes requested):

Dunqing (approved):

Dunqing:
Thank you!

PR #3243, opened 2024-05-12
chore(linter): change the category of all react-perf rules to perf

### LINKED ISSUE #3242: Suspense fallback attribute should not contain JSX
Hi oxc team!

I'm unsure about the specifics, since I'm still learning React, so I want to report this as a mismatch with React docs.

According to the [React docs](https://react.dev/reference/react/Suspense), the following is valid:

```tsx
const Component = () => {
  return <Suspense fallback={<span>Loading...</span>}>...</Suspense>;
};
```

However oxlint tells me that "JSX attribute values should not contain other JSX.", which sounds like I should change it to this:

```tsx
const Component = () => {
  const loadingEleRef = useRef<ReactNode>(null);

  if (loadingEleRef.current === null) {
    loadingEleRef.current = <span>Loading...</span>;
  }

  return <Suspense fallback={loadingEleRef.current}>...</Suspense>;
};
```

What's the correct way?

Dunqing:
close: #3242

### DIFF
--- a/crates/oxc_linter/src/rules/react_perf/jsx_no_jsx_as_prop.rs
+++ b/crates/oxc_linter/src/rules/react_perf/jsx_no_jsx_as_prop.rs
@@ -34,7 +34,7 @@ declare_oxc_lint!(
     /// <Item callback={this.props.jsx} />
     /// ```
     JsxNoJsxAsProp,
-    correctness
+    perf
 );
 
 impl Rule for JsxNoJsxAsProp {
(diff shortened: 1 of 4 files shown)


Dunqing:
* **#3243** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #3258, opened 2024-05-13
fix(transformer): do no add __self when the jsx is inside constructor

Dunqing:


Dunqing:
* **#3258** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_traverse/src/walk.rs
```diff
@@ -2228,7 +2228,7 @@ pub(crate) unsafe fn walk_function<'a, Tr: Traverse<'a>>(
 ) {
     traverser.enter_function(&mut *node, ctx);
     ctx.push_stack(Ancestor::FunctionId(ancestor::FunctionWithoutId(node)));
-    let has_scope = matches!(ctx.ancestor(2).unwrap(), Ancestor::MethodDefinitionValue(_));
+    let has_scope = !matches!(ctx.ancestor(2).unwrap(), Ancestor::MethodDefinitionValue(_));
```
Dunqing, line 2231: I found a bug about the `ScopeFlags` of the constructor method is incorrect. The `JSXOpeningElement` is inside the constructor method, But I get the scope is `ScopeFlags::Function` and the parent scope is `ScopeFlags::Constructor`. It seems that both construct and function have pushed a scope flag.

overlookmotel: Ah. I will investigate. This shouldn't happen.

Boshen (approved):
Shall we merge and iterate?

crates/oxc_transformer/src/react/mod.rs
```diff
@@ -78,8 +78,29 @@ impl<'a> React<'a> {
...
+
+        if (!is_constructor
+            || ctx
+                .find_ancestor(|ancestor| match ancestor {
+                    Ancestor::ClassBody(class) => FinderRet::Found(class.super_class()),
+                    _ => FinderRet::Continue,
+                })
+                .is_none())
+            && self.options.is_jsx_self_plugin_enabled()
+        {
```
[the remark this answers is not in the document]
Dunqing: > On the scopes logic itself, it looks _almost_ correct to me. But...
> 
> 1. What about if we're not in a function at all? Then `__self` shouldn't be added. I guess this means checking for `ScopeFlags::Top` and making `is_constructor = true` in that case (and rename it `is_constructor_or_top_level`).

You are right. This is the case because we are currently writing the plugin based on Babel's test cases for the current lack of such tests so I haven't done anything based on that yet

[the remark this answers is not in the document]
Dunqing: Should we add `ScopeFlags::Function` for MethodDefinition? Like what VisitMut did?

overlookmotel:
> Shall we merge and iterate?

I don't think we should. If we can rebase on top of #3273, then we can make sure the logic is correct before merging.

crates/oxc_syntax/src/scope.rs
```diff
@@ -42,6 +42,10 @@ impl ScopeFlags {
         self.contains(Self::StrictMode)
     }
 
+    pub fn is_block(&self) -> bool {
```
overlookmotel, line 45: I don't think this is correct. `*self == Self::StrictMode` can be true when in a function but not within a block within that function.

Dunqing: That's what I want, I just want to know if in the BlockStatement

overlookmotel: Yes, I think just `self.is_empty()` is what you need for that.

With `is_block` written as it is, `is_block()` would be true when inside `foo` here:

```
'use strict';
function foo() {
  // `ctx.scope().is_block()` is `true`.
  // `ctx.scope().is_empty()` is `false`.
  // `ctx.scope().is_function()` is `true`.
}
```

overlookmotel: Just to say: I don't like the design of `ScopeFlags`. I think it's very confusing!

Dunqing: > Yes, I think just `ScopeFlags::is_empty` is what you need for that.
> 
> `ctx.scope().is_empty()` answers the question "am I in a nested block here?"
> 
> With `is_block` written as it is, `is_block()` would be true when inside `foo` here:
> 
> ```
> 'use strict';
> function foo() {
>   // `ctx.scope().is_block()` is `true`.
>   // `ctx.scope().is_empty()` is `false`.
>   // `ctx.scope().is_function()` is `true`.
> }
> ```

No. BlockStatement will inherit StrictMode. So actually result is

```ts
'use strict';
function foo() {
  // ScopeFlags::StrictMode | ScopeFlags::Function
  // `ctx.scope().is_block()` is `false`.
  // `ctx.scope().is_empty()` is `false`.
  // `ctx.scope().is_function()` is `true`.

  {
     // BlockStatement
     // ScopeFlags::StrictMode
     // `ctx.scope().is_block()` is `true`.
     // `ctx.scope().is_empty()` is `false`.
  }
}
```

You can check in [Playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAIDsgYCAgICAgICTHcrm5BJ12ajhv4gxb1aOKaBCbo9ZL1gx%2BdsX%2B2Uf9RZ%2FjolWJDOoDIMVPusJRxBm6byPrmkDH4xfhqNZbKCJEC7QmR2R2tH8zOQ%2Bu9NBF2tjJyUsHbE3dbo%2FTXe%2BmNWJbHOoIGeyk5W3jC9o34UYRQlHUnyW7zoifwuClYagYuI2LSMmVd90oCcdKBuG1rQeCwPz5ayL0uVjEk6%2Bf1%2FYerI%3D)

overlookmotel: Ah ha! Now I understand why we disagree. We're talking at cross-purposes. You are familiar with `ScopeFlags` from semantic, whereas I'm not.

I copied how `ScopeFlags` inherit from parent scope in `Traverse` from `VisitMut`. But I think `VisitMut`'s logic for it is different from semantic's.

So I am talking about how `ScopeFlags` behaves in `Traverse`, whereas you are talking about how it behaves in `Semantic`. We are both right, but talking about different things!

Obviously, this misalignment is bad. We should change the behavior in both `VisitMut` and `Traverse` to match `Semantic`, and then this confusion will go away.

overlookmotel: Oh no. Actually scratch all that. I was talking rubbish. `self.is_empty() || *self == Self::StrictMode` is correct, I just read it wrong.

There *is* a misalignment between `Semantic` and `Traverse`, but that's not the problem here.

Sorry I've wasted your time with this.

crates/oxc_transformer/src/react/mod.rs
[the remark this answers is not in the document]
overlookmotel: > Should we add `ScopeFlags::Function` for MethodDefinition? Like what VisitMut did?

Yes I think we should. But from what I can see, `Visit` and `VisitMut` use `ScopeFlags::empty()` for class methods. If we change this in `Traverse`, we should change it in `Visit` and `VisitMut` too, right?

https://github.com/oxc-project/oxc/blob/6fd7a3c42712487430ddfd627f61654a32ab58ce/crates/oxc_ast/src/visit/visit.rs#L1379-L1390

https://github.com/oxc-project/oxc/blob/6fd7a3c42712487430ddfd627f61654a32ab58ce/crates/oxc_ast/src/visit/visit_mut.rs#L1416-L1431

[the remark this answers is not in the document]
overlookmotel: Ah no, now I get it! I missed some logic. `Visit::walk_function` sets the `ScopeFlags::Function` bit. This is my mistake - I translated the logic wrongly in Traverse. I'll fix it now.

crates/oxc_syntax/src/scope.rs
Dunqing: Haha, It's okay. By the way, I don't like the implementation of `is_block`. I think we should add `ScopeFlags::Block` instead of `ScopeFlags::Empty`.  That will be easy to implement `is_block`

crates/oxc_transformer/src/react/mod.rs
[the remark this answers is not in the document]
overlookmotel: #3277

overlookmotel:
> @overlookmotel Please review this again. I think we can merge first

@Dunqing Please see https://github.com/oxc-project/oxc/pull/3258#discussion_r1600030436 above

crates/oxc_transformer/src/react/mod.rs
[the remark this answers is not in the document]
Dunqing: > If you can merge that and then rebase this PR on top, then I'll review again. I do think there's still one problem remaining (but I've been wrong about everything else so far, so maybe not!?)

merged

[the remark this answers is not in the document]
overlookmotel: I'm going to resolve this conversation and start again on the up-to-date code.

overlookmotel:
@Dunqing I'm going to merge this. I do think there's still a couple of problems, but I'll do a follow-up PR.

overlookmotel:
Turns out Dunqing's implementation was exactly correct. All my comments saying there were problems were completely wrong!

PR #3276, opened 2024-05-14
refactor(linter): rewrite react/require-render-return

### LINKED ISSUE #3245: perf(linter): react/require_render_return
react/require_render_return is slow

image

It seems like we need follow https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/require-render-return.js

mysteryven:
Closes: #3245

It's gone now, curious about why didn't have performance improvement.

CleanShot 2024-05-14 at 20 57 17@2x

This picture generated by running all the rules.

### DIFF
--- a/crates/oxc_linter/src/rules/react/require_render_return.rs
+++ b/crates/oxc_linter/src/rules/react/require_render_return.rs
@@ -1,18 +1,15 @@
-use oxc_ast::{
-    ast::{Argument, ClassElement, Expression, FunctionBody, ObjectPropertyKind},
-    AstKind,
-};
+use oxc_ast::{ast::Expression, AstKind};
 use oxc_diagnostics::OxcDiagnostic;
 
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
+use oxc_semantic::AstNodeId;
+use oxc_span::{GetSpan, Span};
+use rustc_hash::FxHashSet;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/react/require_render_return.rs
```diff
@@ -52,69 +49,116 @@ declare_oxc_lint!(
...
-                        _ => None,
-                    })
-                {
-                    if is_arrow_function_expression || has_return_in_fn_body(fn_body) {
-                        return;
-                    }
-
-                    ctx.diagnostic(require_render_return_diagnostic(fn_body.span));
-                }
+        for node in ctx.nodes().iter() {
```
Boshen, line 55: Is it possible to not iterate through nodes in in `run_once` for correctness rules?

Boshen:
I don't get this ... https://github.com/jsx-eslint/eslint-plugin-react/blob/d97e3ed96afe77a56fdc6fc7bdec11c28bc256e2/lib/rules/require-render-return.js#L45-L47

Why does it need to save the component and then evaluate it on program exit?

Boshen:
If we are inside the render method, are we able to find explicit returns through the control flow graph 🤔 ?

mysteryven:
After reviewing the hook rule, I have a poor concept of cfg, it will be fun to give it a try!

PR #3287, opened 2024-05-15
feat(transformer): do not add self attribute in react/jsx plugin

Dunqing:
follow-up: https://github.com/oxc-project/oxc/pull/3258.


Boshen:
Can you restack this on top of main?

Boshen (approved):

PR #3293, opened 2024-05-15
feat(transformer_conformance): skip some tests that are known to fail

Dunqing:
I don't think we need to be completely consistent with Babel's output. Because Babel's output doesn't always make sense.

### DIFF
--- a/tasks/transform_conformance/babel.snap.md
+++ b/tasks/transform_conformance/babel.snap.md
@@ -1,6 +1,6 @@
 commit: 4bd1b2c2
 
-Passed: 310/362
+Passed: 310/351
 
 # All Passed:
 * babel-preset-react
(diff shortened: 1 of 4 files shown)


Dunqing:
* **#3293** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3293?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):
I agree.

overlookmotel:
As usual, I miss the chance to comment before the PR is merged, but...

Rather than disabling these tests entirely, could we convert them into "exec" tests which test that *behavior* of the code after transform is what it should be with assertions? If we make sure those assertions pass on both Babel's output and Oxc's, then we know we're fulfilling the spirit of Babel's tests, even if we're not producing the exact same output.

It's great that we don't have to write tons of tests because we can just use Babel's, but when we can't do that, we do still want *some* tests, don't we?

Dunqing:
> As usual, I miss the chance to comment before the PR is merged, but...
> 
> Rather than disabling these tests entirely, could we convert them into "exec" tests which test that _behavior_ of the code after transform is what it should be with assertions? If we make sure those assertions pass on both Babel's output and Oxc's, then we know we're fulfilling the spirit of Babel's tests, even if we're not producing the exact same output.
> 
> It's great that we don't have to write tons of tests because we can just use Babel's, but when we can't do that, we do still want _some_ tests, don't we?

Yes, But to convert to `exec` tests we need to patch the code. It's going to be a lot of work. I think we can just make sure the results are as consistent as possible for now. Once the transformer is almost complete we can optimize it and write our own tests.

PR #3296, opened 2024-05-15
perf(lexer): use bitshifting when parsing known integers

DonIsaac:
## What This PR Does

- perf(lexer): use bit shifting when parsing hex, octal, and binary integers instead of `mul_add`-ing on `f64`s. Check out the difference in assembly generated [here](https://godbolt.org/z/zMEKaeYzh)
- perf(lexer): skip redundant utf8 check when parsing BigInts
- refactor(lexer): remove `unsafe` usage (as per @overlookmotel's request [here](https://github.com/oxc-project/oxc/pull/3283#issuecomment-2111814598))
- test(lexer): add numeric parsing unit tests

I don't expect this PR to have a large performance improvement, since the most common case (`Kind::Decimal`) is not affected. We could do this, however, by splitting `Kind::Decimal` into `Kind::DecimalFloat` and `Kind::DecimalInt` when the lexer encounters a `.`


crates/oxc_parser/src/lexer/number.rs
```diff
@@ -120,5 +124,193 @@ pub fn parse_big_int(s: &str, kind: Kind) -> Result<BigInt, &'static str> {
...
+
+#[cfg(test)]
+#[allow(clippy::unreadable_literal, clippy::mixed_case_hex_literals)]
+mod test {
+
+    use super::{parse_float, parse_int, Kind};
+
+    /// Assert that all (&str, number) items in an [`Iterator`] parse to the
+    /// expected number. `number` will be casted to an [`f64`].
+    macro_rules! assert_all_eq {
```
Boshen, line 141: Do we really need macros instead of simple functions 😅 . This file is at the bottom of the dependency tree, it is valuable to save as much compilation time as possbile.

Boshen:
> use bit shifting 

Can you comment why it's better on the spot for future contributors.

crates/oxc_parser/src/lexer/number.rs
overlookmotel: I would also prefer to lose the macros here, but not for compile time reasons.

I feel like tests should be as "stupid" as possible so it's really clear exactly what they test and to verify that they're testing what you think they are.

In my view, having a pretty simple macro to shorten a repeated pattern is fine, but this macro is too clever! (by which I mean it takes brain power to understand)

To remove the macro while also avoiding repetition, could you make `assert_all_eq!` (or at least parts of it) into a normal function?

overlookmotel:
Just to say, the bitshift algorithm is brilliant! Very smart.

Interesting that compiler unrolls the loop. Looks like it unrolls in batches of 4 characters, which should be good for this use case. If it had used batches of 8 chars for octals, that'd likely be a de-opt because JS rarely contains numbers which are as large as 8 digits (decimal or octal).

DonIsaac:
> > use bit shifting
> 
> Can you comment why it's better on the spot for future contributors.

In this PR discussion, or in the code?

overlookmotel:
> > > use bit shifting
> > 
> > 
> > Can you comment why it's better on the spot for future contributors.
> 
> In this PR discussion, or in the code?

I think Boshen means in the code.

crates/oxc_parser/Cargo.toml
```diff
@@ -24,12 +24,13 @@ oxc_ast         = { workspace = true }
 oxc_syntax      = { workspace = true }
 oxc_diagnostics = { workspace = true }
 
-static_assertions = { workspace = true }
 assert-unchecked  = { workspace = true }
 bitflags          = { workspace = true }
-rustc-hash        = { workspace = true }
 num-bigint        = { workspace = true }
+num-traits        = { workspace = true }
```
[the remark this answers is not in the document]
Boshen: Ok let's keep it for now.

crates/oxc_parser/src/lexer/number.rs
DonIsaac: Removed

Boshen (approved):

Boshen (approved):
Brilliant work! Thank you!

Issue #2870, opened 2024-03-30
Transformer: support babel `targets`


Boshen:
https://babeljs.io/docs/options#targets

Boshen:
https://github.com/oxc-project/oxc/pull/3371

PR #3298, opened 2024-05-16
feat(transformer): add preset-env

Dunqing:
Related issues: https://github.com/oxc-project/oxc/issues/2870

The `targets` is an option for preset-env

This preset has a lot of test failures, so let's skip it for now.

### DIFF
--- a/crates/oxc_transformer/src/env/mod.rs
+++ b/crates/oxc_transformer/src/env/mod.rs
@@ -1,0 +1,21 @@
+mod options;
+
+use std::rc::Rc;
+
+pub use self::options::EnvOptions;
+/// [Preset Env](https://babel.dev/docs/babel-preset-env)
+///
+/// This preset is a smart preset that allows you to use the latest JavaScript without needing to micromanage
+/// which syntax transforms (and optionally, browser polyfills) are needed by your target environment(s).
+/// This both makes your life easier and JavaScript bundles smaller!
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#3298** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
Is `@babel/preset-env` a pre-step for babel plugins? It feels like the API is something like

`Transformer::from_preset_env()` and produces a transformer with all the plugins set correctly.

What do you think?

Dunqing:
> Is `@babel/preset-env` a pre-step for babel plugins? It feels like the API is something like
> 
> `Transformer::from_preset_env()` and produces a transformer with all the plugins set correctly.
> 
> What do you think?

Yes. The `@babel/preset-env` just returns a series plugins. `Transformer::from_preset_env()` is simpler than preset-env. I like it.

Dunqing:
We decided to use `Transformer::from_babel_config()` instead of `preset-env`. This function will handle what preset-env does and return the transformer options and errors for unsupported plugins/options.

This function signature like this
```rust
pub fn from_babel_config(config: serde_json::Value) -> Result<TransformOptions, Vec<Error>> {
}
```

PR #3302, opened 2024-05-16
refactor(ast): store `ScopeId` in AST nodes

overlookmotel:
Add `scope_id` field to all AST nodes which can constitute a scope (`Program`, `Function`, `BlockStatement` etc).

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -14,7 +14,7 @@ use oxc_syntax::{
         AssignmentOperator, BinaryOperator, LogicalOperator, UnaryOperator, UpdateOperator,
     },
     reference::{ReferenceFlag, ReferenceId},
-    scope::ScopeFlags,
+    scope::{ScopeFlags, ScopeId},
     symbol::SymbolId,
 };
 #[cfg(feature = "serialize")]
@@ -46,7 +46,7 @@ export interface FormalParameterRest extends Span {
     scope(ScopeFlags::Top),
     strict_if(self.source_type.is_strict() || self.directives.iter().any(Directive::is_use_strict))
 )]
-#[derive(Debug, Hash)]
+#[derive(Debug)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(tag = "type", rename_all = "camelCase"))]
 pub struct Program<'a> {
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#3302** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
@rzvxa Can you help us review this PR stack, I want to have a second pair of eyes 😅

rzvxa:
> @rzvxa Can you help us review this PR stack, I want to have a second pair of eyes 😅

Sure thing, Although it might take me a few hours before reviewing this.

@overlookmotel Quick question, why not store node_id? so we can have both Scope and CFG information deferred from that? I wasn't following the chain of events that led to this PR, Can you bring me up to date with the context of problem we are trying to solve here?

rzvxa:
> > > @rzvxa Can you help us review this PR stack, I want to have a second pair of eyes 😅
> > 
> > 
> > Sure thing, Although it might take me a few hours before I'm done with the review.
> > @overlookmotel Quick question, why not store node_id? so we can have both Scope and CFG information deferred from that? I wasn't following the chain of events that led to this PR, Can you bring me up to date with the context of problem we are trying to solve here?
> 
> We are building up the prerequisite for [#3251 (comment)](https://github.com/oxc-project/oxc/discussions/3251#discussioncomment-9416826)
> 
> We want to put scope and symbol table into ast traverse so we can access the data at least.

So why not store node_id? It can point to both the scope and symbol data associated with the node, It has a bit more overhead because of indirection but would future proof the AST referencing since in the future we can easily add and remove information associated with that ID. I have no objection, Just curious to know the reasoning.

Boshen:
> Sure thing, Although it might take me a few hours before I'm done with the review.

@overlookmotel will be afk for the next few days so take your time 😀 

> So why not store node_id? 

I'm too eager to have a working solution right now for finding all the edge cases, so we are going after smaller changes right now. Once we have everything passing, we'll reevaluate our decisions and infrastructure (for the 10th time).

rzvxa (approved):

overlookmotel:
Surprised (but pleased) to see the addition of `ScopeId` to a bunch of AST types barely registers on benchmarks. There's basically no cost to doing this.

overlookmotel:
@rzvxa Not ignoring your questions above, just have to run now. Will come back to it.

overlookmotel:
> So why not store node_id? It can point to both the scope and symbol data associated with the node, It has a bit more overhead because of indirection but would future proof the AST referencing since in the future we can easily add and remove information associated with that ID. I have no objection, Just curious to know the reasoning.

@rzvxa To come back (finally) on this question...

First of all, this v1 implementation is a rough attempt to implement what we need right now (scope and binding information) as swiftly as possible so we can proceed with the React+TS transforms that we're currently blocked on. The easiest way to do that without re-inventing any wheels was just to reuse `Semantic`.

However, we can't use all of `Semantic`. `AstNode` contains `AstKind`, which contains refs to AST nodes. This is incompatible with having `&mut` refs to AST nodes in the `enter_*` and `exit_*` visitors. It's the same problem as before - aliasing references.

So currently, in terribly inefficient fashion, we run semantic on the AST, and then throw away most of the data it produces. Clearly, this will need a complete rethink once we have transformer milestone 1 complete. So none of this is set in stone.

But at this point, given the problem of aliasing references, I'm unclear what `AstNodeId` would give us - since we can't use the node ID to get back to an AST node. Am I missing something?

More broadly, the approach I was thinking of was to start by giving the transformer the *minimum* of information it needs, and then adding more if and when it becomes clear that we need it - rather than starting out by giving it the whole kitchen sink and then trying to pare it back. The control flow graph, for example, I think we should try to avoid using in the transformer if at all possible. *Maybe* we need it in some cases, but hopefully not. If we do use it, then we have to somehow keep it in sync in every transform, which will become quite a burden.

The other big reason why I don't have a clear answer to your question is simply that I don't really understand `Semantic` and its uses yet. I've not had time to dig into it properly. I've been delaying doing that until after the immediate need (the missing TS transforms) is completed. So apologies that I just don't know enough to engage with your question as thoroughly as I'd like.

PR #3315, opened 2024-05-16
fix(tasks/lint_rules): Integrate react and react-hooks rules

### LINKED ISSUE #2174: ☂️ eslint-plugin-react-hooks

leaysgur:
Fixes #2174 , now `react-hooks` plugin and its rules are managed by `react` plugin issue.

### DIFF
--- a/tasks/lint_rules/src/eslint-rules.cjs
+++ b/tasks/lint_rules/src/eslint-rules.cjs
@@ -219,7 +219,7 @@ const loadPluginNextRules = (linter) => {
 
 /**
  * @typedef {{
- *   npm: Array<string>;
+ *   npm: string[];
  *   issueNo: number;
  * }} TargetPluginMeta
  * @type {Map<string, TargetPluginMeta>}
--- a/tasks/lint_rules/src/eslint-rules.cjs
+++ b/tasks/lint_rules/src/eslint-rules.cjs
@@ -181,12 +181,12 @@ const loadPluginReactRules = (linter) => {
 
     linter.defineRule(prefixedName, rule);
   }
-};
 
-/** @param {import("eslint").Linter} linter */
-const loadPluginReactHooksRules = (linter) => {
+  // `react-hooks` plugin is available along with `react` plugin
   for (const [name, rule] of Object.entries(pluginReactHooksAllRules)) {
-    const prefixedName = `react-hooks/${name}`;
+    // This may be conflict with `react` plugin
+    // (but `react-hooks` plugin has only 2 rules, so it's fine...!)
+    const prefixedName = `react/${name}`;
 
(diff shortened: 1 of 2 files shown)


rzvxa (approved):
Nice!

leaysgur:
BTW, will linter configurations be also supported?

```json
{
  "rules": {
    // ...
    "react-hooks/rules-of-hooks": "error",
    "react-hooks/exhaustive-deps": "warn"
  }
}
```

> https://github.com/facebook/react/tree/main/packages/eslint-plugin-react-hooks#custom-configuration

If we need to change like this, might there be compatibility issues...?

```json
{
  "rules": {
    // ...
    "react/rules-of-hooks": "error",
    "react/exhaustive-deps": "warn"
  }
}
```

rzvxa:
Do you mean in the long run? or would it cause issues with this task?

I guess we can alias the old rule name for a while, Or even better we can anticipate the wrong option and produce a nice error message pointing the user toward the right option.

leaysgur:
It was not about this task, but about the implementation of the Linter itself.

I thought that `oxlint` would be a drop-in replacement for `eslint`, allowing the use of existing `eslintrc.json` as is. (Although there are some unsupported rules)

I may just be assuming this on my own though...! 😅

rzvxa:
Well, I don't know if that is the goal or not, What I know is that there are a few differences in our rules even not considering the unsupported rules, But with a few compromises, we can get this effect for sure.

I just think providing an automated way of upgrading could be easier; Since once we hit a stable release it is going to be a one-way trip.
We can't keep the backward compatibility forever since it would contain a bunch of legacy options and soon enough we would also have our own legacy options to worry about. What we can always do is provide an auto upgrade and/or nice error messages while detecting incompatible rules, And suggesting possible replacement options if there is one.

PR #3321, opened 2024-05-17
feat(linter): eslint/no-constructor-return

woai3c:


Boshen:
This approach is slow, it is preferred to use the control flow graph 

https://github.com/eslint/eslint/blob/5c28d9a367e1608e097c491f40b8afd0730a8b9e/lib/rules/no-constructor-return.js#L36-L41

woai3c:
> This approach is slow, it is preferred to use the control flow graph
> 
> https://github.com/eslint/eslint/blob/5c28d9a367e1608e097c491f40b8afd0730a8b9e/lib/rules/no-constructor-return.js#L36-L41

Ok, Is there a demo related to the control flow graph in oxc? I don't understand how to implement it in oxc.

Boshen:
> > This approach is slow, it is preferred to use the control flow graph
> > https://github.com/eslint/eslint/blob/5c28d9a367e1608e097c491f40b8afd0730a8b9e/lib/rules/no-constructor-return.js#L36-L41
> 
> Ok, Is there a demo related to the control flow graph in oxc? I don't understand how to implement it in oxc.

CFG is a bit complicated, you may start from https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/examples/cfg.rs

and also the linter rules that use it, e.g. eslint/getter_return.rs

woai3c:
> > > This approach is slow, it is preferred to use the control flow graph
> > > https://github.com/eslint/eslint/blob/5c28d9a367e1608e097c491f40b8afd0730a8b9e/lib/rules/no-constructor-return.js#L36-L41
> > 
> > 
> > Ok, Is there a demo related to the control flow graph in oxc? I don't understand how to implement it in oxc.
> 
> CFG is a bit complicated, you may start from https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/examples/cfg.rs
> 
> and also the linter rules that use it, e.g. eslint/getter_return.rs

@Boshen @rzvxa Hi, I have been reading the code for [require_render_return](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/react/require_render_return.rs#L93). While implementing `no-constructor-return`, I encountered a problem. I need to access the return statement node for two reasons: first, to display the position of the return statement span, and second, to check if the return statement node is in the block statement.

Consider the following example:
```js
class C { constructor(a) { if (!a) { return } else { a() } } }
``` 
The code above is correct. The `return` is used for control flow. Therefore, I would like to know how to access the return statement node using `neighbors_filtered_by_edge_weight`.

image

rzvxa:
As far as I know, this information is lost in our current control flow implementation. For some other reason - which isn't that far off from what you need here #3238 - I'm working on adding the list of instructions to the basic blocks and reworking some parts of cfg to make it more versatile. So if things go according to the plan soon it should be possible to somehow access our nodes in basic blocks.

If you only need to know whether a return has an expression or not you can use the `AssignmentValue` - which is what you are already doing in this screenshot - is that not sufficient for this problem? What do you need from the return statement? maybe we can work around it.

woai3c:
> As far as I know, this information is lost in our current control flow implementation. For some other reason - which isn't that far off from what you need here #3238 - I'm working on adding the list of instructions to the basic blocks and reworking some parts of cfg to make it more versatile. So if things go according to the plan soon it should be possible to somehow access our nodes in basic blocks.
> 
> If you only need to know whether a return has an expression or not you can use the `AssignmentValue` - which is what you are already doing in this screenshot - is that not sufficient for this problem? What do you need from the return statement? maybe we can work around it.

I need to access the return statement node because the error information related to the span is incorrect.

image

The second season is that I need to know if the return statement node in a  block statement. consider the following example:
```ts
class C { constructor(a) { if (!a) { return } else { a() } } }
```
The above code should be correct. Because the `return` is used for flow control. 

I'm not sure if there's a way to solve these two problems. Looking forward to your suggestions. Here is my code:
```rs
impl Rule for NoConstructorReturn {
        if !matches!(node.kind(), AstKind::ArrowFunctionExpression(_) | AstKind::Function(_)) {
            return;
        }

        let Some(parent) = ctx.nodes().parent_node(node.id()) else {
            return;
        };

        if contains_return_statement(node, ctx) {
            match parent.kind() {
                AstKind::MethodDefinition(method)
                    if method.kind == MethodDefinitionKind::Constructor =>
                {
                    ctx.diagnostic(no_constructor_return_diagnostic(method.span));
                }
                _ => {}
            };
        }
    }
}

const KEEP_WALKING_ON_THIS_PATH: bool = true;
const STOP_WALKING_ON_THIS_PATH: bool = false;

fn contains_return_statement<'a>(node: &AstNode<'a>, ctx: &LintContext<'a>) -> bool {
    let cfg = ctx.semantic().cfg();
    let state = neighbors_filtered_by_edge_weight(
        &cfg.graph,
        node.cfg_ix(),
        &|edge| match edge {
            // We only care about normal edges having a return statement.
            EdgeType::Normal => None,
            // For these two type, we flag it as not found.
            EdgeType::NewFunction | EdgeType::Backedge => Some(FoundReturn::No),
        },
        &mut |basic_block_id, _state_going_into_this_rule| {
            // If its an arrow function with an expression, marked as founded and stop walking.
            if let AstKind::ArrowFunctionExpression(arrow_expr) = node.kind() {
                if arrow_expr.expression {
                    return (FoundReturn::Yes, STOP_WALKING_ON_THIS_PATH);
                }
            }

            for entry in cfg.basic_block_by_index(*basic_block_id) {
                if let BasicBlockElement::Assignment(to_reg, val) = entry {
                    if matches!(to_reg, Register::Return)
                        && matches!(val, AssignmentValue::NotImplicitUndefined)
                    {
                        return (FoundReturn::Yes, STOP_WALKING_ON_THIS_PATH);
                    }
                }
            }

            (FoundReturn::No, KEEP_WALKING_ON_THIS_PATH)
        },
    );

    state.iter().any(|&state| state == FoundReturn::Yes)
}

#[derive(Clone, Copy, Debug, Default, PartialEq)]
enum FoundReturn {
    #[default]
    No,
    Yes,
}

#[test]
fn test() {
    use crate::tester::Tester;

    let pass = vec![
        "function fn() { return }",
        "function fn(kumiko) { if (kumiko) { return kumiko } }",
        "const fn = function () { return }",
        "const fn = function () { if (kumiko) { return kumiko } }",
        "const fn = () => { return }",
        "const fn = () => { if (kumiko) { return kumiko } }",
        "return 'Kumiko Oumae'",
        "class C {  }",
        "class C { constructor() {} }",
        "class C { constructor() { let v } }",
        "class C { method() { return '' } }",
        "class C { get value() { return '' } }",
        "class C { constructor(a) { if (!a) { return } else { a() } } }",
        "class C { constructor() { function fn() { return true } } }",
        "class C { constructor() { this.fn = function () { return true } } }",
        "class C { constructor() { this.fn = () => { return true } } }",
    ];

    let fail = vec![
        // "class C { constructor() { return } }",
        "class C { constructor() { return '' } }",
        "class C { constructor(a) { if (!a) { return '' } else { a() } } }",
    ];

    Tester::new(NoConstructorReturn::NAME, pass, fail).test_and_snapshot();
}
```

rzvxa:
Can you test this approach and see how it performs?

```rust
impl Rule for NoConstructorReturn {
        let AstKind::ReturnStatement(ret) = node.kind() else { return };
        let Some(is_in_constructor_root) = is_in_constructor_root(ctx, node.id()) else { return };

        if is_in_constructor_root {
            ctx.diagnostic(no_constructor_return_diagnostic(ret.span));
        } else if ret.argument.is_some() && is_definitely_in_constructor(ctx, node.id()) {
            ctx.diagnostic(no_constructor_return_diagnostic(ret.span));
        }
    }
}

fn is_constructor(node: &AstNode<'_>) -> bool {
    matches!(
        node.kind(),
        AstKind::MethodDefinition(MethodDefinition { kind: MethodDefinitionKind::Constructor, .. })
    )
}

/// Checks to see if the given node is in the root of a constructor.
/// Returns `None` if it isn't possible for this node to be in a constructor.
fn is_in_constructor_root(ctx: &LintContext, node_id: AstNodeId) -> Option<bool> {
    ctx.nodes().ancestors(node_id).nth(3).map(|id| ctx.nodes().get_node(id)).map(is_constructor)
}

fn is_definitely_in_constructor(ctx: &LintContext, node_id: AstNodeId) -> bool {
    ctx.nodes()
        .ancestors(node_id)
        .map(|id| ctx.nodes().get_node(id))
        .skip_while(|node| !node.kind().is_function_like())
        .nth(1)
        .is_some_and(is_constructor)
}
```

Obviously, it isn't as good as what you've been doing here, And we might want to switch back to that when we have the improved control flow, But I suspect that the snippet above should perform OKish at least for now.

rzvxa:
The simpler form of this was already present in your cases - although it was commented out - and in the original [eslint plugin](https://github.com/eslint/eslint/blob/5c28d9a367e1608e097c491f40b8afd0730a8b9e/tests/lib/rules/no-constructor-return.js#L48). However, You are right that case [passes in eslint](https://eslint.org/play/#eyJ0ZXh0IjoiY2xhc3MgQyB7IGNvbnN0cnVjdG9yKCkgeyAgeyByZXR1cm4gfSB9IH0iLCJvcHRpb25zIjp7InJ1bGVzIjp7Im5vLWNvbnN0cnVjdG9yLXJldHVybiI6WyJlcnJvciJdfSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=).

But do these 2 have a real difference? Why should one pass and the other one fail? There are other rules that would point out the redundant block and return statement, Is it possible that it is just a dormant bug happening in a stupid edge case that nobody noticed?

```js
class C { constructor() { return } }
class C { constructor() { { { return } } } }
```

Well in either case you can just ignore the edit and it would pass the test.

woai3c:
@rzvxa I have replaced the code with your suggestion. And the performance is ok. Could you please check if there are other any other issues?

woai3c:
> It might actually be [a bug](https://github.com/eslint/eslint/issues/18486#issuecomment-2125095858), although a different one. So if you don't mind let's wait to see if this behavior is going to change or not. On the bright side, we might get to ship this with the new cfg, It may or may not perform better; Worth a benchmark in my opinion.

ok, I have seen the discussion about the rule. It seems that there are different opinions within the ESLint team.

rzvxa:
@woai3c Hello, How are you doing my friend? I hope you've been having an amazing day!
I just wanted to inform you about the recent progress regarding this rule. The ESLint team decided to go with allowing all implicit undefined returns from constructors, There are some changes in passing and failing cases for this rule which you can check out [here](https://github.com/eslint/eslint/blob/f6534d14033e04f6c7c88a1f0c44a8077148ec6b/tests/lib/rules/no-constructor-return.js#L45).

So if you implement the required changes I can approve this PR for merge.

There is also #3381 which is almost ready; It provides `AstNodeId` appropriate for each CFG instruction so with the help of this you can do what you wanted originally and implement it using CFG alone.

The decision to wait for the new CFG to merge or move on with this rule as is and later on expect it to be refactored to use CFG(either by you or somebody else) is yours.

woai3c:
> @woai3c Hello, How are you doing my friend? I hope you've been having an amazing day! I just wanted to inform you about the recent progress regarding this rule. The ESLint team decided to go with allowing all implicit undefined returns from constructors, There are some changes in passing and failing cases for this rule which you can check out [here](https://github.com/eslint/eslint/blob/f6534d14033e04f6c7c88a1f0c44a8077148ec6b/tests/lib/rules/no-constructor-return.js#L45).
> 
> So if you implement the required changes I can approve this PR for merge.
> 
> There is also #3381 which is almost ready; It provides `AstNodeId` appropriate for each CFG instruction so with the help of this you can do what you wanted originally and implement it using CFG alone.
> 
> The decision to wait for the new CFG to merge or move on with this rule as is and later on expect it to be refactored to use CFG(either by you or somebody else) is yours.

Thank you for the update! That's great news. I will implement the required changes a bit later.

crates/oxc_linter/src/rules/eslint/no_constructor_return.rs
```diff
@@ -0,0 +1,123 @@
...
+impl Rule for NoConstructorReturn {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::ReturnStatement(ret) = node.kind() else { return };
+        if ret.argument.is_none() {
+            return;
+        }
+
+        let Some(is_in_constructor_root) = is_in_constructor_root(ctx, node.id()) else { return };
+
+        if is_in_constructor_root || is_definitely_in_constructor(ctx, node.id()) {
```
rzvxa, line 59: I think we don't need the `is_in_constructor_root` check anymore; So it can be removed along with its function, Other than that it looks good to me.

Thanks again for this contribution and your patience.

woai3c: ok, got it.

woai3c:
@rzvxa Hi, rzvxa. I have addressed the issue you reviewed. Thank you for your feedback.

rzvxa:
> @rzvxa Hi, rzvxa. I have solved the issue you reviewed. Thank you for your feedback.

Thank you❤️

Issue #2059, opened 2024-01-16
[Linter] Rule configuration severity is not enforced


jbanety:
I encountered an issue with the configuration of the @typescript-eslint/no-explicit-any rule in the `eslintrc.json` file. Despite the rule being set to "error", warnings are still being generated instead of errors.

### Configuration
`eslintrc.json`
```json
{
  "rules": {
    "@typescript-eslint/no-explicit-any": "error"
  }
}
```

### Command line output
```shell
⚠ typescript-eslint(no-explicit-any): Unexpected any. Specify a different type.
     ╭─[src/helpers/htmlTemplate.ts:144:1]
 144 │ 
 145 │ export function renderHtml(template: string, data: Record<string, any> = {}): string {
     ·                                                                   ───
 146 │   // Render template
     ╰────
  help: Use `unknown` instead, this will force you to explicitly, and safely, assert the type is correct.

Finished in 20ms on 34 files with 83 rules using 16 threads.
Found 62 warnings and 0 errors.
```

PR https://github.com/oxc-project/oxc/pull/2031 fixed the `off` problem but not the severity.
Tested on the latest 0.2.1.

jbanety:
I'm trying to do a PR to correct the issue.
It seems that the replace (here : https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/config/mod.rs#L97) makes the associated severity disappear. 
It's always Allow.

jbanety:
Also, the `oxlint -D all` command only outputs warnings.

jbanety:
From what I understand, severity is hardcoded into the rules.
So we can't change the gravity by configuration at the moment?

jbanety:
@Boshen I can help on this one.
I have already looked a little at how we could overcome this problem but it affects the design of Oxlint.

Boshen:
Changing a diagnostic to an error was never implemented. The current design treats everything as a warning.

I would consider it as a bug if vscode raises any errors at the moment.

---

We can also start a topic on changing the current implementation to distinguish warnings and errors when we supply "--deny" / "--warn" from the cli and "error" / "warn" from the config.

alisnic:
The whole error/warning flow is currently broken from my tests. Given a `file.tsx` with the following contents:

```
var Hello = <a target="_blank" href="https://example.com/"></a>;
```

Running with `--deny-warnings` has no effect:

```
$ npx oxlint@latest --deny-warnings

[...]

PR #3335, opened 2024-05-17
refactor(linter): remove `with_rule_name` from the tight loop

Boshen:


Boshen:
Re performance regression: I need this change for https://github.com/oxc-project/oxc/issues/2059, will rethink about speeding this up later.

Issue #3320, opened 2024-05-17
Parser Fuzz Errors


Boshen:
We are getting popular, and I'm anxious about our parser crashing or hanging so people could DOS (denial of service) the services that use oxc.

We have a fuzzer setup in https://github.com/oxc-project/oxc/tree/main/fuzz

It currently reports a "timeout" (it could still complete) on parsing a rewind

```
m< $<{3[   $<{3[  $<{3[ m< m$<{3[ m< mm< $<{3[   $<{3[  $<{3[ m< m$<{3[ m< m$<{3[  $<{3[ m< m$<{3[ m< $<{3[  $<{3[ m< m$<{3[ m< [
```

Apparently it rewinds on every `$<` causing O(N^n) time complexity.

Deciphering the cause and finding the fix is going to be very difficult 😅

PR #3340, opened 2024-05-18
refactor(parser): start porting arrow function parsing from tsc

Boshen:
relates #3320


Boshen:
This will speed up again once everything is ported.

PR #3358, opened 2024-05-20
feat(linter): eslint-plugin-jest/no-duplicate-hooks

eryue0220:

Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-duplicate-hooks.ts)


mysteryven (approved):
Nice!

mysteryven:
Currently, many jest rules need to call `collect_possible_jest_call_node` to glob all nodes first, they will do these things for themselves, I want to change back and avoid calling `run_once`.

And I intend to let jest rules align to other kinds of rules, It will help other people to review codes too.

I will try it in `jest/valid-expect-in-promise`, you are always welcome to work on rules except `jest/valid-expect-in-promise`. :) 

After I finish it, I'll ask for your advice!

PR #3360, opened 2024-05-20
refactor(sourcemap): using binary search to search original position

underfin:
The ast span is not ordering at rolldown, eg the module original ast is `a,b,c`, after mutate could be `b,c,a`. So here revert changes from [here](https://github.com/oxc-project/oxc/pull/2728).

### DIFF
--- a/crates/oxc_codegen/src/sourcemap_builder.rs
+++ b/crates/oxc_codegen/src/sourcemap_builder.rs
@@ -325,7 +325,7 @@ mod test {
 
     #[test]
     fn builder_with_unordered_position() {
-        assert_mapping("ab", &[(0, 0, 0), (2, 0, 2), (1, 0, 1)]);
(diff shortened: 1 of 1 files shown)


Boshen:
Do we need to fix the performance issue 😅

overlookmotel:
> Do we need to fix the performance issue 😅

In my view, ultimately the best solution is to be calculating UTF16/line-column spans in the lexer (as discussed in #959). That should be the most performant way to support sourcemaps' use case.

Unclear when we're going to do that, but we could leave worrying about performance until then if we think it's not urgent to fix the perf right now.

underfin:
Yeah. The performance is little regression at rolldown, we could skip it at now.

Boshen:
:shipit:

Boshen:
Wait ... can we add some test cases?

Boshen (approved):

PR #3363, opened 2024-05-20
feat(linter/eslint): Implement no-useless-concat

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/no-useless-concat)

---

I haven't implemented one condition with the multiline string test case and honestly I don't really understand why it is allowed? And also not sure how I would implement that in oxlint.

Another issue is that maybe the output isn't great, the underlining matches the whole BinaryExpression, for example:
```
+   ╭─[no_useless_concat.tsx:1:1]
+ 1 │ foo + `a` + `b`
+   · ───────────────
```

So maybe instead the diagnostic should get two spans passed, for each Expression::StringLiteral or Expression::TemplateLiteral, that would also allow the help text to show that it can be written as `"ab"`.

But an automatic fixxer would be more helpful I reckon :)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_useless_concat.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_useless_concat.rs
@@ -1,9 +1,9 @@
-use oxc_ast::ast::Expression;
+use oxc_ast::ast::{BinaryExpression, Expression};
 use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
-use oxc_syntax::operator::BinaryOperator;
+use oxc_span::{GetSpan, Span};
+use oxc_syntax::{identifier::is_line_terminator, operator::BinaryOperator};
 
 use crate::{context::LintContext, rule::Rule, AstNode};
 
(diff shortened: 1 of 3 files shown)


Boshen:
> I haven't implemented one condition with the multiline string test case and honestly I don't really understand why it is allowed?

People often write paragraphs with this style (from some random tutorial):

```javascript
let learnCoding = 'How to start learning web development?' +
' - Learn HTML' +
' - Learn CSS' +
' - Learn JavaScript' +
' Use freeCodeCamp to learn all the above and much, much more!'
```

To pass this test, you need to scan for newlines between these two expressions I think.

Boshen:
> Another issue is that maybe the output isn't great, the underlining matches the whole BinaryExpression

We try to minimize the span, so maybe from the start of the first string to the last?

```
+   ╭─[no_useless_concat.tsx:1:1]
+ 1 │ foo + `a` + `b`
+   ·       ────────
```

crates/oxc_linter/src/rules/eslint/no_useless_concat.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// ### Why is this bad?
+    ///
+    /// It’s unnecessary to concatenate two strings together.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// var foo = "a" + "b";
+    /// ```
+    NoUselessConcat,
+    suspicious
```
Boshen, line 33: This rule produces useless code, which fits in our category of `correctness`. I'll test this in real codebases and see if I can change this to `correctness`.

Or if you wanna try out yourself: https://github.com/oxc-project/oxlint-ecosystem-ci

Boshen (approved):

Boshen (changes requested):
@jelly would you like to address the questions?

jelly:
> @jelly would you like to address the questions?

Yes, will do! Likely later this week. I opened the other PR because it was close to the finish line and things finally clicked :)

crates/oxc_linter/src/rules/eslint/no_useless_concat.rs
Boshen: It seems like people love to concatenate strings ...

Boshen (approved):
I ported the eslint source code 😅

PR #3379, opened 2024-05-21
feat(linter/eslint): Implement default_case rule

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/default-case)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/default_case.rs
+++ b/crates/oxc_linter/src/rules/eslint/default_case.rs
@@ -1,0 +1,277 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use regex::Regex;
+use regex::RegexBuilder;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn default_case_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint(default-case): Require default cases in switch statements.")
+        .with_help("Add a default case.")
+        .with_labels([span0.into()])
+}
+
+#[derive(Debug, Default, Clone)]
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/default_case.rs
```diff
@@ -0,0 +1,277 @@
...
+
+#[derive(Debug, Clone)]
+pub struct DefaultCaseConfig {
+    comment_pattern: Regex,
+}
+
+impl Default for DefaultCaseConfig {
+    fn default() -> Self {
+        Self {
+            comment_pattern: RegexBuilder::new(r"^no default$")
```
Boshen, line 27: If the default is a simple string match, can we make this a fast path by defining

`comment_pattern: Option<Regex>`,

and just check for `no default` when it's `None`?

Constructing the regex and running it both introduce a small performance overhead.

Boshen (changes requested):

crates/oxc_linter/src/rules/eslint/default_case.rs
jelly: Ah, that makes sense! I'm still learning Rust :)

I'm not sure if I pushed the most Rust idiomatic approach but the tests are happy.

Boshen:
I pushed some changes:

* demonstrate the `None` case
* removed the allocation from `let default_cases = cases.iter().filter(|v| v.test.is_none()).collect::<Vec<_>>();`
* made the code more compact to demonstrate some nice Rust APIs

Boshen (approved):

jelly:
> I pushed some changes:
> 
>     * demonstrate the `None` case
> 
>     * removed the allocation from `let default_cases = cases.iter().filter(|v| v.test.is_none()).collect::<Vec<_>>();`, the code no longer allocates any heap memory 😁
> 
>     * made the code more compact to demonstrate some nice Rust APIs

Thanks for the explanation!

PR #3381, opened 2024-05-21
improvement(semantic/cfg): rework basic blocks.

rzvxa:
I've replaced the `BasicBlockElement` with an `Instruction` type which would keep both the instruction kind and its associated AstNodeId.
I also removed the register scheme in the control flow in favor of a simpler approach using explicit enums.

https://github.com/oxc-project/oxc/pull/3381#issuecomment-2126622774


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3381?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3381** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3380** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

mysteryven:
Look forward to this series of works! Looks like we can get `AstNode` directly by visiting CFG after this, Will it provide more features?

rzvxa:
> Look forward to this series of works. Looks like we can get `AstNode` directly by visiting CFG after this, Will it provide more features?

This PR will introduce `Instruction` to replace `BasicBlockElement`, They're really similar but since `Instruction` can be any statement, We would be able to check if a given statement in a BasicBlock comes exactly before or after a jump/return; On top of this, we can build better inference and a more ergonomic API that might mimic the path segment stacks more closely.

So there are 3 major points that I'm trying to achieve.

1. Associate a CFG instruction with its origin node if one exists(we have dummy blocks with no node so it is optional)
2. Be more specific on instruction types and express a jump as an instruction(in addition to its edge which already exists). This helps us infer whether a jump happens before or after another statement, It might be unnecessary but it's too soon for me to say for sure. Could be possible to provide this information with a better API instead of adding redundant information?
3. Keep track of the unwinding path and provide a way to follow the error path up the stack.

With all that said, I'm pretty much-adding features and refactoring stuff as I find shortcomings while implementing something. So no cleaver querying and/or super specific CFG visitors are going to come out of this; Unless somebody reports an actual use case for it.

rzvxa:
Reviewing this stack is becoming more cumbersome by the minute, I'll mark some parts of it that I'm sure are correct for review.
Is there a kind soul out there who would put up with these changes and give it a spin?

I'll write a summary for each PR I mark as "ready".

PR #3382, opened 2024-05-22
feat(rulegen): skip the test case when it has a languageOption and only has the ecmaVersion option

### LINKED ISSUE #3366: new-rule for eslint doesn't handle `languageOptions`
I was a bit confused after having run `new-rule` and valid test cases error'ing out as languageOptions are not respected when generating a new rule.

For example:
```
{ code: "globalThis.Math.pow(a, b)", languageOptions: { ecmaVersion: 2019 } },
```

https://github.com/eslint/eslint/blob/main/tests/lib/rules/prefer-exponentiation-operator.js#L73

Dunqing:
close: #3366

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/prefer_exponentiation_operator.rs
+++ b/crates/oxc_linter/src/rules/eslint/prefer_exponentiation_operator.rs
@@ -1,92 +1,35 @@
-use oxc_ast::{ast::Expression, match_member_expression, AstKind};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::Span;
 
-use crate::{
-    ast_util::is_method_call, context::LintContext, globals::GLOBAL_OBJECT_NAMES, rule::Rule,
-    AstNode,
-};
+use crate::{context::LintContext, rule::Rule, AstNode};
 
 #[derive(Debug, Default, Clone)]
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#3382** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (commented):
We skip because the code will end up being exactly the same?

What happens if there two snippets with the same code, and they both have `ecmaVersion`, do we end up removing both?

Dunqing:
> We skip because the code will end up being exactly the same?
> 

We don't need to care about these test cases. These test cases are mainly for checking unsupported syntax.

https://github.com/eslint/eslint/blob/main/tests/lib/rules/prefer-exponentiation-operator.js#L73
> What happens if there two snippets with the same code, and they both have `ecmaVersion`, do we end up removing both?

Yes.

Dunqing:
I'll look later at other rules that use the `ecmaVersion` and what they are doing to make sure we're removing really useless test cases

Dunqing:
I have checked [no-eval](https://github.com/eslint/eslint/blob/389744be255717c507fafc158746e579ac08d77e/tests/lib/rules/no-eval.js) rule. There are some tests that only have the `ecmaVersion` option, but the behavior of their test code is different, so we can't remove them

PR #3394, opened 2024-05-23
feat(transformer): add `TraverseCtx::generate_uid`

overlookmotel:
Add `TraverseCtx::generate_uid` method.

This is modelled on Babel's `scope.generateUid()` method. As discussed in https://github.com/oxc-project/oxc/discussions/3251#discussioncomment-9416826, this is required to fix most of the remaining failing tests in transformer Milestone 1.

I have implemented this to work as closely as possible to Babel, so that it will generate same output as Babel for our tests. However, as mentioned in the code comments, this means it's a pretty expensive function to call. Those code comments suggest 2 ways in which we could make it much more efficient, but we'd need to decide  how we're going to handle divergence from Babel before we can decide which route to go.

I've left it as a `TODO(improve-on-babel)` for now.

PR #3395, opened 2024-05-23
fix(transformer): use UIDs in TS namespace transforms

overlookmotel:
Use the `TraverseCtx::generate_uid` method introduced in #3395 to fix some of the TS namespace test cases.

But... I honestly have no idea what I'm doing here!

I am running up against a combination of 3 different areas where I know very little:

1. I am unfamiliar with `Semantic`.
2. I am unfamiliar with Oxc's conventions for writing transforms.
3. I don't "speak" Typescript!

This PR should not be merged as is. I'm pretty sure it's wrong. I've done it mostly just to test out `generate_uid`.

In particular:

1. Is `SymbolFlags::FunctionScopedVariable` the right flags to use in all cases here?
2. `generate_uid` creates a symbol, and registers a binding for it in the scope tree. So if there are any branches in this logic where `name` doesn't actually get used after `generate_uid` is called, then it's not right.
3. Identifiers which are added to AST should also have corresponding `ReferenceId`s created for them (but I imagine this is also missing in many other places in the transformer).

On point 2: We could split up `generate_uid` into 2 functions. One function would find a valid UID name *but not register a binding for it*. If you want to actually use that name, you'd call a 2nd function to generate the binding. Would that be a better API?

Could someone help me out to progress this please?

(Sorry my lack of knowledge is a bit useless here. I will learn all these things in time, but just trying to be honest about where I'm at right now. I'm sure I could figure it out myself, but it would take me hours, whereas others will probably look at it and know what to do in about 5 mins.)

### DIFF
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -151,6 +151,10 @@ impl ScopeTree {
         self.bindings[scope_id].insert(name, symbol_id);
     }
 
+    pub fn remove_binding(&mut self, scope_id: ScopeId, name: &CompactStr) {
+        self.bindings[scope_id].shift_remove(name);
+    }
+
     pub(crate) fn add_unresolved_reference(
         &mut self,
         scope_id: ScopeId,
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -92,8 +92,8 @@ impl<'a> Transformer<'a> {
 }
 
 impl<'a> Traverse<'a> for Transformer<'a> {
-    fn enter_program(&mut self, program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
-        self.x0_typescript.transform_program(program);
+    fn enter_program(&mut self, program: &mut Program<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.x0_typescript.transform_program(program, ctx);
     }
 
     fn exit_program(&mut self, program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
--- a/crates/oxc_transformer/src/typescript/mod.rs
+++ b/crates/oxc_transformer/src/typescript/mod.rs
@@ -68,8 +68,8 @@ impl<'a> TypeScript<'a> {
 
 // Transforms
 impl<'a> TypeScript<'a> {
-    pub fn transform_program(&self, program: &mut Program<'a>) {
-        self.transform_program_for_namespace(program);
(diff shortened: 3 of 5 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3395?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3395** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3394** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Boshen (approved):
The code was setup this way so you can just replace it, and it's working :grin:

Dunqing (approved):
> In particular:
> 
> 1. Is `SymbolFlags::FunctionScopedVariable` the right flags to use in all cases here?

Correct!

> 2. `generate_uid` creates a symbol, and registers a binding for it in the scope tree. So if there are any branches in this logic where `name` doesn't actually get used after `generate_uid` is called, then it's not right.

Yes. 

> 3. Identifiers which are added to AST should also have corresponding `ReferenceId`s created for them (but I imagine this is also missing in many other places in the transformer).

Yes, we need to bind all unresolved references for the new binding.

> 
> On point 2: We could split up `generate_uid` into 2 functions. One function would find a valid UID name *but not register a binding for it*. If you want to actually use that name, you'd call a 2nd function to generate the binding. Would that be a better API?
> 

I think we can have all three API's at the same time, but not now, not until we really need it!

Haha, I think this PR we can merge.  You did the right thing.

Dunqing:
The remaining test cases are for me to take over, and I'm eager to start using these APIs!

overlookmotel:
I've been through this more thoroughly now. Fixed one problem with a scope binding being registered where no actual binding is created.

But found another problem:

https://github.com/oxc-project/oxc/blob/2fcf8712c840cf0ac0d60e93d6d2c80551ac1cea/crates/oxc_transformer/src/typescript/namespace.rs#L136-L140

I'm not quite sure best way to tackle that. If you guys are happy to merge this in its partially-broken state, with the "TODO" comment, then please do so.

Dunqing:
> I'm not quite sure best way to tackle that. If you guys are happy to merge this in its partially-broken state, with the "TODO" comment, then please do so.

Let's merge this, and I will take a try at this problem

PR #3406, opened 2024-05-25
feat(lint/eslint): implement require-await

todor-a:
Implements https://eslint.org/docs/latest/rules/require-await.

Related to https://github.com/oxc-project/oxc/issues/479.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/require_await.rs
+++ b/crates/oxc_linter/src/rules/eslint/require_await.rs
@@ -1,9 +1,10 @@
 use oxc_ast::{
-    ast::{ArrowFunctionExpression, AwaitExpression, ForOfStatement, PropertyKey},
+    ast::{ArrowFunctionExpression, AwaitExpression, ForOfStatement, Function, PropertyKey},
     AstKind, Visit,
 };
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
+use oxc_semantic::ScopeFlags;
 use oxc_span::Span;
 
 use crate::{context::LintContext, rule::Rule, AstNode};
@@ -99,6 +100,7 @@ impl<'a> Visit<'a> for AwaitFinder {
     }
 
     fn visit_arrow_expression(&mut self, _expr: &ArrowFunctionExpression<'a>) {}
+    fn visit_function(&mut self, _func: &Function<'a>, _flags: Option<ScopeFlags>) {}
 }
 
 #[test]
--- a/crates/oxc_linter/src/rules/eslint/require_await.rs
+++ b/crates/oxc_linter/src/rules/eslint/require_await.rs
@@ -1,0 +1,124 @@
+use oxc_ast::{
+    ast::{Expression, Statement},
+    AstKind,
+};
+use oxc_diagnostics::OxcDiagnostic;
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/require_await.rs
```diff
@@ -0,0 +1,124 @@
...
+                        ctx.diagnostic(require_await_diagnostic(func.span));
+                    }
+                }
+                _ => {}
+            }
+        }
+    }
+}
+
+fn has_await<'a>(statements: &'a [Statement<'a>]) -> bool {
```
mysteryven, line 64: This function may ignore some cases like:

```js
async function foo() {
	{	
		await doSomething()
	}
}
```

What about implementing a visit pattern for it:

```rust
struct AwaitFinder {
    found: bool,
}

impl<'a> Visit<'a> for AwaitFinder {
    fn visit_await_expression(&mut self, expr: &AwaitExpression) {
        if self.found {
            return;
        }
        self.found = true;
    }
    fn visit_for_of_statement(&mut self, stmt: &ForOfStatement) {
        if stmt.r#await {
            self.found = true;
        }
    }
	// you may need this if you visit from function body.
    fn visit_function(&mut self, func: &oxc_ast::ast::Function<'a>, flags: Option<oxc_semantic::ScopeFlags>) {
        return;
    }
	// 	you may need this if you visit from function body.
    fn visit_arrow_expression(&mut self, expr: &oxc_ast::ast::ArrowFunctionExpression<'a>) {
        return;
    }
}


 if func.r#async {
      let mut finder = AwaitFinder { found: false };
      finder.visit_function_body(body);
      if finder.found {
          return;
      }
      ctx.diagnostic(require_await_diagnostic(func.span));
  }
```

crates/oxc_linter/src/snapshots/require_await.snap
```diff
@@ -0,0 +1,63 @@
+---
+source: crates/oxc_linter/src/tester.rs
+expression: require_await
+---
+  ⚠ eslint(require-await): Async function has no 'await' expression.
+   ╭─[require_await.tsx:1:1]
+ 1 │ async function foo() { doSomething() }
```
mysteryven, line 7: Would you like shrink it function name `foo`?

https://oxc-project.github.io/docs/contribute/linter.html#pin-point-the-error-message-to-the-shortest-code-span

crates/oxc_linter/src/rules/eslint/require_await.rs
todor-a: nice! just learned something new. and good catch about that corner case

crates/oxc_linter/src/snapshots/require_await.snap
todor-a: done

mysteryven (approved):
Thank you!

mysteryven:
I added: 

```diff
  fn visit_arrow_expression(&mut self, _expr: &ArrowFunctionExpression<'a>) {}
+ fn visit_function(&mut self, _func: &Function<'a>, _flags: Option<ScopeFlags>) {}
```

It is needed for the following case :)

```js
async function foo() { async function bar() { await doSomething() } } // this is a fail case
```

PR #3407, opened 2024-05-25
fix(linter): memorize visited block id in `neighbors_filtered_by_edge_weight`

### LINKED ISSUE #3396: linter: react/require-render-return hangs
https://github.com/elastic/kibana/blob/main/x-pack/plugins/ml/public/application/timeseriesexplorer/timeseriesexplorer.js

I temporarily moved it to nursery ecdffcf

mysteryven:
closes: #3396

we visit the same node too many times, I picked some result from run require-render-return rule on [timeserieseexploere.js](https://github.com/elastic/kibana/blob/main/x-pack/plugins/ml/public/application/timeseriesexplorer/timeseriesexplorer.js)

```bash
    NodeIndex(64): 368640,
    NodeIndex(67): 737280,
    NodeIndex(70): 2949120,
    NodeIndex(73): 5971968,
    NodeIndex(76): 11943936,
    NodeIndex(43): 184320,
    NodeIndex(71): 2985984,
    NodeIndex(65): 368640,
    NodeIndex(68): 1474560,
    NodeIndex(74): 5971968,
    NodeIndex(77): 23887872,
    NodeIndex(44): 184320,
    NodeIndex(41): 73728,
    NodeIndex(35): 36864,
    NodeIndex(66): 737280,
    NodeIndex(69): 1474560,
    NodeIndex(72): 2985984,
    NodeIndex(75): 11943936,
```


mysteryven:
* **#3407** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -1078,9 +1078,9 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         /* cfg - bb after if statement joins consequent and alternate */
         let after_if_graph_ix = self.cfg.new_basic_block();
 
-        if stmt.alternate.is_some() {
-            self.cfg.put_unreachable();
-        }
+        // if stmt.alternate.is_some() {
```
mysteryven, line 1081: For the code below, the return statement be treated as unreachable, would you like to give me some advice to fix this? @rzvxa 

```js
class Hello extends React.Component {
  render() {
    if (this.a > this.b) {
      foo1()
    } else {
      foo2()
    }

    return 'div'
  }
}
```

rzvxa: I was considering removing the unreachable instructions so if it was a new rule I would've suggested waiting until the end of this weekend for me to finish my PR, But since there is an issue out in the wild let me take a closer look and see what workaround are possible.

The unreachable is redundant here, We should be able to remove it once memorizing the already visited nodes. You should also do the same treatment to the conditional expressions. Let me know if it breaks the `rules_of_hooks` since that's the only other rule doing extensive checks on conditional flows.

rzvxa:
@mysteryven I didn't mean to force push here, I just stacked my PR on top of it, Didn't know it would result in this sorry about it.

rzvxa:
Well now that the cat is out of the bag, I'll go ahead and restack it properly creating an intermediate PR containing the appropriate CFG changes and stack these 2 on top of that. Sorry for the inconvenience!

rzvxa (approved):

crates/oxc_semantic/src/builder.rs
mysteryven: I removed the redundant return instructions in if statement and condition statement, there are a bunch of cfg's snapshot updated, I think it's ok since you will refactor it soon :)

 > Let me know if it breaks the rules_of_hooks since that's the only other rule doing extensive checks on conditional flows.
 
 Review it again, find it safe for this rule in my view.

Boshen:
I screwed up the rebase 🤔

PR #3420, opened 2024-05-26
feat(atom): get &str from Atom<'a> with lifetime of 'a

DonIsaac:
Change `Atom<'a>::as_str(&self) -> &str` to `Atom<'a>::as_str(&self) -> &'a str`. 

This API is more ergonomic for external `oxc` consumers relying on `&str` data collected while traversing an AST.

I also enhanced some nearby doc comments and implemented some `From<...>` traits while I was at it.

### DIFF
--- a/crates/oxc_span/src/atom.rs
+++ b/crates/oxc_span/src/atom.rs
@@ -17,8 +17,8 @@ pub const MAX_INLINE_LEN: usize = 16;
 
 /// An inlinable string for oxc_allocator.
 ///
-/// Use [CompactStr] with [Atom::to_compact_str] or [Atom::into_compact_str] for the
-/// lifetimeless form.
+/// Use [CompactStr] with [Atom::to_compact_str] or [Atom::into_compact_str] for
+/// the lifetimeless form.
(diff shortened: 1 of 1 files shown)


overlookmotel:
@DonIsaac You are absolutely right. The lifetime of the `&str` returned should be `'a`. Atoms are immutable, so the ref is guaranteed to remain valid for lifetime of the arena `'a`.

overlookmotel:
One small point on process:

It's preferable if you're able to split up PRs so each distinct change is its own PR. It makes it easier to review and also to "follow the trail" through git history if someone later on wants to understand why something is as it is. In this case, it'd be preferable to have 1 PR for the lifetime change, and a 2nd PR for the changes to comments.

No need to make changes to this PR - just a request for the future.

Do you have access to Graphite? This makes it pretty easy to stack up a bunch of changes as separate PRs.

overlookmotel (approved):

DonIsaac:
I don't have access to Graphite but I can get that set up.

overlookmotel:
@Boshen Do we need to do anything our end to get Don set up on Graphite?

PR #3431, opened 2024-05-27
fix(transformer): use UIDs for React imports

overlookmotel:
Use UIDs for local var names for React JSX imports e.g. `_jsx` in `import { jsx as _jsx } from "react/jsx-runtime";`.

### DIFF
--- a/crates/oxc_transformer/src/react/jsx/mod.rs
+++ b/crates/oxc_transformer/src/react/jsx/mod.rs
@@ -588,6 +588,8 @@ impl<'a> ReactJsx<'a> {
                 // in `update_fragment` after import is added and correct var name is known.
                 // We have to do like this so that imports are in same order as Babel's output,
                 // in order to pass Babel's tests.
+                // TODO(improve-on-babel): Remove this workaround if output doesn't need to match
+                // Babel's exactly.
                 if self.is_script() {
                     self.get_static_member_expression("_reactJsxRuntime", "Fragment")
                 } else {
--- a/crates/oxc_transformer/src/react/jsx/mod.rs
+++ b/crates/oxc_transformer/src/react/jsx/mod.rs
@@ -73,10 +73,10 @@ impl<'a> ReactJsx<'a> {
             jsx_self: ReactJsxSelf::new(ctx),
             jsx_source: ReactJsxSource::new(ctx),
             jsx_runtime_importer,
+            import_create_element: None,
             import_jsx: None,
             import_jsxs: None,
             import_fragment: None,
-            import_create_element: None,
         }
     }
 
--- a/crates/oxc_transformer/src/react/jsx/mod.rs
+++ b/crates/oxc_transformer/src/react/jsx/mod.rs
@@ -7,6 +7,7 @@ use oxc_ast::{ast::*, AstBuilder};
 use oxc_span::{CompactStr, GetSpan, Span, SPAN};
 use oxc_syntax::{
     identifier::{is_irregular_whitespace, is_line_terminator},
+    symbol::SymbolFlags,
     xml_entities::XML_ENTITIES,
 };
 use oxc_traverse::TraverseCtx;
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#3431** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/react/jsx/mod.rs
```diff
@@ -446,7 +472,11 @@ impl<'a> ReactJsx<'a> {
             }
         }
 
-        self.add_import(e, has_key_after_props_spread, need_jsxs);
+        self.add_import(e, has_key_after_props_spread, need_jsxs, ctx);
+
+        if matches!(e, JSXElementOrFragment::Fragment(_)) {
```
Dunqing, line 477:
```suggestion
        if is_fragment
```

crates/oxc_transformer/src/react/jsx/mod.rs
```diff
@@ -569,6 +598,23 @@ impl<'a> ReactJsx<'a> {
         }
     }
 
+    fn update_fragment(&self, arg: &mut Argument<'a>) {
```
Dunqing, line 601: Is there a way we don't need to update the fragment? I previously tried in https://github.com/oxc-project/oxc/pull/3014/files#diff-5f1a38af256ca90636d6989af9db67db76862a43bda7665e4e005c73cd1828e9

Dunqing (changes requested):

crates/oxc_transformer/src/react/jsx/mod.rs
overlookmotel: Yes, good point. Have made this change.

crates/oxc_transformer/src/react/jsx/mod.rs
overlookmotel: This is not required for correct output, but it's an another annoying example where we need to do weird stuff to produce exactly same output as Babel.

We can generate the import earlier, and then generate correct UID for `_Fragment` to use in `get_fragment`. We then don't need to update the fragment later on.

But the problem is that `transform_jsx` calls itself recursively for its children. So changing whether `_Fragment` import is generated before/after the children are visited changes the order imports are registered in, which makes the output different from Babel:

```js
import { Fragment as _Fragment, jsx as _jsx2 } from 'react/jsx-runtime';
```

vs:

```js
import { jsx as _jsx2, Fragment as _Fragment } from 'react/jsx-runtime';
```

This difference makes many of the JSX tests fail.

I've made a `TODO(improve-on-babel)` comment to point out that we can simplify this if we change our testing strategy. But for now, I think we need to keep it, unfortunately.

Dunqing: > We can generate the import earlier, and then generate correct UID for _Fragment to use in get_fragment. We then don't need to update the fragment later on.

You don't need to generate the import earlier. Instead, you can generate UIDs when you need them, and then use the generated UID in `add_require_jsx_runtime`.

overlookmotel: Yes, it's true that would work. I'm just not sure it makes it any less complicated - we just move the complexity somewhere else. But you obviously feel strongly about it, so I'm happy to do it the way you prefer.

Can we possibly merge this as it is now, and I'll do a follow-on to refactor and get rid of `update_fragment`?

crates/oxc_transformer/src/react/jsx/mod.rs
```diff
@@ -554,13 +584,14 @@ impl<'a> ReactJsx<'a> {
...
+                // in order to pass Babel's tests.
+                // TODO(improve-on-babel): Remove this workaround if output doesn't need to match
+                // Babel's exactly.
                 if self.is_script() {
-                    let object_name = if self.options.development {
-                        "_reactJsxDevRuntime"
-                    } else {
-                        "_reactJsxRuntime"
-                    };
-                    self.get_static_member_expression(object_name, "Fragment")
```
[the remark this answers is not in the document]
overlookmotel: Answered in comment above.

crates/oxc_transformer/src/react/jsx/mod.rs
Dunqing: Of course!

Dunqing (approved):

crates/oxc_transformer/src/react/jsx/mod.rs
overlookmotel: OK great. Thanks. I've opened issue #3464 for the refactoring follow-on.

PR #3436, opened 2024-05-27
feat(linter): eslint-plugin-jest/no-large-snapshot

eryue0220:

Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-large-snapshots.ts)


mysteryven:
I'll take a close look at these PRs after several hours.

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
```diff
@@ -0,0 +1,554 @@
...
+    }
+
+    fn get_line_count(span: Span, ctx: &LintContext) -> usize {
+        let start = span.start as usize;
+        let start = if start == 0 { 1 } else { start };
+        let end = span.end as usize;
+        let start = ctx.source_text()[..start].lines().count();
+        let end = ctx.source_text()[start..end].lines().count();
+
+        end - start
```
mysteryven, line 293: Why we need `end - start`, looks like `ctx.source_text()[start..=end].lines().count()` is the line count

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
```diff
@@ -0,0 +1,554 @@
...
+                Err(_) => snapshot_name == allowed_snapshot,
+            }
+        })
+    }
+
+    fn check_all_paths_are_absolute(&self) {
+        let all_paths_are_absolute =
+            self.allowed_snapshots.keys().all(|p| Path::new(p).is_absolute());
+
+        assert!(all_paths_are_absolute,
```
mysteryven, line 281: Would you like to remove  `assert` here?(I noticed the `throw Error` in jest plugin:) )

mysteryven (changes requested):

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
```diff
@@ -0,0 +1,554 @@
...
+        let start = span.start as usize;
+        let start = if start == 0 { 1 } else { start };
+        let end = span.end as usize;
+        let start = ctx.source_text()[..start].lines().count();
+        let end = ctx.source_text()[start..end].lines().count();
+
+        end - start
+    }
+
+    fn get_span_from_expression(expr: &Expression) -> Span {
```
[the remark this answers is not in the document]
eryue0220: Thanks for your suggestion.

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
eryue0220: Yes, It was my first thought, but the [jest](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-large-snapshots.ts#L32) plugin wast not. 😅 The test will fail if just return the `end`.

crates/oxc_span/src/source_type.rs
```diff
@@ -71,7 +71,8 @@ impl Default for SourceType {
 }
 
 /// Valid file extensions
-pub const VALID_EXTENSIONS: [&str; 8] = ["js", "mjs", "cjs", "jsx", "ts", "mts", "cts", "tsx"];
+pub const VALID_EXTENSIONS: [&str; 9] =
+    ["js", "mjs", "cjs", "jsx", "ts", "mts", "cts", "tsx", "snap"];
```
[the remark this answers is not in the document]
eryue0220: If it was removed, the test would fail as well, or oxc will not check the `.snap` type? or is there some other method or plugin can bypass this check?

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
eryue0220: Done.

crates/oxc_span/src/source_type.rs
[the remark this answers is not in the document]
mysteryven: > Is there some other method or plugin can bypass this check?

Currently, no. We can comment out these failed tests.

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
mysteryven: I still curious bit about this, image a condition, if `..start` has three lines, and `start..end` has only one line, `end-start` is a negative number.

crates/oxc_span/src/source_type.rs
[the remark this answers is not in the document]
eryue0220: Done

crates/oxc_linter/src/rules/jest/no_large_snapshots.rs
eryue0220: Done, I changed the position of `start` and `end` declaration.

mysteryven: Feels like not correct.

```rs
 fn get_line_count(span: Span, ctx: &LintContext) -> usize {
        let start = span.start as usize;
        let end = span.end as usize;
        ctx.source_text()[start..=end].lines().count() - 1
    }
```
Test this works for test case, please check if it right.

mysteryven:
I want to merge this PR, so I updated the `get_line_count ` fn myself, if you find this way it not a correct, you can pull another PR :)

mysteryven (approved):

PR #3455, opened 2024-05-28
feat(linter): implement @typescript-eslint/explicit-function-return-type

kaykdm:

original implementation

- code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/explicit-function-return-type.ts
- test: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/explicit-function-return-type.test.ts
- doc: https://typescript-eslint.io/rules/explicit-function-return-type/

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/explicit_function_return_type.rs
+++ b/crates/oxc_linter/src/rules/typescript/explicit_function_return_type.rs
@@ -57,8 +57,7 @@ declare_oxc_lint!(
     /// ```javascript
     /// ```
     ExplicitFunctionReturnType,
-    restriction, // TODO: change category to `correctness`, `suspicious`, `pedantic`, `perf`, `restriction`, or `style`
-             // See <https://oxc-project.github.io/docs/contribute/linter.html#rule-category> for details
+    restriction,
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/src/rules/typescript/explicit_function_return_type.rs
```diff
@@ -0,0 +1,2134 @@
...
+        // 	function foo(): any {
+        // 	  anyValue = () => () => console.log('aa');
+        // 	}
+        // 	      ",
+        //     Some(
+        //         serde_json::json!([        {          "allowTypedFunctionExpressions": true,        },      ]),
+        //     ),
+        //     None,
+        //     None,
+        // ),
```
kaykdm, line 1642: I don't know why this case will be invalid.
it seems like this should be valid

```
let anyValue: any;
function foo(): any {
  anyValue = () => () => console.log('aa'); // Error
}


function foo2(): any {
  const anyValue: any = () => () => console.log('aa'); // No Error as expected
}
```

PR #3459, opened 2024-05-29
refactor(typescript/namespace): reuse TSModuleBlock's scope id

Dunqing:


Dunqing:
* **#3460** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3460?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3459** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3459?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
merge conflict

crates/oxc_transformer/src/typescript/namespace.rs
```diff
@@ -369,6 +377,7 @@ impl<'a> TypeScript<'a> {
                 params,
                 Some(body),
             );
+            function.scope_id.set(scope_id);
```
overlookmotel, line 380: Could set the correct scope flags here with:

```rs
if let Some(scope_id) = scope_id {
    function.scope_id.set(Some(scope_id));
    let is_strict = ctx.scopes().get_flags(scope_id).is_strict_mode();
    let flags = ScopeFlags::Function.with_strict_mode(is_strict);
    *ctx.scopes_mut().get_flags_mut(scope_id) = flags;
}
```

(I *think* that's correct - or are TS files always strict mode?)

crates/oxc_transformer/src/typescript/namespace.rs
```diff
@@ -150,9 +151,15 @@ impl<'a> TypeScript<'a> {
             &ctx.generate_uid_in_current_scope(real_name, SymbolFlags::FunctionScopedVariable),
         );
 
+        // TODO: This is workaround for now.
```
overlookmotel, line 154: Personally I think this can be a permanent solution rather than temporary.

The advantage of re-using the existing scope is that then all child scopes have the correct parent, and bindings within the block are bound in correct scope already. So we don't have to do any more work updating the scope tree the way we would if we created a new scope.

crates/oxc_transformer/src/typescript/namespace.rs
```diff
@@ -150,9 +151,15 @@ impl<'a> TypeScript<'a> {
...
+        // Reuse TSModuleBlock's scope id in transformed function.
+        // We also need to update the scope flags to FunctionScopedVariable
+        let mut scope_id = None;
+
         let namespace_top_level = if let Some(body) = decl.body {
             match body {
                 TSModuleDeclarationBody::TSModuleBlock(mut block) => {
+                    scope_id = block.scope_id.get();
```
overlookmotel, line 162: Can we get the `scope_id` in the other branch for `TSModuleDeclarationBody::TSModuleDeclaration` as well?

overlookmotel:
@Dunqing I've made some comments, but I don't fully understand TS namespaces. So feel free to ignore them and merge if you want.

Unless I'm understanding incorrectly, there should always be a `TSModuleBlock` we can get `scope_id` from unless the namespace is empty. And if the namespace is empty, we can exit from `handle_nested` early as the statement just gets deleted. Do I have that right?

If you like, merge this and then I'll have a go at fixing the "binding registered in wrong scope" problem I left unsolved previously, and you could check my work? Maybe that's easier than going back and forth with comments?

crates/oxc_transformer/src/typescript/namespace.rs
Dunqing: `TSModuleDeclarationBody::TSModuleDeclaration` has no scope id. You can see https://oxc-project.github.io/oxc/playground/?code=3YCAAIACgICAgICAgICXYPwo06VHxJ7qhgzVoEhzbUbnGCOcafvmPsPEMeUylof9IpeQ%2BId0nsXmmyubkBQxGJ1%2B8VzwpE5XnX5OFaO%2F721rUgVJ3398R6CA

crates/oxc_transformer/src/typescript/namespace.rs
Dunqing: I agreed

crates/oxc_transformer/src/typescript/namespace.rs
Dunqing: Yes, It's correct. we can use `ScopeFlags::Function | ScopeFlags::StrictMode?` directly

Dunqing:
> Unless I'm understanding incorrectly, there should always be a `TSModuleBlock` we can get `scope_id` from unless the namespace is empty. And if the namespace is empty, we can exit from `handle_nested` early as the statement just gets deleted. Do I have that right?

Yes. That's right.

> If you like, merge this and then I'll have a go at fixing the "binding registered in wrong scope" problem I left unsolved previously, and you could check my work? Maybe that's easier than going back and forth with comments?

Sure!

crates/oxc_transformer/src/typescript/namespace.rs
overlookmotel: I meant can we use the `scope_id` from the `TSModuleBlock` within the inner `TSModuleDeclaration`:
namespace AST

overlookmotel: Anyway, I'll follow-up, as discussed below.

PR #3465, opened 2024-05-29
feat(oxc_codegen): preserve annotate comment

IWANABETHATGUY:
1. Copy tests from https://github.com/evanw/esbuild/blob/efa3dd2d8e895f7f9a9bef0d588560bbae7d776e/internal/bundler_tests/bundler_dce_test.go#L3833-L3971
2. Add option to preserve annotate comment like `/* #__NO_SIDE_EFFECTS__ */` and `/* #__PURE__ */`

### DIFF
--- a/crates/oxc_codegen/src/lib.rs
+++ b/crates/oxc_codegen/src/lib.rs
@@ -8,14 +8,18 @@
 //! Code adapted from
 //! * [esbuild](https://github.com/evanw/esbuild/blob/main/internal/js_printer/js_printer.go)
 
+mod annotation_comment;
 mod context;
 mod gen;
 mod gen_ts;
 mod operator;
 mod sourcemap_builder;
 
+use std::ops::Range;
+
 #[allow(clippy::wildcard_imports)]
 use oxc_ast::ast::*;
+use oxc_ast::{Comment, TriviasMap};
 use oxc_span::{Atom, Span};
 use oxc_syntax::{
     identifier::is_identifier_part,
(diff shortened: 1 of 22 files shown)


crates/oxc_codegen/src/lib.rs
```diff
@@ -72,6 +80,16 @@ pub struct Codegen<const MINIFY: bool> {
     indentation: u8,
 
     sourcemap_builder: Option<SourcemapBuilder>,
+    comment_gen_related: Option<CommentGenRelated>,
```
Dunqing, line 83: Can we put it in `CodegenOptions`? Also, it doesn't look like this feature is used anywhere. Do we need this?

IWANABETHATGUY: IMO, `Options` should store configuration like `boolean` type, but `CommentGenRelated` store trivias, and `MoveMap`, I don't think it is a good idea to put it under `CodegenOptions`

IWANABETHATGUY: more details you could refer: https://github.com/rolldown/rolldown/issues/1269

Dunqing: Ok, I got it. You should add some tests for it.

IWANABETHATGUY: https://github.com/oxc-project/oxc/pull/3465/files#diff-8b2f51c395bbba32ab8432d172a0c5a85ab39159eac6ec709d29d0b77b2625d3R207

Issue #3392, opened 2024-05-23
Align visitation order in `Traverse` with `Visit` and `VisitMut`


overlookmotel:
`Traverse` is codegen-ed from the AST node type definitions. The codegen generates a walker which visits children of a node in the order which they're defined in the AST types. e.g. for:

https://github.com/oxc-project/oxc/blob/8a1db6738c761b87a2581275e728445c1bb63e5a/crates/oxc_ast/src/ast/js.rs#L52-L60

Visitors are called in order matching the above:

* `enter_program`
* `enter_directive` + `exit_directive` (for each directive)
* `enter_hashbang` + `exit_hashbang`
* `enter_statements` + `exit_statements`
* `exit_program`

On the other hand, `Visit` and `VisitMut` were written by hand. It's likely that for some types the visitation order is different from the types' field order. So it will be different from `Traverse`'s visitation order.

This mismatch is going to cause problems at some point as contributors who are used to `Visit`/`VisitMut` will assume that `Traverse` behaves the same, and be confused when it doesn't, or unwittingly introduce bugs due to incorrect assumptions.

I think best way to tackle this is:

* Write a codegen for `Visit` and `VisitMut` which does visitation in AST node field order (like `Traverse` does).
* Diff current `Visit` vs codegen-ed `Visit`.
* Where they don't match, change the field order in AST types, and run codegen again.
* Keep doing that until the codegen-ed `Visit` and current hand-written `Visit` match exactly.
* Throw away the old hand-written `Visit` and `VisitMut` and use the codegen from here on.

Once we switch to codegen, any changes to AST types will automatically update `Visit` + `VisitMut` to match.

In addition: We know there are also some discrepancies with `Visit`/`VisitMut` not visiting some child nodes. Likely in some cases this is a mistake (inevitable when writing this much code by hand), but in other cases it's intentional (like not visiting some TS nodes where it's pointless to do so).

The process above will reveal all such discrepancies, and then we can decide which are mistakes and which are intentional, and we can then adapt `Visit`, `VisitMut` and `Traverse` to make them all consistent and all reflecting what we decide is the "correct" visitation policy.

Boshen:
@rzvxa FYI

rzvxa:
@overlookmotel I think this can be closed after merging #4001 and #4007

rrandom:
Can anybody tell me what is the difference between these two traits?

[...]

PR #3488, opened 2024-05-31
refactor(ast): move scope from `TSModuleBlock` to `TSModuleDeclaration`

### LINKED ISSUE #3471: Multiple scopes for `namespace foo.bar`
If I understand correctly, this:

```ts
namespace foo.bar {
  export const qux = 123;
}
```

is just shorthand for:

```ts
namespace foo {
  export namespace bar {
    export const qux = 123;
  }
}
```

Therefore, `namespace foo.bar` essentially represents 2 scopes - 1 for `foo`, and a 2nd for `bar`.

I propose to encode this in the AST by either:

1. Add `scope_id` field to the `TSModuleDeclarationBody::TSModuleDeclaration` variant
or
2. Remove `scope_id` field from `TSModuleBlock` and add it to `TSModuleDeclaration` instead.

Creating multiple scopes seems semantically correct to me, and would make transforming it to JS much simpler.

Just opening this issue because I'm pretty rubbish with TS, and want to check my reasoning before going ahead with it. Am I misunderstanding anything?

cc @Boshen @Dunqing

overlookmotel:
Closes #3471.

Remove scope from `TSModuleBlock` and add scope to `TSModuleDeclaration` instead.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -868,8 +868,8 @@ pub enum TSTypePredicateName<'a> {
     This(TSThisType),
 }
 
-#[visited_node]
-#[derive(Debug, Hash)]
+#[visited_node(scope(ScopeFlags::TsModuleBlock), enter_scope_before(body))]
+#[derive(Debug)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(tag = "type", rename_all = "camelCase"))]
 pub struct TSModuleDeclaration<'a> {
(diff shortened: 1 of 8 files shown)


overlookmotel:
* **#3488** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -931,27 +953,14 @@ pub enum TSModuleDeclarationBody<'a> {
...
 #[cfg_attr(feature = "serialize", serde(tag = "type", rename_all = "camelCase"))]
 pub struct TSModuleBlock<'a> {
     #[cfg_attr(feature = "serialize", serde(flatten))]
     pub span: Span,
     pub body: Vec<'a, Statement<'a>>,
-    pub scope_id: Cell<Option<ScopeId>>,
-}
-
-impl<'a> TSModuleBlock<'a> {
-    pub fn new(span: Span, body: Vec<'a, Statement<'a>>) -> Self {
```
Dunqing, line 946: I think we can keep this method

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -868,8 +868,8 @@ pub enum TSTypePredicateName<'a> {
     This(TSThisType),
 }
 
-#[visited_node]
-#[derive(Debug, Hash)]
+#[visited_node(scope(ScopeFlags::TsModuleBlock), enter_scope_before(body))]
```
Dunqing, line 871: I'm wondering if `scope(ScopeFlags::TsModuleBlock), enter_scope_before(body)` is still necessary, as it seems that the `oxc_traverse` hasn't changed with it!

Dunqing (changes requested):

crates/oxc_ast/src/ast/ts.rs
overlookmotel: `enter_scope_before(...)` is still used by the `oxc_traverse` codegen. You're right that the `scope(...)` part is ignored at present, but I thought best leave them there for now because it'll be needed for #3392.

crates/oxc_ast/src/ast/ts.rs
Dunqing: I see. Actually, `PrivateIdentifier` and `IdentifierName` both have new methods, but they don't have `scope_id`. It doesn't matter. I just to say. We can add a `new` method for all AST nodes

Dunqing (approved):

crates/oxc_ast/src/ast/ts.rs
overlookmotel: Ah right, so there is no pattern! OK, I propose to leave this as is for now. As you say, we can add `new` methods for all AST node types later on if we think that'd be of benefit.

PR #3489, opened 2024-05-31
fix(transformer): use correct scope for TS namespaces

overlookmotel:
Re-use scope ID from `TSModuleDeclaration` for the scope of the function which replaces it in transformed output.

This should mean the scope tree is left in correct state after the transform.

NB: Still incomplete - we also need to transfer the binding ID from `X` in `namespace X {}` to `let X` in output.

### DIFF
--- a/crates/oxc_transformer/src/typescript/namespace.rs
+++ b/crates/oxc_transformer/src/typescript/namespace.rs
@@ -133,36 +133,37 @@ impl<'a> TypeScript<'a> {
         parent_export: Option<Expression<'a>>,
         ctx: &mut TraverseCtx,
     ) -> Option<Statement<'a>> {
+        // Skip empty declaration e.g. `namespace x;`
+        let body = decl.body?;
+
         let mut names: FxHashSet<Atom<'a>> = FxHashSet::default();
 
-        let real_name = decl.id.name();
+        let TSModuleDeclarationName::Identifier(IdentifierName { name: real_name, .. }) = decl.id
+        else {
+            return None;
+        };
 
-        // TODO: This binding is created in wrong scope.
-        // Needs to be created in scope of function which `transform_namespace` creates below.
-        let name = self.ctx.ast.new_atom(
-            &ctx.generate_uid_in_current_scope(real_name, SymbolFlags::FunctionScopedVariable),
-        );
+        // Reuse `TSModuleDeclaration`'s scope in transformed function
+        let scope_id = decl.scope_id.get().unwrap();
+        let name = self.ctx.ast.new_atom(&ctx.generate_uid(
+            &real_name,
+            scope_id,
+            SymbolFlags::FunctionScopedVariable,
+        ));
 
-        // Reuse TSModuleBlock's scope id in transformed function.
-        let scope_id = decl.scope_id.get();
-
-        let namespace_top_level = if let Some(body) = decl.body {
-            match body {
-                TSModuleDeclarationBody::TSModuleBlock(block) => block.unbox().body,
-                // We handle `namespace X.Y {}` as if it was
-                //   namespace X {
-                //     export namespace Y {}
-                //   }
-                TSModuleDeclarationBody::TSModuleDeclaration(declaration) => {
-                    let declaration = Declaration::TSModuleDeclaration(declaration);
-                    let export_named_decl =
-                        self.ctx.ast.plain_export_named_declaration_declaration(SPAN, declaration);
-                    let stmt = Statement::ExportNamedDeclaration(export_named_decl);
-                    self.ctx.ast.new_vec_single(stmt)
-                }
+        let namespace_top_level = match body {
+            TSModuleDeclarationBody::TSModuleBlock(block) => block.unbox().body,
+            // We handle `namespace X.Y {}` as if it was
+            //   namespace X {
+            //     export namespace Y {}
+            //   }
+            TSModuleDeclarationBody::TSModuleDeclaration(declaration) => {
+                let declaration = Declaration::TSModuleDeclaration(declaration);
+                let export_named_decl =
+                    self.ctx.ast.plain_export_named_declaration_declaration(SPAN, declaration);
+                let stmt = Statement::ExportNamedDeclaration(export_named_decl);
+                self.ctx.ast.new_vec_single(stmt)
             }
-        } else {
-            self.ctx.ast.new_vec()
         };
 
         let mut new_stmts = self.ctx.ast.new_vec();
@@ -172,6 +173,7 @@ impl<'a> TypeScript<'a> {
                 Statement::TSModuleDeclaration(decl) => {
                     if decl.id.is_string_literal() {
                         self.ctx.error(ambient_module_nested(decl.span));
+                        continue;
                     }
 
                     let module_name = decl.id.name().clone();
@@ -228,6 +230,7 @@ impl<'a> TypeScript<'a> {
                             Declaration::TSModuleDeclaration(module_decl) => {
                                 if module_decl.id.is_string_literal() {
                                     self.ctx.error(ambient_module_nested(module_decl.span));
+                                    continue;
                                 }
 
                                 let module_name = module_decl.id.name().clone();
@@ -268,7 +271,7 @@ impl<'a> TypeScript<'a> {
             return None;
         }
 
-        Some(self.transform_namespace(&name, real_name, new_stmts, parent_export, scope_id, ctx))
+        Some(self.transform_namespace(name, real_name, new_stmts, parent_export, scope_id, ctx))
     }
 
     // `namespace Foo { }` -> `let Foo; (function (_Foo) { })(Foo || (Foo = {}));`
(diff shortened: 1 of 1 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3489?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3489** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3489?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3488** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3488?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
Damn! Something in conformance is failing. I'll fix.

crates/oxc_transformer/src/typescript/namespace.rs
```diff
@@ -133,36 +133,40 @@ impl<'a> TypeScript<'a> {
...
-        // Needs to be created in scope of function which `transform_namespace` creates below.
-        let name = self.ctx.ast.new_atom(
-            &ctx.generate_uid_in_current_scope(real_name, SymbolFlags::FunctionScopedVariable),
-        );
+        let real_name = match decl.id {
+            TSModuleDeclarationName::Identifier(ident) => ident.name,
+            TSModuleDeclarationName::StringLiteral(lit) => {
+                // TODO: Need to convert to valid identifier
+                lit.value
+            }
```
Dunqing, line 146: We don't care about `StringLiteral`, that only works for `module "x"`. Here we only transform `namespace`. You can check the TSModuleDeclaration's kind field. https://oxc-project.github.io/oxc/playground/?code=3YCAAIChgICAgICAgIC3mMl9fWP2DOEtS3aewAVF3MlN9R8zpdZMUnii0U5sQf%2Fxyrjgn39L%2BQCA

overlookmotel:
@boshen @Dunqing Can I ask for help here?

Conformance is failing for various TS compiler test cases due to panic on this line:

https://github.com/oxc-project/oxc/blob/1f1449f4fbb725b51b1ab7b20d4449805fb1b23b/crates/oxc_transformer/src/typescript/namespace.rs#L150

i.e. semantic has not set a `scope_id` for this `TSModuleDeclaration`.

Here's one that fails:

`typescript/tests/cases/compiler/declareDottedModuleName.ts`

```rs
// @declaration: true
module M {
    module P.Q { } // This shouldnt be emitted
}

module M {
    export module R.S { }  //This should be emitted
}

module T.U { // This needs to be emitted
}
```

The common factor among the failing tests appears to be the `@declaration: true` directive. When I remove that line, the tests pass. Does that directive disable semantic / change it's behavior somehow?

crates/oxc_transformer/src/typescript/namespace.rs
Dunqing: But `module "d"` is not a valid typescript code. [Check this](https://www.typescriptlang.org/play/?target=1&lib=true&ts=5.5.0-dev.20240531#code/LYewJgrgNgpgBAIgM4LgbwL4CgthgYygEMAneUSWRFdDIA)

overlookmotel:
I assume the answer to my question is this:

https://github.com/oxc-project/oxc/blob/ff7e8c7448b3c781166b4634b06491b37fdc8535/crates/oxc_semantic/src/builder.rs#L141-L151

Semantic does not visit TS definition files.

Next questions:
1. Is that correct behavior?
2. What should transformed output be for a TS definition file? Pass it through unchanged? Output an empty file (the way all `declare` statements are stripped out of a `.ts` file)?

crates/oxc_transformer/src/typescript/namespace.rs
overlookmotel: Babel test case `typescript/module-namespace/head/input.ts` includes this illegal syntax:

```ts
namespace N {}
namespace M.N.O {}
module M {}
module "m" {}
```

Should we be catching this and rejecting it in semantic?

overlookmotel: I've skipped that Babel test case for transformer conformance, so we can assume `module "d"` is not possible here.

overlookmotel: Have opened #3498 for catching this illegal syntax in semantic.

overlookmotel:
Thanks for coming back. But I'm seeing something different from Babel.

Here's the TS conformance test case `compiler/defaultDeclarationEmitShadowedNamedCorrectly.ts` run through Babel: [Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAgDMN5odEBbaAtACwwGcA6NTsjIwAOWZGlAAqUIk6hmoKsiyNQAIm7B2XVQG4y0AB4ixoDDjTRkVRLGigAKhxwBzTgB4ACgBoHAPlAA3mSgIaBCBB56oYQE9noAvvpGouJUxDiwmLhyiADW0J4-AML2or4AFAYEAaA40ADuoOUAlAQloqAAZIFCSkKcEfGg8a0OTq6FoO3I_kHRyNBoxMg4tcTw8NKyiDgAnglkScbiZhZWNnYefbJBiUcpoDDW6-IIMrIAsrtFyiJ15oFgqFelh-hFrgdDMdakxoJwhBdQABlZSLcaA6J0cRfH7CXD0cQAXlAACYdKBKEi2IhIFhGjA3qgsjgfJwsKBOLtGAAjLCbLigXDwXagYicRDcuigABuGEQhDYdk40HgVAAtIJjkCQlCHrBcJxXgt0HZiYw8tBysxuE9EC9mgcgA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.24.6&externalPlugins=%40babel%2Fplugin-transform-typescript%407.24.6&assumptions=%7B%7D)

Or is it just that Babel's not understanding from the `@declaration: true` directive that it's a TS definition file?

In [TS playground](https://www.typescriptlang.org/play/?filetype=d.ts#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAgDMN5odEBbaAtACwwGcA6NTsjIwAOWZGlAAqUIk6hmoKsiyNQAIm7AMOIcTSqA3GWgAPEWNBa00ZFUSxooACoccAc04AeAAoAaJwD5QAG8yUDDQIQIvQ3DCAkdDAF8jU1FxKmIcWExcOUQAa2hvPwBhR1F-AApjAiDQHGgAd1BKgEoCMtFQADJgoSUhTijE0ET2pxd3YtBO5ECQ2ORoNGJkHHrieHhpWUQcAE8kshSzcUtrW3tQLwHZEOSTtNAYW03xBBlZAFl9kuURBo4cQLcL9LCDKK3I4mU71JjQThCOwOADKymWk2CoXCdHEPz+wlw9HEAF5QAAmfSgSgotiISBYZowD6oHI4PycLCgTj7RgAIyw2y4oFw8H2oGInEQfLooAAbhhEIQ2A5ONB4FQALSCU7YsIwp6wXCcd5LdAOMmMArQSrMbgvRBvVpHIA) it produces errors!

Dunqing:
> Or is it just that Babel's not understanding from the `@declaration: true` directive that it's a TS definition file?
> 
> In [TS playground](https://www.typescriptlang.org/play/?filetype=d.ts#code/PTAEAEBMFMGMBsCGAnRAXAlgewHYC5Q1kBXaAKBAgDMN5odEBbaAtACwwGcA6NTsjIwAOWZGlAAqUIk6hmoKsiyNQAIm7AMOIcTSqA3GWgAPEWNBa00ZFUSxooACoccAc04AeAAoAaJwD5QAG8yUDDQIQIvQ3DCAkdDAF8jU1FxKmIcWExcOUQAa2hvPwBhR1F-AApjAiDQHGgAd1BKgEoCMtFQADJgoSUhTijE0ET2pxd3YtBO5ECQ2ORoNGJkHHrieHhpWUQcAE8kshSzcUtrW3tQLwHZEOSTtNAYW03xBBlZAFl9kuURBo4cQLcL9LCDKK3I4mU71JjQThCOwOADKymWk2CoXCdHEPz+wlw9HEAF5QAAmfSgSgotiISBYZowD6oHI4PycLCgTj7RgAIyw2y4oFw8H2oGInEQfLooAAbhhEIQ2A5ONB4FQALSCU7YsIwp6wXCcd5LdAOMmMArQSrMbgvRBvVpHIA) it produces errors!

Yes, the directive only works in typescript. But I don't see any mention of documentation about `@declaration` in the typescript document. BTW our typescript coverage test has already handled this directive https://github.com/oxc-project/oxc/blob/6f71541db0da3da7da0f79e33aeb462789782a24/tasks/coverage/src/typescript.rs#L238

overlookmotel:
Conformance finally passes! Assuming that #3500 is correct, I think this is ready for a last review.

NB: I had to disable 3 conformance tests which contain illegal `module "m" {}` syntax (without `declare`).

overlookmotel:
I changed my mind. Now just skipping ambient modules, so was able to re-enable the failing tests containing this invalid syntax.

Dunqing (approved):

PR #3495, opened 2024-06-01
chore: bump oxc-browserslist to 0.17.0

Boshen:


Boshen:
Shaved off another 20s from the conformance build.

PR #3514, opened 2024-06-03
fix(codegen): wrong escape string

### LINKED ISSUE #3492: Unicode escape issue in `oxc_codegen::Codegen`
Related to: https://github.com/kuma-ui/kuma-ui/issues/415

When using `oxc_codegen::Codegen` in WebAssembly context, the generated code contains Unicode escape sequences for non-ASCII characters. This results in the rendered text being displayed as escaped sequences rather than the original characters.

### Steps to Reproduce:

1. Use the oxc_codegen::Codegen to generate code from an AST.
2. Ensure the input source code contains non-ASCII characters.
3. Observe that the generated code includes Unicode escape sequences instead of the original characters.

**Example**: https://github.com/kuma-ui/kuma-ui/blob/main/packages/wasm/src/lib.rs

In this example, my library uses `Codegen` after manipulating the AST from the original source code string. When non-ASCII characters are present in the source text, they are escaped in the generated code.

If there is already a workaround or if I missed something in the documentation, I apologize in advance.

Dunqing:
close: #3492

### DIFF
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -1339,7 +1339,9 @@ fn print_unquoted_str<const MINIFY: bool>(s: &str, quote: char, p: &mut Codegen<
             '\u{a0}' => {
                 p.print_str(b"\\xA0");
             }
-            _ => p.print_str(c.escape_default().to_string().as_bytes()),
+            _ => {
+                p.print_str(c.encode_utf8([0; 4].as_mut()).as_bytes());
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#3514** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
Looks like escape is slow! Also, I use `encode_utf8` instead of `to_string`

PR #3531, opened 2024-06-04
feat(linter): add fixer for no-single-promise-in-promise-methods

DonIsaac:
Fixes look like this:

```rs
let fix = vec![
    ("Promise.all([null]).then()", "Promise.resolve(null).then()", None),
    ("let x = await Promise.all([foo()])", "let x = await foo()", None),
    ("let x = await (Promise.all([foo()]))", "let x = await (foo())", None),
];
```


crates/oxc_linter/src/rules/unicorn/no_single_promise_in_promise_methods.rs
```diff
@@ -57,9 +57,38 @@ impl Rule for NoSinglePromiseInPromiseMethods {
...
-        if !is_promise_method_with_single_element_array(call_expr) {
+        // if !is_method_call(
+        //     call_expr,
+        //     Some(&["Promise"]),
+        //     Some(&["all", "any", "race"]),
+        //     Some(1),
+        //     Some(1),
+        // ) {
+        //     return false;
+        // }
```
Boshen, line 68: Are these commented out code intended?

crates/oxc_linter/src/rules/unicorn/no_single_promise_in_promise_methods.rs
```diff
@@ -5,9 +5,9 @@ use oxc_ast::{
 use oxc_diagnostics::OxcDiagnostic;
 
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
+use oxc_span::{GetSpan as _, Span};
```
Boshen, line 8:
```suggestion
use oxc_span::{GetSpan, Span};
```

Boshen (changes requested):

Boshen (approved):

PR #3580, opened 2024-06-07
refactor(transformer): replace `&’a Trivias` with `Rc<Trivias>`

Dunqing:
`Transformer` needs to borrow `Trivias`. https://github.com/rolldown/rolldown/pull/1364/files/8be1cc80520e1a466cd7b2d3fb8b637db9e529a9#r1630711060

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -108,10 +108,9 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     #[must_use]
-    pub fn with_trivias(mut self, trivias: Trivias) -> Self {
-        let trivias = Rc::new(trivias);
-        self.trivias = Rc::clone(&trivias);
-        self.jsdoc = JSDocBuilder::new(self.source_text, trivias);
+    pub fn with_trivias(mut self, trivias: Rc<Trivias>) -> Self {
+        self.trivias = trivias;
+        self.jsdoc = JSDocBuilder::new(self.source_text, Rc::clone(&self.trivias));
         self
     }
 
--- a/crates/oxc_wasm/src/lib.rs
+++ b/crates/oxc_wasm/src/lib.rs
@@ -172,6 +172,8 @@ impl Oxc {
             .parse();
 
         self.comments = self.map_comments(&ret.trivias);
+        let trivias = Rc::new(ret.trivias);
+
         self.save_diagnostics(ret.errors.into_iter().map(Error::from).collect::<Vec<_>>());
 
         self.ir = format!("{:#?}", ret.program.body).into();
@@ -179,7 +181,7 @@ impl Oxc {
         let program = allocator.alloc(ret.program);
 
         let semantic_ret = SemanticBuilder::new(source_text, source_type)
-            .with_trivias(ret.trivias)
+            .with_trivias(Rc::clone(&trivias))
             .with_check_syntax_error(true)
             .build(program);
 
@@ -229,15 +231,9 @@ impl Oxc {
 
         if run_options.transform() {
             let options = TransformOptions::default();
-            let result = Transformer::new(
-                &allocator,
-                &path,
-                source_type,
-                source_text,
-                semantic.trivias(),
-                options,
-            )
-            .build(program);
+            let result =
+                Transformer::new(&allocator, &path, source_type, source_text, trivias, options)
+                    .build(program);
             if let Err(errs) = result {
                 self.save_diagnostics(errs);
             }
--- a/crates/oxc_language_server/src/linter.rs
+++ b/crates/oxc_language_server/src/linter.rs
@@ -287,7 +287,7 @@ impl IsolatedLintHandler {
 
             let program = allocator.alloc(ret.program);
             let semantic_ret = SemanticBuilder::new(javascript_source_text, source_type)
-                .with_trivias(ret.trivias)
+                .with_trivias(Rc::new(ret.trivias))
                 .with_check_syntax_error(true)
                 .build(program);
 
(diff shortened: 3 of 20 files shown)


Dunqing:
* **#3580** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
No we should never clone this heavy data structure, we need to `Rc` or `Arc` it before hand.

overlookmotel:
I also ran into problems with trying to share `Trivias` in another context. It's problematic that when you create a reference `&Trivias`, its lifetime doesn't live as long as `Program<'a>`.

I'm not sure `Rc`/`Arc` will solve this problem as it'll still not live as long as we need it to.

In my view the best solution would be to store `Trivias` in the allocator, and hand out a `TriviasRef<'a>` with same lifetime as `Program<'a>`. Then it'd be easy to send both around the place together.

Ideally the actual `BTreeMap` data would be stored in an arena, so it can be cheap to drop, but I can't find a `BTreeMap` implementation which supports arena allocators. So for now we could just add it as a field on `Allocator`. `Allocator` owns `Trivias`, and it'd get dropped when allocator is dropped.

Would that solve the problem that this PR is targeting?

Dunqing:
> Would that solve the problem that this PR is targeting?

I think would solve this problem. I will give it a try!

overlookmotel:
OK great. We should also add a function to `Allocator` that resets both the bumpalo arena and `Trivias`:

```rs
impl Allocator {
  fn reset(&mut self) {
    self.bump.reset();
    self.trivias.reset();
  }
}

impl Trivias {
  fn reset(&mut self) {
    self.comments.clear();
    self.irregular_whitespaces.clear();
  }
}
```

Otherwise `Trivias` will keep growing each time the allocator is reset.

Dunqing:
I tried to give `Trivias` the same lifetime as `Program`. However, this approach didn't solve this problem. The `Transformer` always needs to borrow the `Trivias`. So we cannot move out `ast` at the same time.
image

In my opinion, the only viable solution is to make `Trivias` cloneable. By using `Rc`, we can address this problem. I found the same solution [here](https://github.com/oxc-project/oxc/blob/532510d7143b1c6010cf078cd158a8af17bc982d/crates/oxc_semantic/src/builder.rs#L110-L115)

I don't know much about lifetime, so maybe there's a better solution I don't know about.

overlookmotel:
Oh dear! I'm not good with lifetimes either. I don't think any of us are!

If you can solve it with `Rc`, probably best to do that for now to unblock us. I don't think it's the optimal solution, but we can come back to it later if we want to try to remove `Rc`. Reasoning about the lifetimes in transformer is a brain-bender.

crates/oxc_wasm/src/lib.rs
```diff
@@ -170,16 +170,17 @@ impl Oxc {
         let ret = Parser::new(&allocator, source_text, source_type)
             .allow_return_outside_function(parser_options.allow_return_outside_function)
             .parse();
+        let trivias = Rc::new(ret.trivias);
 
-        self.comments = self.map_comments(&ret.trivias);
+        self.comments = self.map_comments(&trivias);
```
overlookmotel, line 175: I think you can leave this as `&ret.trivias` (`map_comments` only needs to temporarily borrow `trivias`). Then wrap `trivias` in an `Rc` further down:

```rs
let semantic_ret = SemanticBuilder::new(source_text, source_type)
    .with_trivias(Rc::new(&ret.trivias))
```

That'd remove one `Rc::clone`.

crates/oxc_semantic/src/builder.rs
```diff
@@ -108,8 +108,8 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     #[must_use]
-    pub fn with_trivias(mut self, trivias: Trivias) -> Self {
-        self.trivias = Rc::new(trivias);
+    pub fn with_trivias(mut self, trivias: Rc<Trivias>) -> Self {
+        self.trivias = trivias;
         self.jsdoc = JSDocBuilder::new(self.source_text, &self.trivias);
```
overlookmotel, line 113: Please see #3586.

crates/oxc_wasm/src/lib.rs
Dunqing: Yes!

crates/oxc_wasm/src/lib.rs
```diff
@@ -234,7 +236,7 @@ impl Oxc {
                 &path,
                 source_type,
                 source_text,
-                semantic.trivias(),
+                Rc::clone(&trivias),
```
overlookmotel, line 239: I think this can just be `trivias`. No need to clone the `Rc`, as it's not used again after this.

overlookmotel:
I'd suggest making the change above, but it's not so important.

Besides that, this change is motivated by a need in Rolldown, and I'm not familiar enough with Rolldown to say "LGTM", as I don't understand the context. But if it solves the problem, everything else looks good to me.

crates/oxc_wasm/src/lib.rs
Dunqing: Thanks for catching that

Dunqing:
> Besides that, this change is motivated by a need in Rolldown, and I'm not familiar enough with Rolldown to say "LGTM", as I don't understand the context. But if it solves the problem, everything else looks good to me.

Thank you for reviewing this! I tested this change in Rolldown and it worked.

overlookmotel:
> Thank you for reviewing this! I tested this change in Rolldown and it worked.

OK great. Merge it then I guess!

PR #3585, opened 2024-06-08
feat(linter): eslint-plugin-jest/max-nested-describes

eryue0220:

Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/max-nested-describe.ts)

### DIFF
--- a/crates/oxc_linter/src/rules/jest/max_nested_describe.rs
+++ b/crates/oxc_linter/src/rules/jest/max_nested_describe.rs
@@ -152,18 +152,19 @@ impl MaxNestedDescribe {
         let AstKind::CallExpression(call_expr) = node.kind() else {
             return;
         };
-
-        if is_type_of_jest_fn_call(
+        let is_describe_call = is_type_of_jest_fn_call(
             call_expr,
             possible_jest_node,
             ctx,
             &[JestFnKind::General(JestGeneralFnKind::Describe)],
-        ) && !describes_hooks_depth.contains(&scope_id)
+        );
+
+        if is_describe_call && !describes_hooks_depth.contains(&scope_id)
         {
             describes_hooks_depth.push(scope_id);
         }
 
-        if describes_hooks_depth.len() > self.max {
+        if is_describe_call && describes_hooks_depth.len() > self.max {
             ctx.diagnostic(exceeded_max_depth(
                 describes_hooks_depth.len(),
                 self.max,
--- a/crates/oxc_linter/src/rules/jest/max_nested_describe.rs
+++ b/crates/oxc_linter/src/rules/jest/max_nested_describe.rs
@@ -159,8 +159,7 @@ impl MaxNestedDescribe {
             &[JestFnKind::General(JestGeneralFnKind::Describe)],
         );
 
-        if is_describe_call && !describes_hooks_depth.contains(&scope_id)
-        {
+        if is_describe_call && !describes_hooks_depth.contains(&scope_id) {
             describes_hooks_depth.push(scope_id);
         }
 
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/jest/max_nested_describe.rs
```diff
@@ -0,0 +1,397 @@
...
+        collect_possible_jest_call_node, is_type_of_jest_fn_call, JestFnKind, JestGeneralFnKind,
+        PossibleJestNode,
+    },
+};
+
+fn exceeded_max_depth(current: usize, max: usize, span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn(
+        "eslint-plugin-jest(max-nested-describe): Enforces a maximum depth to nested describe calls.",
+    )
+    .with_help(format!("Too many nested describe calls ({current:?}) - maximum allowed is {max:?}"))
```
DonIsaac, line 20: Prefer `Display` syntax over `Debug`, e.g. `({current})` over `({current:?})` if possible

crates/oxc_linter/src/rules/jest/max_nested_describe.rs
```diff
@@ -0,0 +1,397 @@
...
+            .and_then(serde_json::Number::as_u64)
+            .map_or(5, |v| usize::try_from(v).unwrap_or(5));
+
+        Self { max }
+    }
+
+    fn run_once(&self, ctx: &LintContext) {
+        let mut describes_hooks_depth: Vec<ScopeId> = vec![];
+        let mut possibles_jest_nodes = collect_possible_jest_call_node(ctx);
+        possibles_jest_nodes.sort_by_key(|n| n.node.id());
```
DonIsaac, line 135: What's the benefit of sorting here? Is it worth the extra runtime overhead it adds?

crates/oxc_linter/src/snapshots/max_nested_describe.snap
```diff
@@ -0,0 +1,353 @@
...
+ 10 │ │                                           });
+ 11 │ ╰─▶                                     });
+ 12 │                                     });
+    ╰────
+  help: Too many nested describe calls (6) - maximum allowed is 5
+
+  ⚠ eslint-plugin-jest(max-nested-describe): Enforces a maximum depth to nested describe calls.
+    ╭─[max_nested_describe.tsx:8:41]
+  7 │                                         describe('over limit', function () {
+  8 │ ╭─▶                                         it('should get something', () => {
```
DonIsaac, line 21: Are we intentionally adding diagnostics to `it` and `expect`s as well as `describe`s?

eryue0220: In this rule, no. For `expect`, it has another related rule called `max_expect`.

crates/oxc_linter/src/rules/jest/max_nested_describe.rs
eryue0220: It's more easy to handle the iteration, or it will need to check its node.id and parent recursively.

crates/oxc_linter/src/rules/jest/max_nested_describe.rs
eryue0220: Done~

crates/oxc_linter/src/snapshots/max_nested_describe.snap
DonIsaac: In that case, you may want to update your reporting logic to skip diagnostics on `it`, `test`, and `expect`

mysteryven (approved):

mysteryven:
I merge this because all comment be resolved and look good to me.

PR #3586, opened 2024-06-08
refactor(semantic): pass `Rc` by value

overlookmotel:
Same as #3550. 

`Rc<T>` is already a reference, so instead of passing an `&Rc<T>` to a function and then `Rc::clone()` it in the function, it's better to clone it first and pass `Rc<T>` to the function.

`Rc<T>` and `&Rc<T>` are both 8 bytes, so it introduces no additional overhead to the function call, and reduces indirection.

This is a very small optimization. Am only submitting these changes for purpose of code tidying - making the patterns around `Rc` consistent and optimal throughout the codebase.

We should probably look if we can remove some of these `Rc`s entirely and replace them with plain `&` refs. I suspect `Rc` is not actually required in most places and we're only using it to avoid dealing with lifetimes, but it's sub-optimal as `Rc::clone` has a cost, whereas copying a `&` ref has none.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -100,7 +100,7 @@ impl<'a> SemanticBuilder<'a> {
             symbols: SymbolTable::default(),
             module_record: Arc::new(ModuleRecord::default()),
             label_builder: LabelBuilder::default(),
-            jsdoc: JSDocBuilder::new(source_text, &trivias),
+            jsdoc: JSDocBuilder::new(source_text, trivias),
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#3587** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3587?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3586** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3586?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #3589, opened 2024-06-08
up(linter): add rule fixer

DonIsaac:
Adds a new `RuleFixer` struct that gets passed to `ctx.diagnostic_with_fix`.

This PR is half proposal and half implementation. I'd love thoughts and feedback on its design (including whether or not it adds enough benefit to make it worth merging this PR).

### DIFF
--- a/crates/oxc_ast/src/span.rs
+++ b/crates/oxc_ast/src/span.rs
@@ -341,7 +341,6 @@ impl<'a> GetSpan for Argument<'a> {
     }
 }
 
-// impl_get_span!(ArrayExpression);
 impl<'a> GetSpan for ArrayExpression<'a> {
     fn span(&self) -> Span {
         self.span
--- a/crates/oxc_ast/src/span.rs
+++ b/crates/oxc_ast/src/span.rs
@@ -2,6 +2,33 @@ use oxc_span::{GetSpan, Span};
 
 use crate::ast::*;
 
+macro_rules! impl_get_span {
+    ($Ty:ident) => {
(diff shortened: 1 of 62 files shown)


crates/oxc_span/src/span.rs
Boshen, line 1: Regarding all these changes, we should have a separate PR copying all the methods from https://docs.rs/text-size/1.1.1/text_size/struct.TextRange.html

crates/oxc_ast/src/span.rs
```diff
@@ -2,6 +2,33 @@ use oxc_span::{GetSpan, Span};
 
 use crate::ast::*;
 
+macro_rules! impl_get_span {
```
Boshen, line 5: Let's have as less macro as possible in these crates, a simple copy paste is totally fine here.

I have a rule https://oxc.rs/docs/contribute/rules.html#development-policy

> Avoid heavy macros or generics or any Rust techniques that would penalize compilation speed.

It gets heavy when we repeat these.

Boshen (commented):
The direction is good, we should just port most of eslint 😅

crates/oxc_span/src/span.rs
DonIsaac: https://github.com/oxc-project/oxc/pull/3592

crates/oxc_ast/src/span.rs
DonIsaac: Removed

Boshen (approved):

Boshen:
Definitely a better API!

PR #3591, opened 2024-06-09
test(linter): add linter to coverage task

DonIsaac:
Makes `cargo coverage` run the linter. It can be run in isolation with `cargo coverage linter`.

The linter is run with fixes and all rules enabled (except nursery rules) in an identical fashion to `oxlint -W all --fix`. The fixed code is then re-parsed and checked for errors. Tests will fail if parsing fails or if semantic analysis produces more errors the second time around than when the original source code is analyzed.

I've already found several problems with fixes, including 2 panics. One of them was fixed in #3590, and the other is fixed in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,6 +1378,7 @@ dependencies = [
  "oxc_ast",
  "oxc_codegen",
  "oxc_diagnostics",
+ "oxc_linter",
  "oxc_minifier",
  "oxc_parser",
  "oxc_prettier",
--- a/crates/oxc_linter/src/fixer.rs
+++ b/crates/oxc_linter/src/fixer.rs
@@ -25,7 +25,7 @@ pub struct FixResult<'a> {
     pub messages: Vec<Message<'a>>,
 }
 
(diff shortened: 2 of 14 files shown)


Boshen:
Should we add this to https://github.com/oxc-project/oxlint-ecosystem-ci/actions/workflows/ecosystem-ci.yml instead 🤔

DonIsaac:
It appears to add 50s to the conformance CI job. If we think this is too high I can move it to the ecosystem CI job, though it may not catch some of the weirder fixes (e.g. labels)

Boshen:
> It appears to add 50s to the conformance CI job. If we think this is too high I can move it to the ecosystem CI job, though it may not catch some of the weirder fixes (e.g. labels)

Let's try and move this to ecosystem CI with fixes applied to running `oxlint`, and then check whether the fixed file is good, maybe by running oxlint over it again?

Boshen:
Moving to https://github.com/oxc-project/oxlint-ecosystem-ci/issues/11

PR #3602, opened 2024-06-10
refactor(ast): make `AstBuilder` `Copy`

overlookmotel:
`AstBuilder` only contains a shared ref `&Allocator`. So make it `Copy` and pass around `AstBuilder<'a>` instead of `&AstBuilder<'a>` - 1 less level of indirection.

Hopefully this will have little effect on benchmarks as all `AstBuilder`'s methods are marked `#[inline]`, but this change will minimize impact if compiler decides not to inline for some reason.

### DIFF
--- a/crates/oxc_ast/src/ast_builder.rs
+++ b/crates/oxc_ast/src/ast_builder.rs
@@ -20,6 +20,7 @@ use oxc_syntax::{
 use crate::ast::*;
 
 /// AST builder for creating AST nodes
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#3602** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #3605, opened 2024-06-10
feat(linter): typescript-eslint no-useless-empty-export

keita-hino:
partof: #2180

docs: https://typescript-eslint.io/rules/no-useless-empty-export/
code: https://github.com/typescript-eslint/typescript-eslint/blob/584db29ec44ce4e9cb71afac35d48994889168e6/packages/eslint-plugin/src/rules/no-useless-empty-export.ts
test: https://github.com/typescript-eslint/typescript-eslint/blob/ac397f18176a9defd8c189b5b6b4e5d0b7582210/packages/eslint-plugin/tests/rules/no-useless-empty-export.test.ts

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/no_useless_empty_export.rs
+++ b/crates/oxc_linter/src/rules/typescript/no_useless_empty_export.rs
@@ -44,6 +44,11 @@ declare_oxc_lint!(
 
 impl Rule for NoUselessEmptyExport {
     fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let module_record = ctx.semantic().module_record();
+        if module_record.not_esm || !module_record.export_default_duplicated.is_empty() {
+            return;
+        }
+
         match node.kind() {
             AstKind::Program(program) => {
                 check_node(&program.body, ctx);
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/typescript/no_useless_empty_export.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    NoUselessEmptyExport,
+    correctness
+);
+
+impl Rule for NoUselessEmptyExport {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::Program(program) => {
+                check_node(&program.body, ctx);
```
Boshen, line 49: It seems like we can add a fast path here.

Can you take a look at `ctx.semantic().module_record()` and see if there are any useful information that you can leverage?

keita-hino: I have modified it to skip processing when it's `not_esm` or `export_default_duplicated`. Is this approach correct?
https://github.com/oxc-project/oxc/pull/3605/commits/790949fe2713d5d6bb7a25b0c90ad552c9879199

Boshen:
I realized that we can just check if module record has empty exports when we see `export {}`.

This TypeScript case happens to rare, better to ignore it 😅 

```
import _ = require('_');
export {};
```

keita-hino:
@Boshen Thank you for the fix🙏

PR #3622, opened 2024-06-11
feat(coverage): add `--debug` option for finding crash scene

### LINKED ISSUE #3497: Catch panics in conformance tests
While working on #3489, I have been hitting conformance test cases where the transformer panics. It's difficult to debug because, while stack trace tells you where the panic arises, it doesn't tell you on which test case.

Could we make the conformance runner catch panics and report them as test fails, rather than crashing the process?

I believe we would need to:

1. Wrap running cases in `std::panic::catch_unwind`.
2. Run coverage with profile including `panic = "unwind"` instead of `panic = "abort"`.

If `panic = "unwind"` would slow down coverage a lot, could we keep `panic = "abort"` as the default but add a separate command `just coverage-with-unwind` which you can use when you are trying to debug panics?

Boshen:
The `--debug` options turns off rayon and prints the paths so we can
find the crashing file by looking at the last printed out file.

closes #3497

### DIFF
--- a/tasks/coverage/README.md
+++ b/tasks/coverage/README.md
@@ -1,13 +1,11 @@
 # Coverage
 
-The parser is tested against [test262], [babel] and TypeScript for conformance.
+Tools are tested against [test262], [babel] and [TypeScript] for conformance.
 
-Note: tests against regexp are disabled for now.
-
-Clone the test files beforehand:
+Clone the test repositories beforehand:
 
 ```bash
-git submodule update --init
+just submodules
 ```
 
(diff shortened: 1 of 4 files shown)


Boshen:
* **#3622** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3622?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Thanks for doing this. Will help a lot!

PR #3623, opened 2024-06-11
fix(transformer): do not rename accessible identifier references

### LINKED ISSUE #3620: transformer: incorrect enum value
babel enum/mix-references/input.ts

In:
```js
var x = 10;

enum Foo {
    a = 10,
    b = a,
    c = b + x,
}
```

Current output:

```
var x = 10;
var Foo = function(Foo) {
	Foo[Foo['a'] = 10] = 'a';
	Foo[Foo['b'] = 10] = 'b';
	Foo[Foo['c'] = Foo.b + Foo.x] = 'c';
	return Foo;
}(Foo || {});
```

This produces an incorrect transform, `Foo.x` is undefined producing `Foo.c = NaN`

We should transform it correctly even we don't compute enum values.

Dunqing:
close: #3620

In `Babel`, the expected output is:

```ts
var x = 10;
var Foo = function(Foo) {
       Foo[Foo['a'] = 10] = 'a';
       Foo[Foo['b'] = 10] = 'b';
       Foo[Foo['c'] = 30] = 'c';
       return Foo;
}(Foo || {});
```

IMO, `Foo.b + x` is enough, because `x` is not a const variable. The output same as with `typescript`

### DIFF
--- a/crates/oxc_transformer/src/typescript/enum.rs
+++ b/crates/oxc_transformer/src/typescript/enum.rs
@@ -535,6 +535,16 @@ impl<'a, 'b> VisitMut<'a> for IdentifierReferenceRename<'a, 'b> {
                 None
             }
             Expression::Identifier(ident) => {
+                // If the identifier is binding in current/parent scopes,
+                // and it is not a member of the enum,
+                // we don't need to rename it.
+                // `var c = 1; enum A { a = c }` -> `var c = 1; enum A { a = c }
+                if !self.previous_enum_members.contains_key(&ident.name)
+                    && self.ctx.scopes().has_binding(self.ctx.current_scope_id(), &ident.name)
+                {
+                    return;
+                }
+
                 // TODO: shadowed case, e.g. let ident = 1; ident; // ident is not an enum
                 // enum_name.identifier
                 let ident_reference = IdentifierReference::new(SPAN, self.enum_name.clone());
--- a/tasks/transform_conformance/oxc.snap.md
+++ b/tasks/transform_conformance/oxc.snap.md
@@ -1,6 +1,6 @@
 commit: 4bd1b2c2
 
-Passed: 3/3
+Passed: 4/4
 
 # All Passed:
 * babel-plugin-transform-typescript
--- a/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/enum-member-reference/input.ts
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/enum-member-reference/input.ts
@@ -1,0 +1,7 @@
+var x = 10;
+
+enum Foo {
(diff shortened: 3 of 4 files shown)


Dunqing:
* **#3623** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
The main thing is to fix the incorrectness. But ideally, I don't think `Foo.b + x` is quite enough. Going one step further to `10 + x` would be fine though (i.e. resolve `Foo.b` to `10`).

We could go a step further and deduce that because `x` is never written to, it's effectively a const, and then const fold it down to `20`. But that feels like a job for the minifier. But I doubt minifier could deduce that `Foo.b` is a const too, hence why I think that step should happen in the transform (ideally).

PS: I don't know where `30` comes from in comment above. If input is what's given in #3620, the value should be `20`, no?

Dunqing:
> The main thing is to fix the incorrectness. But ideally, I don't think `Foo.b + x` is quite enough. Going one step further to `10 + x` would be fine though (i.e. resolve `Foo.b` to `10`).

A similar [example](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=G4QwTgBAHhC8EAoCUcB8EDeBfA3AKFEgDEB7EuCAMwFcA7AYwBcBLE2hUklDPASE4DaggOQhhAXQoBGAAyT4o4fn5khq4QCMJ0uRU1K-gkfW3xOAOg0QA1NHkRhJ5WACmjamFoRO-LBzIQAD6BmFhIOEA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=script&lineWrap=true&presets=react%2Ctypescript&prettier=false&targets=&version=7.24.7&externalPlugins=%40babel%2Fplugin-transform-typescript%407.24.6&assumptions=%7B%7D)

> 
> We could go a step further and deduce that because `x` is never written to, it's effectively a const, and then const fold it down to `20`. But that feels like a job for the minifier. But I doubt minifier could deduce that `Foo.b` is a const too, hence why I think that step should happen in the transform (ideally).
> 

Yes, `Foo.b` can be converted to `20`, but the current implementation can't do it. The way it is currently implemented, renaming and resolving constant variables do not happen at the same time. Once we find that we can't convert an initializer to a constant value, we stop resolving it, and wait until all the initializers have resolved and rename it.

> PS: I don't know where `30` comes from in comment above. If input is what's given in #3620, the value should be `20`, no?

Ah, Yes, it is `20`, I made a mistake

Issue #3528, opened 2024-06-04
`JSXIdentifierReference` type


overlookmotel:
`Foo` in `<Foo />` is equivalent to an `IdentifierReference`. It should have a `reference_id`, and semantic should resolve its binding.

`JSXIdentifier` is also used in other positions e.g. as a `JSXAttributeName`, so would probably need a separate type `JSXIdentifierReference` for this specific use.

Could also differentiate between `<div />` and `<Div />` in parser, with `JSXIdentifierReference` only used for the latter.

Boshen:
I confirmed with tsc, which uses a plain `Identifier`:

```typescript
export type JsxTagNameExpression =
    | Identifier
    | ThisExpression
    | JsxTagNamePropertyAccess
    | JsxNamespacedName
    ;
```

which means we can remove `JsxIdentifier` and use `IdentifierReference` and `IdentifierName` directly.

```rust
pub enum JSXElementName<'a> {
    /// `<Apple />`
    IdentifierReference(Box<'a, IdentifierReference<'a>>),
    /// `<div />
    IdentifierName(Box<'a, IdentifierName<'a>>),
    /// `<Apple:Orange />`
    NamespacedName(Box<'a, JSXNamespacedName<'a>>),
    /// `<Apple.Orange />`
    MemberExpression(Box<'a, JSXMemberExpression<'a>>),
}
```

but, in order to make it compatible with estree, we need to add at least one new AST node `JSXIdentifierReference `
```

:-/

**Comment by overlookmotel** [comment]
Ditto for:

```rs
pub enum JSXMemberExpressionObject<'a> {
    /// `<Foo.bar />`
    IdentifierReference(Box<'a, IdentifierReference<'a>>),
    /// `<foo.bar />`
    IdentifierName(Box<'a, IdentifierName<'a>>),
    MemberExpression(Box<'a, JSXMemberExpression<'a>>),
}
```

I think we can use `IdentifierReference` and `IdentifierName` directly and get ESTree compatibility with a custom `Serialize` impl.

overlookmotel:
Looks like `JSXElementName` also needs a `ThisExpression` variant. `<this />` is valid syntax.

overlookmotel:
Dunqing did bulk of this work in #5223.

Remaining bits to cover: #5352, #5353, #5354.

Closing in favour of those individual issues.

PR #3627, opened 2024-06-11
fix(transformer): arrow function transform alter `</this>`

overlookmotel:
Arrow function transform transforms `this` in `<this>` (JSX opening element), but was missing out transforming `</this>` (JSX closing element).

### DIFF
--- a/crates/oxc_transformer/src/es2015/arrow_functions.rs
+++ b/crates/oxc_transformer/src/es2015/arrow_functions.rs
@@ -119,34 +119,21 @@ impl<'a> ArrowFunctions<'a> {
     }
 
     /// Change <this></this> to <_this></_this>, and mark it as found
-    pub fn transform_jsx_opening_element(&mut self, elem: &mut JSXOpeningElement<'a>) {
+    pub fn transform_jsx_element_name(&mut self, name: &mut JSXElementName<'a>) {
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#3627** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
Currently, the arrow function plugin has almost no test cases, can you add a test for this change?

overlookmotel:
@Dunqing This has been merged this before I got a chance to add the test you requested.

To be honest, I don't think it's worthwhile adding tests at this stage for 2 reasons:

First, as noted in the code comments added in #3634, there are various problems with the transform at present, and some are not resolvable until TS transforms are modified to update the scope tree. So in my view, the best approach to testing is to later on solidify the implementation, and then enable the Babel conformance tests for arrow function transforms.

Secondly, if we do #3528, then we can remove the special handling of `<this>` in arrow function transform, as `this` as a JSX element name will just be a standard `ThisExpression` (https://github.com/oxc-project/oxc/issues/3528#issuecomment-2150843938). So tests that we add now for this will become redundant.

But if you disagree, that's fine - please just let me know, and I'll add one.

PR #3645, opened 2024-06-12
feat(linter): add `parserOptions.emitDecoratorMetadata` and `parserOptions.experimentalDecorators` for config file

mysteryven:
These two options are needed in [`consistent-type-import`](https://typescript-eslint.io/rules/consistent-type-imports/#caveat-decorators--experimentaldecorators-true--emitdecoratormetadata-true).

At first, I think `parserOptions` is not useful for us, and want to hard-code `experimentalDecorators` and `emitDecoratorMetadata` to `false` directly. But I changed the thought after reading:

1. https://typescript-eslint.io/blog/changes-to-consistent-type-imports-with-decorators
2. https://github.com/typescript-eslint/typescript-eslint/pull/8335 

We don't have type-aware linting for now, and require these two options to decide how we perform: if we have both `experimentalDecorators: true` and `emitDecoratorMetadata: true`, then it will not report any errors within any files that contain decorators.

https://github.com/typescript-eslint/typescript-eslint/blob/e408b93e48e5199d83a8d99d1e27126d2dd8bc8f/packages/eslint-plugin/src/rules/consistent-type-imports.ts#L303-L336

### DIFF
--- a/crates/oxc_linter/src/config/mod.rs
+++ b/crates/oxc_linter/src/config/mod.rs
@@ -1,5 +1,6 @@
 mod env;
 mod globals;
+mod parser_options;
 mod rules;
 mod settings;
 
@@ -13,7 +14,7 @@ use serde::Deserialize;
 use crate::{rules::RuleEnum, AllowWarnDeny, RuleWithSeverity};
 
 pub use self::{
-    env::OxlintEnv, globals::OxlintGlobals, rules::OxlintRules,
+    env::OxlintEnv, globals::OxlintGlobals, parser_options::OxlintParseOptions, rules::OxlintRules,
     settings::jsdoc::JSDocPluginSettings, settings::OxlintSettings,
 };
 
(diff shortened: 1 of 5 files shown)


mysteryven:
* **#3645** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3645?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac:
@mysteryven it looks like we've gotten consistent-type-imports shipped, do we still need this PR?

mysteryven:
> @mysteryven it looks like we've gotten consistent-type-imports shipped, do we still need this PR?

Unfortunately, this issue still exists: https://typescript-eslint.io/blog/changes-to-consistent-type-imports-with-decorators/#consistent-type-imports-caused-runtime-breakage, there are [some cases](https://github.com/oxc-project/oxc/blob/f7da22da185e1384b8257d3756fc7c1abc34676c/crates/oxc_linter/src/rules/typescript/consistent_type_imports.rs#L1341-L1351) are commented out temporarily.

Add `parserOptions.*` only for this case is not really worthy IMO, it's looks strange. It's OK for me to close this.

Maybe the [`FrameworkFlags`](https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/frameworks.rs#L7) and read something in [`tsconfig.json`](https://www.typescriptlang.org/tsconfig/#experimentalDecorators) is better and more ergonomic?

mysteryven:
Need a better way.

PR #3653, opened 2024-06-13
build(website): use typescript

DonIsaac:


Boshen:
It was intentional to not have a typescript dependency.

The playground is being rebuilt here https://github.com/oxc-project/playground so I'll merge this so we can move things over easily.

Boshen:
@DonIsaac Can you DM me or raise an issue before raising huge PRs like this, it's somewhat of maintenance burden 😅

PR #3742, opened 2024-06-18
refactor(semantic): rename `cfg` macro to `control_flow`.

rzvxa:
As requested here https://github.com/oxc-project/oxc/pull/3737#discussion_r1644713121

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -31,7 +31,7 @@ use crate::{
     Semantic,
 };
 
-macro_rules! cfg {
+macro_rules! control_flow {
     (|$self:ident, $cfg:tt| $body:expr) => {
         if let Some(ref mut $cfg) = $self.cfg {
             $body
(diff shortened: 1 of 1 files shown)


rzvxa:
* **#3742** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3742?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
Too much variance coming from a small benchmark file, I'll look into it tomorrow.

overlookmotel (approved):

overlookmotel:
> Too much variance coming from a small benchmark file, I'll look into it tomorrow.

My guess is the variance is coming from `SourcemapBuilder::generate_line_offset_tables` which builds a large `Vec`. There's a large `Vec` reallocation (`RawVec::grow_one` followed by memcpy) at start of the flamegraph. If by chance the global allocator says that `Vec` can grow in place, then the memcpy would disappear, which would be a large jump in perf. But whether that happens or not will be non-deterministic depending on state of global allocator (or even the OS).

That's just a guess, though...

PR #3746, opened 2024-06-18
chore(benchmark): enable `cfg` for linter benchmarks.

rzvxa:
The benchmark result on this PR - if we filter out the unstable benchmarks - can also be used to show the total impact of CFG enable rules.

PR #3745, opened 2024-06-18
perf(linter): only generate `cfg` when we have a rule which depends on it.

rzvxa:

PR #3747, opened 2024-06-18
feat(linter): add `CFGLintContext` for rules with `#[use_cfg]`.

rzvxa:
related to this https://github.com/oxc-project/oxc/pull/3744#issuecomment-2176735840

Now only rules declared with the `#[use_cfg]` attribute would get to use `ctx.cfg` method otherwise they have to check for `cfx.semantic().cfg()` to be `Some`.

If this gets merged we should be able to unwrap the control flow unchecked; Since our current linter runner is getting its context from outside we should check for this not being `None` before iterating and panicking early on.

### DIFF
--- a/crates/oxc_linter/src/context.rs
+++ b/crates/oxc_linter/src/context.rs
@@ -88,11 +88,6 @@ impl<'a> LintContext<'a> {
 pub struct CFGLintContext<'a>(LintContextData<'a>);
 
 impl<'a> CFGLintContext<'a> {
-    #[allow(dead_code)]
-    pub(super) fn into_data(self) -> LintContextData<'a> {
-        self.0
-    }
-
     /// # Panics
     /// If rule doesn't have `#[use_cfg]` in it's declaration it would panic.
     pub fn cfg(&self) -> &ControlFlowGraph {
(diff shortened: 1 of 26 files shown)


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3747?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3747** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3745** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3745?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3744** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3744?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3746** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3746?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rzvxa:
@overlookmotel Is there any way to transmute `LintContext` to `CFGLintContext`? If we want to do this are we forced to make them `#[repr(C)]`? Some parts of my current implementation are a bit wanky because of this.

crates/oxc_linter/src/context.rs
```diff
@@ -34,8 +73,68 @@ pub struct LintContext<'a> {
     severity: Severity,
 }
 
+#[derive(Clone)]
```
overlookmotel, line 76: I *think* the transmute between `&LintContext`/`&CFGLintContext` and `&LintContextData` is sound as you've made both structs `#[repr(transparent)]`.

But if someone changed the type defs without understanding this need, it'd lead to UB.

To make this unlikely to happen, I think it needs a comment here at the location of the type definitions explaining what is going on here, the motivation for it, and saying at minimum "It is **essential** that `LintContext` and `LintContextData` have identical memory layouts so that `From` impls below are sound.".

Boshen:
Closing the stack as too complicated.

PR #3761, opened 2024-06-19
refactor(linter): `LintContext` can now only be constructed with a cfg enabled semantic.

rzvxa:
It has the same spirit as #3747 but with a much simpler approach. I've used the fact that @Boshen mentioned about linter always using CFG so now we assert `semantic.cfg().is_some()` in the `LintContext::new` because of this assertion we can have a `LintContext::cfg` that unwraps unchecked.
Eliminates unnecessary checks in our hot paths.

It has the best of both worlds, No complicated typing yet we still get the CFG as a non-optional value without extra ASM or branching.

### DIFF
--- a/crates/oxc_language_server/src/linter.rs
+++ b/crates/oxc_language_server/src/linter.rs
@@ -287,6 +287,7 @@ impl IsolatedLintHandler {
 
             let program = allocator.alloc(ret.program);
             let semantic_ret = SemanticBuilder::new(javascript_source_text, source_type)
+                .with_cfg(true)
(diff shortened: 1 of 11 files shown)


rzvxa:
* **#3761** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3761?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
I'm not sure, this graphite merge queue is confusing sometimes (probably due to interacting with github).

PR #3766, opened 2024-06-19
test(linter): enable `no-fallthrough` test with `disable-next-line`.

rzvxa:
I thought we didn't have this, But now that I've read the `LintContext` I have realized it wasn't true.


rzvxa:
@Boshen I'll merge it since it has no sensitive code, Just adding a passing test.

PR #3776, opened 2024-06-19
ci: custom global allocator for benchmarks

overlookmotel:
Use a custom global allocator for benchmarks. Will hopefully fix the wild variance in many benchmarks that we've been seeing, which I believe is caused by unpredictable behavior in the system allocator.

### DIFF
--- a/tasks/benchmark/src/lib.rs
+++ b/tasks/benchmark/src/lib.rs
@@ -1,1 +1,39 @@
+use std::alloc::{GlobalAlloc, Layout, System};
+
 pub use criterion::*;
+
+#[global_allocator]
+static GLOBAL: NeverGrowInPlaceAllocator = NeverGrowInPlaceAllocator;
+
+/// Global allocator for use in benchmarks.
+///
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#3776** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3776?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
I'm not entirely sure if this is going to work. The CodSpeed results *seem* to suggest it's doing what's intended - large extra costs for memcpy on some of the most flaky benchmarks. But some new calls to `malloc_consolidate` have also appeared in places, and I'm not sure what that is.

Those oddities suggest this change will not succeed in making allocation costs entirely deterministic, but hopefully it'll be a step in the right direction.

Hopefully after the initial jumps in bench results on this PR, we'll see them settle down and be more consistent in future. Hopefully...

My suggestion is that we merge this and see how it goes over next week or so. If we stop seeing wild variation in benchmarks, it's probably all good, and we can try re-enabling the semantic benchmarks. If not, we can revert this change and try something else.

PR #3794, opened 2024-06-20
feat(traverse): disable syntax check and disable build module record

Boshen:
These can be skipped because :
 * semantic errors are not passed to the caller
 * module record is not used


Boshen:
What ... who's using what?

Boshen:
Found the issue https://github.com/oxc-project/oxc/issues/3796

PR #3798, opened 2024-06-20
feat(isolated-declarations): infer undefined as any type

Dunqing:
In the TypeScript Playground, it is inferred as the `undefined` type, but in `transpileDeclaration`, it is inferred as the `any` type. We should make sure that the output is the same as transpileDeclaration's. See: https://github.com/microsoft/TypeScript/pull/58912#issuecomment-2177327619

### DIFF
--- a/crates/oxc_isolated_declarations/src/inferrer.rs
+++ b/crates/oxc_isolated_declarations/src/inferrer.rs
@@ -17,7 +17,7 @@ impl<'a> IsolatedDeclarations<'a> {
     pub fn infer_type_from_expression(&self, expr: &Expression<'a>) -> Option<TSType<'a>> {
         match expr {
             Expression::BooleanLiteral(_) => Some(self.ast.ts_boolean_keyword(SPAN)),
-            Expression::NullLiteral(_) => Some(self.ast.ts_null_keyword(SPAN)),
+            Expression::NullLiteral(lit) => Some(self.ast.ts_any_keyword(lit.span)),
             Expression::NumericLiteral(_) | Expression::BigintLiteral(_) => {
                 Some(self.ast.ts_number_keyword(SPAN))
             }
@@ -30,7 +30,7 @@ impl<'a> IsolatedDeclarations<'a> {
                 }
             }
             Expression::Identifier(ident) => match ident.name.as_str() {
-                "undefined" => Some(self.ast.ts_undefined_keyword(SPAN)),
+                "undefined" => Some(self.ast.ts_any_keyword(ident.span)),
                 _ => None,
             },
             Expression::FunctionExpression(func) => {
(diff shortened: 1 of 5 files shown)


Dunqing:
* **#3798** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3798?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jakebailey:
I'm like 99% certain this is a bug in `transpileDeclaration`.

Boshen:
> I'm like 99% certain this is a bug in `transpileDeclaration`.

Thank you for noticing us.

We are conforming against https://github.com/microsoft/TypeScript/tree/main/tests/cases/transpile so we will just port bugs if they occur (because we don't really understand the underlying logic).

We will correct this once upstream is fixed.

jakebailey:
Ah, you're testing this with strict off; `undefined` doesn't "exist" without strictNullChecks, hence what you're seeing here.

Boshen:
@Dunqing Let's gather a few cases where the playground and the `transpileDeclaration` API differ. I think it's mostly when errors occur, one or the other will reach to the type checker to get the types, but our implementation don't have type inference obviously.

Dunqing:
> Ah, you're testing this with strict off; `undefined` doesn't "exist" without strictNullChecks, hence what you're seeing here. With this test on TS:

The API document doesn't mention `strictNullChecks: false`. Is this a bug or just forget to add it?
https://github.com/Dunqing/TypeScript/blob/d8086f14b6b97c0df34a0cc2f56d4b5926a0c299/src/services/transpile.ts#L68-L82

jakebailey:
That comment only shows the compiler options which differ from the defaults. strictNullChecks is disabled by default (unfortunately). You should pull compiler options from the resolved tsconfig that would be used to compile the code normally via tsc/tsserver/etc.

Dunqing:
> That comment only shows the compiler options which differ from the defaults ("If no options are provided - it will use a set of default compiler options.").
> 
> strictNullChecks is disabled by default (unfortunately). You should pull compiler options from the resolved tsconfig that would be used to compile the code normally via tsc/tsserver/etc.

Thank you for explaining this. Unfortunately, we're not going to read tsconfig.json yet.

Dunqing:
I think for now we just need to be consistent with the default configuration of `transpileDeclaration`

jakebailey:
I think you'd be better off assuming strict=true, if anything.

jakebailey:
If you write "undefined" to the output, then non-strict consumers will still get it right. Definitely don't write out any...

Dunqing:
> I think you'd be better off assuming strict=true, if anything.

> If you write "undefined" to the output, then non-strict consumers will still get it right. Definitely don't write out any...

Ok, I agree. FLY: The strict in tsconfig.json created by `tsc --init` is set to true.

PR #3812, opened 2024-06-21
feat(linter/eslint-plugin-react): Implement prefer-es6-class

jelly:
Rule Detail:
[link](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prefer-es6-class.md)

---

I am not sure if this is the most efficient way to implement the rule. If function calls are expensive in Rust, I could move the check inside the `let AstKind::` condition.

Another option is to make this a `run_once` rule, as I assume `ctx.semantic.classes()` can be iterated over for the `never` case. But for the `always` case we'd have to iterate over `ctx.nodes()`. So not sure if that's an improvement.

BTW: rule generation didn't work as it tries to generate with `prefer-es-6-class` which is not the name of the react rule.

### DIFF
--- a/crates/oxc_linter/src/rules/react/prefer_es6_class.rs
+++ b/crates/oxc_linter/src/rules/react/prefer_es6_class.rs
@@ -1,0 +1,185 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{
+    context::LintContext,
+    rule::Rule,
+    utils::{is_es5_component, is_es6_component},
+    AstNode,
+};
+
+fn unexpected_es6_class_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Component should use createClass instead of es6 class")
+        .with_label(span0)
+}
+
+fn expected_es6_class_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Component should use es6 class instead of createClass")
+        .with_label(span0)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct PreferEs6Class {
+    prefer_es6_class_option: PreferES6ClassOptionType,
+}
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// React offers you two ways to create traditional components: using the ES5
+    /// create-react-class module or the new ES6 class system.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// This rule enforces a consistent React class style.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// var Hello = createReactClass({
+    ///   render: function() {
+    ///     return <div>Hello {this.props.name}</div>;
+    ///   }
+    /// });
+    /// ```
+    PreferEs6Class,
+    style,
+);
+
+impl Rule for PreferEs6Class {
+    fn from_configuration(value: serde_json::Value) -> Self {
+        let obj = value.get(0);
+
+        Self {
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/prefer_es_6_class.snap
```diff
@@ -0,0 +1,36 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint-plugin-react(prefer-es6-class): Component should use createClass instead of es6 class
```
DonIsaac, line 4: I think you have these diagnostics backwards

jelly: Yup, fixed and also rewrote `with_labels` to `with_label`. Thanks for spotting.

crates/oxc_linter/src/snapshots/prefer_es_6_class.snap
```diff
@@ -0,0 +1,36 @@
...
+ 5 │ │                   return <div>Hello {this.props.name}</div>;
+ 6 │ │                 }
+ 7 │ ╰─▶             });
+ 8 │                 
+   ╰────
+
+  ⚠ eslint-plugin-react(prefer-es6-class): Component should use es6 class instead of createClass
+   ╭─[prefer_es_6_class.tsx:2:25]
+ 1 │     
+ 2 │ ╭─▶             var Hello = createReactClass({
```
DonIsaac, line 19: Consider only highlighting the `createReactClass` function name; this code block could get quite long

crates/oxc_linter/src/snapshots/prefer_es_6_class.snap
```diff
@@ -0,0 +1,36 @@
...
+ 4 │ │                   return <div>Hello {this.props.name}</div>;
+ 5 │ │                 }
+ 6 │ ╰─▶             });
+ 7 │                 
+   ╰────
+
+  ⚠ eslint-plugin-react(prefer-es6-class): Component should use createClass instead of es6 class
+   ╭─[prefer_es_6_class.tsx:2:13]
+ 1 │     
+ 2 │ ╭─▶             class Hello extends React.Component {
```
DonIsaac, line 30: Consider only highlighting the class definition (e.g. `class Hello extends React.Component`). The entire class could get quite long.

crates/oxc_linter/src/rules/react/prefer_es6_class.rs
```diff
@@ -0,0 +1,185 @@
...
+    AstNode,
+};
+
+fn unexpected_es6_class_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Component should use createClass instead of es6 class")
+        .with_label(span0)
+}
+
+fn expected_es6_class_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Component should use es6 class instead of createClass")
```
DonIsaac, line 19:
```suggestion
    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Components should use ES6 classes instead of createClass")
```

crates/oxc_linter/src/rules/react/prefer_es6_class.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use crate::{
+    context::LintContext,
+    rule::Rule,
+    utils::{is_es5_component, is_es6_component},
+    AstNode,
+};
+
+fn unexpected_es6_class_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Component should use createClass instead of es6 class")
```
DonIsaac, line 14:
```suggestion
    OxcDiagnostic::warn("eslint-plugin-react(prefer-es6-class): Components should use createClass instead of ES6 classes")
```

PR #3814, opened 2024-06-21
feat(linter/import): Implement max-dependencies

jelly:
Rule Detail:
[link](https://github.com/import-js/eslint-plugin-import/blob/v2.29.1/docs/rules/max-dependencies.md)

---

This lacks the handling of `require()` which seems to be the case for most existing `import` rules.

Another "issue" could be if you have:
```
import { foo } from "./foo";
import { bar } from "./foo";
```

But then again there should be another rule to filter these duplicate imports out and combine them into one.


crates/oxc_linter/src/rules/import/max_dependencies.rs
```diff
@@ -0,0 +1,202 @@
...
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use serde_json::Value;
+
+use crate::{context::LintContext, rule::Rule};
+
+fn max_dependencies_diagnostic(x0: &str, span1: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!("eslint-plugin-import(max-dependencies): {x0:?}"))
+        .with_help("Reduce the number of dependencies in this file")
+        .with_labels([span1.into()])
```
DonIsaac, line 11: use `with_label` if you only need to add a single label

crates/oxc_linter/src/rules/import/max_dependencies.rs
```diff
@@ -0,0 +1,202 @@
...
+                module_record.import_entries.iter().filter(|entry| entry.is_type).count();
+
+            module_count -= type_imports;
+        }
+
+        if module_count <= self.max {
+            return;
+        }
+
+        let Some(entry) = module_record.import_entries.get(self.max) else {
```
DonIsaac, line 101: Could this be moved before the `if self.ignore_type_imports` check? It could save CPU cycles by skipping iterations over import entries.

crates/oxc_linter/src/rules/import/max_dependencies.rs
```diff
@@ -0,0 +1,202 @@
...
+            r"
+             import './foo.js';
+             import './bar.js';",
+            None,
+        ),
+        // (
+        //     r"
+        //      import './foo.js';
+        //      import './bar.js';
+        //      const a = require('./foo.js');
```
DonIsaac, line 136: Do you intend on adding CJS support?

jelly: I don't, other rules also do not support CJS nor does the `module_record` know about them.

I also saw this rule ignore `require()`.
https://github.com/oxc-project/oxc/blob/09bc0fba23cd9ef9ce59e21e94f2100ef41b7b2e/crates/oxc_linter/src/rules/import/named.rs#L168

DonIsaac (approved):

PR #3815, opened 2024-06-21
feat(tasks/ast_codegen): prototype for codegen AST related code

rzvxa:
Part of https://github.com/oxc-project/oxc/issues/3819

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ oxc_macros       = { path = "crates/oxc_macros" }
 oxc_linter       = { path = "crates/oxc_linter" }
 oxc_prettier     = { path = "crates/oxc_prettier" }
 oxc_tasks_common = { path = "tasks/common" }
+oxc_ast_codegen  = { path = "tasks/oxc_ast_codegen" }
 
 napi        = "2.16.6"
 napi-derive = "2.16.5"
@@ -179,11 +180,12 @@ base64-simd         = "0.8"
 cfg-if              = "1.0.0"
 schemars            = "0.8.21"
 oxc-browserslist    = "1.0.1"
+prettyplease        = "0.2.20"
 criterion2          = { version = "0.11.0", default-features = false }
 daachorse           = { version = "1.0.0" }
 
 [workspace.metadata.cargo-shear]
-ignored = ["napi", "oxc_traverse"]
+ignored = ["napi", "oxc_traverse", "oxc_ast_codegen", "prettyplease"]
 
 [profile.dev]
 # Disabling debug info speeds up local and CI builds,
--- a/tasks/ast_codegen/src/main.rs
+++ b/tasks/ast_codegen/src/main.rs
@@ -1,0 +1,181 @@
+// TODO: remove me please!
+#![allow(dead_code)]
+mod defs;
+mod generators;
+mod linker;
+mod schema;
+
+use std::{borrow::Cow, cell::RefCell, collections::HashMap, io::Read, path::PathBuf, rc::Rc};
+
+use itertools::Itertools;
+use proc_macro2::TokenStream;
+use syn::parse_file;
+
+use defs::TypeDef;
+use generators::{AstGenerator, AstKindGenerator};
+use linker::{linker, Linker};
+use schema::{Inherit, Module, REnum, RStruct, RType, Schema};
+
+type Result<R> = std::result::Result<R, String>;
+type TypeId = usize;
+type TypeName = String;
+type TypeTable = Vec<TypeRef>;
+type IdentTable = HashMap<TypeName, TypeId>;
+type TypeRef = Rc<RefCell<RType>>;
+
+#[derive(Default)]
+struct AstCodegen {
+    files: Vec<PathBuf>,
+    generators: Vec<Box<dyn Generator>>,
+}
+
+trait Generator {
+    fn name(&self) -> &'static str;
+    fn generate(&mut self, ctx: &CodegenCtx) -> GeneratorOutput;
+}
+
+#[derive(Debug)]
+enum GeneratorOutput {
+    None,
+    One(TokenStream),
+    Many(HashMap<String, TokenStream>),
+}
+
+struct CodegenCtx {
+    modules: Vec<Module>,
+    ty_table: TypeTable,
+    ident_table: IdentTable,
+}
+
+struct CodegenResult {
+    /// One schema per definition file
+    schema: Vec<Schema>,
+    outputs: Vec<(/* generator name */ &'static str, /* output */ GeneratorOutput)>,
+}
+
+impl CodegenCtx {
+    fn new(mods: Vec<Module>) -> Self {
+        // worst case len
+        let len = mods.iter().fold(0, |acc, it| acc + it.items.len());
+        let defs = mods.iter().flat_map(|it| it.items.iter());
+
+        let mut ty_table = TypeTable::with_capacity(len);
+        let mut ident_table = IdentTable::with_capacity(len);
+        for def in defs {
+            if let Some(ident) = def.borrow().ident() {
+                let ident = ident.to_string();
+                let type_id = ty_table.len();
+                ty_table.push(TypeRef::clone(def));
+                ident_table.insert(ident, type_id);
+            }
+        }
+        Self { modules: mods, ty_table, ident_table }
+    }
+
+    fn find(&self, key: &TypeName) -> Option<TypeRef> {
+        self.ident_table.get(key).map(|id| TypeRef::clone(&self.ty_table[*id]))
+    }
+
+    fn type_id<'b>(&'b self, key: &'b TypeName) -> Option<&'b TypeId> {
+        self.ident_table.get(key)
+    }
+}
+
+impl AstCodegen {
+    #[must_use]
+    fn add_file<P>(mut self, path: P) -> Self
+    where
+        P: AsRef<str>,
+    {
+        self.files.push(path.as_ref().into());
+        self
+    }
+
+    #[must_use]
+    fn with<G>(mut self, generator: G) -> Self
+    where
+        G: Generator + 'static,
+    {
+        self.generators.push(Box::new(generator));
+        self
+    }
+
+    fn generate(self) -> Result<CodegenResult> {
+        let modules = self
+            .files
+            .into_iter()
+            .map(Module::from)
+            .map(Module::load)
+            .map_ok(Module::expand)
+            .collect::<Result<Result<Vec<_>>>>()??;
+
+        let ctx = CodegenCtx::new(modules);
+        ctx.link(linker)?;
+
+        let outputs = self
+            .generators
+            .into_iter()
+            .map(|mut gen| (gen.name(), gen.generate(&ctx)))
+            .collect_vec();
+
+        let schema = ctx.modules.into_iter().map(Module::build).collect::<Result<Vec<_>>>()?;
+        Ok(CodegenResult { schema, outputs })
+    }
+}
+
+fn files() -> std::array::IntoIter<String, 4> {
+    fn path(path: &str) -> String {
+        format!("crates/oxc_ast/src/ast/{path}.rs")
+    }
+
+    [path("literal"), path("js"), path("ts"), path("jsx")].into_iter()
+}
+
+#[allow(clippy::print_stdout)]
+fn main() -> Result<()> {
+    let CodegenResult { schema, .. } = files()
+        .fold(AstCodegen::default(), AstCodegen::add_file)
+        .with(AstGenerator)
+        .with(AstKindGenerator)
+        .generate()?;
+
+    // NOTE: Print AstKind
+    // println!(
+    //     "{}",
+    //     outputs
+    //         .into_iter()
+    //         .find(|it| it.0 == AstKindGenerator.name())
+    //         .map(|(_, output)| {
+    //             let GeneratorOutput::One(result) = output else { unreachable!() };
+    //             prettyplease::unparse(&parse_file(result.to_string().as_str()).unwrap())
+    //         })
+    //         .unwrap()
+    // );
+
+    // NOTE: Print AST
+    // println!(
+    //     "{}",
+    //     outputs
+    //         .into_iter()
+    //         .find(|it| it.0 == AstGenerator.name())
+    //         .map(|(_, output)| {
+    //             let GeneratorOutput::Many(results) = output else { unreachable!() };
+    //
+    //             results
+    //                 .into_iter()
+    //                 .map(|(k, v)| {
+    //                     format!(
+    //                         "file \"{}\":\n{}",
+    //                         k,
+    //                         prettyplease::unparse(&parse_file(v.to_string().as_str()).unwrap())
+    //                     )
+    //                 })
+    //                 .join("\n //-nextfile")
+    //         })
+    //         .unwrap()
+    // );
+
+    let schema = serde_json::to_string_pretty(&schema).map_err(|e| e.to_string())?;
+    println!("{schema}");
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1320,6 +1320,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "oxc_ast_codegen"
+version = "0.0.0"
+dependencies = [
+ "itertools 0.13.0",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "serde",
+ "serde_json",
+ "syn",
+]
+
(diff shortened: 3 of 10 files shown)


rzvxa:
* **#3815** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/ast_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        // for s in schema.definitions {}
+
+        Ok(String::new())
+    }
+}
+
+use analyzer::*;
+use schema::*;
+mod schema {
```
Boshen, line 52: @overlookmotel this is what I want, can you double confirm this is what you want as well?

tasks/ast_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    #[derive(Debug, Default)]
+    pub struct EnumMeta {
+        inherits: Vec<Inherit>,
+    }
+
+    #[derive(Debug)]
+    pub struct AstEnum {
+        pub tt: ItemEnum,
```
Boshen, line 116: nit: I try to avoid abbreviations. I will never know what this stands for the first time.

rzvxa: You are right, I'll change them in the cleanup phase. I just named stuff similar to Rust's macro convention.

crates/oxc_ast/src/ast_impl/js.rs
Boshen, line 1: Let me cherry pick these `ast_impl` changes to main, I'm getting conflicts from my other PRs :-(

rzvxa: Ok go ahead I was about to take a break anyway, I haven't moved everything around but It would be better if I do those in another PR and stack this on top of that.

Boshen: Done https://github.com/oxc-project/oxc/pull/3821, please copy things over again once its merged, there are merge conflicts.

overlookmotel:
What I had in mind for the "schema" is more abstracted, and would not use `syn`'s types (kind of an IR). e.g.:

```rs
// `TypeId` is an index into `Types`
type TypeId = usize;
struct Types(Vec<TypeDef>);

enum TypeDef {
  Struct(StructDef),
  Enum(EnumDef),
  Box(BoxDef),
  Vec(VecDef),
  Option(OptionDef),
  Cell(CellDef),
  Primitive(PrimitiveDef),
  StrRef(StrRefDef),
  // I think that's all of them?
}

struct StructDef {
  name: String,
  fields: Vec<StructFieldDef>,
  has_lifetime: bool,
}

struct StructFieldDef {
  name: String,
  type_id: TypeId,
}

struct EnumDef {
  name: String,
  variants: Vec<EnumVariantDef>,
  // For `@inherits` inherited enum variants
  inherits: Vec<TypeId>,
  has_lifetime: bool,
}

struct EnumVariantDef {
  name: String,
  type_id: Option<TypeId>, // `None` for fieldless variants
  discriminant: Option<u8>,
}

// ... and so on ...
```

My thought was codegen proceeds in 2 steps:

1. Parse `.rs` files and build `Types`.
2. Generate code from `Types` (without needing original `syn` AST).

The types above are similar to what `layout_inspect` uses: https://github.com/overlookmotel/layout_inspect/blob/master/layout_inspect/src/defs.rs

`StructDef` etc can have more fields added later on if we need them, but I think initially this is all that's needed.

`Types` could alternatively be a hashmap keyed by type name, but I thought a `Vec` which is indexed into would be simpler. It may come in useful later for every AST type to have a simple integer ID.

Does this make any sense? What do you think?

tasks/ast_codegen/src/main.rs
overlookmotel: Please see below...

rzvxa:
> What I had in mind for the "schema" is more abstracted, and would not use `syn`'s types (kind of an IR). e.g.:
> 
> ```rust
> // `TypeId` is an index into `Types`
> type TypeId = usize;
> struct Types(Vec<TypeDef>);
> 
> enum TypeDef {
>   Struct(StructDef),
>   Enum(EnumDef),
>   Box(BoxDef),
>   Vec(VecDef),
>   Option(OptionDef),
>   Cell(CellDef),
>   Primitive(PrimitiveDef),
>   StrRef(StrRefDef),
>   // I think that's all of them?
> }
> 
> struct StructDef {
>   name: String,
>   fields: Vec<StructFieldDef>,
>   has_lifetime: bool,
> }
> 
> struct StructFieldDef {
>   name: String,
>   type_id: TypeId,
> }
> 
> struct EnumDef {
>   name: String,
>   variants: Vec<EnumVariantDef>,
>   // For `@inherits` inherited enum variants
>   inherits: Vec<TypeId>,
>   has_lifetime: bool,
> }
> 
> struct EnumVariantDef {
>   name: String,
>   type_id: Option<TypeId>, // `None` for fieldless variants
>   discriminant: Option<u8>,
> }
> 
> // ... and so on ...
> ```
> 
> My thought was codegen proceeds in 2 steps:
> 
>     1. Parse `.rs` files and build `Types`.
> 
>     2. Generate code from `Types` (without needing original `syn` AST).
> 
> 
> The types above are similar to what `layout_inspect` uses (and is what AST transfer will need): https://github.com/overlookmotel/layout_inspect/blob/master/layout_inspect/src/defs.rs
> 
> `StructDef` etc can have more fields added later on if we need them, but I think initially this is all that's needed.
> 
> `Types` could alternatively be a hashmap keyed by type name, but I thought a `Vec` which is indexed into would be simpler. It may come in useful later for every AST type to have a simple integer ID.
> 
> Does this make any sense? What do you think?

I'll add the simple output, I keep the syn type in the building process so we can expand macros and generate AST in the future if neede.

you are looking at Module sturct not the Schema

overlookmotel:
> you are looking at the Module sturct not the Schema.
> I basically drop the syn types and keep their meta after the build as the schema which then can be outputed as either json or binary.

Sorry if I misunderstood. I don't quite understand what you're doing, but no matter. I'm sure you have a plan!

rzvxa:
> > you are looking at the Module sturct not the Schema.
> > I basically drop the syn types and keep their meta after the build as the schema which then can be outputed as either json or binary.
> 
> Sorry if I misunderstood. I don't quite understand what you're doing, but no matter. I'm sure you have a plan!

It is totally on me, Currently, the code is full of unused types and dirty prototypes, I'm focused on getting the whole pipeline to work and then going back to clean it up.

I also opted for interior mutability and cloning where needed since there aren't many performance concerns in this crate. I thought it would make development easier if we could be a little careless with how we pass around stuff, After all this script only runs once in a while and the runtime is minimal so not having to fight with borrow checker constantly is much more favorable than having a well-optimized build script.

tasks/ast_codegen/schema.json
rzvxa, line 1: @overlookmotel temporarily added the generated schema to the git repo so you can give it a look. Some stuff might be missing currently but we can iterate on it as our requirements change.

rzvxa:
@Boshen @overlookmotel This is almost ready, You may want to give it a spin.

overlookmotel:
Rez and I just discussed this PR and surrounding questions on video chat.

I *think* conclusion of our conversation was:

* We may want to remove the `R*` types later on, and do codegen purely from the `TypeDef` schema.
* We don't need to do that now - LGTM.
* AST files (which are the *source* for the codegen, not codegen's output) will need to include proc macros in some cases.
* We can make the bulk of the work happen in the codegen/build script so those proc macros can be "dumb", and therefore fast to compile and run. i.e. minimize impact of the proc macros.
* The question of whether to codegen the AST types themselves remains unresolved, but we now understand each others' arguments for/against better, and we are closer to convergence. This question doesn't need to be answered right now anyway.

@rzvxa Please correct me if any of the above does not accurately describe our conversation.

overlookmotel:
One oddity which affects this PR:

Codegen is only parsing AST types with a `#[visited_node]` attr. But there are a few types which are part of AST, but do not have a `#[visited_node]`. e.g. `JSXOpeningFragment` and `JSXClosingFragment`.

Maybe these are the only 2, or maybe there are a few more - I'm afraid I can't remember.

Reason they don't have a `#[visited_node]` attr is because we don't want to generate `enter_jsx_opening_fragment` visitors in `Traverse`. But they *may* need to be parsed by the codegen anyway. Certainly AST transfer needs them to be part of the schema, but I don't know if they also need to be included in `AstKind`?

Maybe we should add another no-op attr macro `#[unvisited_node]` to add to these types, so that codegen can find them too.

Not sure if this needs to be addressed before merging this PR or not - @rzvxa I think you can answer that better than me.

rzvxa:
> One oddity which affects this PR:
> 
> Codegen is only parsing AST types with a `#[visited_node]` attr. But there are a few types which are part of AST, but do not have a `#[visited_node]`. e.g. `JSXOpeningFragment` and `JSXClosingFragment`.
> 
> Maybe these are the only 2, or maybe there are a few more - I'm afraid I can't remember.
> 
> Reason they don't have a `#[visited_node]` attr is because we don't want to generate `enter_jsx_opening_fragment` visitors in `Traverse`. But they _may_ need to be parsed by the codegen anyway. Certainly AST transfer needs them to be part of the schema, but I don't know if they also need to be included in `AstKind`?
> 
> Maybe we should add another no-op attr macro `#[unvisited_node]` to put on these types, so that codegen can find them too.
> 
> Not sure if this needs to be addressed before merging this PR or not - @rzvxa I think you can answer that better than me.

Is there any type in our current `ast/*.rs` files that aren't necessary for AST transfer?
### A:
We can parse them all and mark their metadata as `visitable` if we need them all.

If that's not the case then we have 2 options:

### B:
Either extract these unnecessary types and move them somewhere else so that we can go to `A`.

### C:
Introduce a new marker attribute such as `#[ast_node]` to address this issue and similar to `A` we load all types in the ast files and mark them with both `ast_node` and `visitable`.

At the generator level, we filter through these using their metadata and schema information depending on our needs.

rzvxa:
> I don't know if they also need to be included in AstKind?

This I don't know either, Maybe @Boshen can better answer this.

Boshen:
I don't know either haha, better to replicate what's hand written right now, without introducing more conflicts 😅

rzvxa:
> Reason they don't have a #[visited_node] attr is because we don't want to generate enter_jsx_opening_fragment visitors in Traverse.

May I know your reasoning for this decision? There are currently only 20 types that we don't visit, Shouldn't we add all of them to the traverse (and visitors) just for the sake of completeness? If we terminate the walk on these nodes they would have empty implementations by default so they wouldn't affect the performance.

All of these types seem like stuff that we might want to visit one way or another, Either for future linters/transformers or when it comes to working on the minifier and the formatter(which most probably would want to know about things like JSX tags).

overlookmotel:
> May I know your reasoning for this decision?

For `JSXOpeningFragment` and `JSXClosingFragment` I just thought it was pointless to visit them as they're only `Span`s. We don't have `visit_span` (and I don't think we should).

> `Unvisited names: [ "RegExp", ... ]`

How many of these are structs? Most look like fieldless enums e.g. `FunctionType`, which personally I don't think we should visit.

rzvxa:
> How many of these are structs?

```
Unvisited Item names: [
    "enum AccessorPropertyType",
    "enum ClassType",
    "enum FormalParameterKind",
    "enum FunctionType",
    "enum ImportOrExportKind",
    "enum MethodDefinitionKind",
    "enum MethodDefinitionType",
    "enum PropertyDefinitionType",
    "enum PropertyKind",
    "enum TSAccessibility",
    "enum TSMappedTypeModifierOperator",
    "enum TSMethodSignatureKind",
    "enum TSModuleDeclarationKind",
    "enum TSTypeOperatorOperator",
    "enum VariableDeclarationKind",
    "struct EmptyObject",
    "struct JSXClosingFragment",
    "struct JSXOpeningFragment",
    "struct RegExp",
    "struct TemplateElementValue",
]
```

overlookmotel:
Rzvxa suggested on another channel that maybe we should replace `#[visited_node]` with `#[ast_node(visit)]`. Then *unvisited* types could be just `#[ast_node]`.

I have avoided the name `ast_node` because it is reminiscent of SWC's `#[ast_node]` macro which generates a ginormous amount of code. I wanted to make clear that our macro does not do that, so avoided using the same name. But maybe "ast_node" is the best name for it, as these types are - quite literally - AST nodes.

rzvxa:
> > How many of these are structs?
> 
> ```
> Unvisited Item names: [
>     "enum AccessorPropertyType",
>     "enum ClassType",
>     "enum FormalParameterKind",
>     "enum FunctionType",
>     "enum ImportOrExportKind",
>     "enum MethodDefinitionKind",
>     "enum MethodDefinitionType",
>     "enum PropertyDefinitionType",
>     "enum PropertyKind",
>     "enum TSAccessibility",
>     "enum TSMappedTypeModifierOperator",
>     "enum TSMethodSignatureKind",
>     "enum TSModuleDeclarationKind",
>     "enum TSTypeOperatorOperator",
>     "enum VariableDeclarationKind",
>     "struct EmptyObject",
>     "struct JSXClosingFragment",
>     "struct JSXOpeningFragment",
>     "struct RegExp",
>     "struct TemplateElementValue",
> ]
> ```

I checked all of the types, None of the enums have either named or unnamed fields. And structures as @overlookmotel explained are useless to visit. So I guess we won't need to add these to our AstKind since they won't get visited and subsequently, they won't need an `AstNode` or `AstKind` associated with them.

rzvxa:
> I have avoided the name ast_node because it is reminiscent of SWC's #[ast_node] macro

Oh, those poor souls who have PTSD (post-traumatic SWC disorder) won't appreciate the new name at all🥺

overlookmotel:
> I checked all of the types, None of the enums have either named or unnamed fields. And structures as @overlookmotel explained are useless to visit. So I guess we won't need to add these to our AstKind since they won't get visited and subsequently, they won't need an AstNode or AstKind associated with them.

Thanks. I checked them all too and was about the write the same thing!

So... I think conclusion is: AST transfer needs them all in its schema, but we can do that later. For purposes of codegen, we don't need any of them. So it's not a blocker to this PR. LGTM!

rzvxa:
> So... I think conclusion is: AST transfer needs them all in its schema, but we can do that later. For purposes of codegen, we don't need any of them. So it's not a blocker to this PR. LGTM!

Well, good news mate, I've already taken care of it😅

Boshen:
@overlookmotel can you help me review and merge this one? This is beyond me 😅 

I'll chime in when the edit - save workflow is up and running.

rzvxa:
@overlookmotel I suggest you only review #3852 there is some stuff I've changed after this PR and I was too lazy to cherry-pick and merge them here.

rzvxa:
@overlookmotel Just to make your life easier, I'm adding comments to help with the review process, I'll mention you in up stack PR as soon as I submit my docs commit.

tasks/ast_codegen/schema.json
overlookmotel: Thanks. I had a look. It's not exactly what AST transfer needs, but we can iterate on it later.

tasks/ast_codegen/src/main.rs
overlookmotel: rzvxa and I have discussed. In my opinion this PR is good as first step. We can iterate on it later.

Cargo.toml
```diff
@@ -179,11 +180,12 @@ base64-simd         = "0.8"
 cfg-if              = "1.0.0"
 schemars            = "0.8.21"
 oxc-browserslist    = "1.0.1"
+prettyplease        = "0.2.20"
 criterion2          = { version = "0.11.0", default-features = false }
 daachorse           = { version = "1.0.0" }
 
 [workspace.metadata.cargo-shear]
-ignored = ["napi", "oxc_traverse"]
+ignored = ["napi", "oxc_traverse", "oxc_ast_codegen", "prettyplease"]
```
overlookmotel, line 188: Small nit: Does `prettyplease` need to be included here as well as in `tasks/ast_codegen/Cargo.toml`?

rzvxa: That's why I told you to only review the up stack PR, It is one of those small changes I did in the next PR and haven't cherry-picked to merge here.

rzvxa: Yes, it does, otherwise I get errors from shear.

overlookmotel: Ah OK.

overlookmotel (approved):

PR #3825, opened 2024-06-22
feat(linter): implement @typescript-eslint/no-non-null-assertion

kaykdm:

original implementation

- code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-non-null-assertion.ts
- test: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/no-non-null-assertion.test.ts
- doc: https://typescript-eslint.io/rules/no-non-null-assertion/

typescript-eslint has a feature that is manually fixable by editor suggestions on this rule, but oxc does not have one as far as I know, so the implementation is simple compared to typescript-eslint

### DIFF
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -137,6 +137,7 @@ mod typescript {
     pub mod no_misused_new;
     pub mod no_namespace;
     pub mod no_non_null_asserted_optional_chain;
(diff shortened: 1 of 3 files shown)


DonIsaac (approved):
LGTM

Issue #2958, opened 2024-04-13
ast: revisit `Modifiers`


Boshen:
The `Modifiers` data is used for error recoverable, but it causes a lot of confusion while working on the AST. Consider removing it.

Boshen:
I want a data structure to clearly indicate the allowed modifiers, with other disallowed modifiers attached to it.

overlookmotel:
Would it be OK to:

Only allow valid modifiers to be encoded in the AST.

When parser finds a modifier in an invalid location, it raises an error and then ignores it (i.e. doesn't put it in AST) and can then continue parsing the rest of the file.

Then we get a simpler AST which we can rely on being valid, but don't cause the parser to grind to a halt when it hits something illegal.

Does that count as recoverable?

Boshen:
AST no longer uses `Modifiers` but parser still uses it for maximum control.

PR #3838, opened 2024-06-23
feat(semantic): check for illegal symbol modifiers

DonIsaac:


crates/oxc_semantic/tests/integration/util/mod.rs
```diff
@@ -64,11 +81,68 @@ impl<'a> SemanticTester<'a> {
...
+                semantic_ret
+                    .errors
+                    .iter()
+                    .map(ToString::to_string)
+                    .intersperse("\n\n".to_owned())
+                    .collect::<String>()
+            ),
+            _ => semantic_ret.semantic,
+        }
+        // if !semantic_ret.errors.is_empty() {
```
Boshen, line 125: Some commented out code

Boshen (commented):
I was thinking about removing `Modifiers` from the AST nodes in https://github.com/oxc-project/oxc/issues/2958

I'll accept this PR and refactor things out later.

crates/oxc_semantic/tests/integration/util/mod.rs
DonIsaac: removed

PR #3842, opened 2024-06-23
feat(linter): accept multiple fixes when fix code

mysteryven:
Want to add this because of there are bunch of fix codes in [consistent-type-import](https://github.com/typescript-eslint/typescript-eslint/blob/e408b93e48e5199d83a8d99d1e27126d2dd8bc8f/packages/eslint-plugin/src/rules/consistent-type-imports.ts#L610-L945) are not easy to port.

### DIFF
--- a/crates/oxc_linter/src/context.rs
+++ b/crates/oxc_linter/src/context.rs
@@ -9,7 +9,7 @@ use oxc_syntax::module_record::ModuleRecord;
 use crate::{
     config::OxlintRules,
     disable_directives::{DisableDirectives, DisableDirectivesBuilder},
-    fixer::{Fix, Message, RuleFixer},
+    fixer::{CompositeFix, Message, RuleFixer},
     javascript_globals::GLOBALS,
     AllowWarnDeny, OxlintConfig, OxlintEnv, OxlintGlobals, OxlintSettings,
 };
@@ -171,14 +171,16 @@ impl<'a> LintContext<'a> {
     }
 
     /// Report a lint rule violation and provide an automatic fix.
-    pub fn diagnostic_with_fix<F: FnOnce(RuleFixer<'_, 'a>) -> Fix<'a>>(
-        &self,
-        diagnostic: OxcDiagnostic,
-        fix: F,
-    ) {
+    pub fn diagnostic_with_fix<C, F>(&self, diagnostic: OxcDiagnostic, fix: F)
+    where
+        C: Into<CompositeFix<'a>>,
+        F: FnOnce(RuleFixer<'_, 'a>) -> C,
+    {
         if self.fix {
             let fixer = RuleFixer::new(self);
-            self.add_diagnostic(Message::new(diagnostic, Some(fix(fixer))));
+            let composite_fix: CompositeFix = fix(fixer).into();
+            let fix = composite_fix.normalize_fixes(self.source_text());
+            self.add_diagnostic(Message::new(diagnostic, Some(fix)));
         } else {
             self.diagnostic(diagnostic);
         }
--- a/crates/oxc_linter/src/fixer.rs
+++ b/crates/oxc_linter/src/fixer.rs
@@ -22,6 +22,92 @@ impl<'a> Fix<'a> {
     }
 }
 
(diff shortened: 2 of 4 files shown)


mysteryven:
* **#3842** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3842?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (commented):
What's the difference between this and https://github.com/oxc-project/oxc/blob/d76f34b1300069adc8def5f9fd1b21e05f78092d/crates/oxc_linter/src/fixer.rs#L29 ?

Ideally we should only have 1 - a port from eslint.

mysteryven:
I need to compose multi `Fix` at one time(one line code may have to delete and replace). And have three thoughts at first, one is shown as this PR, and another is change our `Fix`

```diff
#[derive(Debug, Clone, Default)]
	pub struct Fix<'a> {
-		pub contents: Cow<'a, str>,
+		pub contents: Vec<Cow<'a, str>>,
-		pub spans: Span,
+		pub spans: Vec<Span>,
}

 ctx.diagnostic_with_fix(
      no_import_type_side_effects_diagnostic(import_decl.span),
      |fixer| {
          let raw = ctx.source_range(import_decl.span);
          let mut string_fixer = StringFixer::new(raw);

          let mut fix = fixer.new_fix();
          fix.replace(span, "abc");
          fix.delete(span, "abc");
		fix
 })
```

I checked [rule-fixer.js](https://github.com/eslint/eslint/blob/219aecb78bc646d44bad27dc775a9b3d3dc58232/lib/linter/rule-fixer.js), and found change Fix is not keep same with ESLint, ESLint produces one fix in RuleFixer's method, so I didn't change `RuleFixer`

The third option is let the fixer return `Vec<fix>`, I thought it will influence many files that have a fixer.

```diff
#[derive(Clone)]
pub struct Message<'a> {
    pub error: OxcDiagnostic,
    pub start: u32,
    pub end: u32,
+ pub fixs: Option<Vec<Fix<'a>>>,
- pub fix: Option<Fix<'a>>,
    fixed: bool,
}
```

Finally I chose option 1. I thought this way only need to change the least things =.=

Now, looks like option 3 is better, are we to accept it? or would you like to give me some advice?

Boshen:
My question is always how does eslint do it? Can we align our implementation with theirs?

crates/oxc_linter/src/context.rs
```diff
@@ -166,14 +166,15 @@ impl<'a> LintContext<'a> {
     }
 
     /// Report a lint rule violation and provide an automatic fix.
-    pub fn diagnostic_with_fix<F: FnOnce(RuleFixer<'_, 'a>) -> Fix<'a>>(
+    pub fn diagnostic_with_fix<F: FnOnce(RuleFixer<'_, 'a>) -> CompositeFix<'a>>(
```
DonIsaac, line 169: Maybe return `C : Into<CompositeFix<'a>>`?

crates/oxc_linter/src/fixer.rs
```diff
@@ -20,6 +20,83 @@ impl<'a> Fix<'a> {
     pub fn new<T: Into<Cow<'a, str>>>(content: T, span: Span) -> Self {
         Self { content: content.into(), span }
     }
+
+    pub fn into_composite_fix(self) -> CompositeFix<'a> {
+        CompositeFix::Single(self)
+    }
+}
+
+pub enum CompositeFix<'a> {
```
DonIsaac, line 29: Consider implementing `From<Fix<'a>>` and `From<Vec<Fix'a>>`

crates/oxc_linter/src/context.rs
mysteryven: Nice suggestion!

Boshen (commented):
Are there unit tests from eslint? We should port them over.

mysteryven:
> Are there unit tests from eslint? We should port them over.

Added now, and it really exposed my bug 😥

PR #3863, opened 2024-06-23
fix(semantic): do not bind type/value references to value/type symbols

### LINKED ISSUE #3799: Incorrect reference resolve in function parameter's type
```typescript
import * as foo from 'foo';

function bar(foo: foo) {}
```

We treat function parameter `foo`'s type refers to itself.

CleanShot 2024-06-21 at 10 00 04@2x

CleanShot 2024-06-21 at 10 03 02@2x

1. [our playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAIC1gICAgICAgIC0G8rnONK887cdP8rOL8UC6QnA9HB4rpM9rpIjLjycp13Au8%2FJhmDB5RAx7SoW4A0Y7lXumf54QLmA) 
2. [typescript playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wGYQpqERwDkWEhA3AFBVUYCuAdgMYzAT0oAUJAXJtgJRwA3hThwKAXyA)

DonIsaac:
Fixes #3799

Only symbols with the desired "meaning" (value/type) get resolved when binding references to symbol ids.

I also updated some code in `oxc_traverse`, but to be honest I'm not sure what's going on there and I'd love some input on what to do there. It seems to be used by `oxc_semantic`, but not completely?

## References
- [TSC, `createNameResolver`](https://github.com/microsoft/TypeScript/blob/5d70bf894efe9014d24d4ba439807311e5b33fb8/src/compiler/utilities.ts#L11221)

## Todo
- [ ] Emit an error if a reference is unresolved because of a meaning difference (ts error code 2749 - [example](https://ts-ast-viewer.com/#code/FAYw9gdgzgLgBAFTgXjgRmAMwK4RDAS0jkzDAAoAPALkQEo4BvYOOAJwFMZs2I5K4AakTAAvsAlA))
- [ ] Type annotations should count as a type+read reference for type symbols. Maybe do this in another PR?


overlookmotel:
Thanks for putting this together @DonIsaac. A couple of responses:

1. This PR aims to resolve reference identifiers correctly depending on whether they are types or variables. I'm not sure if this is what we want to do or not. As mentioned in https://github.com/oxc-project/oxc/issues/3799#issuecomment-2184079221, maybe the best solution is not to resolve type identifier references at all. i.e. The scope tree becomes purely for variable bindings and type bindings/references are ignored.

2. If we *do* want to include type bindings in the scope tree, and resolve references to them, then this PR is a good way to go about it, but it's not complete. In particular, we need to figure out how to support 2 bindings with same name in the same scope (1 variable binding + 1 type binding). Also, I don't believe semantic at present visits all type identifier references.

```ts
type Foo = number;
const Foo: Foo = 123;
function f(foo: Foo) { // `Foo` here refers to `type Foo`
    return Foo; // `Foo` here refers to `const Foo`
}
```

I think before going further, we should figure out what our goal is.

DonIsaac:
For anyone reading this PR thread, the conversation is taking place on #3799.

Dunqing:
@Boshen and I are studying how TyepScript resolves references and what `container` does. We'll probably be refactoring our Scope soon. maybe we'll remove the ScopeTree, and in its place, we might have a TypesScript-like container implementation.

Boshen:
We are studying how tsc does this. Closing for house keeping. cc @Dunqing You may want to copy over some of the tests from this PR.

PR #3876, opened 2024-06-24
feat(napi/transform): add `crate-type: "lib"`

underfin:
The pr add a `crate-type: lib` for transform binding, it will be used for rolldown export binding. I tested the binary size 895k -> 944K, it could be acceptable.

Here bump the napi library version due to rolldown napi version is used `3.0.0-alpha`, different napi version make rolldown release build error.

It means the binding library need to publish to crate.io, @Boshen Could you add it to ci pipeline for it? I'm not familiar for it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1078,13 +1078,12 @@ dependencies = [
 
 [[package]]
 name = "napi"
-version = "2.16.6"
+version = "3.0.0-alpha.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Boshen:
This doesn't contain oxc_transformr yet, please create an issue if you want one 😅

underfin:
Yeah. I konw i will work on it at next.

Boshen:
> It means the binding library need to publish to crate.io

Ok, I'll take a look, feels pretty weird though 🤔

PR #3881, opened 2024-06-24
feat(linter): implement @typescript-eslint/prefer-namespace-keyword

kaykdm:

original implementation

- code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/prefer-namespace-keyword.ts
- test: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/prefer-namespace-keyword.test.ts
- doc: https://typescript-eslint.io/rules/prefer-namespace-keyword/

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/prefer_namespace_keyword.rs
+++ b/crates/oxc_linter/src/rules/typescript/prefer_namespace_keyword.rs
@@ -1,0 +1,95 @@
+use oxc_ast::{
+    ast::{TSModuleDeclarationKind, TSModuleDeclarationName},
+    AstKind,
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Default, Clone)]
+pub struct PreferNamespaceKeyword;
+
+declare_oxc_lint!(
+    /// ### What it does
+    /// Require using namespace keyword over module keyword to declare custom TypeScript modules.
+    ///
+    /// ### Why is this bad?
+    /// TypeScript historically allowed a form of code organization called "custom modules" (module Example {}), later renamed to "namespaces" (namespace Example).
+    ///
+    /// Namespaces are an outdated way to organize TypeScript code. ES2015 module syntax is now preferred (import/export).
+    ///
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/typescript/prefer_namespace_keyword.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// declare module foo {}
+    /// ```
+    PreferNamespaceKeyword,
+    restriction,
+);
+
+fn prefer_namespace_keyword_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("typescript-eslint(prefer-namespace-keyword): Use 'namespace' instead of 'module' to declare custom TypeScript modules.")
+        .with_help("'Require using `namespace` keyword over `module` keyword to declare custom TypeScript modules")
+        .with_labels([span0.into()])
```
DonIsaac, line 37: prefer `with_label` when only labelling a single span

crates/oxc_linter/src/rules/typescript/prefer_namespace_keyword.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::TSModuleDeclaration(decl) = node.kind() else { return };
+        if decl.kind != TSModuleDeclarationKind::Module {
+            return;
+        }
+        let TSModuleDeclarationName::Identifier(id) = &decl.id else { return };
+
+        ctx.diagnostic_with_fix(prefer_namespace_keyword_diagnostic(decl.span), |fixer| {
+            // replace `module` with `namespace`
+            fixer.replace(Span::new(id.span.start - 7, id.span.start - 1), "namespace")
```
DonIsaac, line 50: This looks like it will break if there is more than once space between `module` and the identifier. Is there no span for the `module` keyword?

Boshen:
Extending ["plugin:@typescript-eslint/stylistic"](https://typescript-eslint.io/users/configs#stylistic) in an [ESLint configuration](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#extending-configuration-files) enables this rule.

This appears to be a stylistic rule, which we don't accept. Let me make a note of this so we don't accept more stylistic rules.

PR #3889, opened 2024-06-24
fix(oxlint): properly report error

lucab:
This fixes an incorrect error formatting in the lint runner logic.


rzvxa:
Thanks for your contribution!

PR #3888, opened 2024-06-24
feat(ast_codegen): generate `ast_kind.rs`.

rzvxa:
[List of 68 blacklisted items](https://github.com/oxc-project/oxc/pull/3888/files#diff-59cbea4a1444e752992821d3429a5f27bf1027a530977d075ec89aa55995912aR11)

PR #3855, opened 2024-06-23
refactor(ast): add span field to the `BindingPattern` type.

rzvxa:
So we don't have to introduce a special case while generating `GetSpan` implementations for all of our Ast types.

PR #3852, opened 2024-06-23
feat(ast_codegen): add `ImplGetSpanGenerator`.

rzvxa:
This one is ready to replace the handwritten module.

PR #3890, opened 2024-06-24
feat(ast)!: generate `ast_builder.rs`.

rzvxa:
### Every structure has 2 builder methods:

1. `xxx` e.g. `block_statement`
```rust
    #[inline]
    pub fn block_statement(self, span: Span, body: Vec<'a, Statement<'a>>) -> BlockStatement<'a> {
        BlockStatement { span, body, scope_id: Default::default() }
    }
```
2. `alloc_xxx` e.g. `alloc_block_statement`
```rust
    #[inline]
    pub fn alloc_block_statement(
        self,
        span: Span,
        body: Vec<'a, Statement<'a>>,
    ) -> Box<'a, BlockStatement<'a>> {
        self.block_statement(span, body).into_in(self.allocator)
    }
```

### We generate 3 types of methods for enums:

1. `yyy_xxx` e.g. `statement_block`
```rust
    #[inline]
    pub fn statement_block(self, span: Span, body: Vec<'a, Statement<'a>>) -> Statement<'a> {
        Statement::BlockStatement(self.alloc(self.block_statement(span, body)))
    }
```
2. `yyy_from_xxx` e.g. `statement_from_block`
```rust
    #[inline]
    pub fn statement_from_block<T>(self, inner: T) -> Statement<'a>
    where
        T: IntoIn<'a, Box<'a, BlockStatement<'a>>>,
    {
        Statement::BlockStatement(inner.into_in(self.allocator))
    }
```
3. `yyy_xxx` where `xxx` is inherited e.g. `statement_declaration`
```rust
    #[inline]
    pub fn statement_declaration(self, inner: Declaration<'a>) -> Statement<'a> {
        Statement::from(inner)
    }
```

------------

### Generic parameters:

We no longer accept `Box<'a, ADT>`, `Atom` or `&'a str`, Instead we use `IntoIn<'a, Box<'a, ADT>>`, `IntoIn<'a, Atom<'a>>` and `IntoIn<'a, &'a str>` respectively.
It allows us to rewrite things like this:
```rust
let ident = IdentifierReference::new(SPAN, Atom::from("require"));
let number_literal_expr = self.ast.expression_numeric_literal(
    right_expr.span(),
    num,
    raw,
    self.ast.new_str(num.to_string().as_str()),
    NumberBase::Decimal,
);
```
As this:
```rust
let ident = IdentifierReference::new(SPAN, "require");
let number_literal_expr = self.ast.expression_numeric_literal(
    right_expr.span(),
    num,
    raw,
    num.to_string(),
    NumberBase::Decimal,
);
```


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3890?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3890** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3890?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3888** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3852** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3852?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3815** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3815?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3855** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
This one is trickier, because all kinds of weird methods have been added to AST builder over time, and I don't think there's much of a pattern to what's implemented and what's not, and there's not always consistency in naming of methods. I think we might need to standardize the existing methods in the existing manually-written impl, and standardize the naming of those methods, before moving over to a codegen-ed version.

There's also a danger of generating lots more code than we have at present, and hurting compile times, if we try to make the codegen-ed version cover all possible cases. If the amount of code generated is excessive, maybe we need attrs on the AST types which indicate which builders to generate?

```rs
#[visited_node]
#[builder(Expression)]
struct BooleanLiteral {
    #[cfg_attr(feature = "serialize", serde(flatten))]
    pub span: Span,
    pub value: bool,
}
```

This would generate:

```rs
impl<'a> AstBuilder<'a> {
    // Always generated by default
    fn boolean_literal(self, span: Span, value: bool) -> BooleanLiteral {
        BooleanLiteral { span, value }
    }

    // Generated because of `#[builder(Expression)]` attr
    fn boolean_literal_expression(self, span: Span, value: bool) -> Expression<'a> {
        Expression::BooleanLiteral(self.alloc(BooleanLiteral { span, value }))
    }
}
```

An example of what we could have (but haven't) generated:

```rs
impl<'a> AstBuilder<'a> {
    fn boolean_literal_array_expression_element(self, span: Span, value: bool) -> ArrayExpressionElement<'a> {
        ArrayExpressionElement::BooleanLiteral(self.alloc(BooleanLiteral { span, value }))
    }
}
```

overlookmotel:
Probably what I suggested above is too complicated. Maybe better way to tackle this is incrementally?

e.g. Start by replacing all the methods for `Expression`s and `Statement`s as 1 PR. Then continue whittling away the easy/standard ones (e.g. `Declaration`s) in further PRs, until most of manual impls file is cleared out, and then it's easier to see what oddities remain.

For all `Expression`s, we could generate:

```rs
impl BooleanLiteral {
    #[inline]
    fn new(span: Span, value: bool) -> BooleanLiteral {
        BooleanLiteral { span, value }
    }
}

impl<'a> AstBuilder<'a> {
    #[inline]
    fn boolean_literal(self, span: Span, value: bool) -> Expression<'a> {
        Expression::BooleanLiteral(self.alloc(BooleanLiteral::new(span, value)))
    }
}
```

If consumer wants an `Argument::BooleanLiteral`, they can convert themselves with `Argument::from(ast.boolean_literal(span, value))`.

As I see it, the main value of `AstBuilder` is to hold the allocator and so provide the nicer API `self.ast.boolean_literal(span, value)` rather than `BooleanLiteral::expression(span, value, self.allocator)`. So any function which doesn't require the allocator doesn't need to be in `AstBuilder`.

rzvxa:
@overlookmotel let me mull this one for a little while. I'll come back to you when I have figured out a working plan to discuss.

rzvxa:
Right now I can think of 3 options to unify all of our builder calls(both in terms of name and signature),

1. have a bunch of `enum_variant` methods like this:
```rust
fn statement_block_statement(self, it: BlockStatement<'a>) -> Statement<'a> {
    Statement::BlockStatement(self.alloc(it))
}

fn block_statement(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
) -> BlockStatement<'a> {
    BlockStatement {
        span,
        body,
        scope_id: Default::default(),
    }
}
```

then we use it like this:

```rust
let block = self.ast.block_statement(span, body);
let stmt = self.ast.statement_block_statement(block);
```

2.

We have methods like these:

```rust
fn statement_block_statement(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
    ) -> Statement<'a> {
    Statement::BlockStatement(self.alloc(self.block_statement(span, body)))
}

fn block_statement(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
) -> BlockStatement<'a> {
    BlockStatement {
        span,
        body,
        scope_id: Default::default(),
    }
}
```

and use like this:

```rust
let stmt = self.ast.statement_block_statement(span, body);
```

This is nice but it would mean that if you want to wrap an already existing `block` in a `Statement` enum you have to do something like this:

```rust
let stmt = Statement::BlockStatement(self.alloc(block));
```

Or rename the method we previously called `statement_block_statement` to `new_statement_block_statement`; and generate more functions to use like this(similar to `1`):

```rust
fn statement_block_statement(self, it: BlockExpression<'a>) -> Expression<'a> {
    Statement::BlockStatement(self.alloc(it))
}
```

3.

We generate these methods:

```rust
fn block_statement(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
) -> Box<'a, BlockStatement<'a>> {
    self.alloc(BlockStatement {
        span,
        body,
        scope_id: Default::default(),
    })
}
```

we also generate `From` implementations for variant initialization.

If we want to initialize a try we write something like this:


```rust
let try = self.ast.try_statement(
    span,
    self.ast.block_statement(block_span, block_body)
);
```

If we want a `Statement` from our block we just do this:

```rust
let stmt = self.ast.block_statement(span, body).into();
```


#### TL;DR

So to sum up here all 3 approaches that I can think of together:

```rust
// 1
let block = self.ast.block_statement(span, body);
let block_stmt = self.ast.statement_block_statement(block);
// 2
let block_stmt = self.ast.statement_block_statement(span, body);
// 3
let block_stmt = self.ast.block_statement(span, body).into();


// 1
let try_stmt = self.ast.try_statement(
    span,
    self.ast.alloc(self.ast.block_statement(block_span, block_body))
    handler,
    finalizer,
)
// 2
let try_stmt = self.ast.try_statement(
    span,
    self.ast.alloc(self.ast.block_statement(block_span, block_body))
    handler,
    finalizer,
)
// 3
let try_stmt = self.ast.try_statement(
    span,
    self.ast.block_statement(block_span, block_body).into()
    handler,
    finalizer,
)
```


@Boshen @overlookmotel I would like to know your thoughts on this, And what are your suggestions.

rzvxa:
There is actually a fourth and fifth option which @overlookmotel mentioned these are the 3 that I can think of.

overlookmotel:
Not a complete answer, but some thoughts...

In my opinion, our guiding principles here should be:

1. Make the most common action possible in minimal number of keystrokes. In the case of e.g. `BooleanLiteral`, most commonly you want an `Expression`. It already requires very verbose code to construct even simple AST nodes in the transformer, and we should try very hard not to make it worse!
2. Avoid adding methods which don't allocate to AST builder. They can be implemented directly on the types e.g. `BooleanLiteral::new(span, value)` -> `BooleanLiteral` vs `ast.boolean_literal(span, value)` -> `Expression`.
3. Ensure our methods don't encourage/require using `Box::unbox`. It is inefficient to unbox a value which is in the arena back onto stack, only to allocate it back into arena again in a new `Box`. I worry that APIs which return `Box<T>` will encourage this antipattern.

Personally, I'm not a fan of `.into()`. When reading code, I often find myself asking "but into what?". In a code editor *sometimes* Rust Analyser answers that question with annotations, but not always. And when reviewing PRs in Github obviously you don't get any type info that isn't explicit in the code. So I find `.into()` obfuscates meaning.

Obviously people do use `.into()`, and I don't want to make that illegal or anything! But personally I don't think we should *encourage* it as a core idiom of interacting with the AST builder, as we'll have `.into()` everywhere. For this reason, I'm not a fan of rzvxa's solution no. 3 above.

rzvxa:
> 3. Ensure our methods don't encourage/require using Box::unbox. It is inefficient to unbox a value which is in the arena back onto stack, only to allocate it back into arena again in a new Box. I worry that APIs which return Box<T> will encourage this antipattern.

I mostly agree with your 2 other points but this one can be resolved by generating our functions like this:

```rust
    pub fn ts_conditional_type<R>(
        self,
        span: Span,
        check_type: TSType<'a>,
        extends_type: TSType<'a>,
        true_type: TSType<'a>,
        false_type: TSType<'a>,
    ) -> R
    where
        R: From<Box<'a, TSConditionalType<'a>>>,
    {
        self.alloc(TSConditionalType { span, check_type, extends_type, true_type, false_type }).into()
    }
```

But this introduces the inconvenience of generics and doesn't address all of our needs.

> Avoid adding methods which don't allocate to AST builder. 

It all comes down to what we want our node constructions to look like, Based on this [comment](https://github.com/oxc-project/oxc/issues/2818#issuecomment-2027539381) from Boshen I assumed that we prefer to build everything through our builder as a mean to centralize this process, If we ever want to go down this road we might want to add a private marker to all of our types and enforce their creation only though the builder or constructors in some cases.

Personally, I prefer 

```rust
Something::new(span, value)
// to
Something {
    span,
    value,
    scope_id: Default::default(),
    symbol_id: Default::default(),
    other_stuff: Default::default(),
}
// or even this
Something {
    span,
    value,
    ...Something::default()
}
```

----------------

About your concerns with `into`, I agree with you (Especially this part) 
> And when reviewing PRs in Github obviously you don't get any type info 

But I think it is a nice pattern compared to the alternative which is having overly verbose function calls to assign something which isn't going to live long as it is and is designed to be assigned somewhere immediately.

If we don't want to centralize the ast creation to the builder we can nuke it and go for something much less verbose like this(which is inspired by overlookmotel's idea):

We introduce a set of `FromIn` and `IntoIn` which would look like this:
```rust
trait IntoIn<'a, T> {
    #[inline]
    fn into_in(self, allocator: &'a Allocator) -> T;
}
trait FromIn<'a, T> {
    #[inline]
    fn from_in(T, allocator: &'a Allocator) -> Self;
}
```

Then instead of writing this:

```rust
let kind = VariableDeclarationKind::Let;
let declarations = {
    let ident = BindingIdentifier::new(SPAN, name.clone());
    let pattern_kind = self.ctx.ast.binding_pattern_identifier(ident);
    let binding = self.ctx.ast.binding_pattern(pattern_kind, None, false);
    let decl = self.ctx.ast.variable_declarator(SPAN, kind, binding, None, false);
    self.ctx.ast.new_vec_single(decl)
};
Declaration::VariableDeclaration(self.ctx.ast.variable_declaration(
    SPAN,
    kind,
    declarations,
    Modifiers::empty(),
))
```

We should be able to write this:

```rust
let kind = VariableDeclarationKind::Let;
let ident = BindingIdentifier::with_span(span, name.clone());
let pattern_kind = self.ctx.ast.binding_pattern_identifier(ident);
let binding = self.ctx.ast.binding_pattern(pattern_kind, None, false);
VariableDeclaration::with_span(
    span,
    kind,
    VariableDeclaration::with_span(span, kind, binding, None, false).into_in(self.alloc),
    Modifiers::empty(),
).into_in(self.alloc)
```

Notice that I've called `with_span` constructor, For the transformer/minifer and other stuff we can use `new` which would use the default `SPAN`.
Plus note that as a bonus because of having the conversion traits we no longer need a special approach for stuff like wrapping your items in `new_vec_single`, We can hide all of this verbosity behind the trait calls.

Most of the time you don't need to know how type A is assigned to type B, That's the implementation detail!

Boshen:
For completeness I pick `2. More methods`. The extra amount of code it gonna generate are just all enum variants, which is like 100 - 200 methods?

For constructors, I want to have everything go through the builder to make sure AST nodes will be possible (even just for experimentation) in the future. The builder will have a counter, and each builder call will increment the counter, which is really nice!

Boshen:
```rust
fn block_statement(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
) -> Statement<'a> {
    Statement::BlockStatement(self.alloc(self.block_statement(span, body)))
}

fn block_statement_node(
    self,
    span: Span,
    body: Vec<'a, Statement<'a>>,
) -> BlockStatement<'a> {
    BlockStatement {
        span,
        body,
        scope_id: Default::default(),
    }
}
```

rzvxa:
@Boshen I'm back at this PR,

Quick question about the builder,

Since `block_statement_node` might be confusing because of its similarity to `AstNode`,
What do you think about this semantic to hide the enum variants:

```rust
self.ast.statement().block_statement(span, value);
```

This way we won't need to name our enum factory functions like this `statement_block_statement`.
So instead of this:

```rust
let block = self.ast.block_statement(span, body);
let block_stmt = self.ast.statement_block_statement(block);
```

We can do this:

```rust
let block_stmt = self.ast.statement().block_statement(span, body);
```
And if we want to wrap an already exiting block in a statement we can do this:
```rust
let block = self.ast.block_statement(span, body);
// ...
// somewhere else
let stmt = self.ast.statement().from_block_statement(block);
```

Boshen:
```
let stmt = self.ast.statement().from_block_statement(block);
```

This indirection seems unintuitive to me 🤔

rzvxa:
> ```
> let stmt = self.ast.statement().from_block_statement(block);
> ```
> 
> This indirection seems unintuitive to me 🤔

You are most probably right, That's why I went with what we talked about earlier. We may want to iterate on the builder later on but I'm unsure how we can make it more ergonomic.

Similar to visitors I'll make it to work as is then spend some time on upcoming PRs trying to improve the semantics/performance of it.

This morning I was studying [ast_snippert.rs](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_oxc_utils/src/ast_snippet.rs) from the rolldown team, I think we would be better off with a FromIn trait implemented for some stuff.
With that we can accept `IntoIn<Atom<'a>>` from `&str`, and a bunch of other sources, We can also easily pass in a box or unboxed variant and have it boxed in the builder or any other place with access to allocator regardless of its previous state.

We can also have a trait called `WithSpan` that would set the span for a given node, This way for transformers we can call the builder like so:

```rust
self.ast.statement_block_statement(body);
```

While we call it like this in the parser:

```rust
self.ast.statement_block_statement(body).with_span(span);
```

We can make it into a fully-fledged builder pattern but I think that would be overkill.

#### I'm thinking out loud so people can see my thought process and give feedback.

Boshen (commented):
EPIC!

rzvxa:
@Boshen I'd appreciate it if you let me migrate the rolldown codebase. I was on edge about adding some methods because we already generate quite a bit of code but if rolldown needs those I'll change my decision.

Boshen:
> @Boshen I'd appreciate it if you let me migrate the rolldown codebase. I was on edge about adding some methods because we already generate quite a bit of code but if rolldown needs those I'll change my decision.

👍 Would be helpful if we reduce boilerplate in Rolldown, but this is just a stretch goal.

crates/oxc_ast/src/ast_builder_impl.rs
```diff
@@ -0,0 +1,219 @@
...
+            Span::default(),
+            VariableDeclarationKind::Var,
+            self.new_vec(),
+            false,
+        );
+        let empty_decl = Declaration::VariableDeclaration(self.alloc(empty_decl));
+        mem::replace(decl, empty_decl)
+    }
+
+    /* ---------- Constructors ---------- */
```
Boshen, line 110: I guess all code below this will become "templates" or "snippets" in the near future.

Boshen:
We may end up not needing `alloc_xxx` methods at all due to limited usage and call sites, I'll let you decide once you hop around in the codebase.

Dunqing (approved):
Huge work! We need to merge it as soon as possible to avoid conflicts.

crates/oxc_ast/src/ast_builder_impl.rs
rzvxa: Yes, these are just convenient special methods that are too useful to get rid of them at this stage.
The memory manipulation methods are here to stay, We want to put more care into writing/documenting them.

rzvxa:
> > @Boshen I'd appreciate it if you let me migrate the rolldown codebase. I was on edge about adding some methods because we already generate quite a bit of code but if rolldown needs those I'll change my decision.
> 
> 👍 Would be helpful if we reduce boilerplate in Rolldown, but this is just a stretch goal.
> 
> But yes, we need a follow PR in PR once this gets released.

After this release, a cleanup on their side is recommended if not necessary, Since they've rolled their own `IntoIn` trait which now can be replaced with the `oxc_allocator::IntoIn`. We have to go through all of their usage of the `ast_builder` anyway so I think it could be a good time to also get rid of this obsolete trait, And redundant calls to `AstBuilder::atom` and `AstBuilder::new_str` as they now are constructible from strings and string slices.

It can also be beneficial as I'd be able to use that PR as a means to familiarize myself with our downstream usages.

rzvxa:
> We can merge once @overlookmotel finishes the review, along with a branch in a rolldown fork with oxc pointing to this branch.

It is based on #4124, Give it a look and let me know if there is anything I should consider changing.
https://github.com/rzvxa/rolldown/tree/migrate-to-new-builder

overlookmotel (approved):
I do have some thoughts/potential changes to propose, but given the size of the PR, let's get it merged ASAP, and then iterate.

PR #3894, opened 2024-06-25
perf(parser): add a few more inline hints to cursor functions

Boshen:


crates/oxc_parser/src/cursor.rs
```diff
@@ -113,6 +123,7 @@ impl<'a> ParserImpl<'a> {
 
     /// Move to the next token
     /// Checks if the current token is escaped if it is a keyword
+    #[inline]
```
Boshen, line 126:
```suggestion
```

PR #3896, opened 2024-06-25
feat: oxc transform binding

### LINKED ISSUE #3877: Transform binding should including oxc_transform, like esbuild `tranform` api.

underfin:
close https://github.com/oxc-project/oxc/issues/3877


DonIsaac:
Can we please have the napi SourceMap bindings conform to the contract used by [@jridgewell/trace-mapping](https://github.com/jridgewell/trace-mapping/blob/5a658b10d9b6dea9c614ff545ca9c4df895fee9e/src/types.ts#L14)? I need this for Jest compatibility.

underfin:
> Can we please have the napi SourceMap bindings conform to the contract used by [@jridgewell/trace-mapping](https://github.com/jridgewell/trace-mapping/blob/5a658b10d9b6dea9c614ff545ca9c4df895fee9e/src/types.ts#L14)? I need this for Jest compatibility.

Here has some different fields typing, but it is ok if using typing cast. Because the js ecosystem `SourceMap` typing is not same, oxc can't let everyone's happy, it is difficult to use directly, I hasn't a good idea for it, if you have some advice, we could discuss at here.

PR #3897, opened 2024-06-25
perf(diagnostics): optimize string-buffer reallocations

lucab:
This improves `GraphicalReportHandler` logic by using a string-buffer with a capacity hint. That avoids growing the buffer from zero each time, saving a bunch of reallocations.

### DIFF
--- a/crates/oxc_diagnostics/src/graphic_reporter.rs
+++ b/crates/oxc_diagnostics/src/graphic_reporter.rs
@@ -1122,9 +1122,9 @@ impl GraphicalReportHandler {
         let mut column = context_data.column();
         let mut offset = context_data.span().offset();
         let mut line_offset = offset;
+        let mut line_str = String::with_capacity(context.len());
+        let mut lines = Vec::with_capacity(1);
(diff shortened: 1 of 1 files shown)


lucab:
For context, those buffer resizes were visible in memory profiles of `oxlint`.
When doing a default linting run on https://github.com/microsoft/vscode, they resulted in ~5k reallocs:
allocs

Boshen:
Nice! You can also send a PR to `miette` as well!

Boshen:
Quick question: how did you discover this?

lucab:
@Boshen just having a look at the memory usage pattern of `oxlint`. The tool I'm using is https://github.com/KDE/heaptrack. The "temporary allocations" bucket shows reallocations and allocations directly followed by deallocation.

Boshen:
Thank you! TIL heaptrack

Feel free to raise an issue if you find anything else that is interesting.

And if you wanna have fun with the memory allocator, we have a backlog here https://github.com/oxc-project/backlog/issues/49

PR #3899, opened 2024-06-25
refactor(ast)!: revert adding `span` field to the `BindingPattern` type.

rzvxa:
Since this is a temporary solution in the time that we are waiting for the `#[span]` hint, And there are already other workarounds used in our `ast_codegen` I propose removing it right away - sorry in my opinion adding it in the first place was a mistake - in favor of adding an edge case in the codegen. It is better to do the refactoring in the codegen instead of the production code which people may depend on.


rzvxa:
* **#3899** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):

PR #3902, opened 2024-06-25
feat(sourcemap): impl `std::fmt::Display` for `Error`

DonIsaac:


crates/oxc_sourcemap/src/error.rs
```diff
@@ -1,3 +1,5 @@
+use std::fmt;
+
 #[derive(Debug)]
 pub enum Error {
```
rzvxa, line 4: Maybe we should implement the [Error](https://doc.rust-lang.org/std/error/trait.Error.html) trait as well, Just for the good measure.
Since we already have both the `Debug` and `Display` traits implemented; An empty `Error` implementation like this would be more than sufficient to standardize our error type.
```rust
impl std::error:Error for Error {}
```

rzvxa: FYI, It is especially useful for the end users who use `Result<T, Box<dyn Error>>` or [anyhow](https://github.com/dtolnay/anyhow/blob/eb976a4f53cd7db16dd84dd5e56690780fb3dcf0/src/error.rs#L30) to catch their errors.

DonIsaac: Done

rzvxa (approved):
Thanks for the change, LGTM!

PR #3932, opened 2024-06-26
improvement(ast_codegen): better output formatting.

rzvxa:
We use prettyplease for formatting our generated code, Which has a lot of advantages over rustfmt, But sometimes it conflicts with our cargo fmt config,
Since it isn't configurable I just added a `cargo fmt` step after the generation process to prevent conflicts.
We might be able to not rely on prettyplease at all, But We have to test that one after we have most of our generated code, rustfmt usually bails on nested calls and macros so it might not work as our only solution. In contrast to that prettyplease is used in both bindgen and cargo-expand and is created for this exact reason(formatting generated code).

### DIFF
--- a/tasks/ast_codegen/src/fmt.rs
+++ b/tasks/ast_codegen/src/fmt.rs
@@ -1,13 +1,22 @@
+use std::{
+    path::PathBuf,
+    process::{Command, Stdio},
+};
+
 use lazy_static::lazy_static;
 use proc_macro2::TokenStream;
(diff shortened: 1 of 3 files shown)


rzvxa:
* **#3932** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #3933, opened 2024-06-26
chore(napi/transform): support wasm build

Dunqing:
~~Encountered the same problem as https://github.com/oxc-project/oxc-resolver/issues/190. Fixed in napi@2.16.7, but oxc use `3.x`~~

Fixed in napi@3.0.0-alpha.3


Dunqing:
* **#3933** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/actions/pnpm/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Pnpm
```
Boshen, line 1: This is not needed, can inline this.

Boshen (changes requested):

.github/actions/pnpm/action.yml
Boshen: Or maybe refactor the other scripts later?

Dunqing: Yes, refactor later

PR #3948, opened 2024-06-27
feat(parser): parse modifiers with `parse_modifiers`

### LINKED ISSUE #3929: parser: finish parsing of `Modifiers`
https://github.com/oxc-project/oxc/blob/main/crates/oxc_parser/src/modifiers.rs

DonIsaac:
Closes #3929

### DIFF
--- a/crates/oxc_parser/src/diagnostics.rs
+++ b/crates/oxc_parser/src/diagnostics.rs
@@ -1,6 +1,8 @@
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_span::Span;
 
+use crate::modifiers::Modifier;
+
 #[cold]
 pub fn redeclaration(x0: &str, declare_span: Span, redeclare_span: Span) -> OxcDiagnostic {
     OxcDiagnostic::error(format!("Identifier `{x0}` has already been declared")).with_labels([
@@ -399,7 +401,43 @@ pub fn jsx_element_no_match(span0: Span, span1: Span, name: &str) -> OxcDiagnost
         .with_labels([span0, span1])
 }
 
+// ================================= MODIFIERS =================================
+
 #[cold]
-pub fn modifier_cannot_be_used_here(span: Span, name: &str) -> OxcDiagnostic {
-    OxcDiagnostic::error(format!("'{name}' modifier cannot be used here.")).with_label(span)
+pub fn modifier_cannot_be_used_here(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!("'{}' modifier cannot be used here.", modifier.kind))
+        .with_label(modifier.span)
+}
+
+/// TS(1030)
+#[cold]
+pub fn modifier_already_seen(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!("TS1030: '{}' modifier already seen.", modifier.kind))
+        .with_label(modifier.span)
+        .with_help("Remove the duplicate modifier.")
+}
+
+/// TS(1273)
+#[cold]
+pub fn cannot_appear_on_a_type_parameter(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!(
+        "'{}' modifier cannot be used on a type parameter.",
+        modifier.kind
+    ))
+    .with_label(modifier.span)
+}
+/// TS(1090)
+pub fn cannot_appear_on_a_parameter(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!(
+        "TS1090: '{}' modifier cannot appear on a parameter.",
+        modifier.kind
+    ))
+    .with_label(modifier.span)
+}
+
+/// TS(18010)
+#[cold]
+pub fn accessibility_modifier_on_private_property(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error("An accessibility modifier cannot be used with a private identifier.")
+        .with_label(modifier.span)
 }
--- a/crates/oxc_parser/src/js/class.rs
+++ b/crates/oxc_parser/src/js/class.rs
@@ -8,7 +8,7 @@ use crate::{
     diagnostics,
     lexer::Kind,
     list::NormalList,
-    modifiers::{ModifierKind, Modifiers},
+    modifiers::{ModifierFlags, ModifierKind, Modifiers},
     Context, ParserImpl, StatementContext,
 };
 
@@ -25,7 +25,11 @@ impl<'a> ParserImpl<'a> {
         stmt_ctx: StatementContext,
         start_span: Span,
     ) -> Result<Statement<'a>> {
-        let decl = self.parse_class_declaration(start_span, &Modifiers::empty())?;
+        let modifiers = self.parse_modifiers(
+            /* allow_decorators */ true, /* permit_const_as_modifier */ false,
+            /* stop_on_start_of_class_static_block */ true,
+        );
+        let decl = self.parse_class_declaration(start_span, &modifiers)?;
 
         if stmt_ctx.is_single_statement() {
             self.error(diagnostics::class_declaration(Span::new(
@@ -86,14 +90,11 @@ impl<'a> ParserImpl<'a> {
         }
         let body = self.parse_class_body()?;
 
-        for modifier in modifiers.iter() {
-            if !matches!(modifier.kind, ModifierKind::Declare | ModifierKind::Abstract) {
-                self.error(diagnostics::modifier_cannot_be_used_here(
-                    modifier.span,
-                    modifier.kind.as_str(),
-                ));
-            }
-        }
+        self.verify_modifiers(
+            modifiers,
+            ModifierFlags::DECLARE | ModifierFlags::ABSTRACT,
+            diagnostics::modifier_cannot_be_used_here,
+        );
 
         Ok(self.ast.class(
             r#type,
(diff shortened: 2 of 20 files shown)


crates/oxc_parser/src/modifiers.rs
```diff
@@ -102,39 +124,125 @@ pub struct Modifier {
...
+    }
+}
+
+impl GetSpan for Modifiers<'_> {
+    fn span(&self) -> Span {
+        let Some(modifiers) = &self.modifiers else { return SPAN };
+        // SAFETY: One of Modifier's invariants is that Some(modifiers) always
+        // contains a non-empty Vec; otherwise it must be `None`.
+        unsafe {
+            use assert_unchecked::assert_unchecked;
```
DonIsaac, line 242: @Boshen is it OK to use `unsafe` here? Or should I use safe code?

crates/oxc_parser/src/modifiers.rs
```diff
@@ -178,20 +286,52 @@ impl ModifierKind {
...
         let mut modifiers = self.ast.new_vec();
         while self.at_modifier() {
             let span = self.start_span();
             let modifier_flag = self.cur_kind().into();
-            flags.set(modifier_flag, true);
             let kind = self.cur_kind();
             self.bump_any();
-            modifiers.push(Self::modifier(kind, self.end_span(span)));
+            let modifier = Self::modifier(kind, self.end_span(span));
+            self.verify_modifier(flags, &modifier);
```
DonIsaac, line 330: I'd rather do this in the checker, but we don't persist modifiers in the AST and so we lose spans

crates/oxc_parser/src/modifiers.rs
Boshen: `debug_assert` should be fine here?

crates/oxc_parser/src/modifiers.rs
```diff
@@ -178,20 +252,52 @@ impl ModifierKind {
         }
     }
 }
+impl From<Kind> for ModifierKind {
+    // NOTE (@DonIsaac) should we be using TryFrom here? or does unreachable!()
```
Boshen, line 256: `unreachable` is too unsafe here, better to return a None, and return unexpected token on the call site.

Need to be extra safe for a parser.

crates/oxc_parser/src/modifiers.rs
```diff
@@ -178,20 +252,52 @@ impl ModifierKind {
...
+
+impl std::fmt::Display for ModifierKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
 
 impl<'a> ParserImpl<'a> {
-    pub(crate) fn eat_modifiers_before_declaration(&mut self) -> (ModifierFlags, Modifiers<'a>) {
+    pub(crate) fn eat_modifiers_before_declaration(&mut self) -> Modifiers<'a> {
```
Boshen, line 287: I'd eventually align all the method names with tsc's parser.ts as well so we know where to look when we try to debug parsing.

crates/oxc_parser/src/modifiers.rs
```diff
@@ -231,55 +337,40 @@ impl<'a> ParserImpl<'a> {
...
+            if modifier.is_static() {
                 has_seen_static_modifier = true;
             }
+            self.check_for_duplicate_modifiers(modifier_flags, &modifier);
+            modifier_flags.set(modifier.kind.into(), true);
             modifiers.push(modifier);
+            has_leading_modifier = true;
         }
 
         // parse trailing decorators, but only if we parsed any leading modifiers
```
Boshen, line 376: I wonder where does trailing decorators occur lol

crates/oxc_parser/src/modifiers.rs
```diff
@@ -289,47 +380,63 @@ impl<'a> ParserImpl<'a> {
...
+        {
             return None;
         }
-        // }
         Some(Self::modifier(kind, self.end_span(span)))
     }
 
+    fn next_token_is_open_brace(&mut self) -> bool {
+        self.bump_any();
+        self.cur_kind() == Kind::LCurly
```
Boshen, line 437: There's a shorthand `self.at(Kind::LCurly)`

Boshen (changes requested):

tasks/transform_conformance/babel.snap.md
```diff
@@ -445,7 +445,9 @@ Passed: 475/927
 * opts/optimizeConstEnums/input.ts
 * opts/rewriteImportExtensions/input.ts
 
-# babel-plugin-transform-typescript (131/151)
+# babel-plugin-transform-typescript (129/151)
+* class/accessor-allowDeclareFields-false/input.ts
```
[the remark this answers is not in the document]
DonIsaac: I'm not sure, those cases are using `private` on private class elements.

```ts
abstract class Foo {
  accessor prop: number = 1;
  static accessor prop2: number = 1;
  accessor #prop3: number = 1;
  accessor [prop4]: number = 1;
  private accessor prop5: number = 1;
  abstract accessor prop6: number;
  private accessor #p: any;

  accessor a!: any;
  accessor aa!: any;
  abstract accessor #s;
  abstract accessor f = 1;
  readonly accessor g;
}
```

crates/oxc_parser/src/modifiers.rs
DonIsaac: I couldn't figure out what exactly was going on with this method so I left it as is, simply changing the return signature...

crates/oxc_parser/src/modifiers.rs
DonIsaac: I think this is b/c of the syntax discrepancy between ts experimental decorators and
the ecmascript proposal?

crates/oxc_parser/src/js/class.rs
```diff
@@ -25,7 +25,8 @@ impl<'a> ParserImpl<'a> {
         stmt_ctx: StatementContext,
         start_span: Span,
     ) -> Result<Statement<'a>> {
-        let decl = self.parse_class_declaration(start_span, &Modifiers::empty())?;
+        let modifiers = self.parse_modifiers(true, false, true);
```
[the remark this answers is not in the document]
DonIsaac: fixed

crates/oxc_parser/src/modifiers.rs
DonIsaac: Fixed

crates/oxc_parser/src/modifiers.rs
Boshen: > I couldn't figure out what exactly was going on with this method

Me neither, I didn't write it. We can get rid of it and port from tsc instead.

crates/oxc_parser/src/js/list.rs
```diff
@@ -265,6 +265,7 @@ impl<'a> SeparatedList<'a> for FormalParameterList<'a> {
         let span = p.start_span();
         p.eat_decorators()?;
 
+        // TODO: use parse_modifiers
```
Boshen, line 268: Better to not have a TODO in the parser.

crates/oxc_parser/src/modifiers.rs
```diff
@@ -76,65 +98,117 @@ impl ModifierFlags {
...
+    /// # Invariants
+    /// `flags` must correctly reflect the [`ModifierKind`]s within
+    ///  `modifiers`. E.g., if `modifiers` is empty, then so is `flags``.
+    #[must_use]
+    fn new(modifiers: Vec<'a, Modifier>, flags: ModifierFlags) -> Self {
+        if modifiers.is_empty() {
+            debug_assert!(flags.is_empty());
+            Self::empty()
+        } else {
+            // todo: loop over `modifiers` and check integrity of `flags`
```
Boshen, line 159: Better to not have a TODO in the parser.

crates/oxc_parser/src/diagnostics.rs
```diff
@@ -399,7 +401,35 @@ pub fn jsx_element_no_match(span0: Span, span1: Span, name: &str) -> OxcDiagnost
...
+#[cold]
+pub fn modifier_cannot_be_used_here(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!("'{}' modifier cannot be used here.", modifier.kind))
+        .with_label(modifier.span)
+}
+
+/// TS(1030)
+#[cold]
+pub fn modifier_already_seen(modifier: &Modifier) -> OxcDiagnostic {
+    OxcDiagnostic::error(format!("'{}' modifier already seen.", modifier.kind))
```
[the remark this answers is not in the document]
DonIsaac: I'm thinking of adding error codes to `OxcDiagnostic`, what do you think?

crates/oxc_parser/src/modifiers.rs
DonIsaac: Ive spent two hours staring at this code, `parse_modifers` and typescript's implementation. When I try to use `parse_modifiers` where this function is consumed, I get an infinite loop over `Kind::Const`. I can't figure it out. Any insights?

crates/oxc_parser/src/js/list.rs
DonIsaac: fixed

Boshen (approved):

crates/oxc_parser/src/modifiers.rs
Boshen: Let's track this in the issue.

crates/oxc_parser/src/diagnostics.rs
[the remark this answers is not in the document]
Boshen: It's unnecessary for us, we don't produce i18n error messages.

PR #3949, opened 2024-06-27
feat(linter): add react/jsx-curly-brace-presence

DonIsaac:
Note that this PR does not implement a fixer, but one is available.


crates/oxc_linter/src/rules/react/jsx_curly_brace_presence.rs
```diff
@@ -0,0 +1,1208 @@
...
+/// <App>{' '}</App>
+/// <App>{'     '}</App>
+/// <App>{/* comment */ <Bpp />}</App> // the comment makes the container necessary
+/// ```
+///
+/// ## When Not To Use It
+///
+/// You should turn this rule off if you are not concerned about maintaining
+/// consistency regarding the use of curly braces in JSX props and/or
+/// children as well as the use of unnecessary JSX expressions.
```
mysteryven, line 273:
```suggestion
```

Need remove this?

crates/oxc_linter/src/rules/react/jsx_curly_brace_presence.rs
```diff
@@ -0,0 +1,1208 @@
...
+
+    [this_container_idx.checked_sub(1), this_container_idx.checked_add(1)]
+        .into_iter()
+        // [prev id, next id] -> [prev node, next node], removing out-of-bounds indices
+        .filter_map(|idx| idx.and_then(|idx| children.get(idx)))
+        .any(oxc_ast::ast::JSXChild::is_expression_container)
+}
+
+#[test]
+fn test_debug() {
```
mysteryven, line 776: Is this need to clean up? 🤔

crates/oxc_linter/src/rules/react/jsx_curly_brace_presence.rs
```diff
@@ -0,0 +1,1208 @@
...
+        ("<App prop={true}>foo</App>", None),
+        ("<App prop>foo</App>", None),
+        (r"<App prop='bar'>{'foo \n bar'}</App>", None),
+        ("<App prop={ ' ' }/>", None),
+        ("<MyComponent prop='bar'>foo</MyComponent>", Some(json!([{ "props": "never" }]))),
+        (r#"<MyComponent prop="bar">foo</MyComponent>"#, Some(json!([{ "props": "never" }]))),
+        ("<MyComponent>foo</MyComponent>", Some(json!([{ "children": "never" }]))),
+        (r#"<MyComponent>{<App/>}{"123"}</MyComponent>"#, Some(json!([{ "children": "never" }]))),
+        // NOTE: causes parser to panic
+        // (r#"<App>{"foo 'bar' "foo" bar"}</App>"#, Some(json!([{ "children": "never" }]))),
```
mysteryven, line 873: Looks like this case should be: `<App>{"foo 'bar' \"foo\" bar"}</App>`

mysteryven (approved):

PR #3953, opened 2024-06-28
chore: bump napi 3.0.0-alpha.3

underfin:
Keep version same as rolldown, avoid release build error.

@Boshen Could you release the `oxc_transform_napi` library to crate.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,8 +100,8 @@ oxc_prettier     = { path = "crates/oxc_prettier" }
 oxc_tasks_common = { path = "tasks/common" }
 oxc_ast_codegen  = { path = "tasks/oxc_ast_codegen" }
 
-napi        = "3.0.0-alpha.1"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -100,8 +100,8 @@ oxc_prettier     = { path = "crates/oxc_prettier" }
 oxc_tasks_common = { path = "tasks/common" }
 oxc_ast_codegen  = { path = "tasks/oxc_ast_codegen" }
 
-napi        = "3.0.0-alpha.1"
-napi-derive = "3.0.0-alpha.1"
+napi        = "3.0.0-alpha.3"
+napi-derive = "3.0.0-alpha.2"
```
DonIsaac, line 104: should this be `3.0.0-alpha.3` as well?

Brooooooklyn: napi-derive alpha.3 doesn't release yet

Issue #3887, opened 2024-06-24
Replace `NormalList` and `SeparatedList` trait implementation with `parseDelimitedList` and `parseList` 


Boshen:
I find `parseDelimitedList` and `parseList` from typescript's parser.ts a lot nicer than our trait implementations, our version is too boilerplate heavy.

PR #3962, opened 2024-06-28
up(parser): gracefully handle illegal list separators

DonIsaac:
Continue parsing when an illegal list separator is found when parsing a `SeparatedList`. Illegal but handle-able separators are specified via `illegal_separator()`. This defaults to `;`, but can be changed or disabled by returning `None`.

For example:
```ts
const x = {
  a: 1;
  b: 2
}
```

instead of panicking, the parser will produce a valid AST and still report the illegal `;` in the object literal expression.

### DIFF
--- a/crates/oxc_parser/src/list.rs
+++ b/crates/oxc_parser/src/list.rs
@@ -42,6 +42,13 @@ pub trait SeparatedList<'a>: Sized {
         Kind::Comma
     }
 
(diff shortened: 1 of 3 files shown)


Boshen:
Will probably improve list parsing at some point: https://github.com/oxc-project/oxc/issues/3887

Boshen:
This PR bloats the code too much. Let's defer the change and rewrite the whole thing instead.

Issue #3289, opened 2024-05-15
Change type of padding in lexer `Token`?


overlookmotel:
#3283 showed a mysterious 5% speed-up on lexer benchmarks (see https://github.com/oxc-project/oxc/pull/3283#issuecomment-2111835472).

It's unclear what caused this, as the changes to the lexer in that PR actually make the lexer do *more* work, not less.

Only explanation I can see is that for some reason, replacing one of the `u8` padding bytes in `Token` with a `bool` produced better codegen somehow (the mysteries of the compiler!)

We could try converting the other 4 padding bytes `u32` to bools too `[bool; 4]`, and see if that makes a difference.

Ideally we'd use a type which has no "active bits" at all instead of `bool`, so compiler can know it can leave these fields uninitialized. A type which is like `()`, but with size 1 instead of size 0. But does such a type exist? Generally such a type would be completely useless!

overlookmotel:
Actually I think I may have figured it out!

Changing the field from `u8` to `bool` altered the type layout.

After #3283:

```rs
#[derive(Debug, Clone, Copy, Default)]
pub struct Token {
    pub kind: Kind,
    pub start: u32,
    pub end: u32,
    pub is_on_new_line: bool,
    pub escaped: bool,
    has_separator: bool,
    _padding2: u32,
}

const _: () = {
    assert!(std::mem::offset_of!(Token, start) == 0);
    assert!(std::mem::offset_of!(Token, end) == 4);
    assert!(std::mem::offset_of!(Token, kind) == 12);
    assert!(std::mem::offset_of!(Token, is_on_new_line) == 13);
    assert!(std::mem::offset_of!(Token, escaped) == 14);
    assert!(std::mem::offset_of!(Token, has_separator) == 15);
    assert!(std::mem::offset_of!(Token, _padding2) == 8);
};
```

Whereas before:

```rs
#[derive(Debug, Clone, Copy, Default)]
pub struct TokenBefore {
    pub kind: Kind,
    pub start: u32,
    pub end: u32,
    pub is_on_new_line: bool,
    pub escaped: bool,
    _padding: u8,
    _padding2: u32,
}

const _: () = {
    assert!(std::mem::offset_of!(TokenBefore, start) == 0);
    assert!(std::mem::offset_of!(TokenBefore, end) == 4);
    assert!(std::mem::offset_of!(TokenBefore, kind) == 13);
    assert!(std::mem::offset_of!(TokenBefore, is_on_new_line) == 14);
[...]

PR #3964, opened 2024-06-29
perf(lexer): move padding to the end of `Token` struct

DonIsaac:
Experiment; follows up on #3289.

Before:
image

After:
image

Maybe this will prevent a shift instruction when accessing `kind`?

### DIFF
--- a/crates/oxc_parser/src/lexer/token.rs
+++ b/crates/oxc_parser/src/lexer/token.rs
@@ -13,6 +13,11 @@ pub struct Token {
     /// End offset in source
     pub end: u32,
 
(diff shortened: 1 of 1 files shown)


DonIsaac:
Maybe it's _better_ for `_padding` to be in the middle?
cc: @overlookmotel

Boshen:
```
pub struct Token {
    /// Start offset in source
    pub start: u32,

    /// End offset in source
    pub end: u32,

    /// Token Kind
    pub kind: Kind,
```

Regressed lexer by 8%.

Boshen:
No performance improvements found.

PR #3971, opened 2024-06-29
feat(linter): implement @typescript-eslint/no-dynamic-delete

kaykdm:

original implementation

- code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-dynamic-delete.ts
- test: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/no-dynamic-delete.test.ts
- doc: https://typescript-eslint.io/rules/no-dynamic-delete/

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
+++ b/crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
@@ -1,0 +1,164 @@
+use oxc_ast::{ast::Expression, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use oxc_syntax::operator::UnaryOperator;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Default, Clone)]
+pub struct NoDynamicDelete;
+
+declare_oxc_lint!(
+    /// ### What it does
+    /// Disallow using the delete operator on computed key expressions.
+    ///
+    /// ### Why is this bad?
+    /// Deleting dynamically computed keys can be dangerous and in some cases not well optimized.
+    /// Using the delete operator on keys that aren't runtime constants could be a sign that you're using the wrong data structures.
+    /// Consider using a Map or Set if you’re using an object as a key-value collection.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// const container: { [i: string]: 0 } = {};
+    /// delete container['aa' + 'b'];
+    /// ```
+    NoDynamicDelete,
+    restriction,
+);
+
+fn no_dynamic_delete_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn(
+        "typescript-eslint(no-dynamic-delete): Do not delete dynamically computed property keys.",
+    )
+    .with_help("Disallow using the `delete` operator on computed key expressions")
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/snapshots/no_dynamic_delete.snap
```diff
@@ -0,0 +1,92 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ typescript-eslint(no-dynamic-delete): Do not delete dynamically computed property keys.
+   ╭─[no_dynamic_delete.tsx:3:17]
+ 2 │             const container: { [i: string]: 0 } = {};
+ 3 │             delete container['aa' + 'b'];
```
DonIsaac, line 7: consider covering either the entire statement or just the computed key.

crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
```diff
@@ -0,0 +1,162 @@
...
+        "typescript-eslint(no-dynamic-delete): Do not delete dynamically computed property keys.",
+    )
+    .with_help("Disallow using the `delete` operator on computed key expressions")
+    .with_label(span0)
+}
+
+impl Rule for NoDynamicDelete {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::UnaryExpression(expr) = node.kind() else { return };
+        if expr.operator != UnaryOperator::Delete {
```
DonIsaac, line 41: prefer `matches!` to remove a function call dispatch (`matches!` is a const macro`)

crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
```diff
@@ -0,0 +1,162 @@
...
+        	const value = 1;
+        	delete value;
+        	    ",
+        "
+        	const value = 1;
+        	delete -value;
+        	    ",
+        "
+        	const container: { [i: string]: 0 } = {};
+        	delete container['aaa'];
```
DonIsaac, line 103: What happens if you do
```ts
// this?
delete container[('aaa')]
// or this?
delete container['aaa' + 'bbb']
```
Should we allow these cases?

crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
kaykdm: fixed it!

crates/oxc_linter/src/snapshots/no_dynamic_delete.snap
kaykdm: fixed it by covering the entire statement

crates/oxc_linter/src/rules/typescript/no_dynamic_delete.rs
kaykdm: yeah you're right. `delete container[('aaa')]` should be allowed.
fixed it by using get_inner_expression function.

It seems like `delete container['aaa' + 'bbb']` should NOT be allowed. There is an invalid test case for this: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/no-dynamic-delete.test.ts#L64-L71

Boshen (approved):

PR #3972, opened 2024-06-29
perf(semantic): use `Atom<'a>` for `Reference`s

DonIsaac:
Relates to [this issue](https://github.com/oxc-project/backlog/issues/31) on the backlog.

PR #3974, opened 2024-06-29
Revert "perf(semantic): use `Atom<'a>` for `Reference`s"

Boshen:
Reverts oxc-project/oxc#3972

@DonIsaac As it turns out we can't have lifetimes on these structures because semantic data is exposed to downstream users to use freely, detached from the ast and allocator.


DonIsaac:
Does adding lifetimes break downstream consumers?

Boshen:
> Does adding lifetimes break downstream consumers?

Yes, I forgot to mention Rolldown. https://github.com/rolldown/rolldown/blob/87ec563e8bc1a2241d80a40452c8dc7c6c59ad06/crates/rolldown_common/src/module/normal_module.rs#L42

PR #3976, opened 2024-06-29
Various Isolated declarations fixes

escaton:
I'm playing with "isolated-declarations" in a project with 30k files.
These are fixes I did while attempting to get a successful typecheck.

For now, there is at least one test for each case, and I plan to write more later, but I think I'll need help to make them better.

### DIFF
--- a/crates/oxc_isolated_declarations/src/class.rs
+++ b/crates/oxc_isolated_declarations/src/class.rs
@@ -66,7 +66,11 @@ impl<'a> IsolatedDeclarations<'a> {
                         .value
                         .as_ref()
                         .and_then(|expr| {
-                            let ts_type = self.infer_type_from_expression(expr);
+                            let ts_type = if property.readonly {
+                                self.transform_expression_to_ts_type(expr)
+                            } else {
+                                self.infer_type_from_expression(expr)
+                            };
                             if ts_type.is_none() {
                                 self.error(property_must_have_explicit_type(property.key.span()));
                             }
@@ -75,6 +79,9 @@ impl<'a> IsolatedDeclarations<'a> {
                         .map(|ts_type| self.ast.ts_type_annotation(SPAN, ts_type))
                 })
         };
+
+        // TODO if inferred type_annotations is TSLiteral, it should stand as value,
+        // so `field = 'string'` remain `field = 'string'` instead of `field: 'string'`
 
         self.ast.class_property(
             property.r#type,
--- a/crates/oxc_isolated_declarations/src/class.rs
+++ b/crates/oxc_isolated_declarations/src/class.rs
@@ -324,18 +324,18 @@ impl<'a> IsolatedDeclarations<'a> {
 
         let mut has_private_key = false;
         let mut elements = self.ast.new_vec();
-        let mut is_function_overloads = false;
+        // let mut is_function_overloads = false;
         for element in &decl.body.body {
             match element {
                 ClassElement::StaticBlock(_) => {}
                 ClassElement::MethodDefinition(ref method) => {
-                    if method.value.body.is_none() {
-                        is_function_overloads = true;
-                    } else if is_function_overloads {
-                        // Skip implementation of function overloads
-                        is_function_overloads = false;
-                        continue;
-                    }
+                    // if method.value.body.is_none() {
+                    //     is_function_overloads = true;
+                    // } else if is_function_overloads {
+                    //     // Skip implementation of function overloads
+                    //     is_function_overloads = false;
+                    //     continue;
+                    // }
                     if method.key.is_private_identifier() {
                         has_private_key = true;
                         continue;
--- a/crates/oxc_isolated_declarations/src/diagnostics.rs
+++ b/crates/oxc_isolated_declarations/src/diagnostics.rs
@@ -139,3 +139,8 @@ pub fn type_containing_private_name(name: &str, span: Span) -> OxcDiagnostic {
     ))
     .with_label(span)
 }
+
+pub fn destructuring_export(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::error("TS9999: Not supported yet")
+        .with_label(span)
+}
--- a/crates/oxc_isolated_declarations/tests/deno/mod.rs
+++ b/crates/oxc_isolated_declarations/tests/deno/mod.rs
@@ -389,12 +389,18 @@ export function foo(a: any): number {
 
     #[test]
     fn dts_class_type_param() {
-        transform_dts_test("export class A<T> extends B<T> implements C<T>{};", "export declare class A<T> extends B<T> implements C<T> {}");
+        transform_dts_test(
+            "export class A<T> extends B<T> implements C<T>{};",
+            "export declare class A<T> extends B<T> implements C<T> {}",
+        );
     }
 
     #[test]
     fn dts_fn_type_param() {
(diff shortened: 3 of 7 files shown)


crates/oxc_isolated_declarations/src/class.rs
```diff
@@ -324,18 +331,18 @@ impl<'a> IsolatedDeclarations<'a> {
...
                 ClassElement::StaticBlock(_) => {}
                 ClassElement::MethodDefinition(ref method) => {
-                    if method.value.body.is_none() {
-                        is_function_overloads = true;
-                    } else if is_function_overloads {
-                        // Skip implementation of function overloads
-                        is_function_overloads = false;
-                        continue;
-                    }
+                    // if method.value.body.is_none() {
```
escaton, line 339: TODO either fix, or uncomment

crates/oxc_isolated_declarations/tests/deno/mod.rs
```diff
@@ -16,9 +16,13 @@ mod tests {
 
         let ret = IsolatedDeclarations::new(&allocator).build(&program);
         let actual = CodeGenerator::new().build(&ret.program).source_text;
-        let expected_program = Parser::new(&allocator, expected, source_type).parse().program;
-        let expected = CodeGenerator::new().build(&expected_program).source_text;
+        // let expected_program = Parser::new(&allocator, expected, source_type).parse().program;
```
escaton, line 19: TODO revert

crates/oxc_isolated_declarations/src/diagnostics.rs
```diff
@@ -139,3 +139,8 @@ pub fn type_containing_private_name(name: &str, span: Span) -> OxcDiagnostic {
     ))
     .with_label(span)
 }
+
+pub fn destructuring_export(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::error("TS9999: Not supported yet")
```
escaton, line 144: I wanted to give feedback, that a certain transformation is not supported yet, instead of silently generating wrong `.d.ts`

escaton:
@Boshen hi!
I'm not sure is you already addressed some of these issues, so let me know if you think this PR can be useful. Of course after some polish.

Boshen:
Thank you in advance. I have copied over the codegen changes in https://app.graphite.dev/github/pr/oxc-project/oxc/3980 so I can make a quicker release.

Boshen:
> @Boshen hi! I'm not sure if you already addressed some of these issues, so let me know if you think this PR can be useful. Of course after some polish.

cc @Dunqing

crates/oxc_isolated_declarations/src/diagnostics.rs
Dunqing: I agree, can you provide the examples that need to report this error?

Dunqing: The `TSXXXX` error code is identical to the error reported by typescript. If you want to report an error that typescript doesn't have, you don't need to add `TSXXXX`.

Dunqing:
Thank you for working on this! If you need to add test, please add in [fixtures](https://github.com/oxc-project/oxc/blob/main/crates/oxc_isolated_declarations/tests/fixtures). The `deno.rs` is copy from https://github.com/denoland/deno_graph/blob/main/src/fast_check/transform_dts.rs#L932-#L1532

crates/oxc_isolated_declarations/src/diagnostics.rs
escaton: > can you provide the examples that need to report this error?

1. I added it [here](https://github.com/escaton/oxc/blob/bc09ef80aa5830f65bff90396bb3571efa2de25f/crates/oxc_isolated_declarations/src/declaration.rs#L59) because `return None;` leads to `export const {a} = b()` is transformed to `export declare const;` instead of `export declare const a: unknown;`.
2. I also considered adding it [here](https://github.com/escaton/oxc/blob/8fd3ae99af82d1bc107f9e7a3ee42204b2365df9/crates/oxc_isolated_declarations/src/class.rs#L344) because with current `// Skip implementation of function overloads` it generates incorrect code, and I don't have ideas how to fix it right away.

Dunqing: > > can you provide the examples that need to report this error?
> 
> 1. I added it [here](https://github.com/escaton/oxc/blob/bc09ef80aa5830f65bff90396bb3571efa2de25f/crates/oxc_isolated_declarations/src/declaration.rs#L59) because `return None;` causes `export const {a} = b()` to be transformed into `export declare const;` instead of `export declare const a: unknown;`.

This is a bug. Our goal is to keep the output the same as TypeScript, including diagnostics. if it's not the same, it's considered a bug. So we should report `binding_element_export` here

> 2. I also considered adding it [here](https://github.com/escaton/oxc/blob/8fd3ae99af82d1bc107f9e7a3ee42204b2365df9/crates/oxc_isolated_declarations/src/class.rs#L344) because with current `// Skip implementation of function overloads` it generates incorrect code, and I don't have ideas how to fix it right away.
> 
> Methods in abstract classes could be empty, and it is recognized as overloading:
> 
> ```
> export abstract class A {
>   b(): void;
>   c(): void {};
>   d(): void {}
> }
> ```
> 
> ->
> 
> ```
> export abstract class A {
>   b(): void;
>   d(): void {}
> }
> ```

This code seems to be incorrect. [TypeScript Playground](https://www.typescriptlang.org/play/?isolatedDeclarations=true#code/KYDwDg9gTgLgBAQwEYGcZQQY3pgNglFOAQTgG8AoOapACgEoAuOANwgEsATAbiuswbM2XcgF9e1OJ0GsOnMRVEUgA). We assume that the transforms code is correct

Dunqing: The first case is fixed in #3976, The second case has a similar [issue](https://github.com/oxc-project/oxc/pull/4019), and fixed in #4024

Dunqing:
All known bugs are about to be fixed. Please Feel free to send an RP fix if you find other bugs.

crates/oxc_isolated_declarations/src/diagnostics.rs
escaton: fixed in https://github.com/oxc-project/oxc/pull/4025, right?

Dunqing: Yes!!

escaton:
I think these commits are still actual
- [fix inferring literal types for readonly calss fileds](https://github.com/oxc-project/oxc/pull/3976/commits/8fd3ae99af82d1bc107f9e7a3ee42204b2365df9)
- [fix global declarations with references](https://github.com/oxc-project/oxc/pull/3976/commits/0cc038b6b9845178f916c307cab0d3310cfaabe0)

I'll create another PR for them with proper tests

Dunqing:
> @Dunqing I think these commits are still actual
> 
> * [fix inferring literal types for readonly calss fileds](https://github.com/oxc-project/oxc/pull/3976/commits/8fd3ae99af82d1bc107f9e7a3ee42204b2365df9)
> * [fix global declarations with references](https://github.com/oxc-project/oxc/pull/3976/commits/0cc038b6b9845178f916c307cab0d3310cfaabe0)
> 
> I'll create another PR for them with proper tests

Yes, go ahead!

PR #3982, opened 2024-06-30
fix(isolated-declarations): if declarations is referenced in `declare global` then keep it

### LINKED ISSUE #3981: isolatedDeclaration: The related types inside `declare global {}` should also be kept in the result even if they are not exported
For example:

```typescript
function MyFunction() {
  return 'here is my function'
}

function MyOtherFunction() {
  return 'here is my other function'
}

export { MyOtherFunction }

declare global {
  interface Window {
    MyFunction: typeof MyFunction
  }
}

window.MyFunction = MyFunction
```

MyFunction is not exported but is attached to the global `window` object, so the MyFunction types should be included in the `d.ts` file even if it is not exported.

The following are the results of the comparison using tsc and oxc-transform:

```typescript
// by tsc
declare function MyFunction(): string;
declare function MyOtherFunction(): string;
export { MyOtherFunction };
declare global {
    interface Window {
        MyFunction: typeof MyFunction;
    }
}
```

```typescript
// by oxc-transform
declare function MyOtherFunction(): string;
export { MyOtherFunction };
declare global {
	interface Window {
		MyFunction: typeof MyFunction;
	}
}
```

The file generated by oxc-transform is missing: `declare function MyFunction(): string;`

Dunqing:
close: #3981

### DIFF
--- a/crates/oxc_isolated_declarations/src/lib.rs
+++ b/crates/oxc_isolated_declarations/src/lib.rs
@@ -144,6 +144,7 @@ impl<'a> IsolatedDeclarations<'a> {
                         }
                         Declaration::TSModuleDeclaration(decl) => {
                             if decl.kind.is_global() {
+                                self.scope.visit_ts_module_declaration(decl);
                                 transformed_indexes.push(new_stmts.len());
                             }
                         }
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#3982** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
Seems merging in Graphite cannot auto close linked issues? Can you check it? @Boshen

PR #3984, opened 2024-06-30
feat(linter): add fixer for @typescript-eslint/consistent-type-imports

mysteryven:
This rule's fixer is complicated, I do line by line copy as possible.

https://github.com/typescript-eslint/typescript-eslint/blob/7b13dae347075d467ea95d0f986b38c3048906a2/packages/eslint-plugin/src/rules/consistent-type-imports.ts

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/consistent_type_imports.rs
+++ b/crates/oxc_linter/src/rules/typescript/consistent_type_imports.rs
@@ -1,12 +1,24 @@
-use std::ops::Deref;
+use std::{error::Error, ops::Deref};
 
-use oxc_ast::{ast::ImportDeclarationSpecifier, AstKind};
+use itertools::Itertools;
+use oxc_ast::{
+    ast::{
+        ImportDeclaration, ImportDeclarationSpecifier, ImportDefaultSpecifier,
+        ImportNamespaceSpecifier, ImportSpecifier, Statement,
+    },
+    AstKind,
+};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_semantic::SymbolId;
-use oxc_span::{CompactStr, Span};
+use oxc_span::{CompactStr, GetSpan, Span};
 
-use crate::{context::LintContext, rule::Rule, AstNode};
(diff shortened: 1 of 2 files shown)


mysteryven:
* **#3984** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/rules/typescript/consistent_type_imports.rs
```diff
@@ -1532,1053 +2179,1096 @@ fn test() {
...
+            "
+            import Foo from 'foo';
+            let foo: Foo;
+                      ",
+            "
+            import type Foo from 'foo';
+            let foo: Foo;
+                      ",
+            Some(
+                serde_json::json!([            { "prefer": "type-imports", "fixStyle": "inline-type-imports" },          ]),
```
rzvxa, line 2225: I'd like to review this rule but as you've mentioned it is quite complicated, So depending on how my day goes it might take me a few hours to a day before I'm done with it. Don't let my slow review stop you from merging it if anyone else has the time to leave a review before me.

----------

On another note:
There are some formatting issues in the test cases, You may want to expand your tabs as spaces(that's usually why my code formatting is broken because I don't expand tabs by default and GitHub uses 8 tab size by default as opposed to 4 which I use locally).

mysteryven: Now the test cases in `fix` are formatted. 

> how my day goes it might take me a few hours to a day before I'm done with it.

Good luck to you =.=, and looking forward to your suggestion.

rzvxa (approved):

Boshen:
This is a big ask, but can you test this in https://github.com/oxc-project/oxlint-ecosystem-ci manually? 

The repos are cloned locally, you can check `git diff` after the fix.

More over, I think we should add idempotence test to the fixer: run fix, then run fix agin on the fixed code, the result should be the same.

Boshen:
Merge first, integrate with ecosystem ci later.

PR #3985, opened 2024-06-30
refactor(semantic)!: Do not expose `CompactStr` from `Reference`

DonIsaac:
Changes `Reference::name()` to return a `&str` instead of a `&CompactStr`. As such, it is technically a breaking change. I scanned through Rolldown to see how they use `Reference`s, and I do not think this PR will break anything.

Doing this will give us flexibility to change the representation of a reference name in the future. I'm thinking of making this an `Rc<CompactStr>` or something similar to reduce allocations within semantic.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_global_assign.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_global_assign.rs
@@ -66,7 +66,7 @@ impl Rule for NoGlobalAssign {
                 let reference = symbol_table.get_reference(reference_id);
                 if reference.is_write() {
                     let name = reference.name();
-                    if !self.excludes.contains(name) && ctx.env_contains_var(name) {
+                    if !self.excludes.iter().any(|ex| ex == name) && ctx.env_contains_var(name) {
                         ctx.diagnostic(no_global_assign_diagnostic(name, reference.span()));
                     }
                 }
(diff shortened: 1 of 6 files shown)


Dunqing:
image

It seems to have some impact on performance

overlookmotel:
@DonIsaac Just to say, in my view we shouldn't be using `CompactString` / `CompactStr` at all in `Semantic`. We could just use `Atom` and avoid conversion costs to/from `CompactStr`.

Please see https://github.com/oxc-project/backlog/issues/32 which is a work-in-progress shopping list of various optimizations we could make to semantic. But... my intent is to review all the code which uses `Semantic`, try to think about its design "holistically", and then write up a RFC proposal to revamp it. I hope to do that later this month, or more likely next month.

There's plenty of room for improvement, but in my view, at this point it'll be more fruitful to tackle it as a whole with a defined roadmap, rather than making lots of incremental changes.

However... if you're keen to get busy on this, I'm pretty sure replacing `CompactStr` with `Atom` in semantic will be a good idea. If you'd like to do that, please open an issue for it and make sure it's agreed that's a good way forwards (I think it is, but others may disagree with me) before spending your time doing a PR.

overlookmotel:
Oh I just saw #3972 and #3974 (am catching up on issues in reverse order). Damn! Clearly this needs more thought...

PR #3954, opened 2024-06-28
feat(ast_codegen): add visit generator

rzvxa:
~~The generated code is only here for the sake of my own comparison (instead of manually keeping a backup of the old generated file). I would clean this up as soon as it is ready, submit some parts of it as the down stack, and stack the actual generated code on top of this. So please don't let the huge diff distract you, It won't have many conflicts since almost all of these are the generated visit code, which is completely contained to its own module(other than some minor renaming refactors).~~

The order of function definitions is a bit different, I've used a depth-first search, We can switch to a breadth-first one to align functions more closely to the original.

PR #3994, opened 2024-07-01
refactor(ast): reorder fields to reflect their visit order.

rzvxa:


rzvxa:
* **#3954** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3994** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3994?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@rzvxa I'm late on this one, but a question: Does this new order align with `Visit`'s current visitation order? Or just to "what makes sense"?

rzvxa:
> @rzvxa I'm late on this one, but a question: Does this new order align with `Visit`'s current visitation order? Or just to "what makes sense"?

It is both, That's the order they would get interpreted; And we already visit them in this interpretation order, Right now we have the exact same visit order as our field definitions, If we find a conflict with the interpretation order we probably should consider it as a bug(unless we have a good reason for keeping it out of order)

PR #4010, opened 2024-07-01
fix(linter): find disabled directives using the message's `Span`.

### LINKED ISSUE #4005: linter: `no-empty-file` does not respect disable directive comments
When `oxlint -D no-empty-file` is run on a file like this:
```js
/* eslint-disable no-empty-file */
```

`no-empty-file`'s reported error is not filtered out by `DisableDirectives`.

rzvxa:
fixes #4005

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_empty_file.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_empty_file.rs
@@ -58,7 +58,13 @@ impl Rule for NoEmptyFile {
             return;
         }
 
-        ctx.diagnostic(no_empty_file_diagnostic(Span::new(0, 0)));
+        let mut span = program.span;
+        // only show diagnostic for the first 100 characters to avoid huge diagnostic messages with
+        // empty programs containing a bunch of comments.
+        // NOTE: if the enable/disable directives come after the first 100 characters they won't be
+        // respected by this diagnostic.
+        span.end = std::cmp::min(span.end, 100);
+        ctx.diagnostic(no_empty_file_diagnostic(span));
     }
 }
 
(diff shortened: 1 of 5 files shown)


rzvxa:
* **#4010** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

crates/oxc_linter/src/rules/unicorn/no_empty_file.rs
```diff
@@ -58,7 +58,7 @@ impl Rule for NoEmptyFile {
             return;
         }
 
-        ctx.diagnostic(no_empty_file_diagnostic(Span::new(0, 0)));
+        ctx.diagnostic(no_empty_file_diagnostic(program.span));
```
Boshen, line 61: This will spit out a huge diagnostic if the file has lots of commented code. Pointing to the first line should be fine here?

Boshen (changes requested):
Also need a test case.

crates/oxc_linter/src/rules/unicorn/no_empty_file.rs
rzvxa: I've limited it to the first 100 characters to avoid searching for the new line.
Disable comments without any leading or trailing spaces and with only a single rule(`no-empty-file`) is 34 characters, So a hundred should be plenty for multi-rule comments or weirdly formatted ones.

Let me know if this doesn't work for you or if you insist on limiting it to exactly the first line.

Boshen (approved):

PR #4030, opened 2024-07-02
fix(isolated_declarations): Fix readonly specifier on class constructor params

escaton:


escaton:
@Dunqing Here is one more :)

Dunqing:
I compared our output with the typescript's output and the result is still incorrect. But we can fix this in another PR

https://www.typescriptlang.org/play/?noCheck=false&isolatedDeclarations=true&isolatedModules=true#code/KYDwDg9gTgLgBAYwDYEMDOa4CEUC84DeAUHHFMCgCYQB2SAnnGFBGAIxwC8cA5GhAFtgcNDCgBLGgHMeAbhJMWYAExdeKGhBgALYFBFjJM+QF8gA

escaton:
You are right, I missed that. If you have an idea where is the problem, I'm happy to fix it in this PR

Dunqing:
> You are right, I missed that. If you have an idea where is the problem, I'm happy to fix it in this PR

We need to replace this with  `if param.accessibility.is_some() || param.readonly {`

https://github.com/oxc-project/oxc/blob/78c66d3f059a6c964c9e85a58314eb905a3723fe/crates/oxc_isolated_declarations/src/class.rs#L234

escaton:
Done.
There is still a little difference: TSC emits no type for private `prop2`
Oxc:
```
export declare class Boo {
	readonly prop: number;
	private readonly prop2: number;
	readonly prop3: number;
	constructor(prop?: number, prop2?: number, prop3?: number);
}

```
TSC:
```
export declare class Boo {
    readonly prop: number;
    private readonly prop2;
    readonly prop3: number;
    constructor(prop?: number, prop2?: number, prop3?: number);
}
```

But I don't think it is critical. Let me create a separate issue for that.

Dunqing (approved):
Thank you!

escaton:
https://github.com/oxc-project/oxc/issues/4033

PR #4035, opened 2024-07-02
feat(isolated-declarations): support optional class methods

escaton:
Example class
```
export class A {
  m1?(): void;
  m2(): void {}
}
```

Before the fix:
```
export declare class A {
  m1(): void;
}
```

with the fix:
```
export declare class A {
  m1?(): void;
  m2(): void;
}
```


escaton:
@Dunqing one more ;)

Dunqing (approved):

PR #4038, opened 2024-07-02
feat(linter/eslint): Implement no-multi-str

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/no-multi-str)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_multi_str.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_multi_str.rs
@@ -1,0 +1,80 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_multi_str_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint(no-multi-str): Unexpected multi string.").with_label(span0)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct NoMultiStr;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Disallow multiline strings.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// Some consider this to be a bad practice as it was an undocumented feature of JavaScript
+    /// that was only formalized later.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// var x = "Line 1 \
+    ///  Line 2";
+    /// ```
+    NoMultiStr,
+    style,
+);
+
+impl Rule for NoMultiStr {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::StringLiteral(literal) = node.kind() {
+            let source = literal.span.source_text(ctx.source_text());
+            // https://github.com/eslint/eslint/blob/9e6d6405c3ee774c2e716a3453ede9696ced1be7/lib/shared/ast-utils.js#L12
+            let position = source
+                .chars()
+                .position(|ch| matches!(ch, '\r' | '\n' | '\u{2028}' | '\u{2029}'))
+                .unwrap_or(0);
+            if position != 0 {
+                // We found the "newline" character but want to highlight the '\', so go back one
+                // character.
+                let multi_span_start =
+                    literal.span.start + u32::try_from(position).unwrap_or_default() - 1;
+                ctx.diagnostic(no_multi_str_diagnostic(Span::new(
+                    multi_span_start,
+                    multi_span_start + 1,
+                )));
+            }
+        }
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+
+    let pass = vec![
+        "var a = 'Line 1 Line 2';",
+        "var a = <div>
+			<h1>Wat</h1>
+			</div>;", // { "ecmaVersion": 6, "parserOptions": { "ecmaFeatures": { "jsx": true } } }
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/no_multi_str.rs
```diff
@@ -0,0 +1,74 @@
...
+impl Rule for NoMultiStr {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        // https://github.com/eslint/eslint/blob/9e6d6405c3ee774c2e716a3453ede9696ced1be7/lib/shared/ast-utils.js#L12
+        lazy_static! {
+            static ref MULTILINE_REGEX: Regex = Regex::new(r"\r\n|[\r\n\u2028\u2029]").unwrap();
+        }
+
+        if let AstKind::StringLiteral(literal) = node.kind() {
+            let source = literal.span.source_text(ctx.source_text());
+            if MULTILINE_REGEX.is_match(source) {
```
mysteryven, line 45: Shall we remove the Regex? We prefer avoid the regex crate when possible.

https://oxc.rs/docs/contribute/rules.html#development-policy

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
```diff
@@ -0,0 +1,74 @@
...
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use regex::Regex;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_multi_str_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint(no-multi-str): .").with_label(span0)
```
mysteryven, line 11: Looks like we forget to add error message :)

crates/oxc_linter/src/snapshots/no_multi_str.snap
```diff
@@ -0,0 +1,32 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint(no-multi-str): .
+   ╭─[no_multi_str.tsx:1:9]
+ 1 │ ╭─▶ var x = 'Line 1 \
```
mysteryven, line 6: Is it possible only to point out on the `\` ?

jelly: I suppose we can do that by `source.find('\')` and then creating a new Span with the StringLiteral.span.start + the index?

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
jelly: Woops, added locally, will push when I resolved the other comments.

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
jelly: Is it really slower then doing `.contains` and later `.find()` again to find the `\`?

mysteryven: I test on my machine with code, It's only a small diff:

```rust
if source.chars().any(|ch| {
    matches!(ch, '\r' | '\n' | '\u{2028}' | '\u{2029}' )
}) {
    ctx.diagnostic(no_multi_str_diagnostic(literal.span));
}
```

For vscode repo

```
hyperfine -w 5 -i -n oxlint_with_reg "./oxlint -A all -D no-multi-str" -n oxlint_with_contains "./oxlint_contains -A all -D no-multi-str"

oxlint_with_reg ran
    1.01 ± 0.02 times faster than oxlint_with_contains

```

For only `bignewlines.js`

```
hyperfine -w 5 -i -n oxlint_with_reg "./oxlint -A all -D no-multi-str bignewlines.js" -n oxlint_with_contains "./oxlint_contains -A all -D no-multi-str bignewlines.js"

 oxlint_with_reg ran
    1.02 ± 0.03 times faster than oxlint_with_contains

```

With real testing, I also this it's OK to use RegExp in this rule :)

crates/oxc_linter/src/snapshots/no_multi_str.snap
mysteryven: LGTM

jelly: I like the proposed `matches!` but would like to combine finding the character's position and match into one. It seems that `find` has [some issues](https://stackoverflow.com/a/46943949) with non-ASCII characters. So I went for finding the position, I assume that is as fast as `any`.

Not sure if I handled the `unwrap` well, but it seems impossible for the first position to be 0 and testing finds it will never be true:

```
  × eslint(no-multi-str): Unexpected multi string.
   ╭─[foo.js:7:12]
 6 │
 7 │ const b = "\
   ·            ─
 8 │ wjeifwe";
   ╰────
```

jelly: I've added an extra test for this case.

mysteryven: Our `start` is the `byte` offset from the start of the source text, I think `find` is ok here 🤔

https://github.com/oxc-project/oxc/blob/051ceb65396ea089f3213875afc8130112cb599d/crates/oxc_parser/src/lexer/mod.rs#L294-L295

mysteryven:
Thanks for your patient, I want to merge this nice PR, so I made a tiny change about using `find`, hope you never mind.

jelly:
> Thanks for your patient, I want to merge this nice PR, so I made a tiny change about using `find`, hope you never mind.

No problem! Should have added a test :)

PR #4045, opened 2024-07-03
refactor(ast)!: store trivia comments in a sorted slice

### LINKED ISSUE #4271: Convert `TriviasMap` (comments) into a `Vec`
I *think* `TriviasMap` (which contains `Comment`s) could be converted to a `Vec`.

If I remember right, it was a `Vec` originally, but was converted to a `BTreeMap` to work around a problem with comments getting inserted twice when lexer looks ahead and is then rewound.

This problem could still be avoided if it was a `Vec`. Comments are inserted in source text order, so duplicates could be avoided by checking before pushing whether the last entry's span start is equal or greater than the comment being inserted. If it is, don't push it again.

Seeking by span start could be still performed efficiently with a binary search.

As it's a flat `Vec`, there would be no significant gain to storing it in arena. `std::vec::Vec` is fine.

As mentioned in [oxc#3580](https://github.com/oxc-project/oxc/pull/3580#issuecomment-2154587480), `Trivias` could also be stored in `Allocator` so you can get a `&'a Trivias` with same lifetime as `Program<'a>`.

lucab:
This gets rid of `TriviasMap`, introducing `SortedComments` in order to store trivia comments in a sorted slice.

Closes: https://github.com/oxc-project/oxc/issues/4271

### DIFF
--- a/crates/oxc_ast/src/trivia.rs
+++ b/crates/oxc_ast/src/trivia.rs
@@ -1,8 +1,8 @@
 //! Trivias such as comments and irregular whitespaces
 
 use std::{
-    collections::btree_map::{BTreeMap, Range},
-    ops::{Deref, RangeBounds},
+    iter::FusedIterator,
+    ops::{Bound, Deref, RangeBounds},
     sync::Arc,
 };
 
@@ -37,15 +37,16 @@ impl CommentKind {
     }
 }
 
-pub type TriviasMap = BTreeMap<u32, Comment>;
+/// Sorted set of unique trivia comments, in ascending order by starting position.
+pub type SortedComments = Box<[(u32, Comment)]>;
 
 #[derive(Debug, Clone, Default)]
 pub struct Trivias(Arc<TriviasImpl>);
 
 #[derive(Debug, Default)]
 pub struct TriviasImpl {
-    /// Keyed by span.start
-    comments: TriviasMap,
+    /// Unique comments, ordered by increasing span-start.
+    comments: SortedComments,
 
     irregular_whitespaces: Vec<Span>,
 }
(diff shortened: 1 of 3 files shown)


Boshen (approved):
This is amazing! Love the renaming to `SortedComments`.

Boshen:
You can run `just ready` or just `just r` to quickly verify everything.

crates/oxc_ast/src/trivia.rs
```diff
@@ -59,26 +60,124 @@ impl Deref for Trivias {
 }
 
 impl Trivias {
-    pub fn new(comments: TriviasMap, irregular_whitespaces: Vec<Span>) -> Trivias {
+    pub fn new(comments: SortedComments, irregular_whitespaces: Vec<Span>) -> Trivias {
```
lucab, line 63: I think this method can be switched to `pub(crate)` instead. That may allow dropping a bunch of `pub` from a chain of related structs. If that's ok, I'll be happy to explore that in a followup.

Boshen: 👍 Go ahead

Boshen:
No performance improvement from the parser benchmarks is somewhat expected.

Let me double check the semantic benchmark, I think I forgot to turn jsdoc parsing back on (which reads the comments).

Boshen:
I updated the benchmark on main, a quick rebase should reveal if there are any performance changes.

lucab:
Rebased, there are indeed some small perf improvements.

crates/oxc_ast/src/trivia.rs
lucab: Nevermind, this is involved in a cross-crate call (`oxc_parser -> oxc_ast`) thus it needs to stay public.

PR #4051, opened 2024-07-04
docs(ast): improve AST node documentation

### LINKED ISSUE #2803: ast: improve AST node documentation
While I was working with `syn`, I found the documentation on AST nodes really helpful, we should do the same.

https://docs.rs/syn/latest/syn/index.html#structs

re-taro:
partof #2803 

I even wrote a VariableDeclaration.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -58,89 +58,89 @@ inherit_variants! {
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[serde(untagged)]
 pub enum Expression<'a> {
-    /// See [`BooleanLiteral`] for ast node details.
+    /// See [`BooleanLiteral`] for AST node details.
     BooleanLiteral(Box<'a, BooleanLiteral>) = 0,
-    /// See [`NullLiteral`] for ast node details.
+    /// See [`NullLiteral`] for AST node details.
     NullLiteral(Box<'a, NullLiteral>) = 1,
-    /// See [`NumericLiteral`] for ast node details.
+    /// See [`NumericLiteral`] for AST node details.
     NumericLiteral(Box<'a, NumericLiteral<'a>>) = 2,
-    /// See [`BigIntLiteral`] for ast node details.
+    /// See [`BigIntLiteral`] for AST node details.
     BigIntLiteral(Box<'a, BigIntLiteral<'a>>) = 3,
-    /// See [`RegExpLiteral`] for ast node details.
+    /// See [`RegExpLiteral`] for AST node details.
     RegExpLiteral(Box<'a, RegExpLiteral<'a>>) = 4,
-    /// See [`StringLiteral`] for ast node details.
+    /// See [`StringLiteral`] for AST node details.
     StringLiteral(Box<'a, StringLiteral<'a>>) = 5,
-    /// See [`TemplateLiteral`] for ast node details.
+    /// See [`TemplateLiteral`] for AST node details.
     TemplateLiteral(Box<'a, TemplateLiteral<'a>>) = 6,
 
-    /// See [`IdentifierReference`] for ast node details.
+    /// See [`IdentifierReference`] for AST node details.
     Identifier(Box<'a, IdentifierReference<'a>>) = 7,
 
-    /// See [`MetaProperty`] for ast node details.
+    /// See [`MetaProperty`] for AST node details.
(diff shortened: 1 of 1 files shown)


crates/oxc_ast/src/ast/js.rs
```diff
@@ -419,7 +459,7 @@ impl<'a> Expression<'a> {
     }
 }
 
-/// Identifier Name
+/// Fundamental syntactic structure used for naming variables, functions, and properties. It must start with a Unicode letter (including $ and _) and can be followed by Unicode letters, digits, $, or _.: `var` in `let var = 1;`
```
Boshen, line 462: `cargo doc` shows the first line as the title, and the next lines as descriptions.

image

So the first line should be simple to understand 
image

re-taro: I see, understanding.

crates/oxc_ast/src/ast/js.rs
```diff
@@ -535,18 +579,17 @@ pub struct ArrayExpression<'a> {
 }
 
 inherit_variants! {
-/// Array Expression Element
-///
-/// Inherits variants from [`Expression`]. See [`ast` module docs] for explanation of inheritance.
```
overlookmotel, line 540: Could you possibly restore these comments about inheritance? I know they're repeated throughout this file, but it's quite confusing, so I think best to err on side of making sure a reader sees the comment and has a chance of understanding what's going on!

crates/oxc_ast/src/ast/js.rs
```diff
@@ -101,49 +102,88 @@ inherit_variants! {
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(untagged))]
 pub enum Expression<'a> {
+    /// See [`BooleanLiteral`] for ast node details.
```
overlookmotel, line 105: This is a real nit, but could we please use capitalized "AST" rather than "ast"? This isn't consistent throughout Oxc's codebase as present, but since you're going to the trouble of documenting the AST, it'd be great to set a good pattern here.

overlookmotel:
Just want to say thanks *loads* for tackling this. Documentation isn't always the most "glamorous" work, but it is a priority for the project to be approachable to new contributors, and the AST is the core of Oxc, so this is really important work. So, yeah, just wanted say, really appreciate you working on this.

overlookmotel:
@re-taro #4413 also added more doc comments to AST. Would be keen to get your changes merged too on top of that. Would you have time to rebase on latest master and resolve conflicts, and then we can get it merged?

re-taro:
@overlookmotel 

Yes, we can get on with it now. But it's still a work in progress, so is there a problem with merging it in stages?

overlookmotel:
Thanks for coming back so swiftly. Making changes in stages is totally fine, as long as an intermediate stage still leaves Oxc in a working state (CI green). In fact, it's good! Shorter-lived branches = less merge conflicts, which takes up less time for PR authors, and we can all keep moving swiftly.

re-taro:
ok! I'll get the conflicts out of ASAP and put it up for review!

crates/oxc_ast/src/ast/js.rs
overlookmotel: This change has been made.

overlookmotel: This has now been addressed.

overlookmotel:
Oh I'm sorry. I misunderstood what you meant by incomplete. I thought you meant that you intended to add comments to more AST nodes, but hadn't done them all yet.

I'm afraid Boshen's comment above I think does need addressing before we merge.

Just to make sure what Boshen was saying is clear, for e.g. `IdentifierName`, this would be ideal:

```rs
/// Identifier name.
///
/// Fundamental syntactic structure used for naming variables, functions, and properties.
/// It must start with a Unicode letter (including `$` and `_`) and can be followed by
/// Unicode letters, digits, `$`, or `_`.
///
/// See: [13.1 Identifiers](https://tc39.es/ecma262/#sec-identifiers)
///
/// ## Example
/// ```js
/// let foo = 1;
///     ^^^
/// ```
#[ast(visit)]
/*...etc...*/
pub struct IdentifierName<'a> {
```

On a few types, this PR removes some useful text which was there previously e.g. `ObjectExpression`. Could you please restore that text?

Boshen:
The current style is good and will show what I want on `cargo docs`. Let's merge some of this and continue iterating on it.

re-taro:
fmm… can’t merge it?

PR #4052, opened 2024-07-04
feat(linter): eslint-plugin-jest/prefer-hooks-in-order

eryue0220:

Rule Detail: [link](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-hooks-in-order.ts)


crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
```diff
@@ -0,0 +1,1279 @@
...
+                    call_expr,
+                    ctx,
+                );
+            };
+        }
+    }
+}
+
+impl PreferHooksInOrder {
+    fn check<'a>(
```
mysteryven, line 165: Seems we can be more simple:

1. Group `nodes` by `scope_id`
2. In each group, check if it's ordered, in a group, we only need a record a `previousHookIndex` and `currentHookIndex` like jest does.

Do you have interest to try it out?

crates/oxc_linter/src/snapshots/prefer_hooks_in_order.snap
```diff
@@ -0,0 +1,409 @@
...
+---
+  ⚠ eslint-plugin-jest(prefer-hooks-in-order): Prefer having hooks in a consistent order.
+   ╭─[prefer_hooks_in_order.tsx:6:21]
+ 5 │                         });
+ 6 │ ╭─▶                     beforeAll(() => {
+ 7 │ │                           createMyDatabase();
+ 8 │ ╰─▶                     });
+ 9 │                     };
+   ╰────
+  help: `"beforeAll"` hooks should be before any `"afterAll"` hooks
```
mysteryven, line 13: Nitpick: Looks like should be: `help: "beforeAll" hooks should be before any "afterAll" hooks`

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
```diff
@@ -0,0 +1,1279 @@
...
+    ///         });
+    ///     });
+    ///     afterAll(() => {
+    ///         removeMyDatabase();
+    ///     });
+    /// });
+    /// ```
+    ///
+    ///
+    /// This rule is compatible with [eslint-plugin-vitest](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/prefer_hooks_in_order.md),
```
mysteryven, line 131: This link is broken 👀

crates/oxc_linter/src/snapshots/prefer_hooks_in_order.snap
eryue0220: Done.

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
eryue0220: Done.

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
eryue0220: Updated.

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
```diff
@@ -0,0 +1,1271 @@
...
+    fn check<'a>(
+        previous_hook_index: &mut i32,
+        possible_jest_node: &PossibleJestNode<'a, '_>,
+        call_expr: &'a CallExpression<'_>,
+        ctx: &LintContext<'a>,
+    ) {
+        let Some(ParsedJestFnCallNew::GeneralJestFnCall(jest_fn_call)) =
+            parse_jest_fn_call(call_expr, possible_jest_node, ctx)
+        else {
+            *previous_hook_index = -1;
```
mysteryven, line 175: Curious about why we `-1` here.

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
```diff
@@ -0,0 +1,1271 @@
...
+            return;
+        };
+
+        if !matches!(jest_fn_call.kind, JestFnKind::General(JestGeneralFnKind::Hook)) {
+            *previous_hook_index = -1;
```
mysteryven, line 180: Same question as above, why non-hook jest fn can let hook index minus 1

crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
eryue0220: Because of these two cases:

```js
describe('my test', () => {
    afterEach(() => {});
    afterAll(() => {});

    doSomething();

    beforeAll(() => {});
    beforeEach(() => {});
});

describe('my test', () => {
    afterEach(() => {});
    afterAll(() => {});

    it('test', () => {});

    beforeAll(() => {});
    beforeEach(() => {});
  });
```

After a non-hook call or non-jest call, it need to reset the index, or it would fail.

mysteryven (approved):
Thank you!

Issue #4060, opened 2024-07-05
Oxc crate v0.17.0 is yanked


Boshen:
There is a huge behavior change in the visitor where expression visits are missed.

e.g.

Before:
```rust
    pub fn walk_expression_array_element<'a, V: Visit<'a>>(visitor: &mut V, expr: &Expression<'a>) {
        let kind = AstKind::ExpressionArrayElement(visitor.alloc(expr));
        visitor.enter_node(kind);
        visitor.visit_expression(expr);
        visitor.leave_node(kind);
    }
```

v0.17.0:

```rust
    pub fn walk_expression_array_element<'a, V: Visit<'a>>(visitor: &mut V, it: &Expression<'a>) {
        let kind = AstKind::ExpressionArrayElement(visitor.alloc(it));
        visitor.enter_node(kind);
        match it {
            Expression::BooleanLiteral(it) => visitor.visit_boolean_literal(it),
            Expression::NullLiteral(it) => visitor.visit_null_literal(it),
            Expression::NumericLiteral(it) => visitor.visit_numeric_literal(it),
            Expression::BigIntLiteral(it) => visitor.visit_big_int_literal(it),
            Expression::RegExpLiteral(it) => visitor.visit_reg_exp_literal(it),
```

THIS IS A BUG.

rzvxa:
Let's keep this one open for a few days in case of any other issues.

rzvxa:
closed for housekeeping since it seems to be fixed now.

PR #4061, opened 2024-07-05
fix(ast_codegen, ast): visit `ExpressionArrayElement` as `Expression`.

rzvxa:
hotfix for #4060

I just added an edge case, We can come back to it later on and make it a standard in our codegen.

The diff is big because it causes the code to reorder, The edge case generated code doesn't follow the order in which they are defined in the source of truth(eg `js.rs`)

### DIFF
--- a/crates/oxc_ast/src/generated/visit.rs
+++ b/crates/oxc_ast/src/generated/visit.rs
@@ -1690,57 +1690,7 @@ pub mod walk {
     pub fn walk_expression_array_element<'a, V: Visit<'a>>(visitor: &mut V, it: &Expression<'a>) {
         let kind = AstKind::ExpressionArrayElement(visitor.alloc(it));
         visitor.enter_node(kind);
-        match it {
-            Expression::BooleanLiteral(it) => visitor.visit_boolean_literal(it),
-            Expression::NullLiteral(it) => visitor.visit_null_literal(it),
-            Expression::NumericLiteral(it) => visitor.visit_numeric_literal(it),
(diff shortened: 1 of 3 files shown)


rzvxa:
* **#4061** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4061?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
Not just `Expression`, we need to visit the corresponding `AST` for all `@inherit XXX`

rzvxa:
> Not just `Expression`, we need to visit the corresponding `AST` for all `@inherit XXX`

We already do this for the inherited enums, `ExpressionArrayElement` is special because it doesn't exist, I use the `visit_as` attribute to visit an arbitrary `Expression` as a non-existing type `ExpressionArrayElement`, That's actually what causes this bug, Since I was generating a new walk for this based on the `Expression` supertype.

It also happens in `FinallyClause` and `ClassHeritage` but those 2 are structures so It doesn't need to consider inheritance.
For now, I used an edge case, But we should standardize the behavior or `visit_as` on enums to act this way out of the gate.

Dunqing:
Thanks for the explanation, I see the root cause

rzvxa:
> Thanks for the explanation, I see the root cause

Your question actually showed me another edge case, `ClassHeritage` is also an alias for `Expression` I have to add that one to our edge case too.

Boshen:
Confirmed in Rolldown by testing locally.

PR #4064, opened 2024-07-06
fix(isolated_declarations): correct emit for private static methods

MichaelMitchell-at:
Fixes issue where `static` gets dropped in `private static` methods.


Dunqing (approved):
Thanks!

PR #4068, opened 2024-07-06
fix(isolated_declarations): Infer type of template literal expressions as string

MichaelMitchell-at:
In a "non-`const`" context, a template literal string can just be inferred a `string`. This is consistent with TypeScript's behavior.

### DIFF
--- a/crates/oxc_isolated_declarations/src/declaration.rs
+++ b/crates/oxc_isolated_declarations/src/declaration.rs
@@ -77,7 +77,9 @@ impl<'a> IsolatedDeclarations<'a> {
                     } else {
                         init = Some(self.ast.copy(init_expr));
                     }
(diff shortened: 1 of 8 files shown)


Dunqing:
The `infer_type_from_expression` method can infer `TemplateLiteral` as a string.  But there is a mistake here, when `expression` exists but can be inferred, we should infer that it is a `string` type

https://github.com/oxc-project/oxc/blob/adee7280d74a05e5443630ee52ee7b42017489b4/crates/oxc_isolated_declarations/src/inferrer.rs#L28-L34

MichaelMitchell-at:
@Dunqing I've updated the logic. Please have another look.

Dunqing (approved):

PR #4070, opened 2024-07-06
refactor(isolated-declarations): reorganize scope tree

lucab:
This reorganizes scope fields into levels which atomically follow the enter/leave pattern. It avoids tracking the same lenght and capacity separately for each property, and does not need to grow multiple vectors while traversing.

### DIFF
--- a/crates/oxc_isolated_declarations/src/scope.rs
+++ b/crates/oxc_isolated_declarations/src/scope.rs
@@ -8,89 +8,106 @@ use oxc_span::Atom;
 use oxc_syntax::scope::ScopeFlags;
 use rustc_hash::FxHashSet;
 
+/// Declaration scope.
+#[derive(Debug)]
+struct Scope<'a> {
+    type_bindings: FxHashSet<Atom<'a>>,
+    value_bindings: FxHashSet<Atom<'a>>,
+    type_references: FxHashSet<Atom<'a>>,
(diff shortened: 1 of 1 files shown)


Dunqing (approved):
Great work. Thank you!

overlookmotel:
@lucab Just out of interest, did you see any performance change from this PR? I think it was before we added a benchmark on CodSpeed for isolated-declarations. I'm looking at revamping `Semantic`, and would be interested to know whether this move from "struct of arrays" style to "array of structs" was good or bad for perf.

lucab:
@overlookmotel I didn't have good benchmarks to compare, but running some small examples through `hyperfine` and `valgrind` didn't show relevant changes in either directions. We can try reverting it (it seems @Dunqing may be doing that now) as see what was the real delta.

That said, here there was no consumer using the SoA to slice through a single field, but instead all Vecs were pushed/popped at the same time. `Semantic` may or may not have different access patterns to this one.

Dunqing:
From the benchmark [results](https://codspeed.io/oxc-project/oxc/branches/revert-4070-ups/isolated-decl-scopes), there is no difference in performance. However, this makes the code organization clearer.

overlookmotel:
OK cool. Thanks both for coming back. You're right lucab that `Semantic` may well have different access patterns, so a different approach may be appropriate there. But still interesting to know it made little difference in this case.

PR #4072, opened 2024-07-06
fix(lexer): incorrect lexing of large hex/octal/binary literals

### LINKED ISSUE #3347: Incorrect parsing of large binary/octal/hex literals
@DonIsaac I'm afraid I found an edge case with #3296.

Because of the switch from using floating point to integers, we need to handle arithmetic overflow now.

For numbers which evaluate to more than `u64::MAX`, parsing is incorrect.

e.g.: `0x10000000000000000` should be evaluated as `2.0f64.powf(64.0)`, but instead produces `0`.

[Playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAIDzgICAgICAgICXYPzioZ7SxU1dOdzNYEZUBh88G3nYsxpCi9Z0WSpDMLcz2B5pa4JLNbDfSPI%2BVqnhinaki%2FiGtv9%2FV9oAgA%3D%3D)

DonIsaac:
Closes #3347. Implementation follows the approach described by @overlookmotel [here](https://github.com/oxc-project/oxc/issues/3347#issuecomment-2119004288).

### DIFF
--- a/crates/oxc_parser/src/lexer/number.rs
+++ b/crates/oxc_parser/src/lexer/number.rs
@@ -25,10 +25,8 @@ pub fn parse_float(s: &str, has_sep: bool) -> Result<f64, &'static str> {
 /// This function assumes `s` has had all numeric separators (`_`) removed.
 /// Parsing will fail if this assumption is violated.
 fn parse_int_without_underscores(s: &str, kind: Kind) -> Result<f64, &'static str> {
-    if kind == Kind::Decimal {
-        return parse_float_without_underscores(s);
-    }
     match kind {
+        Kind::Decimal => parse_float_without_underscores(s),
         Kind::Binary => Ok(parse_binary(&s[2..])),
         Kind::Octal => {
             let s = if s.starts_with("0o") || s.starts_with("0O") {
@@ -49,6 +47,17 @@ fn parse_float_without_underscores(s: &str) -> Result<f64, &'static str> {
     s.parse::<f64>().map_err(|_| "invalid float")
 }
 
+/// b'0' is 0x30 and b'1' is 0x31.
+///
+/// So we can convert from binary digit to its value with `c & 1`.
+/// This is produces more compact assembly than `c - b'0'`.
+///
+/// <https://godbolt.org/z/1vvrK78jf>
+const fn binary_byte_to_value(c: u8) -> u8 {
+    debug_assert!(c == b'0' || c == b'1');
+    c & 1
+}
+
 /// NOTE: bit shifting here is is safe and much faster than f64::mul_add.
 /// It's safe because we're sure this number is an integer - if it wasn't, it
 /// would be a [`Kind::Float`] instead. It's fast because shifting usually takes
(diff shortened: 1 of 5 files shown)


crates/oxc_parser/src/lexer/number.rs
```diff
@@ -212,10 +260,48 @@ mod test {
...
+    static_assertions::const_assert_eq!(hex_byte_to_value(b'a'), 10);
+    static_assertions::const_assert_eq!(hex_byte_to_value(b'f'), 15);
+
     #[test]
-    #[allow(clippy::excessive_precision)]
+    #[allow(clippy::excessive_precision, clippy::cast_precision_loss)]
     fn test_int_precision() {
         assert_eq!(parse_int("9007199254740991", Kind::Decimal, false), Ok(9007199254740991.0));
+        assert_eq!(
+            parse_int("0x10000000000000000", Kind::Hex, false),
```
Boshen, line 280: Can you move these to `tasks/coverage/misc/pass`? I wanna check regression for all the components as well.

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -83,72 +82,121 @@ fn parse_binary(s: &str) -> f64 {
...
-    // https://godbolt.org/z/9rYTsMoMM
-    const fn byte_to_value(c: u8) -> u8 {
-        debug_assert!(c >= b'0' && c <= b'7');
-        c & 7
-    }
-    #[cfg(test)]
-    {
-        static_assertions::const_assert_eq!(byte_to_value(b'0'), 0);
-        static_assertions::const_assert_eq!(byte_to_value(b'7'), 7);
+    const MAX_FAST_OCTAL_LEN: usize = 21;
```
overlookmotel, line 99: Please add a comment as to why 21 is the "magic number" here.

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -83,72 +82,121 @@ fn parse_binary(s: &str) -> f64 {
...
+
+    debug_assert!(!s.is_empty());
+    debug_assert!(!s.starts_with("0o") && !s.starts_with("0O"));
+
+    let without_leading_zeros = s.trim_start_matches('0');
```
overlookmotel, line 104: I think this logic can be simplified to minimize the the number of operations before getting into the fast path (which will be almost all cases). Just this would do it:

```rs
if s.len() > MAX_FAST_OCTAL_LEN {
  parse_octal_slow(s)
} else {
  // Fast path code here
}
```

I know it was me who originally suggested trimming off leading zeros, but with the simpler (better) version of `parse_octal_slow` that you've used, it's not necessary. `trim_start_matches` is one of the annoying stdlib functions which processes text as `char`s, so it's quite slow. If you wanted to retain the removal of leading zeros, you could implement it as loop over bytes.

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -83,72 +82,121 @@ fn parse_binary(s: &str) -> f64 {
...
-        if c < b'A' {
-            c & 15 // 0-9
-        } else {
-            (c & 15) + 9 // A-F or a-f
+    const MAX_FAST_HEX_LEN: usize = 16;
+
+    debug_assert!(!s.is_empty());
+    debug_assert!(!s.starts_with("0x"));
+
+    let without_leading_zeros = s.trim_start_matches('0');
```
overlookmotel, line 167: Same as above - I think this logic can be simplified.

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -83,72 +82,121 @@ fn parse_binary(s: &str) -> f64 {
...
+    match without_leading_zeros.len() {
+        0 => 0.0,
+        1 => return hex_byte_to_value(without_leading_zeros.as_bytes()[0]) as f64,
+        n if n > MAX_FAST_HEX_LEN => parse_hex_slow(without_leading_zeros),
+        _ => {
+            let mut result = 0_u64;
+            let mut has_seen_non_zero = false;
```
overlookmotel, line 174: I'm not clear on the purpose of `has_seen_non_zero` here. Since leading zeros are trimmed off already, unless I'm missing something, `has_seen_non_zero` will always be set to `true` on the first turn of the loop, and `if !has_seen_non_zero` branch will never be taken.

Even if that wasn't the case, I *suspect* (but don't know for sure) that the extra branching on `has_seen_non_zero` is probably more costly than just performing the 2 (very cheap) bitwise ops.

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -212,10 +260,48 @@ mod test {
...
+            parse_int("0x10000000000000000", Kind::Hex, false),
+            Ok(0x10000000000000000_i128 as f64)
+        );
+        assert_eq!(
+            parse_int("0o40000000000000000", Kind::Octal, false),
+            Ok(0o40000000000000000_i128 as f64)
+        );
+        assert_eq!(
+            parse_int(
+                "0b0001000000000000000000000000000000000000000000000000000000000000",
```
overlookmotel, line 289: I don't think this test is covering the case that we need it to. You have 60 zeros after the 1 here. If you add 4 more, I think you'll find the test fails. `parse_binary` will need changing in same manner as `parse_octal` and `parse_hex` to fix that.

overlookmotel (changes requested):

overlookmotel:
Hi @DonIsaac. Thanks very much for tackling this.

Apologies if some of my comments above appear nit-picky. Maybe they are, given that these functions are not in a hot path, so a few extra instructions aren't a bit deal. But I just figured that since I noticed these things, I might as well point them out. Feel free to ignore them if you don't feel it's worth the bother.

One other thing: Could you please add a comment at the top somewhere explaining what the reason for the fast path/slow path setup is, so a future reader can follow why it's like this?

Boshen:
@DonIsaac Can you wrap up this PR?

crates/oxc_parser/src/lexer/number.rs
DonIsaac: it was my attempt to skip leading zeroes; I forgot to clean this up, apologies

crates/oxc_parser/src/lexer/number.rs
DonIsaac: Done

crates/oxc_parser/src/lexer/number.rs
DonIsaac: Done

crates/oxc_parser/src/lexer/number.rs
DonIsaac: Done

crates/oxc_parser/src/lexer/number.rs
DonIsaac: Good catch, fixed

crates/oxc_parser/src/lexer/number.rs
```diff
@@ -55,33 +66,47 @@ fn parse_float_without_underscores(s: &str) -> Result<f64, &'static str> {
...
-        result |= byte_to_value(*c) as u64;
+        result |= binary_byte_to_value(*c) as u64;
     }
     result as f64
 }
 
+#[cold]
+#[inline(never)]
+fn parse_binary_slow(s: &str) -> f64 {
+    debug_assert!(!s.is_empty());
```
overlookmotel, line 99: Small nit: This line is unnecessary. Function will work fine for zero-length strings, and it's checked above anyway.

DonIsaac: Fixed

overlookmotel (approved):

overlookmotel:
Nice one @DonIsaac.

If CI passes, and you have no further changes to make, feel free to merge this.

PR #4073, opened 2024-07-06
fix(minifier): omit dce `undefined` which can be a shadowed variable

Boshen:
```
 // Shadowed `undefined` as a variable should not be erased.
    test(
        "function foo(undefined) { if (!undefined) { } }",
        "function foo(undefined){if(!undefined){}}",
    );
```

I'm not using the cheap `ident.reference_id.get().is_some()` here yet because I don't know what I'm doing - how should minifier consume `Semantic`?


Boshen:
Passed tests in Rolldown.

PR #4077, opened 2024-07-07
fix(isolated_declarations): Remove nested AssignmentPatterns from inside parameters

MichaelMitchell-at:
The default values in destructured parameters are retained in declarations, which can cause captured variables to be part of the emit when they shouldn't be. This can also lead to unnecessary isolated declaration errors when those variables are themselves missing type annotations and can't be inferred.

For example:
```ts
const x = 42;
const y = '';
export function fooGood3({a = x, b: [{c = y}]}: object): void {}
```

before this change will be emitted as:
```ts
declare const x = 42;
declare const y = '';
export declare function fooGood3({ a = x, b: [{ c = y }] }: object): void;
```

and after this change will be emitted as:
```ts
export declare function fooGood3({ a, b: [{ c }] }: object): void;
```

### DIFF
--- a/crates/oxc_isolated_declarations/src/function.rs
+++ b/crates/oxc_isolated_declarations/src/function.rs
@@ -9,6 +9,7 @@ use crate::{
         function_must_have_explicit_return_type, implicitly_adding_undefined_to_type,
         parameter_must_have_explicit_type,
     },
+    formal_parameter_binding_pattern::FormalParameterBindingPattern,
     IsolatedDeclarations,
 };
 
@@ -64,6 +65,8 @@ impl<'a> IsolatedDeclarations<'a> {
             } else {
                 self.ast.copy(&param.pattern)
             };
+
+        FormalParameterBindingPattern::remove_assignments_from_kind(self.ast, &mut pattern.kind);
 
         if is_assignment_pattern || pattern.type_annotation.is_none() {
             let type_annotation = pattern
--- a/crates/oxc_isolated_declarations/src/formal_parameter_binding_pattern.rs
+++ b/crates/oxc_isolated_declarations/src/formal_parameter_binding_pattern.rs
@@ -1,0 +1,24 @@
+use oxc_ast::{
+    ast::BindingPatternKind, visit::walk_mut::walk_binding_pattern_kind, AstBuilder, VisitMut,
+};
+
+pub struct FormalParameterBindingPattern<'a> {
+    ast: AstBuilder<'a>,
+}
+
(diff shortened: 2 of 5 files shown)


crates/oxc_isolated_declarations/src/function.rs
```diff
@@ -42,6 +42,54 @@ impl<'a> IsolatedDeclarations<'a> {
         }
     }
 
+    fn copy_binding_pattern_kind_without_assignments(
```
MichaelMitchell-at, line 45: I feel that this approach is not very satisfying as any changes to the AST will require updating this code. A simpler approach might be to copy the entire kind and then modify the copy in place using a visitor, but it doesn't seem like this is possible with the visitor/walker abstractions I found in the code.

crates/oxc_isolated_declarations/src/function.rs
```diff
@@ -65,6 +113,8 @@ impl<'a> IsolatedDeclarations<'a> {
                 self.ast.copy(&param.pattern)
             };
 
+        pattern.kind = self.copy_binding_pattern_kind_without_assignments(&pattern.kind);
```
MichaelMitchell-at, line 116: This might be an unnecessary copy if no default values exist, but I feel that detecting whether a copy is necessary creates extra work in a different way.

crates/oxc_isolated_declarations/src/function.rs
Dunqing: You can do this in a new structure and impl `VisitMut` for it. This way you can modify the AST

crates/oxc_isolated_declarations/src/function.rs
```diff
@@ -42,6 +42,54 @@ impl<'a> IsolatedDeclarations<'a> {
...
+                            property.computed,
+                        )
+                    }),
+                    self.ast.allocator,
+                ),
+                self.ast.copy(&pattern.rest),
+            ),
+            BindingPatternKind::ArrayPattern(pattern) => self.ast.array_pattern(
+                SPAN,
+                Vec::from_iter_in(
```
Dunqing, line 75: Use `self.ast.new_vec_from_iter` instead of it

Dunqing (changes requested):

crates/oxc_isolated_declarations/src/function.rs
MichaelMitchell-at: Updated code to use the visitor approach.

Dunqing (approved):
That's what I thought!

PR #4079, opened 2024-07-07
feat(linter): implement unicorn/no-useless-undefined

### LINKED ISSUE #3870: feat(linter): eslint-plugin-unicorn/no-useless-undefined
To get started, run

```sh
just new-unicorn-rule no-useless-undefined
```

- [Rule docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v54.0.0/docs/rules/no-useless-undefined.md)
- [Rule source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v54.0.0/rules/no-useless-undefined.js)

cblh:
Resolves #3870

Hey there, thought I'd give this a try as it's tagged "good first issue" :) Let me know if there's anything that needs to change

Another part of the code also needs modification:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-useless-undefined.js#L95

I tried to modify it. If there are any issues with these modifications, please let me know.               
```
fn is_has_function_return_type(node: &AstNode, ctx: &LintContext<'_>) -> bool {
    let Some(parent_node) = ctx.nodes().parent_node(node.id()) else {
        return false;
    };
    match parent_node.kind() {
        AstKind::ArrowFunctionExpression(arrow_func_express) => {
            arrow_func_express.return_type.is_some()
        }
        AstKind::Function(func) => func.return_type.is_some(),
        _ => is_has_function_return_type(parent_node, ctx),
    }
}
```


crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::IdentifierReference(undefined_literal)
+                if undefined_literal.name == "undefined" =>
+            {
+                let Some(parent_node) = ctx.nodes().parent_node(node.id()) else { return };
+                let parent_node_kind = parent_node.kind();
+
+                // `return undefined`
+                if let AstKind::ReturnStatement(ret_stmt) = parent_node_kind {
```
DonIsaac, line 144: Use a match statement here.

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+    "same",
+    "notSame",
+    "strictSame",
+    "strictNotSame",
+];
+
+fn is_match_ignore_func_name(name: &str) -> bool {
+    COMPARE_FUNCTION_NAMES.contains(&name)
+        // `array.push(undefined)`
+        || name == "push"
```
DonIsaac, line 73: Create a static set for these names. Also, consider sorting them and using a binary search for faster checks.

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+fn warn() -> OxcDiagnostic {
+    OxcDiagnostic::warn(
+        "eslint-plugin-unicorn(no-useless-undefined): Do not use useless `undefined`.",
+    )
+    .with_help("Consider using `null` instead.")
```
DonIsaac, line 14:
```suggestion
    .with_help("Consider removing `undefined` or using `null` instead.")
```

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+fn is_function_bind_call(call_expr: &CallExpression) -> bool {
+    !call_expr.optional
+        && matches!(&call_expr.callee, Expression::StaticMemberExpression(member_expr)
+        if member_expr.property.name.as_str() == "bind")
+}
+
+fn is_undefined(arg: &Argument) -> bool {
```
DonIsaac, line 116: We already have an utility function that does this, consider using it.

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+        (r"foo.bind(bar, baz, undefined);", None),
+        (r"foo?.bind(bar, undefined);", None),
+        // `checkArguments: false`
+        (r"foo(undefined, undefined);", options_ignore_arguments()),
+        (r"foo.bind(undefined);", options_ignore_arguments()),
+        // `checkArrowFunctionBody: false`
+        (r"const foo = () => undefined", options_ignore_arrow_function_body()),
+    ];
+
+    let fail = vec![(r"let foo = undefined;", None)];
```
DonIsaac, line 358: Should we be checking for cases such as parameter initializers and assignment expressions?

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+    let options_ignore_arguments = || Some(serde_json::json!({ "checkArguments": false }));
+    let options_ignore_arrow_function_body =
+        || Some(serde_json::json!({"checkArrowFunctionBody": false}));
+    let pass = vec![
```
DonIsaac, line 296: Can you please add cases for member expression assignment and class property assignment? These should all pass:
```ts
const x = { a: undefined }
const y: any = {}
y.foo = undefined

class Foo () {
  public x: number | undefined = undefined }
}

function foo(x: string | undefined = undefined) { } // maybe fix into `x?: string`?
```

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+                // `return undefined`
+                if let AstKind::ReturnStatement(ret_stmt) = parent_node_kind {
+                    ctx.diagnostic_with_fix(
+                        no_useless_undefined_diagnostic(undefined_literal.span),
+                        |fixer| {
+                            let comments_range = ctx
+                                .semantic()
+                                .trivias()
+                                .comments_range(ret_stmt.span.start..ret_stmt.span.end);
+                            if let Some((_, comment)) = comments_range.last() {
```
DonIsaac, line 152: This code can be simplified.

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,403 @@
...
+        }
+        Expression::StaticMemberExpression(static_assertions) => {
+            let name = static_assertions.property.name.as_str();
+            is_match_ignore_func_name(name)
+        }
+        _ => false,
+    }
+}
+
+fn is_function_bind_call(call_expr: &CallExpression) -> bool {
```
DonIsaac, line 110: Consider utilizing `ast_utils::is_method_call`

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
cblh: I will add cases for member expression assignment and class property assignment，but`function foo(x: string | undefined = undefined) { } maybe fix into `x?: string`?` should be fix like that: `x?: string`

I try add these case on:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-useless-undefined.mjs

but get `?: string | undefined`

```
valid(2): function foo(x: string | undefined = undefined) { }
  Should have no errors but had 1: [
    {
      ruleId: 'rule-to-test/no-useless-undefined',
      severity: 1,
      message: 'Do not use useless `undefined`.',
      line: 1,
      column: 38,
      nodeType: 'Identifier',
      messageId: 'no-useless-undefined',
      endLine: 1,
      endColumn: 47,
      fix: { range: [Array], text: '?: string | undefined' }
    }
  ]
```

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
cblh: I attempted to use phf::phf_set! like this file, but I'm confused why using set still requires binary search.
https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/unicorn/prefer_add_event_listener.rs#L68

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
cblh: Yes it has been added.

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,670 @@
...
+}
+fn no_useless_undefined_diagnostic(span0: Span) -> OxcDiagnostic {
+    warn().with_label(span0)
+}
+fn no_useless_undefined_diagnostic_spans(spans: Vec<Span>) -> OxcDiagnostic {
+    warn().with_labels(spans)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct NoUselessUndefined {
```
DonIsaac, line 24: in from_confuguration, these default to `true`. However, `#[derive(Default)]` will automatically default booleans to false. Maybe implement Default manually?

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,670 @@
...
+				undefined,
+				undefined,
+			)
+		",
+            None,
+        ),
+        ("function foo([bar = undefined] = []) {}", None),
+        ("foo(bar, undefined, undefined);", None),
+        ("let a = undefined, b = 2;", None),
+        // TODO: Handle parenthesized undefined
```
DonIsaac, line 546: There's a method to get the inner expression for parenthesis; I don't remember the name off the top of my head but I'll find it for you.

cblh: Thank you

jordan-boyer: if you have an expression you can use the without_parenthesized() method =)

DonIsaac (approved):
LGTM: I ran the fixer and saw no parse errors on a second pass. I'd like another review from a second pair of eyes if possible, @mysteryven think you could give me a hand?

crates/oxc_linter/src/rules/unicorn/no_useless_undefined.rs
```diff
@@ -0,0 +1,687 @@
...
+    ///
+    /// ### Example
+    /// ```javascript
+    /// // bad
+    /// let foo = undefined;
+    /// // good:
+    /// let foo;
+    /// ```
+    NoUselessUndefined,
+    correctness,
```
Boshen, line 51: Should be `pedantic`?

cblh: It should be `pedantic`, because checking function parameters will cause false positives in some functions. like：compare function names 、array function

mysteryven:
```js
test('extension is not outdated when there is no local and gallery', () => {
    const extension = instantiationService.createInstance(Extension, () => ExtensionState.Installed, () => undefined, undefined, undefined, undefined, undefined);
    assert.strictEqual(extension.outdated, false);
});
```

Our fixer need to run twice to fix this, Tested on `unicorn/no-useless-undefined`, it needs once.

Other parts are look good to me.

cblh:
> ```js
> test('extension is not outdated when there is no local and gallery', () => {
>     const extension = instantiationService.createInstance(Extension, () => ExtensionState.Installed, () => undefined, undefined, undefined, undefined, undefined);
>     assert.strictEqual(extension.outdated, false);
> });
> ```
> 
> https://github.com/microsoft/vscode/blob/c37f4a49e28831cec1aec1af0f626edaada1f7ce/src/vs/workbench/contrib/extensions/test/electron-sandbox/extension.test.ts#L29-L32
> 
> Our fixer need to run twice to fix this, tested on `unicorn/no-useless-undefined`, it needs once.
> 
> Other parts are also look good to me.

ok, I will fix it

cblh:
@mysteryven @DonIsaac 

```rust
            if i64::from(start) <= last_pos {
                return;
            }
```

https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/fixer.rs#L278

I found that `span` is an interval that is closed on the left and open on the right. It should not be `<=`, which will cause the continuous interval to be skipped. It should be `<`.

The test case needs to be modified to allow continuous interval modification, but I found that it will cause `no_useless_escape` test failure

https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/fixer.rs#L520
```rust
    const REMOVE_START: Fix = Fix::delete(Span::new(0, 4));
    const REPLACE_ID: Fix = Fix { span: Span::new(4, 10), content: Cow::Borrowed("foo") };
    fn apply_two_fix_when_the_start_the_same_as_the_previous_end() {
        let result = get_fix_result(vec![
            create_message(remove_start(), Some(REMOVE_START)),
            create_message(replace_id(), Some(REPLACE_ID)),
        ]);
        assert_eq!(result.fixed_code, TEST_CODE.replace("var answer", "foo"));
        assert!(result.fixed);
    }
```

mysteryven:
> I found that span is an interval that is closed on the left and open on the right. It should not be <=, which will cause the continuous interval to be skipped. It should be <.

Since it not the issue about current rule, let's merge this PR first.

mysteryven (approved):

PR #4087, opened 2024-07-07
feat(linter/eslint): Implement no-label-var

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/no-label-var)


crates/oxc_linter/src/rules/eslint/no_label_var.rs
```diff
@@ -0,0 +1,69 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_label_var_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint(no-label-var): Found identifier with same name as label.")
```
DonIsaac, line 9:
```suggestion
    OxcDiagnostic::warn("eslint(no-label-var): Found an identifier with the same name as a label.")
```

crates/oxc_linter/src/rules/eslint/no_label_var.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl Rule for NoLabelVar {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::LabeledStatement(labeled_stmt) = node.kind() else { return };
+
+        let scope = ctx.scopes();
+        if scope.ancestors(node.scope_id()).any(|v| {
+            scope.get_bindings(v).iter().any(|(k, _)| k.as_str() == labeled_stmt.label.name)
+        }) {
+            ctx.diagnostic(no_label_var_diagnostic(labeled_stmt.span));
```
DonIsaac, line 48: Consider adding label spans for the label and variable declaration nodes.

jelly: I looked at what the eslint rule did, and it just selects the `full label declaration`. It would be nicer to also mark the declarations I'll look into doing that.

crates/oxc_linter/src/rules/eslint/no_label_var.rs
jelly: Adjusted and then re-adjusted as I added a the label spans with text. Maybe I went a bit overboard but it looks nice:

```
  × eslint(no-label-var): Found identifier 'x' with the same name as a label.
   ╭─[foo.js:3:5]
 2 │
 3 │ var x = foo;
   ·     ┬
   ·     ╰── Identifier 'x' found here.
 4 │
 5 │
 6 │ function bar() {
 7 │ x:
   · ┬
   · ╰── Label with the same name.
 8 │   for (;;) {
   ╰────
```

DonIsaac (approved):

PR #3824, opened 2024-06-22
feat(regex_parser): Implement `RegExp` parser

leaysgur:
Part of #1164

## Progress updates 🗞️ 

Waiting for the review and advice, while thinking how to handle escaped string when `new RegExp(pat)`.

## TODOs

- [x] `RegExp(Literal = Body + Flags)#parse()` structure
- [x] Base `Reader` impl to handle both unicode(u32) and utf-16(u16) units
- [x] Global `Span` and local offset conversion
- [x] Design AST shapes
  - [x] Keep `enum` size small by `Box<'a, T>`
  - [x] Rework AST shapes 
- [x] Split body and flags w/ validating literal
- [x] Parse `RegExpFlags`
- [x] Parse `RegExpBody` = `Pattern`
- [x] Parse `Pattern` > `Disjunction`
- [x] Parse `Disjunction` > `Alternative`
- [x] Parse `Alternative` > `Term`
- [x] Parse `Term` > `Assertion`
	- [x] Parse `BoundaryAssertion`
	- [x] Parse `LookaroundAssertion`
- [x] Parse `Term` > `Quantifier`
- [x] Parse `Term` > `Atom`
	- [x] Parse `Atom` > `PatternCharacter`
	- [x] Parse `Atom` > `.`
	- [x] Parse `Atom` > `\AtomEscape`
		- [x] Parse `\AtomEscape` > `DecimalEscape`
		- [x] Parse `\AtomEscape` > `CharacterClassEscape`
			- [x] Parse `CharacterClassEscape` > `\d, \D, \s, \S, \w, \W`
			- [x] Parse `CharacterClassEscape` > `\p{UnicodePropertyValueExpression}, \P{UnicodePropertyValueExpression}`
		- [x] Parse `\AtomEscape` > `CharacterEscape`
			- [x] Parse `CharacterEscape` > `ControlEscape`
			- [x] Parse `CharacterEscape` > `c AsciiLetter`
			- [x] Parse `CharacterEscape` > `0`
			- [x] Parse `CharacterEscape` > `HexEscapeSequence`
			- [x] Parse `CharacterEscape` > `RegExpUnicodeEscapeSequence`
			- [x] Parse `CharacterEscape` > `IdentityEscape`
		- [x] Parse `\AtomEscape` > `kGroupName`
	- [x] Parse `Atom` > `[CharacterClass]`
    	- [x] Parse `[CharacterClass]` > `ClassContents` > `[~UnicodeSetsMode] NonemptyClassRanges`
[...]

PR #4096, opened 2024-07-07
feat(linter): implement eslint/no-useless-backreference

Sysix:
This is my first work on oxc in rust. So please tell me when something can be done better.
I have already some ideas to optimize to code. For that there should me mostly a ToDo in the Code.

Currently on work:
- [ ] Flag check for strings, see `report_invalid_back_reference(s)`  call
- [ ] same capture group names in alternatives
- [ ] into a negative lookaround tests
- [ ] report multiple invalid
- [ ] tests, tests, tests

There are also some outcommented test. These only makes more parser work and are (in my opinion) rare.
Please check if they are needed:

- [ ] not the global RegExp 
- [ ]  invalid syntax
   - `eslint` marks them as ok
   - currently we dont care and report it
   -  maybe wait for #3824 validator?
- [ ] contacting variables (example `p = '\\1', s = '(a)'; new RegExp(p + s);` should fail)

Linter Docs:
https://eslint.org/docs/latest/rules/no-useless-backreference


Sysix:
I encountered problems with nested (negative) look-behinds/aheads. 
I will wait for https://github.com/oxc-project/oxc/pull/3824 before reinventing the wheel.

PR #4099, opened 2024-07-07
fix(isolated_declarations): Emit computed properties when they are well known symbols

### LINKED ISSUE #4016: isolatedDeclarations does not support well-known symbols as properties
Tested version: 0.16.3
```
> require('oxc-transform').isolatedDeclaration('hello.ts', `
    export class Foo {
      [Symbol.hasInstance](): boolean {
        return false;
      }
    }
`)
{
  sourceText: 'export declare class Foo {}\n',
  errors: [
    'TS9038: Computed property names on class or object literals cannot be inferred with --isolatedDeclarations.'
  ]
}
```

Expected: 
```ts
export declare class Foo {
    [Symbol.hasInstance](): boolean;
}
```
Actual:
```ts
export declare class Foo {}
```
[TS behavior for reference](https://www.typescriptlang.org/dev/bug-workbench/?target=7&noCheck=true&ts=5.6.0-dev.20240701#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU133DpoGEALYga0dmx4wUTjQDuAUQC28aEIiRRkmdGhYCAMXgksaeHSIAPAHQFj0SDnngAZtqx0yU3aH1HzlrIYAONWNFBSMkhIUA0aGlAAbxxQUABtAGUATykAIxoSY05ggEk6GDI6RCwAXQAKAEo0DMysIujYuNBYLGgMWDpQGzISSCwAbiaAXxxRoA)

MichaelMitchell-at:
Fixes https://github.com/oxc-project/oxc/issues/4016.

To fully comply with the reference behavior of the TypeScript compiler, we need to be able to detect the case when `Symbol` or `globalThis.Symbol` do not refer to the global `Symbol`. I achieve this by creating a similar structure to `ScopeTree` called `GlobalSymbolBindingTracker` (will take suggestions for a better name) and have it visit the entire program before we begin transformation. We can keep track of which computed properties reference non-global `Symbol` by storing a set of their spans, which should each be uniquely identifying since we're in the context of a single file.
Later, if and when we transform the computed property into a type declaration, we can refer back to this set to see if it should be allowed.

### DIFF
--- a/crates/oxc_isolated_declarations/src/global_symbol_binding_tracker.rs
+++ b/crates/oxc_isolated_declarations/src/global_symbol_binding_tracker.rs
@@ -1,0 +1,251 @@
+use std::cell::Cell;
+
+#[allow(clippy::wildcard_imports)]
+use oxc_ast::ast::*;
+#[allow(clippy::wildcard_imports)]
+use oxc_ast::{visit::walk::*, Visit};
+use oxc_span::{Atom, GetSpan, Span};
+use oxc_syntax::scope::{ScopeFlags, ScopeId};
+use rustc_hash::FxHashSet;
+
+pub struct GlobalSymbolBindingTracker {
+    depth: u8,
+    symbol_binding_depth: Option<u8>,
+    global_this_binding_depth: Option<u8>,
+    computed_properties_using_non_global_symbol: FxHashSet<Span>,
+    computed_properties_using_non_global_global_this: FxHashSet<Span>,
+}
+
+impl GlobalSymbolBindingTracker {
+    pub fn new() -> Self {
+        Self {
+            depth: 0,
+            symbol_binding_depth: None,
+            global_this_binding_depth: None,
+            computed_properties_using_non_global_symbol: FxHashSet::default(),
+            computed_properties_using_non_global_global_this: FxHashSet::default(),
+        }
+    }
+
+    fn does_computed_property_reference_non_global_symbol(&self, key: &PropertyKey) -> bool {
+        self.computed_properties_using_non_global_symbol.contains(&key.span())
+    }
+
+    fn does_computed_property_reference_non_global_global_this(&self, key: &PropertyKey) -> bool {
+        self.computed_properties_using_non_global_global_this.contains(&key.span())
+    }
+
+    pub fn does_computed_property_reference_well_known_symbol(&self, key: &PropertyKey) -> bool {
+        if let PropertyKey::StaticMemberExpression(expr) = key {
+            if let Expression::Identifier(identifier) = &expr.object {
+                identifier.name == "Symbol"
+                    && !self.does_computed_property_reference_non_global_symbol(key)
+            } else if let Expression::StaticMemberExpression(static_member) = &expr.object {
+                if let Expression::Identifier(identifier) = &static_member.object {
+                    identifier.name == "globalThis"
+                        && static_member.property.name == "Symbol"
+                        && !self.does_computed_property_reference_non_global_global_this(key)
+                } else {
+                    false
+                }
+            } else {
+                false
+            }
+        } else {
+            false
+        }
+    }
+
+    fn handle_name_binding(&mut self, name: &Atom) {
+        match name.as_str() {
+            "Symbol" if self.symbol_binding_depth.is_none() => {
+                self.symbol_binding_depth = Some(self.depth);
+            }
+            "globalThis" if self.global_this_binding_depth.is_none() => {
+                self.global_this_binding_depth = Some(self.depth);
+            }
+            _ => {}
+        }
+    }
+}
+
+impl<'a> Visit<'a> for GlobalSymbolBindingTracker {
+    fn enter_scope(&mut self, _: ScopeFlags, _: &Cell<Option<ScopeId>>) {
+        self.depth += 1;
+    }
+
+    fn leave_scope(&mut self) {
+        if self.symbol_binding_depth == Some(self.depth) {
+            self.symbol_binding_depth = None;
+        }
+        if self.global_this_binding_depth == Some(self.depth) {
+            self.global_this_binding_depth = None;
+        }
+        self.depth -= 1;
+    }
+
+    fn visit_ts_type(&mut self, _: &TSType<'a>) {
+        // Optimization: we don't need to traverse down into types.
+    }
+
+    fn visit_statement(&mut self, statement: &Statement<'a>) {
+        // Optimizations: Only try to visit parts of statements containing other statements.
+        match statement {
+            Statement::TSEnumDeclaration(_)
+            | Statement::TSTypeAliasDeclaration(_)
+            | Statement::TSInterfaceDeclaration(_) => (),
+            Statement::WhileStatement(stmt) => walk_statement(self, &stmt.body),
+            Statement::DoWhileStatement(stmt) => walk_statement(self, &stmt.body),
+            Statement::ForStatement(stmt) => walk_statement(self, &stmt.body),
+            Statement::ForInStatement(stmt) => walk_statement(self, &stmt.body),
+            Statement::ForOfStatement(stmt) => walk_statement(self, &stmt.body),
+            Statement::IfStatement(stmt) => {
+                walk_statement(self, &stmt.consequent);
+                if let Some(alt) = &stmt.alternate {
+                    walk_statement(self, alt);
+                }
+            }
+            Statement::SwitchStatement(stmt) => {
+                walk_switch_cases(self, &stmt.cases);
+            }
+            _ => walk_statement(self, statement),
+        }
+    }
+
+    fn visit_binding_pattern(&mut self, pattern: &BindingPattern<'a>) {
+        if let BindingPatternKind::BindingIdentifier(ident) = &pattern.kind {
+            self.handle_name_binding(&ident.name);
+        }
+        walk_binding_pattern(self, pattern);
+    }
+
+    fn visit_assignment_pattern(&mut self, pattern: &AssignmentPattern<'a>) {
+        // If the left side of the assignment already has a type annotation, we don't need to visit the right side.
+        if pattern.left.type_annotation.is_some() {
+            self.visit_binding_pattern(&pattern.left);
+            return;
+        }
+        walk_assignment_pattern(self, pattern);
+    }
+
+    fn visit_variable_declarator(&mut self, decl: &VariableDeclarator<'a>) {
+        // If the variable already has a type annotation, we don't need to visit the initializer.
+        if decl.id.type_annotation.is_some() {
+            self.visit_binding_pattern(&decl.id);
+        } else {
+            walk_variable_declarator(self, decl);
+        }
+    }
+
+    fn visit_assignment_expression(&mut self, expr: &AssignmentExpression<'a>) {
+        // If the left side of the assignment is a member expression, it won't affect emitted declarations.
+        if expr.left.is_member_expression() {
+            return;
+        }
+        walk_assignment_expression(self, expr);
+    }
+
+    fn visit_function(&mut self, func: &Function<'a>, flags: ScopeFlags) {
+        // Async and generator functions always need explicit declarations.
+        if func.generator || func.r#async {
+            return;
+        }
+
+        if let Some(id) = func.id.as_ref() {
+            self.handle_name_binding(&id.name);
+        }
+
+        // If the function already has a return type annotation, we don't need to visit the body.
+        if func.return_type.is_some() {
+            return;
+        }
+
+        walk_function(self, func, flags);
+    }
+
+    fn visit_arrow_function_expression(&mut self, func: &ArrowFunctionExpression<'a>) {
+        // If the arrow function already has a return type annotation, we don't need to visit the body.
+        if func.return_type.is_some() {
+            return;
+        }
+
+        walk_arrow_function_expression(self, func);
+    }
+
+    fn visit_expression(&mut self, expr: &Expression<'a>) {
+        match expr {
+            Expression::ArrowFunctionExpression(_)
+            | Expression::Identifier(_)
+            | Expression::ArrayExpression(_)
+            | Expression::AssignmentExpression(_)
+            | Expression::ClassExpression(_)
+            | Expression::FunctionExpression(_)
+            | Expression::ObjectExpression(_)
+            | Expression::ParenthesizedExpression(_) => {
+                // Expressions whose types can be inferred, but excluding trivial ones
+                // whose types will never contain a Symbol computed property.
+                walk_expression(self, expr);
+            }
+            _ => (),
+        }
+    }
+
+    fn visit_declaration(&mut self, declaration: &Declaration<'a>) {
+        match declaration {
+            Declaration::VariableDeclaration(_) | Declaration::FunctionDeclaration(_) => {
+                // handled in BindingPattern and Function
+            }
+            Declaration::ClassDeclaration(decl) => {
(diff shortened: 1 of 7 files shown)


Dunqing:
Have you checked [scope.rs](https://github.com/oxc-project/oxc/blob/main/crates/oxc_isolated_declarations/src/scope.rs#L50)?

MichaelMitchell-at:
> Have you looked at [scope.rs](https://github.com/oxc-project/oxc/blob/main/crates/oxc_isolated_declarations/src/scope.rs#L50)?

At the locations that we check `report_property_key`, `self.scope` hasn't visited the necessary parts of the tree yet, so we can't consult whether the `Symbol` or `globalThis` value bindings are in scope. It seems that we don't actually walk the tree until after the object or class has already been transformed.

Stack trace for `report_property_key`
```
   2: oxc_isolated_declarations::class::<impl oxc_isolated_declarations::IsolatedDeclarations>::report_property_key
   3: oxc_isolated_declarations::types::<impl oxc_isolated_declarations::IsolatedDeclarations>::transform_object_expression_to_ts_type::{{closure}}
   4: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut
   5: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::find_map
   6: <core::iter::adapters::filter_map::FilterMap<I,F> as core::iter::traits::iterator::Iterator>::next
   7: oxc_allocator::arena::Vec<T>::from_iter_in
   8: oxc_ast::ast_builder::AstBuilder::new_vec_from_iter
   9: oxc_isolated_declarations::types::<impl oxc_isolated_declarations::IsolatedDeclarations>::transform_object_expression_to_ts_type
  10: oxc_isolated_declarations::inferrer::<impl oxc_isolated_declarations::IsolatedDeclarations>::infer_type_from_expression
  11: oxc_isolated_declarations::return_type::FunctionReturnType::infer
  12: oxc_isolated_declarations::inferrer::<impl oxc_isolated_declarations::IsolatedDeclarations>::infer_function_return_type::{{closure}}
  13: core::option::Option<T>::and_then
  14: oxc_isolated_declarations::inferrer::<impl oxc_isolated_declarations::IsolatedDeclarations>::infer_function_return_type
  15: oxc_isolated_declarations::function::<impl oxc_isolated_declarations::IsolatedDeclarations>::transform_function
  16: oxc_isolated_declarations::declaration::<impl oxc_isolated_declarations::IsolatedDeclarations>::transform_declaration
  17: oxc_isolated_declarations::module::<impl oxc_isolated_declarations::IsolatedDeclarations>::transform_export_named_declaration
  18: oxc_isolated_declarations::IsolatedDeclarations::transform_statements_on_demand
  19: oxc_isolated_declarations::IsolatedDeclarations::transform_program
  20: oxc_isolated_declarations::IsolatedDeclarations::build
```

Stack trace for `visit_binding_pattern` (where the value binding would be added)
```
   2: <oxc_isolated_declarations::scope::ScopeTree as oxc_ast::generated::visit::Visit>::visit_binding_pattern
   3: oxc_ast::generated::visit::walk::walk_formal_parameter
   4: oxc_ast::generated::visit::Visit::visit_formal_parameter
   5: oxc_ast::generated::visit::walk::walk_formal_parameter_list
   6: oxc_ast::generated::visit::Visit::visit_formal_parameter_list
   7: oxc_ast::generated::visit::walk::walk_formal_parameters
   8: oxc_ast::generated::visit::Visit::visit_formal_parameters
   9: oxc_ast::generated::visit::walk::walk_function
  10: <oxc_isolated_declarations::scope::ScopeTree as oxc_ast::generated::visit::Visit>::visit_function
  11: oxc_ast::generated::visit::walk::walk_declaration
  12: <oxc_isolated_declarations::scope::ScopeTree as oxc_ast::generated::visit::Visit>::visit_declaration
  13: oxc_isolated_declarations::IsolatedDeclarations::transform_statements_on_demand
  14: oxc_isolated_declarations::IsolatedDeclarations::transform_program
  15: oxc_isolated_declarations::IsolatedDeclarations::build
```

I'm not sure how to achieve this without either:
- updating the scope as we transform, which seems too difficult.
- take some kind of multi-phase approach where we implement a visitor that tracks value bindings and whenever it visits a computed property that's an identifier whose name is either "Symbol" or "globalThis", it sets some state on the identifier that lets it know whether or not it's referring to the global "Symbol"/"globalThis". The visitor runs before transformation begins, so that later the state on the identifier can be read from `report_property_key`.

MichaelMitchell-at:
Just brainstorming a different approach: When `Symbol` or `globalThis` has a value binding in scope we could report an error if using `Symbol.*` and `globalThis.Symbol.*` respectively as computed properties. This is more strict than the TypeScript compiler's behavior, which only reports an error if  using `Symbol.*` and `globalThis.Symbol.*` respectively as computed properties would affect emitted types. I think this could be worth considering, at least in the short-term, if we compared this behavior to the current behavior of oxc_transform:

Right now:
- oxc_transform over-reports errors when computed properties are`Symbol.*` or `globalThis.Symbol.*` and would affect type emit.
- it's not possible to rewrite the following code in a way that satisfies oxc_transform while preserving the same type semantics.
```ts
export class Foo {
  [Symbol.iterator]: ...
}
```

With this different approach:
- oxc_transform now over-reports errors when computed properties are`Symbol.*` or `globalThis.Symbol.* and `Symbol` or `globalThis` have an in-scope value binding.
- The code can easily be rewritten to avoid the errors, e.g. just use different variable names.

Dunqing:
Can you explain more what the difference between this approach and typescript's behavior?

MichaelMitchell-at:
> Can you explain more what the difference between this approach and typescript's behavior?

With an alternative approach this code would have an isolated declarations error:
```ts
export function foo(Symbol: any): void {
  const x = {[Symbol.iterator]: 1};
}
```
because `Symbol` is bound in the current scope, and we are trying to use `Symbol.*` as a property key. In TypeScript this wouldn't be an isolated declarations error because `x` clearly has no effect on emit.

Dunqing:
Let's try this alternative approach.

MichaelMitchell-at:
Ok, I might not get to this until next weekend

MichaelMitchell-at:
@Dunqing actually I thought of an approach that matches the same behavior as TypeScript, but I'm not sure if the performance/memory usage would be considered acceptable by the project. Please take a look.

crates/oxc_isolated_declarations/src/global_symbol_binding_tracker.rs
```diff
@@ -0,0 +1,176 @@
...
+        let scope = Scope::new();
+        self.levels.push(scope);
+    }
+
+    fn leave_scope(&mut self) {
+        self.levels.pop();
+    }
+
+    fn visit_ts_type(&mut self, _: &TSType<'a>) {
+        // Optimization: we don't need to traverse down into types.
```
MichaelMitchell-at, line 107: There might be other paths that can be avoided, but I'm not sure which other ones would provide as much value.

crates/oxc_isolated_declarations/src/lib.rs
```diff
@@ -77,6 +81,10 @@ impl<'a> IsolatedDeclarations<'a> {
         &mut self,
         program: &Program<'a>,
     ) -> oxc_allocator::Vec<'a, Statement<'a>> {
+        // Collect information about global Symbol usage within computed
+        // properties before performing any transformations.
+        self.global_symbol_binding_tracker.visit_program(program);
```
MichaelMitchell-at, line 86: We could probably avoid allocations for this most of the time by first checking if there is both an occurrence of overriding the global Symbol and references to Symbol *anywhere* in the program, not necessarily in the same scope, but I'm not sure what the cost of that extra traversal would be.

crates/oxc_isolated_declarations/src/global_symbol_binding_tracker.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn new() -> Self {
+        Self { has_symbol_binding: false, has_global_this_binding: false }
+    }
+}
+
+/// Linear tree of declaration scopes.
+pub struct GlobalSymbolBindingTracker<'a> {
+    levels: Vec<'a, Scope>,
+    computed_properties_using_non_global_symbol: FxHashSet<Span>,
+    computed_properties_using_non_global_global_this: FxHashSet<Span>,
```
MichaelMitchell-at, line 30: I think in theory we could just store the `start` instead of the whole `Span` and still be correct, but my guess is that it wouldn't make much difference in performance.

Dunqing:
Sorry for the late response. I will review this week. Also, I wanted to ask, do you think we merge after TypeScript 5.6 is released or before?

MichaelMitchell-at:
> Sorry for the late response. I will review this week.

No problem, no rush :)

> Also, I wanted to ask, do you think we merge after TypeScript 5.6 is released or before?

I think it's ok to wait, though I pinged the TypeScript repo to see if they might include the change in a new patch release of 5.5.

Dunqing (changes requested):
Looks like we can skip `visit_statement` in `GlobalSymbolBindingTracker`.  We just need to collect bindings.

MichaelMitchell-at:
> Looks like we can skip `visit_statement` in `GlobalSymbolBindingTracker`. We just need to collect bindings.

I believe we need to traverse into the bodies of `if`/`for`/`while` loops and there are some cases where declaration is allowed, for example `for (const Symbol of []) { ... }`.

Dunqing:
> I believe we need to traverse into the bodies of `if`/`for`/`while` loops and there are some cases where declaration is allowed, for example `for (const Symbol of []) { ... }`.

Oh yes. You are right. Can we skip some statements that are not binding, traversing the whole AST has a performance impact.

MichaelMitchell-at:
@Dunqing I made some optimizations to bring the benchmark back to baseline performance, but each one needs careful attention to make sure there isn't a scenario that I didn't think of.

Dunqing:
> @Dunqing I made some optimizations to bring the benchmark back to baseline performance, but each one needs careful attention to make sure there isn't a scenario that I didn't think of.

Well done!

Dunqing:
Can you add a test([like this](https://www.typescriptlang.org/play/?target=99&isolatedDeclarations=true&noCheck=true&ts=5.6.0-dev.20240723&noUnusedLocals=true#code/GYVwdgxgLglg9mABAIwIYCcAUBlAngW2TgBsAuRAbwF8AaRAc2LjWIBUALGAZ3OoEpKAKESJ0AUygh0SCsJGIA2nkIkAdDChj0qKHHQBdcpj7kAbnBgATRAF4AfJVpyRCxs1RtOXVcqLF1mtq6BkYmiOZWtg7UNHJUgvGCYgAeAA56UI5AA)) that uses non-global `Symbol` and `globalThis` in non-export declarations? I would like to see have any errors in this test.

Dunqing:
> I think it's ok to wait, though I pinged the TypeScript repo to see if they might include the change in a new patch release of 5.5.

I saw that v5.5.4 doesn't include this change. We have more time to improve this!

MichaelMitchell-at:
> Can you add a test([like this](https://www.typescriptlang.org/play/?target=99&isolatedDeclarations=true&noCheck=true&ts=5.6.0-dev.20240723&noUnusedLocals=true#code/GYVwdgxgLglg9mABAIwIYCcAUBlAngW2TgBsAuRAbwF8AaRAc2LjWIBUALGAZ3OoEpKAKESJ0AUygh0SCsJGIA2nkIkAdDChj0qKHHQBdcpj7kAbnBgATRAF4AfJVpyRCxs1RtOXVcqLF1mtq6BkYmiOZWtg7UNHJUgvGCYgAeAA56UI5AA)) that uses non-global `Symbol` and `globalThis` in non-export declarations? I would like to see have any errors in this test.

Sure, just added them

crates/oxc_isolated_declarations/src/lib.rs
Dunqing: Is it possible to check on demand? For example, we initially only check for the root scope. When we visit the function we will check the function. This will only check for declarations that will be exported

Dunqing (changes requested):

crates/oxc_isolated_declarations/src/lib.rs
MichaelMitchell-at: Possibly, I will think about it some more. I think one of the main challenges comes just from organizing the code and performing state management. It feels like similar to with `ScopeTree` we would need to tell `GlobalSymbolBindingTracker` when it is entering or leaving a scope, but the logic will diverge from `ScopeTree` (otherwise we would have just used `ScopeTree` for this purpose), which may result in the code looking more confusing.

(Just voicing my thoughts out loud) I think what feels redundant with my current approach is that the short-circuit optimizations I take are making assumptions about how the tree will be traversed for emitting declarations, so that logic kind of feels like it's being reproduced in two places. The code could be potentially cleaner if that logic could be pulled out. What I imagine is something like alternative `Visitor` traits that only traverse the parts of the AST relevant to emitted declarations. Such traits would already have the "optimizations" I wrote naturally built-in. What's harder for me to imagine right now is how easy it would be to re-implement declaration transforms on top of such traits.

Please let me know your thoughts.

Dunqing: > Possibly, I will think about it some more. I think one of the main challenges comes just from organizing the code and performing state management. It feels like similar to with `ScopeTree` we would need to tell `GlobalSymbolBindingTracker` when it is entering or leaving a scope, but the logic will diverge from `ScopeTree` (otherwise we would have just used `ScopeTree` for this purpose), which may result in the code looking more confusing.
> 

Only `Function` and `TSModuleDeclaration` will have nested scopes, which I don't think should be too much trouble. We could try.

> (Just voicing my thoughts out loud) I think what feels redundant with my current approach is that the short-circuit optimizations I take are making assumptions about how the tree will be traversed for emitting declarations, so that logic kind of feels like it's being reproduced in two places. The code could be potentially cleaner if that logic could be pulled out. What I imagine is something like alternative `Visitor` traits that only traverse the parts of the AST relevant to emitted declarations. Such traits would already have the "optimizations" I wrote naturally built-in. What's harder for me to imagine right now is how easy it would be to re-implement declaration transforms on top of such traits.
> 

Adding another trait instead of `Visitor` may significantly increase maintenance costs. Currently `Visitor` is automatically generated with `ast-codegen`

MichaelMitchell-at: > Adding another trait instead of Visitor may significantly increase maintenance costs

That's a good point. I suppose it would be necessary for such a trait to extend `Visit` and only override a small subset of the method implementations.

> Only `Function` and `TSModuleDeclaration` will have nested scopes, which I don't think should be too much trouble. We could try.

I may not be able to get to this soon, but as you said, we have time!

DonIsaac:
@Dunqing what do we need to do to get this over the finish line?

Dunqing:
Feel free to convert it to open and ping me if it ready to review

MichaelMitchell-at:
I'm about to go on vacation so will probably not be able to touch this for at least another couple weeks. If anyone wants to iterate on this in the meantime, please feel free :)

Dunqing:
> I'm about to go on vacation so will probably not be able to touch this for at least another couple weeks. If anyone wants to iterate on this in the meantime, please feel free :)

Thank you for your hard work on Isolated Declaration! Have a nice holiday!!

Boshen:
Close as stale. Feel free to comeback when ts is updated.

PR #4107, opened 2024-07-08
perf(semantic): keep a single map of unresolved references

lucab:
This reworks `ScopeTree` in order to keep a single (root) map of unresolved references. The `SemanticBuilder` keeps track of all intermediate ones while walking scopes, and it can get rid of all non-root ones once done.

Ref: https://github.com/oxc-project/backlog/issues/32

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -9,6 +9,7 @@ use oxc_cfg::{
     IterationInstructionKind, ReturnInstructionKind,
 };
 use oxc_diagnostics::OxcDiagnostic;
+use oxc_index::{index_vec, IndexVec};
 use oxc_span::{CompactStr, SourceType, Span};
 use oxc_syntax::{module_record::ModuleRecord, operator::AssignmentOperator};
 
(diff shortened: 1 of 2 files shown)


Boshen (approved):
Clever! Thank you!

Boshen:
@lucab Given your recent contributions, I have sent you invitation to the repo with write permission so you can use graphite. https://github.com/oxc-project/oxc/invitations

lucab:
For reference, I tried to push `unresolved_references` directly into `Semantic` (as suggested in https://github.com/oxc-project/backlog/issues/32) but eventually found out that at least the JSX transformer needs to access it through `ScopeTree` (it doesn't seem to have access to `Semantic` in its traversing context):
https://github.com/oxc-project/oxc/blob/c3f08ce8e02f45dd788130b63ebba3b0636f5a62/crates/oxc_transformer/src/react/jsx.rs#L983-L992

overlookmotel:
Nice work @lucab!

I can see potential for one further optimization...

When exiting a scope, `UnresolvedReferences` for that scope is drained, and then that `FxHashMap` is never used again. We end up creating a short-lived `FxHashMap` for every scope.

Instead of `unresolved_references` being a `Vec` indexed by scope ID, could it be a stack? When you exit a scope, `UnresolvedReferences` for that scope would be cleared, but left in `unresolved_references`. Then when entering next sibling scope, you reuse the same `UnresolvedReferences` as before. This would reduce allocations, and also you'd mostly be using a `FxHashMap` which is already warm in cache.

Does that make sense, and do you think it could work/make a difference?

overlookmotel:
> I tried to push `unresolved_references` directly into `Semantic`

Personally, I think `ScopeTree` is the best place for it. Unresolved references could be considered to be the "global scope".

PR #4109, opened 2024-07-08
feat(linter/eslint): Implement prefer-numeric-literals

jelly:
Rule Detail:
[link](https://eslint.org/docs/latest/rules/prefer-numeric-literals)

---

This lacks the fix, which I intend to work on as a follow up or if preferred I can leave the PR open an hack on it this week.

One interesting note is that `eslint(radix)` does not handle:

```
        "function *f(){ yield(Number).parseInt('11', 2) }", // { "ecmaVersion": 6 },
```

Maybe it is an idea to introduce a helper `is_parseint_stmt` which returns the `call_expr`? As the code is almost identical to the `radix` rule.


jelly:
Just send https://github.com/oxc-project/oxc/pull/4110 for the radix issue.

jelly:
Implemented the fixer, but I need to figure out if there is space on the start of the span and end of the previous span are only one apart and I'm not sure how to do that.

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
```diff
@@ -0,0 +1,301 @@
...
+    PreferNumericLiterals,
+    style,
+);
+
+impl Rule for PreferNumericLiterals {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::CallExpression(call_expr) = node.kind() {
+            match &call_expr.callee.without_parenthesized() {
+                Expression::Identifier(ident) if ident.name == "parseInt" => {
+                    if ctx.symbols().get_symbol_id_from_name("parseInt").is_none() {
```
mysteryven, line 58: This code may let our rule doesn't work (Even thought I think rare people who will write code like this):

```js
// 1. Use `parseInt` here
parseInt();

{
  // 2. And then define another function called `parseInt` in this scope
  function parseInt() {
    // snip
  }
}
```

What about using `get_symbol_id_of_variable(ident, ctx).is_none()`

https://github.com/oxc-project/oxc/blob/2334515247f2db6eca56e8df4522dd11d4046057/crates/oxc_linter/src/ast_util.rs#L261-L269

mysteryven:
> Implemented the fixer, but I need to figure out if there is space on the start of the span and end of the previous span are only one apart and I'm not sure how to do that.

Sound difficult for us, do you have any thoughts @DonIsaac

DonIsaac:
> > Implemented the fixer, but I need to figure out if there is space on the start of the span and end of the previous span are only one apart and I'm not sure how to do that.
> 
> 
> 
> Sound difficult for us, do you have any thoughts @DonIsaac 

Just saw this, my bad! I'll review this tomorrow.

jelly:
> > > Implemented the fixer, but I need to figure out if there is space on the start of the span and end of the previous span are only one apart and I'm not sure how to do that.
> > 
> > 
> > Sound difficult for us, do you have any thoughts @DonIsaac
> 
> Just saw this, my bad! I'll review this tomorrow.

So I know now that I can look up the parent_node easily in the fixer. But I think the AST does not really record the previous node.

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
```diff
@@ -0,0 +1,301 @@
...
+                            }
+                        }
+                    }
+                }
+                Expression::ChainExpression(chain_expr) => {
+                    if let Some(member_expr) = chain_expr.expression.as_member_expression() {
+                        if let Expression::Identifier(ident) = &member_expr.object() {
+                            if ident.name == "Number"
+                                && member_expr.static_property_name() == Some("parseInt")
+                                && ctx.symbols().get_symbol_id_from_name("Number").is_none()
```
DonIsaac, line 88: Refactor all of these to reduce code duplication.

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
```diff
@@ -0,0 +1,301 @@
...
+        return;
+    }
+
+    let string_arg = &call_expr.arguments[0];
+    if !is_string_type(string_arg) || !string_arg.to_expression().is_immutable_value() {
+        return;
+    }
+
+    let radix_arg = &call_expr.arguments[1];
+    if radix_arg.to_expression().is_specific_raw_number_literal("2") {
```
DonIsaac, line 135: This can be rewritten to reduce duplication as well

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
```diff
@@ -0,0 +1,301 @@
...
+        // (
+        //     "function *f(){ yield/**/Number.parseInt('67', 8) }",
+        //     "function *f(){ yield/**/0o67 }",
+        //     None,
+        // ),
+        // ("function *f(){ yield(parseInt('A', 16)) }", "function *f(){ yield(0xA) }", None),
+        ("parseInt('11', 2)+5", "0b11+5", None),
+        ("Number.parseInt('17', 8)+5", "0o17+5", None),
+        ("parseInt('A', 16)+5", "0xA+5", None),
+        // ("parseInt('11', 2)in foo", "0b11 in foo", None),
```
DonIsaac, line 279: Why don't these cases work?

jelly: These cases don't work as the fixer will create invalid Javascript, for example:
```
  left: "0b11 in foo"
 right: "0b11in foo"
```

jelly: Decided to drop the code fixer for now, I can work on that as an improvement.

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
jelly: I am not sure if I understand what this does, is that an extra check to check if `parseInt` is not undefined?

jelly:
I've pushed work on a fixer to https://github.com/jelly/oxc/tree/eslint/numeric-literals-fixer, the tests pass but the eslint tests aren't great and allow for creation of invalid syntactical code for example:

`parseInt('', 16);` and `parseInt('1234.5', 8) => 0o1234.5`. So I'll need to make the fixer conditional, not entirely sure yet how I check fractions.

mysteryven:
> not entirely sure yet how I check fractions.

Looks [`from_str_radix`](https://doc.rust-lang.org/nightly/std/primitive.i64.html#method.from_str_radix) can solve this: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb53c9ce0762620dd8d97fcb6d9e1e62).

crates/oxc_linter/src/rules/eslint/prefer_numeric_literals.rs
mysteryven: Sorry for my unclear comment, you can add this to your fail tests, the `parseInt` in first line should be detected.

```javascript
  parseInt("767", 8) === 503; // should report
  function foo() {
      function parseInt() {
          throw new Error()
      }   
  }
```

jelly:
> > not entirely sure yet how I check fractions.
> 
> Looks [`from_str_radix`](https://doc.rust-lang.org/nightly/std/primitive.i64.html#method.from_str_radix) can solve this: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb53c9ce0762620dd8d97fcb6d9e1e62).

Thanks! That would solve everything I needed for the fixer!

Boshen:
@mysteryven Would you like to help out and land this?

DonIsaac:
> I've pushed work on a fixer to https://github.com/jelly/oxc/tree/eslint/numeric-literals-fixer, the tests pass but the eslint tests aren't great and allow for creation of invalid syntactical code for example:
> 
> `parseInt('', 16);` and `parseInt('1234.5', 8) => 0o1234.5`. So I'll need to make the fixer conditional, not entirely sure yet how I check fractions.

If you cannot figure out a perfectly safe fix, consider making it dangerous. https://github.com/oxc-project/oxc/blob/732f4e2591b01e1bc658c7ae16fdae9e3172c969/crates/oxc_linter/src/fixer/fix.rs#L178

mysteryven (approved):
Let's merge this, the fixer part will be the next separate PR.

Issue #4111, opened 2024-07-08
Oxlint crashes, due stack overflow and panics


qarmin:
Oxlint version - 0f026089d104800b963568f6703f91b737ac188f

command
```
oxlint file -D all -D nursery --fix
```

```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
timeout: the monitored command dumped core
```

folder may contain also other crashes like 
```
2024-07-08T09:04:58.1425461Z thread 'main' panicked at crates/oxc_diagnostics/src/reporter/graphical.rs:28:65:
2024-07-08T09:04:58.1427255Z called `Result::unwrap()` on an `Err` value: Error
2024-07-08T09:04:58.1430402Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-07-08T09:04:58.1435215Z timeout: the monitored command dumped core
```
```
2024-07-08T09:41:20.0318660Z thread '<unnamed>' panicked at crates/oxc_linter/src/rules/eslint/array_callback_return/mod.rs:160:69:
2024-07-08T09:41:20.0320174Z called `Option::unwrap()` on a `None` value
2024-07-08T09:41:20.0321273Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-07-08T09:41:20.0322577Z timeout: the monitored command dumped core
```
```
2024-07-08T07:34:13.8760591Z thread '<unnamed>' panicked at crates/oxc_linter/src/rules/unicorn/prefer_dom_node_dataset.rs:117:36:
2024-07-08T07:34:13.8762530Z byte index 5 is not a char boundary; it is inside '\u{8d}' (bytes 4..6) of `contnteditable`
2024-07-08T07:34:13.8767065Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-07-08T07:34:13.8767997Z timeout: the monitored command dumped core
```

7zip file(needs to remove .zip extension) - [zzz.7z.zip](https://github.com/user-attachments/files/16126175/zzz.7z.zip)

Files with crash output in each file  - [A.zip](https://github.com/user-attachments/files/16142825/A.zip)

rzvxa:
It is probably related to https://github.com/oxc-project/backlog/issues/58

rzvxa:
@qarmin I'd like to investigate this issue.
[...]

PR #4121, opened 2024-07-08
fix(linter): panic in `get_enclosing_function`

DonIsaac:
Fixes one of the panics described in #4111

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -201,7 +201,7 @@ pub fn get_enclosing_function<'a, 'b>(
         {
             return Some(current_node);
         }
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/ast_util.rs
```diff
@@ -201,7 +201,7 @@ pub fn get_enclosing_function<'a, 'b>(
         {
```
rzvxa, line 201: It isn't a change request, Mostly thinking out loud; While you are at it maybe we can also replace this condition with `current_node.kind().is_function_like()` for the sake of consistency.

DonIsaac: This sounds like a good idea; I can make a separate PR.

PR #4098, opened 2024-07-07
feat(ast_codegen): add ast builder generator.

rzvxa:

PR #4090, opened 2024-07-07
feat(span): add various implementations of `FromIn` for `Atom`.

rzvxa:

PR #4088, opened 2024-07-07
feat(allocator): introduce `FromIn` and `IntoIn` traits.

rzvxa:

PR #4122, opened 2024-07-08
refactor(ast/ast_builder)!: shorter allocator utility method names.

rzvxa:
This PR serves two purposes, First off it would lower the amount of characters we have to type in for a simple operation such as wrapping an expression in a vector. Secondly, it would follow the generated names more closely since nowhere else in the builder we do have `new_xxx`, We always say `xxx` since a builder always constructs something.

```
new_vec -> vec
new_vec_single -> vec1*
new_vec_from_iter -> vec_from_iter
new_vec_with_capacity -> vec_with_capacity
new_str -> str
new_atom -> atom
```

`*` This one is the main motivation behind this PR, It saves 10 characters!

### DIFF
--- a/crates/oxc_ast/src/ast_builder_impl.rs
+++ b/crates/oxc_ast/src/ast_builder_impl.rs
@@ -27,34 +27,34 @@ impl<'a> AstBuilder<'a> {
     }
 
     #[inline]
-    pub fn new_vec<T>(self) -> Vec<'a, T> {
+    pub fn vec<T>(self) -> Vec<'a, T> {
         Vec::new_in(self.allocator)
     }
 
     #[inline]
-    pub fn new_vec_with_capacity<T>(self, capacity: usize) -> Vec<'a, T> {
+    pub fn vec_with_capacity<T>(self, capacity: usize) -> Vec<'a, T> {
         Vec::with_capacity_in(capacity, self.allocator)
     }
 
     #[inline]
-    pub fn new_vec_single<T>(self, value: T) -> Vec<'a, T> {
-        let mut vec = self.new_vec_with_capacity(1);
+    pub fn vec1<T>(self, value: T) -> Vec<'a, T> {
+        let mut vec = self.vec_with_capacity(1);
         vec.push(value);
         vec
     }
 
     #[inline]
-    pub fn new_vec_from_iter<T, I: IntoIterator<Item = T>>(self, iter: I) -> Vec<'a, T> {
+    pub fn vec_from_iter<T, I: IntoIterator<Item = T>>(self, iter: I) -> Vec<'a, T> {
         Vec::from_iter_in(iter, self.allocator)
     }
 
     #[inline]
-    pub fn new_str(self, value: &str) -> &'a str {
+    pub fn str(self, value: &str) -> &'a str {
         String::from_str_in(value, self.allocator).into_bump_str()
     }
 
     #[inline]
-    pub fn new_atom(self, value: &str) -> Atom<'a> {
+    pub fn atom(self, value: &str) -> Atom<'a> {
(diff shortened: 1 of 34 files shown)


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4122?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4122** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4117** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4116** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4116?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3890** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/3890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4098** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4090** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4088** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4088?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
I agree with the logic of this change. I'd suggest `vec_from_iter` rather than `vec_from`. That might be an abbreviation too far (from what?) - and also it's not much used.

rzvxa:
> I agree with the logic of this change. I'd suggest `vec_from_iter` rather than `vec_from`. That might be an abbreviation too far (from what?) - and also it's not much used.

Sorry, my summary didn't cover all of the changes, I've changed the signature of our former `new_vec_from_iter` to this:

```rust
    #[inline]
    pub fn vec_from<T, I: IntoIterator<Item = T>>(self, iter: I) -> Vec<'a, T> {
        Vec::from_iter_in(iter, self.allocator)
    }
```


So it's no longer just "from_iter", It accepts anything that can be iterated even something like `Some(T)`. That's why I've made the name more generic as well.

overlookmotel:
I still prefer `vec_from_iter` as it [matches std's naming convention](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_iter). We could conceivably want to add e.g. `vec_from_slice` later on. But I don't feel so strongly about it.

rzvxa:
> I agree with the logic of this change. I'd suggest `vec_from_iter` rather than `vec_from`. That might be an abbreviation too far (from what?) - and also it's not much used.

At first, I thought it was accepting iterators before so I sent a false comment. I removed it to write an accurate answer instead.

I chose this name because it doesn't accept `Iterator`, It accepts `IntoIterator` which can be a variety of things such as `Some(T)` so that's why I went with this generic name. Let me know if you don't follow my reasoning or if you have a relatively strong objection, I'm flexible for changing it.

rzvxa:
> I still prefer `vec_from_iter` as it [matches std's naming convention](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_iter) (`Vec::from_iter` also accepts anything which can be iterated). We could conceivably want to add e.g. `vec_from_slice` later on. But I don't feel so strongly about it.

I agree with your reasoning, Just pushed a new commit addressing this.

PR #4125, opened 2024-07-08
feat(semantic): check for abstract initializations and implementations

DonIsaac:


rzvxa (approved):

tasks/coverage/parser_typescript.snap
```diff
@@ -11074,6 +11069,46 @@ Expect to Parse: "conformance/salsa/plainJSRedeclare3.ts"
...
+
+  × TS(1245): Method 'bb' cannot have an implementation because it is marked abstract.
+   ╭─[conformance/classes/classDeclarations/classAbstractKeyword/classAbstractAccessor.ts:5:17]
+ 4 │    abstract set b(x: string);
+ 5 │    abstract set bb(x: string) {} // error
+   ·                 ──
+ 6 │ }
+   ╰────
+
+  × TS(1245): Method 'constructor' cannot have an implementation because it is marked abstract.
```
rzvxa, line 11104: Is this intentional? Shouldn't it raise a syntax error instead? I think the parser is mistaking `abstract constructor` as an `abstract <method>`.
https://www.typescriptlang.org/play/?#code/IYIwzgLgTsDGEAJYBthjAgggg3g0kM8SA9gHaECu8JUAFAJQ4C+zQA

DonIsaac: I believe you are correct; I'll add checks to the parser

rzvxa: Oh Thanks, I'll close #4128 in favor of your future PR since you are already knee-deep in the modifiers😆

rzvxa:
@DonIsaac I think for the constructors we should also add `verify_modifiers` in the parse phase. Let me know what are your thoughts on it.

PR #4132, opened 2024-07-09
refactor(ast)!: store span details inside comment struct

### LINKED ISSUE #4069: Include start offset in `Comment`
#4045 replaced `TriviasMap` with a sorted `Vec` (bravo @lucab!)

After that change, there's no need to use a `Vec<(u32, Comment)>` - can change it to `Vec<Comment>` and move `start` field into `Comment` itself. Or, probably even better, make it `struct Comment { span: Span, kind: CommentKind }` to match AST nodes.

This would feel more natural to me - `start` was only external to `Comment` previously as it was key for the map.

Side note: I'm not sure if `SortedComments` should be a boxed slice rather than a `Vec`. Possible that some transforms may want to add/remove comments.

lucab:
This tweaks `Comment` definition in order to internally store the start and end position of its span.

Closes: https://github.com/oxc-project/oxc/issues/4069

### DIFF
--- a/crates/oxc_ast/src/trivia.rs
+++ b/crates/oxc_ast/src/trivia.rs
@@ -79,8 +79,8 @@ impl Trivias {
         }))
     }
 
-    pub fn comments(&self) -> impl Iterator<Item = (CommentKind, Span)> + '_ {
-        self.comments.iter().map(|comment| (comment.kind, comment.span))
+    pub fn comments(&self) -> impl Iterator<Item = &Comment> {
+        self.comments.iter()
     }
 
     pub fn comments_range<R>(&self, range: R) -> CommentsRange<'_>
--- a/crates/oxc_ast/src/trivia.rs
+++ b/crates/oxc_ast/src/trivia.rs
@@ -12,13 +12,20 @@ use oxc_span::Span;
 #[derive(Debug, Clone, Copy)]
 pub struct Comment {
     pub kind: CommentKind,
-    pub end: u32,
+    /// The span of the comment text (leading/trailing delimiters not included).
+    span: Span,
 }
 
 impl Comment {
     #[inline]
-    pub fn new(end: u32, kind: CommentKind) -> Self {
-        Self { kind, end }
+    pub fn new(start: u32, end: u32, kind: CommentKind) -> Self {
+        let span = Span::new(start, end);
+        Self { kind, span }
+    }
+
+    /// Return the span of the comment text (without delimiters).
+    pub fn span(&self) -> &Span {
+        &self.span
     }
 }
 
@@ -41,7 +48,7 @@ impl CommentKind {
 }
 
 /// Sorted set of unique trivia comments, in ascending order by starting position.
-pub type SortedComments = Box<[(u32, Comment)]>;
+pub type SortedComments = Box<[Comment]>;
 
 #[derive(Debug, Clone, Default)]
 pub struct Trivias(Arc<TriviasImpl>);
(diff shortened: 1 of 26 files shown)


crates/oxc_ast/src/trivia.rs
```diff
@@ -12,13 +12,20 @@ use oxc_span::Span;
...
     #[inline]
-    pub fn new(end: u32, kind: CommentKind) -> Self {
-        Self { kind, end }
+    pub fn new(start: u32, end: u32, kind: CommentKind) -> Self {
+        let span = Span::new(start, end);
+        Self { kind, span }
+    }
+
+    /// Return the span of the comment text (without delimiters).
+    pub fn span(&self) -> &Span {
```
Boshen, line 27: Shall we try adding `#[inline]` here? `.span()` is called a lot in the jsdoc builder, hence the performance regression (my guess).

Boshen:
Weird, where's the performance regression coming from ...

crates/oxc_ast/src/trivia.rs
overlookmotel: Definitely `#[inline]`. Also, `Span` is `Copy`, so we should also return `Span` not `&Span`.

crates/oxc_ast/src/trivia.rs
```diff
@@ -12,13 +12,20 @@ use oxc_span::Span;
 #[derive(Debug, Clone, Copy)]
 pub struct Comment {
     pub kind: CommentKind,
-    pub end: u32,
+    /// The span of the comment text (leading/trailing delimiters not included).
+    span: Span,
```
overlookmotel, line 16: What's the reason for making the `span` field private?

crates/oxc_ast/src/trivia.rs
```diff
@@ -126,14 +137,14 @@ impl<'a> CommentsRange<'a> {
...
+    type Item = &'c Comment;
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.current_start < self.current_end {
-            for (start, comment) in &self.comments[self.current_start..self.current_end] {
+            for comment in &self.comments[self.current_start..self.current_end] {
                 self.current_start = self.current_start.saturating_add(1);
-                if self.range.contains(start) {
-                    return Some((start, comment));
+                if self.range.contains(&comment.span.start) {
```
overlookmotel, line 146: I think this can be `comment.span.start` (without `&`).

overlookmotel:
Is `comments_range` on a hot path? If so, we could experiment with storing comments in SOA style with separate `Vec`s for `start`, `end` and `kind`. This might speed up searching (or maybe not, I can't really imagine what the memory access patterns are).

I still think the `Comment` type which is returned to caller should be as it is after this PR (as it's the most user-friendly format), but the backing storage could be different.

Boshen:
> Is comments_range on a hot path? 

It seems to be on the wrong hot path, I need to fix it. The problem is unrelated to this PR.

crates/oxc_ast/src/trivia.rs
lucab: There is a `pub` to private change in here too (as @overlookmotel already pointed out below) which is possibly preventing some cross-crate inlining/optimization which was previously possible.
I'll try to expose it again in the next round and see if that fixes the benchmark regression.

crates/oxc_ast/src/trivia.rs
lucab: That doesn't seem to build, the trait method takes a reference there: https://doc.rust-lang.org/std/ops/trait.RangeBounds.html#method.contains

Boshen:
Comments processing shouldn't be on a hot path, merging this one and fixing the hot path problem in https://github.com/oxc-project/oxc/pull/4140

PR #4162, opened 2024-07-10
perf(semantic): store unresolved refs in a stack

lucab:
This tweaks `SemanticBuilder` logic in order to accumulate unresolved
references in a stack, getting rid of the previous index-vector which
is not required under the current access pattern.

Ref: https://github.com/oxc-project/oxc/pull/4107#issuecomment-2214167393

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -9,7 +9,6 @@ use oxc_cfg::{
     IterationInstructionKind, ReturnInstructionKind,
 };
 use oxc_diagnostics::OxcDiagnostic;
-use oxc_index::{index_vec, IndexVec};
 use oxc_span::{CompactStr, SourceType, Span};
 use oxc_syntax::{module_record::ModuleRecord, operator::AssignmentOperator};
 
@@ -66,9 +65,9 @@ pub struct SemanticBuilder<'a> {
     pub nodes: AstNodes<'a>,
     pub scope: ScopeTree,
     pub symbols: SymbolTable,
-    /// NOTE(lucab): lazy vector, always access this through the
-    /// `unresolved_references_by_scope()` helper.
-    unresolved_references: IndexVec<ScopeId, UnresolvedReferences>,
+
+    // LIFO stack used to accumulate unresolved refs while traversing scopes.
+    unresolved_references: Vec<UnresolvedReferences>,
 
     pub(crate) module_record: Arc<ModuleRecord>,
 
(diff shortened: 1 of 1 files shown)


lucab:
* **#4162** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Nice work! However, I think we can potentially do even better.

What I had in mind is that you never call `unresolved_references.pop()`. Instead, hold on to the hash map and re-use it. That way we avoid generating (and allocating) new hash map when entering a scope most of the time.

`unresolved_references` is `Vec<UnresolvedReferences>`, indexed by *scope depth*.

e.g. with this input:

```js
function foo() {}
function bar() {}
```

* Enter top level scope:
  * Depth is 0.
  * Push a `UnresolvedReferences` hash map to `unresolved_references`.
  * Current hashmap is `unresolved_references[0]`.
* Enter scope for `foo`:
  * Increment depth to 1.
  * Push another `UnresolvedReferences` hash map to `unresolved_references`.
  * Current hashmap is `unresolved_references[1]`.
* Exit scope for `foo`:
  * Clear current hashmap (`unresolved_references[1]`).
  * Decrement depth to 0.
  * Do NOT pop from `unresolved_references`.
  * Current hashmap is now `unresolved_references[0]`.
* Enter scope for `bar`:
  * Increment depth to 1.
  * Do NOT push to `unresolved_references`.
  * Current hashmap is `unresolved_references[1]` (i.e. reuse same hash map that was used for `foo`)
 * Exit scope for `bar`: same as for `foo`
 * Exit top level scope: same
 
Does this make any sense? I *think* it'll be a further perf boost due to reducing allocations, and reusing allocations which are already warm in the cache.

lucab:
@overlookmotel I see, you want to reuse previously allocated hashmaps (sorry I missed that point from your original request). Mind if I try to do that in a separate PR after landing this one, so that it's easier to see the performance delta?

overlookmotel:
> Do you mind if I try to do that in a separate PR after landing this one

Yes, sure. Whichever way you prefer. In any case, this PR is already a really nice perf boost. Also... feel free *not* to do it at all! It's completely up to you what you want to work on. I'm just making suggestions, not commands!

overlookmotel (approved):

lucab:
Yes, this one is ready to merge.

PR #4192, opened 2024-07-11
refactor(semantic): correct scope in CatchClause

### LINKED ISSUE #4186: ast-codegen: visit `it.body.body` or`it.body` in CatchClause?
### CatchClause 
https://github.com/oxc-project/oxc/blob/07506ece83e7101f618093afa74be23babfc58c3/crates/oxc_ast/src/ast/js.rs#L1307-L1313

### In semantic builder:
https://github.com/oxc-project/oxc/blob/07506ece83e7101f618093afa74be23babfc58c3/crates/oxc_semantic/src/builder.rs#L1351-L1363

### In `Visit`
https://github.com/oxc-project/oxc/blob/07506ece83e7101f618093afa74be23babfc58c3/crates/oxc_ast/src/generated/visit.rs#L3691-L3707

The difference between them is that one calls `visit_statements(it.body.body)` and the other `visit_block_statements(it.body)`

Because we will create a scope within the `CatchClause`, we do not need to create a scope again in the `BlockStatement`.

Is there a better way to resolve this problem? Such as changing the `body` to `Vec<'a, Statements<'a>> `

Dunqing:
close: #4186

CatchClause has two scopes. The first one is `CatchClause`, which will add a `CatchParameter` to it. The second one is `Block`, which will add binding that declares in the current block scope.

The spec has a syntax error about `CatchParameter`
- It is a Syntax Error if any element of the BoundNames of CatchParameter also occurs in the LexicallyDeclaredNames of Block.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1300,7 +1300,7 @@ pub struct TryStatement<'a> {
 }
 
 #[visited_node]
-#[scope(if(self.param.is_some()))]
+#[scope]
 #[derive(Debug)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[cfg_attr(feature = "serialize", serde(tag = "type"))]
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1348,19 +1348,6 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         self.leave_node(kind);
     }
 
-    fn visit_catch_clause(&mut self, clause: &CatchClause<'a>) {
-        let kind = AstKind::CatchClause(self.alloc(clause));
-        self.enter_scope(ScopeFlags::empty(), &clause.scope_id);
-        clause.scope_id.set(Some(self.current_scope_id));
-        self.enter_node(kind);
-        if let Some(param) = &clause.param {
-            self.visit_catch_parameter(param);
-        }
-        self.visit_statements(&clause.body.body);
-        self.leave_node(kind);
-        self.leave_scope();
-    }
-
     fn visit_finally_clause(&mut self, clause: &BlockStatement<'a>) {
         let kind = AstKind::FinallyClause(self.alloc(clause));
         self.enter_scope(ScopeFlags::empty(), &clause.scope_id);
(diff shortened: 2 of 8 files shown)


Dunqing:
* **#4192** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4192?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -321,7 +321,17 @@ impl<'a> SemanticBuilder<'a> {
         excludes: SymbolFlags,
         report_error: bool,
     ) -> Option<SymbolId> {
-        let symbol_id = self.scope.get_binding(scope_id, name)?;
+        let symbol_id = self.scope.get_binding(scope_id, name).or_else(|| {
+            self.scope.get_parent_id(scope_id).and_then(|pid| {
+                let flags = self.scope.get_flags(pid);
+                // It is a Syntax Error if any element of the BoundNames of CatchParameter also occurs in the LexicallyDeclaredNames of Block.
```
Boshen, line 327: I don't think this check is in the right place 🤔

Dunqing: I tried another way. Can you take a look at https://github.com/oxc-project/oxc/pull/4192/commits/3ed49909c2a226961b92601e11cee42108c4973a?

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1300,7 +1300,7 @@ pub struct TryStatement<'a> {
 }
 
 #[visited_node]
-#[scope(if(self.param.is_some()))]
```
overlookmotel, line 1303: @Dunqing What was reason for removing `if(self.param.is_some())`? If no catch param, we don't need a scope, right?

I think it's arguable that it would be simpler if we didn't have any conditional scopes, but if we do that, probably we should do it consistently.

Dunqing: Yes, just to be consistent with the spec

overlookmotel: OK, so is plan to make all scopes unconditional?

overlookmotel: Personally, I don't think we *need* to adhere to the spec. The spec defines what *observable behavior* must be, but implementations are free to achieve that through any means they choose. V8, for example, diverges from the example algorithms in the spec very heavily for purpose of better performance.

Dunqing: I agree. I have found that there is a performance impact of `boa` reporting errors according to the specification

https://github.com/boa-dev/boa/blob/5b1621abd2b702b99bd6b4491973095f9b27c362/core/parser/src/parser/statement/try_stm/catch.rs#L92-L97

overlookmotel: Thanks for coming back. OK. I would suggest reverting this to `#[scope(if(self.param.is_some()))]` for now then. If we keep scopes as conditional, it's quite easy to make them all unconditional later. But it's harder to go in other direction - you have to go through all the scopes and reason about which *could* be conditional.

Personally I think the most important thing either way is to be consistent, as it's one less oddity to reason about.

But that's just my opinion. You can disagree, and I won't argue further.

Dunqing: Ok. I will send a PR later

Issue #4161, opened 2024-07-10
Should class declarations contain a binding for class name in scope for the class?


overlookmotel:
Should class declarations contain a binding for class name in the scope for the class?

`Klass` in example below has 2 bindings - 1st at top level, 2nd within the class body.

```js
class Klass {
    static tryWrite() {
        Klass = 123;
    }
}

const CopyOfKlass = Klass;

// Reassignment of `Klass` in top level scope succeeds
Klass = 123;
console.log(Klass); // Logs 123

// Reassignment of `Klass` within class fails.
// `TypeError: Assignment to constant variable at Klass.tryWrite`
CopyOfKlass.tryWrite();
```

The outer binding is reassignable, whereas the binding within the class is read-only.

Dunqing:
In the spec. The `class` doesn't have a scope.

In the `TypeScript` container. Will create a container for `class`. But doesn't add a binding for class name to the `container`

overlookmotel:
> In the spec. The class doesn't have a scope.

Is it possible for the spec to be wrong?? (or wrong in terms of what we need). I do think in this case there are 2 bindings. The above example demonstrates this.

Dunqing:
> Is it possible for the spec to be wrong?? (or wrong in terms of what we need). 

Unlikely. But we are aligning with the TypeScript container now.

I do think in this case there are 2 bindings. The above example demonstrates this.

IMO, we don't need to add a binding to the class scope. The error in the above example is a runtime error that we don't need to care about

rzvxa:
I think the class identifier much like the `this` keyword means something special in the context of class and that's why you can't reassign it

overlookmotel:
Maybe the spec has some "special" way of thinking about this. But for our purposes, you have 2 variables with the same name which are independent of each other and one shadows the other - that sounds a lot like scopes to me!

overlookmotel:
> The error in the above example is a runtime error that we don't need to care about

I think we will need to care about it in future. e.g. a transform which converts a `class` into a `function`. And minifier may be able to perform optimizations if it understands that one binding is read-only.

rzvxa:
But it will cause the scopes to be wrong in some cases.

```js
class Klass {
  y() {
    class Foo {
      y() {
        console.log(Klass, Foo);
      }
    }
    return new Foo();
  }
}

new Klass().y().y();
```

What should `Klass` and `Foo` point to in the `console.log` line?

overlookmotel:
[...]

PR #4195, opened 2024-07-11
feat(semantic): align class scope with typescript

Dunqing:
```ts
class Klass <T>   extends Root       <R>                   {}
//    ^^^^^ ^^^           ^^^^       ^^^                   ^^
//    id type_paramter super_class super_type_parameters  body
```
I reorder fields according to the order above

The class scope is not defined in the spec. But we need to create a scope for `class` to store `TypeParamters`


tasks/coverage/parser_test262.snap
```diff
@@ -2,12 +2,186 @@ commit: a1587416
...
 Expect Syntax Error: "language/import/import-assertions/json-invalid.js"
 Expect Syntax Error: "language/import/import-assertions/json-named-bindings.js"
 Expect Syntax Error: "language/import/import-attributes/json-invalid.js"
 Expect Syntax Error: "language/import/import-attributes/json-named-bindings.js"
+Expect to Parse: "built-ins/Function/prototype/toString/class-expression-explicit-ctor.js"
+
+  × Identifier `A` has already been declared
+    ╭─[built-ins/Function/prototype/toString/class-expression-explicit-ctor.js:10:5]
+  9 │ 
+ 10 │ let A = /* before */class /* a */ A /* b */ extends /* c */ B /* d */ { /* e */ constructor /* f */ ( /* g */ ) /* h */ { /* i */ ; /* j */ } /* k */ m /* l */ ( /* m */ ) /* n */ { /* o */ } /* p */ }/* after */;
```
Boshen, line 16: Oh no!

Dunqing: Don't worry! It has already been fixed in https://github.com/oxc-project/oxc/pull/4141

Boshen: Can we merge these 2 PRs, this is confusing :-/

overlookmotel:
> The class scope is not defined in the spec

I think it is - please see #4161.

tasks/coverage/parser_test262.snap
Dunqing: Done

Dunqing:
Blocked on https://github.com/oxc-project/oxc/issues/4200

Dunqing:
The regression tests have been fixed in #4156. Should I merge into this PR?

Dunqing (commented):
Not adding symbols to scope may cause some confusion. For example, in mangler, we need to rename bindings, but we can't find the symbols for function expression and class expression in scope.

 I'll revert this change

crates/oxc_semantic/src/builder.rs
```diff
@@ -1586,6 +1546,20 @@ impl<'a> SemanticBuilder<'a> {
...
+                if let Some(node) = self.nodes.parent_node(self.current_node_id) {
+                    if let AstKind::Class(class) = node.kind() {
+                        if class.is_expression() {
+                            // We must to bind class expression when enter BindingIdentifier,
+                            // because we should add binding to current scope
+                            let previous_node_id = self.current_node_id;
+                            self.current_node_id = node.id();
+                            class.bind(self);
+                            self.current_node_id = previous_node_id;
+                        }
```
Dunqing, line 1559: Bind `ClassExpression` in `BindingIdentifier` to add the binding to the correct scope

PR #4199, opened 2024-07-11
fix: avoid print extra semicolon after accessor property

IWANABETHATGUY:
## Before
```js
export default class Foo { @x @y accessor #aDef = 1 }
```
**output** 
```bash
Original:
export default class Foo { @x @y accessor #aDef = 1 }


Printed:
export default class Foo {
        accessor #aDef=1;;
}

```


IWANABETHATGUY:
Directly merged, since graphite did not support merge from from a forked repo

PR #4206, opened 2024-07-11
perf(semantic): reuse existing map of unresolved refs

lucab:


lucab:
I was hoping for this to have some visible performance gain, in particular at the root scope where it avoids draining and then recreating the map, but CodSpeed only shows a 1% improvement. I guess that isn't a very hot path.

Boshen:
Code is shorter, and every 1% counts :shipit:

overlookmotel:
@lucab I think the main gain here was not replacing `drain` with `retain`, but that you removed `.collect` into a new temporary `Vec` (a good change - that intermediate `Vec` was pointless).

PR #4203, opened 2024-07-11
refactor(ast/visit): fire node events as the outermost one.

rzvxa:
I'm going to be AFK today(till about 9 PM UTC). Meanwhile, I Didn't want to be a blocker so here we go.
It would fix the #4200 merge if you guys find it in the correct order otherwise feel free to close it.

PR #4208, opened 2024-07-11
refactor(semantic): convert scope fields to IndexVecs

### LINKED ISSUE #4269: Use `IndexVec` not `FxHashMap` for all fields of `ScopeTree`
Semantic's `ScopeTree` consists mostly of `IndexVec`s, but 2 fields are `FxHashMap`s.

https://github.com/oxc-project/oxc/blob/95e9b6986caad41c7991ae14cf1b943f326b46bc/crates/oxc_semantic/src/scope.rs#L21-L32

I'm not sure why `child_ids` and `node_ids` are hash maps, but imagine they could be changed to `IndexVec`s, which have faster lookup. If for some reason a few scopes don't get records in these fields at present, we could just insert dummies in those cases. Then `IndexId` can be used to index them.

lucab:
Closes: https://github.com/oxc-project/oxc/issues/4269

### DIFF
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -18,13 +18,12 @@ pub(crate) type UnresolvedReferences = FxHashMap<CompactStr, Vec<ReferenceId>>;
 /// `SoA` (Struct of Arrays) for memory efficiency.
 #[derive(Debug, Default)]
 pub struct ScopeTree {
-    /// Maps a scope to the parent scope it belongs in
+    /// Maps a scope to the parent scope it belongs in.
     parent_ids: IndexVec<ScopeId, Option<ScopeId>>,
-
-    /// Maps a scope to direct children scopes
-    child_ids: FxHashMap<ScopeId, Vec<ScopeId>>,
-    // Maps a scope to its node id
-    node_ids: FxHashMap<ScopeId, AstNodeId>,
+    /// Maps a scope to direct children scopes.
+    child_ids: IndexVec<ScopeId, Vec<ScopeId>>,
+    /// Maps a scope to its node id.
+    node_ids: IndexVec<ScopeId, AstNodeId>,
     flags: IndexVec<ScopeId, ScopeFlags>,
     bindings: IndexVec<ScopeId, Bindings>,
     pub(crate) root_unresolved_references: UnresolvedReferences,
@@ -48,10 +47,10 @@ impl ScopeTree {
         // have recursive closures
         fn add_to_list(
             parent_id: ScopeId,
-            child_ids: &FxHashMap<ScopeId, Vec<ScopeId>>,
+            child_ids: &IndexVec<ScopeId, Vec<ScopeId>>,
             items: &mut Vec<ScopeId>,
         ) {
-            if let Some(children) = child_ids.get(&parent_id) {
+            if let Some(children) = child_ids.get(parent_id) {
                 for child_id in children {
                     items.push(*child_id);
                     add_to_list(*child_id, child_ids, items);
@@ -67,11 +66,11 @@ impl ScopeTree {
     }
 
     pub fn get_child_ids(&self, scope_id: ScopeId) -> Option<&Vec<ScopeId>> {
-        self.child_ids.get(&scope_id)
+        self.child_ids.get(scope_id)
     }
 
     pub fn get_child_ids_mut(&mut self, scope_id: ScopeId) -> Option<&mut Vec<ScopeId>> {
-        self.child_ids.get_mut(&scope_id)
+        self.child_ids.get_mut(scope_id)
     }
 
     pub fn descendants_from_root(&self) -> impl Iterator<Item = ScopeId> + '_ {
@@ -132,7 +131,7 @@ impl ScopeTree {
     pub fn set_parent_id(&mut self, scope_id: ScopeId, parent_id: Option<ScopeId>) {
         self.parent_ids[scope_id] = parent_id;
         if let Some(parent_id) = parent_id {
-            self.child_ids.entry(parent_id).or_default().push(scope_id);
+            self.child_ids[parent_id].push(scope_id);
         }
     }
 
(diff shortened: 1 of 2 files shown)


lucab:
* **#4208** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/scope.rs
```diff
@@ -132,7 +132,7 @@ impl ScopeTree {
     pub fn set_parent_id(&mut self, scope_id: ScopeId, parent_id: Option<ScopeId>) {
         self.parent_ids[scope_id] = parent_id;
         if let Some(parent_id) = parent_id {
-            self.child_ids.entry(parent_id).or_default().push(scope_id);
+            self.child_ids[parent_id].push(scope_id);
```
DonIsaac, line 135: This could panic. `parent_id` should be inserted if it doesn't exist.

DonIsaac (dismissed):

crates/oxc_semantic/src/scope.rs
lucab: I think the logic in `add_scope()` already takes care of this.
In particular, if we have a `ScopeId` for the parent here, it means that as some point we got it through a `let parent_id = add_scope(...);`. Within that execution, there has been a `self.child_ids.push(...)` call for the `parent_id` scope, which in turn ensures that this `self.child_ids[parent_id]` here always sees a populated entry.
Does this make sense?

crates/oxc_semantic/src/scope.rs
```diff
@@ -184,16 +184,20 @@ impl ScopeTree {
...
+        _ = self.node_ids.push(AstNodeId::dummy());
 
-        if let Some(parent_id) = parent_id {
-            self.child_ids.entry(parent_id).or_default().push(scope_id);
-        }
+        let parent = match parent_id {
+            Some(id) => vec![id],
+            None => vec![],
+        };
+        _ = self.child_ids.push(parent);
```
overlookmotel, line 194: I don't think this is correct. Unless I've misunderstood, it's setting a scope's parent as it's own child. I think these 5 lines should be:

```rs
// Set this scope as child of parent scope
if let Some(parent_id) = parent_id {
    self.child_ids[parent_id].push(scope_id);
}
// Init list of child scopes for this scope (currently empty)
_ = self.child_ids.push(vec![]);
```

crates/oxc_semantic/src/scope.rs
```diff
@@ -184,16 +184,20 @@ impl ScopeTree {
         let scope_id = self.parent_ids.push(parent_id);
         _ = self.flags.push(flags);
         _ = self.bindings.push(Bindings::default());
+        // XXX: should this be a real invalid sentinel value or an Option?
+        _ = self.node_ids.push(AstNodeId::dummy());
```
overlookmotel, line 188: I think sentinel (as you have it now) is better. `Option<AstNodeId>` would be 16 bytes, rather than 8.

Once #4203 is merged, node IDs will be generated before scope is created, so we can add `node_id` to params of this function, and push the actual `AstNodeId` here. Then we can remove `set_node_id`.

overlookmotel (changes requested):

overlookmotel:
Thanks for continuing your rampage through all the semantic optimization issues!

There's a modest perf improvement, and as Boshen said on another issue, any gain is worthwhile. A bunch of 1%/2% gains add up pretty fast to something significant.

I think I see why it's not as much as I originally expected, though. I thought `node_id` was set for all nodes which have scopes, but I now see it's only set for functions (I don't know why! If knowing the `AstNodeId` of a scope is useful, we should record it for *all* scopes). But the result is that the `node_ids` `IndexVec` is much larger than the hash map it replaces. The greater size is offsetting the gain of faster accesses.

Dunqing:
> I think I see why it's not as much of a speed-up as I originally expected. I thought `node_id` was set for all nodes which have scopes, but I now see it's only set for functions (I don't know why! If knowing the `AstNodeId` of a scope is useful, we should record it for *all* scopes). But the result is that the `node_ids` `IndexVec` is much larger than the hash map it replaces. The greater size is offsetting the gain of faster accesses.


we should set the `node_id` for all nodes that have a scope, initially I only needed to use it in functions, so I only set for functions

crates/oxc_semantic/src/scope.rs
lucab: Ok, I've dropped the comment.

My concerns were that:
 * `dummy()` isn't a real **invalid** sentinel value. To the contrary, ID 0 is the most common valid ID. This setup seems prone to hard-to-debug logic bugs.
 * `AstNodeId` is not well-suited for an `Option<AstNodeId>` as it does not support niche-optimization (as you pointed out).

Maybe longer term this is going to be a moot point, but I guess it makes sense to turn `AstNodeId` into a `NonZeroUsize`.

overlookmotel: Yes, I agree with your concerns. It's a bit messy.

The background which I imagine you're not aware of because the info was buried in various issues is that I'm planning to do a "holistic" review of `Semantic` and make an RFC to significantly revamp it. This kind of issue is one I'll be looking at. That review will take me a month or so - first I need to familiarize myself with all the ways `Semantic` is used in both Oxc and Rolldown (there's a lot of Oxc I don't know at present - I've never touched the linter, for example).

But the perf improvements you've been working on were definitely going to be part of my conclusions, so getting them done now, in advance of my RFC, is really helpful as (a) it gives us perf boost *now*, and (b) it narrows the scope of what the RFC will have to cover.

overlookmotel (approved):

overlookmotel:
#4203 is merged now, so we can move recording `AstNodeId` into `add_scope` (see https://github.com/oxc-project/oxc/pull/4208#discussion_r1674797228).

Merge this PR now, and do that in a follow-up? Or amend this PR? Which do you prefer @lucab?

lucab:
@overlookmotel I'm fine with merging this small one as is. I started having a look at the new scope/node coupling, but I still need to explore some corners and I doubt I'll find more time this week.

overlookmotel:
@lucab OK great. Merging now. If you have time, please open an issue for moving recording `AstNodeId` into `add_scope` (https://github.com/oxc-project/oxc/pull/4208#discussion_r1674797228), so we don't forget about it.

lucab:
Tracked at https://github.com/oxc-project/oxc/issues/4234.

overlookmotel:
Thank you!

PR #4213, opened 2024-07-12
perf(semantic): recycle unresolved references hash maps

### LINKED ISSUE #4169: Optimize unresolved references in semantic
#4162 boosted perf of semantic by turning `unresolved_references` into a stack.

I think we can potentially do even better. From https://github.com/oxc-project/oxc/pull/4162#issuecomment-2220267268:

What I had in mind is that you never call `unresolved_references.pop()`. Instead, hold on to the hash map and re-use it. That way we avoid generating (and allocating) new hash map when entering a scope most of the time.

`unresolved_references` is `Vec<UnresolvedReferences>`, indexed by *scope depth*.

e.g. with this input:

```js
function foo() {}
function bar() {}
```

* Enter top level scope:
  * Depth is 0.
  * Push a `UnresolvedReferences` hash map to `unresolved_references`.
  * Current hashmap is `unresolved_references[0]`.
* Enter scope for `foo`:
  * Increment depth to 1.
  * Push another `UnresolvedReferences` hash map to `unresolved_references`.
  * Current hashmap is `unresolved_references[1]`.
* Exit scope for `foo`:
  * Clear current hashmap (`unresolved_references[1]`).
  * Decrement depth to 0.
  * Do NOT pop from `unresolved_references`.
  * Current hashmap is now `unresolved_references[0]`.
* Enter scope for `bar`:
  * Increment depth to 1.
  * Do NOT push to `unresolved_references`.
  * Current hashmap is `unresolved_references[1]` (i.e. reuse same hash map that was used for `foo`)
 * Exit scope for `bar`: same as for `foo`
 * Exit top level scope: same
 
I *think* this would be a further perf boost due to reducing allocations, and reusing allocations which are already warm in the cache.

cc @lucab (though no obligation on you to do this)

overlookmotel:
Closes #4169.

Reduce allocations while building `Semantic` by recycling hash maps used for tracking unresolved references, rather than creating a new one for every scope.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -56,6 +56,9 @@ pub struct SemanticBuilder<'a> {
     pub current_node_flags: NodeFlags,
     pub current_symbol_flags: SymbolFlags,
     pub current_scope_id: ScopeId,
+    /// Current scope depth.
+    /// 0 is global scope. 1 is `Program`. Incremented on entering a scope, and decremented on exit.
+    pub current_scope_depth: usize,
     /// Stores current `AstKind::Function` and `AstKind::ArrowFunctionExpression` during AST visit
     pub function_stack: Vec<AstNodeId>,
     // To make a namespace/module value like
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4213** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@lucab FYI this is what I was getting at when I suggested re-using hash maps. Sorry I explained it so badly!

Dunqing (approved):
That's great. I learned a lot.

PR #4217, opened 2024-07-12
fix(linter): not ignore adjacent spans when fixing

### LINKED ISSUE #4204: The adjacent `span` is ignored in fixer
```
test('extension is not outdated when there is no local and gallery', () => {
    const extension = instantiationService.createInstance(Extension, () => ExtensionState.Installed, () => undefined, undefined, undefined, undefined, undefined);
    assert.strictEqual(extension.outdated, false);
});
```

[microsoft/vscode@c37f4a4/src/vs/workbench/contrib/extensions/test/electron-sandbox/extension.test.ts#L29-L32](https://github.com/microsoft/vscode/blob/c37f4a49e28831cec1aec1af0f626edaada1f7ce/src/vs/workbench/contrib/extensions/test/electron-sandbox/extension.test.ts#L29-L32)

Our fixer needs to run twice to fix this, tested on unicorn/no-useless-undefined, it needs once.

----

Why our fixer need to run twice 

```rust
            if i64::from(start) <= last_pos {
                return;
            }
```

https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/fixer.rs#L278

I found that `span` is an interval that is closed on the left and open on the right. It should not be `<=`, which will cause the continuous interval to be skipped. It should be `<`.

The test case has to be modified to allow continuous interval modification, but I found that it will cause `no_useless_escape` test failure

https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/fixer.rs#L520
```rust
    const REMOVE_START: Fix = Fix::delete(Span::new(0, 4));
    const REPLACE_ID: Fix = Fix { span: Span::new(4, 10), content: Cow::Borrowed("foo") };
    fn apply_two_fix_when_the_start_the_same_as_the_previous_end() {
        let result = get_fix_result(vec![
            create_message(remove_start(), Some(REMOVE_START)),
            create_message(replace_id(), Some(REPLACE_ID)),
        ]);
        assert_eq!(result.fixed_code, TEST_CODE.replace("var answer", "foo"));
        assert!(result.fixed);
    }
```

_Originally posted by @cblh in https://github.com/oxc-project/oxc/issues/4079#issuecomment-2222794692_

mysteryven:
fixes: #4204

### DIFF
--- a/crates/oxc_linter/src/snapshots/no_useless_escape.snap
+++ b/crates/oxc_linter/src/snapshots/no_useless_escape.snap
@@ -181,12 +181,6 @@ source: crates/oxc_linter/src/tester.rs
    ·            ──
    ╰────
 
-  ⚠ eslint(no-useless-escape): Unnecessary escape character '{'
-   ╭─[no_useless_escape.tsx:1:14]
- 1 │ var foo = `\$\{{${foo}`;
-   ·              ──
-   ╰────
-
   ⚠ eslint(no-useless-escape): Unnecessary escape character '$'
    ╭─[no_useless_escape.tsx:1:12]
  1 │ var foo = `\$a${foo}`;
--- a/crates/oxc_linter/src/fixer.rs
+++ b/crates/oxc_linter/src/fixer.rs
@@ -316,7 +316,7 @@ impl<'a> Fixer<'a> {
             if start > end {
                 return;
             }
-            if i64::from(start) <= last_pos {
+            if i64::from(start) < last_pos {
                 return;
             }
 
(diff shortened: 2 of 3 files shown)


mysteryven:
* **#4217** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/snapshots/no_useless_escape.snap
```diff
@@ -181,12 +181,6 @@ source: crates/oxc_linter/src/tester.rs
    ·            ──
    ╰────
 
-  ⚠ eslint(no-useless-escape): Unnecessary escape character '{'
-   ╭─[no_useless_escape.tsx:1:14]
- 1 │ var foo = `\$\{{${foo}`;
```
mysteryven, line 186: `\$\{{${foo}` only need to report on:

```js
\$\{{${foo}
^
```

DonIsaac (approved):

PR #4228, opened 2024-07-12
perf(semantic): use `FxHashMap` instead of `IndexMap`

Dunqing:


overlookmotel:
I did wonder why we were using `IndexMap`. Is there any reason that decision was made?

Dunqing:
> I did wonder why we were using `IndexMap`, as iteration order isn't important. Is there any reason it was chosen?

I don't know why. I just found it slower than `FxHashMap`.

Dunqing:
We got 1%~3% performance improvement and removed 1 dependency!
image

Boshen:
https://github.com/oxc-project/oxc/commit/f8828db68c33988aefc5b96b19929fe60a170dc8

The mangler is sensitive to binding position.

overlookmotel:
> The mangler is sensitive to binding position.

Ah ha. So that's why `IndexMap`.

But... binding insertion order and position don't always align. They will if the AST comes direct from the parser. But what if a transform alters e.g. function params? Then you could end up with the 1st param being the last binding.

Boshen:
https://github.com/oxc-project/oxc/blob/be82c286d63f4660cde7df590a9586eec284232f/crates/oxc_mangler/src/lib.rs#L71

Symbols and scopes are recreated, since mangler is the last stage just before codegen of the pipeline, even after the compressor.

overlookmotel:
I assume (perhaps incorrectly?) that our intention is that ultimately transform and compress will update symbols and scopes as they modify the AST, so we won't need to keep recreating them. So at that point, the binding insertion order / position mismatch could become a problem.

But perhaps we should leave this as is for now.

Dunqing:
Close this because we're not sure yet that we can safely replace it with a `FxHashMap`.

PR #4235, opened 2024-07-12
feat(transformer): decode xml character entity `&#xhhhh` and `&#nnnn;`

Boshen:


Boshen:
I need to find the right place for some tests.

Boshen:
idempotency test should be enough.

PR #4237, opened 2024-07-12
feat(codegen): configurable indentation styles

DonIsaac:
Add an indentation variant to `CodegenOptions`, then use it in `print_indent`.

This PR also splits options into a separate `options` submodule because of how long option implementations were getting.


Boshen:
Our codegen is not a formatter. We deliberately try to avoid complexity and reduce `if` checks to make it performant.

The `quote` option was added so that we can match against different test suite.

PR #4239, opened 2024-07-12
perf(semantic): reduce cloning `CompactStr`s when declaring references

overlookmotel:
The hash map `Entry` API requires an owned key. Avoid using it so don't have to clone the reference name `CompactStr` unless  the hashmap need an insertion.

Using `Atom` instead of `CompactStr` as hash map key (#3972) would be a better long-term solution, but that's [not possible currently](https://github.com/oxc-project/oxc/pull/3974).

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -347,13 +347,17 @@ impl<'a> SemanticBuilder<'a> {
     ///
     /// # Panics
     pub fn declare_reference(&mut self, reference: Reference) -> ReferenceId {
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#4239** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Oh dear. That didn't work.

overlookmotel:
Od dear. That didn't work.

PR #4242, opened 2024-07-13
feat(coverage): enable regexp in test262

Boshen:
@leaysgur This enables all test262 regexp tests, feel free to decide when's the right time to integrate.

It seems like we need to add some pointing spans on the diagnostics.

It's somewhat slow to run `just c`, so I always use `just example parser` for local development.


leaysgur:
> feel free to decide when's the right time to integrate.

It looks like it will take a little longer, but understood. Thank you!

And I will catch up with the content of this PR as well.
(BTW, I will be on an early summer vacation next week and will be unavailable. 🏖️  )

crates/oxc_parser/src/js/expression.rs
```diff
@@ -337,20 +337,32 @@ impl<'a> ParserImpl<'a> {
...
-    pub(crate) fn parse_literal_regexp(&mut self) -> RegExpLiteral<'a> {
+    pub(crate) fn parse_literal_regexp(&mut self) -> Result<RegExpLiteral<'a>> {
+        use oxc_regexp_parser::{ParserOptions, PatternParser};
         let span = self.start_span();
 
         // split out pattern
         let (pattern_end, flags) = self.read_regex();
         let pattern_start = self.cur_token().start + 1; // +1 to exclude `/`
         let pattern = &self.source_text[pattern_start as usize..pattern_end as usize];
+        if let Err(diagnostic) = PatternParser::new(
```
IWANABETHATGUY, line 348: Reparsing each `regex` eagerly in place sounds not reasonable, can we introduce another visit pass (enabled by a option) and reparse each regex then emit **diagnostic** at the end?

IWANABETHATGUY: Most downstream users of `parser` may not care about whether the regex semantic is correct or not like formater, bundler.

leaysgur:
FYI:

- The previous CodSpeed result was one where the entire source code(!) was passed to the regexp parser, not the regular expression part
- I have not yet been started parser perf improvements

crates/oxc_parser/src/js/expression.rs
leaysgur: But actually, I'm also concerned about how to finish this PR.

With current code, `oxc_parser` parses `RegExp` and reports errors, but does not seem to hold the parsed results.

@Boshen What were your original plans?

leaysgur: I've thought about it a little, and to organize my thoughts..., I'll answer my own questions.

> RegExp literals(/abc/) are checked, but RegExp object calls(new RegExp("abc")) are not checked

It's OK.
In the case of `new RegExp("string")`, the code is just parsed, and if there’s an error in RegExp, it will occur at runtime.

On the other hand, in the case of `/string/`, the syntax must satisfy the requirements of a literal, so it should produce an error during the parsing stage, before runtime.

```js
console.log("START");
const a = new RegExp("a{", "u");
console.log("END");
```

This will log "START".

```js
console.log("START");
const a = /a{/u;
console.log("END");
```

This, however, will not log anything.

One thing that concerns me, though, is that if invalid literal is treated as error at the parser stage, it would make it impossible to implement rules like `eslint/no-invalid-regexp`? 

> How to reuse parsed result

Perhaps parse it again from Semantic, just like with JSDoc...?

Boshen:
Continue in https://github.com/oxc-project/oxc/pull/4998

PR #4250, opened 2024-07-14
perf(semantic): fast path for resolving references in scopes with no bindings

overlookmotel:
Scopes with no bindings are fairly common e.g.:

```js
if (foo) {
    // This block contains no bindings
    doStuff(foo);
}
```

Fast path for this case when resolving references in semantic.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -369,6 +369,8 @@ impl<'a> SemanticBuilder<'a> {
         let parent_refs = iter.nth(self.current_scope_depth - 1).unwrap();
         let current_refs = iter.next().unwrap();
 
+        // Dummy comment to run benchmarks again
+
         let bindings = self.scope.get_bindings(self.current_scope_id);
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4250** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4250?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This good for the `semantic[RadixUIAdoptionSection.jsx]` benchmark (+2%) and bad for `semantic[checker.ts]` and `semantic[pdf.mjs]` (-1%).

Reason is that `RadixUIAdoptionSection.jsx` contains a single function with a bunch of references in it, and no references in top-level scope, so it takes the fast path, and gets quite a bit of benefit from it. The other benchmarks will use the fast path far less.

But if the function had even a single parameter (which most React components do) then the fast path wouldn't kick in.

Therefore, on balance I think this change is not good - it's only benefitting an uncommon case.

PR #4252, opened 2024-07-14
Implement most of the Eslint plugin promise

jelly:
One rule [promise/param-names](https://github.com/standard/eslint-config-standard/blob/27470979b50a4fcdff8e1d01d82d1bcdf4da3cdd/src/index.ts#L273) is part of the [eslint-config-standard](https://www.npmjs.com/package/eslint-config-standard) so would be useful for oxlint to support. But only supporting one rule seems a bit unreasonable so I looked at at least implementing a few of the rules which got out of hand fast.... 

So this PR implements 10 out of the 15 rules. Some rules lack the explanation why this is bad because such documentation does not exists, there are multiple issues about it https://github.com/eslint-community/eslint-plugin-promise/issues/214 and https://github.com/eslint-community/eslint-plugin-promise/issues/109

Some of the leftover rules are a bit debatable if they are useful such as [no-native](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-native.md), disallowing native Promise implementations. And most folks seem to [disable the rule](https://github.com/search?q=%27promise%2Fno-native%27&type=code&p=1).

Leftover rules:

* always-return
* no-multiple-resolved
* no-nesting
* prefer-await-to-callbacks
* no-native

---

I am happy to split up this PR into multiple smaller PR's if this is too big to review.

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -403,3 +403,34 @@ pub fn is_function_node(node: &AstNode) -> bool {
         _ => false,
     }
 }
+
+pub fn is_promise(call_expr: &CallExpression) -> bool {
+    let Some(member_expr) = call_expr.callee.get_member_expr() else {
+        return false;
+    };
+
+    let Some(prop_name) = member_expr.static_property_name() else {
+        return false;
+    };
+
+    // hello.then(), hello.catch(), hello.finally()
+    if matches!(prop_name, "then" | "catch" | "finally") {
+        return true;
+    }
+
+    // foo().then(foo => {}), needed?
+    if let Expression::CallExpression(inner_call_expr) = member_expr.object() {
+        return is_promise(inner_call_expr);
+    }
+
+    if member_expr.object().is_specific_id("Promise")
+        && matches!(
+            prop_name,
+            "resolve" | "reject" | "all" | "allSettled" | "race" | "any" | "withResolvers"
+        )
+    {
+        return true;
+    }
+
+    false
+}
--- a/crates/oxc_linter/src/rules/promise/avoid_new.rs
+++ b/crates/oxc_linter/src/rules/promise/avoid_new.rs
@@ -1,0 +1,69 @@
+use oxc_ast::{ast::Expression, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn avoid_new_promise_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-promise(avoid-new): Avoid creating new promises")
+        .with_label(span0)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct AvoidNew;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Disallow creating new promises outside of utility libs.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// If you dislike the new promise style promises.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// new Promise((resolve, reject) => { ... });
+    /// ```
+    AvoidNew,
+    style,
+);
+
+impl Rule for AvoidNew {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::NewExpression(expr) = node.kind() else {
+            return;
+        };
+
+        let Expression::Identifier(ident) = &expr.callee else {
+            return;
+        };
+
+        if ident.name == "Promise" && ctx.semantic().is_reference_to_global_variable(ident) {
+            ctx.diagnostic(avoid_new_promise_diagnostic(expr.span));
+        }
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+
+    let pass = vec![
+        "Promise.resolve()",
+        "Promise.reject()",
+        "Promise.all()",
+        "new Horse()",
+        "new PromiseLikeThing()",
+        "new Promise.resolve()",
+    ];
+
+    let fail = vec![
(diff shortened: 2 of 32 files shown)


jelly:
Ok, this PR is clearly way too big and I just discovered the ability to make a util::promise module which seems to be good to refactor a bunch of rules. Moving this into a draft and I'll split up some commits.

crates/oxc_linter/src/rules/promise/avoid_new.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::NewExpression(expr) = node.kind() else {
+            return;
+        };
+
+        let Expression::Identifier(ident) = &expr.callee else {
+            return;
+        };
+
+        if ident.name == "Promise" && ctx.semantic().is_reference_to_global_variable(ident) {
```
jelly, line 43: I'm wondering what is common in oxlint here, there is:

```
        let AstKind::NewExpression(new_expr) = node.kind() else {
            return;
        };
        if !new_expr.callee.is_specific_id("Array") {
            return;
        }
```

Which would also run for:

```
class Array {};

const list = new Array(5).map(_ => {});
```

So that seems like it should be refactored?

jelly: And maybe we should get `is_global_id("Array", ctx)` which also checks if the global reference.

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,190 @@
...
+    let Some(prop_name) = member_expr.static_property_name() else {
+        return false;
+    };
+
+    // hello.then(), hello.catch(), hello.finally()
+    if matches!(prop_name, "then" | "catch" | "finally") {
+        return true;
+    }
+
+    // foo().then(foo => {}), needed?
```
jelly: I still have no idea what case this handles :-)

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,190 @@
...
+    if let Expression::CallExpression(inner_call_expr) = member_expr.object() {
+        return is_promise(inner_call_expr);
+    }
+
+    if member_expr.object().is_specific_id("Promise")
+        && matches!(
+            prop_name,
+            "resolve" | "reject" | "all" | "allSettled" | "race" | "any" | "withResolvers"
```
jelly: This should be moved to a utility, see previous rule which checks literally the same thing.

crates/oxc_linter/src/ast_util.rs
```diff
@@ -403,3 +403,34 @@ pub fn is_function_node(node: &AstNode) -> bool {
         _ => false,
     }
 }
+
+pub fn is_promise(call_expr: &CallExpression) -> bool {
```
jelly, line 407: Move to utils::promise instead.

crates/oxc_linter/src/rules/promise/param_names.rs
```diff
@@ -0,0 +1,179 @@
...
+                }
+                _ => continue,
+            }
+        }
+    }
+}
+
+impl ParamNames {
+    fn check_parameter_names(&self, params: &FormalParameters, ctx: &LintContext) {
+        if params.items.len() < 1 {
```
jelly, line 103: is_empty() instead?

jelly:
Two rules need to be branch aware: `always-return` `no-multiple-resolved`, which I am not sure if already implemented in oxint.

`no-native` doesn't feel very useful to implement, also mostly disabled from a github code search.

`no-nesting` `prefer-await-to-callbacks`, seem a bit complicated.

DonIsaac:
@jelly we do, there's a control flow graph in `Semantic`. I'm not super familiar with the code though, @rzvxa is more familiar with it.

DonIsaac:
@jelly I'll save my review for when you've split this PR up. Feel free to request one when you're ready.

rzvxa:
> Two rules need to be branch aware: `always-return` `no-multiple-resolved`, which I am not sure if already implemented in oxint.

Yes, we have a CFG. While it isn't documented, It's already used in some rules that you can study; Some of these rules are pretty similar to what you want for example `getter_return` and `require_render_return`(I suggest reading this one first since the former is in nursery and is a tad bit more complicated) should be really close to the `always_return` and can be used as a skeleton to build `no_multiple_resolved` on top of it.

To find all rules relying on CFG use the `find-reference` command of your LSP client on the `LintContext::cfg` method.

The graph itself uses [petgraph](https://docs.rs/petgraph/latest/petgraph/) and you can use its `depth-first`/`breadth-first`-search among other walk strategies, But for small subgraphs, I suggest our in-house search methods(which are more efficient on small subgraphs), You'd find them at `crates/oxc_cfg/src/visit.rs`.

Since we don't have a CFG visualization on our website(yet) you can create a `test.js` file in the root of the project - this is an ignored file - and run `cargo run -p oxc_semantic --example cfg` to generate semantic/CFG details as RON and DOT.

Then you can copy the content of the `test.dot` file into this [webpage](https://dreampuf.github.io/GraphvizOnline/)(or the DOT visualizer of your choice) to get a visualization of it.

Keep in mind the output of this dotfile is really dirty and is meant to be used for deep debugging purposes. To get a more simple dot output you should use `DisplayDot`(located at `crates/oxc_cfg/src/dot.rs`) and `DebugDot`(located at `crates/oxc_semantic/src/dot.rs`) traits. Those don't have an example that you can run on demand but it should be pretty trivial to set one up.

Since as of now I'm the only one who used our new CFG, I'm looking forward to your feedback and bug reports😅
Let me know if there is anything that I can help with, I'm always happy to chat. Just hit me up here or on Discord.

jelly:
Split off the easy part in https://github.com/oxc-project/oxc/pull/4293

mysteryven:
Seems there are only two rules not implemented(promise/no-promise-in-callback, promise/no-return-wrap), can we close this as a placeholder?

jelly:
Sure!

PR #3969, opened 2024-06-29
feat(semantic): add some scopes to align TypeScript's container

Dunqing:
This pr is a prototype to fix https://github.com/oxc-project/oxc/issues/2023. We have added scope to some nodes

PR #4253, opened 2024-07-14
feat(ast): align ts ast scope with typescript

### LINKED ISSUE #2023: Incorrect symbol references
https://oxc-project.github.io/oxc/playground/?code=3YCAAIC2gICAgICAgIC0G8rnONK88wEEqaAR7TB6kap9ZsyLue0RBjpWSJG6bXYDMwOdJ7L3juP1dt5lItxqFGL6b5l0%2F36F7IA%3D

image

The correct result is that the reference should be in `SymbolId 2`.

Dunqing:
close: #3969
close: #2023

We need to add scope according to [this](https://github.com/microsoft/TypeScript/blob/d8086f14b6b97c0df34a0cc2f56d4b5926a0c299/src/compiler/binder.ts#L3883-L3962). There are still some ASTs that need to be added to the scope.

---

Context from @Boshen:

Before this whole journey of fixing symbols and scopes I asked @Dunqing to debug through binder.ts via a debugger to fully understand how it does resolution.

We then agreed to align the implementation so that when a problem occurs, we can debug through both implementations and find where our problem is.

tsc doesn't have a specification, so we need to align with the reference implementation instead.


Dunqing:
* **#4253** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/coverage/parser_typescript.snap
Boshen, line 1: This is a big ask, but did you verify the changes here?

Dunqing: Yes. I verified. Since the scope was corrected. So the incorrect redeclare error is gone

PR #4262, opened 2024-07-15
perf(semantic): reduce overhead of cfg recording ast nodes

overlookmotel:
Control flow graph builder records AST node IDs in a temp structure `ast_nodes_records`. Only the first AST node ID inserted into the `Vec` is ever used, so turn it into just a single value.

Using `AstNodeId::dummy()` as sentinel for "not set" instead of using `Option<AstNodeId>` to reduce size of the records stack. `Option<AstNodeId>` is 16 bytes.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -92,7 +92,7 @@ pub struct SemanticBuilder<'a> {
 
     pub class_table_builder: ClassTableBuilder,
 
-    ast_nodes_records: Vec<Vec<AstNodeId>>,
+    ast_node_records: Vec<AstNodeId>,
 }
 
 pub struct SemanticBuilderReturn<'a> {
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4263** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4263?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4262** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa (approved):
Nice, I wanted to make this change for a long time, I'm glad you went ahead and refactored this.😍

crates/oxc_semantic/src/builder.rs
```diff
@@ -250,16 +250,19 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     fn record_ast_nodes(&mut self) {
```
rzvxa, line 252: this should be `record_ast_node` now. Maybe we can rename the other method to something like `push_ast_record`?

rzvxa: @Boshen Feel free to resolve my nitpicks. None of my comments are a blocking change, Just suggestions.

PR #4263, opened 2024-07-15
perf(semantic): do not record ast nodes for cfg if cfg disabled

overlookmotel:
Control flow graph builder records AST node IDs in a temp structure `ast_node_records`. Disable this if CFG is disabled.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -250,18 +250,26 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     fn record_ast_nodes(&mut self) {
-        self.ast_node_records.push(AstNodeId::dummy());
+        if self.cfg.is_some() {
+            self.ast_node_records.push(AstNodeId::dummy());
+        }
     }
 
     #[allow(clippy::unnecessary_wraps)]
     fn retrieve_recorded_ast_node(&mut self) -> Option<AstNodeId> {
-        Some(self.ast_node_records.pop().expect("there is no ast node record to stop."))
+        if self.cfg.is_some() {
+            Some(self.ast_node_records.pop().expect("there is no ast node record to stop."))
+        } else {
+            None
+        }
     }
 
     fn record_ast_node(&mut self) {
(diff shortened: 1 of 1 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4263?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4263** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4263?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4262** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rzvxa (commented):
Wouldn't it be better if we do this at the call site? We might want to use these for other stuff. We can wrap both the `record_ast_node` and `retrive_recorded_ast_node` in `control_flow!` macros.
And since these 2 wouldn't get called without CFG we can remove the guards around `record_ast_node`.

overlookmotel:
> Wouldn't it be better if we do this at the call site? We might want to use these for other stuff. We can wrap both the `record_ast_node` and `retrive_recorded_ast_node` in `control_flow!` macros.

Yes, we could do that. But I didn't see the point of adding `control_flow!` at the call sites, because ultimately the macro does the same thing as the check I've put in `record_ast_nodes` etc.

Although... the macro would move the `if self.cfg.is_some()` check inline within `visit_*` functions. So it might make it faster when CFG is disabled (e.g. in transformer), as it'd remove a function call (if compiler doesn't inline these functions anyway - both a quite trivial).

We could alternatively try to get same optimization by marking these functions `#[inline]`. I'll try that first.

> And since these 2 wouldn't get called without CFG we can remove the guards around `record_ast_node`.

Yes, you're right. That guard could be removed.

PR #4265, opened 2024-07-15
refactor(semantic): directly record `current_node_id` when adding a scope

### LINKED ISSUE #4234: semantic: directly record real node ID when adding a scope to the scope-tree
From https://github.com/oxc-project/oxc/pull/4208#discussion_r1674797228:

> Once https://github.com/oxc-project/oxc/pull/4203 is merged, node IDs will be generated before scope is created, so we can add node_id to params of this function, and push the actual AstNodeId here. Then we can remove set_node_id.

In particular, this is about refactoring the logic here below and getting rid of the dummy node ID:
https://github.com/oxc-project/oxc/blob/be82c286d63f4660cde7df590a9586eec284232f/crates/oxc_semantic/src/scope.rs#L182-L195

Dunqing:
close: #4234

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -189,7 +189,7 @@ impl<'a> SemanticBuilder<'a> {
     /// # Panics
     pub fn build(mut self, program: &Program<'a>) -> SemanticBuilderReturn<'a> {
         if self.source_type.is_typescript_definition() {
-            let scope_id = self.scope.add_scope(None, ScopeFlags::Top);
+            let scope_id = self.scope.add_scope(None, AstNodeId::dummy(), ScopeFlags::Top);
             program.scope_id.set(Some(scope_id));
         } else {
             self.visit_program(program);
@@ -449,7 +449,7 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
             flags = self.scope.get_new_scope_flags(flags, parent_scope_id);
         }
 
-        self.current_scope_id = self.scope.add_scope(parent_scope_id, flags);
+        self.current_scope_id = self.scope.add_scope(parent_scope_id, self.current_node_id, flags);
         scope_id.set(Some(self.current_scope_id));
 
         if let Some(parent_scope_id) = parent_scope_id {
@@ -471,8 +471,6 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         if !flags.is_top() {
             self.bind_function_or_class_expression();
         }
-
-        self.add_current_node_id_to_current_scope();
     }
 
     fn leave_scope(&mut self) {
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#4265** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -189,7 +189,7 @@ impl<'a> SemanticBuilder<'a> {
     /// # Panics
     pub fn build(mut self, program: &Program<'a>) -> SemanticBuilderReturn<'a> {
         if self.source_type.is_typescript_definition() {
-            let scope_id = self.scope.add_scope(None, ScopeFlags::Top);
+            let scope_id = self.scope.add_scope(None, AstNodeId::dummy(), ScopeFlags::Top);
```
lucab, line 192: Why not `self.current_node_id` here?

crates/oxc_traverse/src/context/scoping.rs
```diff
@@ -121,7 +121,7 @@ impl TraverseScoping {
     /// `flags` provided are amended to inherit from parent scope's flags.
     pub fn create_scope_child_of_current(&mut self, flags: ScopeFlags) -> ScopeId {
         let flags = self.scopes.get_new_scope_flags(flags, self.current_scope_id);
-        self.scopes.add_scope(Some(self.current_scope_id), flags)
+        self.scopes.add_scope(Some(self.current_scope_id), AstNodeId::dummy(), flags)
```
lucab, line 124: I think this will result in multiple scopes having the same node ID (0), is this a bug in the logic?

overlookmotel: I can answer this one. `Traverse` is a mutating visitor. For this reason, it does not use all of `Semantic`. It can't use `AstNodes` because that contains `AstKind`s which contain references to all actual AST nodes. Holding that would be a violation of Rust's aliasing rules, because visitor functions also get passed `&mut` references to AST nodes => would result in having a live `&` and `&mut` reference to same AST node simultaneously.

So... long story short... `Traverse` uses only an incomplete subset of `Semantic` and that doesn't include `AstNodes`, so it uses dummy `AstNodeId`s.

This is part of what I hope to look at when I review `Semantic`. We need some way to merge the 2 paradigms.

crates/oxc_semantic/src/builder.rs
overlookmotel: Good question. I guess we could do that.

overlookmotel: Please feel free to submit a PR.

Dunqing: We don't build semantic for `.d.ts` file now. So we haven't  current_node_id. We may need to build semantic for `d.ts` in the future. Rolldown needs this to bundle types to one entry point.

PR #4266, opened 2024-07-15
perf(semantic): reduce index_vec allocation

Dunqing:
Combining some SymbolTable fields to the `Symbol` struct, we can reduce allocating `IndexVec`. After my changes, I suddenly realized the api is much easier to use now. We no longer need the `symbol_id` to get the various field value you want, you just need the `Symbol`

### DIFF
--- a/crates/oxc_semantic/src/symbol.rs
+++ b/crates/oxc_semantic/src/symbol.rs
@@ -77,7 +77,11 @@ impl SymbolTable {
         self.len() == 0
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = SymbolId> + '_ {
+    pub fn iter(&self) -> impl Iterator<Item = &Symbol> + '_ {
+        self.symbols.iter()
+    }
+
+    pub fn symbol_ids(&self) -> impl Iterator<Item = SymbolId> + '_ {
         self.symbols.iter_enumerated().map(|(symbol_id, _)| symbol_id)
     }
 
@@ -87,6 +91,10 @@ impl SymbolTable {
 
     pub fn names(&self) -> impl Iterator<Item = &CompactStr> {
         self.symbols.iter().map(|symbol| &symbol.name)
+    }
+
+    pub fn resolved_references(&self) -> impl Iterator<Item = &Vec<ReferenceId>> {
+        self.symbols.iter().map(|symbol| &symbol.resolved_references)
     }
 
     pub fn name(&self, symbol_id: SymbolId) -> &CompactStr {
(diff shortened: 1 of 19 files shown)


Dunqing:
* **#4266** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4266?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Interesting that this change from struct-of-arrays to array-of-structs is a gain on benchmarks. However, there is another possible approach to tackling this which may be even better: https://github.com/oxc-project/backlog/issues/11

Dunqing:
> Interesting that this change from struct-of-arrays to array-of-structs is a gain on benchmarks. However, there is another possible approach to tackling this which may be even better: [oxc-project/backlog#11](https://github.com/oxc-project/backlog/issues/11)
> 
> However, since this PR is a gain at present, we could merge it now and attempt [oxc-project/backlog#11](https://github.com/oxc-project/backlog/issues/11) later on?

This PR is just an attempt at my guess, I'm not sure we want to merge it. I know you will be optimizing the whole `semantic`, and you probably have further improvements that can be done better. So it's up to you whether to merge it or not!

Boshen:
SoA improves write performance but regresses read performance (notice drop in linter).

I think we should keep SoA unless there is a win across the table.

Dunqing:
> SoA improves write performance but regresses read performance (notice drop in linter).

There are further optimizations we can make in linter, Such as we can pass `Symbol` to `run_on_symbol`. This may give a small performance boost

crates/oxc_semantic/src/symbol.rs
```diff
@@ -23,82 +23,130 @@ const TS_APPEND_CONTENT: &'static str = r#"
...
+#[derive(Debug)]
+#[cfg_attr(feature = "serialize", derive(Serialize, Tsify), serde(rename_all = "camelCase"))]
+pub struct Symbol {
+    pub span: Span,
+    pub name: CompactStr,
+    pub flag: SymbolFlags,
+    pub scope_id: ScopeId,
+    /// Pointer to the AST Node where this symbol is declared
+    pub declaration: AstNodeId,
+    pub resolved_references: Vec<ReferenceId>,
```
DonIsaac, line 35: Should we encapsulate these fields for future flexibility?

crates/oxc_semantic/src/symbol.rs
```diff
@@ -23,82 +23,130 @@ const TS_APPEND_CONTENT: &'static str = r#"
 export type IndexVec<I, T> = Array<T>;
 "#;
 
+#[derive(Debug)]
+#[cfg_attr(feature = "serialize", derive(Serialize, Tsify), serde(rename_all = "camelCase"))]
+pub struct Symbol {
```
DonIsaac, line 28: Consider adding a `GetSpan` impl.

Dunqing:
Closing this because we tried another better way to improve more in #4328

PR #4272, opened 2024-07-15
perf(semantic): inline ast record functions

overlookmotel:
Inline these functions so that when CFG is disabled, it doesn't cost a function call just for the trivial `if self.cfg.is_some()` check.

Based on @rzvxa's suggestion in https://github.com/oxc-project/oxc/pull/4263#pullrequestreview-2176762670.

PR #4273, opened 2024-07-15
perf(semantic): remove a branch from `record_ast_node`

overlookmotel:
Remove an unnecessary branch from `record_ast_node`. As suggested by @rzvxa in https://github.com/oxc-project/oxc/pull/4263#pullrequestreview-2176762670.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -275,6 +275,7 @@ impl<'a> SemanticBuilder<'a> {
                 *record = self.current_node_id;
             }
         }
+        // Dummy comment to re-run benchmark. TODO Remove me.
     }
 
     pub fn current_scope_flags(&self) -> ScopeFlags {
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -275,7 +275,6 @@ impl<'a> SemanticBuilder<'a> {
                 *record = self.current_node_id;
             }
         }
-        // Dummy comment to re-run benchmark. TODO Remove me.
     }
 
     pub fn current_scope_flags(&self) -> ScopeFlags {
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -268,11 +268,11 @@ impl<'a> SemanticBuilder<'a> {
 
     #[inline]
     fn record_ast_node(&mut self) {
-        if self.cfg.is_some() {
-            if let Some(record) = self.ast_node_records.last_mut() {
-                if *record == AstNodeId::dummy() {
-                    *record = self.current_node_id;
-                }
+        // NB: If CFG is disabled, `ast_node_records` is always empty, so this is skipped.
+        // No need for a `if self.cfg.is_some()` guard.
+        if let Some(record) = self.ast_node_records.last_mut() {
+            if *record == AstNodeId::dummy() {
+                *record = self.current_node_id;
             }
         }
     }


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4273?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4273** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4273?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4272** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4272?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rzvxa:
Huh, It is consistently slower? How does that make sense?

overlookmotel:
@rzvxa I've run the benchmarks 3 times, and each time it shows a negative effect. NB: CFG is not enabled on semantic benchmark.

I am guessing what's going on is one of:

* Because `record_ast_nodes` and `retrieve_recorded_ast_node` also call `self.cfg.is_some()`, this value is warm in cache so faster to check.
* Because `record_ast_nodes` and `retrieve_recorded_ast_node` also call `self.cfg.is_some()`, using `self.cfg.is_some()` here too allows compiler to see it's a repeated check and make some optimization.
* `Vec::last_mut` is slower than `Option::is_some`. This surprises me as I'd assume first thing `Vec::last_mut` would do is a bounds check.

But this is just guesswork. No idea really! Any thoughts?

Regardless, given the (small) perf regression, I think we should not merge this.

rzvxa:
Yes, Let's close it. I initially suggested this to have fewer checks in the CFG pass but since the linter benchmark(which contains CFG) is untouched it clearly wasn't the case.

Sorry to waste your time mate😓

overlookmotel:
Not a waste of time at all. Negative results are also results.

overlookmotel:
I thought about it more, and I have another theory. Maybe the thing is that `Vec::last_mut` is a more complicated function, and often used, so compiler doesn't inline it. So the difference between `Vec::last_mut` and `Option::is_some` is that `Vec::last_mut` involves a function call. Maybe.

@rzvxa Still no idea if this is correct, but thought you might be interested.

rzvxa:
Thanks for having this conversation with me.

> I thought about it more, and I have another theory. Maybe the thing is that `Vec::last_mut` is a more complicated function, and often used, so compiler doesn't inline it. So the difference between `Vec::last_mut` and `Option::is_some` is that `Vec::last_mut` adds the overhead a function call. Added to that, it may be optimized to expect the `Vec` to have entries, so returning `None` is the cold path and incurs a branch mis-prediction.
> 
> Maybe.
> 
> @rzvxa Still no idea if this is correct, but thought you might be interested.

I've tested this with a few other approaches such as checking indices instead of using the last_mut and checks for `is_some` have outperformed them all. Rustc is just really good at handling options(as it should be).

overlookmotel:
> Thanks for having this conversation with me.
> 
> > I thought about it more, and I have another theory. Maybe the thing is that `Vec::last_mut` is a more complicated function, and often used, so compiler doesn't inline it. So the difference between `Vec::last_mut` and `Option::is_some` is that `Vec::last_mut` adds the overhead a function call. Added to that, it may be optimized to expect the `Vec` to have entries, so returning `None` is the cold path and incurs a branch mis-prediction.
> > Maybe.
> > @rzvxa Still no idea if this is correct, but thought you might be interested.
> 
> I've tested this with a few other approaches such as checking indices instead of using the last_mut and checks for `is_some` have outperformed them all. Rustc is just really good at handling options(as it should be).

Or maybe the explanation was this one:

* Because `record_ast_nodes` and `retrieve_recorded_ast_node` also call `self.cfg.is_some()`, this value is warm in cache, so faster to check.

PR #4275, opened 2024-07-15
refactor(semantic): alter syntax of `control_flow!` macro

overlookmotel:
Previously:

```rs
let ix = control_flow!(|self, cfg| cfg.current_node_ix);
```

after this PR:

```rs
let ix = control_flow!(self, |cfg| cfg.current_node_ix);
```

It expands to:

```rs
let ix = if let Some(ref mut cfg) = self.cfg {
    cfg.current_node_ix
} else {
    Default::default()
};
```

So rationale for this change is that it makes it clearer that `self` is passed *in* and `cfg` comes *out* into the "closure".

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -32,7 +32,7 @@ use crate::{
 };
 
 macro_rules! control_flow {
-    (|$self:ident, $cfg:tt| $body:expr) => {
+    ($self:ident, |$cfg:tt| $body:expr) => {
         if let Some(ref mut $cfg) = $self.cfg {
             $body
         } else {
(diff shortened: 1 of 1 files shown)


overlookmotel:
@rzvxa Out of interest, why `if let Some(ref mut cfg) = self.cfg` rather than `if let Some(cfg) = &mut self.cfg`? Is there a subtle difference between the two?

rzvxa:
> @rzvxa Out of interest, why `if let Some(ref mut cfg) = self.cfg` rather than `if let Some(cfg) = &mut self.cfg`? Is there a subtle difference between the two?

Not a functional one, I find this too noisy(lots of symbols).

```rust
if let Some($cfg) = &mut $self.cfg {
    $body
} else {
}
```

rzvxa (approved):

overlookmotel:
Thanks for satisfying my curiosity!

PR #4278, opened 2024-07-15
refactor(semantic): move function/class-specific code into specific visitors

overlookmotel:
Instead of calling `bind_function_or_class_expression` for every scope, and then branching on the AST node kind, insert the relevant code into the visitors for functions and classes. This reduces work on all other kinds of scopes e.g. block statements.

PR #4283, opened 2024-07-15
refactor(ast): pass final `ScopeFlags` into `visit_function`

overlookmotel:
We have a strange workaround for `visit_function` where we pass in `ScopeFlags`, to support creating the scope inside `Function`, but setting different flags for `MethodDefinition`s.

Previously `visit_function` took `Option<ScopeFlags>` and then did `flags.unwrap_or(ScopeFlags::empty()) | ScopeFlags::Function` to it. Personally, I found this confusing. When I was looking at `MethodDefinition`, I was wondering "It's a function, why doesn't it set Function flag too?"

This changes makes it more explicit and clear what `ScopeFlags` everything has.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -79,7 +79,7 @@ pub enum Expression<'a> {
     ChainExpression(Box<'a, ChainExpression<'a>>) = 16,
     ClassExpression(Box<'a, Class<'a>>) = 17,
     ConditionalExpression(Box<'a, ConditionalExpression<'a>>) = 18,
-    #[visit_args(flags = None)]
+    #[visit_args(flags = ScopeFlags::Function)]
     FunctionExpression(Box<'a, Function<'a>>) = 19,
     ImportExpression(Box<'a, ImportExpression<'a>>) = 20,
     LogicalExpression(Box<'a, LogicalExpression<'a>>) = 21,
@@ -970,7 +970,7 @@ pub struct BlockStatement<'a> {
 #[cfg_attr(feature = "serialize", serde(untagged))]
 pub enum Declaration<'a> {
     VariableDeclaration(Box<'a, VariableDeclaration<'a>>) = 32,
-    #[visit_args(flags = None)]
+    #[visit_args(flags = ScopeFlags::Function)]
     FunctionDeclaration(Box<'a, Function<'a>>) = 33,
     ClassDeclaration(Box<'a, Class<'a>>) = 34,
     UsingDeclaration(Box<'a, UsingDeclaration<'a>>) = 35,
@@ -1437,8 +1437,8 @@ pub struct BindingRestElement<'a> {
 /// Function Definitions
 #[visited_node]
 #[scope(
-    // TODO: `ScopeFlags::Function` is not correct if this is a `MethodDefinition`
-    flags(flags.unwrap_or(ScopeFlags::empty()) | ScopeFlags::Function),
+    // `flags` passed in to visitor via parameter defined by `#[visit_args(flags = ...)]` on parents
+    flags(flags),
     strict_if(self.is_strict()),
 )]
 #[derive(Debug)]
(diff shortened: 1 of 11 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4283?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4283** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4283?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4278** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rzvxa (changes requested):
What do you think about this?

```rust
quote! {
    if #strict_if {
        #flags | ScopeFlags::StrictMode
    } else {
        #flags
    }
}
```

Instead of this?

https://github.com/oxc-project/oxc/blob/91ac3f242d51c0b9ebe67cfe8512dc1049be7e65/tasks/ast_codegen/src/generators/visit.rs#L495-L501

We no longer have to do `unwrap` stuff here so it generates a more concise code.

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1437,8 +1437,8 @@ pub struct BindingRestElement<'a> {
 /// Function Definitions
 #[visited_node]
 #[scope(
-    // TODO: `ScopeFlags::Function` is not correct if this is a `MethodDefinition`
-    flags(flags.unwrap_or(ScopeFlags::empty()) | ScopeFlags::Function),
+    // `flags` passed in to visitor via parameter defined by `#[visit_args(flags = ...)]` on parents
```
rzvxa, line 1440: To describe this I wanted to add something like this to the `visited_node`.

```rust
#[visited_node(args(flags: ScopeFlags))]
struct X;
```

It does absolutely nothing but express that this visited node has an extra argument called `flags` which is `ScopeFlags`. What do you think about it? I'm not sure if having this over a comment is beneficial so I'm looking forward to hearing from you.

Since with this marker, we still have to implement flags as an edge case(as we are doing right now) it might not be worth it and just add to the code noise.

I wasn't sure then and I'm not sure now.

overlookmotel: I would really like to find a way to remove this business with the extra arg entirely. It's the only place we do it.

If we had separate `FunctionDeclaration`, `FunctionExpression` and `ObjectMethod` types (along with class `MethodDefinition`), then each of them could hold the scope, and there'd be no need for this weird workaround. In my view, that'd also have other advantages.

So personally I think don't bother finessing this too much - let's just get rid of it entirely!

overlookmotel:
> What do you think about this?
> 
> ```rust
> quote! {
>     if #strict_if {
>         #flags | ScopeFlags::StrictMode
>     } else {
>         #flags
>     }
> }
> ```

We're into micro-optimization territory here but... What we ideally want is for compiler to do this in straight line code with a `cmov` instruction (instead of a branch). [Godbolt shows](https://godbolt.org/z/oxeh67TjY) that both versions produce this when `#flags` is simple, but I wonder if `#flags` was more complicated (e.g. a function call), that'd encourage compiler towards a branch.

Quite possibly it makes no difference at all either way, but `if #strict_if { #flags =| ScopeFlags::StrictMode; }` is closer to what we want the compiler to produce, so on balance I think it's a safer bet.

rzvxa (approved):

crates/oxc_ast/src/ast/js.rs
overlookmotel: https://github.com/oxc-project/backlog/issues/28 is relevant.

PR #4287, opened 2024-07-15
ci: run tests and conformance on all PRs in a stack

overlookmotel:
Run tests (on Ubuntu only) and conformance on all PRs in a stack.

Before this change, they run only on the bottom PR in stack.

When I'm working on my desktop machine, it is quite slow, so I prefer to push and let CI run tests + conformance rather than waiting for them to run locally.

@Boshen Are you OK with this change? I'm sure you had a reason for disabling them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,18 +40,40 @@ jobs:
           graphite_token: ${{ secrets.GRAPHITE_TOKEN }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
-  test:
+  test-ubuntu:
+    name: Test (ubuntu-latest)
+    runs-on: ubuntu-latest
+    env:
+      # https://github.com/mozilla/sccache/blob/main/docs/S3.md
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4287** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4287?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
What if instead of doing this we add a commit message flag such as `[no optimize]` and use a prep step to check it? CI Optimize saves a lot of time when it works, But I agree. I also would love to be able to run these on demand. We can also add a toggle to the manual CI run so we can fire one manually [here](https://github.com/oxc-project/oxc/actions/workflows/ci.yml).

rzvxa:
I also found this, Maybe we can configure it to also run on the topmost PR
image
image

Boshen:
> When I'm working on my desktop machine, it is quite slow

Sounds like a problem, what is slow exactly?

---

I had a commit somewhere removing `optimize_ci`, but forgot to push, I'll get rid of it soon, it's a confusing feature.

overlookmotel:
> I had a commit somewhere removing optimize_ci, but forgot to push, I'll get rid of it soon, it's a confusing feature.

Can we merge this PR in meantime?

I actually think it's good to skip running tests on MacOS + Windows until just before merge. These rarely fail if tests on Ubuntu pass, so it's 2 less CI tasks to run.

Boshen:
I removed `optimized_ci` and the windows job on main.

overlookmotel:
Thanks Boshen.

PR #4289, opened 2024-07-16
refactor(ast): rename the `visited_node` marker to `ast`.

### LINKED ISSUE #4282: Rename `#[visited_node]` attr to `#[ast]`
At present we only have `#[visited_node]` attrs on types which are visited.

We will also need to add an attribute to all types which are part of AST but are *not* visited, for:

* #4284
* https://github.com/oxc-project/oxc/issues/4296
* AST transfer

Currently:

```rs
#[visited_node]
enum Expression<'a> { /* ... */ }
```

I propose:

```rs
#[ast(visit)]
enum Expression<'a> { /* ... */ }
```

or:

```rs
#[ast]
#[visit]
enum Expression<'a> { /* ... */ }
```

Types which are part of AST, but which are *not* visited would omit the `visit` part e.g.:

```rs
#[ast]
pub enum FunctionType { /* ... */ }
```

### Bikeshedding

We could call the attr something other than `#[ast]`. I considered `#[ast_node]` but:

1. That name will be familiar to some from SWC, and I don't think we should use it because these attrs do something very different - they're no-op markers for codegen, rather than macros generating a huge amount of trait impls and other code.
2. Some of these types (e.g. `FunctionType`) are not AST *nodes* per se.

The common factor between all these types is that they're part of the AST, so `#[ast]` seems descriptive.

But I'm not at all set on that name, if someone has a better suggestion.

cc @Boshen @rzvxa.

rzvxa:
closes #4282
I went with `#[ast(visit)]` but we can change it to 2 separate markers as suggested by @overlookmotel in the issue.

### DIFF
--- a/crates/oxc_ast/src/ast_impl/js.rs
+++ b/crates/oxc_ast/src/ast_impl/js.rs
@@ -1,6 +1,3 @@
-// NB: `#[visited_node]` attribute on AST nodes does not do anything to the code in this file.
-// It is purely a marker for codegen used in `oxc_traverse`. See docs in that crate.
-
 use crate::ast::*;
 
 use std::{cell::Cell, fmt, hash::Hash};
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -8,7 +8,7 @@
 use std::cell::Cell;
 
 use oxc_allocator::{Box, Vec};
-use oxc_ast_macros::visited_node;
+use oxc_ast_macros::ast;
 use oxc_span::{Atom, SourceType, Span};
 use oxc_syntax::{
     operator::{
@@ -27,7 +27,7 @@ use serde::Serialize;
 #[cfg(feature = "serialize")]
 use tsify::Tsify;
 
-#[visited_node]
+#[ast(visit)]
 #[scope(
     flags(ScopeFlags::Top),
     strict_if(self.source_type.is_strict() || self.directives.iter().any(Directive::is_use_strict)),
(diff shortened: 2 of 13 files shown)


rzvxa:
* **#4289** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_ast/src/ast_impl/js.rs
```diff
@@ -1,6 +1,3 @@
-// NB: `#[visited_node]` attribute on AST nodes does not do anything to the code in this file.
```
overlookmotel, line 1: Can we keep these comments, but just change `#[visited_node]` to `#[ast]`? The intent is for `#[ast]` to become a real macro in future, but it isn't right now - and anyway when it does we'll need to have a comment here explaining what it does.

rzvxa: These are removed from the implementation files where we don't have any occurrence of the `visited_node`(and now `ast`). I can move them to a separate PR if you'd prefer. That's just a cleanup I've missed when separating ast files into x.rs and x_impl.rs

overlookmotel: Oh I'm sorry, my mistake. I wasn't paying attention to which files these comments were in. No need to put in a separate PR. Just ignore me!

overlookmotel:
@rzvxa Is this stack ready for review? (or at least bottom 3 PRs of it?)

rzvxa:
> @rzvxa Is this stack ready for review? (or at least bottom 3 PRs of it?) Or you still working on it?

Yes, the bottom 3 are ready to be reviewed.

overlookmotel (approved):

PR #4293, opened 2024-07-16
feat(linter): Add eslint-plugin-promise rules: avoid-new, no-new-statics, params-names

jelly:
This introduces the `eslint-plugin-promise` plugin and implements three relatively simple rules.

Split off from https://github.com/oxc-project/oxc/pull/4252

### DIFF
--- a/crates/oxc_linter/src/rules/promise/avoid_new.rs
+++ b/crates/oxc_linter/src/rules/promise/avoid_new.rs
@@ -1,0 +1,69 @@
+use oxc_ast::{ast::Expression, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn avoid_new_promise_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("eslint-plugin-promise(avoid-new): Avoid creating new promises")
+        .with_label(span0)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct AvoidNew;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Disallow creating new promises outside of utility libs.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// If you dislike the new promise style promises.
+    ///
+    /// ### Example
+    /// ```javascript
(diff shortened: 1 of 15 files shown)


crates/oxc_linter/src/rules/promise/avoid_new.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// ### Why is this bad?
+    ///
+    /// If you dislike the new promise style promises.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// new Promise((resolve, reject) => { ... });
+    /// ```
+    AvoidNew,
+    style,
```
mysteryven, line 30: As it disallows the usage of `new promise()`, we can change it to `restriction`.

crates/oxc_linter/src/rules/promise/no_new_statics.rs
```diff
@@ -0,0 +1,106 @@
+use oxc_ast::{ast::Expression, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn static_promise_diagnostic(span0: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn(
+        "eslint-plugin-promise(no-new-statics): Disallow calling `new` on a Promise static method",
```
mysteryven, line 10: Nit: Seems the error message about this rule is `Avoid calling 'new' on 'Promise.{{ name }}()`, let's keep the same with it.

https://github.com/eslint-community/eslint-plugin-promise/blob/ecbce9faef0c8e05636287cf9c851949be6d70cc/rules/no-new-statics.js#L26

crates/oxc_linter/src/snapshots/no_new_statics.snap
```diff
@@ -0,0 +1,46 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint-plugin-promise(no-new-statics): Disallow calling `new` on a Promise static method
+   ╭─[no_new_statics.tsx:1:1]
+ 1 │ new Promise.resolve()
```
mysteryven, line 6: Nit: do we need to shrink to point on `new`?

jelly: Why not, it seems to make sense and was easy to do :)

Issue #3291, opened 2024-05-15
Convert lexer to iterate over bytes instead of chars


overlookmotel:
### Why the lexer is slower than it could be

`Chars` iterator is really slow. Lexer should iterate byte-by-byte rather than char-by-char.

In almost all cases, we're only matching against ASCII characters anyway (e.g. `self.next_eq('.')`), so calculating the Unicode char is completely pointless, as we only care if it's an ASCII `.` anyway. Surprisingly, the compiler seems not to be able to see this, and doesn't optimize out the char calculation itself. It generates a *lot* of pointless and slow code.

### How to make it faster

[`Source`](https://github.com/oxc-project/oxc/blob/main/crates/oxc_parser/src/lexer/source.rs) was introduced with intention to ease the transition to byte-by-byte iteration through source code.

#2352 got some heavy perf improvements from using it (along with other tricks), but I have not been able to find the time to complete the work.

The following APIs should be converted to take/return `u8` and use `Source::peek_byte` instead of `Source::peek_char`:

 - [x] `peek`
 - [x] `peek2`
 - [x] `next_eq`

Usages of these APIs should be refactored unless they're directly preceded by a `peek`:

 - [x] `next_char`
 - [x] `consume_char`

### Suggested implementation

`Source::next_byte` is an unsafe function, as it can break the invariant that `Source` must always be positioned on a UTF8 character boundary (i.e. not in the middle of a multi-byte Unicode char). It's preferable not to use it, to avoid littering the lexer with unsafe code.

However, `Source::next_char` is written to be much more transparent to the compiler than `Chars::next` is. So the compiler is able to optimize safe code like:

```rs
const dot_was_consumed = match lexer.source.peek_byte() {
  b'.' => {
    lexer.consume_char().unwrap();
    true
  }
  _ => false,
}
```

down to equivalent of:

```rs
const dot_was_consumed = match lexer.source.peek_byte() {
  b'.' => {
    unsafe {
      // This is a single assembler operation
      lexer.source.set_position( lexer.source.position().add(1) );
    }
    true
  }
  _ => false,
}
```

(at least that's what I remember from a few months ago when I checked it with [Godbolt](https://godbolt.org/)).

(originally mentioned in https://github.com/oxc-project/oxc/pull/3250#issuecomment-2108578763)

overlookmotel:
Closed in #4298 and #4304. Performance gain was sadly not very large. But every little helps, and if you don't try, you don't know...

PR #4298, opened 2024-07-16
perf(parser): optimize conditional advance on ASCII values

lucab:
Part of https://github.com/oxc-project/oxc/issues/3291.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -263,14 +263,16 @@ impl<'a> Lexer<'a> {
         self.source.peek_char2()
     }
 
-    /// Peek the next character, and advance the current position if it matches
-    #[inline]
-    fn next_eq(&mut self, c: char) -> bool {
-        let matched = self.peek() == Some(c);
-        if matched {
-            self.source.next_char().unwrap();
-        }
-        matched
+    /// Peek the next byte, and advance the current position if it matches
+    /// the given ASCII char.
+    #[allow(clippy::inline_always)]
+    #[inline(always)]
+    fn next_ascii_char_eq(&mut self, b: u8) -> bool {
+        // TODO: can be replaced by `std::ascii:Char` once stabilized.
+        // https://github.com/rust-lang/rust/issues/110998
+        assert!(b.is_ascii());
+        // SAFETY: `b` is a valid ASCII char.
+        unsafe { self.source.advance_if_ascii_eq(b) }
     }
 
     fn current_offset(&self) -> Span {
--- a/crates/oxc_parser/src/lexer/byte_handlers.rs
+++ b/crates/oxc_parser/src/lexer/byte_handlers.rs
@@ -209,8 +209,8 @@ ascii_byte_handler!(LIN(lexer) {
 // !
 ascii_byte_handler!(EXL(lexer) {
     lexer.consume_char();
-    if lexer.next_eq('=') {
-        if lexer.next_eq('=') {
+    if lexer.next_ascii_char_eq(b'=') {
+        if lexer.next_ascii_char_eq(b'=') {
             Kind::Neq2
         } else {
             Kind::Neq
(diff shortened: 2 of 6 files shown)


crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -263,14 +263,17 @@ impl<'a> Lexer<'a> {
...
+    /// the given ASCII char.
     #[inline]
-    fn next_eq(&mut self, c: char) -> bool {
-        let matched = self.peek() == Some(c);
-        if matched {
-            self.source.next_char().unwrap();
+    fn next_ascii_char_eq(&mut self, b: u8) -> bool {
+        // TODO: can be replaced by `std::ascii:Char` once stabilized.
+        // https://github.com/rust-lang/rust/issues/110998
+        if !b.is_ascii() {
```
overlookmotel, line 272: Maybe try replacing `if !b.is_ascii() { return false; }` with `assert!(b.is_ascii())`.

You'd think that'd create more verbose assembly, but counter-intuitively, sometimes `assert!` causes compiler to backtrack further and prove that the assertion cannot fail (which it can't if calling code is e.g. `self.next_ascii_char_eq(b'.')`, and therefore removes it entirely, where it won't do so for a normal `if` statement.

overlookmotel: Also, just as an experiment, you could remove `if !b.is_ascii() { return false; }` and not replace it with anything. This would be completely unsafe, and so could not be the final version, but it'd be interesting to see if that affected benchmarks.

overlookmotel:
Surprising and disappointing that this has no effect on benchmarks. I really thought it would, and if I remember right when I played with this back in Dec/Jan it did produce a speed-up.

I'm not sure what's going on! The problem previously was that `Chars` iterator was really slow as compiler didn't recognise that `self.remaining().chars().next() == Some('.')` only requires checking 1 byte, and generated a bunch of code to read a potentially multi-byte `char` before checking if it's `'.'`. Maybe between then and now `Chars` iterator has become more optimized?

overlookmotel:
Nope. `Chars` iterator still generates a ton more code on Rust 1.79.0 than checking by byte.

https://godbolt.org/z/dsv93Pnce

I'll take another look at this tomorrow with fresh eyes. I really hope I haven't wasted your time.

crates/oxc_parser/src/lexer/mod.rs
lucab: I've added both alternatives as separate commits, so you can compare the benchmark results in CodSpeed.
There aren't huge differences among all options, although some lexer and parser cases do see a 1% improvement.
There is also https://github.com/oxc-project/oxc/pull/4304 stacked on top of this and pushing the conversion further along; it shows similar numbers.

overlookmotel:
I am really struggling to understand why this isn't doing more on benchmarks.

Here is a more complete version of before vs after on Godbolt: https://godbolt.org/z/fjc96GfeE

* Old version `next_eq_dot_old` is 55 instructions.
* New version `next_eq_dot_new` is 11.

So the old code is way more verbose. It also requires 2 branches before it gets to the `cmp esi, 46` instruction, which is where it does the actual check for "is this byte `.`?". Granted, both those branches are almost never taken - the 1st branch is only taken when have reached EOF, and the 2nd only taken if the byte read is not ASCII (very rare) - so branch predictor will speculatively continue down the common path without delay. But still, I'd have expected that:

1. Even a few extra instructions in a path as hot as this would make a real difference, at least to the lexer benchmarks.
2. The excessive amount of code generated might cause more instruction cache to be consumed and therefore cause more instruction cache misses.

But no!

I've also tried adding `#[inline(always)]` to both the relevant functions to make sure inlining is definitely happening. But I didn't expect that to make any difference, and sure enough it doesn't.

The Godbolt link above also includes an alternative version `advance_if_ascii_eq2` which converts the branch which *is* unpredictable to straight line code. But I doubt that'll help much in practice - in actual code, we're always branching on the result of `next_ascii_char_eq` too, so that'd re-introduce the branch again.

@lucab I don't know how used you are to looking at assembly like this, but any ideas?

@strager I don't know if you're out there (and I know you deny being an expert in such matters!), but I'm a bit lost here and would really appreciate your thoughts if you have time.

DonIsaac:
Any updates on this PR?

Boshen:
I see no objections, merging.

We can also use `/usr/bin/time` to measure CPU instruction counts.

PR #4299, opened 2024-07-16
feat(linter): eslint-plugin-vitest/expect-expect 

eryue0220:
support [eslint-plugin-vitest/expect-expect](https://github.com/veritem/eslint-plugin-vitest/blob/main/src/rules/expect-expect.ts)


crates/oxc_linter/src/rules/jest/expect_expect.rs
```diff
@@ -486,5 +499,250 @@ fn test() {
         ),
     ];
 
-    Tester::new(ExpectExpect::NAME, pass, fail).with_jest_plugin(true).test_and_snapshot();
+    let pass_vitest = vec![
+        // Note: it's conflict with jest rule. In jest rule, it should fail.
+        // ("test.skip(\"skipped test\", () => {})", None),
```
mysteryven, line 504: We can add this condition when it's vitest rule, to verify it works, you can change it to:

```js
import { test } from 'vitest'
test.skip("skipped test", () => {})
```

https://github.com/veritem/eslint-plugin-vitest/blob/e885ca206506477342f6bfdc320c681373aba615/src/rules/expect-expect.ts#L93

eryue0220: Ok, let me try try 😂

eryue0220: Doen.

PR #4301, opened 2024-07-16
feat(linter): eslint-plugin-vitest/no-alias-methods

eryue0220:
support [eslint-plugin-vitest/no-alias-methods](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/no-alias-method.md)


crates/oxc_linter/src/snapshots/no_alias_methods.snap
```diff
@@ -105,3 +106,24 @@ source: crates/oxc_linter/src/tester.rs
    ·               ──────────────
    ╰────
   help: Replace "toThrowError" with its canonical name of "toThrow"
+
+  ⚠ eslint-plugin-jest(no-alias-methods): Unexpected alias "toBeCalled"
```
DonIsaac, line 110: Shouldn't this say `eslint-plugin-vitest`?

eryue0220: Whether it's `eslint-plugin-vitest` or not, depending on the code includes the [vitest](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/utils/mod.rs#L55). 

I think it's ok that we guarantee all of test cases pass from the vitest and jest. If need more precision, or we could add a configuration to specify the framework. 

@mysteryven  WDYT

mysteryven: @DonIsaac It's truly annoying while vitest test cases say jest things, the vitest rules are from jest rules with minor modifications, I didn't figure out a better way to reuse codes, so used `is_using_vitest` rudely. It will have problem if you use rules from command:

```bash
npx oxlint --vitest-plugin -D no-focused-tests
```

```bash
  × eslint-plugin-jest(no-focused-tests): Unexpected focused test.
   ╭─[a.js:1:4]
 1 │ it.only('test', () => {
   ·    ────
 2 │     // ...
   ╰────
  help: Remove focus from test.
```

Let's find a better way to solve the problem :(  I am responsible for this issue, we can merge eryue's PR.

DonIsaac (approved):

PR #4305, opened 2024-07-16
feat(linter): eslint/no-else-return

poyoho:


rzvxa:
Hi, First off thanks for your contribution😍

I have some thoughts on this, I know the eslint rule is implemented in the same way you've implemented this one. But the performance isn't so great with this approach.
That's why I was thinking maybe we can use the control flow graph for this purpose instead. But that would require a major change in what you've already done.

So before going with that - or ruling it out - is it possible for you to do these 2 experiments to see the benchmark for it?

1. running the rule on `return` statements and going up to see if we are in an if statement or not(which IMHO I think wouldn't improve the performance or even worse it might degrade it even further)
2. running the rule on `node.is_function_like`s and check they contain an appropriate if statement to lint

@mysteryven @DonIsaac Do you guys want to chime in?

Side note: It seems like this PR still is a work in progress, It would be better if you keep a WIP PR as `Draft` and only mark it for review when it is ready to be reviewed and merged.

poyoho:
I tinker repeatedly executing `always_return` on the same if_stmt to make the performace bad. may be I can save the if_stmt the result of `always_return` the reduce run this function?

DonIsaac:
hey @poyoho :wave: thanks so much for your PR! Do you need some help getting this PR over the finish line?

poyoho:
I will try again

crates/oxc_linter/src/rules/eslint/no_else_return.rs
```diff
@@ -0,0 +1,384 @@
...
+
+fn naive_has_return(node: &Statement) -> bool {
+    match node {
+        Statement::BlockStatement(block) => {
+            let Some(last_child) = block.body.last() else {
+                return false;
+            };
+            matches!(last_child, Statement::ReturnStatement(_))
+        }
+        node => matches!(node, Statement::ReturnStatement(_)),
```
DonIsaac, line 112: This should be part of the match statement its nested in

DonIsaac (approved):

DonIsaac:
👏 👏 👏

poyoho:
Do you know how to fix `CI / Test (ubuntu-latest) and CI / Test (macos-14)` @DonIsaac

DonIsaac:
You need to specify a fix kind in declare_oxc_lint. Check out other rules for examples.

DonIsaac:
🎉🎉🎉

PR #4324, opened 2024-07-17
chore(tasks/rulegen): support eslint-plugin-import rule generation

jelly:


jelly:
This seems to be a bit too hard to implement, closing.

PR #4328, opened 2024-07-17
perf(semantic): calculate number of nodes, scopes, symbols, references before visiting AST

Dunqing:
Note by @overlookmotel 20/7/24: Even though this PR shows as merged, that was by [some weird malfunction of Github](https://github.com/oxc-project/oxc/pull/4328#issuecomment-2239472555). It was immediately reverted, and the merge scrubbed from history of `main` branch. #4367 continues the work.


overlookmotel:
Is this related to https://github.com/oxc-project/backlog/issues/31?

A couple of thoughts:

1. I'd imagined counting them in parser, rather than doing a whole extra AST pass in Semantic.
2. Even doing it in Semantic, as in this PR, if you use the collected counts to initialize `ScopeTree` + `SymbolTable`'s `Vec`s with sufficient capacity, it may turn out that delivers a perf boost which is greater than the loss shown at present on this PR - so overall it could be positive.

overlookmotel:
Wow! This is banging! Much more than I expected.

Only problem is `semantic[RadixUIAdoptionSection.jsx]`. Personally, I think that particular benchmark is quite important because it's the most representative of actual application code.

I think the reason it's suffering is:

* The loss is the extra traversal in `Collect`.
* The gain is the `Vec`s not growing, with the memory copying involved.
* The cost of growing the `Vec`s is massive when they're big, not so much when they're small.
* So for the large file benchmarks, the gain outweighs the loss, but for little `RadixUIAdoptionSection.jsx`, the loss is the stronger factor.

In my view we should not merge this as is because of the significant negative effect on that 1 benchmark. But what this experiment *has* done is show us there's a massive speed boost there for the taking, and that we should probably prioritize it above other perf work.

Boshen:
I'm in favor of merging this and have a new issue on removing this ast pass, which is a clearer task.

overlookmotel:
Does Rolldown's process involve running Semantic at all? If it does, merging this I'd imagine could hurt them, because `RadixUIAdoptionSection.jsx` is probably the closest of our benchmarks to real-world app code Rolldown processes.

A compromise would be to use a heuristic to enable/not enable the extra pass. e.g. only run on source files larger than 10 KB. That'd probably put us in the green for all benchmarks. What do you think of that option?

DonIsaac:
I've opened a similar PR tackling this in a slightly different way: #4332. Sorry I didn't know what this one was doing, I was a little confused by the title! #4332 uses source text length as a heuristic for reserving memory instead of performing a traversal. It seems to perform worse overall than @Dunqing's approach, which I find quite interesting.

overlookmotel:
By chance, @DonIsaac has tackled the same issue today on #4332 - but taking a different approach.

Don's approach is to estimate the required size of the `Vec`s from source text size. This has pluses and minuses:

* Advantage: Avoid the overhead of the `Collect` pass.
* Disadvantage: Potentially getting the estimate wrong.

He's getting even better results on some benchmarks than this PR, but worse on others.

If I can throw in my 2 cents as well...

From my (fairly little) knowledge of OS-level memory stuff, I believe that allocating *too much* memory is not necessarily a problem. On architectures with virtual memory (which includes all the ones Oxc targets except WASM), allocating a huge slab of memory (e.g. 1 GB) doesn't actually consume 1 GB of physical memory. It only consumes 1 GB of *virtual* memory space. Pages of physical memory to "back" that virtual memory will only be allocated when that chunk of the memory is first written to. Like I said, I could be wrong about this - it's not an area I really know - but that's my rough understanding.

Which is all a long way of saying: possibly the best strategy (except on WASM) is to estimate too high because that costs very little, whereas estimating too low will lead to the `Vec`s reallocating, which is *really* expensive when they're large.

The big question though is where is this -6% perf penalty on the `RadixUIAdoptionSection.jsx` benchmark coming from? Don's approach is also getting -6%, despite taking a different approach. So the explanation I gave in [comment above](https://github.com/oxc-project/oxc/pull/4328#issuecomment-2233938510) is probably wrong.

I don't really have much idea on that one. A good first step might be to figure out what the optimal size of all these `Vec`s is for each benchmark (i.e. how many scopes/symbols/nodes are actually needed for each) and test initializing the `Vec`s to exactly those sizes. That'd show us the upper bound on perf improvement we can get, and then we could measure different strategies against that, and try to figure out where the -6% is coming from. @DonIsaac would you like to test that out on your PR?

DonIsaac:
Sure, I can try that out. Another possible option is to record symbols/AST nodes/etc that we find during the parse step. Thoughts?

overlookmotel:
> Sure, I can try that out. Another possible option is to record symbols/AST nodes/etc that we find during the parse step. Thoughts?

Yes, that is a good option, and it's what I originally had in mind when I wrote https://github.com/oxc-project/backlog/issues/31.

But:

1. Your PR #4332 has raised another way of doing it which I think is also worth investigating.
2. Counting everything in the parser is going to be a bit complicated, and we should do that as a follow-up.

Sorry, I think I've made everything too complicated! As [Boshen's said above](https://github.com/oxc-project/oxc/pull/4328#issuecomment-2233956849), let's get this PR merged, and then iterate on it.

So I propose:

1. Merge this PR now.
2. Follow-up PR ASAP to add a guard which disables the `Collect` pass if file is below a certain size, to get the `RadixUIAdoptionSection` benchmark back up to where it was before (which we must merge before next Oxc release, to avoid hurting downstream consumers).
3. Open another issue to discuss further improvement, where we can also consider your approach from #4332 and/or counting in parser.

@Dunqing, @DonIsaac Does that sound OK to you?

Dunqing:
> 1. Merge this PR now.
> 2. Follow-up PR ASAP to add a guard which disables the `Collect` pass if file is below a certain size, to get the `RadixUIAdoptionSection` benchmark back up to where it was before (which we must merge before next Oxc release, to avoid hurting downstream consumers).
> 3. Open another issue to discuss further improvement, where we can also consider your approach from [perf(semantic): initialize builder storage with pre-allocated capacity #4332](https://github.com/oxc-project/oxc/pull/4332) and/or counting in parser.
> 
> @Dunqing, @DonIsaac Does that sound OK to you?

Sounds good to me. Both of these approaches look good. Which one you would favor, I have no opinion on that.

Boshen:
109.1 µs -> 116.6 µs

Don't forget to look at the absolute numbers, -6% is huge but 7 µs is small.

crates/oxc_semantic/src/builder.rs
```diff
@@ -100,6 +100,39 @@ pub struct SemanticBuilderReturn<'a> {
     pub errors: Vec<OxcDiagnostic>,
 }
 
+#[derive(Default)]
+pub struct Collector {
```
Boshen, line 104: We should move this to another file and have some doc comments explaning the why before merge.

Boshen:
> Counting everything in the parser is going to be complicated

AST builder can do the counting :-)

overlookmotel:
> AST builder can do the counting :-)

Yes, but currently `AstBuilder` is `Copy` which we'll need to change to make it stateful, some places bypass the builder, and counting scopes is dependent on some surrounding context because some scopes are conditional. All totally doable, but not a completely trivial amount of work.

> Don't forget to look at the absolute numbers, -6% is huge but 7 µs is small.

I thought % change was more important, on grounds that a large application might contain 10,000+ files this size (particularly if you include `node_modules`), so the perf impact will scale. Is my logic off?

Boshen:
Let's run oxlint locally on the vscode repo and see if there are any significant performance change.

Dunqing:
I tested this PR on `Rolldown`. The [benchmark](https://codspeed.io/rolldown/rolldown/branches/try-next-oxc) shows 1% ~ 3% faster performance!

Also, I tested locally using `oxlint`. The benchmark results for all three repos are 2% to 4% slower than before. I'm not sure what is causing this

### cal.com
```shell
hyperfine --warmup 30 "/Users/qing/p/github/oxc/target/debug/oxlint-before packages"  "/Users/qing/p/github/oxc/target/debug/oxlint packages"  -i
Benchmark 1: /Users/qing/p/github/oxc/target/debug/oxlint-before packages
  Time (mean ± σ):     553.3 ms ±  12.6 ms    [User: 3756.0 ms, System: 84.5 ms]
  Range (min … max):   542.8 ms … 583.4 ms    10 runs

Benchmark 2: /Users/qing/p/github/oxc/target/debug/oxlint packages
  Time (mean ± σ):     572.0 ms ±  17.7 ms    [User: 3830.0 ms, System: 84.8 ms]
  Range (min … max):   558.9 ms … 619.2 ms    10 runs

Summary
  /Users/qing/p/github/oxc/target/debug/oxlint-before packages ran
    1.03 ± 0.04 times faster than /Users/qing/p/github/oxc/target/debug/oxlint packages
```

### TypeScript
```shell
hyperfine --warmup 3 "/Users/qing/p/github/oxc/target/debug/oxlint-before src"  "/Users/qing/p/github/oxc/target/debug/oxlint src"  -i
Benchmark 1: /Users/qing/p/github/oxc/target/debug/oxlint-before src
  Time (mean ± σ):      4.942 s ±  0.194 s    [User: 10.907 s, System: 0.061 s]
  Range (min … max):    4.784 s …  5.427 s    10 runs

Benchmark 2: /Users/qing/p/github/oxc/target/debug/oxlint src
  Time (mean ± σ):      5.163 s ±  0.353 s    [User: 11.002 s, System: 0.079 s]
  Range (min … max):    4.811 s …  5.878 s    10 runs

Summary
  /Users/qing/p/github/oxc/target/debug/oxlint-before src ran
    1.04 ± 0.08 times faster than /Users/qing/p/github/oxc/target/debug/oxlint src
```

### VSCode

```shell
hyperfine --warmup 5 "/Users/qing/p/github/oxc/target/debug/oxlint-before src" "/Users/qing/p/github/oxc/target/debug/oxlint src" -i
Benchmark 1: /Users/qing/p/github/oxc/target/debug/oxlint-before src
  Time (mean ± σ):      4.319 s ±  0.211 s    [User: 29.747 s, System: 0.259 s]
  Range (min … max):    4.063 s …  4.598 s    10 runs

Benchmark 2: /Users/qing/p/github/oxc/target/debug/oxlint src
  Time (mean ± σ):      4.405 s ±  0.152 s    [User: 31.028 s, System: 0.259 s]
  Range (min … max):    4.236 s …  4.771 s    10 runs

Summary
  /Users/qing/p/github/oxc/target/debug/oxlint-before src ran
    1.02 ± 0.06 times faster than /Users/qing/p/github/oxc/target/debug/oxlint src
```

overlookmotel:
I think I've figured out the source of the problem (but not the solution!).

### The cause

For reasons I don't understand, `Collector` is getting the wrong counts. Here's a comparison of the counts `Collector` gets vs the length of the `Vec`s in actuality after running `visit_program`:

(under-estimates in **bold**, overestimates in *italics*)

#### RadixUIAdoptionSection.jsx

| Type      | Collector count | Actual count |
|-----------|-----------------|--------------|
| node      | 313             | **392**      |
| scope     | 2               | 2            |
| symbol    | 11              | 11           |
| reference | 68              | *28*         |

#### antd.js

| Type      | Collector count | Actual count |
|-----------|-----------------|--------------|
| node      | 441296          | **571656**   |
| scope     | 12318           | 12318        |
| symbol    | 33946           | *33833*      |
| reference | 96414           | 96414        |

#### cal.com.tsx

| Type      | Collector count | Actual count |
|-----------|-----------------|--------------|
| node      | 120086          | **152935**   |
| scope     | 2501            | 2501         |
| symbol    | 8333            | *5233*       |
| reference | 24612           | *16749*      |

#### checker.ts

| Type      | Collector count | Actual count |
|-----------|-----------------|--------------|
| node      | 212822          | **303404**   |
| scope     | 10233           | 10233        |
| symbol    | 15040           | **15056**    |
| reference | 75598           | *75587*      |

#### pdf.mjs

| Type      | Collector count | Actual count |
|-----------|-----------------|--------------|
| node      | 90239           | **108347**   |
| scope     | 3856            | 3856         |
| symbol    | 4530            | 4530         |
| reference | 13578           | 13578        |

The code used to generate these numbers [here](https://github.com/oxc-project/oxc/commit/aa42e9153630baf20f28764254652a765d4ab193).

### What's going on?

`Collector`'s count of AST nodes is consistently well below the actual number. I don't see any mistake in `Collector`'s code, so I suspect there's a bug somewhere else in `SemanticBuilder` - probably it's generating 2 AST node records for some AST nodes.

There's also one oddity in `checker.ts` - `Collector` there is underestimating the number of symbols too. That's weird.

The over-estimates on symbol and reference counts are expected and are not a big problem. As I mentioned [above](https://github.com/oxc-project/oxc/pull/4328#issuecomment-2234367769), over-estimating is not too expensive - it's under-estimating which has a massive cost.

If `Collector` got accurate counts, these are the perf gains we'd get:

https://codspeed.io/oxc-project/oxc/branches/test-visit2 (*excluding* overhead of `Collector` visit pass)
https://codspeed.io/oxc-project/oxc/branches/test-visit3 (*including* overhead of `Collector` visit pass)

Up to +30% improvement!

### Next steps

It looks like first thing we should do is figure out why `SemanicBuilder` is generating more `AstNode` records than it should. Fixing that will be a perf improvement in itself, and also will make this PR deliver all the perf gain which it has potential to.

overlookmotel:
Linter benchmark doesn't include running semantic, whereas transformer benchmark does. Probably that's the reason. The +1% on linter benchmark shown on `test-visit3` branch is probably just noise.

Dunqing:
> Linter benchmark doesn't include running semantic, whereas transformer benchmark does. Probably that's the reason. The +1% on linter benchmark shown on `test-visit3` branch is probably just noise.

I'm looking in the wrong

But as far as I know, `linter` can't do anything without `semantic`.
https://github.com/oxc-project/oxc/blob/58f6ec21f3209de88c7db786e07012703854b860/tasks/benchmark/benches/linter.rs#L31-L35

overlookmotel:
Actually the % increases shown on CodSpeed for `test-visit2` branch (which doesn't run `Collector` visit) are inaccurate. Codspeed is comparing to the wrong base, for some reason.

Actual improvement for `antd.js`:

Main: 175.8 ms
test-visit2: 120.8 ms
Improvement: 46% !!!

That's what we can expect if we move counting into parser, or use a simpler estimation method (like #4332).

All I can say is... Holy guacumole!

overlookmotel:
> But as far as I know, `linter` can't do anything without `semantic`.

Yes, true. But the linter benchmark generates `Semantic` outside of the loop which is measured. This is the measurement loop:

https://github.com/oxc-project/oxc/blob/58f6ec21f3209de88c7db786e07012703854b860/tasks/benchmark/benches/linter.rs#L50-L55

Whereas transformer benchmark measure includes both parsing and semantic (generating `Semantic` happens within `Transformer::build`):

https://github.com/oxc-project/oxc/blob/58f6ec21f3209de88c7db786e07012703854b860/tasks/benchmark/benches/transformer.rs#L20-L36

Dunqing:
> I think I've figured out the source of the problem (but not the solution!).
> 
> ### The cause
> For reasons I don't understand, `Collector` is getting the wrong counts. Here's a comparison of the counts `Collector` gets vs the length of the `Vec`s in actuality after running `visit_program`:
> 
> (under-estimates in **bold**, overestimates in _italics_)
> 
> #### RadixUIAdoptionSection.jsx
> Type	Collector count	Actual count
> node	313	**392**
> scope	2	2
> symbol	11	11
> reference	68	_28_
> #### antd.js
> Type	Collector count	Actual count
> node	441296	**571656**
> scope	12318	12318
> symbol	33946	_33833_
> reference	96414	96414
> #### cal.com.tsx
> Type	Collector count	Actual count
> node	120086	**152935**
> scope	2501	2501
> symbol	8333	_5233_
> reference	24612	_16749_
> #### checker.ts
> Type	Collector count	Actual count
> node	212822	**303404**
> scope	10233	10233
> symbol	15040	**15056**
> reference	75598	_75587_
> #### pdf.mjs
> Type	Collector count	Actual count
> node	90239	**108347**
> scope	3856	3856
> symbol	4530	4530
> reference	13578	13578
> The code used to generate these numbers [here](https://github.com/oxc-project/oxc/commit/aa42e9153630baf20f28764254652a765d4ab193).
> 
> ### What's going on?
> `Collector`'s count of AST nodes is consistently well below the actual number. I don't see any mistake in `Collector`'s code, so I suspect there's a bug somewhere else in `SemanticBuilder` - probably it's generating 2 AST node records for some AST nodes.
> 
> There's also one oddity in `checker.ts` - `Collector` there is underestimating the number of symbols too. That's weird.
> 
> The over-estimates on symbol and reference counts are expected and are not a big problem. As I mentioned [above](https://github.com/oxc-project/oxc/pull/4328#issuecomment-2234367769), over-estimating is not too expensive - it's under-estimating which has a massive cost.
> 
> If `Collector` got accurate counts, these are the perf gains we'd get:
> 
> https://codspeed.io/oxc-project/oxc/branches/test-visit2 (_excluding_ overhead of `Collector` visit pass) https://codspeed.io/oxc-project/oxc/branches/test-visit3 (_including_ overhead of `Collector` visit pass)
> 
> Up to +30% improvement!
> 
> ### Next steps
> It looks like first thing we should do is figure out why `SemanticBuilder` is generating more `AstNode` records than it should. Fixing that will be a perf improvement in itself, and also will make this PR deliver all the perf gain which it has potential to.

Ah, I see, Thanks for explaining this!

Boshen:
Nice findings! I'll take over the investigation so you guys can work on some business logic 😁

overlookmotel:
Dunqing found the bug!

codspeed

Dunqing:
> Dunqing found the bug!
I forgot these identifiers also is a node 😅

crates/oxc_semantic/src/builder.rs
```diff
@@ -100,6 +100,58 @@ pub struct SemanticBuilderReturn<'a> {
...
+    #[inline]
+    fn enter_node(&mut self, _: AstKind<'a>) {
+        self.node += 1;
+    }
+    #[inline]
+    fn enter_scope(&mut self, _: ScopeFlags, _: &Cell<Option<ScopeId>>) {
+        self.scope += 1;
+    }
+    #[inline]
+    fn leave_node(&mut self, _: AstKind<'a>) {}
```
overlookmotel, line 121: I think this empty method + `leave_scope` below can be removed.

crates/oxc_semantic/src/builder.rs
```diff
@@ -100,6 +100,58 @@ pub struct SemanticBuilderReturn<'a> {
...
+    }
+    fn visit_jsx_identifier(&mut self, _: &JSXIdentifier<'a>) {
+        self.reference += 1;
+        self.node += 1;
+    }
+    fn visit_jsx_member_expression_object(&mut self, it: &JSXMemberExpressionObject<'a>) {
+        if let JSXMemberExpressionObject::MemberExpression(expr) = &it {
+            self.visit_jsx_member_expression(expr);
+        } else {
+            self.node += 1;
```
overlookmotel, line 141: Since you're counting `JSXIdentifier` as a reference, shouldn't this branch also have `self.reference += 1;`?

Boshen (commented):
Should we have a debug assert check that final vec length is similar to node count?

overlookmotel:
> Should we have a debug assert check that final vec length is similar to node count?

Yes it'd be a good idea to add some assertions.

* Node count should be 100% accurate.
* Reference count should be 100% accurate (I think).
* Scope count + Symbol count in `Collector` should be `>=` actual counts.

I don't think we should add assertions bounding how much `Collector` overestimates the last 2. Some weird files could causes a large over-estimate. e.g.: 

* `var x; var x; /* 1000 more... */ var x;` (over-estimate symbols)
* `try {} catch {}; try {} catch {}; /* 1000 more... */` (over-estimate scopes)

crates/oxc_semantic/src/builder.rs
Dunqing: https://github.com/oxc-project/oxc/blob/b4a8d675e0dd9999f2098000fecbb4330470d45e/crates/oxc_semantic/src/builder.rs#L1966-L1975

`JSXIdentifier` is not always a reference. you can see my latest [commit](https://github.com/oxc-project/oxc/pull/4328/commits/b4a8d675e0dd9999f2098000fecbb4330470d45e).

overlookmotel: Two things:

In that case shouldn't it call `self.visit_jsx_identifier`?

I don't think we necessarily need to make the counts 100% accurate. Over-estimating is fine. It may be that the cost of complicating `Collector` to make an accurate count (and so making it a bit slower) is worse than the cost of allocating too much space in the `Vec`s. It's *under*-estimating that we absolutely need to avoid.

overlookmotel:
@Dunqing I see you're still working on this, so going to leave you alone to get on with it. Please shout when you're happy with it and would like a review.

Dunqing:
I corrected the issue with the `Collector's` incorrect count. Currently 100% accurate.

Dunqing:
Maybe we can do more, even count the number of each scope bindings?

crates/oxc_semantic/src/builder.rs
Dunqing: > Two things:
> 
> In that case shouldn't it call `self.visit_jsx_identifier`?

If I call the `visit_jsx_identifier`, the reference count will increase. But it's not a `reference`

overlookmotel:
> I corrected the issue with the `Collector's` incorrect count. Currently 100% accurate.

I assume you're only referring to nodes count here?

Is symbol count for `checker.ts` right now? Previously it was an under-estimate, which is a problem.

I don't think we can make symbol (aka binding) count completely accurate without building scope tree (which would defeat the point of `Collector`) because of `var x; var x;` (2 x binding identifiers, but only 1 symbol).

As I said above, I don't think we necessarily need to make the counts 100% accurate. Over-estimating is fine. It may be that the cost of complicating `Collector` to make an accurate count (and so making it a bit slower) is worse than the cost of allocating too much space in the `Vec`s. It's *under*-estimating that we absolutely need to avoid.

> Maybe we can do more, even count the number of each scope bindings?

Yes, maybe that would be good - could size the bindings hash map for each scope correctly up front. But that's a bit more complicated. I'd suggest investigating that in a follow-up.

Have you finished all your changes and ready for review?

Dunqing:
> I assume you're only referring to nodes count being accurate here?
> 
> Is symbol count for `checker.ts` right now? Previously it was an under-estimate, which is a problem.
> 
> I don't think we can make symbol (aka binding) count completely accurate without building scope tree (which would defeat the point of `Collector`) because of `var x; var x;` (2 x binding identifiers, but only 1 symbol).
> 
> As I said above, I don't think we necessarily need to make the counts 100% accurate. Over-estimating is fine. It may be that the cost of complicating `Collector` to make an accurate count (and so making it a bit slower) is worse than the cost of allocating too much space in the `Vec`s. It's _under_-estimating that we absolutely need to avoid.
> 

I got you mean. You are right. The counting is not complicated as it stands, we are not dealing with `var x, var x` so comparing the actual number is an overestimate

Dunqing:
Rolldown's [codspeed](https://codspeed.io/rolldown/rolldown/branches/try-next-oxc) shows 1% ~ 4% faster than before

overlookmotel:
I checked counts on latest version (2nd `fix: stack overflow` commit):

* Radix
  * All estimated exactly.
* antd
  * Symbols over-estimated a bit (no problem).
* cal.com:
  * Nodes slightly over-estimated (shouldn't happen).
  * Symbols very over-estimated (no problem, expected).
  * References over-estimated a bit (shouldn't happen but not a big deal).
* checker:
  * Nodes slightly over-estimated (shouldn't happen).
  * Symbols slightly under-estimated (**problem**).
  * References over-estimated a bit (shouldn't happen but not a big deal).
 * pdf.mjs
   * All estimated exactly.

The most problematic one is `Collector` undercounting scopes in `checker.ts`. That means the scopes `Vec`s are re-allocating during `SemanticBuilder`'s traversal, which will be a bit of a slow-down.

crates/oxc_semantic/src/builder.rs
overlookmotel: Unless I'm reading `reference_jsx_identifier` wrong, it *does* create a reference for a `JSXIdentifier` whose parent is a `JSXMemberExpressionObject`. `Some(AstKind::JSXMemberExpressionObject(_)) => {}` lets execution fall through to next block of code which creates the reference.

That would be correct behavior - in `<Foo.bar />`, `Foo` is a reference.

But this isn't easy to reason about. #3528 would make it much simpler.

Dunqing:
> The most problematic one is `Collector` undercounting scopes in `checker.ts`. That means the scopes `Vec`s are re-allocating during `SemanticBuilder`'s traversal, which will be a bit of a slow-down.

I found nodes over-estimated caused by visit_arrow_function missing visit `return_type`. 

When #4364 is merged. Only symbols are very over-estimated.

```shell
[crates/oxc_semantic/src/builder.rs:289:13] collector = Collector {
    node: 314759,
    scope: 10554,
    symbol: 15834,
    reference: 78443,
}
[crates/oxc_semantic/src/builder.rs:292:13] Collector {
    node: self.nodes.len(),
    scope: self.scope.len(),
    symbol: self.symbols.len(),
    reference: self.symbols.references.len(),
} = Collector {
    node: 314759,
    scope: 10554,
    symbol: 15674,
    reference: 78443,
}
[Finished running. Exit status: 0]
```

Dunqing:
I don’t why this PR is merged !!!! I don't click the merge button. And I don't see a revert button

Dunqing:
Please see #4367

milesj:
Y'all are CRACKED

PR #4332, opened 2024-07-17
perf(semantic): initialize builder storage with pre-allocated capacity

DonIsaac:
## What This PR Does

Initializes `nodes`, `scopes`, `symbols`, and several other internal stacks with some capacity already allocated. This should reduce the number of re-allocations `Semantic` needs to make while traversing a program.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -102,6 +102,8 @@ pub struct SemanticBuilderReturn<'a> {
 
 impl<'a> SemanticBuilder<'a> {
     pub fn new(source_text: &'a str, source_type: SourceType) -> Self {
+        // Initialize scope tree with a capacity of 1/256th of the source text length. The bitwise
+        // or of two ensures there's at least 2 slot even for small programs.
         let scope = ScopeTree::with_capacity(source_text.len() >> 8 | 2);
         let current_scope_id = scope.root_scope_id();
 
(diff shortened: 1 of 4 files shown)


DonIsaac:
@overlookmotel any insight into why this makes some benchmarks perform better while others perform worse?

overlookmotel:
@DonIsaac Ha! By chance, Dunqing has been working on exactly the same thing today, but taking a slightly different approach - #4328. It's unfortunate his PR title is quite mysterious "test: visit" so you didn't see it - I think he intended it initially as a quick temporary test.

But... actually it's quite useful you've investigated this other heuristic approach because Dunqing's approach has run into same problem you have here with `RadixUIAdoptionSection.jsx`.

Let's join forces!

I'll post more comment on #4328 to continue the conversation in one place.

DonIsaac:
Love it! LMK if you're reachable on Discord as well.

CC: @Dunqing

crates/oxc_semantic/src/builder.rs
```diff
@@ -124,11 +126,12 @@ impl<'a> SemanticBuilder<'a> {
             current_reference_flag: ReferenceFlag::empty(),
             current_scope_id,
             current_scope_depth: 0,
-            function_stack: vec![],
-            namespace_stack: vec![],
-            nodes: AstNodes::default(),
+            function_stack: Vec::with_capacity(4),
+            // Most TS modules are not nested, and those that are are usually only 1 level deep.
+            namespace_stack: Vec::with_capacity(2),
+            nodes: AstNodes::with_capacity(source_text.len() >> 4),
```
overlookmotel, line 132: @DonIsaac Just a side note: There's no need to write `>> 4` here. When you multiply or divide by a power of 2, the compiler will convert that to a bitshift operation. So `/ 16` is preferable, as it's easier to read for us humans, and you don't lose any perf.

PR #4346, opened 2024-07-18
refactor(ast): scope is created only if CatchClause has param

### LINKED ISSUE #4345: Make `CatchClause` scope optional
`CatchClause` does not need a scope if there's no param e.g. `try {} catch {}`. Add back `#[scope(if(self.param.is_some()))]`.

As discussed in https://github.com/oxc-project/oxc/pull/4192#pullrequestreview-2171374436, there's an argument for making all scopes unconditional. But in my opinion, either way the most important thing is to be consistent, so scopes-related code is easier to reason about.

At present, our policy is not to create a scope if none is needed (because it has no bindings). So we should keep to that policy for `CatchClause`. If we want to change the policy, we can do that, but we should do it consistently for *all* scopes.

Dunqing:
close: #4345

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1300,7 +1300,7 @@ pub struct TryStatement<'a> {
 }
 
 #[visited_node]
-#[scope(flags(ScopeFlags::CatchClause))]
+#[scope(flags(ScopeFlags::CatchClause), if(self.param.is_some()))]
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#4346** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4346?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):

PR #4348, opened 2024-07-18
fix(sourcemap): avoid negative line if token_chunks has same prev_dst…

underfin:
…_line


DonIsaac:
Can we add test cases for this?

underfin:
Yeah. I already uopdate related test.

PR #4349, opened 2024-07-18
refactor(linter): use diagnostic codes in lint rules

DonIsaac:
> This PR is (unfortunately) quite large, but all changes are needed in tandem for this to work properly.

## What This PR Does

Updates the linter to populate diagnostics reported by rules with error codes statically derived from `RuleMeta` + `RuleEnum`.

Doing so required changing how we handle vitest rules. I know @mysterven was hoping to refactor that part of the code, and I think this approach is an improvement (but could probably be cleaned up further).

### 1. Auto-Populate Error Codes
`LintContext` now sets an error code scope + error code number for diagnostics reported by lint rules. `LintContext` will not clobber existing codes set by rules, allowing for rule-specific override behavior (e.g. to use `eslint-plugin-react-hooks` as an error scope).

In order to accomplish this, I had to update every diagnostic factory for every rule. While doing this I found some incorrect error messages, or messages that could be easily improved. This is where a large majority of the snapshot diffs come from. Additionally, I was able to reduce string allocations from `format!` usages in diagnostic factories, especially within jest rules.

### 2. Framework and Library Detection
This PR adds `FrameworkFlags`, which specify what (if any) set of libraries and frameworks are being used by a project and/or file. They are passed in two ways:

1. `LintOptions` can specify a set of `framework_hints` that apply to the entire target codebase. Right now these are always empty, but I'm thinking in the future we could sniff `package.json`. It may be helpful for enabling/disabling default rules.
2. When `Linter` gets run on a file, framework information is sniffed from the `LintContext`. Right now, we are only checking for `vitest` imports in `ModuleRecord` and test path prefixes from `source_path`. It may be useful to do something similar for React/NextJS rules in the future. I know that [next/no-html-link-for-pages](https://nextjs.org/docs/messages/no-html-link-for-pages) could benefit greatly from this.

### DIFF
--- a/crates/oxc_diagnostics/src/lib.rs
+++ b/crates/oxc_diagnostics/src/lib.rs
@@ -100,6 +100,11 @@ impl Diagnostic for OxcDiagnostic {
             .map(Box::new)
             .map(|b| b as Box<dyn Iterator<Item = LabeledSpan>>)
     }
+
+    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        // self.code.is_some().then(|| Box::new(&self.code) as Box<dyn Display>)
+        None
+    }
 }
 
 impl OxcDiagnostic {
@@ -138,21 +143,29 @@ impl OxcDiagnostic {
 
     #[inline]
     pub fn with_error_code_scope<T: Into<Cow<'static, str>>>(mut self, code_scope: T) -> Self {
-        self.inner.code.scope = Some(code_scope.into());
+        self.inner.code.scope = match self.inner.code.scope {
+            Some(scope) => Some(scope),
+            None => Some(code_scope.into()),
+        };
         debug_assert!(
             self.inner.code.scope.as_ref().is_some_and(|s| !s.is_empty()),
             "Error code scopes cannot be empty"
         );
+
         self
     }
 
     #[inline]
     pub fn with_error_code_num<T: Into<Cow<'static, str>>>(mut self, code_num: T) -> Self {
-        self.inner.code.number = Some(code_num.into());
+        self.inner.code.number = match self.inner.code.number {
+            Some(num) => Some(num),
+            None => Some(code_num.into()),
+        };
         debug_assert!(
             self.inner.code.number.as_ref().is_some_and(|n| !n.is_empty()),
             "Error code numbers cannot be empty"
         );
+
         self
     }
 
--- a/crates/oxc_linter/src/snapshots/consistent_test_it.snap
+++ b/crates/oxc_linter/src/snapshots/consistent_test_it.snap
@@ -1,15 +1,14 @@
 ---
 source: crates/oxc_linter/src/tester.rs
-assertion_line: 216
 ---
-  ⚠ eslint-plugin-jest(consistent-test-it): Enforce `test` and `it` usage conventions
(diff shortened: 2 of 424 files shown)


DonIsaac:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4349?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4349** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4336** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4336?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4335** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4335?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4334** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4334?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_diagnostics/src/lib.rs
```diff
@@ -100,6 +100,11 @@ impl Diagnostic for OxcDiagnostic {
             .map(Box::new)
             .map(|b| b as Box<dyn Iterator<Item = LabeledSpan>>)
     }
+
+    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        // self.code.is_some().then(|| Box::new(&self.code) as Box<dyn Display>)
+        None
```
DonIsaac, line 106: Right now, I'm rendering codes via `Display` for `OxcDiagnostic` for two reasons:
1. The graphical reporter renders error codes above error messages. (NOTE: do we want to change this behavior?)
2. Doing this lets stringification of diagnostics keep the error code.

Any opinions on how we should proceed?

mysteryven: Or replace our `plugin-name(rule-name)` part with error code?

CleanShot 2024-07-19 at 15 35 02@2x

DonIsaac: cc: @Boshen

mysteryven (approved):

mysteryven:
Let's merge this, It's good to me, and Donlaac needs to resolve conflict every time a linter rule added. The `code` question above needs Boshen's advice then.

crates/oxc_linter/src/snapshots/consistent_test_it.snap
```diff
@@ -108,98 +107,98 @@ assertion_line: 216
...
-  ⚠ eslint-plugin-jest(consistent-test-it): Enforce `test` and `it` usage conventions
+  ⚠ eslint-plugin-vitest(consistent-test-it): Enforce `test` and `it` usage conventions
    ╭─[consistent_test_it.tsx:1:1]
  1 │ test.skip("foo")
    · ────
    ╰────
   help: Prefer using "it" instead of "test"
 
-  ⚠ eslint-plugin-jest(consistent-test-it): Enforce `test` and `it` usage conventions
+  ⚠ eslint-plugin-vitest(consistent-test-it): Enforce `test` and `it` usage conventions
```
armano2, line 138: are you sure that this is correct?

PR #4352, opened 2024-07-18
fix: impl PartialEq<str> for CompactStr

DonIsaac:
I need this for another PR that I'm working on, but I think adding this just makes sense.

### DIFF
--- a/crates/oxc_span/src/atom.rs
+++ b/crates/oxc_span/src/atom.rs
@@ -297,6 +297,18 @@ impl PartialEq<CompactStr> for &str {
     }
 }
 
+impl PartialEq<CompactStr> for str {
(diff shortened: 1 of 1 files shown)


overlookmotel:
Nice. It has always been an annoyance that you can't compare `&compact_str == "foo"`.

But do you want to also add the reverse comparison `impl PartialEq<CompactStr> for str`? This PR at present makes `&compact_str == "foo"` valid, but not `"foo" == &compact_str`.

DonIsaac:
> `impl PartialEq<CompactStr> for str`?

Done

overlookmotel (approved):

overlookmotel:
Thanks Don!

PR #4355, opened 2024-07-19
chore: git ignore `.DS_Store` files

overlookmotel:
Git ignore Mac OS metadata `.DS_Store` files.


Boshen:
Please add this to your global git ignore. A local git ignore is project only.

overlookmotel:
Ah thanks! I wasn't aware there was a global git ignore. Sorted.

PR #4356, opened 2024-07-19
perf(traverse): speed up finding UID binding name

overlookmotel:
Speed up finding UID var name. Previously iterated through all `SymbolTable::references` and checked each for a string match. Replace that with a single hashmap lookup on `ScopeTree::root_unresolved_references`.

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -447,21 +447,8 @@ impl TraverseScoping {
     }
 
     fn name_is_unique(&self, name: &str) -> bool {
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4356** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4356?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
CodSpeed is showing -1% on `transformer[pdf.mjs]` benchmark. I believe this to be erroneous - just noise. The flame graph shows the part which has got slower is semantic, which is unaffected by this change.

PR #4309, opened 2024-07-17
feat(ast): add `#[ast]` attribute to non-visited AST types.

rzvxa:
Mark everything mentioned in https://github.com/oxc-project/oxc/pull/3815#issuecomment-2186736258 as AST.

We now error on the occurrence of non-ast items in the source of truth. It doesn't make sure that all fields and variants are `#[ast]` and therefore `repr_stable` but there are only a handful of non-AST types used here(mainly Atom and Span). Since we don't have access to the external types we can't make sure of it unless we find a way to const assert it.

The best we can do until then is to check all field/variant types to be either `#[ast]` or in a white list. I can add this check to the codegen in an upcoming PR.

PR #4371, opened 2024-07-19
refactor(ast_codegen): better visit marker parsing.

### LINKED ISSUE #4281: Reduce number of attrs for `#[visited_node]` used by AST codegen
Currently:

```rs
#[visited_node]
pub enum Expression<'a> {
    #[visit_args(flags = ScopeFlags::Function)]
    FunctionExpression(Box<'a, Function<'a>>),
    /* ... */
}

#[visited_node]
pub struct TryStatement<'a> {
    /* ... */
    #[visit_as(FinallyClause)]
    pub finalizer: Option<Box<'a, BlockStatement<'a>>>,
}
```

Could we reduce the number of "special" attrs by changing `visit_args(...)` / `visit_as(...)` to `visit(args(...))`  / `visit(as(...))`?

```rs
#[visited_node]
pub enum Expression<'a> {
    #[visit(args(flags = ScopeFlags::Function))]
    FunctionExpression(Box<'a, Function<'a>>),
    /* ... */
}

#[visited_node]
pub struct TryStatement<'a> {
    /* ... */
    #[visit(as(FinallyClause))]
    pub finalizer: Option<Box<'a, BlockStatement<'a>>>,
}
```

@rzvxa What do you think? Not a big deal, but I think it'd be neater. I suspect we may end up adding more attrs later on (it may be useful, for instance, to tag all TS-only fields `#[typescript]`) and it could get out of hand.

rzvxa:
closes #4281

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1,6 +1,6 @@
-// NB: `#[span]`, `#[scope(...)]`, `#[visit(...)]`, `#[visit_as(...)]` and `#[visit_args(...)]` do
-// not do anything to the code, They are purely markers for codegen used in
-// `tasts/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in that crate.
+// NB: `#[span]`, `#[scope(...)]` and `#[visit(...)]` do NOT do anything to the code.
+// They are purely markers for codegen used in
+// `tasks/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in those crates.
 
 // Silence erroneous warnings from Rust Analyser for `#[derive(Tsify)]`
 #![allow(non_snake_case)]
@@ -79,7 +79,7 @@ pub enum Expression<'a> {
     ChainExpression(Box<'a, ChainExpression<'a>>) = 16,
     ClassExpression(Box<'a, Class<'a>>) = 17,
     ConditionalExpression(Box<'a, ConditionalExpression<'a>>) = 18,
-    #[visit_args(flags = ScopeFlags::Function)]
+    #[visit(args(flags = ScopeFlags::Function))]
     FunctionExpression(Box<'a, Function<'a>>) = 19,
     ImportExpression(Box<'a, ImportExpression<'a>>) = 20,
     LogicalExpression(Box<'a, LogicalExpression<'a>>) = 21,
@@ -252,7 +252,7 @@ pub enum ArrayExpressionElement<'a> {
     /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#arrays>
     Elision(Elision) = 65,
     // `Expression` variants added here by `inherit_variants!` macro
-    // TODO: support for attributes syntax here so we can use `#[visit_as(ExpressionArrayElement)]`
+    // TODO: support for attributes syntax here so we can use `#[visit(as(ExpressionArrayElement))]`
     @inherit Expression
 }
 }
@@ -966,7 +966,7 @@ pub struct BlockStatement<'a> {
 #[serde(untagged)]
 pub enum Declaration<'a> {
     VariableDeclaration(Box<'a, VariableDeclaration<'a>>) = 32,
-    #[visit_args(flags = ScopeFlags::Function)]
+    #[visit(args(flags = ScopeFlags::Function))]
     FunctionDeclaration(Box<'a, Function<'a>>) = 33,
     ClassDeclaration(Box<'a, Class<'a>>) = 34,
     UsingDeclaration(Box<'a, UsingDeclaration<'a>>) = 35,
@@ -1292,7 +1292,7 @@ pub struct TryStatement<'a> {
     pub span: Span,
     pub block: Box<'a, BlockStatement<'a>>,
     pub handler: Option<Box<'a, CatchClause<'a>>>,
-    #[visit_as(FinallyClause)]
+    #[visit(as(FinallyClause))]
     pub finalizer: Option<Box<'a, BlockStatement<'a>>>,
 }
 
@@ -1428,7 +1428,7 @@ pub struct BindingRestElement<'a> {
 /// Function Definitions
 #[ast(visit)]
 #[scope(
-    // `flags` passed in to visitor via parameter defined by `#[visit_args(flags = ...)]` on parents
+    // `flags` passed in to visitor via parameter defined by `#[visit(args(flags = ...))]` on parents
     flags(flags),
     strict_if(self.is_strict()),
 )]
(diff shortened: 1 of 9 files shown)


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4371?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4371** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4371?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4309** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4289** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1597,7 +1597,7 @@ pub struct Class<'a> {
     pub id: Option<BindingIdentifier<'a>>,
     #[scope(enter_before)]
     pub type_parameters: Option<Box<'a, TSTypeParameterDeclaration<'a>>>,
-    #[visit_as(ClassHeritage)]
+    #[visit(alias(ClassHeritage))]
```
rzvxa, line 1600: Using `alias` instead of `as` because `as` introduces an edge case since it is a keyword and not a `Path` item.

We are already dealing with such edge cases for `#[scope(if(...))]` but that one feels natural to the user, On the other hand `as(...)` doesn't.

overlookmotel: How much work would it be to enable `visit(as(...))` (rather than `alias`)?

If it's not too much work, I think `as` would be preferable. I know `as` has a specific meaning in Rust, but the meaning of the attr in English really is "visit this field *as* as a `ClassHeritage`", so "as" is I think the best word to use. "Alias" is a bit more mysterious.

As with `visit(args(...))`, I hope we can eventually standardize and remove this special attr by adding an actual `ClassHeritage` AST node type (or by not having a specific visitor for it, if there's some other way to fulfil the use case). But who knows when/if we'll get to that, so I think it'd be better to leave the code now in best state we can, rather than assuming it doesn't much matter because we'll be removing it anyway.

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1,6 +1,6 @@
-// NB: `#[span]`, `#[scope(...)]`, `#[visit(...)]`, `#[visit_as(...)]` and `#[visit_args(...)]` do
-// not do anything to the code, They are purely markers for codegen used in
-// `tasts/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in that crate.
+// NB: `#[span]`, `#[scope(...)]` and `#[visit(...)]` do NOT do anything to the code,
+// They are purely markers for codegen used in
+// `tasts/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in those crates.
```
overlookmotel, line 3: Small nits:

1. Typo: `tasks` not `tasts`.
2. Can we add `#[ast]` to the list of attrs which don't do anything in this comment?
3. Full stop not comma after "do NOT do anything to the code" (that really is a micro-nit!)

crates/oxc_ast/src/ast/js.rs
rzvxa: > How much work would it be to enable visit(as(...)) (rather than alias)?

It is quite trivial, We try to parse it as the keyword and if it fails we proceed to other options. It would be quite similar to this(but a bit more complicated since it is nested):

https://github.com/oxc-project/oxc/blob/759180e95c1c23e432f3f41bf8c6a8245080c335/tasks/ast_codegen/src/markers.rs#L76-L82

> As with visit(args(...)), I hope we can eventually standardize and remove this special attr by adding an actual ClassHeritage AST node type (or by not having a specific visitor for it, if there's some other way to fulfil the use case). But who knows when/if we'll get to that, so I think it'd be better to leave the code now in best state we can, rather than assuming it doesn't much matter because we'll be removing it anyway.

I totally agree, I've introduced this just to avoid having edge cases/hacky workarounds in the generator.

------------

I'll request a review when I'm done refactoring it.

overlookmotel (changes requested):

crates/oxc_ast/src/ast/js.rs
overlookmotel: Thanks Rez. I've reviewed the rest of PR, and no further comments other than this one (and my nits below).

crates/oxc_ast/src/ast/js.rs
rzvxa: > 2. Can we add #[ast] to the list of attrs which don't do anything in this comment?

We can but it would be removed in the next PR, We actually do something now in our `ast` attribute.

I'll fix the other 2, Let me know what are your thoughts on this one.

overlookmotel: Ok cool, feel free to leave out that one if you prefer. We'll need to put a longer comment on the files once we do the whole `#[repr(C)]` thing, to explain what `#[ast]` does.

overlookmotel:
@rzvxa I'm sorry - I've caused merge conflicts with #4375.

rzvxa:
> @rzvxa I'm sorry - I've caused merge conflicts with #4375.

No worries, I'll merge after #4376 gets in.

crates/oxc_ast/src/ast/js.rs
```diff
@@ -1,6 +1,6 @@
-// NB: `#[span]`, `#[scope(...)]`, `#[visit(...)]`, `#[visit_as(...)]` and `#[visit_args(...)]` do
-// not do anything to the code, They are purely markers for codegen used in
-// `tasts/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in that crate.
+// NB: `#[span]`, `#[scope(...)]` and `#[visit(...)]` do NOT do anything to the code.
+// They are purely markers for codegen used in
+// `tests/ast_codegen` and `crates/oxc_traverse/scripts`. See docs in those crates.
```
overlookmotel, line 3: Oh dear! Shouldn't it be `tasks` not `tests`?

rzvxa: Oh you are right, I don't know why have difficulties spelling this path😆 I'll take care of it.

overlookmotel (approved):

overlookmotel:
Thanks very much for doing this, and sorry for tortuous review process.

PR #4367, opened 2024-07-19
perf(semantic): calculate number of nodes, scopes, symbols, references before visiting AST

Dunqing:
context: #4328

PR #4372, opened 2024-07-20
feat(semantic): add `Symbol`

DonIsaac:
# What This PR Does

Adds `Symbol`, which is a thin abstraction over `Semantic` for interacting with a specific symbol. It basically just wraps a `SymbolId`, making it a little nicer to interact with the symbol table.

### DIFF
--- a/crates/oxc_semantic/src/lib.rs
+++ b/crates/oxc_semantic/src/lib.rs
@@ -32,7 +32,7 @@ use rustc_hash::FxHashSet;
 pub use crate::{
     reference::{Reference, ReferenceFlag, ReferenceId},
     scope::ScopeTree,
-    symbol::SymbolTable,
+    symbol::{Symbol, SymbolTable},
 };
 
 pub struct Semantic<'a> {
@@ -60,6 +60,10 @@ pub struct Semantic<'a> {
 }
 
 impl<'a> Semantic<'a> {
+    pub fn get_symbol(&self, symbol_id: SymbolId) -> Symbol<'_, 'a> {
+        Symbol::new(self, symbol_id)
+    }
+
     pub fn into_symbol_table_and_scope_tree(self) -> (SymbolTable, ScopeTree) {
         (self.symbols, self.scopes)
     }
--- a/crates/oxc_semantic/src/symbol.rs
+++ b/crates/oxc_semantic/src/symbol.rs
@@ -1,5 +1,7 @@
 #![allow(non_snake_case)] // Silence erroneous warnings from Rust Analyser for `#[derive(Tsify)]`
 
+use std::{fmt, hash};
+
 use oxc_ast::ast::Expression;
 use oxc_index::IndexVec;
 use oxc_span::{CompactStr, Span};
@@ -13,7 +15,7 @@ use serde::Serialize;
 use tsify::Tsify;
 
 use crate::{
-    node::AstNodeId,
+    node::{AstNode, AstNodeId},
     reference::{Reference, ReferenceId},
 };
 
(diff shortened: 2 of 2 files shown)


Boshen:
This indirection feels redundant to me - you'll need to get familiarized yourself of data-orientated-design and our SoA strucutre.

overlookmotel:
Personally, I do like the thrust of this PR - I think the storage medium (SoA) can be thought of an implementation detail, and it's appropriate to provide an abstraction with a more "normal" API.

However:

1. I'm not sure if this specific implementation of that abstraction is ideal. It may be that `Symbol` borrowing `&'s Semantic<'a>` is a perf regression, and also may get us in trouble with borrow-checker when it comes to mutating symbols.
2. As I think has been mentioned elsewhere, I'm going to be doing an "in the round" review of `Semantic` over the next month, and producing an RFC. My feeling is that at this point, we'd be better off deciding an overall end goal for `Semantic` and plotting a roadmap for to get there, rather than continuing with incremental peacemeal changes. So for now, it be my preference if we could call a halt to modifying `Semantic`'s APIs, as it may be wasted effort if my RFC proposes ripping it all up and starting again!

FYI, I've also stopped doing perf work on `Semantic` now. There was some low-hanging fruit which seemed worthwhile to pluck asap (e.g. #4367). But in my view, we've now got most of the perf gains that we can get with ease, and at this point we'd be better off putting our efforts into other areas until RFC is ready. Hope this makes sense.

rzvxa:
Not directly related to this PR, But maybe we should put some effort toward researching any existing solution like `MultiArrayList` in zig or implement one ourselves. It is a really handy data structure that makes any struct into an SoA without drastic changes to the API for the users.

https://github.com/ziglang/zig/blob/master/lib/std/multi_array_list.zig

(have I mentioned this before or am I having a Dejavu moment?)

DonIsaac:
I agree that the various SoA structures should be an implementation detail. I've found that interacting with them can be quite verbose (particularly in lint rules), and this seems like a cleaner way of doing so.

> 1. I'm not sure if the specific implementation of that abstraction in this PR is ideal. It may be that `Symbol` borrowing `&'s Semantic<'a>` is a perf regression, and also may get us in trouble with borrow-checker when it comes to mutating symbols.

This _could_ be a perf regression, but if it was, it would only affect consumers using an owned `Semantic`. In most cases, we pass around an `Rc<Semantic>`, which requires a single pointer dereference, the same way this does. Additionally, passing around a `Symbol` does not incur reference counting overhead. And, since all methods are inlined, we shouldn't incur any overhead for function calls. In most cases (e.g. lint rules), the generated assembly _should_ be almost exactly the same. The one potential caveat is the inclusion of `SymbolFlags` in `Symbol` - use cases that don't use them may see a slight regression. Because of this, I've marked `Symbol` as `#[non_exhaustive]` so we can remove it later without breaking its API.

Regarding updates to the symbol table, `Symbol` would not be able to do these.
1. In the case of `SemanticBuilder`, internal `_mut()` methods should continue to be used
2. In the case of  the Transformer, Minifier, and ID, most changes are on the AST, which is out of scope of `Symbol`
3. Mangling will not break `Symbol` as it only changes identifier names, meaning `Symbol::name()` would continue to return the correct identifier name. `SymbolId`s do not change.

> 2. As I think has been mentioned elsewhere, I'm going to be doing an "in the round" review of `Semantic` over the next month, and producing an RFC. My feeling is that at this point, we'd be better off deciding an overall end goal for `Semantic` and plotting a roadmap for how to get there, rather than continuing with incremental peacemeal changes. So for now, it be my preference if we could call a halt to modifying `Semantic`'s APIs, as it may be wasted effort if my RFC proposes ripping it all up and starting again!

In my opinion, having a wrapper gives us the flexibility to refactor `Semantic` while reducing the effects on downstream code. Right now, APIs for `Semantic`, `SymbolTable`, etc are directly tied to the internal SoA format. We can't change internal formats without breaking our contract. With a wrapper, we could refactor how internal structures are organized, then just change the implementation of `Symbol`'s methods.

overlookmotel:
I'm sorry @DonIsaac but I don't think we should move forwards with this right now. Boshen has voiced opposition (and I don't want to pull him away from his holiday to argue the case with him). And I've said that, while I do like the thrust of what you're doing here, I'm also opposed to making changes to `Semantic`'s APIs right now.

I've added a note on https://github.com/oxc-project/backlog/issues/11 to come back to this later on, so your work won't be wasted.

So I'm not saying "no", but I am saying "not now". I hope you can understand my reasoning.

DonIsaac:
Understandable, let's move forward on Semantic first. I'll close this PR.

overlookmotel:
Thanks for understanding Don. It's never easy saying "no", and rarely feels great receiving that answer either. So I appreciate you accepting my rationale, even though it's no doubt not what you wanted to hear.

PR #4373, opened 2024-07-20
feat(ast, ast_macros): apply stable repr to all `#[ast]` enums

### LINKED ISSUE #4296: Make AST `#[repr(C)]`
I am in favour of making all AST types `#[repr(C)]`.

The effect of `#[repr(C)]` is to make the memory layout of types predictable.

### 1: Treat Rust types as a serialization format

(copied from https://github.com/oxc-project/oxc/pull/3115)

Oxc's direction appears to be increasingly to build up control over the fundamental primitives we use, in order to unlock performance and features. We have our own allocator, our own custom implementations for `Box` and `Vec`, our own `IndexVec`. The AST is the central building block of Oxc, and taking control of its memory layout feels like a step in this same direction.

Oxc has a major advantage over other similar libraries in that it keeps all the AST data in an arena. This opens the door to treating the AST either as Rust types or as pure data (just bytes). That data can be moved around and manipulated beyond what Rust natively allows.

However, to enable that, the types need to be well-specified, with completely stable layouts. `#[repr(C)]` is the only tool Rust provides to do this.

Once the types are `#[repr(C)]`, various features become possible:

* Cheap transfer of the AST across boundaries without ser/deser - the property used by AST Transfer.
* Having multiple versions of the AST (standard, read-only, traversable), and these AST representations can be converted to one other at zero cost via transmute - the property used by Traversable AST scheme (Traversable AST is now abandoned, but we could use the same ability in other ways).
* Caching AST data on disk (https://github.com/oxc-project/oxc/discussions/3079) or transferring across network.
* Stuff we haven't thought of yet!

Allowing the AST to be treated as pure data will likely unlock other "next level" features further down the track (caching for "edge bundling" comes to mind).

### 2: AST transfer

This isn't *necessary* for AST transfer, but it's desirable, as then we can get rid of a bunch of code in current implementation which has to do introspection of the type layouts at runtime. If types are `#[repr(C)]`, it can be done in build script instead.

### 3: Memory access optimizations

Control of in-memory type layouts can also allow some runtime optimizations.

For example, if all `span` field was always placed at the same offset in every struct, then `GetSpan::span` could be reduced to a single CPU instruction with no branching.

## The problem with `#[repr(C)]`

`#[repr(C)]` does have the annoying requirement that you have to order the fields of structs in the exact order that they'll be laid out in memory. Without `#[repr(C)]`, Rust automatically re-orders the fields to minimize padding (broadly speaking, order fields in descending order of type alignment). But with `#[repr(C)]` you have to do this manually.

e.g. `BinaryExpression` needs to be defined as:

```rs
#[repr(C)]
pub struct BinaryExpression<'a> {
    pub left: Expression<'a>,
    pub right: Expression<'a>,
    pub span: Span,
    pub operator: BinaryOperator,
}
```

This is not ideal. We want field order as written in `*.rs` files to be:

1. The order of visitation.
2. The order that makes sense semantically - memory layout should be an invisible implementation detail.

## Proposed solutions

```rs
#[repr_stable(field_order(left, right, span, operator))]
pub struct BinaryExpression<'a> {
    pub span: Span,
    pub left: Expression<'a>,
    pub operator: BinaryOperator,
    pub right: Expression<'a>,
}
```

`repr_stable` is a proc macro which re-orders the fields in the memory layout. But the field order *as written* can be the same as it is now - a semantically sensible order.

Alternatively: Rather than specifying memory field order in the `#[repr_stable]` attribute, codegen (https://github.com/oxc-project/oxc/pull/3815) can figure out what memory layout should be and pass this info to the macro via "side channel" of the AST schema. That way, all types will be optimized for ideal memory layout without any manual work.

That also opens the door to doing most of the work in the codegen, and making the proc macro "dumb", so minimizing its compile-time impact. Pushing this to its furthest extent, something like "Option 2d: Cache macro expansion output" in oxc-project/oxc#4297 would make the macro basically free (almost 0 compile time cost).

rzvxa:
closes #4296

### DIFF
--- a/crates/oxc_ast_macros/src/lib.rs
+++ b/crates/oxc_ast_macros/src/lib.rs
@@ -1,30 +1,54 @@
 use proc_macro::TokenStream;
-use std::str::FromStr;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
 
-/// Attach to AST node type (struct or enum), to signal to codegen to create visitor for this type.
+/// returns `#[repr(C, u8)]` if `enum_` has any non-unit variant,
+/// Otherwise it would return `#[repr(u8)]`.
+fn enum_repr(enum_: &syn::ItemEnum) -> TokenStream2 {
+    if enum_.variants.iter().any(|var| !matches!(var.fields, syn::Fields::Unit)) {
+        quote!(#[repr(C, u8)])
+    } else {
+        quote!(#[repr(u8)])
+    }
+}
+
+/// This attribute serves two purposes,
+/// First, it is a marker for our codegen to detect AST types. Furthermore.
+/// It is also a lightweight macro; All of its computation is cached and
+/// it only applies the following changes without any complex operation:
 ///
-/// Macro's role is not to generate code - it's purely a means to communicate information to the codegen.
+/// * Prepend `#[repr(C, u8)]` to fieldful enums e.g. `enum E { X: u32, Y: u8 }`
+/// * Prepend `#[repr(u8)]` to unit (fieldless) enums e.g. `enum E { X, Y, Z, }`
+/// * Prepend `#[derive(oxc_ast_macros::Ast)]`
 ///
-/// Only thing macro does is add `#[derive(Ast)]` to the item.
-/// Deriving `Ast` does nothing, but supports `#[scope]`, `#[visit]`, and other attrs on struct fields.
-/// These "helper" attributes are also signals to the codegen, and do nothing in themselves.
-///
-/// This is a workaround for Rust not supporting helper attributes for `proc_macro_attribute` macros,
-/// so we need to use a derive macro to get that support.
-///
-/// Use native Rust `TokenStream`, to avoid dependency on slow-compiling crates like `syn` and `quote`.
 #[proc_macro_attribute]
 #[allow(clippy::missing_panics_doc)]
 pub fn ast(_args: TokenStream, input: TokenStream) -> TokenStream {
-    let mut stream = TokenStream::from_str("#[derive(::oxc_ast_macros::Ast)]").unwrap();
-    stream.extend(input);
-    stream
+    let input = syn::parse_macro_input!(input as syn::Item);
+
+    let repr = match input {
+        syn::Item::Enum(ref enum_) => enum_repr(enum_),
+        // In future, we'll add `#[repr(C)]` to structs, but at present this is disabled
+        syn::Item::Struct(_) => TokenStream2::default(),
+
+        _ => {
+            unreachable!()
+        }
+    };
+
+    let expanded = quote! {
+        #[derive(::oxc_ast_macros::Ast)]
+        #repr
+        #input
+    };
+    TokenStream::from(expanded)
 }
 
 /// Dummy derive macro for a non-existent trait `Ast`.
 ///
 /// Does not generate any code.
 /// Only purpose is to allow using `#[scope]`, `#[visit]`, and other attrs in the AST node type defs.
+/// These "marker" attributes are used in codegen.
 #[proc_macro_derive(Ast, attributes(scope, visit, span, serde, tsify))]
 pub fn ast_derive(_item: TokenStream) -> TokenStream {
     TokenStream::new()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1355,6 +1355,11 @@ dependencies = [
 [[package]]
 name = "oxc_ast_macros"
 version = "0.23.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
 
 [[package]]
 name = "oxc_benchmark"
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -52,7 +52,6 @@ inherit_variants! {
 ///
 /// [`ast` module docs]: `super`
 #[ast(visit)]
-#[repr(C, u8)]
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[serde(untagged)]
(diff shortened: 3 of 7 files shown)


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4373?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4373** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4373?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4371** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4309** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4289** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -16,9 +25,23 @@ use std::str::FromStr;
 #[proc_macro_attribute]
 #[allow(clippy::missing_panics_doc)]
 pub fn ast(_args: TokenStream, input: TokenStream) -> TokenStream {
```
overlookmotel, line 27: Doc comment above this function needs updating to reflect that this is a real macro now and adds `#[repr]` attribute to types.

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -1,5 +1,14 @@
 use proc_macro::TokenStream;
-use std::str::FromStr;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+
+fn enum_repr(enum_: &syn::ItemEnum) -> TokenStream2 {
```
overlookmotel, line 5: Please add doc comment on this function clarifying when it's `#[repr(C)]` and when `#[repr(C, u8)]`.

overlookmotel (changes requested):

overlookmotel:
Aside from comments above, this looks good.

But... I suggest we hold off merging until rest of the stack is complete. With just this PR, the types' memory layouts are not optimal, and you can see it a bit on benchmarks. Let's wait until macro+codegen rearranges the fields so they end up in same order as they were before `#[repr(C)]` was applied.

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -1,30 +1,54 @@
...
 
-/// Attach to AST node type (struct or enum), to signal to codegen to create visitor for this type.
-///
-/// Macro's role is not to generate code - it's purely a means to communicate information to the codegen.
-///
-/// Only thing macro does is add `#[derive(Ast)]` to the item.
-/// Deriving `Ast` does nothing, but supports `#[scope]`, `#[visit]`, and other attrs on struct fields.
-/// These "helper" attributes are also signals to the codegen, and do nothing in themselves.
+/// returns `#[repr(C, u8)]` if `enum_` has any non-unit variant,
+/// Otherwise it would return `#[repr(C)]`.
```
overlookmotel, line 6:
```suggestion
/// Otherwise it would return `#[repr(u8)]`.
```

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -1,30 +1,54 @@
...
+fn enum_repr(enum_: &syn::ItemEnum) -> TokenStream2 {
+    if enum_.variants.iter().any(|var| !matches!(var.fields, syn::Fields::Unit)) {
+        quote!(#[repr(C, u8)])
+    } else {
+        quote!(#[repr(C)])
```
overlookmotel, line 11: `#[repr(u8)]` is what we want for fieldless enums.

```suggestion
        quote!(#[repr(u8)])
```

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -1,30 +1,54 @@
...
 #[allow(clippy::missing_panics_doc)]
 pub fn ast(_args: TokenStream, input: TokenStream) -> TokenStream {
-    let mut stream = TokenStream::from_str("#[derive(::oxc_ast_macros::Ast)]").unwrap();
-    stream.extend(input);
-    stream
+    let input = syn::parse_macro_input!(input as syn::Item);
+
+    let repr = match input {
+        syn::Item::Enum(ref enum_) => enum_repr(enum_),
+        syn::Item::Struct(_) => TokenStream2::default(),
```
overlookmotel, line 32:
```suggestion
        // In future, we'll add `#[repr(C)]` to structs, but at present this is disabled
        syn::Item::Struct(_) => TokenStream2::default(),
```

crates/oxc_ast_macros/src/lib.rs
```diff
@@ -1,30 +1,54 @@
...
+/// This attribute serves two purposes,
+/// First, it is a marker for our codegen to detect AST types. Furthermore.
+/// It is also a lightweight macro; All of its computation is cached and
+/// it only applies the following changes without any complex operation:
 ///
-/// This is a workaround for Rust not supporting helper attributes for `proc_macro_attribute` macros,
-/// so we need to use a derive macro to get that support.
+/// * Prepend `#[repr(C)]` to structs
+/// * Prepend `#[repr(C, u8)]` to enums
+/// * Prepend `#[repr(C)]` to unit enums
```
overlookmotel, line 22:
```suggestion
/// * Prepend `#[repr(C, u8)]` to fieldful enums e.g. `enum E { X: u32, Y: u8 }`
/// * Prepend `#[repr(u8)]` to unit (fieldless) enums e.g. `enum E { X, Y, Z, }`
```

overlookmotel:
@rzvxa Aside from comments above, and resolving conflicts, I think this PR + 3 more above it in this stack are ready to merge.

overlookmotel (approved):

PR #4374, opened 2024-07-20
fix(codegen): print shorthand for all `{ x }` variants

### LINKED ISSUE #4340: codegen: as possible as should be printing shorthand in `ObjectProperty`
See: https://github.com/rolldown/rolldown/pull/1656#discussion_r1682067931
Caused by https://github.com/oxc-project/oxc/pull/4319

### In
```js
const { foo } = obj;
```

### Out
```js
const { foo: foo } = obj;
```

### Expected
```js
const { foo } = obj;
```

Boshen:
closes #4340


crates/oxc_codegen/src/gen.rs
```diff
@@ -1079,18 +1079,10 @@ impl<'a, const MINIFY: bool> GenExpr<MINIFY> for ParenthesizedExpression<'a> {
...
-                    p.add_source_mapping_for_name(self.span, &name);
-                    p.print_str(&name);
-                    return;
-                }
-            }
-        }
-        p.add_source_mapping_for_name(self.span, &self.name);
-        p.print_str(&self.name);
+        let name = p.get_identifier_reference_name(self);
+        let name = CompactStr::new(name);
```
Boshen, line 1083: This is not ideal, let me try again without it. borrower checker was getting in the way.

PR #2669, opened 2024-03-10
refactor: reduce `cfg_attr` boilerplate with `SerAttrs` derive

overlookmotel:
Closes #2641.

Also added `tsify` attribute to the `SerAttrs` derive macro, so `#[cfg_attr(feature = "wasm", tsify(...))]` can also be reduced to `#[tsify(...)]`.

PR #4375, opened 2024-07-20
refactor(ast): reduce `#[cfg_attr]` boilerplate in AST type defs

overlookmotel:
Remove a ton of `#[cfg_attr(feature = "serialize", serde(...))]` boilerplate from AST type definitions.

Before: `#[cfg_attr(feature = "serialize", serde(flatten))]`
After: `#[serde(flatten)]`

This is a reprise of #2669, which was later reverted, but this time doing it using our existing zero-cost `#[ast]` dummy macro attr, so no compile time penalty this time around.

This makes no difference to either runtime or compile time behavior, purely removes the `cfg_attr` boilerplate and makes the code easier to read.

### DIFF
--- a/crates/oxc_ast/src/ast/jsx.rs
+++ b/crates/oxc_ast/src/ast/jsx.rs
@@ -23,9 +23,9 @@ use super::{inherit_variants, js::*, literal::*, ts::*};
 #[ast(visit)]
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
-#[cfg_attr(feature = "serialize", serde(tag = "type", rename_all = "camelCase"))]
+#[serde(tag = "type", rename_all = "camelCase")]
 pub struct JSXElement<'a> {
-    #[cfg_attr(feature = "serialize", serde(flatten))]
+    #[serde(flatten)]
     pub span: Span,
     pub opening_element: Box<'a, JSXOpeningElement<'a>>,
     pub closing_element: Option<Box<'a, JSXClosingElement<'a>>>,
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#4375** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4375?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_ast/src/ast/jsx.rs
```diff
@@ -23,9 +23,9 @@ use super::{inherit_variants, js::*, literal::*, ts::*};
 #[ast(visit)]
 #[derive(Debug, Hash)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
```
rzvxa, line 25: Can't we also put these in `#[ast]`? Similar to how we are adding ast derive.

rzvxa: To clarify I'm talking about this line:

```rust
#[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
```

overlookmotel: You mean add `#[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]` in `#[ast]` macro so we can remove it from type defs?

If that's what you mean, I think this would be a more "controversial" change, because it hides away things from the reader ("How does this type get serialized? It doesn't implement `Serialize`!"). In comparison, the changes in this PR don't really do that - they just strip out some largely meaningless syntax noise. Pretty sure Boshen would be against going so far as hiding `#[derive]`s.

rzvxa (approved):

PR #4377, opened 2024-07-20
perf: replace some CompactStr usages with Cows

DonIsaac:
Reduce memory allocations in semantic and linter by using `Cow<'a, str>` over `CompactStr`


overlookmotel (approved):
Can't see significant speed-up on benchmarks, but this does seem a good change in principle, so merging.

@DonIsaac Probably the reason it's not having a more dramatic impact on perf is that `CompactStr` has an inline optimization for strings of 24 bytes or less. So it only allocates to heap if strings are longer than that, which is relatively rare.

We should probably also try [beef](https://crates.io/crates/beef) at some point to replace `Cow` - it takes less memory and claims to be faster.

PR #4379, opened 2024-07-20
refactor(semantic): var hoisting

### LINKED ISSUE #4323: semantic: var hoisting
Several attempts was made to pass all tests in test262 + babel + typescript regarding var hoisting in strict / non-strict mode.

Scope tree ended up with a `var` binding everywhere:

```
{
  {
    var x
  } 
}

x = 1
```

produces

```
Scope1 (ScopeFlags(StrictMode | Top)) {
  Bindings: {
    x (SymbolFlags(FunctionScopedVariable))
  }
  Scope2 (ScopeFlags(StrictMode)) {
    Bindings: {
      x (SymbolFlags(FunctionScopedVariable))
    }
    Scope3 (ScopeFlags(StrictMode)) {
      Bindings: {
        x (SymbolFlags(FunctionScopedVariable))
      }
    }
  }
}
```

The consequence of this is:

* unintuitive
* performance killer
* breaks mangler https://github.com/oxc-project/oxc/blob/3df9e697ccf1944359edf3b94fe569eb5a00b364/crates/oxc_mangler/src/lib.rs#L111-L123

To understand the test cases, try different strategies in https://github.com/oxc-project/oxc/blob/3df9e697ccf1944359edf3b94fe569eb5a00b364/crates/oxc_semantic/src/binder.rs#L46-L54

and run `just c` to see changes in snapshot.

Dunqing:
close: #4323

This PR refactors the var hoisting logic to avoid inserting the same symbol into every scope that can be hosted.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -15,6 +15,7 @@ use oxc_cfg::{
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_span::{Atom, CompactStr, SourceType, Span};
 use oxc_syntax::{module_record::ModuleRecord, operator::AssignmentOperator};
+use rustc_hash::FxHashMap;
 
 use crate::{
     binder::Binder,
@@ -27,7 +28,7 @@ use crate::{
     module_record::ModuleRecordBuilder,
     node::{AstNodeId, AstNodes, NodeFlags},
     reference::{Reference, ReferenceFlag, ReferenceId},
-    scope::{ScopeFlags, ScopeId, ScopeTree},
+    scope::{Bindings, ScopeFlags, ScopeId, ScopeTree},
     symbol::{SymbolFlags, SymbolId, SymbolTable},
     unresolved_stack::UnresolvedReferencesStack,
     JSDocFinder, Semantic,
@@ -66,6 +67,7 @@ pub struct SemanticBuilder<'a> {
     // to value like
     pub namespace_stack: Vec<SymbolId>,
     current_reference_flag: ReferenceFlag,
+    pub hoisting_variables: FxHashMap<ScopeId, FxHashMap<Atom<'a>, SymbolId>>,
 
     // builders
     pub nodes: AstNodes<'a>,
@@ -114,6 +116,7 @@ impl<'a> SemanticBuilder<'a> {
             function_stack: vec![],
             namespace_stack: vec![],
             nodes: AstNodes::default(),
+            hoisting_variables: FxHashMap::default(),
             scope,
             symbols: SymbolTable::default(),
             unresolved_references: UnresolvedReferencesStack::new(),
@@ -357,14 +360,16 @@ impl<'a> SemanticBuilder<'a> {
     }
 
     pub fn check_redeclaration(
-        &mut self,
+        &self,
         scope_id: ScopeId,
         span: Span,
         name: &str,
         excludes: SymbolFlags,
         report_error: bool,
     ) -> Option<SymbolId> {
-        let symbol_id = self.scope.get_binding(scope_id, name)?;
+        let symbol_id = self.scope.get_binding(scope_id, name).or_else(|| {
+            self.hoisting_variables.get(&scope_id).and_then(|symbols| symbols.get(name).copied())
+        })?;
         if report_error && self.symbols.get_flag(symbol_id).intersects(excludes) {
             let symbol_span = self.symbols.get_span(symbol_id);
             self.error(redeclaration(name, symbol_span, span));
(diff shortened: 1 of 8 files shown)


Dunqing:
* **#4379** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/coverage/parser_typescript.snap
```diff
@@ -4639,6 +4639,19 @@ Expect to Parse: "compiler/withStatementInternalComments.ts"
...
+     ╭─[conformance/async/es6/asyncWithVarShadowing_es6.ts:130:14]
+ 129 │     }
+ 130 │     catch ({ x }) {
+     ·              ┬
+     ·              ╰── `x` has already been declared here
+ 131 │         var x;
+     ·             ┬
+     ·             ╰── It can not be redeclared here
+ 132 │     }
+     ╰────
```
Dunqing, line 4654: It is correct

https://esbuild.github.io/try/#dAAwLjIzLjAALS1taW5pZnkAYXN5bmMgZnVuY3Rpb24gZm4yNih4KSB7CiAgICB0cnkgewoKICAgIH0KICAgIGNhdGNoICh7IHggfSkgewogICAgICAgIHZhciB4OwogICAgfQp9

Boshen: I remember this one! I was so confused :-/

Boshen:
https://github.com/oxc-project/monitor-oxc/actions/workflows/ci.yml can check PR now.

crates/oxc_semantic/src/builder.rs
```diff
@@ -67,6 +68,7 @@ pub struct SemanticBuilder<'a> {
     // to value like
     pub namespace_stack: Vec<SymbolId>,
     current_reference_flag: ReferenceFlag,
+    pub hosting_variables: FxHashMap<ScopeId, Vec<SymbolId>>,
```
Boshen, line 71: `hosting` -> `hoisting`.

Boshen: This state feels so weird. Shall we sit on this PR for at least 2 days before merging too quickly?

Dunqing: Ok

Let me explain why need it. This state replaces add binding to every hoisting scope. Since we're only inserting bindings into the top-level hoisting scopes now. This will result in us not being able to check if a variable declared after the promoted variable has been declared or not

For example:

```js
function func() {
  {
    {
      var v = 0;
    }
    let v = 0;
  }
}
```

After hoisted, will be like this:

```js
function func() {
  var v;
  {
    {
      v = 0;
    }
    let v = 0;
  }
}
```
Should throw an error “Identifier `v` has already been declared” for `let v = 0;`. If without `hoisting_varaibles` we can't do that

tasks/coverage/parser_typescript.snap
overlookmotel: What?? In NodeJS, this runs without error:

```js
try {
    throw { x: 123 };
} catch (x) {
    var x = 456;
}
```

and this throws `SyntaxError: Identifier 'x' has already been declared`:

```js
try {
    throw { x: 123 };
} catch ({ x }) {
    var x = 456;
}
```

But look what this does:

```js
try {
    throw {x: 123};
} catch (x) {
    console.log(x); // Logs {x: 123}
    var x = 456;
    console.log(x); // Logs 456
    var y = 789;
}
console.log(x); // Logs undefined (!!!)
console.log(y); // Logs 789
```

Note that the last `console.log(x)` does NOT throw `ReferenceError: x is not defined`. So `var x` has created a hoisted binding in outer scope, but it's not initialized (unlike `var y = 789;` which creates binding *and* sets its value in outer scope). Is this correct behavior??

Dunqing: I can't explain why this is the case, from the example it seems to be because `CatchClause` can't change the value of the `CatchParameter` internally

Dunqing:
> https://github.com/oxc-project/monitor-oxc/actions/workflows/ci.yml can check PR now.

I saw `mangler` job succeed, and the `transformer` job failed but it has nothing to do with the changes to that PR
https://github.com/oxc-project/monitor-oxc/actions/runs/10034040791/job/27728055130

tasks/coverage/parser_typescript.snap
Boshen: Read the spec. It tried to fix for es2015.

Boshen:
I'm going to stare at this PR while I'm on the plane to Sydney tomorrow.

Boshen (approved):
Rebase and merge~

PR #4383, opened 2024-07-21
perf(semantic): simplify logic in `enter_scope` + `leave_scope`

overlookmotel:
`SemanticBuilder::enter_scope` contained multiple checks purely to handle when called for `Program` scope (where scope has no parent). This is a very uncommon case as `Program` is visited only once, but these checks run for every scope.

So don't call `enter_scope` from `visit_program`, and inline the special logic for root scope there instead, to simplify `enter_scope`.

A branch in `leave_scope` also gets more predictable (always taken).

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -123,7 +123,7 @@ impl<'a> SemanticBuilder<'a> {
             current_symbol_flags: SymbolFlags::empty(),
             current_reference_flag: ReferenceFlag::empty(),
             current_scope_id,
-            current_scope_depth: 0,
+            current_scope_depth: 1, // Start with depth for `Program`
             function_stack: vec![],
             namespace_stack: vec![],
             nodes: AstNodes::default(),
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4383** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -460,31 +460,26 @@ impl<'a> SemanticBuilder<'a> {
...
-                // Clone the `CatchClause` bindings and add them to the current scope.
-                // to make it easier to check redeclare errors.
-                let bindings = self.scope.get_bindings(parent_scope_id).clone();
-                self.scope.get_bindings_mut(self.current_scope_id).extend(bindings);
-            }
+        if self.scope.get_flags(parent_scope_id).is_catch_clause() {
+            // Clone the `CatchClause` bindings and add them to the current scope.
+            // to make it easier to check redeclare errors.
+            let bindings = self.scope.get_bindings(parent_scope_id).clone();
+            self.scope.get_bindings_mut(self.current_scope_id).extend(bindings);
```
Dunqing, line 482: After merging this PR. I'd also like to move this logic to `BlockStatement`

overlookmotel: Good idea!

overlookmotel:
I ran the benchmarks several times and perf impact was inconclusive - too much noise in the bench results. But this *should* be a little faster.

Dunqing (approved):

PR #4390, opened 2024-07-21
perf(semantic): remove bounds checks on unresolved references stack

overlookmotel:
`SemanticBuilder` keeps a stack of hashmaps to track unresolved references. This stack only grows, never shrinks. Use that invariant to remove bounds checks.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -26,8 +26,9 @@ use crate::{
     module_record::ModuleRecordBuilder,
     node::{AstNodeId, AstNodes, NodeFlags},
     reference::{Reference, ReferenceFlag, ReferenceId},
-    scope::{ScopeFlags, ScopeId, ScopeTree, UnresolvedReferences},
+    scope::{ScopeFlags, ScopeId, ScopeTree},
     symbol::{SymbolFlags, SymbolId, SymbolTable},
+    unresolved_stack::UnresolvedReferencesStack,
     JSDocFinder, Semantic,
 };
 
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#4390** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
More than anything else, this is an experiment to see what effect on perf this kind of optimization has.

In this case, answer seems to be only a little bit, but enter/exit scope isn't a particularly hot path. Likely we could use this same technique on hotter paths for a larger boost.

crates/oxc_semantic/src/builder.rs
```diff
@@ -100,18 +95,103 @@ pub struct SemanticBuilderReturn<'a> {
     pub errors: Vec<OxcDiagnostic>,
 }
 
+// Stack used to accumulate unresolved refs while traversing scopes.
+// Indexed by scope depth. We recycle `UnresolvedReferences` instances during traversal
+// to reduce allocations, so the stack grows to maximum scope depth, but never shrinks.
+// See: <https://github.com/oxc-project/oxc/issues/4169>
+// This stack abstraction uses the invariant that stack only grows to avoid bounds checks.
+struct UnresolvedReferencesStack {
```
Boshen, line 103: Move to a new file? Seems out of place.

overlookmotel: Done. NB this is only used in `SemanticBuilder`, not `Semantic`, so I don't think it should be in `scope.rs`.

Dunqing (approved):

PR #4329, opened 2024-07-17
refactor(semantic)!: remove name from `reference`

Dunqing:
related to: https://github.com/oxc-project/backlog/issues/32

PR #4394, opened 2024-07-21
perf(semantic): use `Atom` instead of `CompactStr` for `UnresolvedReferencesStack`

Dunqing:
Just a try

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -351,13 +351,12 @@ impl<'a> SemanticBuilder<'a> {
     /// Declare an unresolved reference in the current scope.
     ///
     /// # Panics
-    pub fn declare_reference(&mut self, reference: Reference) -> ReferenceId {
-        let reference_name = reference.name().clone();
+    pub fn declare_reference(&mut self, name: &str, reference: Reference) -> ReferenceId {
         let reference_flag = *reference.flag();
         let reference_id = self.symbols.create_reference(reference);
 
         self.unresolved_references[self.current_scope_depth]
-            .entry(reference_name)
+            .entry(CompactStr::from(name))
             .or_default()
             .push((reference_id, reference_flag));
         reference_id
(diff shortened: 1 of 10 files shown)


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4394?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4394** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4329** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4329?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -351,12 +359,12 @@ impl<'a> SemanticBuilder<'a> {
     /// Declare an unresolved reference in the current scope.
     ///
     /// # Panics
-    pub fn declare_reference(&mut self, name: &str, reference: Reference) -> ReferenceId {
+    pub fn declare_reference(&mut self, name: &Atom<'a>, reference: Reference) -> ReferenceId {
```
overlookmotel, line 362: A small thing: `Atom<'a>` is just a wrapper around `&'a str`. So it's better to pass around `Atom`s, rather than `&Atom`s (`&Atom` is equivalent to `&&str` - double-reference). `atom.clone()` is very cheap (compiler treats it as if `Atom` was `Copy`).

We could make `Atom` `Copy` now, but I think we should not, to avoid breakage if we later make `Atom` an interned string.

overlookmotel:
Nice! Significant speed-up.

Dunqing:
Cannot reopen this PR, so I create another PR #4401 instead

PR #4401, opened 2024-07-22
perf(semantic): use `Atom` instead of `CompactStr` for `UnresolvedReferencesStack`

Dunqing:
related: #4394

The `UnresolvedReferencesStack` is only used for resolving references, and the `Atom` clone is cheap, So we can safely change `CompactStr`to `Atom`


Dunqing:
image

Boshen (approved):

overlookmotel (approved):

PR #4415, opened 2024-07-23
Revert "Release crates v0.22.0"

Dunqing:
Reverts oxc-project/oxc#4414

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1276,7 +1276,7 @@ checksum = "caff54706df99d2a78a5a4e3455ff45448d81ef1bb63c22cd14052ca0e993a3f"
 
 [[package]]
 name = "oxc"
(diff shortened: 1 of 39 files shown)


IWANABETHATGUY:
It looks like `cargo` doesn't allow publishing crates with cyclic dev-dependencies, https://github.com/rust-lang/cargo/issues/4242, there are some workarounds: 
https://github.com/madsmtm/objc2/commit/5a891843844c69baf4c47f710f64ce59f2813249
https://github.com/entropyxyz/entropy-core/pull/837/files

IWANABETHATGUY:
We could use path instead of `workspace`, 
https://github.com/oxc-project/oxc/blob/main/crates/oxc_mangler/Cargo.toml#L31

PR #4416, opened 2024-07-23
chore: fix failed publish caused by cyclic dev dependencies 

IWANABETHATGUY:
1. This looks like caused by https://github.com/rust-lang/cargo/issues/4242

## Ref
1. https://github.com/rust-lang/futures-rs/pull/2305

### DIFF
--- a/crates/oxc_mangler/Cargo.toml
+++ b/crates/oxc_mangler/Cargo.toml
@@ -28,7 +28,7 @@ oxc_index    = { workspace = true }
 itertools    = { workspace = true }
 
 [dev-dependencies]
-oxc_codegen   = { workspace = true }
+oxc_codegen   = { path = "../oxc_codegen" }
 oxc_parser    = { workspace = true }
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
Or we could remove `version` in https://github.com/IWANABETHATGUY/oxc/blob/89e037b231f75451e7152daf6b81e649fde2bd7c/Cargo.toml#L79-L79, instead.

Dunqing:
Do we have a way to verify this is work

IWANABETHATGUY:
I don't know, cc  @Boshen

IWANABETHATGUY:
Because the dev cyclic dependency still exists, you can't make sure if this works until you try to make a new release, AFAIK, this is probably the best workaround, since `future-rs` and many linked pull request also use this trick.

Dunqing:
Still failed https://github.com/oxc-project/oxc/actions/runs/10053226875/job/27785666460

PR #4438, opened 2024-07-23
feat(linter): add typescript-eslint/prefer-keyword-namespce

areklaw2:
Added rule for https://typescript-eslint.io/rules/prefer-namespace-keyword/


DonIsaac:
Should this rule ignore modules inside `declare global`?

```ts
declare global {
  module foo { }
}
```

aowalke2:
> Should this rule ignore modules inside `declare global`?
> 
> ```ts
> declare global {
>   module foo { }
> }
> ```

No it should not ignore modules inside declare glabal

DonIsaac:
Oops I think I merged before benchmarks completed, I'll revert if there's a regression.

PR #4440, opened 2024-07-24
feat(linter): eslint-plugin-vitest/no-import-node-test

eryue0220:
support [eslint-plugin-vitest/no-import-node-test](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/no-import-node-test.md)


mysteryven (approved):

crates/oxc_linter/src/rules/vitest/no_import_node_test.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::ImportDeclaration(import_decl) = node.kind() {
+            if import_decl.source.value.eq("node:test") {
+                let span = import_decl.source.span;
+
+                ctx.diagnostic_with_fix(no_import_node_test(span), |fixer| {
+                    fixer.replace(span, "\"vitest\"")
+                });
+            }
+        };
```
mysteryven, line 57: Let's try use [module record](https://github.com/oxc-project/oxc/blob/a207923af1b87953e89ecbf0fe272cf7134b71ef/crates/oxc_syntax/src/module_record.rs#L17)

mysteryven (changes requested):
Let's try module record to see if it can improve our performance👀

To run benchmark for vitest rule, we also need to modify: https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/tasks/benchmark/benches/linter.rs#L44-L51

eryue0220:
> Let's try module record to see if it can improve our performance👀
> 
> To run benchmark for vitest rule, we also need to modify:
> 
> https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/tasks/benchmark/benches/linter.rs#L44-L51

still the same problem. :(

mysteryven:
> still the same problem. :(

We can try to use module record. e.g. 

1. https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/rules/import/no_self_import.rs#L30-L46
2. https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/frameworks.rs#L83-L85

eryue0220:
Thanks, it works!!

mysteryven (approved):
Nice!

PR #4413, opened 2024-07-22
docs(ast): add doc comments to more AST nodes

DonIsaac:
# What This PR Does
Adds more doc comments and examples to AST nodes

PR #4471, opened 2024-07-25
docs(ast): auto-generate doc comments for AstBuilder methods

DonIsaac:
# What This PR Does

Modifies `ast_codegen` to auto-generate rustdoc comments for generated `AstBuilder` methods. As we add more doc comments to AST node fields, the generated documentation will get better.

image

### DIFF
--- a/crates/oxc_ast/src/generated/ast_builder.rs
+++ b/crates/oxc_ast/src/generated/ast_builder.rs
@@ -26,26 +26,61 @@ pub struct AstBuilder<'a> {
 }
 
 impl<'a> AstBuilder<'a> {
+    /// Builds a [`BooleanLiteral`]
+    ///
+    /// If you want the built node to be allocated in the memory arena, use [`AstBuilder::alloc_boolean_literal`] instead.
+    ///
+    /// ## Parameters
+    /// - span: The [`Span`] covering this node
+    /// - value
     #[inline]
     pub fn boolean_literal(self, span: Span, value: bool) -> BooleanLiteral {
         BooleanLiteral { span, value }
     }
 
+    /// Builds a [`BooleanLiteral`] and stores it in the memory arena.
+    ///
+    /// Returns a [`Box`] containing the newly-allocated node. If you want a stack-allocated node, use [`AstBuilder::boolean_literal`] instead.
+    ///
+    /// ## Parameters
+    /// - span: The [`Span`] covering this node
+    /// - value
     #[inline]
     pub fn alloc_boolean_literal(self, span: Span, value: bool) -> Box<'a, BooleanLiteral> {
         Box::new_in(self.boolean_literal(span, value), self.allocator)
     }
 
+    /// Builds a [`NullLiteral`]
+    ///
+    /// If you want the built node to be allocated in the memory arena, use [`AstBuilder::alloc_null_literal`] instead.
+    ///
+    /// ## Parameters
+    /// - span: The [`Span`] covering this node
     #[inline]
     pub fn null_literal(self, span: Span) -> NullLiteral {
         NullLiteral { span }
     }
 
+    /// Builds a [`NullLiteral`] and stores it in the memory arena.
+    ///
+    /// Returns a [`Box`] containing the newly-allocated node. If you want a stack-allocated node, use [`AstBuilder::null_literal`] instead.
+    ///
+    /// ## Parameters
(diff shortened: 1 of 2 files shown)


DonIsaac:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4471?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4471** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4471?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4413** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4413?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -96,16 +123,43 @@ pub enum TSEnumMemberName<'a> {
...
+///
+/// A type that is a literal value. Wraps a [`TSLiteral`].
+///
+/// ## Example
+/// ```ts
+/// const x: 'foo' = 'foo';
+/// //       ^^^^^
+///
+/// type NonZero<N> = N extends 0 ? never : N;
+/// //                          ^
```
Dunqing, line 159: wrong pointing

Dunqing (dismissed):

Dunqing:
Looks good to me. Add these docs very useful for contributors who are new to `oxc`!  Also, I have noticed #2803 is doing the same thing. @Boshen cc

Boshen (commented):
LGTM

tasks/ast_codegen/src/generators/ast_builder.rs
```diff
@@ -309,6 +347,139 @@ impl ToTokens for Param {
     }
 }
 
+#[derive(Debug)]
+struct DocComment<'p> {
```
rzvxa, line 351: It'll be nice if we move this type to somewhere else so we can use it across the board. I haven't spent time documenting the generated code but this is a nice way to do it.

rzvxa:
This is an amazing addition, LGTM. We can move the `DocComment` to somewhere else in an upcoming PR.

crates/oxc_ast/src/generated/ast_builder.rs
```diff
@@ -3489,16 +5470,36 @@ impl<'a> AstBuilder<'a> {
...
     }
 
+    /// Builds a [`BindingPattern`]
+    ///
+    /// If you want the built node to be allocated in the memory arena, use [`AstBuilder::alloc_binding_pattern`] instead.
+    ///
+    /// ## Parameters
+    /// - kind
+    /// - type_annotation
+    /// - optional
```
rzvxa, line 5502: It would be super cool if we add an example here to let the user know what is the ideal usage(especially when some optional arguments are None).

e.g.
```rust
let id = builder.binding_pattern(
    binding_kind,
    Option::<TSTypeAnnotation>::None,
    false
);
```

The compiler tells the user to do this which is less than ideal.

```rust
let id = builder.binding_pattern::<Option<TSTypeAnnotation>>(binding_kind, None, false);
```

rzvxa: I just want to add a lot of issues with lots of booleans and optional types would go away as we move toward traced constructors and template macros.
Let me know if you have any suggestions to help with the interface.

overlookmotel:
A couple of LGTM comments, but no-one has merged? Is there any reason not to at this point? @rzvxa I think you're the one who's most on top of this?

crates/oxc_ast/src/ast/ts.rs
DonIsaac: image

tasks/ast_codegen/src/generators/ast_builder.rs
DonIsaac: Where would you like it moved? And could this be done in a following PR?

crates/oxc_ast/src/generated/ast_builder.rs
DonIsaac: I agree with the examples. I think this could be tackled in a following PR, so I'm resolving this comment.

PR #4476, opened 2024-07-26
perf(sourcemap): pre allocate String buf while encoding

Brooooooklyn:
I'm trying llama3.1:70b following [local-ai-copilot](https://developer.ibm.com/tutorials/awb-local-ai-copilot-ibm-granite-code-ollama-continue/) to find some potential performance improvement.

The code was not completed by AI; it only identified potential optimizations, which I am testing for effectiveness.

### DIFF
--- a/crates/oxc_sourcemap/src/encode.rs
+++ b/crates/oxc_sourcemap/src/encode.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 #[cfg(feature = "concurrent")]
 use rayon::prelude::*;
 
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
* **#4476** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4476?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Dunqing:
lol

overlookmotel:
The benchmark result here is not great, but I do think llama is onto something. There is a lot of memory-copying and allocation going on here, and it's largely unnecessary.

I guess the first question is: Does `SourceMap::to_json_string` needs to return a `String`?

Presumably main use case for this method is that you call it and then write the output to file. For that use case, the result type only needs to be iterated over, and not indexed into or sliced, and therefore it could be a type which either:

* under the hood is more like a `Vec<Cow<str>>` or 
* builds output lazily and implements [std::io::Read](https://doc.rust-lang.org/nightly/std/io/trait.Read.html)

Brooooooklyn:
Maybe there are some bugs in Codespeed? I don't touch any codes in the regression cases, and this pr shows the similar regression: https://github.com/oxc-project/oxc/pull/4477#issuecomment-2252260195

overlookmotel:
Yes, CodSpeed is using "base" of commit [ccb1835](https://github.com/oxc-project/oxc/commit/ccb1835a459f902e97c137c5edb8c193cf74aa1a) for comparison, whereas the merge commit that benchmarks are running on is based on [0c13442](https://github.com/oxc-project/oxc/commit/0c134422bacb1986b080ca5ea4103e5cedc12012). So the apparent regression is likely due to  [0c13442](https://github.com/oxc-project/oxc/commit/0c134422bacb1986b080ca5ea4103e5cedc12012), not this PR.

What is less clear is why upgrading to Rust 1.80.0 is having such a negative effect on `prepass` benchmark.

overlookmotel:
#4479 has updated CodSpeed's records of benchmarks to include the regression post-Rust 1.80.0 upgrade.

Do you want to rebase on latest main, so we can get an accurate picture of perf impact of this PR?

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -26,37 +28,38 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
...
+    contents.push("\"names\":[".into());
     let names = sourcemap
         .names
         .iter()
         .map(|x| serde_json::to_string(x.as_ref()))
         .collect::<std::result::Result<Vec<_>, serde_json::Error>>()
         .map_err(Error::from)?;
-    buf.push_str(&names.join(","));
-    buf.push_str("],\"sources\":[");
+    contents.push(Cow::Owned(names.join(",")));
```
overlookmotel, line 51: Rather than `join`ing here, how about allocating slots in `contents` for each name, and pushing them individually?

```rs
// NB 17 not 16 to handle case where `sourcemap.names.len() == 0`
let mut contents = PreAllocatedString::new(17 + sourcemap.names.len() * 2);
// ...
contents.push("\"names\":[".into());
for name in &sourcemap.names {
    let encoded = serde_json::to_string(name.as_ref()).map_err(Error::from)?;
    contents.push(Cow::Owned(encoded));
    contents.push(Cow::Borrowed(","));
}
contents.pop();
```

Could do same for all the other strings which are currently `join`ed. It should have biggest effect with `source_contents`, as will avoid copying those long strings twice.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -26,37 +28,39 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
...
-    buf.push_str("],\"sources\":[");
-    let sources = sourcemap
-        .sources
-        .iter()
-        .map(|x| serde_json::to_string(x.as_ref()))
-        .collect::<std::result::Result<Vec<_>, serde_json::Error>>()
-        .map_err(Error::from)?;
-    buf.push_str(&sources.join(","));
+    // Remove the last `,`.
+    contents.pop();
```
overlookmotel, line 53: I think this needs a check that `names` in not empty. If it is empty, then the loop won't execute and `.pop()` will pop off `"names":[`.

```rs
if !sourcemap.names.is_empty() {
    for n in &sourcemap.names {
        contents.push(serde_json::to_string(n.as_ref())?.into());
        contents.push(",".into());
    }
    contents.pop();
}
```

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -26,37 +28,39 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
...
+    contents.push(Cow::Borrowed("],\"sources\":["));
+    for s in &sourcemap.sources {
+        contents.push(serde_json::to_string(s.as_ref())?.into());
+        contents.push(",".into());
+    }
+    // Remove the last `,`.
+    contents.pop();
```
overlookmotel, line 60: Same as above. Need to handle if `sources` is empty.

overlookmotel (changes requested):

overlookmotel (approved):

PR #4480, opened 2024-07-26
fix(semantic): params in `export default (function() {})` flagged as `SymbolFlags::Export`

### LINKED ISSUE #4475: bug(semantic): params in `export default (function() {})` flagged as `SymbolFlags::Export`
> Found in #4445 

Parameters in parenthesis-wrapped function expressions in `ExportDefaultDeclaration`s are being incorrectly flagged as exports.
```ts
// correct: `a` is not flagged as exported
export default function (a) { }
// incorrect: `a` is flagged as exported
export default (function(a) {});
```

image

Dunqing:
The code changes in `builder.rs` optimize the export flag check in the `SemanticBuilder` implementation. Instead of setting the export flag to `true` for all cases except `ExportDefaultDeclarationKind::ClassDeclaration(ref class)`, the flag is now set to `true` only for `ExportDefaultDeclarationKind::ClassDeclaration` and `false` for all other cases. This change improves the efficiency of the export flag check.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1675,7 +1675,8 @@ impl<'a> SemanticBuilder<'a> {
                         func.id.is_some()
                     }
                     ExportDefaultDeclarationKind::ClassDeclaration(ref class) => class.id.is_some(),
-                    _ => true,
+                    ExportDefaultDeclarationKind::TSInterfaceDeclaration(_) => true,
+                    _ => false,
                 } {
                     self.current_symbol_flags |= SymbolFlags::Export;
                 }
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#4480** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

DonIsaac:
How about this case?
```ts
export default function foo() {
}
```
will `foo` still be marked as exported?

DonIsaac:
Yeah, this is breaking default exports for interfaces an enums
```ts
export default interface Foo {}
```

Dunqing:
> Yeah, this is breaking default exports for interfaces.
> 
> ```ts
> export default interface Foo {}
> ```

Ah, Thanks you found this. I also noticed we even don't have `Foo` symbol. 😅

DonIsaac (approved):

PR #4483, opened 2024-07-26
ci: use deterministic allocator for benchmarks

overlookmotel:
An attempt to reduce variance in benchmarks. Just testing at this stage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1375,6 +1375,8 @@ dependencies = [
  "oxc_transformer",
  "serde",
  "serde_json",
+ "spin",
+ "talc",
 ]
 
 [[package]]
@@ -2535,6 +2537,9 @@ name = "spin"
 version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
+dependencies = [
+ "lock_api",
+]
 
 [[package]]
 name = "static_assertions"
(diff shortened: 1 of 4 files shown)


overlookmotel:
* **#4483** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4483?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This does seem to work in that it removes almost all variance from the benchmarks.

Results for 5 dummy commits on #4485 which is on top of this PR:

benches

Remaining problem is that changing the allocator is showing as quite a significant speed-up on many benchmarks. Probably this allocator is faster than system allocator! Need to slow it down somehow so it matches system allocator better.

Boshen:
> Need to slow it down somehow so it matches system allocator better.

Probably doesn't matter, we are interested in relative change, not the absolute value from codspeed.

overlookmotel:
> Probably doesn't matter, we are interested in relative change, not the absolute value from codspeed.

Problem is that I think it will give us unrealistic *relative* results too.

Let's say we introduce a change that removes a bunch of allocations, but requires some extra work to do that (caching structs, bookkeeping etc). That is very likely to be a performance gain in real world, but if the allocator we use for benchmarks makes allocation unrealistically cheap (as this one does), benchmarks will lie to us and tell us it's a perf regression. For example, with this allocator, benchmarks probably would have told us #4213 was a bad change.

Ditto for using an arena allocator - benchmarks may tell us that using bumpalo is hurting perf (because its not much cheaper than our unrealistically cheap "global" allocator), when in fact it's a large perf gain.

NB: Allocations are a small part of the code overall. So if we're seeing 10% perf boost on some benchmarks, probably that means this allocator is ~double the speed of the system one. That's a very big discrepancy.

We might have more luck with https://crates.io/crates/dlmalloc which it sounds like is a closer analogue to the default system allocator (from `libc`), and so may reduce this discrepancy. I'll look into that.

overlookmotel:
Have asked for help on CodSpeed Discord: https://discord.com/channels/1065233827569598464/1065686090452828251/threads/1267428778691399690

Boshen:
Merge or close?

overlookmotel:
Close!

PR #4484, opened 2024-07-26
fix(syntax): correct `is_reserved_keyword_or_global_object`'s incorrect function calling.

7086cmd:
It may be a problem, but doesn't matter previously.

Formerly, the `is_reserved_keyword_or_global_object` is `is_reserved_keyword(s) || is_reserved_keyword(s)`. I think it should be `is_reserved_keyword(s) || is_global_object(s)` according to its name.

Also, the `.idea` may be because I am using RustRover, which may automatically create `.idea` folder. So I ignore it in `.gitignore`.

I think I can contribute to `oxc` more when I am free.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 target/
 **/*.rs.bk
+.idea
 
 /node_modules/
 /website/node_modules/
--- a/crates/oxc_syntax/src/keyword.rs
+++ b/crates/oxc_syntax/src/keyword.rs
@@ -2,7 +2,7 @@ use phf::{phf_set, Set};
 
 #[inline]
 pub fn is_reserved_keyword_or_global_object(s: &str) -> bool {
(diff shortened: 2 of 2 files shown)


DonIsaac (approved):

overlookmotel:
@7086cmd @DonIsaac Just a quick note: `.idea` and files like that shouldn't be in `.gitignore`. Please see #4355 for the reason (when I tried to add `.DS_Store`!)

DonIsaac:
@overlookmotel thanks for the heads up, I'll make a PR removing it tomorrow

overlookmotel:
No worries. I already removed it, just wanted to let you know why.

7086cmd:
> @7086cmd @DonIsaac Just a quick note: `.idea` and files like that shouldn't be in `.gitignore`. Please see #4355 for the reason (when I tried to add `.DS_Store`!)

Thanks! I understood.

PR #4487, opened 2024-07-26
perf(sourcemap): use simd to escape JSON string

Brooooooklyn:
Also optimize the memory allocation in string escape. The default size in `serde_json` is 1024 for String type, we pre allocate `string.len() * 2 + 2` for every string to reduce re-allocate in escaping.

I've tried to hand write SIMD implementation, but it's too complex, so I uses the `v_jsonescape` here. But it doesn't support `aarch64` and `wasm32` simd implementation, we need to contribute to it!

### DIFF
--- a/crates/oxc_sourcemap/src/encode.rs
+++ b/crates/oxc_sourcemap/src/encode.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 #[cfg(feature = "concurrent")]
 use rayon::prelude::*;
 
-use crate::error::{Error, Result};
+use crate::Error;
 use crate::JSONSourceMap;
 /// Port from https://github.com/getsentry/rust-sourcemap/blob/master/src/encoder.rs
 /// It is a helper for encode `SourceMap` to vlq sourcemap string, but here some different.
@@ -25,9 +25,9 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
     }
 }
 
-// Here using `serde_json::to_string` to serialization `names/source_contents/sources`.
+// Here using `serde_json` to serialize `names` / `source_contents` / `sources`.
 // It will escape the string to avoid invalid JSON string.
-pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String> {
+pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String, Error> {
     let max_segments = 12
         + sourcemap.names.len() * 2
         + sourcemap.sources.len() * 2
@@ -49,10 +49,11 @@ pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String> {
     }
 
     contents.push("\"names\":[".into());
-    contents.push_list(sourcemap.names.iter().map(to_json_string))?;
+    contents.push_list(sourcemap.names.iter().map(escape_json_string))?;
 
     contents.push("],\"sources\":[".into());
-    contents.push_list(sourcemap.sources.iter().map(to_json_string))?;
+
+    contents.push_list(sourcemap.sources.iter().map(escape_json_string))?;
 
     // Quote `source_content` in parallel
     if let Some(source_contents) = &sourcemap.source_contents {
@@ -61,18 +62,19 @@ pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String> {
             if #[cfg(feature = "concurrent")] {
                 let quoted_source_contents: Vec<_> = source_contents
                     .par_iter()
-                    .map(to_json_string)
+                    .map(escape_json_string)
                     .collect();
                 contents.push_list(quoted_source_contents.into_iter())?;
             } else {
-                contents.push_list(source_contents.iter().map(to_json_string))?;
+                contents.push_list(source_contents.iter().map(escape_json_string));
             }
         };
     }
 
     if let Some(x_google_ignore_list) = &sourcemap.x_google_ignore_list {
         contents.push("],\"x_google_ignoreList\":[".into());
-        contents.push_list(x_google_ignore_list.iter().map(|ignore| Ok(ignore.to_string())))?;
+        contents
+            .push_list(x_google_ignore_list.iter().map(std::string::ToString::to_string).map(Ok))?;
     }
 
     contents.push("],\"mappings\":\"".into());
@@ -83,11 +85,6 @@ pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String> {
     debug_assert!(contents.num_segments() <= max_segments);
 
     Ok(contents.consume())
-}
-
-#[inline]
-fn to_json_string<S: AsRef<str>>(s: S) -> Result<String> {
-    serde_json::to_string(s.as_ref()).map_err(Error::from)
 }
 
 #[allow(clippy::cast_possible_truncation)]
@@ -214,9 +211,9 @@ impl<'a> PreAllocatedString<'a> {
     }
 
     #[inline]
-    fn push_list<I>(&mut self, mut iter: I) -> Result<()>
+    fn push_list<I>(&mut self, mut iter: I) -> Result<(), Error>
     where
-        I: Iterator<Item = Result<String>>,
+        I: Iterator<Item = Result<String, Error>>,
     {
         let Some(first) = iter.next() else {
             return Ok(());
@@ -227,7 +224,6 @@ impl<'a> PreAllocatedString<'a> {
             self.push(Cow::Borrowed(","));
             self.push(Cow::Owned(other?));
         }
-
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
* **#4487** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @Brooooooklyn and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Dunqing (dismissed):
conflict

overlookmotel:
SIMD! Now we're talking!

But please don't merge before I've reviewed in full. I'm afraid I'm tied up this weekend, so that won't be until Monday.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -238,6 +286,21 @@ impl<'a> PreAllocatedString<'a> {
     }
 }
 
+#[inline]
+fn needs_escaping(s: &str) -> bool {
+    let bytes = s.as_bytes();
+
+    // Check for control characters (0x00-0x1F)
+    if memchr_iter(b'\x00', bytes).next().is_some() {
```
overlookmotel, line 294: Doesn't this only check for bytes which are 0, rather than all control chars (0x00-0x1F)?

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -45,43 +49,77 @@ pub fn encode_to_string(sourcemap: &SourceMap) -> Result<String> {
...
-                    .collect::<std::result::Result<Vec<_>, serde_json::Error>>()
-                    .map_err(Error::from)?;
-            } else {
-                let quote_source_contents = source_contents
-                    .iter()
-                    .map(|x| serde_json::to_string(x.as_ref()))
-                    .collect::<std::result::Result<Vec<_>, serde_json::Error>>()
-                    .map_err(Error::from)?;
+                    .filter_map(|s| {
+                        if needs_escaping(s) {
```
overlookmotel, line 91: Can probably omit `needs_escaping(s)` call here. It'd be extremely uncommon for source content not to include at least 1 line break (which needs escaping), so it will basically always return true.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -255,6 +318,14 @@ fn test_encode() {
     }
 }
 
+#[inline]
+fn serde_escape_content<S: AsRef<str>>(s: S) -> Result<String> {
+    let mut escaped_buf = Vec::with_capacity(s.as_ref().len() + 2);
```
overlookmotel, line 323: Capacity should be larger. Assuming the content does contain escapes, the output will definitely be longer than the input + 2 quotes.

Question is how much extra capacity to allocate. I think it's fairly harmless to allocate *too much* capacity, as this string is just temporary and will be dropped almost immediately, whereas allocating too little capacity will cause the `Vec` to grow which involves reallocation and memory copy. So probably a generous over-estimate like `s.as_ref().len() * 2 + 2` would be OK.

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -27,7 +27,7 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
 
 // Here using `serde_json::to_string` to serialization `names/source_contents/sources`.
```
overlookmotel, line 28:
```suggestion
// Here using `v_jsonescape::b_escape` to serialize `names` / `source_contents` / `sources`.
```

overlookmotel (changes requested):

overlookmotel:
10% speed up!

Unfortunately, I think the way we're using `v_jsonescape` here may be dangerous. The crate is not well-documented, but from what I can see, it's determining at *build time* what CPU features (SSE/AVX) are available, and then it assumes that the CPU where the code *runs* has those same features.

If you build on same machine as you run the code on, no problem. But if you build a binary on a machine which has AVX2 support, that binary will crash if run on another x86 machine which only has SSE2. This seems like a bad idea - people usually expect they're building portable binaries.

[build.rs](https://github.com/zzau13/v_escape/blob/v_escape_codegen-v0.1.8/v_jsonescape/build.rs):

```rs
fn main() {
    #[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
    {
        if is_x86_feature_detected!("sse2") {
            println!("cargo:rustc-cfg=v_escape_sse");
        }
        if is_x86_feature_detected!("avx2") {
            println!("cargo:rustc-cfg=v_escape_avx");
        }
    }
}
```

[Runtime code](https://github.com/zzau13/v_escape/blob/22794972b2d7fc2551c8a269f5f7d79db73202b1/v_jsonescape/src/lib.rs#L1327-L1332):

```rs
pub fn b_escape<B: buf_min::Buffer>(s: &[u8], buf: &mut B) {
    #[allow(unused_unsafe)]
    unsafe {
        _b_escape(s, buf)
    }
}
```

[Runtime code](https://github.com/zzau13/v_escape/blob/v_escape_codegen-v0.1.8/v_jsonescape/src/lib.rs#L1368-L1384):

```rs
pub unsafe fn _b_escape<B: buf_min::Buffer>(bytes: &[u8], buf: &mut B) {
    #[cfg(not(v_escape_avx))]
    {
        #[cfg(not(v_escape_sse))]
        {
            scalar::b_escape(bytes, buf)
        }
        #[cfg(v_escape_sse)]
        {
            ranges::sse::b_escape(bytes, buf)
        }
    }
    #[cfg(v_escape_avx)]
    {
        ranges::avx::b_escape(bytes, buf)
    }
}
```

This problem is mentioned in https://github.com/zzau13/v_escape/issues/54.

It looks like we'd be better off using `v_escape::encode` which [uses runtime CPU feature detection](https://github.com/zzau13/v_escape/blob/v_escape_codegen-v0.1.8/v_jsonescape/src/lib.rs#L1335-L1359).

overlookmotel:
Also, perhaps we should merge #4528 first (if it's correct) so we can see effect on benchmarks once the benchmark is measuring what it's meant to.

Brooooooklyn:
> Unfortunately, I think the way we're using v_jsonescape here may be dangerous. The crate is not well-documented, but from what I can see, it's determining at build time what CPU features (SSE/AVX) are available, and then it assumes that the CPU where the code runs has those same features.

I agree; I can't even run `cargo test` successfully in the `v_escape` repo. I think it's worth maintaining our own simd JSON escape implementation and supporting the aarch64/wasm32 backend in it.

So let's just pre-allocate in this pr and implement JSON escape in upcoming separated PRs.

overlookmotel:
I think we can go forwards with this PR if we switch to using `v_jsonescape::encode` instead of `b_encode`. `encode` [uses runtime CPU feature detection](https://github.com/zzau13/v_escape/blob/v_escape_codegen-v0.1.8/v_jsonescape/src/lib.rs#L1335-L1359), so doesn't have the problem that `b_escape` does.

I agree that longer term, we'd be better off maintaining our own crate for SIMD ops (lexer and various other parts of Oxc could also benefit a lot from SIMD) and support aarch64 + wasm32 (and also AVX512, which would likely be almost x2 speed boost for some workloads on CPUs that support it).

But for now, this PR is producing a significant speed boost for common case of x86_64, so I think it's worthwhile merging it in meantime if we can remove usage of `b_encode`.

Personally, I think implementing a SIMD crate will not be a straightforward task. Not so much the code itself, but building the testing infra around it to make sure it's completely solid on all targets (some discussion of this in https://github.com/oxc-project/oxc/discussions/2285). But maybe you're familiar @Brooooooklyn with the cross-platform testing side of things from napi-rs?

crates/oxc_sourcemap/src/encode.rs
```diff
@@ -25,9 +25,9 @@ pub fn encode(sourcemap: &SourceMap) -> JSONSourceMap {
     }
 }
 
-// Here using `serde_json::to_string` to serialization `names/source_contents/sources`.
+// Here using `v_jsonescape::b_escape` to serialize `names` / `source_contents` / `sources`.
```
overlookmotel, line 28: Ha ha ha back again!

```suggestion
// Here using `serde_json` to serialize `names` / `source_contents` / `sources`.
```

crates/oxc_sourcemap/Cargo.toml
```diff
@@ -19,11 +19,11 @@ workspace = true
...
-rustc-hash  = { workspace = true }
-serde       = { workspace = true, features = ["derive"] }
-serde_json  = { workspace = true }
-base64-simd = { workspace = true }
-cfg-if      = { workspace = true }
+rustc-hash   = { workspace = true }
+serde        = { workspace = true, features = ["derive"] }
+serde_json   = { workspace = true }
+base64-simd  = { workspace = true }
+cfg-if       = { workspace = true }
```
overlookmotel, line 26: Remove from diff (if you can be bothered).

```suggestion
rustc-hash  = { workspace = true }
serde       = { workspace = true, features = ["derive"] }
serde_json  = { workspace = true }
base64-simd = { workspace = true }
cfg-if      = { workspace = true }
```

overlookmotel (approved):

overlookmotel:
Merging! Hooray. @Brooooooklyn I hope you haven't found my approach to review too annoyingly nit-picky. If it is, please say so. I'm new to being maintainer on a popular project. I know Boshen takes a much more light-touch "if the tests pass, merge it" approach, so I'm not sure if I'm out of line.

PR #4498, opened 2024-07-27
perf(mangler): reduce unnecessary allocation

Dunqing:
There seems to be no noticeable change in performance.


Boshen:
Hold on, let me double check the benchmark ... if there is one.

PR #4505, opened 2024-07-27
refactor(semantic): move CatchClause scope binding logic to visit_block_statement

Dunqing:
This logic only needs to run in the BlockStatement

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -307,6 +307,7 @@ impl<'a> SemanticBuilder<'a> {
         }
     }
 
+    #[inline]
     pub fn current_scope_flags(&self) -> ScopeFlags {
         self.scope.get_flags(self.current_scope_id)
     }
@@ -522,14 +523,6 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         self.current_scope_id = self.scope.add_scope(parent_scope_id, self.current_node_id, flags);
         scope_id.set(Some(self.current_scope_id));
 
-        if self.scope.get_flags(parent_scope_id).is_catch_clause() {
-            // Move all bindings from parent scope to current scope
-            // to make it easier to resole references and check redeclare errors.
-            let parent_bindings =
-                self.scope.get_bindings_mut(parent_scope_id).drain(..).collect::<Bindings>();
-            *self.scope.get_bindings_mut(self.current_scope_id) = parent_bindings;
-        }
-
         self.unresolved_references.increment_scope_depth();
     }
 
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#4505** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -674,6 +667,32 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
...
+            Some(self.scope.get_bindings_mut(self.current_scope_id).drain(..).collect::<Bindings>())
+        } else {
+            None
+        };
+
+        self.enter_scope(ScopeFlags::empty(), &it.scope_id);
+
+        if let Some(parent_bindings) = parent_bindings {
+            *self.scope.get_bindings_mut(self.current_scope_id) = parent_bindings;
+        }
```
overlookmotel, line 688: I suggest:

```suggestion
        let parent_scope_id = self.current_scope_id;

        self.enter_scope(ScopeFlags::empty(), &it.scope_id);

        // Move all bindings from parent scope to current scope
        // to make it easier to resole references and check redeclare errors.
        // NOTE: enter_scope hasn't been called yet, so `current_scope_id`` is the parent scope of the block's scope.
        if self.scope.get_flags(parent_scope_id).is_catch_clause() {
            let parent_bindings = std::mem::take(self.scope.get_bindings_mut(parent_scope_id));
            *self.scope.get_bindings_mut(self.current_scope_id) = parent_bindings;
        }
```

1 less branch, and a bit simpler.

NB `take` is better than `drain` + `collect` I think as `drain` + `collect` empties the original hash map, allocates a new one, and copies the contents over. `take` just grabs it and moves it.

Dunqing: Thanks!

overlookmotel (approved):
I've pushed 3 commits for style nits. I hope you don't mind - seems faster than requesting changes.

If you're OK with those changes, please go ahead and merge. Or if you're not, revert any you don't like... and go ahead and merge. All roads leads to merging!

PS: I like this change. Great to be moving logic off the hot path.

overlookmotel:
Oh wait! Why are some semantic benchmarks showing as -1%? Let's see if that still holds true after the benchmarks have run again after my style nit commits.

Dunqing:
> Looks like I was wrong that `mem::take` is better than `drain` + `collect`. I guess `drain` and `collect` has a shortcut fast path I didn't know about. I've switched it back to `drain` + `collect` and now benchmarks are positive.
> 
> Sorry I made that bad suggestion! I guess the lesson is: never assume you understand, always measure!

Agreed! Even with a bad suggestion. I still learned a lot of useful experiences from your attempt! You are the best!

> Also sorry I've hijacked your PR! I will stop fiddling and hand it back to you now.

No worries!

PR #4534, opened 2024-07-29
perf(parser): get and check lookahead token

lucab:


lucab:
* **#4534** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lucab:
I saw this method coming up as an hotspot when profiling, and I was hoping this could help a bit.
It doesn't look like it changes much, but maybe it is still nicer to read.

overlookmotel (approved):

overlookmotel:
> I saw this method coming up as an hotspot when profiling

@lucab That's interesting. I wonder if we can remove the lookahead from the lexer entirely and move lookahead logic purely into the parser in the places where it needs it?

As I understand it, lookahead is only required for a few things (arrow functions, TS type params) but currently every single time we get a token (any token) we have to pass through this code. I don't know the parser well enough to know how difficult it'd be to move the logic though.

PR #4567, opened 2024-07-30
(experiment) perf: create `Allocator` with an initial capacity

DonIsaac:


DonIsaac:
Seems like some benchmarks get a ~1% boost, while a few others get a -1% hit. It was worth a shot ¯\_(ツ)_/¯

PR #4578, opened 2024-07-31
feat(linter): eslint-plugin-vitest/require-local-test-context-for-concurrent-snapshots

eryue0220:
support [eslint-plugin-vitest/require-local-test-context-for-concurrent-snapshots](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/require-local-test-context-for-concurrent-snapshots.md)


crates/oxc_linter/src/utils/jest/parse_jest_fn.rs
```diff
@@ -546,6 +546,13 @@ const VALID_JEST_FN_CALL_CHAINS: [[&str; 4]; 51] = [
     ["describe", "skip", "each", ""],
     ["fdescribe", "", "", ""],
     ["fdescribe", "each", "", ""],
+    // vitest
```
mysteryven, line 549: Since there are too many items in [this file](https://github.com/veritem/eslint-plugin-vitest/blob/9885b9f2b15d316d2d4c2d8f8da410f2aa24c4a1/src/utils/valid-vitest-fn-call-chains.ts#L1), we can add another `VALID_VITEST_FN_CALL_CHAINS` and need to change it to a `set`.
 
 You can use `ctx.frameworks().is_vitest()` to decide which to use.
 
 https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/rules/jest/expect_expect.rs#L140

eryue0220: Done.

crates/oxc_linter/src/utils/jest/parse_jest_fn.rs
```diff
@@ -100,7 +101,8 @@ pub fn parse_jest_fn_call<'a>(
 
         let mut call_chains = Vec::from([Cow::Borrowed(name)]);
         call_chains.extend(members.iter().filter_map(KnownMemberExpressionProperty::name));
-        if !is_valid_jest_call(&call_chains) {
+
+        if !is_valid_jest_call(&call_chains) && !is_valid_vitest_call(&call_chains) {
```
mysteryven, line 105: I think we only need to check `is_valid_vitest_call` if it's using vitest. You can also use `ctx.frameworks().is_vitest` here.

mysteryven, line 105: Seems like this should be:

```js
if ctx.frameworks().is_jest() && !is_valid_jest_call(&call_chains) {
    return None;
}

if ctx.frameworks().is_vitest() && !is_valid_vitest_call(&call_chains) {
    return None;
}
```

After changing this, it will also expose some bugs in our jest rules. Are you interest in fixing them in another PR?

eryue0220: Sure, but I'm not sure where the bug is. 😅

I ran the `cargo test` and test cases are passed.

mysteryven: We need to update this first:

CleanShot 2024-08-03 at 09 46 04@2x

```
pub const fn is_jest(self) -> bool {
    self.contains(Self::Jest)
}
```

eryue0220: Updated

crates/oxc_linter/src/rules/jest/no_disabled_tests.rs
```diff
@@ -153,73 +153,129 @@ fn run<'a>(possible_jest_node: &PossibleJestNode<'a, '_>, ctx: &LintContext<'a>)
...
-        ("it.todo('fill this later')", None),
-        ("var appliedSkip = describe.skip; appliedSkip.apply(describe)", None),
-        ("var calledSkip = it.skip; calledSkip.call(it)", None),
-        ("({ f: function () {} }).f()", None),
-        ("(a || b).f()", None),
-        ("itHappensToStartWithIt()", None),
-        ("testSomething()", None),
-        ("xitSomethingElse()", None),
-        ("xitiViewMap()", None),
+        ("describe('foo', function () {})", None, None, Some(PathBuf::from(jest_path))),
```
mysteryven, line 161: We don't need to change test cases in rules, you can update path at once:

```diff
let path_to_lint = if self.import_plugin {
      assert!(path.is_none(), "import plugin does not support path");
      self.current_working_directory.join(&self.rule_path)
  } else if let Some(path) = path {
      self.current_working_directory.join(path)
+ } else if self.jest_plugin {
+     self.rule_path.with_extension("test.tsx")
  } else {
      self.rule_path.clone()
  };
```

mysteryven (changes requested):
Can you split the bug fixes in another PR? We are supposed this PR only contains things about `require-local-test-context-for-concurrent-snapshots` rule.

crates/oxc_linter/src/rules/jest/no_disabled_tests.rs
eryue0220: Okay, I'll updated it later.

eryue0220: Done.

mysteryven:
This PR looks huge, can you split the bug fixes into another PR?

PR #4580, opened 2024-07-31
(poc) perf(linter): shrink size of `LintContext`

DonIsaac:
Moves file-specific context information into a `LinterContext` struct and refactors `LintContext` to store a reference to it. The result is a much smaller `LintContext` struct with many fewer `Rc` clones. Locally I'm seeing a ~12% speedup on linter benchmarks.


DonIsaac:
Interesting, I was seeing much better benchmarks locally 
image

Boshen (commented):
I'm not sure about the crazy number of lifetimes, need to take a closer look.

`LinterContext` is a very weird state holder ...

DonIsaac:
I agree, which is why I made it a POC. I think there's something here to be done though. I tried an RC as well, but you cannot mutate the inner value, breaking builder methods. This needs further investigation.

PR #4583, opened 2024-07-31
perf(sourcemap): elide bounds checks in VLQ encoding

overlookmotel:
In `oxc_sourcemap`'s VLQ encoding, avoid bounds checks when pushing bytes to the encoded string in the hot loop.

Those bounds checks are quite expensive as they involve a function call to `alloc::raw_vec::RawVec::grow_one`, and that happens on every single pushed byte.

https://godbolt.org/z/44G8jjss3

Not much difference on benchmarks, as VLQ encoding is only a small part of source map generation, but a local benchmark of just VLQ encoding shows this increases performance by 11%.

### DIFF
--- a/crates/oxc_sourcemap/src/encode.rs
+++ b/crates/oxc_sourcemap/src/encode.rs
@@ -112,6 +112,9 @@ fn serialize_sourcemap_mappings(sm: &SourceMap) -> String {
     )
 }
 
+// Max length of a single VLQ encoding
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4583** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #4584, opened 2024-07-31
perf(allocator): use capacity hint

lucab:


lucab:
* **#4584** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

overlookmotel:
@lucab does `Vec::extend` not already ensure sufficient capacity from the size hint before it starts pushing? That's a genuine question, not a criticism! I don't know the answer. But if `extend` doesn't, it should!

lucab:
@overlookmotel unless I misunderstood the code, it performs a check-and-resize step for each element to push: https://docs.rs/allocator-api2/latest/src/allocator_api2/stable/vec/mod.rs.html#2758-2785

overlookmotel:
Thanks for coming back. Interesting. That's pretty damn inefficient!

That said, I guess it *has* to check on each turn of the loop, if the iterator isn't "trusted". I think it's legal (although unfortunate) for an iterator to provide an incorrect size hint (or one that's an estimate).

`std` has the concept of a "trusted iterator" which promises its size hint is 100% accurate, and then `extend` can just do 1 capacity check at the start, and not on every turn of the loop. But I'm not sure if that facility is available in "userland" - I think might require some nightly features.

I hope to replace our current `Vec` implementation with something new (https://github.com/oxc-project/backlog/issues/18), and can try to improve this situation at same time.

Or if you're interested, and up for a bit of an unsafe code brain-bender, would you like to look into it? That's quite a sizeable project though!

NB: https://github.com/oxc-project/backlog/issues/35 is also relevant, though I'm guessing we'll probably only use a chunked vec type to replace some `Vec`s, but not all of them, so we'll likely still need a standard `Vec` type.

PR #4318, opened 2024-07-17
feat(linter): implement `eslint-plugin-promise/no-return-in-finally, prefer-await-to-then` rule

jelly:
Part of https://github.com/oxc-project/oxc/pull/4252

PR #4598, opened 2024-08-01
feat(linter/eslint-plugin-promise): implement valid-params

jelly:
Rule detail:
[link](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/valid-params.md)

---

Split this up from https://github.com/oxc-project/oxc/pull/4318 to make it easier to review and as some of the other rules require some more complicated work.

### DIFF
--- a/crates/oxc_linter/src/rules/promise/valid_params.rs
+++ b/crates/oxc_linter/src/rules/promise/valid_params.rs
@@ -1,0 +1,189 @@
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, utils::is_promise, AstNode};
+
+fn zero_or_one_argument_required_diagnostic(
+    span0: Span,
+    prop_name: &str,
+    args_len: usize,
+) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!(
+        "Promise.{prop_name}() requires 0 or 1 arguments, but received {args_len}"
+    ))
+    .with_label(span0)
+}
+
+fn one_or_two_argument_required_diagnostic(
+    span0: Span,
+    prop_name: &str,
+    args_len: usize,
+) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!(
+        "Promise.{prop_name}() requires 1 or 2 arguments, but received {args_len}"
+    ))
+    .with_label(span0)
+}
+
+fn one_argument_required_diagnostic(
+    span0: Span,
+    prop_name: &str,
+    args_len: usize,
+) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!(
+        "Promise.{prop_name}() requires 1 argument, but received {args_len}"
+    ))
+    .with_label(span0)
+}
+
+fn valid_params_diagnostic(span0: Span, x0: &str) -> OxcDiagnostic {
+    OxcDiagnostic::warn(x0.to_string()).with_label(span0)
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct ValidParams;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Enforces the proper number of arguments are passed to Promise functions.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// Calling a Promise function with the incorrect number of arguments can lead to unexpected
+    /// behavior or hard to spot bugs.
+    ///
+    /// ### Example
+    /// ```javascript
+    /// Promise.resolve(1, 2)
+    /// ```
+    ValidParams,
+    correctness,
+);
+
+impl Rule for ValidParams {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::CallExpression(call_expr) = node.kind() else {
+            return;
+        };
+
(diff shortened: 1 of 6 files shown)


crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    // hello.then(), hello.catch(), hello.finally()
+    if matches!(prop_name, "then" | "catch" | "finally") {
+        return true;
+    }
+
+    member_expr.object().is_specific_id("Promise")
+        && matches!(
+            prop_name,
+            "resolve" | "reject" | "all" | "allSettled" | "race" | "any" | "withResolvers"
```
DonIsaac, line 50: Should we have a shared utility for promise names that can be used by other rules? We may also want to use `phf_set`.

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,177 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::CallExpression(call_expr) = node.kind() else {
+            return;
+        };
+
+        if !is_promise(call_expr) {
+            return;
+        }
+
+        let Some(member_expr) = call_expr.callee.get_member_expr() else {
```
DonIsaac, line 64: These two checks are redundant with logic in is_promise. Maybe have that function return the references you need?

jelly: Seems logical, I checked the other PR and at least `no-return-in-finallly` could re-use that, other rules can't but they can instead call `is_promse().is_some()`

crates/oxc_linter/src/rules/promise/valid_params.rs
jelly: I think I do that in another PR.

https://github.com/oxc-project/oxc/pull/4318/files#diff-0afbabe6273a5175aa7ba4cf2f9282b1b6318e738ab5d6b65b12ae831a469414R5

So  I will sneak that into this one :)

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,148 @@
...
+        let Some(prop_name) = is_promise(call_expr) else {
+            return;
+        };
+
+        let args_len = call_expr.arguments.len();
+
+        match prop_name.as_str() {
+            "resolve" | "reject" => {
+                if args_len > 1 {
+                    ctx.diagnostic(valid_params_diagnostic(call_expr.span, &format!("Promise.{prop_name}() requires 0 or 1 arguments, but received {args_len}")));
```
DonIsaac, line 48: Move message formatting logic into the diagnostics function

DonIsaac: Also, this code creates a String, passes a &str to the diagnostics function, which then creates another String. This results in a duplicate string allocation. The above suggestion fixes this.

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,148 @@
...
+                }
+            }
+            "then" => {
+                if !(1..=2).contains(&args_len) {
```
DonIsaac, line 52: While cleaner, using a range here creates an iterator, which produces unnecessary branches. Consider just using a greater than and less than or equals check.

crates/oxc_linter/src/snapshots/valid_params.snap
```diff
@@ -0,0 +1,146 @@
...
+
+  ⚠ eslint-plugin-promise(valid-params): Promise.allSettled() requires 1 argument, but received 4
+   ╭─[valid_params.tsx:1:1]
+ 1 │ Promise.allSettled({}, function() {}, 1, 2)
+   · ───────────────────────────────────────────
+   ╰────
+
+  ⚠ eslint-plugin-promise(valid-params): Promise.any() requires 1 argument, but received 3
+   ╭─[valid_params.tsx:1:1]
+ 1 │ Promise.any(1, 2, 3)
```
DonIsaac, line 66: Could we add a suggestion for this case? Using `ctx.diagnostic_with_suggestion`?

crates/oxc_linter/src/rules/promise/valid_params.rs
jelly: Sure, I have to figure out how to silence clippy then.

crates/oxc_linter/src/snapshots/valid_params.snap
jelly: Hmm well the arguments can be any type so I haven't figured out how to get the span location yet of the first argument. Or use the source text to determine the location of the comma

crates/oxc_linter/src/rules/promise/valid_params.rs
```diff
@@ -0,0 +1,198 @@
...
+                        args_len,
+                    ));
+                }
+            }
+            _ => {}
+        }
+    }
+}
+
+// fn generate_fix<'a>(
```
DonIsaac, line 112: looks like it needs deleting

DonIsaac:
One last thing, then we're good to go 🎊

crates/oxc_linter/src/rules/promise/valid_params.rs
jelly: Sorry, dropped it.

DonIsaac (approved):

PR #4603, opened 2024-08-02
perf(linter): change react rules and utils to use `Cow` and `CompactStr` instead of `String` 

DonIsaac:


crates/oxc_span/src/atom.rs
```diff
@@ -389,6 +390,25 @@ impl Serialize for CompactStr {
     }
 }
 
+#[cfg(feature = "schemars")]
```
Boshen, line 393: Why are these features part of this PR 🤔 ?

Issue #2085, opened 2024-01-19
How to distinguish `\ud808` and `\\ud808`?


mysteryven:
I want to print the String Literal correctly in Codegen, for `'\ud808'` and `'\\ud808'`, it should be different:

```bash
let x = '\\ud808  \ud808' => let x = "\\ud808  \uD808";
```

Our AST node's value is [the same](https://oxc-project.github.io/oxc/playground/?code=3YCAAICZgICAgICAgIC2mcpqppTW%2BJx5VcAGZWOyq7bkGDnoTy6%2Ff3bCAIA%3D). How can I solve it, need to add a `raw` to String Literal?

```bash
"init": {
      "type": "StringLiteral",
      "start": 8,
      "end": 25,
      "value": "\\ud808  \\ud808",
      "raw": "'\\\\ud808  \\ud808'"
 }
```

Boshen:
Can you give me the test case file? It feels like a parser bug, but I'm not sure.

mysteryven:
My brain is a bit of mush, They are from [ESBuild test case](https://github.com/oxc-project/oxc/blob/0e9104477fc94f01b453a19ee01f7f30b2063490/crates/oxc_minifier/tests/esbuild/mod.rs#L207) and [invalid-sccape-sequences.js](https://github.com/tc39/test262/blob/715dd1073bc060f4ee221e2e74770f5728e7b8a0/test/language/expressions/tagged-template/invalid-escape-sequences.js#L48-L51) respectively.

```
 '\\' => {
      let mut cloned_char = chars.clone();
      let next = cloned_char.next();
      let next_next = cloned_char.clone().next();
      if next.is_some_and(|c| matches!(c, 'u' | 'U'))
          && next_next.is_some_and(|c| c.is_ascii_hexdigit())
      {
          p.print_str(b"\\");
          p.print_str(b"u");
          // <https://mathiasbynens.be/notes/javascript-escapes#unicode>
          cloned_char.take(4).for_each(|c| {
              p.print_str(c.to_string().as_bytes());
          });
          for _ in 0..5 {
              chars.next();
          }
      } else {
          p.print_str(b"\\\\");
      }
  }
```


After passing the ESBuild test case, I converted `\\u0g` to '\u0g' accidentally, which not a valid Unicode, seems I need to know whether `\` is being escaped, our AST not show this.

Boshen:
[...]

PR #4605, opened 2024-08-02
fix(linter/eslint): Fix invalid regexp in no_regex_spaces test

leaysgur:
The same as #4544 

image

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs
@@ -228,7 +228,7 @@ fn test() {
         "var foo = RegExp(' [  ] [  ] ');",
         r"var foo = new RegExp(' \[   ');",
         r"var foo = new RegExp(' \[   \] ');",
-        r"var foo = /[\q{    }]/v;",
+        "var foo = /[\\q{    }]/v;",
         "var foo = new RegExp('[  ');",
         "new RegExp('[[abc]  ]', flags + 'v')",
     ];
--- a/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs
@@ -228,7 +228,7 @@ fn test() {
         "var foo = RegExp(' [  ] [  ] ');",
         r"var foo = new RegExp(' \[   ');",
         r"var foo = new RegExp(' \[   \] ');",
-        r"var foo = /[\q{    }[\q{    }]]/v;",
+        r"var foo = /[\q{    }]/v;",
         "var foo = new RegExp('[  ');",
         "new RegExp('[[abc]  ]', flags + 'v')",
     ];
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs
```diff
@@ -228,7 +228,7 @@ fn test() {
         "var foo = RegExp(' [  ] [  ] ');",
         r"var foo = new RegExp(' \[   ');",
         r"var foo = new RegExp(' \[   \] ');",
-        r"var foo = /[\\q{    }]/v;",
```
mysteryven, line 231: It looks like the Regexp it represents is `/[\q{    }]/v`.

```js
console.log("/[\\q{    }]/v")
```

leaysgur: > console.log("/[\\q{    }]/v")

This is not evaluated as `RegExp`, this is just a `String`. 👀 

image

mysteryven: Maybe I missed some information, but if I understand correct, the test case below 🤔

```js
{ code: "var foo = /[\\q{    }]/v;", languageOptions: { ecmaVersion: 2024 } },
```

is equal to the file content: 

```js
var foo = /[\q{    }]/v;
```

This is what we write in JS file. I tested it on NodeJS:

```js
// filename: a.js
const fs = require('fs')
const assert = require('assert')
let foo = "var foo = /[\\q{    }]/v;\n"

let input = fs.readFileSync('input.js', 'utf8');
assert.strictEqual(foo, input)

// filename: input.js
var foo = /[\q{    }]/v;
```

I know this because I've fallen into a similar trap: #2085

leaysgur: I may also missed something, 😅

But I'd like to fix here is Rust <-> Rust conversion, not Rust <-> JS.

```rs
r"var foo = /[\\q{    }]/v;",
```

By leading `r"..."`, we do not need to escape `\` unless we really need 2 `\`s.

And if there are 2 `\`(= original code), it's not a valid `RegExp` syntax.

mysteryven: In rust side, `r"var foo = /[\q{    }]/v;"` is equal to the test case in the original ESLint test case, they all represent the same JS file which content is:

CleanShot 2024-08-03 at 13 41 41@2x

leaysgur: > if we port it over to our side, it can be `r"var foo = /[\q{    }]/v;"` or `"var foo = /[\\q{    }]/v;"`. 👀

Yes. 

But it was `r"var foo = /[\\q{ }]/v;"`. So I've fixed it by just removing the leading `r` now.

(In my first commit, I tried to fix this by removing the extra `\`. I also added the `NestedClass` pattern to expand the test case. This may have led to misunderstandings. Now I have minimized the diff for sure~.)

mysteryven (approved):

PR #4609, opened 2024-08-02
feat(ast): `GetSpanMut` trait

### LINKED ISSUE #4606: Introduce `span_mut` for `GetSpan`
There's situiation that we need turn a identifier `a` into various form depend on the conditions. And we want to apply the `span` of  `a` to the final form. The current solution make it complex to do it.

```rust
let span_of_a = a.span;
let expr = if rewrite_to_member_expr {
  member_expr!(a, SPAN)
} else {
  a
 }
let expr = if is_callee {
  member_expr_again!(expr, SPAN)
} else {
  expr
};

expr.span = span_of_a;
```

You will see we have to apply the `span` of `a` in the end, otherwise the sourcemap might be wrong, which is caused by nested ast nodes have the same span. I guess this is allowed maybe? But it's doesn't generate accurate sourcemap.

overlookmotel:
Closes #4606.

Introduce `GetSpanMut` trait and implement it on all AST node types.

This has to be a separate trait, rather than adding `span_mut` method to `GetSpan` because `AstKind` implements `GetSpan`, and it only has an immutable `&` ref to AST node it contains.

### DIFF
--- a/crates/oxc_ast/src/generated/span.rs
+++ b/crates/oxc_ast/src/generated/span.rs
@@ -4,12 +4,19 @@
 #![allow(clippy::match_same_arms)]
 
 use crate::ast::*;
-use oxc_span::{GetSpan, Span};
+use oxc_span::{GetSpan, GetSpanMut, Span};
 
 impl GetSpan for BooleanLiteral {
     #[inline]
     fn span(&self) -> Span {
         self.span
+    }
+}
+
+impl GetSpanMut for BooleanLiteral {
+    #[inline]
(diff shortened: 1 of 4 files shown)


overlookmotel:
* **#4609** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4609?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa (approved):
Nice!

PR #4610, opened 2024-08-02
fix(parser): parse `assert` keyword in `TSImportAttributes`

### LINKED ISSUE #4601: Oxc Parser - failed to parse import asserts in type alias
This is surprisingly valid TS:
```ts
type Vite =  typeof import("vite", { assert: { "resolution-mode": "import" } });
```

It looks like oxc fails to parse this in the playground.

[TS playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAaglsaBeKVSQPYDMpwLZgYBOwAFAEQBuCE5ANFAN5QCGAzmxCQFxNTlEIbDABsArsDgYAdgFo8GACa1e5fIRLkoAXx0BKANwAoIA)
[oxc playgroud](https://oxc-project.github.io/oxc/playground/?code=3YCAAIDTgICAgICAgICFnotj2NHm9OWvRF27hVW%2BBV57kTxewbAzOJibNSg%2FZ3tZa4nQcXQgVDHTJWvWxpJyjXacqrdae8XG43Q945iJo%2F1wjMCcPa2v7Dj0Ga9%2FXuAggA%3D%3D)

Boshen:
closes #4601


Dunqing:
TypeScript may deprecate the `assert` keyword in `TSImport` in 6.0  A rule or warning to disable `assert` may be added before the deprecation. We may need to follow-up updates of https://github.com/microsoft/TypeScript/issues/58453

Dunqing (approved):

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -1062,6 +1062,7 @@ pub struct TSImportType<'a> {
 pub struct TSImportAttributes<'a> {
     #[serde(flatten)]
     pub span: Span,
+    pub attributes_keyword: IdentifierName<'a>, // `with` or `assert`
```
armano2, line 1065: q: isn't better to use enum in here?

PR #4611, opened 2024-08-02
fix(docs): mark eslint and typescript alias rules as implemented,

### LINKED ISSUE #4085: Some of the rules should be marked as completed in eslint core rules
`default-parameter-last`, `max-params` and `no-array-constructor` is marked as completed in https://github.com/oxc-project/oxc/issues/479 but not in https://github.com/oxc-project/oxc/issues/2180 😅 Both need to be marked as completed.

We also need to map https://eslint.org/docs/latest/rules/no-new-symbol to the existing rule  no-new-native-nonconstructor

The relevant code is in https://github.com/oxc-project/oxc/tree/main/tasks/lint_rules

Sysix:
Closes #4085

### DIFF
--- a/tasks/lint_rules/src/eslint-rules.cjs
+++ b/tasks/lint_rules/src/eslint-rules.cjs
@@ -56,19 +56,24 @@ const {
 } = require("eslint-plugin-react-perf");
 // https://github.com/vercel/next.js/blob/canary/packages/eslint-plugin-next/src/index.ts
 const { rules: pluginNextAllRules } = require("@next/eslint-plugin-next");
+// https://github.com/eslint-community/eslint-plugin-promise/blob/main/index.js
+const {
+  rules: pluginPromiseRules,
+  configs: pluginPromiseConfigs
+} = require("eslint-plugin-promise");
+// https://github.com/veritem/eslint-plugin-vitest/blob/main/src/index.ts
+const {
+  rules: pluginVitestRules,
+  configs: pluginVitestConfigs
+} = require("eslint-plugin-vitest");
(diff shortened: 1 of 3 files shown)


tasks/lint_rules/src/eslint-rules.cjs
```diff
@@ -242,6 +285,9 @@ exports.ALL_TARGET_PLUGINS = new Map([
   ],
   ["react-perf", { npm: ["eslint-plugin-react-perf"], issueNo: 2041 }],
   ["nextjs", { npm: ["@next/eslint-plugin-next"], issueNo: 1929 }],
+  ["promise", { npm: ["eslint-plugin-promise"], issueNo: 9999 }], // TODO!
+  ["vitest", { npm: ["eslint-plugin-vitest"], issueNo: 9999 }], // TODO!
+  ["tree-shaking", { npm: ["eslint-plugin-tree-shaking"], issueNo: 9999 }], // TODO!
```
Sysix, line 290: tree-shaking has only one rule in node and in oxc. 

You should know better if an Tracking-Issue should be created or not ;)

leaysgur:
@Sysix Hey, thanks for the initiative!

Looking at the code diff, it appears that the content and title of the PR do not match?

And I have addressed the part of updating the `typescript-eslint` status in a separate PR. #4654 

Therefore, can I ask you to reduce the scope of this PR to the part that tracks the status of `plugin-promise` and `plugin-vitest`? (If `plugin-tree-shaking` has a single rule, then I don't think tracking is necessary either)

@Boshen Would you please create a ☂️ issue for `plugin-promise` and `plugin-vitest`?

Sysix:
@leaysgur thanks u, I created another PR: #4668

PR #4615, opened 2024-08-02
feat(ast_codegen): add alignment and size data to the schema.

rzvxa:
This PR generates the layouts in the schema but doesn't use it to reorder.

PR #4614, opened 2024-08-02
refactor(ast): make AST structs `repr(C)`.

rzvxa:
Make structs `#[repr(C)]` without reordering them to benchmark the performance implications of it.

This PR adds 136 bytes of padding in total. You can visit the list of these types [here](https://github.com/oxc-project/oxc/pull/4404/files/)(checkout the `assert_layouts.rs` diff).

Update:

Doesn't seem too bad!

image

* Linter: We can easily get more than 1% performance gain in the linter to even things out
* Prepass: If we fix our issue with Rust `1.80.0` we gain 13% And it should also be possible to find some other areas there to gain back this one percent.

The most notable thing is that `parser` isn't impacted by this, That is our most optimized crate which every percent counts(and it can be hard to gain back any perf regression there).

PR #4612, opened 2024-08-02
refactor(parser): use `ast_builder` in more places.

rzvxa:

Issue #4512, opened 2024-07-28
ast: IdentifierReference's reference_flag is always empty


Dunqing:
SemanticBuilder's already resolved reference usage but never doesn't set `reference_flag` for IdentifierReference.

Although we can get `reference_flag` by `reference_id`, would be great if we could directly `reference_flag`.

Dunqing:
> SemanticBuilder's already resolved reference usage but never doesn't set `reference_flag` for IdentifierReference.

I know the reason now. We can only get the immutable `IdentifierReference` from `AstKind`, so we can't change the `reference_flag`. In this way, we need to change `reference_flag: ReferenceFlag` to `reference_flag: Cell<ReferenceFlag>`

overlookmotel:
Agreed, it should be `Cell<ReferenceFlag>`. Or do we remove it entirely from `IdentifierReference`? Or remove it from `Reference`? It doesn't really need to be in 2 places does it?

Dunqing:
> make transformer/minifier more complex to have to update in 2 places when flags get changed.

Yes! It is better to remove `reference_flag` from `IdentifierReference`

overlookmotel:
> Yes! It is better to remove `reference_flag` from `IdentifierReference`

I agree that's the better one to remove. It makes more sense for that info to be stored in `Reference`.

Annoyingly, removing `reference_flag` from `IdentifierReference` doesn't save any bytes in AST, as that type has excess padding anyway. Whereas removing it from `Reference` would reduce size of `Reference` by 4 bytes.

But still, in my view it's more important that our data structures make logical sense, than that we've optimized every last drop out of them.

overlookmotel:
Closed in #5077.

PR #4617, opened 2024-08-02
fix(ast_codegen): detect "complex" type wrappers

rzvxa:
Detect types such as `Cell<Option<ScopeId>>` and mark them as such! We didn't used to use this method for these options but now we have to check all types to calculate their layouts which means we need to process them correctly(instead of falling to their inner value).

### DIFF
--- a/tasks/ast_codegen/src/util.rs
+++ b/tasks/ast_codegen/src/util.rs
@@ -61,6 +61,8 @@ pub trait StrExt: AsRef<str> {
 
 #[derive(Debug)]
 pub enum TypeIdentResult<'a> {
+    /// We bailed on detecting wrapper
+    Complex(Box<TypeIdentResult<'a>>),
     Ident(&'a Ident),
     Vec(Box<TypeIdentResult<'a>>),
     Box(Box<TypeIdentResult<'a>>),
@@ -81,6 +83,10 @@ impl<'a> TypeIdentResult<'a> {
         Self::Option(Box::new(inner))
     }
 
+    fn complex(inner: Self) -> Self {
+        Self::Complex(Box::new(inner))
+    }
+
     fn reference(inner: Self) -> Self {
         Self::Reference(Box::new(inner))
     }
@@ -88,9 +94,11 @@ impl<'a> TypeIdentResult<'a> {
     pub fn inner_ident(&self) -> &'a Ident {
         match self {
             Self::Ident(it) => it,
-            Self::Vec(it) | Self::Box(it) | Self::Option(it) | Self::Reference(it) => {
-                it.inner_ident()
-            }
+            Self::Complex(it)
+            | Self::Vec(it)
+            | Self::Box(it)
+            | Self::Option(it)
+            | Self::Reference(it) => it.inner_ident(),
         }
     }
 
@@ -116,6 +124,8 @@ pub enum TypeWrapper {
     OptBox,
     OptVec,
     Ref,
+    /// We bailed on detecting the type wrapper
+    Complex,
 }
 
 #[derive(Debug)]
@@ -151,7 +161,7 @@ impl TypeExt for Type {
                                     if seg1.ident == "Option" {
                                         TypeIdentResult::option(inner)
                                     } else {
-                                        inner
+                                        TypeIdentResult::complex(inner)
                                     }
                                 }
                                 Some(GenericArgument::Lifetime(_)) => {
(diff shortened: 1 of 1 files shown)


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4617?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4540** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4540?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4404** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4404?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4615** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4614** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4617** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4612** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
What constitutes "complex"? In practice at present, is it only `Cell<Option<T>>`?

A couple of suggestions:

Could we add `Cell` and `CellOpt` variants to `TypeWrapper`, to explicitly handle these, rather than a catch-all `Complex`?

At present are you assuming that `Cell<Option<T>>` has same size and alignment as `T`? If so, that's only true accidentally, because `ScopeId`, `SymbolId` and `ReferenceId` are all `NonZeroU32` under the hood, and so have a niche. If someone changed `ScopeId` to a `u32`, or wrapped some other field in a `Cell`, this assumption would break.

This isn't entirely academic - Dunqing was talking about making `IdentifierReference::reference_flag` a `Cell` in #4512.

Using the following rules would make type layout calculations more robust in the face of such changes:

* `Cell<T>` is same size and align as `T` (so where `T` is an `Option`, calculate size + alignment of the `Option` using same logic as for other `Option`s).
* `Cell<T>` never has a niche, even if `T` does.

Taking it a bit further (which I'm not suggesting doing now, but we maybe could in future) could make `TypeWrapper` more flexible by turning it into a `Vec<TypeWrapperPart>`. So then you could handle any possible combination of wrappers e.g. `my_field: Option<Vec<Option<Box<Something>>>`.

rzvxa:
> What constitutes "complex"? In practice at present, is it only Cell<Option<T>>?

Any type that has more than one part(e.g. Adt<T>) and `Adt` isn't covered by other wrappers.

> Could we add Cell and CellOpt variants to TypeWrapper, to explicitly handle these, rather than a catch-all Complex?

We sure can but we don't need it at the moment. We never care for misc types such as `Cell<Option<XXX>>`. We do not run any analysis on them. It is always just a simple match to detect these and do a predefined behavior.

> At present are you assuming that Cell<Option<T>> has same size and alignment as T? If so, that's only true accidentally, because ScopeId, SymbolId and ReferenceId are all NonZeroU32 under the hood, and so have a niche. If someone changed ScopeId to a u32, or wrapped some other field in a Cell, this assumption would break.

No; Instead we have the whole type as well-known (as opposed to having well-known wrappers and calculating the layout).

https://github.com/oxc-project/oxc/blob/8cdd9d9795000abbd4af461d255fd8f9adeac31f/tasks/ast_codegen/src/passes/calc_layout.rs#L356-L358

As you can see unlike the other wrappers which we only mark the wrapper part(for example `Vec`) as well-known, These are fully identified types with predefined layouts.

https://github.com/oxc-project/oxc/blob/8cdd9d9795000abbd4af461d255fd8f9adeac31f/tasks/ast_codegen/src/passes/calc_layout.rs#L219-L249

If a type isn't AST And doesn't have well-known information to calculate its layout, We check if the whole type is well-known or not. If it is we use that layout otherwise we fail to do so.

As a follow up we can calculate these, But we don't have to. I assert all of the types so if a well-known layout changes we get an updated assertion or old assertions break(if we don't call just ast with new layouts).


I think the only scenario where calculating Cell is necessary is if we use it to wrap actual AST types. In that case, we should add `Cell` as an actual wrapper.

overlookmotel (approved):

overlookmotel:
Thanks for clarifying. I hadn't clocked that layout for `Cell<Option<ScopeId>>` etc are hard-coded. That's totally fine for now.

PR #4618, opened 2024-08-02
feat(linter/eslint): implement func-names

Sysix:
Rule: https://eslint.org/docs/latest/rules/func-names

Implentation: https://github.com/eslint/eslint/blob/main/lib/rules/func-names.js


crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Require or disallow named function expressions
+    ///
+    /// ### Why is this bad?
+    ///
+    /// If you leave off the function name then when the function throws an exception you are likely
```
DonIsaac, line 44: Please improve this documentation. Consider copying it from the ESLint website.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// to get something similar to anonymous function in the stack trace. If you provide the optional
+    /// name for a function expression then you will get the name of the function expression in the stack trace.
+    ///
+    /// /// ### Example
+    ///
+    /// ```javascript
+    /// Foo.prototype.bar = function bar() {};
```
DonIsaac, line 51: Please provide good/bad examples and indicate which one is which. See the docs for no-unused-vars for an example.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+ */
+fn is_object_or_class_method(parent_node: Option<&AstNode>) -> bool {
+    if parent_node.is_none() {
+        return false;
+    }
+
+    match parent_node.unwrap().kind() {
+        AstKind::MethodDefinition(_) => true,
+        AstKind::ObjectProperty(property) => {
+            property.method
```
DonIsaac, line 69: Consider using the matches! Macro to avoid a function call.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+/**
+ * Determines whether the current FunctionExpression node has a name that would be
+ * inferred from context in a conforming ES6 environment.
+ */
+fn has_inferred_name(function: &Function, parent_node: Option<&AstNode>) -> bool {
+    if is_object_or_class_method(parent_node) {
+        return true;
+    }
+
+    if parent_node.is_none() {
```
DonIsaac, line 86: Redundant, since the above function already performs this check.

DonIsaac: If you need it to be some, consider doing 

`let Some(foo) = bar else { return; };` above.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+                if let Expression::FunctionExpression(function_expression) = &target.init {
+                    return get_function_identifier(function_expression)
+                        == get_function_identifier(function);
+                }
+            }
+
+            false
+        }
+        AstKind::AssignmentPattern(pattern) => {
+            if let BindingPatternKind::BindingIdentifier(_) = pattern.left.kind {
```
DonIsaac, line 142: These branches could be simplified.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+            }
+
+            false
+        }
+        _ => false,
+    }
+}
+
+fn get_function_identifier<'a>(func: &'a Function<'a>) -> Option<&'a Span> {
+    if let Some(id) = &func.id {
```
DonIsaac, line 172: Consider using `Option::map`

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+        return Some(&id.span);
+    }
+
+    None
+}
+
+fn get_function_name<'a>(func: &'a Function<'a>) -> Option<&Atom<'a>> {
+    if let Some(id) = &func.id {
```
DonIsaac, line 180: Here too.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+fn is_invalid_function(
+    func: &Function,
+    config: &FuncNamesConfig,
+    parent_node: Option<&AstNode>,
+) -> bool {
+    let func_name = get_function_name(func);
+
+    if
+    // never
+    (*config == FuncNamesConfig::Never
```
DonIsaac, line 196: Consider pattern matching over config.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+    Never,
+}
+
+impl FuncNamesConfig {
+    pub fn from(raw: &str) -> Self {
+        match raw {
+            "always" => FuncNamesConfig::Always,
+            "as-needed" => FuncNamesConfig::AsNeeded,
+            "never" => FuncNamesConfig::Never,
+            _ => FuncNamesConfig::default(),
```
DonIsaac, line 32: We should tell consumers when their configs are incorrect instead of failing silently.

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+        ("var a = new Date(function() {});", None),
+        ("var test = function(d, e, f) {};", None),
+        ("new function() {}", None),
+        ("Foo.prototype.bar = function() {};", Some(serde_json::json!(["as-needed"]))),
+        ("(function(){}())", Some(serde_json::json!(["as-needed"]))),
+        ("f(function(){})", Some(serde_json::json!(["as-needed"]))),
+        ("var a = new Date(function() {});", Some(serde_json::json!(["as-needed"]))),
+        ("new function() {}", Some(serde_json::json!(["as-needed"]))),
+        ("var {foo} = function(){};", Some(serde_json::json!(["as-needed"]))), // { "ecmaVersion": 6 },
+        // ("({key: foo = function(){}} = {});", Some(serde_json::json!(["as-needed"]))), // { "ecmaVersion": 6 },
```
DonIsaac, line 425: Does this need fixing?

crates/oxc_linter/src/rules/eslint/func_names.rs
```diff
@@ -0,0 +1,507 @@
...
+pub struct FuncNames {
+    default_config: FuncNamesConfig,
+    generators_config: FuncNamesConfig,
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+enum FuncNamesConfig {
+    #[default]
+    Always,
+    AsNeeded,
```
DonIsaac, line 22: I feel like this would make a more useful default, what do you think?

Sysix: From the offical docs:

>This rule has a string option:
> -  "always" (default) requires function expressions to have a name
> - "as-needed" requires function expressions to have a name, if the name isn’t assigned automatically per the ECMAScript specification.
> - "never" disallows named function expressions, except in recursive functions, where a name is needed

crates/oxc_linter/src/rules/eslint/func_names.rs
Sysix: nice catch, this is a passing test, I just copied rfom above to test something, will remove it :)

Sysix:
I added function name detection from `eslint/rules/ast-utiles/` so the warning messages are clearer :)

crates/oxc_linter/src/rules/eslint/func_names.rs
DonIsaac: I agree that this is the original implementation of the rule, but some of their defaults are rather strict and often get changed. I'm thinking of, for example, `eqeqeq`, where `smart` would make a better default. Do you have an opinion on actual usability?

crates/oxc_linter/src/rules/eslint/func_names.rs
DonIsaac: I agree they're doing this. This is a failing of existing configuration parsing. IMO, we've avoided panics in many cases where the user should be notified of misconfiguration.

An example, as requested: https://github.com/oxc-project/oxc/blob/6ff200d072fe55cb720c8dbe3a00f885fd1bbd28/crates/oxc_linter/src/rules/eslint/no_unused_vars/options.rs#L329

crates/oxc_linter/src/rules/eslint/func_names.rs
DonIsaac: https://github.com/oxc-project/oxc/blob/6ff200d072fe55cb720c8dbe3a00f885fd1bbd28/crates/oxc_linter/src/rules/eslint/no_unused_vars/mod.rs#L67

Notice how examples are given based on the configuration provided.

I plan on using these doc comments to build documentation pages on oxc.rs in the near future, fleshed out documentation will help with DX.

crates/oxc_linter/src/rules/eslint/func_names.rs
DonIsaac: By replacing the first line with, for example, "Leaving the name off a function will cause `<anonymous>` to appear in stack traces of errors thrown in it or any function called within it. This makes it more difficult to find where an error is thrown."

DonIsaac:
@Sysix thanks a ton for the PR, we're just about there!

crates/oxc_linter/src/rules/eslint/func_names.rs
Sysix: I don't have an opinion which options should be the default one. But I would prefer to be the same as the eslint one. 

Someone who has in the config just `"eslint/func-names": "error"` will get another result with oxlint then with eslint.

If I need to choose, I would choose `never` :')

crates/oxc_linter/src/rules/eslint/func_names.rs
Sysix: I updated the doc, feel free to modify it as needed :)

crates/oxc_linter/src/rules/eslint/func_names.rs
Sysix: I don't really understand what the expecting result will be.
Here it what is does now:

>failures:
> ---- rules::eslint::func_names::test stdout ----
> thread 'rules::eslint::func_names::test' panicked at crates/oxc_linter/src/rules/eslint/func_names.rs:376:71:
called `Result::unwrap()` on an `Err` value: OxcDiagnostic { inner: OxcDiagnosticInner { message: "Expecting always, as-needed or never, got as-needed2", labels: None, help: None, severity: Error, code: OxcCode { scope: None, number: None } } }

Sysix:
@DonIsaac I implemented the requested changes. Have a look again :) 

Also I check if the `CallExpression` is inside the `invalid_funcs` vec, so we can skip the parent node traveling when the name doesn't match

DonIsaac (approved):

PR #4621, opened 2024-08-03
refactor(ast): put `assert_layouts.rs` behind `debug_assertions`

rzvxa:
#4615 was showing as green on all checks, but failed CI on main once merged because of errors in "Test wasm32-wasip1-threads" job, which only runs on main. This change will make sure any such problem gets caught at the PR stage in future.

### DIFF
--- a/crates/oxc_ast/src/lib.rs
+++ b/crates/oxc_ast/src/lib.rs
@@ -37,7 +37,7 @@ pub mod syntax_directed_operations;
 mod trivia;
 
 mod generated {
-    #[cfg(test)]
+    #[cfg(debug_assertions)]
     pub mod assert_layouts;
     pub mod ast_builder;
     pub mod ast_kind;


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4621?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4621** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4620** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4620?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
@rzvxa Does this definitely solve the problem? Don't look like WASM is tested at all on PRs.

rzvxa:
> @rzvxa Does this definitely solve the problem? Don't look like WASM is tested at all on PRs.

We used to have one, It is changed in #4547.

What if we create a new task that would run wasm tests only if not running on "main"?

overlookmotel:
Maybe we should just remove/comment out the 32-bit layout assertions until we solve this? We don't really need them at present anyway. It's not ideal to run risk turning CI red by accident. Could make life difficult when trying to git bisect a bug later on.

rzvxa:
> Maybe we should just remove/comment out the 32-bit layout assertions until we solve this? We don't really need them at present anyway. It's not ideal to run risk of turning CI red by accident. Could make life difficult when trying to git bisect a bug later on.

Or we can remove the condition of `check wasm-unknown-unknown`, We should be fine if we run a cargo check on the wasm.
This used to be unconditional after we got rid of `ci-optimize`. so there shouldn't be an issue if we make it unconditional once again.

```yaml
  test-wasm32-unknown-unknown:
    name: Check wasm32-unknown-unknown
    # if: ${{ github.ref_name == 'main' }}
    runs-on: ubuntu-latest
`

overlookmotel (approved):

overlookmotel:
> Or we can remove the condition of `check wasm-unknown-unknown`

Agreed. Let's merge this PR and do that in a follow-up.

PR #4625, opened 2024-08-03
refactor(minifier): ast passes infrastructure

Boshen:
After studying [google closure compiler](https://github.com/google/closure-compiler) for several days, I'm leaning towards a multi-ast-pass infrastructure for the minifier.

This is one of the few places where we are going to trade maintainability over performance, given the goal of the minifier is compression size not performance.

All of the terminologies and separation of concerns are aligned with google closure compiler.
 
Infrastructure of `terser` and `esbuild` are not suitable for us to study nor pursuit. Their code are so tightly coupled - I failed to comprehend any of them every time I try to walk through a piece of optimization. Google closure compiler despite being written in Java, is actually the most readable minifier out there.

To improve performance between ast passes, I envision a change detection system over a portion of the code.

Each AST pass adds around 5 - 10% performance regression, which I think is an acceptable trade off. The baseline is the minifier should be able to minify huge files. Individual AST pass can be turned off if the performance is not acceptable.

I also see a future where some of the AST passes are moved to dev mode in Rolldown, eliminating more dev-prod inconsistencies.

Multiple-ast-passes infrastructure can also welcome and facilitate more contributors to learn and contribute.


crates/oxc_minifier/src/compressor.rs
```diff
@@ -0,0 +1,47 @@
...
+        if self.options.substitute_alternate_syntax {
+            SubstituteAlternateSyntax::new(ast, self.options).build(program);
+        }
+        if self.options.remove_dead_code {
+            RemoveDeadCode::new(ast).build(program);
+        }
+        if self.options.collapse {
+            Collapse::new(ast, self.options).build(program);
+        }
+    }
```
Boshen, line 39: Here are the 5 AST passes.

tasks/coverage/minifier_test262.snap
```diff
@@ -2,4 +2,22 @@ commit: a1587416
 
 minifier_test262 Summary:
 AST Parsed     : 46406/46406 (100.00%)
-Positive Passed: 46406/46406 (100.00%)
+Positive Passed: 46388/46406 (99.96%)
+Expect to Parse: "language/expressions/logical-and/S11.11.1_A3_T4.js"
```
Boshen, line 6: Tests are broken due to not having a fixed point.

Boshen:
I'm going to do a local wall-clock measurement, I doubt it's going to be noticeable by human.

Boshen:
This is in a good shape to be iterated upon. So merging.

PR #4640, opened 2024-08-05
feat(website): auto-generate rule docs pages

DonIsaac:
> AI-generated description because I'm lazy
### TL;DR

This PR introduces the ability to generate documentation for linter rules and adds new methods and metadata for rule fix capabilities.

To see what this looks like, please check out https://github.com/oxc-project/oxc-project.github.io/pull/165.

## Screenshots
Hyperlinks to rule doc pages in auto-generated rules table
image

Example of a docs page
image

### What changed?

- Added `RuleFixMeta` to indicate rule fix capabilities
- Introduced methods `is_none` and `is_pending` in `RuleFixMeta`
- Modified `render_markdown_table` in `RuleTableSection` to accept an optional link prefix
- Created new modules for rule documentation and HTML rendering
- Updated `print_rules` function to generate markdown for rules and detailed documentation pages

### How to test?

Run the `linter-rules` task with appropriate arguments to generate the markdown table and documentation pages.
Verify the generated files for correctness and that all metadata is correctly displayed.

### Why make this change?

To enhance the project documentation and provide clear rule fix capabilities, thereby improving the developer experience and easing the integration process.

---

PR #4646, opened 2024-08-05
fix(linter): invalid tags in rule docs

DonIsaac:
These are causing build errors in doc pages in PR #4640


DonIsaac:
Trivial change, so I'll just merge it

PR #3813, opened 2024-06-21
chore(ci): now we mark both ts alias and eslint rules.

rzvxa:
Fixes the unmarked implemented rules in #479.

PR #4654, opened 2024-08-05
fix(tasks/lint_rules): Refactor syncing typescript-eslint and eslint status

### LINKED ISSUE #4085: Some of the rules should be marked as completed in eslint core rules

leaysgur:
Closes #4085 

This issue seemed to have been addressed in #3779 , but partially reverted in #3813 ? 🤔 

Since I wasn't aware of these changes, I've just checked the current implementation through the review requests in #4611 and refactored as the original author.

### DIFF
--- a/tasks/lint_rules/src/eslint-rules.cjs
+++ b/tasks/lint_rules/src/eslint-rules.cjs
@@ -263,7 +263,3 @@ exports.loadTargetPluginRules = (linter) => {
   loadPluginReactPerfRules(linter);
   loadPluginNextRules(linter);
 };
-
-// some typescript rules are some extension of the basic eslint rules
-// we need them later to map them for both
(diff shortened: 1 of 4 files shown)


leaysgur:
The no-new-symbol rule is already deprecated in ESLint v9 and is available as latest in npm. 

However, some modules in our dependency graph still require v8, so it will remain in the recommended category for a while.

PR #4581, opened 2024-07-31
feat(transformer_conformance): test `ScopeTree` and `SymbolTable` for correctness after transformation

Dunqing:
part of #4549 

This test verifies that the correct symbols and scopes were added/changed during the transformation. From the snapshots, most of them are incorrect currently

PR #4658, opened 2024-08-05
fix(transformer/react): the `require` IdentifierReference does not have a `reference_id`

Dunqing:
part of #4581. Looks we had to pass `TraverseCtx` to the `ModuleImports`

### DIFF
--- a/crates/oxc_transformer/src/helpers/module_imports.rs
+++ b/crates/oxc_transformer/src/helpers/module_imports.rs
@@ -3,8 +3,10 @@ use std::cell::{Cell, RefCell};
 use indexmap::IndexMap;
 use oxc_allocator::{Allocator, Vec};
 use oxc_ast::{ast::*, AstBuilder};
+use oxc_semantic::ReferenceFlag;
 use oxc_span::{Atom, SPAN};
 use oxc_syntax::symbol::SymbolId;
+use oxc_traverse::TraverseCtx;
 
 pub struct NamedImport<'a> {
     imported: Atom<'a>,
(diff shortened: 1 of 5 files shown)


Dunqing:
* **#4658** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/helpers/module_imports.rs
```diff
@@ -116,9 +118,13 @@ impl<'a> ModuleImports<'a> {
         &self,
         source: Atom<'a>,
         names: std::vec::Vec<NamedImport<'a>>,
+        ctx: &mut TraverseCtx<'a>,
     ) -> Statement<'a> {
         let var_kind = VariableDeclarationKind::Var;
-        let callee = self.ast.expression_identifier_reference(SPAN, Atom::from("require"));
+        let ident =
+            ctx.create_reference_id(SPAN, Atom::from("require"), None, ReferenceFlag::read());
```
overlookmotel, line 125: Strictly speaking, we should also handle if there's a binding in root scope for a var called `require` e.g. `let require = 123;`. This would create output that doesn't work properly, but at least scopes tree will be correct!

```suggestion
        let symbol_id = ctx.scopes().get_root_binding("require");
        let ident =
            ctx.create_reference_id(SPAN, Atom::from("require"), symbol_id, ReferenceFlag::read());
```

overlookmotel:
Ah ha! You found my mistake!

crates/oxc_transformer/src/helpers/module_imports.rs
Dunqing: Good catch!

overlookmotel (approved):

PR #4664, opened 2024-08-05
refactor(semantic): simplify inherit scope flags from parent scope

### LINKED ISSUE #3862: Inherit `ScopeFlags::StrictMode` from parent scope
There appear to be some discrepancies in when `ScopeFlags::StrictMode` is applied by semantic.

`enter_scope` *looks like* its logic is that `StrictMode` should be set on any scope which is strict mode (including from a `'use strict'` directive somewhere above it). But this doesn't appear to be what it actually does, because it only sets `StrictMode` flag if the *direct* parent is a function with a `'use strict'` directive, or if the entire program is strict mode:

https://github.com/oxc-project/oxc/blob/9e148e97a95f55d49e5976524242d9b243e0c72c/crates/oxc_semantic/src/builder.rs#L414-L440

A couple of examples of weirdness:

```js
// Script - sloppy mode
function foo() {
  'use strict';
  {
    // Scope flags for this block include `StrictMode`
    {
      // Scope flags for this block do NOT include `StrictMode`
    }
  }
}
```

```js
// Script - sloppy mode
class X {
  foo() {
    // Scope flags for this method do NOT include `StrictMode`
  }
}
```

When is `StrictMode` flag *meant* to be set?

If intended behavior is that `StrictMode` flag should be set for any scope which runs in strict context, we can simplify the above logic to:

```rs
if let Some(parent_scope_id) = parent_scope_id {
    let parent_scope_flags = self.scope.get_flags(parent_scope_id);
    flags |= parent_scope_flags & ScopeFlags::StrictMode;

    // inherit flags for non-function scopes
    if !flags.contains(ScopeFlags::Function) {
        flags |= parent_scope_flags & ScopeFlags::Modifiers;
    }
}
```

Dunqing:
close: #3862

This is reaming last part of close #3862

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -510,10 +510,9 @@ impl<'a> SemanticBuilder<'a> {
 
 impl<'a> Visit<'a> for SemanticBuilder<'a> {
     // NB: Not called for `Program`
-    fn enter_scope(&mut self, flags: ScopeFlags, scope_id: &Cell<Option<ScopeId>>) {
+    fn enter_scope(&mut self, mut flags: ScopeFlags, scope_id: &Cell<Option<ScopeId>>) {
         let parent_scope_id = self.current_scope_id;
 
-        let mut flags = flags;
         if !flags.is_strict_mode() && self.current_node_flags.has_class() {
             // NOTE A class definition is always strict mode code.
             flags |= ScopeFlags::StrictMode;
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#4664** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):

PR #4665, opened 2024-08-05
fix(linter): fix false positives in no-confusing-non-null-assertion

goto-bus-stop:
This lint did not check which operator was used, so it would warn you that `a! + b` looks too much like `a !+ b`, which doesn't make much sense.
```
  ! typescript-eslint(no-confusing-non-null-assertion): Confusing combinations of non-null assertion and equal test like "a! + b", which looks very similar to not equal "a !+ b".
     ,-[util.ts:115:19]
 114 |     coordEach(wrapped, (coord) => {
 115 |       coord[0] = (coord[0]! + 360) % 360
     :                   ^^^^^^^^^^^^^^^
 116 |     })
     `----
  help: Remove the "!", or prefix the "=" with it.
```

This patch adds tests for various binary operators and restricts the lint to only those that actually look like `!=`.


Boshen:
I recognize your OG name! Thank you for contributing!

DonIsaac (approved):

PR #4671, opened 2024-08-05
build(deps): compile insta in release mode

DonIsaac:
Compile insta and its diffing library in release mode. After the initial build, I'm seeing ~1s shaved off `cargo test` locally.


Boshen:
NICE! I can't believe that we have so much tests that this is taking into effect.

PR #4168, opened 2024-07-10
feat(ast, ast_codegen): pass the `scope_id` to the `enter_scope` event.

rzvxa:

PR #4688, opened 2024-08-06
refactor(traverse): simplify `ChildScopeCollector` to use `enter_scope` method

Dunqing:
related: https://github.com/oxc-project/oxc/pull/4168

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -2,10 +2,7 @@ use std::{cell::Cell, str};
 
 use compact_str::{format_compact, CompactString};
 #[allow(clippy::wildcard_imports)]
-use oxc_ast::{
-    ast::*,
-    visit::{walk, Visit},
-};
+use oxc_ast::{ast::*, visit::Visit};
 use oxc_semantic::{AstNodeId, Reference, ScopeTree, SymbolTable};
 use oxc_span::{Atom, CompactStr, Span, SPAN};
 use oxc_syntax::{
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#4688** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
The point of `ChildScopeCollector` is to get *direct* child scopes only. So each of the visitors records the `ScopeId` but then also does *not* call `walk_*` to continue traversal.

`ChildScopeCollector` is used when inserting a scope into middle of scope chain. We need only the direct child scopes, in order to alter their parent. More deeply nested scopes do not change their parent, so should not be collected.

This PR changes that behavior, which isn't correct.

overlookmotel:
The point of `ChildScopeCollector` is to get *direct* child scopes only. So each of the visitors records the `ScopeId` but then also does *not* call `walk_*`, so they don't continue traversing deeper.

`ChildScopeCollector` is used when inserting a scope into middle of scope chain. We need only the direct child scopes, in order to alter their parent. More deeply nested scopes do not change their parent, so should not be collected.

This PR changes that behavior, which isn't correct.

Dunqing:
> The point of `ChildScopeCollector` is to get _direct_ child scopes only. So each of the visitors records the `ScopeId` but then also does _not_ call `walk_*`, so they don't continue traversing deeper.
> 
> `ChildScopeCollector` is used when inserting a scope into middle of scope chain. We need only the direct child scopes, in order to alter their parent. More deeply nested scopes do not change their parent, so should not be collected.
> 
> This PR changes that behavior, which isn't correct.

Ah, I see, Is it possible to add a test to cover this incorrect behavior?

overlookmotel:
> Is it possible to add a test to cover this incorrect behavior?

Well I'm hoping that as we add more test coverage to transformer, it'll get covered through imported test cases.

PR #4694, opened 2024-08-06
feat(ast_codegen): process AST-related `syntax` types.

rzvxa:

PR #4689, opened 2024-08-06
refactor(ast): add explicit enum discriminants.

rzvxa:
Related to oxc-project/backlog#123 but doesn't close it.

PR #4682, opened 2024-08-06
refactor(ast_codegen): move away from `RType` in generators.

rzvxa:
This PR introduces `EarlyCtx` and `LateCtx` in place of the old `CodegenCtx`, Early passes operate at the AST level while generators and other late passes operate on the schema.

It will also replace the confusing `RType` name with something more idiomatic ~~(open for suggestions, I haven't found a good name yet)~~ I've named it `AstType` and dropped the `R` prefix for `REnum` and `RStruct`.

There are some qualities of life improvements too, Things like `to_type_elide` can be used to simplify the code.

Related to #4442 (and can potentially mark it "close as fixed").

PR #4696, opened 2024-08-06
feat(ast_codegen, span): process `SourceType` through ast_codegen.

rzvxa:


rzvxa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4696?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4696** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4696?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4694** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4694?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4690** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4690?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4682** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4682?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4691** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4691?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4689** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rzvxa:
@overlookmotel this stack should be ready for review.

overlookmotel (commented):
Great to have these types included in schema.

Personally I wouldn't have split the types into a separate file. Is the rationale for that to give the codegen less code to chew through looking for items with an `#[ast]` attr? If so, I wouldn't worry too much about that - the file is not so large, and it's no big deal if the codegen isn't quite as quick as it could be.

It's annoying not to be able to assert that field offsets are correct because the fields are private. Perhaps at some point we should add a `generated` folder to `oxc_syntax` and `oxc_span` crates, and generate the assertions there so they can access the private fields. We'd need to do that for `CloneIn` and `Serialize` traits if we want to codegen their impls for `SourceType`.

overlookmotel:
One last thing: Since `oxc_span` crate now depends on `oxc_ast_macros`, should we put `#[ast]` on `Span` as well?

rzvxa:
> Great to have these types included in schema.
> 
> Personally I wouldn't have split the types into a separate file. Is the rationale for that to give the codegen less code to chew through looking for items with an `#[ast]` attr? If so, I wouldn't worry too much about that - the file is not so large, and it's no big deal if the codegen isn't quite as quick as it could be.

I intentionally panic on unrelated items in the source read by codegen. It isn't for performance purposes, I don't care for performance in the codegen.

It lowers the clutter in the rust definition files, I would even suggest going as far as changing the file extension to something like `XXX.d.rs`. Otherwise, we have no way of reasoning about what is and isn't included in the codegen(unless you read the source code).

> It's annoying not to be able to assert that field offsets are correct because the fields are private. Perhaps at some point we should add a `generated` folder to `oxc_syntax` and `oxc_span` crates, and generate the assertions there so they can access the private fields. We'd need to do that for `CloneIn` and `Serialize` traits if we want to codegen their impls for `SourceType`.

Yeah, Currently I'm working on the `generate_derive` marker and the `CloneIn` trait. One limitation of our current approach is that we output one file per generator. That means we get only one output `derive_clone_in.rs` file.

We used to have multiple outputs from a single generator but the implementation was no good. We need a way to have one generator per crate or one output per type and fold it into crate/module boundaries.

We might not even need such a thing. In general, I prefer we don't invest time on cross-crate compilation, We shouldn't expand codegen far out of AST. Otherwise, we start generating code all over the place and bloat the project.

I can extract the pre-expanded derives as its own thing if we see a reasonable improvement over vanilla derives.

rzvxa:
> One last thing: Since oxc_span crate now depends on oxc_ast_macros, should we put #[ast] on Span as well?

Sure.

overlookmotel (approved):

overlookmotel:
> Otherwise, we have no way of reasoning about what is and isn't included in the codegen(unless you read the source code).

Personally, I think it's fine to just have `#[ast]` to inform the reader what it being altered by the codegen. I also like your idea of using `#[generate_derive(CloneIn)]` - again, that makes it explicit that some code is being generated for the type, rather than it just happening invisibly.

rzvxa:
> > Otherwise, we have no way of reasoning about what is and isn't included in the codegen(unless you read the source code).
> 
> Personally, I think it's fine to just have `#[ast]` to inform the reader what it being altered by the codegen. I also like your idea of using `#[generate_derive(CloneIn)]` - again, that makes it explicit that some code is being generated for the type, rather than it just happening invisibly.

The problem is when you use `#[ast]` on something expecting it to work similarly to all places you've seen this marker, But it doesn't; The `#[ast]` itself isn't the whole picture. we don't know if it is even included in the codegen or not as long as we haven't checked the main file of the codegen.

In the `#[ast]` macro we can panic if the file isn't already processed by the codegen, which just complicates the compilation, However, it is possible to use it to raise compile errors on illegal usage.

> I'll make a PR to add `#[ast]` to `Span`.

Thank you!

overlookmotel:
> In the #[ast] macro we can panic if the file isn't already processed by the codegen, which just complicates the compilation, However, it is possible to use it to raise compile errors on illegal usage.

I think this would be a good idea. Opened #4704 to discuss.

PR #4697, opened 2024-08-06
perf(linter): add Set

DonIsaac:
Many rules store a `Vec<String>` in their configs for things like ignore lists or JSX tags that should be checked. They are created once (in `from_configuration`) and mostly used for `Vec::contains`. There are a few problems with this:

1. `String`s are always heap-allocated. We should inline when possible, using `CompactStr`.
2. `Vec::contains` is `O(n)`, and is usually used in a hot path.
3. `Vec::contains` requires an `&T`, meaning there are many cases when rules have an `&str` and need to allocate a new `String` just to pass it to contains. This is highly wasteful.
4. Many rules have duplicate logic for parsing a `Vec<String>` from a `Value`.

This PR adds `Set<T>`, which seeks to make this use case more performant. It's very simple ordered set of unique items backed by a `Vec`. Read the doc comments over the struct definition for a more in-depth rationale.

## Improvements
1. `Set<T>` supports `O(log(n))` containment checks at the cost of `O(nlog(n))` initialization. This is fine since we create it once and read from it many times.
2. `Set<T> where T: AsRef<str>>` supports a `contains_str` method that is identical to `contains` but does not force consumers to allocate new strings
3. `Set<CompactStr>` implements `TryFrom<Value>`, which DRYs up rule config parsing. This trait is intentionally not implemented for `Set<String>` to nudge rules towards using `CompactStr`.

Following PRs will modify existing rules to use `Set<CompactStr>` over `Vec<String>`. Since most of these cases are within Jest and VITest rules, I don't expect improvements to appear on benchmarks (rules are skipped when non-applicable, and we have no benchmarks for Jest specs).

### DIFF
--- a/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/mod.rs
+++ b/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/mod.rs
@@ -1,14 +1,14 @@
 use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
+use oxc_span::{CompactStr, Span};
 use serde_json::Value;
 
 use self::listener_map::ListenerMap;
 use crate::{
     context::LintContext,
     rule::Rule,
-    utils::{ModuleFunctions, NodeListenerOptions, WhitelistModule},
+    utils::{ModuleFunctions, NodeListenerOptions, Set, WhitelistModule},
 };
 
 mod listener_map;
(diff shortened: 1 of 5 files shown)


DonIsaac:
* **#4697** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (commented):
I think we should avoid our own data structures for the sake of maintainability and discoverability.

This will make it difficult for contributors to learn - remember most contributors come from JavaScript, we should lower the barrier.

As for optimization, perhaps replacing `String` for `CompactStr` is more than enough. Hot spots can also use `smallvec`. But reading from the heap where the vec is small won't be a performance hot spot because they'll likely live in a nearby cache.

PR #4703, opened 2024-08-06
feat(ast_codegen, span): process `Span` through ast_codegen

overlookmotel:
Add `#[ast]` attr to `Span`.

Due to how AST codegen works, this necessitates putting the type def in a separate file. https://github.com/oxc-project/oxc/pull/4696#issuecomment-2271781995


overlookmotel:
* **#4703** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
It actually wouldn't panic for this specific file, because it doesn't include any other enum/struct definitions. I don't panic on `impl` items, Just on unmarked enum/structs.

But It is better to keep the convention, Even if we want to merge impls and definitions it can be done in one go.

PR #4725, opened 2024-08-07
refactor(minifier): use `oxc_traverse` for AST passes

Boshen:


crates/oxc_minifier/src/ast_passes/substitute_alternate_syntax.rs
```diff
@@ -15,39 +16,48 @@ pub struct SubstituteAlternateSyntax<'a> {
...
+    ) {
         for declarator in decl.declarations.iter_mut() {
-            self.visit_variable_declarator(declarator);
             Self::compress_variable_declarator(declarator);
         }
     }
 
-    fn visit_expression(&mut self, expr: &mut Expression<'a>) {
+    fn enter_expression(&mut self, expr: &mut Expression<'a>, _ctx: &mut TraverseCtx<'a>) {
         // Bail cjs `Object.defineProperty(exports, ...)`
```
Boshen, line 47: @overlookmotel How do I skip visitation?

It needs to bail for this test:

```
#[test]
fn cjs() {
    // Export is undefined when `enumerable` is "!0".
    // https://github.com/nodejs/cjs-module-lexer/issues/64
    test_same(
        "Object.defineProperty(exports, 'ConnectableObservable', {
          enumerable: true,
          get: function() {
            return ConnectableObservable_1.ConnectableObservable;
          }
        });",
    );
}
```

crates/oxc_minifier/src/compressor.rs
```diff
@@ -1,62 +1,80 @@
...
+        self.build_with_symbols_and_scopes(symbols, scopes, program);
+    }
+
+    pub fn build_with_symbols_and_scopes(
+        self,
+        symbols: SymbolTable,
+        scopes: ScopeTree,
+        program: &mut Program<'a>,
+    ) {
+        let mut ctx = TraverseCtx::new(scopes, symbols, self.allocator);
```
Boshen, line 40: The goal is to reuse `ctx` between AST passes, implementation needs to keep symbols and scopes in sync.

overlookmotel: I'll look into how we can do this.

Boshen:
I don't know what I'm doing, so :shipit:

Issue #4284, opened 2024-06-21
`CloneIn` trait for AST nodes


overlookmotel:
We need to be able to clone AST nodes. But we cannot implement `Clone` on most AST types because `oxc_allocator::Box` is not `Clone`.

There is a good reason for that - `Box` stores the value it owns in arena allocator, so there is no `Box::new` method, only `Box::new_in` which takes an `&Allocator` param for the arena to allocate it in.

I propose a new trait `CloneIn`:

```rs
trait CloneIn<'alloc> {
    fn clone_in(&self, allocator: &'alloc Allocator) -> Self;
}
```

Impl for `oxc_allocator::Box`:

```rs
impl<'alloc, T> CloneIn<'alloc> for Box<'alloc, T>
where
    T: CloneIn<'alloc>,
{
    fn clone_in(&self, allocator: &'alloc Allocator) -> Self {
        Box::new_in(self.deref().clone_in(allocator), allocator)
    }
}
```

(I'm actually not sure if I have this right - maybe `'alloc` lifetime should be on `fn clone_in<'alloc>` not `CloneIn<'alloc>`)

Same as `Clone`, `CloneIn` can be implemented on any type where all its fields/variants are also `CloneIn`.

We could make a derive macro for `CloneIn` so can just `#[derive(CloneIn)]` on AST types.

NB: `clone_in` should always clone into the arena of the provided `allocator`, which *may be different* from the arena that `self` is in. i.e. you can use `clone_in` to copy objects from one arena to another.

rzvxa:
I genuinely burst out laughing when I saw `MyTrousers`😆

hyf0:
What's the blocker here? It seems the final form has been shown and there are no technical issues.

hyf0:
Would it be more performant if there something like `size_hint` in `CloneIn` trait to infer allocator?

overlookmotel:
> What's the blocker here? It seems the final form has been shown and there are no technical issues.

No blocker. We've just been working on other things. Is your need for it urgent?

> Would it be more performant if there something like `size_hint` in `CloneIn` trait to infer allocator?

Sorry I don't get you. How do you mean?

hyf0:
> No blocker. We've just been working on other things. Is your need for it urgent?

Not sure how define urgent. But people are writing code like this

[...]

PR #4726, opened 2024-08-07
feat(allocator): introduce `CloneIn` trait.

rzvxa:
Introduce the trait discussed in #4284.

### DIFF
--- a/crates/oxc_allocator/src/clone_in.rs
+++ b/crates/oxc_allocator/src/clone_in.rs
@@ -1,0 +1,91 @@
+use std::cell::Cell;
+
+use crate::{Allocator, Box, Vec};
+
+/// A trait to explicitly clone an object into an arena allocator.
+///
+/// As a convention `Cloned` associated type should always be the same as `Self`,
+/// It'd only differ in the lifetime, Here's an example:
+///
+/// ```
+/// impl<'old_alloc, 'new_alloc> CloneIn<'new_alloc> for Struct<'old_alloc> {
+///     type Cloned = Struct<'new_alloc>;
+///     fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned {
+///         Struct { a: self.a.clone_in(alloc), b: self.b.clone_in(alloc) }
+///     }
+/// }
+/// ```
+///
+/// Implementations of this trait on non-allocated items usually short-circuit to `Clone::clone`;
+/// However, it **isn't** guaranteed.
+///
+pub trait CloneIn<'new_alloc>: Sized {
+    type Cloned;
+
+    fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned;
+}
+
+impl<'alloc, T, C> CloneIn<'alloc> for Option<T>
+where
+    T: CloneIn<'alloc, Cloned = C>,
+{
+    type Cloned = Option<C>;
+    fn clone_in(&self, alloc: &'alloc Allocator) -> Self::Cloned {
+        self.as_ref().map(|it| it.clone_in(alloc))
+    }
+}
+
+impl<'old_alloc, 'new_alloc, T, C> CloneIn<'new_alloc> for Box<'old_alloc, T>
+where
+    T: CloneIn<'new_alloc, Cloned = C>,
+{
+    type Cloned = Box<'new_alloc, C>;
+    fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned {
+        Box::new_in(self.as_ref().clone_in(alloc), alloc)
+    }
+}
+
+impl<'old_alloc, 'new_alloc, T, C> CloneIn<'new_alloc> for Vec<'old_alloc, T>
+where
+    T: CloneIn<'new_alloc, Cloned = C>,
+{
+    type Cloned = Vec<'new_alloc, C>;
+    fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned {
+        Vec::from_iter_in(self.iter().map(|it| it.clone_in(alloc)), alloc)
+    }
+}
+
+impl<'alloc, T: Copy> CloneIn<'alloc> for Cell<T> {
+    type Cloned = Cell<T>;
+    fn clone_in(&self, _: &'alloc Allocator) -> Self::Cloned {
+        Cell::new(self.get())
+    }
+}
+
+impl<'old_alloc, 'new_alloc> CloneIn<'new_alloc> for &'old_alloc str {
+    type Cloned = &'new_alloc str;
+    fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned {
+        alloc.alloc_str(self)
+    }
+}
+
+macro_rules! impl_clone_in {
+    ($($t:ty)*) => {
+        $(
+            impl<'alloc> CloneIn<'alloc> for $t {
+                type Cloned = Self;
+                #[inline(always)]
+                fn clone_in(&self, _: &'alloc Allocator) -> Self {
+                    *self
+                }
+            }
+        )*
+    }
+}
+
+impl_clone_in! {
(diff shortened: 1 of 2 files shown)


rzvxa:
* **#4726** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4726?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl<'a, T, C> CloneIn<'a> for Option<T>
+where
+    T: CloneIn<'a, Cloned = C>,
+{
+    type Cloned = Option<C>;
+    fn clone_in(&self, alloc: &'a Allocator) -> Self::Cloned {
+        self.as_ref().map(|it| it.clone_in(alloc))
+    }
+}
```
overlookmotel, line 34: Can we use `'new_alloc` here instead of `'a`? I know there's only 1 lifetime on this impl, but as you've pointed out, this trait is somewhat confusing, so I think it's better to be explicit.

Ditto for `Cell` impl, and `impl_clone_in!`.

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,89 @@
...
+        Self::Cloned::from_iter_in(self.iter().map(|it| it.clone_in(alloc)), alloc)
+    }
+}
+
+impl<'a, T: Copy> CloneIn<'a> for std::cell::Cell<T> {
+    type Cloned = std::cell::Cell<T>;
+    fn clone_in(&self, _: &'a Allocator) -> Self::Cloned {
+        Self::Cloned::new(self.get())
+    }
+}
```
overlookmotel, line 61: I have a slight preference to import `std::cell::Cell` at top of the file, and use just `Cell` here, as it reduces noise a little. But that's a matter of taste - feel free to ignore me!

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,89 @@
...
+        )*
+    }
+}
+
+impl_clone_in! {
+    usize u8 u16 u32 u64 u128
+    isize i8 i16 i32 i64 i128
+    f32 f64
+    bool char
+}
```
overlookmotel, line 89: Could we / should we do a blanket impl for all `Clone` types instead? Something like:

```rs
impl<'new_alloc, T> CloneIn<'new_alloc> for T
where T: Clone
{
    type Cloned = Self;
    #[inline]
    fn clone_in(&self, _: &'a Allocator) -> Self {
        self.clone()
    }
}
```

I think `Clone` may be optimized by compiler for `Copy` types (but I may be wrong about that).

crates/oxc_allocator/src/clone_in.rs
rzvxa: Higher up the stack I've used `'alloc` for implementations with only one allocator lifetime. Do you think it is sufficient or should I go ahead and refactor them all?

https://github.com/oxc-project/oxc/blob/383afeaff1bcd3bd4ef4170d5a0823180d79a928/crates/oxc_ast/src/generated/derive_clone_in.rs#L14

crates/oxc_allocator/src/clone_in.rs
overlookmotel: Actually no, that's a bad idea. `Atom<'a>` is `Clone`, but it should not be just `Clone`-d.

And we also might want to make `Atom<'a>` `Copy` in future, so we shouldn't do a blanket impl `impl<'a, T: Copy> CloneIn<'a> for T` either.

rzvxa: I wasn't sure if it was safe or not, This implementation for builtins is similar to how std defines them.
I'll test it with the blanket implementation, If there aren't any conflicting implementations I'll refactor the stack. I'll mark this as draft to prevent an accidental merge.

rzvxa: It doesn't work, It breaks implementations for `Atom`, `&'a str`, `Box<'a, T: CloneIn>`, and `Option<T: CloneIn>`.

rzvxa: > Actually no, that's a bad idea. `Atom<'a>` is `Clone`, but it should not be just `Clone`-d.
> 
> And we also might want to make `Atom<'a>` `Copy` in future, so we shouldn't do a blanket impl `impl<'a, T: Copy> CloneIn<'a> for T` either.

I totally missed this comment😅

crates/oxc_allocator/src/clone_in.rs
overlookmotel: Personally I think it's preferable to use `'new_alloc` here so it matches the rest of the impls in this file. But I can see argument for `'alloc` for consistency. Up to you.

crates/oxc_allocator/src/clone_in.rs
overlookmotel: > I totally missed this comment😅

You're too quick!

overlookmotel (approved):

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+impl<'old_alloc, 'new_alloc, T, C> CloneIn<'new_alloc> for Vec<'old_alloc, T>
+where
+    T: CloneIn<'new_alloc, Cloned = C>,
+{
+    type Cloned = Vec<'new_alloc, C>;
+    fn clone_in(&self, alloc: &'new_alloc Allocator) -> Self::Cloned {
+        Self::Cloned::from_iter_in(self.iter().map(|it| it.clone_in(alloc)), alloc)
```
overlookmotel, line 54:
```suggestion
        Vec::from_iter_in(self.iter().map(|it| it.clone_in(alloc)), alloc)
```

or

```suggestion
        Self::from_iter_in(self.iter().map(|it| it.clone_in(alloc)), alloc)
```

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+impl<'alloc, T: Copy> CloneIn<'alloc> for Cell<T> {
+    type Cloned = Cell<T>;
+    fn clone_in(&self, _: &'alloc Allocator) -> Self::Cloned {
+        Self::Cloned::new(self.get())
```
overlookmotel, line 61:
```suggestion
        Cell::new(self.get())
```

overlookmotel (changes requested):

crates/oxc_allocator/src/clone_in.rs
overlookmotel: Wait a second... Why did it break `Box<'a, T: CloneIn>`? `Box` isn't `Clone` is it? It shouldn't be!

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+impl<'old_alloc, 'new_alloc> CloneIn<'new_alloc> for &'old_alloc str {
+    type Cloned = &'new_alloc mut str;
```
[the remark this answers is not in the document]
rzvxa: This was intentional, My reason is that if we don't clone from `&str` to `&mut str` there is no way to actually clone something and get exclusive ownership of it.

Cloning the reference isn't what we mean when we are cloning a `&'alloc str` type. The pointer itself is copy so we are good on that side.
Let me see if I can fix it by providing a `CloneIn` implementation for `str` as opposed to the slice str.

crates/oxc_allocator/src/clone_in.rs
rzvxa: No, You are right box doesn't break. That was my mistake.

overlookmotel (approved):

Issue #4739, opened 2024-08-07
Incorrect `Span` on `BindingIdentifier` 


DonIsaac:
Spans for binding identifiers cover the identifier itself and the type annotation on `BindingPattern`. This was introduced in #2624, and is blocking https://github.com/oxc-project/backlog/issues/84.

## Expected Behavior
```ts
//    ___________ BindingPattern
const foo: number = 1
//    ^^^-------- TSTypeAnnotation
//    ^^^BindingIdentifier
```

## Actual Behavior
```ts
//    ___________ BindingPattern and BindingIdentifier
const foo: number = 1
//       -------- TSTypeAnnotation
```

Boshen:
This is a weird one ...

<img width="778" alt="image" src="https://github.com/user-attachments/assets/d893e5ce-0d28-4891-adf1-5068c12d093b">

Dunqing:
This is also weird in `TypeScript`, Why not `{ pos: 6, end: 7 }`?
<img width="1470" alt="image" src="https://github.com/user-attachments/assets/090bbe46-e6e5-4c32-9e42-7def62e38b76">

Boshen:
Shall we go for the more intuitive version? i.e. 

//    ___________ BindingPattern
const foo: number = 1
//       ^^^^^^^^ TSTypeAnnotation
//    ^^^BindingIdentifier

DonIsaac:
The cause is BindingPattern gets its span from `kind`, which covers the whole BindingPattern. I see two possible solutions:

1) give BindingPattern its own `span` field, which would increase its size to 40 bytes in total
2) derive span as `kind.span().merge(type_annotation.span)`, which would save size at the cost of a) GetSpan performance, b) we could not implement GetSpanMut, and c) we'd need to create a custom GetSpan impl for BindingPattern.

Boshen:
Ahh, I don't like either of these options. What if we accept that it's the way it is and hack it in `no-unused-vars` alone?

DonIsaac:
That's also acceptable assuming nobody else needs this.

magic-akari:
This is weird in SWC.

<img width="856" alt="image" src="https://github.com/user-attachments/assets/6ed09305-2bfa-4a28-9a22-18a6e663f527">
<img width="856" alt="image" src="https://github.com/user-attachments/assets/e01fb106-ef81-44e3-9d62-ceaaac3c48ed">


https://github.com/swc-project/swc/issues/8856

[...]

PR #4740, opened 2024-08-07
fix(linter/no-unused-vars): adjust Symbol spans

DonIsaac:
Workaround until #4739 is fixed. This is not performant.

Part of https://github.com/oxc-project/backlog/issues/84


Dunqing:
It is only fixed in no-unused-vars? I also noticed semantic checker has the same issue

Boshen (commented):
May I decline this PR? We should fix the root cause instead.

Issue #4179, opened 2024-07-11
linter: rule fixer improvements


DonIsaac:
I've been adding and repairing fixers to a lot of rules, but my ad-hoc approach
is not sustainable. This issue is a call for assistance, and tracks fixer
progress for each rule.

## Prerequisites
- [x] #4187
- [x] #4211
- https://github.com/oxc-project/oxlint-ecosystem-ci/issues/19

# Rules

For rules with an existing fixer:
- If the rule has a fixer (using `diagnostic_with_fix`), ensure that there are
  no cases where the fixer could create a syntax error or semantically different
  code.
- If the fix could be correct in some cases, but could produce semantically
  differently different code in others, use `diagnostic_with_fix_dangerous`.
- Link this issue to your PR updating the rule's fixer, and a maintainer will
  update this table. If no changes are needed, leave a comment here and we'll
  also update the table.

For rules without a fixer:
- Check if a fixer could be safely added in some cases. If so, create a PR with
  a fixer implementation that links to this issue.
- If no fixer can (or should) be created for the rule, leave a comment below and
  a maintainer will update the table.

### Legend
- <no emoji>: Not checked
- ✅ Has a working fixer
- 🚧 No fixer, but one could be added
- 🚫 No fixer, and one cannot be added

<details>
<summary>

## Correctness (159):

</summary>

Code that is outright wrong or useless.
| Rule name                                     | Source     | Status  |
| --------------------------------------------- | ---------- | ------- |
| for-direction                                 | eslint     |         |
| no-async-promise-executor                     | eslint     |         |
| no-caller                                     | eslint     |         |
| no-class-assign                               | eslint     |         |
| no-compare-neg-zero                           | eslint     |         |
| no-cond-assign                                | eslint     |         |
| no-const-assign                               | eslint     |         |
| no-constant-binary-expression                 | eslint     |         |
| no-constant-condition                         | eslint     |         |
| no-control-regex                              | eslint     |         |
| no-debugger                                   | eslint     |    ✅     |
| no-delete-var                                 | eslint     |         |
| no-dupe-class-members                         | eslint     |         |
| no-dupe-else-if                               | eslint     |         |
| no-dupe-keys                                  | eslint     |         |
| no-duplicate-case                             | eslint     |         |
| no-empty-character-class                      | eslint     |         |
| no-empty-pattern                              | eslint     |         |
| no-empty-static-block                         | eslint     |         |
| no-ex-assign                                  | eslint     |         |
[...]

PR #4748, opened 2024-08-08
feat(linter): add fixer for eslint/no-compare-neg-zero

heygsc:
part of #4179

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_compare_neg_zero.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_compare_neg_zero.rs
@@ -40,25 +40,31 @@ impl Rule for NoCompareNegZero {
         };
         if Self::should_check(expr.operator) {
             let op = expr.operator.as_str();
-            if is_neg_zero(&expr.left) || is_neg_zero(&expr.right) {
+            let is_left_neg_zero = is_neg_zero(&expr.left);
+            let is_right_neg_zero = is_neg_zero(&expr.right);
+            if is_left_neg_zero || is_right_neg_zero {
                 if expr.operator == BinaryOperator::StrictEquality {
                     ctx.diagnostic_with_suggestion(
                         no_compare_neg_zero_diagnostic(op, expr.span),
                         |fixer| {
-                            //replace x === -0 with Object.is(x, -0)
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_compare_neg_zero.rs
```diff
@@ -39,8 +40,21 @@ impl Rule for NoCompareNegZero {
...
+                    let span = Span::new(start, end);
+                    fixer.replace(span, "")
+                });
+            }
+            if is_neg_zero(&expr.right) {
+                ctx.diagnostic_with_fix(no_compare_neg_zero_diagnostic(op, expr.span), |fixer| {
+                    let start = expr.right.span().start;
+                    let end = start + 1;
+                    let span = Span::new(start, end);
+                    fixer.replace(span, "")
```
DonIsaac, line 56: You can use `fixer.delete` here.

crates/oxc_linter/src/rules/eslint/no_compare_neg_zero.rs
```diff
@@ -39,8 +40,21 @@ impl Rule for NoCompareNegZero {
         };
         if Self::should_check(expr.operator) {
             let op = expr.operator.as_str();
-            if is_neg_zero(&expr.left) || is_neg_zero(&expr.right) {
-                ctx.diagnostic(no_compare_neg_zero_diagnostic(op, expr.span));
+            if is_neg_zero(&expr.left) {
```
mysteryven, line 43: It looks like we need to consider two conditions:

1. Equality( `x === -0` ). It's more ergonomic to replace it with [`Object.is(x, -0)`](https://eslint.org/docs/latest/rules/no-compare-neg-zero#rule-details) , as it will change code behavior, we can it marked as [suggestion](https://github.com/oxc-project/oxc/blob/654395847fdbe2ee2f3174e2c17c8a1946900c74/crates/oxc_linter/src/fixer/fix.rs#L21-L24) fix
2. Comparison ( `x <= -0`). -0 is equal to 0 here, it's safe for us to replace it with `0`.

mysteryven (approved):
Thank you!

Issue #4656, opened 2024-08-05
☂️ eslint-plugin-vitest


Boshen:

shulaoda:
Does `vitest/no-commented-out-tests` seem to be duplicated with `jest/no-commented-out-tests`?

Boshen:
We need to map jest to vitest rules in [tasks/lint_rules](https://github.com/oxc-project/oxc/tree/main/tasks/lint_rules), they have the same implementation.

shulaoda:
`Justfile` seems to not have the command `just new-vitest-rule`. Should we add it or use `just new-jest-rule`? @Boshen

mysteryven:
> `Justfile` seems to not have the command `just new-vitest-rule`. Should we add it or use `just new-jest-rule`? @Boshen

Add this is welcome.

no-yan:
### Rules Implemented but Not Listed in the Table
Some rules are already implemented but not marked in the table. They are located in `src/rules/jest` because they share similar logic between Jest and Vitest.

The following files mention Vitest, suggesting they may also apply to Vitest:

<details><summary>List of files mentioning Vitest</summary>
<p>

```bash
> grep -rl "vitest"  src/rules/jest
./expect_expect.rs
./no_identical_title.rs
./no_conditional_in_test.rs
./valid_expect.rs
./consistent_test_it.rs
./no_disabled_tests.rs
./no_commented_out_tests.rs
./valid_describe_callback.rs
./no_focused_tests.rs
./no_alias_methods.rs
./no_restricted_jest_methods.rs
./no_test_prefixes.rs
./no_conditional_expect.rs
./prefer_hooks_in_order.rs

``` 

</p>
</details> 

### Unable to Register Rules with the Same Name Across Different Plugins

Some rules have the same name as those in the Jest plugin, which prevents them from being registered in Vitest. This is because the macro cannot differentiate identical rule names between plugins.

For more details, see: [#6520](https://github.com/oxc-project/oxc/issues/6520)

camchenry:
@no-yan I've opened a PR which will automatically keep this issue in sync with that list. Thanks for making a note of it. Hopefully that will make this more clear.

https://github.com/oxc-project/oxc/pull/6667

cc @DonIsaac @Boshen

Afsoon:
Working in https://github.com/oxc-project/oxc/pull/16540 I have notice that `vitest/no-restricted-vi-methods` is implemented but the table show as not. As you can see

[...]

PR #4752, opened 2024-08-08
chore(tasks): support init vitest lint rule

shulaoda:
Related to #4656


mysteryven (approved):
Thanks for your quick PR!

PR #4755, opened 2024-08-08
feat(linter/eslint-plugin-vitest): implement prefer-to-be-truthy

shulaoda:
Related to #4656

### DIFF
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -449,6 +449,7 @@ mod promise {
 
 mod vitest {
     pub mod no_import_node_test;
+    pub mod prefer_to_be_truthy;
 }
 
 oxc_macros::declare_all_lint_rules! {
@@ -854,4 +855,5 @@ oxc_macros::declare_all_lint_rules! {
     promise::no_new_statics,
     promise::param_names,
     vitest::no_import_node_test,
+    vitest::prefer_to_be_truthy,
 }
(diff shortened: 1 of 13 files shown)


shulaoda:
The difference in rules between `jest` and `vitest` is not significant, and I may be able to make simple modifications to `jest utilities` for unified use.

mysteryven:
> The difference in rules between `jest` and `vitest` is not significant, and I may be able to make simple modifications to `jest utilities` for unified use.

Most of them are the same, simply modifying is preferred, even it may make the code ugly :(

mysteryven:
Thank you! I am thinking whether we should move `parse_jest_fn_call` to `parse_test_fn_call`, and derive `parse_jest_fn_call` and `parse_vitest_fn_call` from it? Feel likes extends `expectTypeOf` in jest is a bit of a mess.

And need DonIsaac's review.👀

shulaoda:
> Thank you! I am thinking whether we should move `parse_jest_fn_call` to `parse_test_fn_call`, and derive `parse_jest_fn_call` and `parse_vitest_fn_call` from it? Feel likes extends `expectTypeOf` in jest is a bit of a mess.


The current difference between them is too small. If necessary, I can implement it in the next related PR, which won't be too long.

DonIsaac (approved):

DonIsaac:
LGTM, I'll let @mysteryven merge when he's happy with it.

mysteryven:
> I can implement it in the next related PR

It didn't influence other jest rules currently, let's follow the current step and refactor it when needed :)

mysteryven (approved):
Thank you!

PR #4735, opened 2024-08-07
refactor(ast, ast_codegen): use `generate_derive` for implementing `GetSpan` and `GetSpanMut` traits.

rzvxa:

PR #4757, opened 2024-08-08
refactor(ast_codegen): simplify `derive_get_span` generator

overlookmotel:
Simplify `derive_get_span` generator that was introduced in #4735. No change to functionality, just aiming for greater readability.

In particular:

* Move defining idents/tokens which are specific to `GetSpan` / `GetSpanMut` into those specific generators, rather than branching on `MUT` later on.
* Remove `const MUT` param.
* Remove the confusing pairs of closures and functions both called `derive_enum` / `derive_struct`.
* Inline function which generates the impls - prioritizing readability over DRY code.

### DIFF
--- a/tasks/ast_codegen/src/generators/derive_get_span.rs
+++ b/tasks/ast_codegen/src/generators/derive_get_span.rs
@@ -1,6 +1,6 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{Generics, Ident, Type};
+use syn::Ident;
 
 use crate::{
     output,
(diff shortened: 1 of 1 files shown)


overlookmotel:
What do you think @rzvxa? Personally I find this style less convoluted and a lot easier to understand.

NB: Rather than looking at the diff, I'd suggest comparing before and after.

rzvxa (approved):

rzvxa:
> What do you think @rzvxa? Personally I find this style less convoluted and a lot easier to understand.

TBH, I don't have a strong preference here. I'll merge it in if you like it more this way.

PR #4758, opened 2024-08-08
feat(linter): add fixer for eslint/no-eq-null

heygsc:
part of #4179

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_eq_null.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_eq_null.rs
@@ -32,7 +32,7 @@ declare_oxc_lint!(
     /// ```
     NoEqNull,
     restriction,
-    fix
+    fix_dangerous
 );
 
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_eq_null.rs
```diff
@@ -49,10 +50,24 @@ impl Rule for NoEqNull {
                     & binary_expression.left.is_null()
                     & bad_operator
             {
-                ctx.diagnostic(no_eq_null_diagnostic(Span::new(
-                    binary_expression.span.start,
-                    binary_expression.span.end,
-                )));
+                ctx.diagnostic_with_fix(
```
mysteryven, line 53: Maybe `diagnostic_with_dangerous_fix`?

```js
x = undefined
x === null // false
x == null // true
```

DonIsaac (approved):

PR #4770, opened 2024-08-08
feat(linter/eslint-plugin-vitest): implement prefer-to-be-falsy

shulaoda:


mysteryven (approved):
Well done!!

PR #4771, opened 2024-08-09
feat(linter): add fixer for eslint/no-new-wrappers

heygsc:
part of #4179

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_new_wrappers.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_new_wrappers.rs
@@ -1,7 +1,7 @@
 use oxc_ast::{ast::Expression, AstKind};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
+use oxc_span::{GetSpan, Span};
 
 use crate::{context::LintContext, rule::Rule, AstNode};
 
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_new_wrappers.rs
```diff
@@ -91,5 +100,27 @@ fn test() {
         ",
     ];
 
-    Tester::new(NoNewWrappers::NAME, pass, fail).test_and_snapshot();
+    let fix = vec![
+        ("var a = new String('hello');", "var a = String('hello');"),
```
mysteryven, line 104: Thank you, I'm sorry, could I refuse this fixer? This changes the code meaning and seems `String('hello')` is not also good:

```js
typeof new String('hello') // 'object'
typeof String("hello")
```

 I'd like `'hello'`, but it will have other edge cases:
 
```js
new Number(1).toString() // original code

1.toString() // SyntaxError
```

It's acceptable and more safe to change one by one manually.

heygsc: > Thank you, I'm sorry, could I refuse this fixer? This changes the code meaning and seems `String('hello')` is not also good:
> 
> ```js
> typeof new String('hello') // 'object'
> typeof String("hello")
> ```
> 
> I'd like `'hello'`, but it will have other edge cases:
> 
> ```js
> new Number(1).toString() // original code
> 
> 1.toString() // SyntaxError
> ```
> 
> It's acceptable and more safe to change one by one manually.

That makes sense.

PR #4773, opened 2024-08-09
refactor(ast): `oxc_ast` crate re-export AST types from other crates

overlookmotel:
Some AST types are defined in crates outside `oxc_ast` crate. Re-export these types from `oxc_ast` crate for ease - so you don't need to remember where they're defined to use them.

### DIFF
--- a/crates/oxc_ast/src/ast/mod.rs
+++ b/crates/oxc_ast/src/ast/mod.rs
@@ -183,3 +183,12 @@ mod ts;
 use macros::inherit_variants;
 
 pub use self::{js::*, jsx::*, literal::*, ts::*};
+
+// Re-export AST types from other crates
+pub use oxc_span::{Atom, Language, LanguageVariant, ModuleKind, SourceType, Span};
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4773** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@rzvxa @Dunqing I would have asked Boshen to review this, but he's off today. I can't see any downside to this, so I think safe to merge if one of you can confirm that you can't see a downside either.

Dunqing (approved):
Looks good!

rzvxa (approved):
Good call, I think we should move stuff out of `oxc_span` later on. Atom makes sense to be there but I can't say the same thing for the SourceType and its sub-types.

This PR helps if we want to do that.

overlookmotel:
Thanks both for reviewing.

PR #4783, opened 2024-08-09
test(linter): add fixer test for unicorn/no-zero-fractions

heygsc:
Add fixer test.
Some of them have not passed yet, so I have temporarily made comments as todo.

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
@@ -134,5 +134,35 @@ fn test() {
         r"function foo(){return.0+.1}",
     ];
 
-    Tester::new(NoZeroFractions::NAME, pass, fail).test_and_snapshot();
+    let fix = vec![
+        (r"const foo = 1.0", r"const foo = 1"),
+        (r"const foo = 1.0 + 1", r"const foo = 1 + 1"),
+        (r"foo(1.0 + 1)", r"foo(1 + 1)"),
+        (r"const foo = 1.00", r"const foo = 1"),
+        (r"const foo = 1.00000", r"const foo = 1"),
+        (r"const foo = -1.0", r"const foo = -1"),
+        (r"const foo = 123123123.0", r"const foo = 123123123"),
+        (r"const foo = 123.11100000000", r"const foo = 123.111"),
+        (r"const foo = 1.", r"const foo = 1"),
+        (r"const foo = +1.", r"const foo = +1"),
+        (r"const foo = -1.", r"const foo = -1"),
+        // maybe todo
+        // In the following tests, the comments did not pass the fixer.
+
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
```diff
@@ -134,5 +134,35 @@ fn test() {
...
+        (r"const foo = 1.00", r"const foo = 1"),
+        (r"const foo = 1.00000", r"const foo = 1"),
+        (r"const foo = -1.0", r"const foo = -1"),
+        (r"const foo = 123123123.0", r"const foo = 123123123"),
+        (r"const foo = 123.11100000000", r"const foo = 123.111"),
+        (r"const foo = 1.", r"const foo = 1"),
+        (r"const foo = +1.", r"const foo = +1"),
+        (r"const foo = -1.", r"const foo = -1"),
+        // maybe todo
+        // In the following tests, the comments did not pass the fixer.
```
rzvxa, line 150: I believe we should uncomment them, If it results in a completely wrong code we shouldn't suggest it at all, But as far as I can see these are minor changes that wouldn't break the code(for example the parentheses are the most harmless here).
For exponential numbers, it can be fine, If we uncomment them when we fix our codegen to keep exponentials we'd come back to these - now - failing tests and use exponentials again(similar to the original versions).
But if we think changing the number format is unacceptable we shouldn't suggest it at all.

heygsc: > I believe we should uncomment them, If it results in a completely wrong code we shouldn't suggest it at all, But as far as I can see these are minor changes that wouldn't break the code(for example the parentheses are the most harmless here). For exponential numbers, it can be fine, If we uncomment them when we fix our codegen to keep exponentials we'd come back to these - now - failing tests and use exponentials again(similar to the original versions). But if we think changing the number format is unacceptable we shouldn't suggest it at all.

At first, the test only had pass/fail and no fix.
When there are suggests like this, I don't know if these fix tests should be added.

What you said makes sense.
So, how should the PR code be modified? Delete the test related to parentheses and keep the comment for exponential numbers? Or you can modify it.😂

rzvxa: Let's merge it as is and follow up with another PR addressing it, We should either stop suggesting these fixes or we should do it correctly and uncomment the tests.

DonIsaac:
Looks like an existing bug in the dozer that needs addressing?

rzvxa:
Well, it breaks the users' code, So we should either stop suggesting them or fix the issue. At the very least we should temporarily make the fix dangerous for these cases. And even exponentials which isn't breaking in any sense, Is just too inconvenient. Imagine getting all your numbers replaced unintentionally and you just push the changes.
Now you have to go back in and revert a bunch of numbers.

PR #4791, opened 2024-08-09
feat(linter): overhaul unicorn/no-useless-spread

DonIsaac:
I got tired of seeing useless spreads on ternaries and `arr.reduce()` within my company's internal codebase so I overhauled this rule.

## Changes
- add fixer for object spreads
  ```js
  const before = { a, ...{ b, c }, d }
  const after = { a,  b, c, d } // fixer does not dedupe spaces before `b`
  ```
- recursively check for useless clones on complex expressions. This rule now catches and auto-fixes the following cases:
   ```js
  // ternaries when both branches create a new array or object
   const obj = { ...(foo ? { a: 1 } : { b: 2 }) }
  // recursive, so this can support complex cases
  const arr = [ ...(foo ? a.map(fn) : bar ? Array.from(iter) : await Promise.all(bar)) ]
  // reduce functions where the initial accumulator creates a new object or array
   const obj = { ...(arr.reduce(fn, {}) }
  ```

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_useless_spread/mod.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_useless_spread/mod.rs
@@ -1,7 +1,9 @@
+mod const_eval;
+
 use oxc_ast::{
     ast::{
-        match_expression, Argument, ArrayExpression, ArrayExpressionElement, CallExpression,
-        Expression, SpreadElement,
+        ArrayExpression, ArrayExpressionElement, CallExpression, Expression, NewExpression,
+        ObjectExpression, ObjectPropertyKind, SpreadElement,
     },
     AstKind,
 };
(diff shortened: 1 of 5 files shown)


crates/oxc_linter/src/rules/unicorn/no_useless_spread/mod.rs
```diff
@@ -326,153 +338,141 @@ fn check_useless_iterable_to_array<'a>(
...
+    // already diagnosed by first check
+    if matches!(target, Expression::ArrayExpression(_) | Expression::ObjectExpression(_)) {
+        return false;
+    }
 
-                format!("{}.{}", object_name, method.static_property_name().unwrap())
-            },
-        );
+    let hint = target.const_eval();
+    let hint_matches_expr = if is_array { hint.is_array() } else { hint.is_object() };
```
mysteryven, line 387: What's the means of `hint_matches_expr` here? 🤔

DonIsaac: true if
1. we're checking for clones on arrays and `ValueHint` suggests the value is an array
2. vice versa for objects

mysteryven (approved):

PR #4787, opened 2024-08-09
perf(linter): optmize allocations in jest fn parsing

lucab:

PR #4792, opened 2024-08-09
ci: add jest test file to benchmarks

DonIsaac:
This will help bench Jest linter rules. See conversation in #4787 for details.

### DIFF
--- a/tasks/common/src/test_file.rs
+++ b/tasks/common/src/test_file.rs
@@ -61,7 +61,7 @@ impl TestFiles {
         Self { files: vec![file] }
     }
 
-    fn complicated_urls() -> [&'static str; 5] {
+    fn complicated_urls() -> [&'static str; 6] {
         [
             // TypeScript syntax (2.81MB)
             "https://raw.githubusercontent.com/microsoft/TypeScript/v5.3.3/src/compiler/checker.ts",
--- a/tasks/common/src/test_file.rs
+++ b/tasks/common/src/test_file.rs
@@ -73,6 +73,8 @@ impl TestFiles {
             "https://cdn.jsdelivr.net/npm/pdfjs-dist@4.0.269/build/pdf.mjs",
             // ES5 (3.9M)
             "https://cdn.jsdelivr.net/npm/antd@5.12.5/dist/antd.js",
+            // Jest test file from Jest's own repo (8.0K)
+            "https://raw.githubusercontent.com/jestjs/jest/main/e2e/__tests__/coverageReport.test.ts",
         ]
     }
 }


Boshen:
> ## [CodSpeed Performance Report](https://codspeed.io/oxc-project/oxc/branches/don/chore/jest-benchmark)
> ### Merging #4792 will **not alter performance**
> Comparing `don/chore/jest-benchmark` ([683f5f4](https://github.com/oxc-project/oxc/commit/683f5f4e05616e1bada1dd1c667ba271a9b34809)) with `main` ([4dd29db](https://github.com/oxc-project/oxc/commit/4dd29dbc2beac49df28ccf15ea433ffbaf3e9fd4))
> 
> ### Summary
> `✅ 29` untouched benchmarks
> 
> `🆕 4` new benchmarks
> 
> ### Benchmarks breakdown
> Benchmark	`main`	`don/chore/jest-benchmark`	Change
> 🆕	`lexer[coverageReport.test.ts]`	N/A	32.6 µs	N/A
> 🆕	`parser[coverageReport.test.ts]`	N/A	162 µs	N/A
> 🆕	`semantic[coverageReport.test.ts]`	N/A	202.1 µs	N/A
> 🆕	`transformer[coverageReport.test.ts]`	N/A	340.8 µs	N/A

https://github.com/oxc-project/oxc/blob/d191823a0a150a5e8f40526ca52de6567cceeef0/tasks/benchmark/benches/linter.rs#L18-L24

And adding more test files will hurt CI time 😢 

https://github.com/oxc-project/oxc/blob/d191823a0a150a5e8f40526ca52de6567cceeef0/.github/workflows/benchmark.yml#L162-L164

rzvxa:
> And adding more test files will hurt CI time 😢

This isn't directly related to this PR but I just want to express something on my mind.

I believe we can use an on-demand CI pass for heavy stuff. Instead of running some of the CI tasks on push we can use the action button or provide a command through @oxc-bot.

We can benchmark general stuff like before(on push) but provide a command for maintainers and the PR's author to run e2e benchmarks, oxlint-ecosystem, monitor-oxc, or other complicated tasks. We usually only need to run these once per review, Sometime we wouldn't even need to run these on trivial changes which saves a lot of trees(as suggested by @overlookmotel).

I've mentioned it here before https://github.com/oxc-project/backlog/issues/86
I haven't worked with Github's bot API but it shouldn't be too hard to implement. We can also use our own CI runner for these(a small cluster of docker containers). By using such a CI environment we can eliminate a lot of unpredictable variance caused by Github agents. Since with this setup, we can always run a benchmark in the same environment(and round robin between them as new requests queue). We might even be able to time some of our stuff instead of doing system-independent benchmarks.

I also noticed @overlookmotel is working toward filtering CI tasks so they only run on appropriate changes. That can help as well.

I wish Graphite's CI optimization had a better way to strategize tasks and where they should run.

Boshen:
Closing in favor of selective benchmark setup.

PR #4312, opened 2024-07-17
feat(linter): add suggestion for no-console

DonIsaac:
Part of #4179

PR #4793, opened 2024-08-10
fix(linter): improve no-zero-fractions rule for member expressions and scientific notation

### LINKED ISSUE #4788: `unicorn/no-zero-fractions` suggests subpar fixes
This rule fails 3 kinds of fixer tests. We should either stop suggesting them or correct their behavior.

1. Spacing(Invalid code)
For:
```js
function foo(){return.0}
```
We generate:
```js
function foo(){return0}
```
Where it should be:
```js
function foo(){return 0}
```

2. Parenthesized expressions(dangerous):

For:
```js
1.00.toFixed(2)
```
We generate:
```js
1.toFixed(2)
```
Where it should be:
```js
(1).toFixed(2)
```

3. Exponential numbers(Inconvenient):

For:
```js
const foo = 1.e10
```
We generate:
```js
const foo = 110
```
Where it should be:
```js
const foo = 1e10
```

------------

Tests are introduced in #4783

https://github.com/oxc-project/oxc/blob/4dd29dbc2beac49df28ccf15ea433ffbaf3e9fd4/crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs#L149-L164

cblh:
fix: #4788

- Added a check for parent nodes that are MemberExpressions
- If a parent node is a MemberExpression, the fixed number is now wrapped in parentheses to maintain correct syntax
- Add support for scientific notation (e.g., 1e10, 1.e10)
- Handle special cases where a space is needed after certain keywords (return, throw, typeof, void) to prevent the number from being interpreted as a property access


crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
```diff
@@ -146,22 +177,28 @@ fn test() {
...
-        // (r"const foo = 1.e10", r"const foo = 1e10"),
-        // (r"const foo = +1.e-10", r"const foo = +1e-10"),
-        // (r"const foo = -1.e+10", r"const foo = -1e+10"),
+        (r"const foo = 1.e10", r"const foo = 1e10"),
+        (r"const foo = +1.e-10", r"const foo = +1e-10"),
+        (r"const foo = -1.e+10", r"const foo = -1e+10"),
         (r"const foo = (1.).toString()", r"const foo = (1).toString()"),
-        // (r"1.00.toFixed(2)", r"(1).toFixed(2)"),
-        // (r"1.00 .toFixed(2)", r"(1) .toFixed(2)"),
+        (r"1.00.toFixed(2)", r"(1).toFixed(2)"),
```
DonIsaac, line 184: Could this break ASI?
```js
console.log('foo')
1.00.toFixed(2) // -> (1).toFixed(2), parsed as a function call
```

crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
```diff
@@ -66,12 +66,43 @@ impl Rule for NoZeroFractions {
             } else {
                 zero_fraction(number_literal.span, &fmt)
             },
-            |fixer| fixer.replace(number_literal.span, fmt),
+            |fixer| {
+                let mut fixed = fmt.clone();
```
DonIsaac, line 70: Why `clone()` instead of a move?

crates/oxc_linter/src/utils/unicorn/numeric.rs
```diff
@@ -0,0 +1,13 @@
+use lazy_static::lazy_static;
+use regex::Regex;
+
+// Determine whether this node is a decimal integer literal.
+// Copied from https://github.com/eslint/eslint/blob/cc4871369645c3409dc56ded7a555af8a9f63d51/lib/rules/utils/ast-utils.js#L1237
+lazy_static! {
+    static ref DECIMAL_INTEGER_PATTERN: Regex =
+        Regex::new(r"^(?:0|0[0-7]*[89]\d*|[1-9](?:_?\d)*)$").unwrap();
```
DonIsaac, line 8: Does `NumericLiteral#base` not suit your needs?

cblh: This is to check whether 1.0100 becomes 1.01 or 1.00 becomes 1. I found that format_raw's return is_dangling_dot can distinguish this situation, so I will delete this code

crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
cblh: you are right

crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
cblh: i will fix it

cblh: I need help, I don't know how to get the previous token.

DonIsaac: this case is also a problem blocking #4312. I'll add a task to add a `delete_statement` method to `RuleFixer`.

cblh: Sorry, I don't know what `delete_statement` does? The logic I need to implement here is It : 

checks the type and value of tokenBefore:
If tokenBefore is a Punctuator (e.g., a symbol like ;, ], or )), it determines whether a semicolon is necessary based on the context (e.g., the type of the last block node).
If the token type is in tokenTypesNeedsSemicolon, it returns true (semicolon needed).
Special cases like Template strings, ObjectExpression blocks, and certain Identifier cases are handled explicitly. 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/5c92a9aed7793afa00f23e72e25bc58c10adecd4/rules/utils/needs-semicolon.js#L33

Is this also a blocking problem or is there an API available?

DonIsaac: This API does not exist yet, we should move ahead with this improvement for the time being.

DonIsaac (approved):

cblh:
> This API does not exist yet, we should move ahead with this improvement for the time being.

Since it is not a problem with the current rules, let’s merge this PR first. Then join a new issue to follow up on this API

crates/oxc_linter/src/rules/unicorn/no_zero_fractions.rs
cblh: > This API does not exist yet, we should move ahead with this improvement for the time being.

Since it is not a problem with the current rules, let’s merge this PR first. Then join a new issue to follow up on this API

DonIsaac:
Is there still a service disruption for Github Actions?

cblh:
> Is there still a service disruption for Github Actions?

@DonIsaac github action service has been restored

PR #4795, opened 2024-08-10
chore: support windows shell for just

### LINKED ISSUE #4794: Should we add windows shell support for `just`?
When I use my own `Windows` computer for work, I find that it doesn't seem to work.
wecom-temp-27415-49d2fbc0124d193cce822732f0411ecb

shulaoda:
closes #4794

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,5 +1,8 @@
 #!/usr/bin/env -S just --justfile
 
+set windows-shell := ["powershell"]
+set shell := ["bash", "-cu"]
+
 _default:
   @just --list -u
 


Boshen:
Thank you.

rzvxa:
Asking this purely out of curiosity.  How I was using it on my Windows machine until now😆I didn't use the init command though.
Does it only fail on cmd? Because I use PowerShell as my default shell on Windows and that might be the reason it used to work even before this fix.

rzvxa:
Now I noticed in the original issue you were also using PowerShell, So it has to be because of WSL. WSL exposes a bash command to the shell.

PR #4796, opened 2024-08-10
feat(linter): `eslint-plugin-vitest/require-local-test-context-for-concurrent-snapshots`

eryue0220:
see support [eslint-plugin-vitest/require-local-test-context-for-concurrent-snapshots](https://github.com/veritem/eslint-plugin-vitest/blob/main/docs/rules/require-local-test-context-for-concurrent-snapshots.md)


crates/oxc_linter/src/rules/vitest/require_local_test_context_for_concurrent_snapshots.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// test.concurrent('myLogic', (context) => {
+    ///     context.expect(true).toMatchSnapshot();
+    /// })
+    /// ```
+    RequireLocalTestContextForConcurrentSnapshots,
+    style,
+);
+
+impl Rule for RequireLocalTestContextForConcurrentSnapshots {
+    fn run_once(&self, ctx: &LintContext) {
```
mysteryven, line 59: Looks like we don't need to collect function parameters? We can iterate all `expect` call expression and combine you `check` method  

https://github.com/veritem/eslint-plugin-vitest/blob/8b11357b8690c276b869a04140f5238679bab5de/src/rules/require-local-test-context-for-concurrent-snapshots.ts#L33

mysteryven:
Could you split a PR that doesn't contain `require-local-test-context-for-concurrent-snapshots`? Let's merge the bug fixes first, maybe in #4797 ?

eryue0220:
> Could you split a PR that doesn't contain `require-local-test-context-for-concurrent-snapshots`? Let's merge the bug fixes first, maybe in #4797 ?

Done.

mysteryven:
Close because of duplicated: #4951.

Thanks for your work on this rule.

PR #4797, opened 2024-08-10
fix(linter): jest/vitest rule compat

eryue0220:
This pr is for jest/vitest compat and add another jest rule condition


crates/oxc_linter/src/utils/jest/parse_jest_fn.rs
```diff
@@ -597,3 +608,1302 @@ const VALID_JEST_FN_CALL_CHAINS: [[&str; 4]; 51] = [
...
+    "suite.runIf.todo.each",
+    "xtest",
+    "xtest.each",
+    "xit",
+    "xit.each",
+    "fit",
+    "xdescribe",
+    "xdescribe.each",
+    "fdescribe"
+};
```
Boshen, line 1909: This is huge! Any way we can shrink this down?

eryue0220: Sorry for the late, I'll check this.

eryue0220: Updated.

PR #4808, opened 2024-08-10
fix(transformer/typescript): generated assignment for constructor arguments with access modifiers should be injected to the top of the constructor

### LINKED ISSUE #4789: transformer: class constructor argument access modifier special behavior
In TypeScript, the following class (case 1) reports a type error:

```ts
class Foo {
  x = this.foo // error: Property 'foo' is used before its initialization
  foo: any
  constructor(foo: any) {
    this.foo = foo
  }
}
```

Because it will be transformed to:

```js
class Foo {
    constructor(foo) {
        this.x = this.foo; // <-- this.foo not assigned yet!
        this.foo = foo;
    }
}
```

[TS Playground](https://www.typescriptlang.org/play/?target=99#code/MYGwhgzhAEBiD29oG8CwAoa0Ae0C80ALgBYCWEAdAGaLQD0d0ApgE4vwsBc0ACuwA6tCAT2gByGvDHRy0AK4QmAE2gAjJjRZMZhGKQB2pQqTAhSALzDH4+jFkncw+4XejAbEQiznBCHABQO0E7CAJQorlgk5NS0BJKuAL4YiUA)

However, if it uses access modifier on the constructor argument (case 2):

```ts
class Foo {
  x = this.foo
  constructor(public foo: any) {
    console.log(this.foo)
  }
}
```

It will work both in types and at runtime, because it will be transformed to:

```js
class Foo {
    constructor(foo) {
        this.foo = foo; // <-- injected from argument
        this.x = this.foo; // <-- moved from field initializers
        console.log(this.foo);
    }
}
```

[TS Playground](https://www.typescriptlang.org/play/?target=99#code/MYGwhgzhAEBiD29oG8CwAoa0Ae0C80ALgBYCWEAdAGaIZbDwB2EhATgK7CHysAUADuwBGIUsGg14ALmhhGATwCUKOlmgNm8EAFMKIeAHNeJctUSKA3KoC+Ga0A)

Notice how `x = this.foo` is moved into the constructor to be **after** the `this.foo` assignment.

## Oxc's current behavior

Oxc currently transforms case 2 to:

```js
class Foo {
  x = this.foo
  constructor(foo) {
    console.log(this.foo)
    this.foo = foo // <-- injected from argument
  }
}
```

Notice that both `x = this.foo` and `console.log(this.foo)` are executed before the `this.foo` assignment, both leading to runtime errors.

## What needs to be fixed

1. The generated assignment for constructor arguments with access modifiers should be injected to the **top** of the constructor, not the bottom.
2. When constructor argument access modifiers are used, **all** class field initializers need to be moved into the constructor (so their initialization order is preserved), **after** the argument assignments, and **before** existing constructor code.

## Related: Behavior with `useDefineForClassFields: true`

When `useDefineForClassFields` is set to true, case 2 will also throw a type error. And TS's transform out will become:

```js
class Foo {
    foo;
    x = this.foo;
    constructor(foo) {
        this.foo = foo;
        console.log(this.foo);
    }
}
```
[TS Playground](https://www.typescriptlang.org/play/?useDefineForClassFields=true&target=99#code/MYGwhgzhAEBiD29oG8CwAoa0Ae0C80ALgBYCWEAdAGaIZbDwB2EhATgK7CHysAUADuwBGIUsGg14ALmhhGATwCUKOlmgNm8EAFMKIeAHNeJctUSKA3KoC+Ga0A)

I'm not sure if oxc's TS transform currently takes this into account, but this is something we will need to consider.

Dunqing:
fix: #4789

### DIFF
--- a/crates/oxc_transformer/src/typescript/annotations.rs
+++ b/crates/oxc_transformer/src/typescript/annotations.rs
@@ -298,7 +298,8 @@ impl<'a> TypeScriptAnnotations<'a> {
                         )
                     })
                     .statements
-                    .extend(
+                    .splice(
+                        0..0,
                         self.assignments
                             .drain(..)
                             .map(|assignment| assignment.create_this_property_assignment(ctx)),
--- a/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/class-constructor-arguments/input.ts
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/class-constructor-arguments/input.ts
@@ -1,7 +1,7 @@
 class Foo {
   boom: number;
   constructor(public foo, private bar, protected zoo, readonly bang, too) {
-
+    console.log(this.foo, this.bar, this.zoo, this.bang);
   }
 }
 class Bar extends Foo {
--- a/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/class-constructor-arguments/output.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/class-constructor-arguments/output.js
@@ -5,6 +5,7 @@ class Foo {
     this.bar = bar;
     this.zoo = zoo;
     this.bang = bang;
+    console.log(this.foo, this.bar, this.zoo, this.bang);
   }
 }
 class Bar extends Foo {


Dunqing:
* **#4808** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4808?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):

PR #4810, opened 2024-08-10
feat(allocator): introduce `Address`

### LINKED ISSUE #4807: Add `Address` API
Rolldown is using `Span`s as unique identifiers for AST nodes.

[As I've said](https://github.com/oxc-project/oxc/issues/4799#issuecomment-2280937337) I think `Span` is the wrong tool for this job - `Span`s by their nature just aren't unique.

We have been discussing adding `AstNodeId` to all AST nodes, which would be a unique ID, but we've not reached a decision whether to do that or not yet.

There is one other unique identifier available to us - the address of structs in memory. Our bump allocator does not reuse freed memory, and it's guaranteed that the contents of a `Box` never moves in memory. So if `T` is in a `Box<T>`, the memory address of `T` is guaranteed to be stable and unique for as long as the allocator containing the AST is "live".

Add an API like this:

```rs
#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
pub struct Address(usize);

impl<'a, T> oxc_allocator::Box<'a, T> {
    pub fn address(&self) -> Address {
        Address(std::ptr::addr_of!(**self) as usize)
    }
}
```

I know that on face of it, this looks "hacky", but I feel it's completely legitimate. Our choice of using an arena allocator gives us certain extra powers and guarantees - why not take advantage of them?

Memory address of a `&T` is also stable, as long as the `T` is in a `Box<T>` in arena. But to avoid mistakes (e.g. `&T` refers to a `T` on the stack, not in arena), I think better to only implement for `Box`.

overlookmotel:
Closes #4807.

Introduce `Address`. `Address` can be obtained from a `Box<T>` and can act as a unique identifier for an AST node in arena.

NB: It will also be unique across 2 ASTs in different allocators as long as neither allocator is dropped.

### DIFF
--- a/crates/oxc_allocator/src/arena.rs
+++ b/crates/oxc_allocator/src/arena.rs
@@ -212,6 +212,21 @@ impl<'alloc, T: Hash> Hash for Vec<'alloc, T> {
     }
 }
 
+/// Memory address of an AST node in arena.
+///
+/// `Address` is generated from a `Box<T>`.
+/// AST nodes in a `Box` in an arena are guaranteed to never move in memory,
+/// so this address acts as a unique identifier for the duration of the arena's existence.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Address(usize);
+
+impl<'a, T> Box<'a, T> {
+    #[inline]
+    pub fn address(&self) -> Address {
+        Address(ptr::addr_of!(**self) as usize)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#4810** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4810?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #996, opened 2023-10-15
feat(transformer): finish 2016 exponentiation operator

Boshen:
The bulk of the code for review: https://github.com/web-infra-dev/oxc/pull/996/files#diff-4950080c26a449198bc5b125fd72a73afe42cddde9943bc7280fa1cbb7310ced

It took me a while to understand what babel was doing ...

As a bonus, I found swc producing the wrong code for the following snippet:

```javascript
let x = {};
let y = 0;
let z = 0;
x[z++] **= y;
```

Gets transformed into

```javascript
let x = {};
let y = 0;
let z = 0;
var ref = x[z++];
x[z++] = Math.pow(ref, y);
```

Notice how `z++` is evaluated twice.

---

Which means I should ...
- [ ] add more unit tests for this PR

And I also noticed that the benchmark for transformer is wrong ...
- [ ] Fix transformer benchmark because it is transforming the already transformed AST

PR #4813, opened 2024-08-10
feat:  transform exponentiation operator

IWANABETHATGUY:


Boshen:
We had this code before I think, we can copy some of it back.

IWANABETHATGUY:
> We had this code before I think, we can copy some of it back.

yeah, https://github.com/oxc-project/oxc/pull/996, why we remove these code?

Boshen:
> > We had this code before I think, we can copy some of it back.
> 
> yeah, #996, why we remove these code?

Mostly because I wasn't satisfied with the architecture ...

PR #4814, opened 2024-08-10
refactor(linter): use Option to reduce nested level in `eslint/getter-return`

IWANABETHATGUY:
1. Simplify `is_wanted_node`

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/getter_return.rs
+++ b/crates/oxc_linter/src/rules/eslint/getter_return.rs
@@ -119,72 +119,66 @@ impl GetterReturn {
     }
 
     /// Checks whether it is necessary to check the node
-    fn is_wanted_node(node: &AstNode, ctx: &LintContext<'_>) -> bool {
-        if let Some(parent) = ctx.nodes().parent_node(node.id()) {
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/getter_return.rs
```diff
@@ -119,72 +119,66 @@ impl GetterReturn {
...
 
     /// Checks whether it is necessary to check the node
-    fn is_wanted_node(node: &AstNode, ctx: &LintContext<'_>) -> bool {
-        if let Some(parent) = ctx.nodes().parent_node(node.id()) {
-            match parent.kind() {
-                AstKind::MethodDefinition(mdef) => {
-                    if matches!(mdef.kind, MethodDefinitionKind::Get) {
-                        return true;
-                    }
+    fn is_wanted_node(node: &AstNode, ctx: &LintContext<'_>) -> Option<bool> {
```
IWANABETHATGUY, line 122: Since all `None` branch will return `false` finally, so it is safe to directly use `Option` with `?`

DonIsaac (approved):

PR #4852, opened 2024-08-12
feat(linter/jsx-a11y): add fixer for anchor-has-content

DonIsaac:
Add a conditional fix that removes `aria-hidden` from an anchor's child if there
is only a single child. This PR also fixes a false positive on hidden anchors.
It should report visible anchors with hidden content, not hidden anchors.

### DIFF
--- a/crates/oxc_linter/src/rules/jsx_a11y/anchor_has_content.rs
+++ b/crates/oxc_linter/src/rules/jsx_a11y/anchor_has_content.rs
@@ -1,10 +1,14 @@
-use oxc_ast::AstKind;
+use oxc_ast::{
+    ast::{JSXAttributeItem, JSXChild, JSXElement},
+    AstKind,
+};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::Span;
 
 use crate::{
     context::LintContext,
+    fixer::{Fix, RuleFix},
     rule::Rule,
     utils::{
         get_element_type, has_jsx_prop_ignore_case, is_hidden_from_screen_reader,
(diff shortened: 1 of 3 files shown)


DonIsaac:
* **#4852** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/rules/jsx_a11y/anchor_has_content.rs
```diff
@@ -70,7 +69,9 @@ impl Rule for AnchorHasContent {
             };
             if name == "a" {
                 if is_hidden_from_screen_reader(ctx, &jsx_el.opening_element) {
-                    ctx.diagnostic(remove_aria_hidden(jsx_el.span));
+                    // ctx.diagnostic_with_suggestion(remove_aria_hidden(jsx_el.span), |_fixer| {
+                    //     remove_hidden_attributes(&jsx_el.opening_element)
+                    // });
```
camc314, line 74: remove/ uncomment?

crates/oxc_linter/src/rules/jsx_a11y/anchor_has_content.rs
```diff
@@ -114,12 +146,27 @@ fn test() {
...
+        (r"<a aria-hidden>Foo</a>", None, None),
+        (r#"<a aria-hidden="true">Foo</a>"#, None, None),
+        (r"<a hidden>Foo</a>", None, None),
+        (r"<a aria-hidden><span aria-hidden>Foo</span></a>", None, None),
+        (r#"<a hidden="true">Foo</a>"#, None, None),
+        (r#"<a hidden="">Foo</a>"#, None, None),
+        // TODO: should these be failing?
+        (r"<a><Bar hidden /></a>", None, None),
+        (r#"<a><Bar hidden="" /></a>"#, None, None),
+        (r#"<a><Bar hidden="until-hidden" /></a>"#, None, None),
```
camc314, line 160: good question 🙂 probably not? if they are capitalized, that infers that it's a react component rather than a html one. so should be allowed?

camc314 (approved):

crates/oxc_linter/src/rules/jsx_a11y/anchor_has_content.rs
DonIsaac: > good question 🙂 probably not? if they are capitalized, that infers that it's a react component rather than a html one. so should be allowed?

It has less to do with `Bar` and more to do with `hidden`. `aria-hidden` gets reported, but `hidden` does not.

PR #4856, opened 2024-08-12
docs(ast, ast_macros, ast_tools): better documentation for `Ast` helper attributes.

rzvxa:


rzvxa:
* **#4856** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):
The docs added here to `oxc_ast_macros/src/lib.rs` are excellent.

Do you think we should retain *some* comment at top of `oxc_ast/src/ast/js.rs` etc pointing readers to the docs in `oxc_ast_macros`?

People may understand that `#[scope]` attr etc are probably related to the `#[ast]` attr, and so will look there for docs. But if you try "jump to definition" on `generate_derive`, it doesn't take you anywhere, and it's not obvious that `generate_derive` is related to `#[ast]`.

rzvxa:
> The docs added here to `oxc_ast_macros/src/lib.rs` are excellent.
> 
> Do you think we should retain _some_ comment at top of `oxc_ast/src/ast/js.rs` etc pointing readers to the docs in `oxc_ast_macros`?
> 
> People may understand that `#[scope]` attr etc are probably related to the `#[ast]` attr, and so will look there for docs. But if you try "jump to definition" on `generate_derive`, it doesn't take you anywhere, and it's not obvious that `generate_derive` is related to `#[ast]`.

Oh, I didn't notice outer markers aren't pointing to the macro. My initial reasoning behind removing those was to reduce the surface area of our documentation, It is much easier to maintain the docs in one place instead of having to keep multiple places in sync.

I'll bring back the old comments and look for a way so we can have jump to the definition on those attributes as well.

overlookmotel (approved):

overlookmotel:
> Do you think it is worth implementing?

I'm not sure. What do you think? I would definitely say yes if we could do it without adding an extra crate, but if it requires an extra crate, maybe it's not worth the complexity.

I am not really au fait with how proc macro spans interface with Rust Analyser, but is there any other way to achieve this? Like e.g.:

1. In proc macro, set the span of `generated_derive` to be span of `#[ast]`, so at least jump to definition points you in the right direction.
2. Do the `macro_rules! generate_derive` trick you suggested, but put that in `oxc_ast/src/macros.rs`.

Basically, I like this idea, but think introducing an extra crate just for it is maybe overkill. But if you want to go for it, do it!

rzvxa:
> I'm not sure. What do you think? I would definitely say yes if we could do it without adding an extra crate, but if it requires an extra crate, maybe it's not worth the complexity.

I agree on the fact that adding it might be a bit of overkill. Let's see if there comes any other situation that such a thing can help with.

>     1. In proc macro, set the span of `generated_derive` to be span of `#[ast]`, so at least jump to definition points you in the right direction.

This one isn't possible due to the compiler consuming the `#[ast]` tokens before executing the proc-macro on the stream. Evne if it didn't I'm not sure that RA would understand it correctly.

>     2. Do the `macro_rules! generate_derive` trick you suggested, but put that in `oxc_ast/src/macros.rs`.

That's exactly what I did in my test. It works great but can only be used within `oxc_ast`, Since `oxc_ast` depends on both `oxc_syntax` and `oxc_span` so it isn't possible to access these, They form a cyclic dependency graph.

I'll try to think of another way to make it work.
I've already tried generating the dummy macro along with assertions for derive traits however RA doesn't respect it as it doesn't exist before expansion.

PR #4851, opened 2024-08-12
refactor(semantic): make SemanticBuilder opaque

DonIsaac:
change visibility of building-related methods and properties of SemanticBuilder
from `pub` to `pub(crate)`.

Issue #4188, opened 2024-07-11
AST Node ID


Boshen:
## Blocking issues:

* https://github.com/rolldown/rolldown/issues/1578
* https://github.com/oxc-project/oxc/pull/3780

## Related discussions:

* https://github.com/oxc-project/backlog/issues/30

## Previous attempts:

* https://github.com/oxc-project/oxc/issues/2818
* https://github.com/oxc-project/oxc/pull/2861

## Research:

* TODO
* rustc
* other compilers

overlookmotel:
I'm planning to look into this question as part of review of semantic. I've suggested a solution in the meantime on https://github.com/rolldown/rolldown/issues/1578 to unblock them.

rzvxa:
Some related stuff to this

The old issue for this: #2818
Previous attempt at this: #2861
Blocked PR by this missing ID in our own repo: #3780

rzvxa:
Another use case for `AST node ID`. We can remove these methods and free up a few CPU cycles and some memory while building the semantics.

https://github.com/oxc-project/oxc/blob/c65198fa152af560ced3fd2d8f88af1290612900/crates/oxc_semantic/src/builder.rs#L263-L276

Boshen:
Rolldown needs to implement `inject` plugin https://esbuild.github.io/api/#inject, which requires an identifiable import statement to be inserted into the AST.

overlookmotel:
I have suggested to them that they could use either:

1. Index of the statement in `program.body`.
2. Memory address of the `ImportDeclaration`.

```rs
fn addr_of_import(decl: &ImportDeclaration) -> usize {
    decl as *const _ as usize
}
```

Either of the above will work as a unique identifier, and neither depends on `span`.

I'm not saying that we shouldn't add `AstNodeId` to all AST types - maybe we should, maybe we shouldn't. But just that it's not the only way to solve Rolldown's use case.

Boshen:
> I have suggested to them that they could use either:
> 
> 1. Index of the statement in `program.body`.
> 2. Memory address of the `ImportDeclaration`.
> 
> ```rust
> fn addr_of_import(decl: &Box<ImportDeclaration>) -> usize {
>     decl.as_ref() as *const _ as usize
> }
> ```
> 
> Either of the above will work as a unique identifier, and neither depends on `span`.
> 
[...]

PR #4859, opened 2024-08-13
refactor(ast)!: Change order of fields in CallExpression

### LINKED ISSUE #4821: Change order of fields in `CallExpression`
`CallExpression` currently has `arguments` before `callee`:

https://github.com/oxc-project/oxc/blob/4e818a434ccc9100c4b28591e56166304fa590bc/crates/oxc_ast/src/ast/js.rs#L601-L608

Is there a reason for this? It seems like this field order would make more sense:

```rs
pub struct CallExpression<'a> { 
     #[serde(flatten)] 
     pub span: Span, 
     pub callee: Expression<'a>, 
     pub type_parameters: Option<Box<'a, TSTypeParameterInstantiation<'a>>>, 
     pub arguments: Vec<'a, Argument<'a>>, 
     pub optional: bool, // for optional chaining 
 } 
```

cblh:
fix: #4821


cblh:
This test case failed. https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/rules/eslint/no_this_before_super.rs#L384-L387

Dunqing:
> This test case failed.
> 
> https://github.com/oxc-project/oxc/blob/7a75e0f8a77006ef32fa0db45a329d961e70bdf3/crates/oxc_linter/src/rules/eslint/no_this_before_super.rs#L384-L387

Seems as expected, try to fix it

rzvxa:
This PR should be merged with #4851 to minimize our minor version bumps. Ideally, we'd get #4188 in this version too(at least the initial implementation).

crates/oxc_linter/src/rules/eslint/no_this_before_super.rs
```diff
@@ -246,6 +251,27 @@ impl NoThisBeforeSuper {
             }),
         })
     }
+
+    fn contains_this_or_super(arg: &Argument, semantic: &Rc<Semantic>) -> bool {
```
rzvxa, line 255: You are not using this `semantic` parameter, I think you can safely remove it.

rzvxa (changes requested):

crates/oxc_linter/src/rules/eslint/no_this_before_super.rs
cblh: The method `Self::contains_this_or_super_in_args(&call_expr.arguments, semantic)` is used to recursively handle cases like `super(a(b(this.c)))`.

rzvxa: You are absolutely right, Semantic is needed to be passed around, however, that function only uses its `semantic` to call `Self::contains_this_or_super`:

```rust
    fn contains_this_or_super_in_args(args: &[Argument], semantic: &Rc<Semantic>) -> bool {
        args.iter().any(|arg| Self::contains_this_or_super(arg, semantic))
    }
```

And `Self::contains_this_or_super` only uses it to call `Self::contains_this_or_super_in_args`.
I think you've gotten rid of the need for semantic usage however the compiler because of cyclic recursion doesn't notice this to provide a dead_code warning.

Please let me know if I'm missing something.

overlookmotel:
Fixed in #4851.

overlookmotel:
#4851 has already been merged, so I don't think we need to wait to merge this.

@cblh Could you please rebase on main to remove merge conflicts? (don't worry about fixing merge conflicts in `generated/derive_clone_in.rs` - just rebase then run `just ast` again to regenerate the file)

@rzvxa This currently shows as "change requested". If your requested changes have been made, please can you mark that as resolved?

rzvxa:
> #4851 has already been merged, so I don't think we need to wait to merge this.
> 
> @cblh Could you please rebase on main to remove merge conflicts? (don't worry about fixing merge conflicts in `generated/derive_clone_in.rs` - just rebase then run `just ast` again to regenerate the file)
> 
> @rzvxa This currently shows as "change requested". If your requested changes have been made, please can you mark that as resolved?

That one proved not to contain any breaking changes for our downstream, But I guess we can do the same thing with this PR as well.

I'll test it against rolldown and let you know. If it doesn't break anything there we may drop the bang in the title. I'll take care of merging it when CI is green once more.

@cblh We had some naming refactors recently, You should be able to make the CI green by running `just ast` and committing the changes.

cblh:
> > #4851 has already been merged, so I don't think we need to wait to merge this.
> > @cblh Could you please rebase on main to remove merge conflicts? (don't worry about fixing merge conflicts in `generated/derive_clone_in.rs` - just rebase then run `just ast` again to regenerate the file)
> > @rzvxa This currently shows as "change requested". If your requested changes have been made, please can you mark that as resolved?
> 
> That one proved not to contain any breaking changes for our downstream, But I guess we can do the same thing with this PR as well.
> 
> I'll test it against rolldown and let you know. If it doesn't break anything there we may drop the bang in the title. I'll take care of merging it when CI is green once more.
> 
> @cblh We had some naming refactors recently, You should be able to make the CI green by running `just ast` and committing the changes.

done

rzvxa:
Thank you❤️, I'll take care of the rest.

rzvxa (approved):

rzvxa:
@Boshen This PR contains a few breaking changes in the [rolldown/ast_snippet.rs](https://github.com/rolldown/rolldown/blob/main/crates/rolldown_ecmascript/src/ast_snippet.rs).

I'll let you decide on merging it now or waiting to add it with other breaking changes to avoid bumping the minor version multiple times.

It contains a few linter updates so merging would be more desirable to avoid conflicts, However, most of the diff is from the generated code and snapshots. Resolving conflicts here shouldn't be that hard.

Dunqing:
This change may affect the development of the transformer so I would like to merge this ASAP

Dunqing (approved):

PR #2865, opened 2024-03-30
chore: remove `oxc_transformer` for a reimplementation

Boshen:
closes #2860

Issue #4790, opened 2024-08-07
semantic: check symbols and scopes after transformation


Boshen:
cc @overlookmotel how exactly do we want to check symbols and scopes?

overlookmotel:
I would be happy to take this is on if you'd like, Boshen.

Dunqing:
Just a note, we may need to check duplicate AstNodeId, ScopeId, and SymbolId in `AST`. related to https://github.com/oxc-project/oxc/issues/4809

Boshen:
Borrowing the idea of the mangler, we can check whether symbols are the same in scope visitation order.

rzvxa:
> Just a note, we may need to check duplicate AstNodeId, ScopeId, and SymbolId in `AST`. related to #4809

This feels more related to #4804

PR #4876, opened 2024-08-13
feat(transformer): support exponentiation operator plugin

Dunqing:
The implementation copy from the original implementation which removed in https://github.com/oxc-project/oxc/pull/2865.

### DIFF
--- a/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
+++ b/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
@@ -1,0 +1,322 @@
+use std::cell::Cell;
+
+use oxc_allocator::{CloneIn, Vec};
+use oxc_ast::ast::*;
+use oxc_semantic::{ReferenceFlag, SymbolFlags};
+use oxc_span::SPAN;
+use oxc_syntax::operator::{AssignmentOperator, BinaryOperator};
+use oxc_traverse::TraverseCtx;
+
+use crate::context::Ctx;
+
+/// ES2016: Exponentiation Operator
+///
+/// References:
+/// * <https://babel.dev/docs/babel-plugin-transform-exponentiation-operator>
+/// * <https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-exponentiation-operator>
+/// * <https://github.com/babel/babel/blob/main/packages/babel-helper-builder-binary-assignment-operator-visitor>
+pub struct ExponentiationOperator<'a> {
+    _ctx: Ctx<'a>,
+    var_declarations: std::vec::Vec<Vec<'a, VariableDeclarator<'a>>>,
+}
+
+#[derive(Debug)]
+struct Exploded<'a> {
+    reference: AssignmentTarget<'a>,
+    uid: Expression<'a>,
+}
+
+impl<'a> ExponentiationOperator<'a> {
+    pub fn new(ctx: Ctx<'a>) -> Self {
+        Self { _ctx: ctx, var_declarations: vec![] }
+    }
+
+    fn clone_identifier_reference(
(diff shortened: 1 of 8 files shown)


Dunqing:
* **#4876** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/es2016/exponentiation_operator.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        // left **= right
+        if let Expression::AssignmentExpression(assign_expr) = expr {
+            if assign_expr.operator == AssignmentOperator::Exponential {
+                let mut nodes = ctx.ast.vec();
+                let Some(Exploded { reference, uid }) =
+                    self.explode(&mut assign_expr.left, &mut nodes, ctx)
+                else {
+                    return;
+                };
```
overlookmotel, line 98: Where left hand side is a member expression, I think we can avoid cloning it. We could instead get an owned copy of `left` here with `ast.move_expression(&assign_expr.left)`. Then insert that in new position, with no cloning.

```js
// In
obj.prop **= exp;

// Out
// NB: only 1 x `obj` - no cloning needed
var _obj;
(_obj = obj), (_obj.prop = Math.pow(_obj.prop, exp));
```

In above example, there's only 1 `obj` in the output, so no need for cloning. I think this also holds for all `MemberExpression` cases.

In the simpler case where left hand side is just an identifier, we *do* need to clone it:

```js
// In
value **= exp;

// Out
// NB: 2 x `value`s - cloning needed
var _value;
(_value = value), (value = Math.pow(_value, exp))
```

but cloning an identifier is much simpler than cloning an expression.

crates/oxc_transformer/src/es2016/exponentiation_operator.rs
```diff
@@ -0,0 +1,322 @@
...
+            statements.insert(0, Statement::VariableDeclaration(variable));
+        }
+    }
+
+    pub fn transform_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
+        // left ** right
+        if let Expression::BinaryExpression(binary_expr) = expr {
+            if binary_expr.operator == BinaryOperator::Exponential {
+                let left = ctx.ast.move_expression(&mut binary_expr.left);
+                let right = ctx.ast.move_expression(&mut binary_expr.right);
```
overlookmotel, line 85: Thank you for not using `ast.copy` here! ❤️

overlookmotel:
As discussed earlier, I wasn't planning to review these new transform PRs at this stage. I proposed that we merge them and then I do a second pass to review/tidy them later on. I only looked into this one because it informed #4880.

crates/oxc_transformer/src/es2016/exponentiation_operator.rs
Dunqing: Haha, If I still use `ast.copy`, it would cause tons of semantic errors after #4790

Dunqing:
> As discussed earlier, I wasn't planning to review these new transform PRs at this stage. I proposed that we merge them and then I do a second pass to review/tidy them later on. I only looked into this one because it informed #4880.

Let's merge all this stack of PRs first. We can make many improvements later

Dunqing:
> As discussed earlier, I wasn't planning to review these new transform PRs at this stage. I proposed that we merge them and then I do a second pass to review/tidy them later on. I only looked into this one because it informed #4880.
 
Thank you for your review, Let's merge all this stack of PRs first. We can make many improvements later

Issue #4804, opened 2024-08-10
Clone in scope API


overlookmotel:
#4732 introduced `CloneIn` trait.

Currently it clones `SymbolId`, `ScopeId` and `ReferenceId` along with all other parts of AST being cloned.

I don't think this is correct behavior, but am not sure what the correct behavior should be.

If you clone part of an AST which has been through `Semantic`, and insert it elsewhere in same AST, that AST will now contain duplicate IDs.

If you insert it into another AST, the IDs will be invalid. It can cause panic if try to look up the symbol for a `SymbolId`, if the destination AST has less symbols than the source AST, so `SymbolId` is out of bounds. And if it doesn't panic, it'll be the wrong symbol.

Ideally we should provide another API where you provide the destination AST's `ScopeTree` and `SymbolTable`, and what scope the cloned AST fragment will be inserted into, and then it can create new scopes/symbols/references, and patch up all the IDs in the clone automatically for you.

```rs
trait CloneInScope<'a>: Sized {
    type Cloned;
    fn clone_in_scope(
        &self,
        parent_scope_id: ScopeId,
        alloc: &'a Allocator,
        scopes: &mut ScopeTree,
        symbols: &mut SymbolTable,
    ) -> Self::Cloned;
}
```

But in meantime, to avoid invalid IDs, perhaps `CloneIn` should set `scope_id`, `symbol_id` and `reference_id` fields to `None`?

hyf0:
I guess this is also related to the semantics of `PartialEq, Eq, Hash` ... traits on `SymbolId, ScopeId, ReferenceId`.

When you do a comparison between two AST nodes, mostly you don't want to include the `SymbolId, ScopeId, ReferenceId` part.

rzvxa:
👍It is a good idea, I like it. At the very least we should unset those IDs.


I like to know your thoughts on this API. I'm not super familiar with the needs we have for scope tracking so it might be absolute rubbish.

Use this over a `CloneInScope` trait:

```rust
semantic.clone_from(node, scope_id);
```

Then in `Semantic::clone_from` we can clone the node and set its semantic information to turn it into a valid node. Here's how it would happen.

```rust
// where NodeT: CloneIn<T: Scoped>?
let cloned = node.clone_in(alloc); // all IDs are None.
// ...
self.scope_foreign_node(cloned, parent_scope_id); // sets IDs and updates scope via the trait `Scoped`?
```

[...]

PR #4880, opened 2024-08-13
feat(traverse): support `clone_identifier_reference` method in `TraverseCtx`

Dunqing:
related: #4804 

needs from: #4876 

The `clone_identifier_reference` method is used to clone an `IdentifierReference` and create a `Reference` and insert it to `SymbolTable`'s `resolved_references`.

The reason we need this is because we need to make sure that `IdentifierReference`'s `reference_id` is unique

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -326,6 +326,24 @@ impl TraverseScoping {
         let symbol_id = self.scopes.find_binding(self.current_scope_id, name.as_str());
         self.create_reference(name, symbol_id, flag)
     }
+
+    /// Clone `IdentifierReference` based on the original reference's `SymbolId` and name.
+    ///
+    /// This method makes a lookup of the `SymbolId` for the reference. If you need to create multiple
+    /// `IdentifierReference`s for the same binding, it is better to look up the `SymbolId` only once,
+    /// and generate `IdentifierReference`s with `TraverseScoping::create_reference_id`.
+    pub fn clone_identifier_reference<'a>(
+        &mut self,
+        ident: &IdentifierReference<'a>,
+        flag: ReferenceFlag,
+    ) -> IdentifierReference<'a> {
+        let reference =
+            self.symbols().get_reference(ident.reference_id.get().unwrap_or_else(|| {
+                unreachable!("IdentifierReference must have a reference_id");
+            }));
+        let symbol_id = reference.symbol_id();
+        self.create_reference_id(ident.span, ident.name.clone(), symbol_id, flag)
+    }
 }
 
 // Methods used internally within crate
--- a/crates/oxc_traverse/src/context/mod.rs
+++ b/crates/oxc_traverse/src/context/mod.rs
@@ -379,6 +379,21 @@ impl<'a> TraverseCtx<'a> {
     ) -> ReferenceId {
         self.scoping.create_reference_in_current_scope(name, flag)
     }
+
+    /// Clone `IdentifierReference` based on the original reference's `SymbolId` and name.
+    ///
+    /// This method makes a lookup of the `SymbolId` for the reference. If you need to create multiple
+    /// `IdentifierReference`s for the same binding, it is better to look up the `SymbolId` only once,
+    /// and generate `IdentifierReference`s with `TraverseCtx::create_reference_id`.
+    ///
+    /// This is a shortcut for `ctx.scoping.clone_identifier_reference`.
+    pub fn clone_identifier_reference(
+        &mut self,
+        ident: &IdentifierReference<'a>,
+        flag: ReferenceFlag,
+    ) -> IdentifierReference<'a> {
+        self.scoping.clone_identifier_reference(ident, flag)
+    }
 }
 
 // Methods used internally within crate


rzvxa:
Sorry if I'm a little out of the loop here, Out of interest I was wondering where we need 2 copies of the exact same node.
Since it is related to #4804 I'm interested to know about our use cases.
Doesn't it make more sense to have a generic API that would accept both `BindingIdentifier` and `IdentifierReference` and can output an `IdentifierReference`? Something like `ctx.create_reference_to(it)`.

Dunqing:
> Sorry if I'm a little out of the loop here, Out of interest I was wondering where we need 2 copies of the exact same node. Since it is related to #4804 I'm interested to know about our use cases.

For example: 
```js
// In
a.b **= 0

// Out
var _a;
(_a = a), (_a["b"] = Math.pow(_a["b"], 0));
```
The final result, in this case, is the `_a` at least needs to clone three times, and `a` need to clone once

>  Doesn't it make more sense to have a generic API that would accept both `BindingIdentifier` and `IdentifierReference` and can output an `IdentifierReference`? Something like `ctx.create_reference_to(it)`.

It would be better if we have an API to accept both `BindingIdentifier` and `IdentifierReference` and return an`IdentifierReference` which has the same `symbol_id`.

> 
> #### Edit:
> Is there a reason we have something like this in our code:
> 
> ```rust
>             let ident = ctx.create_reference_id(
>                 param.span,
>                 param.name.clone(),
>                 param.symbol_id.get(),
>                 ReferenceFlag::Read,
>             );
> ```
> 
> instead of passing `param` directly?

`IdentifierReference` only has `reference_id`. If we want to get a `symbol_id` through `reference_id`, so we need to get a `reference` by `reference_id` first, and then we can get `symbol_id` through `reference.symbol_id()`

Dunqing:
> > Sorry if I'm a little out of the loop here, Out of interest I was wondering where we need 2 copies of the exact same node. Since it is related to #4804 I'm interested to know about our use cases.
> 
> For example:
> 
> ```js
> // In
> a.b **= 0
> 
> // Out
> var _a;
> (_a = a), (_a["b"] = Math.pow(_a["b"], 0));
> ```
> 
> The final result, in this case, is the `_a` at least needs to clone three times, and `a` need to clone once
> 
> > Doesn't it make more sense to have a generic API that would accept both `BindingIdentifier` and `IdentifierReference` and can output an `IdentifierReference`? Something like `ctx.create_reference_to(it)`. 
> 
> It would be better if we have an API to accept both `BindingIdentifier` and `IdentifierReference` and return an`IdentifierReference` which has the same `symbol_id`. But different parameters have different implementation details
> 
> > #### Edit:
> > Is there a reason we have something like this in our code:
> > ```rust
> >             let ident = ctx.create_reference_id(
> >                 param.span,
> >                 param.name.clone(),
> >                 param.symbol_id.get(),
> >                 ReferenceFlag::Read,
> >             );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of passing `param` directly?
> 
> `IdentifierReference` only has `reference_id`. If we want to get a `symbol_id` through `reference_id`, so we need to get a `reference` by `reference_id` first, and then we can get `symbol_id` through `reference.symbol_id()`

Dunqing:
According to use case #4876, only `clone_identifier_reference` is not enough. `clone_expression` can be quite common.

All of these Expressions include `IdentifierReference`
```js
a.b // MemberExpression
a[b] // MemberExpression
`123 ${variable} 321` // TemplateLiteralExpression
```

Currently, I use `move_expression` instead of `clone_in`, but when an Expression may used to multi-place, then we must **clone** it

The API is best provided by `Traverse`, so we have to pass `ScopeTree` and `SymbolTable`

Dunqing:
I have written this code inside plugins, so does not block any yet

overlookmotel:
The problem with cloning an `IdentifierReference` like this is that it require an unnecessary lookup - you have to get the `Reference` from the `reference_id` to then get the `SymbolId`.

When you're generating multiple `IdentifierReference`s for the same symbol, it's better to look up the `SymbolId` only once, and then reuse it. Or, in this case, `_a` is a UID, so we have the `SymbolId` from the start.

### Another API

There is an API hidden in `helpers` which is good for this use case:

https://github.com/oxc-project/oxc/blob/ea1e64a1f378b2425a03cf88247a62f60a0ee04b/crates/oxc_transformer/src/helpers/bindings.rs#L12-L17

> ```js
> // In
> a.b **= 0
> 
> // Out
> var _a;
> (_a = a), (_a["b"] = Math.pow(_a["b"], 0));
> ```

In this case you could use it as follows:

```rs
// NB: `BindingIdentifier` + `IdentifierReference` type annotations are not required
// - just for clarity in this example

// Create `_a`
let binding = BoundIdentifier::new_uid(
    "a", // Name *without* `_` prefix
    ctx.current_scope_id(),
    SymbolFlags::FunctionScopedVariable,
    ctx
);
let id: BindingIdentifier = binding.create_binding_identifier();

// Then create your `IdentifierReference`s
let ref1: IdentifierReference = binding.create_read_reference(ctx);
let ref2: IdentifierReference = binding.create_read_reference(ctx);
let ref3: IdentifierReference = binding.create_read_reference(ctx);

// Now use them: `var <id>; (<ref1> = a), (<ref2>["b"] = Math.pow(<ref3>["b"], 0))`
```

Actually, this is not correct - `ref1` should have `ReferenceFlag::Write` flag not `Read`. We should add a `BoundIdentifier::create_write_reference` method. I'll make a PR for that now.

A couple of notes:

* The reason for having a `BoundIdentifier` type rather than just using `BindingIdentifier` is that:
  * `BoundIdentifier` is smaller as it doesn't include `Span`.
  * `BoundIdentifier` can have `symbol_id` stored as pure `SymbolId`, not `Cell<Option<SymbolId>>`.
* This `BoundIdentifier` API could probably be better - it could be methods on `TraverseCtx` instead - but I'm not sure if we should change it right now when we're in the middle of intense work on the transformer. What do you think?

### Should we merge this PR?

`clone_identifier_reference` is good if you only need to make a single clone, and you don't know the `SymbolId` already.

But the danger of having this API is that it's then tempting to use it in places where it'd be better to use a more efficient method like `BoundIdentifier`.

### Side note

> The final result, in this case, is the `_a` at least needs to clone three times, and `a` need to clone once.

I don't think `a` should be cloned at all. It should just be moved.

rzvxa:
@Dunqing Thanks for your explanation, Now it makes perfect sense to me.

I have an idea to introduce an `Identifier` trait that can unify these.
We can pass in an `Identifier` and get a `BoundIdentifier` from it, That way should be able to use it wherever we please.

To build upon @overlookmotel snippet it would look something like this:

```rust
// NB: `BindingIdentifier` + `IdentifierReference` type annotations are not required
// - just for clarity in this example

// where we already have `_a` binding identifier as `ident`:

let a_prime: IdentifierReference = ident.scoped_ref(ctx.scoping());

// when we want multiple copies from the same identifier
let binding: BoundIdentifier = ident.scoped_bound(ctx.scoping());

let ref1: IdentifierReference = binding.create_read_reference(ctx);
let ref2: IdentifierReference = binding.create_read_reference(ctx);
let ref3: IdentifierReference = binding.create_read_reference(ctx);


// Similarly if you have access to `_a` as `IdentfierReference` we can still write the exact same code:

let a_double_prime: IdentifierReference = ident_ref.scoped_ref(ctx.scoping());

// when we want multiple copies from the same identifier
let binding: BoundIdentifier = ident_ref.scoped_bound(ctx.scoping());

let ref1: IdentifierReference = binding.create_read_reference(ctx);
let ref2: IdentifierReference = binding.create_read_reference(ctx);
let ref3: IdentifierReference = binding.create_read_reference(ctx);
```

I'm not sure if it is the right time to do such a thing but I can't help but feel like having an `Identifier` trait can simplify a lot of hassles.

overlookmotel:
@rzvxa Your proposed trait is very elegant and flexible. But... my only worry about it is that it's maybe *too* flexible. Same as what I said above about `clone_identifier_reference` method - it makes it too easy to do the wrong (expensive) thing and create a `BoundIdentifier` from an `IdentifierReference` when you already have the `SymbolId`.

I actually think use cases for cloning existing `IdentifierReference`s are quite rare. The transform we've been talking about here is quite typical:

* The original `IdentifierReference` is just moved, and doesn't need to be cloned.
* The only binding that we *do* need to create multiple `IdentifierReference`s for is a UID which we have newly created. So `BoundIdentifier::new_uid` covers us for this.

Here's another transform which follows same pattern (only UIDs needs cloning):

```js
// In
const [x, y, z] = arr;

// Out
var _arr = _slicedToArray(arr, 3); // Helper call
const x = _arr[0], _arr[1], _arr[2];
```

Therefore, adding an API which makes it easy to do the wrong thing, and has few use cases where its the right thing may not on balance be a benefit overall.

The background context is that once we've laid the groundwork, we hope to open up the transformer to contributions (similar to the linter). People who are less familiar with Oxc will tend to just grab whatever API they find that makes the transform work, so there's an argument for limiting our API surface in order to guide people towards the "good path".

I could be wrong that there aren't many places where we'll need to clone existing `IdentifierReference`s. But I suggest we wait and see if we need this API, and only add it if we do.

overlookmotel:
@Dunqing Conversation here has broadened out beyond the content of this specific PR.

Do you think `BoundIdentifier` (especially once #4897 is merged) can cover this, so maybe we don't need `clone_identifier_reference` too?

Sorry I hid `BoundIdentifier` away in a silly place and didn't let anyone know it was there. I got lazy!

overlookmotel:
I was wrong! The exponent operator transform *does* requiring cloning `IdentifierReference`s in some cases. So we probably should add this `clone_identifier_reference` method for that.

Concerning the `Ident` trait idea: Personally I think it's better for this to be an explicitly-named method just for `IdentifierReference`s, rather than a cover-all trait, so potentially bad usages of this method are easier to spot.

crates/oxc_traverse/src/context/scoping.rs
```diff
@@ -326,6 +326,20 @@ impl TraverseScoping {
         let symbol_id = self.scopes.find_binding(self.current_scope_id, name.as_str());
         self.create_reference(name, symbol_id, flag)
     }
+
+    /// Clone IdentifierReference based on the original reference's symbol_id, name, and flag.
+    pub fn clone_identifier_reference<'a>(
```
overlookmotel, line 331: I suggest we add a warning as to what the "bad" usage of this method is:

```suggestion
    /// Clone `IdentifierReference` based on the original reference's `SymbolId`, name, and flag.
    ///
    /// This method makes a lookup of the `SymbolId` for the reference. If you need to create multiple
    /// `IdentifierReference`s for the same binding, it is better to look up the `SymbolId` only once,
    /// and generate `IdentifierReference`s with `TraverseCtx::create_reference_id`.
    pub fn clone_identifier_reference<'a>(
```

overlookmotel: Second thing: Should this method take a `flag` param?

e.g. in the exponent transform:

```js
// In
value **= exp;

// Out
var _value;
(_value = value), value = Math.pow(_value, exp);
```

the 1st `value` should have `Read` flag, and the 2nd `value` should have `Write` flag. Cloning the flag automatically makes it easy to forget this.

Dunqing: Yes, we need to pass a `flag` to it, What type would you expect a `flag` param? If it is `Option<ReferenceFlags>`, the default flag is obtained from `IdentifierReference`.

overlookmotel (approved):

PR #4883, opened 2024-08-13
chore(deps): remove `static_assertions` dependency.

rzvxa:


rzvxa:
* **#4883** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac:
What advantage does `assert!` have over static assertions?

Boshen:
> What advantage does `assert!` have over static assertions?

It's the same thing. We want less dependencies for the whole project hence the removal.

overlookmotel:
@DonIsaac FYI `static_assertions` allows const assertions i.e. assertions which run at compile time rather than runtime.

But Rust's built-in `assert!` now supports this too.

```rs
// This will cause a compilation error
const _: () = assert!(1 + 1 == 3);
```

Issue #4754, opened 2024-08-08
☂️ transformer: ES2024 - ES2016


Boshen:
Need to implement babel plugins equivalent of

https://github.com/evanw/esbuild/blob/9c13ae1f06dfa909eb4a53882e3b7e4216a503fe/compat-table/src/index.ts#L350-L399

- [ ] https://babel.dev/docs/babel-plugin-transform-class-properties
- [ ] https://babel.dev/docs/babel-plugin-transform-class-static-block
- [ ] https://babel.dev/docs/babel-plugin-transform-private-methods
- [ ] https://babel.dev/docs/babel-plugin-transform-private-property-in-object
- [ ] https://babel.dev/docs/babel-plugin-transform-logical-assignment-operators
- [ ] https://babel.dev/docs/babel-plugin-transform-nullish-coalescing-operator
- [ ] https://babel.dev/docs/babel-plugin-transform-optional-chaining
- [ ] https://babel.dev/docs/babel-plugin-transform-optional-catch-binding
- [ ] https://babel.dev/docs/babel-plugin-transform-object-rest-spread
- [ ] https://babel.dev/docs/babel-plugin-transform-exponentiation-operator

(The list may not be complete).

Boshen:
This is not a good first issue yet.

DonIsaac:
Can we add CJS to this list too? I need this for oxc-jest.

Boshen:
> Can we add CJS to this list too? I need this for oxc-jest.

CJS is in the milestone https://github.com/oxc-project/oxc/milestone/6

Boshen:
The list is not that bad actually, we end up with a few groups:

* regex features - transform to regex constructor
* class stuff - `ClassXXX`
* syntax sugars
* Async Generator
* Async Await

7086cmd:
While I generally favor the async/await syntax, it presents a challenging issue in injecting the necessary`polyfill` code through AstBuilder - the formatter pauses :)
.

overlookmotel:
We had `OptionalChain` ticked off the list as done. But actually it's not done yet!

Dunqing:
All plugins have been supported in oxc, Let's pay attention to new transformer issues.

DonIsaac:
🎉🎉🎉

PR #4890, opened 2024-08-14
feat(transformer): support logical-assignment-operators plugin

Dunqing:
part of #4754

The implementation copy from the original implementation which removed in https://github.com/oxc-project/oxc/pull/2865.


Dunqing:
I am going to merge this stack of PRs

Issue #4881, opened 2024-08-13
Standardize conventions for transforms


overlookmotel:
We are planning to build out a ton more transforms over next few months. The transformer is going to get a lot more complicated and unwieldy. To avoid it becoming hard to navigate and understand, I suggest that we adopt a standardized coding style to use on all transforms.

This is prompted in part from the difficulties I found in reviewing #4587.

I propose:

### 1. Make clear what are the "entry points" to transform

* Entry points of transform implemented as `impl Traverse for MyTransform`.
* Those methods have to be called `enter_*` and `exit_*`.
* Those entry points go at top of the file.
* Internal methods implemented lower down in an `impl MyTransform` block. Those methods can be named descriptively.

### 2. Encapsulate logic

All logic for each transform should live in that specific file, with no "leaking" into the parent transform. Each transform is only called into via the standard `enter_*`/`exit_*` entry points.

Only exception is that parent can check if child transform is enabled or not.

i.e. Instead of the parent being quite "fat":

https://github.com/oxc-project/oxc/blob/a9c729a1978ee6c3f43704eeef879acbce2f3d73/crates/oxc_transformer/src/react/mod.rs#L75-L87

Move the JSX-related logic into `jsx.rs`:

```rs
// src/react/mod.rs
fn enter_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) { 
     if self.jsx_plugin {
         self.jsx.enter_expression(self, expr, ctx);
     } 
 }
```

```rs
// src/react/jsx.rs
fn enter_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
    match expr {
        Expression::JSXElement(e) => {
            *expr = self.transform_jsx_element(e, ctx);
        }
        Expression::JSXFragment(e) => {
            *expr = self.transform_jsx_fragment(e, ctx);
        }
        _ => {}
    }
}
```

#### Top of file

Each transform should include comment at top of file including:

* High level explanation of what transform does.
* One "before / after" example.
* Link to Babel plugin.
* Note of any ways in which our implementation diverges from Babel's, and why.

#### Methods

If it's a complicated transform with multiple visitors which interact with each other, add comments explaining how the pieces fit together.

#### Code snippets

`AstBuilder` calls are often very verbose. Preface each chunk of `AstBuilder` calls with a short comment of what this code produces. e.g.:

```rs
// `let Foo;`
let declarations = {
[...]

PR #4899, opened 2024-08-14
docs(transformer): add README including style guide

overlookmotel:
Add a README to `oxc_transformer` crate, including a "style guide" for writing transforms. As discussed in #4881.

### DIFF
--- a/crates/oxc_transformer/README.md
+++ b/crates/oxc_transformer/README.md
@@ -1,0 +1,210 @@
+# Oxc transformer
+
+We mostly model Oxc's transforms on [Babel](https://babeljs.io/)'s implementations.
+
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#4899** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4899?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@Boshen and @Dunqing please could you review as closely as you can and suggest any changes?

I am not at all tied to the content of this document, very happy to make changes if either of you would prefer some things were different.

This is quite boring, but there would not be much point in adopting a style guide if we don't all agree with it, and consequently ignore it / diverge from it! It'd be preferable to work out any differences of opinion upfront.

Issue #4607, opened 2024-08-02
isolatedDeclaration strips module declarations if the file contains imports or exports


MichaelMitchell-at:
Tested version: 0.23.0

```ts
import {} from 'bar';
declare module 'foo' {
    interface Foo {}
    const foo = 42;
}
```

```
require('oxc-transform').isolatedDeclaration('hello.ts', `
  import "bar";
  declare module "foo" {
    interface Foo {}
    const foo = 42;
  }
`)
{ sourceText: 'import "bar";\nexport {};\n', errors: [] }
```

Expected:
```ts
import "bar";
declare module "foo" {
  interface Foo {}
  const foo = 42;
}
```

Actual:
```ts
import "bar";
export {};
```

MichaelMitchell-at:
Another case
```ts
function foo(): void {}
declare namespace foo {
    export let bar = 42;
}
foo.bar = 42;
export {foo};
```

DonIsaac:
I think #4592 addresses this, @MichaelMitchell-at could you please confirm?

MichaelMitchell-at:
> Another case

This was actually a separate issue. Addressed by https://github.com/oxc-project/oxc/pull/4910.

MichaelMitchell-at:
In working on https://github.com/oxc-project/oxc/pull/4911 I realized that the issue actually happens with any import, not just an import with the same name. Updated the title of this issue to reflect that.

Dunqing:
Reopen, because #4916

PR #3872, opened 2024-06-24
feat(isolated-declarations): report error for expando functions

Dunqing:
close: #3703

PR #4910, opened 2024-08-15
fix(isolated_declarations): Support expando functions

MichaelMitchell-at:
I mentioned this issue in a https://github.com/oxc-project/oxc/issues/4607#issuecomment-2264863618 but I later realized this is actually a separate issue from the one that I originally reported.

It seems that https://github.com/oxc-project/oxc/pull/3872 added support for reporting expando function errors, but didn't add support for cases where expando functions are allowed. This PR adds support for not reporting errors when there is a namespace declaration that declares the variable being assigned to.

[TypeScript playground](https://www.typescriptlang.org/play/?noCheck=true&isolatedDeclarations=true&ts=5.5.4#code/KYDwDg9gTgLgBAMwK4DsDGMCWEWIhACgEoAuOANwkwBM4BvAXwCgF8A6AQzDABsBPOAF44xIQD56zJqEiw4aHAGd4AIw5QhI0hSq1BExkzVQ2aDjx6bR+yUzszo8FBwC2wRWA5pgcAHIBleiY4OAc5BRRlOABzfCttShpxWxDYiDYeYBRomAALTQBGAAYmKTCnV3dPbzwIIJDM1XVNABYAJgBuYNDwRzhGuDUAL0Kioq6pVnTjVs6WdmHR8bsAehW4VHLgaiYIqKG44WIyRL0DZgP0mAh-GChMbPjkxiA) for reference

### DIFF
--- a/crates/oxc_isolated_declarations/src/lib.rs
+++ b/crates/oxc_isolated_declarations/src/lib.rs
@@ -404,7 +404,9 @@ impl<'a> IsolatedDeclarations<'a> {
         })
     }
 
-    pub fn report_error_for_expando_function(&self, stmts: &oxc_allocator::Vec<'a, Statement<'a>>) {
+    fn get_assignable_properties_for_namespaces(
+        stmts: &'a oxc_allocator::Vec<'a, Statement<'a>>,
+    ) -> FxHashMap<&'a str, FxHashSet<Atom>> {
         let mut assignable_properties_for_namespace = FxHashMap::<&str, FxHashSet<Atom>>::default();
         for stmt in stmts {
             let Statement::ExportNamedDeclaration(decl) = stmt else { continue };
@@ -465,6 +467,12 @@ impl<'a> IsolatedDeclarations<'a> {
                 }
             }
         }
+        assignable_properties_for_namespace
+    }
+
+    pub fn report_error_for_expando_function(&self, stmts: &oxc_allocator::Vec<'a, Statement<'a>>) {
+        let assignable_properties_for_namespace =
+            IsolatedDeclarations::get_assignable_properties_for_namespaces(stmts);
 
         let mut can_expando_function_names = FxHashSet::default();
         for stmt in stmts {
--- a/crates/oxc_isolated_declarations/src/lib.rs
+++ b/crates/oxc_isolated_declarations/src/lib.rs
@@ -26,7 +26,7 @@ use oxc_allocator::Allocator;
 use oxc_ast::{ast::*, AstBuilder, Visit};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_span::{Atom, SourceType, SPAN};
-use rustc_hash::FxHashSet;
+use rustc_hash::{FxHashMap, FxHashSet};
 
 use crate::scope::ScopeTree;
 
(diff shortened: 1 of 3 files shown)


MichaelMitchell-at:
FYI @Dunqing

crates/oxc_isolated_declarations/src/lib.rs
```diff
@@ -405,6 +405,67 @@ impl<'a> IsolatedDeclarations<'a> {
     }
 
     pub fn report_error_for_expando_function(&self, stmts: &oxc_allocator::Vec<'a, Statement<'a>>) {
+        let mut assignable_properties_for_namespace = FxHashMap::<&str, FxHashSet<Atom>>::default();
+        for stmt in stmts {
+            let Statement::ExportNamedDeclaration(decl) = stmt else { continue };
+            let Some(Declaration::TSModuleDeclaration(decl)) = &decl.declaration else { continue };
```
Dunqing, line 411: What about this case?
```ts
namespace NS { }
NS.a = 0;
export { NS }
```

crates/oxc_isolated_declarations/src/lib.rs
```diff
@@ -405,6 +405,67 @@ impl<'a> IsolatedDeclarations<'a> {
     }
 
     pub fn report_error_for_expando_function(&self, stmts: &oxc_allocator::Vec<'a, Statement<'a>>) {
+        let mut assignable_properties_for_namespace = FxHashMap::<&str, FxHashSet<Atom>>::default();
+        for stmt in stmts {
```
Dunqing, line 409: Looks like we don't need to iterate over `stmts` again. We have an iteration for collecting `can_expando_function_names`

crates/oxc_isolated_declarations/src/lib.rs
```diff
@@ -405,6 +405,67 @@ impl<'a> IsolatedDeclarations<'a> {
...
+                                    .entry(&ident.name)
+                                    .or_default()
+                                    .insert(name);
+                            }
+                        }
+                    }
+                    _ => {}
+                }
+            }
+        }
```
Dunqing, line 467: We need to place it in a new method.

Dunqing (changes requested):
Thank you for working on this!

crates/oxc_isolated_declarations/src/lib.rs
MichaelMitchell-at: In that iteration we also report the error. Since namespace declarations can come after the function declaration, we need to ensure that we've iterated over all statements first. Example:
```ts
export function foo(): void {}
foo.a = 42;  // When we get here we'll report an error because we did not see the namespace declaration yet.
export namespace foo {
    export let a: number;
}
```

crates/oxc_isolated_declarations/src/lib.rs
MichaelMitchell-at: Do you mean the case where there is no function declaration? I believe this is already allowed and handled correctly.

crates/oxc_isolated_declarations/src/lib.rs
MichaelMitchell-at: Ok done

crates/oxc_isolated_declarations/src/lib.rs
Dunqing: > Do you mean the case where there is no function declaration? I believe this is already allowed and handled correctly.

I mean namespace does not export. I saw you only handled `ExportNamedDeclartion`

crates/oxc_isolated_declarations/src/lib.rs
Dunqing: I see, Thanks

crates/oxc_isolated_declarations/src/lib.rs
MichaelMitchell-at: I see. This is consistent with TypeScript's behavior. If the namespace isn't exported, the error will still be reported.
[playground](https://www.typescriptlang.org/dev/bug-workbench/?target=7&noCheck=true&isolatedDeclarations=true&emitDeclarationOnly=true&ts=5.5.0-dev.20240502#code/PTAEAEGcBcCcEsDG0Bco4FcCmAoEEATLRAGwENYzp4B7AOzU133DpoGEALYga0dmx4w4LAFt40ACLFylavQDydEgE9+glvEg1y0LAWmkKVWnUjrmwsiRI0A7gClzoAGbXIliJE72AouOghLx87fwk9AgAxeBIsNDIAOgIE6EgcIPAXGKw6MlE40ETUnCwADwAHGlhoVww6ZFNXGhoACgBKNAA3GngCUABvAF88ACpQMsrq0BGwXPzIcrJELCaaAZxQcYqqmtiasjQ6DFEAIyxYAG4cYZdmhLJQAF5QABYAJiA)

Dunqing: Cool, Can you add a case to cover it?

MichaelMitchell-at: Yes, done

Dunqing (approved):

PR #4916, opened 2024-08-15
Revert "fix(isolated_declarations): Always emit module declarations (#4911)"

Dunqing:
This reverts commit 0fb0b71f0d2615c39e7d7bcee764915aa76ce2d0.

test failed  https://github.com/oxc-project/oxc/actions/runs/10405198969/job/28815418191

We should only emit for 
```ts
declare module "xx" {}
declare global {}
```

Do not emit for
```ts
module x {}
declare module x {}
```

@MichaelMitchell-at cc


MichaelMitchell-at:
@Dunqing sorry and thanks. I put up another fix here https://github.com/oxc-project/oxc/pull/4919

PR #4924, opened 2024-08-15
feat(linter): search for .oxlintrc.json when no --config is provided

### LINKED ISSUE #4600: Auto load config
I tried using `.oxlintrc.json` & `oxlintrc.json` and neither automatically loaded. Other tools tend to automatically load a config file if you're using the file name convention(s) they support.

DonIsaac:
Closes #4600

### DIFF
--- a/apps/oxlint/src/command/lint.rs
+++ b/apps/oxlint/src/command/lint.rs
@@ -1,4 +1,8 @@
-use std::{path::PathBuf, str::FromStr};
+use std::{
+    borrow::Cow,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use bpaf::Bpaf;
 use oxc_linter::{AllowWarnDeny, FixKind};
@@ -62,12 +66,47 @@ pub struct BasicOptions {
     /// Oxlint configuration file (experimental)
     ///  * only `.json` extension is supported
     ///  * tries to be compatible with the ESLint v8's format
+    ///
+    /// If not provided, Oxlint will look for the following files in the current
+    /// directory (in order):
+    ///
+    /// - `.oxlintrc.json`
+    /// - `oxlintrc.json`
+    /// - `.oxlint.json`
+    /// - `oxlint.json
     #[bpaf(long, short, argument("./oxlintrc.json"))]
-    pub config: Option<PathBuf>,
+    config: Option<PathBuf>,
 
     /// TypeScript `tsconfig.json` path for reading path alias and project references for import plugin
     #[bpaf(argument("./tsconfig.json"), hide_usage)]
     pub tsconfig: Option<PathBuf>,
+}
+
+impl BasicOptions {
+    const DEFAULT_OXLINTRC_NAMES: [&'static str; 4] =
+        [".oxlintrc.json", ".oxlint.json", "oxlintrc.json", "oxlint.json"];
+    pub fn config(&self) -> Option<Cow<'_, Path>> {
+        if let Some(config) = &self.config {
+            return Some(config.into());
+        }
+
+        let pwd = std::env::current_dir().ok()?;
+        for entry in std::fs::read_dir(&pwd).ok()? {
+            let Ok(entry) = entry else {
+                continue;
+            };
+            let name_string = entry.file_name();
+            let path = entry.path();
+            let Some(name) = name_string.to_str() else {
+                continue;
+            };
+            if Self::DEFAULT_OXLINTRC_NAMES.contains(&name) && path.is_file() {
+                return Some(Cow::Owned(path));
+            }
+        }
+
+        None
+    }
 }
 
 // This is formatted according to
--- a/apps/oxlint/src/lint/mod.rs
+++ b/apps/oxlint/src/lint/mod.rs
@@ -92,7 +92,7 @@ impl Runner for LintRunner {
         let cwd = std::env::current_dir().unwrap().into_boxed_path();
         let lint_options = LintOptions::default()
             .with_filter(filter)
-            .with_config_path(basic_options.config)
+            .with_config_path(basic_options.config().map(Into::into))
             .with_fix(fix_options.fix_kind())
             .with_react_plugin(enable_plugins.react_plugin)
             .with_unicorn_plugin(enable_plugins.unicorn_plugin)
--- a/tasks/website/src/linter/snapshots/cli.snap
+++ b/tasks/website/src/linter/snapshots/cli.snap
@@ -13,6 +13,10 @@ expression: snapshot
   Oxlint configuration file (experimental)
 * only `.json` extension is supported
 * tries to be compatible with the ESLint v8's format
+
+  If not provided, Oxlint will look for the following files in the current directory (in order):
+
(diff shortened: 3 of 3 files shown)


DonIsaac:
* **#4924** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

apps/oxlint/src/command/lint.rs
```diff
@@ -62,14 +66,49 @@ pub struct BasicOptions {
...
+    config: Option<PathBuf>,
 
     /// TypeScript `tsconfig.json` path for reading path alias and project references for import plugin
     #[bpaf(argument("./tsconfig.json"), hide_usage)]
     pub tsconfig: Option<PathBuf>,
 }
 
+impl BasicOptions {
+    const DEFAULT_OXLINTRC_NAMES: [&'static str; 4] =
+        [".oxlintrc.json", "oxlintrc.json", ".oxlint.json", "oxlint.json"];
```
rzvxa, line 87: Maybe we should add ESLint config names as well. What do you think?

I also think `.oxlint.json` can be higher than `oxlintrc.json`, I have no heuristic to back this up but I feel most projects use hidden files for these configurations.

rzvxa:
Both of our editor extensions are providing a default `.eslintrc` so this wouldn't work as intended in IDEs.

https://github.com/oxc-project/oxc/blob/4081293ce7190f0d1675f82528220923e59c5a84/editors/vscode/package.json#L97-L102

https://github.com/oxc-project/coc-oxc/blob/63f2682fa39c7afb39cc2d248dacb23e04545a74/package.json#L54-L59

We should follow this PR with 2 more to address this change. The `coc-oxc` release pipeline is a bit finicky - since I just bootstrapped it using one of my old CI actions - But anything that bumps the version and goes into the main would trigger a release.

I know you are OK with changing the VSCode one but in case you don't want to bother with the Vim(or even VSCode if you have other tasks in mind) Just let me know so I can take over this small change.

apps/oxlint/src/command/lint.rs
DonIsaac: I'm not doing the former based on Gilad's desire to have separate configs. He wants separation between oxlint/eslint configuration. Thoughts?

DonIsaac:
I'll add a PR to this stack updating the VSCode defaults, but we'll need a separate PR for the CoC plugin

apps/oxlint/src/command/lint.rs
DonIsaac: Also, should we support `.oxlintrc`?

rzvxa: > I'm not doing the former based on Gilad's desire to have separate configs. He wants separation between oxlint/eslint configuration. Thoughts?

Good point; I'm not sure about this, Maybe we can have it at the very end of our check so we only use it if there is no oxlint specific file. I don't know exactly what is our plan to satisfy their configuration requirements, That's why I also think it can wait until we sort the fundamentals out.

> Also, should we support .oxlintrc?

Nice catch, The `.eslint` file is common in the wild so that seems like a nice addition to avoid breaking people's habits. It's bad UX if someone does this which is normal in eslint land and doesn't get the expected behavior or - if we don't want to allow non-json configs as our default files - get a warning.

apps/oxlint/src/command/lint.rs
```diff
@@ -62,14 +66,49 @@ pub struct BasicOptions {
     /// Oxlint configuration file (experimental)
     ///  * only `.json` extension is supported
     ///  * tries to be compatible with the ESLint v8's format
+    ///
+    /// If not provided, Oxlint will look for the following files in the current
+    /// directory (in order):
+    ///
+    /// - `.oxlintrc.json`
+    /// - `oxlintrc.json`
```
rzvxa, line 74: I can't understand how we can guarantee the order here, What am I missing?
If we don't search with the order in mind we can drop the `(in order)` part.

apps/oxlint/src/lint/mod.rs
```diff
@@ -92,7 +92,7 @@ impl Runner for LintRunner {
         let cwd = std::env::current_dir().unwrap().into_boxed_path();
         let lint_options = LintOptions::default()
             .with_filter(filter)
-            .with_config_path(basic_options.config)
+            .with_config_path(basic_options.config().map(Into::into))
```
rzvxa, line 95: I've looked into the LSP, I'm working on a PR updating our LSP to take advantage of this feature. Is it possible to provide a way to resolve the path in the `with_config_path` instead of doing it in the `basic_options`?

It allows me to pass in a `None` value here and let this logic resolve it.

https://github.com/oxc-project/oxc/blob/9c64b1296307cfc39972a20768c7c97b287469c9/crates/oxc_language_server/src/main.rs#L350

rzvxa: Maybe adding a dedicated setter would be better, Or even providing a function that is both public and is used in the `basic_options` implementation so we can share that code.

apps/oxlint/src/command/lint.rs
```diff
@@ -62,14 +66,49 @@ pub struct BasicOptions {
...
+impl BasicOptions {
+    const DEFAULT_OXLINTRC_NAMES: [&'static str; 4] =
+        [".oxlintrc.json", ".oxlint.json", "oxlintrc.json", "oxlint.json"];
+    pub fn config(&self) -> Option<Cow<'_, Path>> {
+        if let Some(config) = &self.config {
+            return Some(config.into());
+        }
+
+        let pwd = std::env::current_dir().ok()?;
+        for entry in std::fs::read_dir(&pwd).ok()? {
```
Boshen, line 94: I'm not sure about the performance overhead of this :-/

Boshen (commented):
This behavior differs from eslint, will this give a surprise? Is implicit better than explicit here?

apps/oxlint/src/command/lint.rs
rzvxa: We can change this to concat our paths and only check the fs stat of those few paths.

I'm more interested in this comment of yours:

> This behavior differs from eslint, will this give a surprise? Is implicit better than explicit here?

I feel it's useful for LSP, How does it differ from eslint? Doesn't eslint use the cwd to find the config if none is provided? In your opinion doing this for the CLI is a bad diversion?

Can't we make this work with something like this: https://eslint.org/docs/latest/use/command-line-interface#--no-eslintrc

I can think of situations where people would want to omit the config flag to run the linter with the default configuration, But most of the time that's not the case. I'm not sure about the implications of this change so please share your thoughts on it.

DonIsaac:
ESLint will default to one of the various .eslintrc file names. This is equivalent behavior, but for our config file. We could also look for an .eslintrc (or other file name) if no oxlint config is found.

Boshen:
I still think we should conform to what eslint does, because all our users come from eslint.

We can merge this if we can reference this feature back to eslint.

Boshen:
Let's revisit this once we improve our config logic.

PR #4934, opened 2024-08-16
fix(isolated_declarations): Class properties should still be lifted from private constructors

MichaelMitchell-at:
Before, this
```ts
export class Bux {
  private constructor(
    public readonly prop: number = 0,
    private readonly prop2: number = 1,
    readonly prop3: number = 1,
  ) {}
}
```
would be emitted as 
```ts
export declare class Bux {
	private constructor();
}
```

Now it will be emitted as
```ts
export declare class Bux {
	readonly prop: number;
	private readonly prop2;
	readonly prop3: number;
	private constructor();
}
```


Dunqing (approved):
Thank you!

PR #4935, opened 2024-08-16
fix(isolated_declarations): namespaces that are default exported should be considered for expando functions

MichaelMitchell-at:
This should be ok but currently has an error reported
```ts
function foo(): void {}

namespace foo {
  export let bar = 42;
}

foo.bar = 42;

export default foo;
```


crates/oxc_isolated_declarations/tests/fixtures/expando-function.ts
```diff
@@ -16,14 +16,19 @@ export namespace foo {
   export let baz = 100;
 }
 
-// namespace must be exported
-namespace foo {
-  export let bar = 42;
-}
```
MichaelMitchell-at, line 22: TS will report an error for `foo.bar = 42` below, but oxc won't report an error now. This is still actually a typecheck error, `Individual declarations in merged declaration 'foo' must be all exported or all local.` so I feel like not having an error in declaration emit is fair game. Arguably that typecheck error should be made into an isolated declaration error, but that's not something oxc needs to worry about until TS implements that.

DonIsaac (approved):

PR #4937, opened 2024-08-16
fix(codegen): add missing `declare` to `PropertyDefinition`

Boshen:
I'm seeing a broken test for 

```rust
    #[test]
    fn dts_class_decl_prop_test() {
        transform_dts_test(
            "export class Foo { declare a: string }",
            "export declare class Foo {
  a: string;
}",
        );
    }
```

can you double check @Dunqing ?


Dunqing:
It is a bug in isolated_declarations.

PR #4938, opened 2024-08-16
chore(coverage): print path relative to snapshot file

Boshen:
Ctrl + click in terminal can open the file with this change.

I don't know about vscode, but maybe a plugin can help?

image

image


DonIsaac (approved):

DonIsaac:
Trying to catch back up with @overlookmotel's diffs? 😆 
image

PR #4945, opened 2024-08-16
fix(linter/react): Fixed false positive with missing key inside React.Children.toArray()

### LINKED ISSUE #4421: eslint-plugin-react(jsx-key) - false positive with missing key inside React.Children.toArray()
This triggers an error with oxlint:

```
{React.Children.toArray(items.map((item) => {
  return (
    <div>
      {item}
    </div>
  );
}))}
```

But it does not with eslint. When using .map inside React.Children.toArray() a key should not be necessary (to my understanding)

ApollonDeParnasse:
Closes: #4421 

Added three new functions in order to check if an element is inside of React.Children.toArray.
Tests added for this fix are exactly the same as those used by eslint-plugin-react.


crates/oxc_linter/src/rules/react/jsx_key.rs
```diff
@@ -400,6 +444,29 @@ fn test() {
...
+
+        Act.Children.toArray([1, 2 ,3].map(x => <App />));
+        Act.Children.toArray(Array.from([1, 2 ,3], x => <App />));
+        Children.toArray([1, 2 ,3].map(x => <App />));
+        Children.toArray(Array.from([1, 2 ,3], x => <App />));
+        ",
+        r"React.Children.toArray([1, 2 ,3].map(x => <App />));",
     ];
 
     let fail = vec![
```
mysteryven, line 472: Could you add a test case `foo.Children.toArray([1, 2 ,3].map(x => <App />));'`? This should fail.

DonIsaac (approved):

DonIsaac:
@mysteryven feel free to merge when you're satisfied.

vHugoObject:
The code needs to be updated for the new test but it will finish it ASAP.

DonIsaac:
@vHugoObject need some help getting this PR over the finish line?

vHugoObject:
Sorry, I got it, I was just busy the last two days.

vHugoObject:
Added the new test and then added new functions so that is_children always verifies imports.

mysteryven (approved):

mysteryven:
Thank you!

PR #4951, opened 2024-08-18
feat(linter/eslint-plugin-vitest): implement require-local-test-context-for-concurrent-snapshots

shulaoda:
Related to #4656


crates/oxc_linter/src/rules/vitest/require_local_test_context_for_concurrent_snapshots.rs
```diff
@@ -0,0 +1,173 @@
...
+    ///     expect(true).toMatchSnapshot();
+    /// })
+    ///
+    /// test.concurrent('myLogic', (context) => {
+    ///     context.expect(true).toMatchSnapshot();
+    /// })
+    /// ```
+    RequireLocalTestContextForConcurrentSnapshots,
+    correctness,
+    suggestion
```
mysteryven, line 83: Nit: we can remove this as it doesn't have a fix.

https://github.com/oxc-project/oxc/blob/a266b4516784eb03e4610b5dc55d0248dc9f554c/tasks/rulegen/template.txt#L37-L39

mysteryven (approved):
Thank you!

Issue #4455, opened 2024-07-23
RFC: Unify crate public APIs


Boshen:
During flights without internet access, I was confused about all the inconsistencies from our crates' public APIs (`oxc_parser`, `oxc_transformer`, `oxc_minifier` etc.) - builder pattern or options pattern?

I'd like to research different Rust public API patterns, so we can have one less thing to do for v1.0.

Boshen:
Oh yes, I forgot to mention another goal is to enable `cargo-semver-checks` after v1.0.0.

obi1kenobi:
Woo! 🎉 Feel free to ping me anytime if you run into any issues with it, or just want a second pair of eyes on anything!

milesj:
Just chiming in here, but I personally prefer builders that use `&mut` instead of immutable `mut` methods. Primarily because the latter makes calling methods in conditionals or blocks _very_ awkward.

hyf0:
https://github.com/oxc-project/oxc/issues/3485 Could be related. Kind of hope the API is friendly for compiler to checkout breaking changes.

Boshen:
We currently have an [`oxc`](https://docs.rs/oxc/0.24.2/oxc/) crate for unifying all the sub-crates.

I plan to expose a `Compiler` API similar to `rustc_driver` and `rustc_interface`.

The `Compiler` API will define the whole pipeline for `source_text` -> `source_text` transformation, i.e. `parser` -> `semantic` -> `transform` -> `minify` -> `codegen`.

Every step should have an before and after interceptor (plugin if you will).

When a user open the docs page of the `oxc` crate, they should be able to copy the example and customize.

Options API seems to be easier for this to work, compared to a builder pattern.

Here's the building block I propose, so I can start changing all the underly crates to be consistent with this pattern:

```rust
struct CompilerOptions {
  parser: ParserOptions,
  semantic: Option<SemanticOptions>,
  transform: Option<TransformOptions>,
  minify: Option<MinifyOptions>,
  codegen: Option<CodegenOptions>
}

struct Compiler {
}

impl Compiler {
  pub fn new(options: CompilerOptions) {
  }
  
  pub fn run(self) {
  }
}
```

Boshen:
I decided to not include any builder patterns in the public APIs, they are super confusing for new comers from JavaScript, who are used to passing objects with spread `{ foo: true, ...options }`.

---

All the tools should conform to

```rust
// `OxcDiagnostic` can also contain warnings.
[...]

PR #4954, opened 2024-08-18
feat(oxc): add `Compiler` and `CompilerInterface`

Boshen:
This PR adds a full compiler pipeline to the `oxc` crate, to stop us from implementing the same pipeline over and over again.

relates #4455

### DIFF
--- a/crates/oxc/src/compiler.rs
+++ b/crates/oxc/src/compiler.rs
@@ -1,0 +1,225 @@
+use std::{mem, ops::ControlFlow, path::Path};
+
+use oxc_allocator::Allocator;
+use oxc_ast::{ast::Program, Trivias};
+use oxc_codegen::{CodeGenerator, CodegenOptions, CommentOptions, WhitespaceRemover};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_parser::{ParseOptions, Parser, ParserReturn};
+use oxc_span::SourceType;
+
+use oxc_minifier::{CompressOptions, Compressor};
+use oxc_semantic::{ScopeTree, SemanticBuilder, SemanticBuilderReturn, SymbolTable};
+use oxc_transformer::{TransformOptions, Transformer, TransformerReturn};
+
+#[derive(Default)]
+pub struct Compiler {
+    printed: String,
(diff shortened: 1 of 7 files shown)


tasks/coverage/parser_babel.snap
```diff
@@ -3,7 +3,7 @@ commit: 12619ffe
 parser_babel Summary:
 AST Parsed     : 2093/2101 (99.62%)
 Positive Passed: 2083/2101 (99.14%)
-Negative Passed: 1380/1501 (91.94%)
+Negative Passed: 1367/1501 (91.07%)
```
Boshen, line 6: I'll investigate when I wake up.

crates/oxc/src/compiler.rs
DonIsaac, line 1: How are we going to handle multi-file outputs, such as source maps and ID?

crates/oxc/src/compiler.rs
```diff
@@ -0,0 +1,226 @@
...
+        &self,
+        program: &Program<'a>,
+        source_text: &'a str,
+        source_type: SourceType,
+        trivias: &Trivias,
+    ) -> SemanticBuilderReturn<'a> {
+        SemanticBuilder::new(source_text, source_type)
+            .with_trivias(trivias.clone())
+            .with_check_syntax_error(true)
+            // .build_module_record(source_path.to_path_buf(), &program)
```
DonIsaac, line 177: This would be very useful to have

tasks/coverage/parser_test262.snap
```diff
@@ -3,11 +3,18 @@ commit: a1587416
 parser_test262 Summary:
 AST Parsed     : 45859/45859 (100.00%)
 Positive Passed: 45859/45859 (100.00%)
-Negative Passed: 3925/3929 (99.90%)
+Negative Passed: 3918/3929 (99.72%)
```
DonIsaac, line 6: What happened here?

crates/oxc/src/compiler.rs
Boshen: Not yet, this is for setting up the most basic things.

DonIsaac:
This is a clean abstraction, I'm in favor of it.

PR #4956, opened 2024-08-18
feat(linter/eslint-plugin-vitest): implement no-restricted-vi-methods

shulaoda:
Related to #4656


mysteryven:
These two rule's difference are trivial, can we combine them in one file?

https://github.com/oxc-project/oxc/pull/4182/files

What's more, the snapshot can be split by using [`with_snapshot_suffix`](https://docs.rs/insta/latest/insta/struct.Settings.html#method.set_snapshot_suffix).

shulaoda:
> These two rule's difference are trivial, can we combine them in one file?

Where should we place the combined rules? `jest` or `vitest`? What should the combined name be?

mysteryven:
> Where should we place the combined rules? `jest` or `vitest`? What should the combined name be?

The original jest rule file is ok. #4182 maybe helpful for you, the initial PR of this serial is #3717  👀

mysteryven (approved):
Thank you!

PR #4971, opened 2024-08-19
feat(linter/eslint-plugin-vitest): implement no-conditional-in-test

shulaoda:
Related to #4656


mysteryven:
Do you want to implement jest/no-conditional-in-test too? It can be placed in path `rules/jest` 👀

shulaoda:
> Do you want to implement jest/no-conditional-in-test too? It can be placed in path `rules/jest` 👀

Of course, it seems that their logic is the same, I just need to change the file position and map it?

mysteryven:
> I just need to change the file position and map it?
 
 I think so.

mysteryven (approved):
Nice! 👍🏻

PR #4979, opened 2024-08-19
feat(oxc_minifier): define plugin with postfix wildcard

IWANABETHATGUY:
1. support `import.meta.env.*` in `ReplaceGlobalDefine`

### DIFF
--- a/crates/oxc_minifier/src/plugins/replace_global_defines.rs
+++ b/crates/oxc_minifier/src/plugins/replace_global_defines.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::sync::Arc;
 
 use oxc_allocator::Allocator;
@@ -18,8 +19,9 @@ pub struct ReplaceGlobalDefinesConfig(Arc<ReplaceGlobalDefinesConfigImpl>);
 
 #[derive(Debug)]
 struct ReplaceGlobalDefinesConfigImpl {
-    identifier_defines: Vec<(/* key */ CompactStr, /* value */ CompactStr)>,
-    dot_defines: Vec<DotDefine>,
+    identifier: Vec<(/* key */ CompactStr, /* value */ CompactStr)>,
+    dot: Vec<DotDefine>,
+    meta_proeperty: Vec<MetaPropertyDefine>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


crates/oxc_minifier/tests/plugins/replace_global_defines.rs
```diff
@@ -42,3 +42,34 @@ fn replace_global_definitions_dot() {
...
+#[test]
+fn replace_global_definitions_dot_with_postfix_wildcard() {
+    {
+        let config = ReplaceGlobalDefinesConfig::new(&[
+            ("process.env.NODE_ENV.*", "production"),
+            ("import.meta.env.*", "undefined"),
+        ])
+        .unwrap();
+        test("process.env.NODE_ENV.test", "production", config.clone());
+        // test("process.env", "process.env", config.clone());
```
Boshen, line 55: What are these commented out tests?

crates/oxc_minifier/src/plugins/replace_global_defines.rs
```diff
@@ -157,39 +180,73 @@ impl<'a> ReplaceGlobalDefines<'a> {
...
+                has_matched_part = true;
+            }
         }
 
-        true
+        has_matched_part
     }
 }
+
+enum StaticMemberExpressionOrMeta<'a> {
```
Boshen, line 249: I think it would be easier to split out dot define and `MetaProperty` in the config instead of mixing them here.

IWANABETHATGUY: Not sure if I understand correctly, do you mean add another field `MetaProperty` in https://github.com/oxc-project/oxc/blob/c65e8fc1ad1c5b71b6a16de44dd2246b293b0660/crates/oxc_minifier/src/plugins/replace_global_defines.rs#L40-L43?

Boshen: yes, because it will be used to test against `MetaProperty` only.

IWANABETHATGUY: Do you mean we only support `import.meta.x.*` syntax rather than general memberExpr wildcard postfix match?
like `process.env.*`?

IWANABETHATGUY: Ok

Boshen: We should only support `process.env.*` if there is a requirement, which is none so far.

crates/oxc_minifier/tests/plugins/replace_global_defines.rs
```diff
@@ -42,3 +42,26 @@ fn replace_global_definitions_dot() {
...
+    {
+        let config = ReplaceGlobalDefinesConfig::new(&[
+            ("import.meta.env.*", "undefined"),
+            ("import.meta.env", "env"),
+        ])
+        .unwrap();
+        test("import.meta.env.result", "undefined", config.clone());
+        test("import.meta.env", "env", config);
+    }
+}
```
Boshen, line 67: It seems like we are lacking tests that don't match the expression, can we add a few more to cover all the lines?

Boshen (commented):
A few more tests and we are ready to merge.

Boshen (approved):

PR #4920, opened 2024-08-15
fix(transformer/typescript): typescript syntax within `SimpleAssignmentTarget` with `MemberExpressions` is not stripped

Dunqing:
close: #4870

I added the `move_identifier_reference` and `move_member_expression` methods used to take ownership in `ast_builder_impl`.  This way can let us get rid of `ast.copy`.

Another possible approach is to add `get_expression_owner` to `SimpleAssignmentTarget` and a `get_inner_expression_owner` method to `Expression`. And add an `into_xxxxx` method for `inherit_variants` macro

The implementation looks like this
```rs
let Some(expression) = self.get_expression_owner() else { return; }
match expr.get_inner_expression_owner() {
    Expression::Identifier(ident) => {
        *target = self.ctx.ast.simple_assignment_target_from_identifier_reference(ident);
    }
    inner_expr @ match_member_expression!(Expression) => {
        *target = SimpleAssignmentTarget::from(
            inner_expr.into_member_expression()
        );
    }
    _ => (),
}
```

PR #4982, opened 2024-08-19
refactor(transformer): TS annotations transform use `move_expression`

overlookmotel:
Follow on after #4920. #4920 removed a usage of `AstBuilder::copy` and replaced with 2 new methods `AstBuilder::move_identifier_reference` and `AstBuilder:: move_member_expression`.

We can instead use `AstBuilder::move_expression` earlier and then unpack the enum again. Hopefully the compiler can see that the 2 `unreachable!()` branches are indeed unreachable and elide them.

`move_expression` is preferable to `move_member_expression` as it only creates 1 temporary `Box<Expression::NullLiteral>` instead of 2.

### DIFF
--- a/crates/oxc_ast/src/ast_builder_impl.rs
+++ b/crates/oxc_ast/src/ast_builder_impl.rs
@@ -69,33 +69,11 @@ impl<'a> AstBuilder<'a> {
         unsafe { std::mem::transmute_copy(src) }
     }
 
-    /// Moves the identifier reference out by replacing it with a dummy identifier reference.
-    #[inline]
-    pub fn move_identifier_reference(
-        self,
-        expr: &mut IdentifierReference<'a>,
-    ) -> IdentifierReference<'a> {
-        let dummy = self.identifier_reference(expr.span(), Atom::empty());
-        mem::replace(expr, dummy)
-    }
-
     /// Moves the expression out by replacing it with a null expression.
     #[inline]
     pub fn move_expression(self, expr: &mut Expression<'a>) -> Expression<'a> {
         let null_expr = self.expression_null_literal(expr.span());
         mem::replace(expr, null_expr)
-    }
-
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#4982** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@Dunqing I think the APIs you suggested in #4920 could be better than this, but I'm not 100% sure whether we need to add something to `inherit_variants!` macro to support it or not. So in meantime I thought best to remove the `move_member_expression` API, as it's relatively expensive (see above).

The equivalent of `to_member_expression` for matching owned expressions is:

```rs
let expr: Expression = get_expression_somehow();
match expr {
    match_member_expression!(Expression) => {
        let member_expr = MemberExpression::try_from(expr).unwrap();
    }
}
```

We could add a `MemberExpression::from` method which is just `MemberExpression::try_from(expr).unwrap()` internally, but I'm not sure if it's a good idea, as it suggests this operation is infallible when in fact it can panic. Maybe `into_member_expression` for an owned `Expression`?

Beyond that, do we need changes to the APIs provided by `inherit_variants!` to implement your `*_owned` APIs? (I'd suggested `*_owned` rather than `*_owner` for naming)

NB: `move_identifier_reference` is unnecessary as `IdentifierReference` is `Clone` anyway - no need to replace the existing one. I prefer to use `move_expression` here and then construct a `SimpleAssignmentTarget::AssignmentTargetIdentifier(ident)` because `ident` is `Box<IdentifierReference>` so we don't need to rebox it (vs using `clone` where we would have to).

Dunqing (approved):

Dunqing:
Looks like `try_from` is enough for now

PR #4983, opened 2024-08-19
fix(transformer): remove an `AstBuilder::copy` call

overlookmotel:
Remove one usage of unsound `AstBuilder::copy` method, using the new `get_inner_expression_mut` method introduced in #4920.

PR #4984, opened 2024-08-19
fix(transformer): remove an `AstBuilder::copy` call for TS `AssignmentTarget` transform

overlookmotel:
Add `AssignmentTarget::get_expression_mut` method and use it to remove an unsound `AstBuilder::copy` call from TS transform for `AssignmentTarget`s.

### DIFF
--- a/crates/oxc_transformer/src/typescript/annotations.rs
+++ b/crates/oxc_transformer/src/typescript/annotations.rs
@@ -250,10 +250,14 @@ impl<'a> TypeScriptAnnotations<'a> {
     }
 
     pub fn transform_assignment_target(&mut self, target: &mut AssignmentTarget<'a>) {
-        if let Some(expr) = target.get_expression() {
-            if let Some(member_expr) = expr.get_inner_expression().as_member_expression() {
-                // SAFETY: `ast.copy` is unsound! We need to fix.
-                *target = AssignmentTarget::from(unsafe { self.ctx.ast.copy(member_expr) });
+        if let Some(expr) = target.get_expression_mut() {
+            let inner_expr = expr.get_inner_expression_mut();
+            if inner_expr.is_member_expression() {
+                let inner_expr = self.ctx.ast.move_expression(inner_expr);
+                let Ok(member_expr) = MemberExpression::try_from(inner_expr) else {
+                    unreachable!();
+                };
+                *target = AssignmentTarget::from(member_expr);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4984?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4984** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4983** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/4983?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_transformer/src/typescript/annotations.rs
```diff
@@ -246,10 +246,14 @@ impl<'a> TypeScriptAnnotations<'a> {
...
     pub fn transform_assignment_target(&mut self, target: &mut AssignmentTarget<'a>) {
-        if let Some(expr) = target.get_expression() {
-            if let Some(member_expr) = expr.get_inner_expression().as_member_expression() {
-                // SAFETY: `ast.copy` is unsound! We need to fix.
-                *target = AssignmentTarget::from(unsafe { self.ctx.ast.copy(member_expr) });
+        if let Some(expr) = target.get_expression_mut() {
+            let inner_expr = expr.get_inner_expression_mut();
+            if inner_expr.is_member_expression() {
+                let inner_expr = self.ctx.ast.move_expression(inner_expr);
+                let Ok(member_expr) = MemberExpression::try_from(inner_expr) else {
```
DonIsaac, line 253: Maybe unwrap instead of using `unreachable!`?

DonIsaac (approved):

crates/oxc_transformer/src/typescript/annotations.rs
DonIsaac: Or at least provide an error message in some way in case this panics

overlookmotel: I used `unreachable!()` because it *is* unreachable.

I know that `unreachable!()` can cause the compiler to backtrack further to try to prove that the code is indeed unreachable, and if it can prove that, it removes the unreachable branch entirely. What I don't know is whether it does the same for `.unwrap()`. Quite possibly it does, but I don't know, so I thought better to play it safe. There is some method in my madness!

To be honest, this whole pattern of using `move_expression` and testing the same thing twice is pretty disgusting. I just don't have a good answer for how to do it better without incurring the wrath of the borrow-checker. But it's certainly better than the UB which `ast.copy` brings into play.

PR #5035, opened 2024-08-21
fix(coverage): correctly check semantic data after transform

### LINKED ISSUE #4999: transformer: incorrect semantic check
I saw nearly all tests have a `Symbols mismatch after transform` error. This error is thrown here

https://github.com/oxc-project/oxc/blob/081e2a37d96a765216f9bb0025208a9fdd699667/crates/oxc_semantic/src/post_transform_checker.rs#L129-L132

After looking into it, what happened here, I found a bug in `PostTransformChecker`. From the code above, it can be seen that this is due to the unequal length of semantic symbols in the two different phases. 

The `previous_collect` is collected in `after_semantic` and the `current_collect` is collected in `after_transform`. This is the cause of the problem. After the transformation, the `Semantic` will definitely change. 

So we shouldn't use `after_semantic`'s semantic to compare with `after_transform`'s semantic. Instead, we should rebuild the semantic after the transformation, and compare it to `after_transform`'s semantic

Boshen:
closes #4999

### DIFF
--- a/tasks/transform_conformance/babel.snap.md
+++ b/tasks/transform_conformance/babel.snap.md
@@ -1,8 +1,9 @@
 commit: 12619ffe
 
-Passed: 260/953
+Passed: 406/953
 
 # All Passed:
+* babel-preset-react
 * babel-plugin-transform-react-display-name
 * babel-plugin-transform-react-jsx-source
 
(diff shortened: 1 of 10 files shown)


tasks/transform_conformance/babel.snap.md
```diff
@@ -1275,97 +1276,88 @@ preset-env: unknown field `shippedProposals`, expected `targets` or `bugfixes`
...
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
+Missing ReferenceId
```
Dunqing, line 1324: What does happen here?

Boshen: let me check.

Dunqing (approved):

overlookmotel:
I'm reading through this now. Please give me half an hour to finish review.

overlookmotel (approved):
This makes sense - it's now comparing post-transform scopes/symbols to a fresh semantic run on same AST.

There are further checks we can do to make sure the 2 sets of scopes/symbols in every way. I'll look into that in a follow-up.

PR #5041, opened 2024-08-21
fix(semantic): incorrect semantic check for label has same name

### LINKED ISSUE #5036: semantic: incorrect semantic check for has same name `LabeledStatement` nested in `LabeledStatement`
## Reproduction: 

This bug was found in https://github.com/oxc-project/monitor-oxc/actions/runs/10482911458/job/29034791059

### Minimal reproduction
[oxc playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAICzgICAgICAgICqk0bEZ9wjbyHIrGpkeLan7ycfX8yDD2f99XdYoe9ohorduk5Wq39%2FbTrAgA%3D%3D)

```js
TOP: {
  L: console.log(1);
  L: console.log(2);
};
```

The above code reports the following errors in the playground
```shell
❗ Identifier `L` has already been declared
❗ Identifier `L` has already been declared
```

### Expected result

There are no semantic errors. I've run the code in Chrome devtools and that works well

heygsc:
fix: #5036


Boshen:
What's wrong? Do you need assistance here?

Boshen:
I didn't write this part of the code, feel free to refactor and make bigger changes.

crates/oxc_semantic/tests/integration/cfg_fixtures/label_sibling.js
Boshen, line 1: You may add tests to https://github.com/oxc-project/oxc/tree/main/tasks/coverage/misc instead of here.

crates/oxc_semantic/src/label.rs
```diff
@@ -54,9 +57,18 @@ impl<'a> LabelBuilder<'a> {
...
-
+        let mut parent_id = None;
+        for labels_in in self.labels.iter() {
+            for labels in labels_in.iter().rev() {
+                if labels.depth + 1 == self.depth {
+                    parent_id = Some(labels.id);
+                    break;
+                }
+            }
+        }
```
Boshen, line 68: Do you feel like the algorithm is inefficient overall? Although labels are rare to have any any impact on the overall performance. But we have so much code for handling labels, which is pretty weird 🤔

heygsc: > Do you feel like the algorithm is inefficient overall? Although labels are rare to have any any impact on the overall performance. But we have so much code for handling labels, which is pretty weird 🤔

Yes, I have thought about it before. Perhaps we can only keep the parent before increasing the depth to avoid traversing it every time.
I was thinking the same as you before, the possibility of labels being too complex is unlikely. I was afraid that if the solution of PR was wrong, there was no need to write it. Loop traversal seems easier to write.🤣

heygsc:
The current method is incorrect. It can't pass:
```js
Parentout: {
    Parent: {
        Son: console.log("Son11");
    }
    Parent: {
        Son: console.log("Son21");
    }
}
```

tasks/coverage/misc/pass/oxc-5036.js
```diff
@@ -1,16 +1,41 @@
-Parent1: {
-    Son1: {
-        Son11: console.log("Son11");
-        Son11: console.log("Son11");
+Parent: {
+    Son: {
```
Boshen, line 2: Should be `Child` 😅

heygsc:
@Boshen  I'm a bit confused about CI/Conformance, it seems that the local `just ready` doesn't have this option. Currently, I am running `cargo coverage` locally but it has failed.

heygsc:
I am running coverage locally and it shows the following error
```shell
cargo coverage
   Compiling oxc_coverage v0.0.0 (/Users/hhh/test/oxc-main/tasks/coverage)
    Finished `coverage` profile [optimized] target(s) in 27.85s
     Running `target/coverage/oxc_coverage`
thread '<unnamed>' panicked at tasks/coverage/src/test262/mod.rs:64:48:
called `Option::unwrap()` on a `None` value
error: Recipe `coverage` was terminated on line 85 by signal 6
```

heygsc:
I am running coverage locally and it shows the following error
```shell
cargo coverage
    Finished `coverage` profile [optimized] target(s) in 27.85s
     Running `target/coverage/oxc_coverage`
thread '<unnamed>' panicked at tasks/coverage/src/test262/mod.rs:64:48:
called `Option::unwrap()` on a `None` value
error: Recipe `coverage` was terminated on line 85 by signal 6
```

Boshen:
> I am running coverage locally and it shows the following error
> 
> ```shell
> cargo coverage
>     Finished `coverage` profile [optimized] target(s) in 27.85s
>      Running `target/coverage/oxc_coverage`
> thread '<unnamed>' panicked at tasks/coverage/src/test262/mod.rs:64:48:
> called `Option::unwrap()` on a `None` value
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> error: Recipe `coverage` was terminated on line 85 by signal 6
> ```

`tasks/coverage/test262` may be dirty, try remove and reclone?

heygsc:
> `tasks/coverage/test262` may be dirty, try remove and reclone?

This step was successful with this method, and the next step has the following new problem. Can a similar solution be used to solve it?
Coverage looks difficult to handle🥹

```shell
[react.development.js] - Downloading [https://cdn.jsdelivr.net/npm/react@17.0.2/cjs/react.development.js] to [/oxc-main/target/react.development.js]
thread 'main' panicked at tasks/common/src/test_file.rs:90:67:
called `Result::unwrap()` on an `Err` value: "Transport(Transport { kind: ConnectionFailed, message: Some(\"tls connection init failed\"), url: Some(Url { scheme: \"https\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"cdn.jsdelivr.net\")), port: None, path: \"/npm/react@17.0.2/cjs/react.development.js\", query: None, fragment: None }), source: Some(Os { code: 54, kind: ConnectionReset, message: \"Connection reset by peer\" }) })"
error: Recipe `coverage` was terminated on line 85 by signal 6
```

Boshen:
> > `tasks/coverage/test262` may be dirty, try remove and reclone?
> 
> This step was successful with this method, and the next step has the following new problem. Can a similar solution be used to solve it? Coverage looks difficult to handle🥹
> 
> ```shell
> [react.development.js] - Downloading [https://cdn.jsdelivr.net/npm/react@17.0.2/cjs/react.development.js] to [/oxc-main/target/react.development.js]
> thread 'main' panicked at tasks/common/src/test_file.rs:90:67:
> called `Result::unwrap()` on an `Err` value: "Transport(Transport { kind: ConnectionFailed, message: Some(\"tls connection init failed\"), url: Some(Url { scheme: \"https\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"cdn.jsdelivr.net\")), port: None, path: \"/npm/react@17.0.2/cjs/react.development.js\", query: None, fragment: None }), source: Some(Os { code: 54, kind: ConnectionReset, message: \"Connection reset by peer\" }) })"
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> error: Recipe `coverage` was terminated on line 85 by signal 6
> ```

Are you able to get https://cdn.jsdelivr.net/npm/react@17.0.2/cjs/react.development.js?

crates/oxc_semantic/src/checker/javascript.rs
```diff
@@ -642,7 +642,17 @@ pub fn check_continue_statement<'a>(
 pub fn check_labeled_statement(ctx: &SemanticBuilder) {
     ctx.label_builder.labels.iter().for_each(|labels| {
         let mut defined = FxHashMap::default();
+        //only need to care about the monotone increasing depth of the array
+        let mut rise_depth = vec![];
```
Boshen, line 646: What is `rise_depth`?

crates/oxc_semantic/src/checker/javascript.rs
```diff
@@ -642,7 +642,17 @@ pub fn check_continue_statement<'a>(
...
         for labeled in labels {
+            rise_depth.push(labeled);
+            // have to traverse because HashMap can only delete one by one
+            while rise_depth.len() > 2
+                && rise_depth[rise_depth.len() - 2].depth >= rise_depth[rise_depth.len() - 1].depth
```
Boshen, line 651: We don't usually use array indices, try an iterator instead.

crates/oxc_semantic/src/checker/javascript.rs
heygsc: > What is `rise_depth`?

An array with monotonically increasing depth values, the name `rise` may not be good enough. (Chinese：深度值，单调上升，的数组)

It's significance, consider an example with comment depth:
```js
Parent: {                           // depth:1
    Child1: {                       // depth:2
        Child2: console.log("1");   // depth:3
    }
    Child1: {                       // depth:2
        Child2: console.log("2");   // depth:3
    }
}
```
We will get this array: [1, 2, 3, 2, 3].

Based on my testing on Chrome Devtools and the use of labels, the check for repeated names only needs to consider whether the nodes on the path from the root node to the child nodes are repeated. No need to consider nodes with the same parent and sibling nodes (as mentioned in issue). (Chinese: 只考虑祖先节点，到子节点路径上是否重复即可。相同parent的节点，以及兄弟节点-就是issue中提到的，无需考虑)

So, finding the path from the root node to the child nodes, is `equivalent` to maintaining a `monotonically increasing path in the depth array`.

Our HashMap only needs to store rise_depth.

Returning to the array mentioned earlier: [1, 2, 3, 2, 3].
The process of get `rise_depth` is as follows:
1
1, 2
1, 2, 3
1, 2, 3, 2=>1, 2, 3(delete), 2=>1,2,2=>1,2
1, 2, 3

Boshen: Maybe it's easier to save the errors in the builder 🤔 I'll let you decide whether you want to explore more approaches.

PR #5042, opened 2024-08-21
refactor(semantic): remove unused function params in transformer checker

overlookmotel:
Pure refactor, just tidying code.


overlookmotel:
I think this is uncontentious, and it's purely refactoring, so merging without review.

PR #5050, opened 2024-08-21
refactor(semantic): store all data in `PostTransformChecker` in transform checker

overlookmotel:
Pure refactor of transform checker. Store all scope data in `PostTransformChecker` so it doesn't need to be passed around. `SemanticData` contains a full set of all semantic data for semantic pass, so we have 2 instances of it for 1. after transform and 2. rebuilt semantic.

### DIFF
--- a/crates/oxc_semantic/src/post_transform_checker.rs
+++ b/crates/oxc_semantic/src/post_transform_checker.rs
@@ -16,11 +16,11 @@
 //!    vs from the fresh semantic analysis.
 //!
 //! We now have 2 sets of semantic data:
-//! * "transformer": Semantic data from after the transformer has run
-//! * "rebuild": Semantic data from the fresh semantic analysis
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#5050** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This is last bit of refactoring. Now I'll start implementing more checks.

Boshen (commented):
You know how I never use top level functions as entry points 😄

overlookmotel:
> You know how I never use top level functions as entry points 😄

Yes, but I was never sure if that was a solid preference or just "how it is". Would you prefer this not to be a top level function?

Issue #5008, opened 2024-08-20
Make all scopes unconditional


overlookmotel:
Almost all nodes which can hold a scope *always* hold a scope. But a small handful only create a scope if required:

* `ForStatement`
* `ForInStatement`
* `ForOfStatement`
* `CatchClause`

I propose that we make all scopes unconditional i.e. all these nodes *always* have a scope.

My rationale is:

* Simplicity - especially in transformer where scope-related logic is already complicated.
* In most JS code, these nodes will create a scope anyway (`for (let item of arr)` is more common than `for (item of arr)`, and `catch (err) {` is much more common than `catch {`).
* Resolving references against a scope which has no bindings is very cheap.
* These are not the most common constructs (vs `if` or `function`), so even if it wasn't so cheap, a few extra unnecessary scopes would not hurt perf much anyway.

If I remember right, Dunqing was in favour of this a while back, but I resisted. At that point I didn't realise how few nodes had conditional scopes, so I thought the perf upside of eliding them was much bigger than it is.

Dunqing:
> * Simplicity - especially in transformer where scope-related logic is already complicated.

Yes, I agreed

Dunqing:
We need to remove these conditional in `SemanticBuilder`
https://github.com/oxc-project/oxc/blob/4e1f4abf8979fa2bb81543481904908309ab684c/crates/oxc_semantic/src/builder.rs#L946-L950

overlookmotel:
Glad you agree!

We would also need to remove the `#[scope(if(...))]` attrs from AST (and remove support for that attr from `ast_tools` and from `Traverse` codegen).

overlookmotel:
@Boshen Would you be happy with this change?

Dunqing:
> @Boshen Would you be happy with this change?

I bet @Boshen totally agree this, he likes the standard way to handle scopes 😀

Boshen:
:shipit:

overlookmotel:
> :shipit:

What does a mouse dressed as a detective mean??!

Boshen:
> > :shipit:
> 
> What does a mouse dressed as a detective mean??!

`:shipit:` https://www.quora.com/On-GitHub-what-is-the-significance-of-the-Ship-It-squirrel

Sorry I'm old.

overlookmotel:
Not as old as me! But thanks for explaining.

overlookmotel:
Implemented in #5114 and preceding PRs.

PR #5064, opened 2024-08-22
docs(transformer): add documentation for `optional-catch-binding` plugin

Dunqing:
follow-up #4881

Issue #5017, opened 2024-08-20
Add separate scopes for function bodies


overlookmotel:
### How it is now

Currently functions only have a single scope which contains bindings for:

* Function name (for function expressions)
* Parameters
* Declarations in body of function

In `SemanticBuilder` we manage to correctly resolve `IdentifierReference`s within function params with an extra call to `resolve_references_for_current_scope` after walking the function params:

https://github.com/oxc-project/oxc/blob/b4407c4e9ae4fd35637c74452eb83997f9494875/crates/oxc_semantic/src/builder.rs#L1955-L1963

i.e. We resolve references in params before any bindings from body of function are added to the scope.

### The problem

In the transformer, we also need to resolve references, but we don't have the advantage that we do in `SemanticBuilder` of constructing `ScopeTree` in order. It already exists, and we have to make sense of it.

Here's an example which doesn't work at present:

```jsx
import React from 'react';
function foo(elem = <Component />) {
    let React;
}
```

In classic mode, JSX transform converts this to:

```js
import React from 'react';
function foo(elem = React.createElement(Component, null)) {
    let React;
}
```

We need to resolve the binding for the inserted reference `React`. Currently, it resolves incorrectly to `let React` inside the function body, as resolution goes up the scope tree and finds a binding for `React` in the function's scope.

### Additional problem

In sloppy mode code, there's also an edge case where you can have 2 bindings for `arguments` in a function - #4232.

### Obvious solution

The obvious solution is to add an additional scope to all function bodies.

Advantage: This is a simple solution which has no "surprises".

Disadvantages:

* More scopes = likely perf hit when resolving references for which the binding is outside current function.
* Makes it more complicated to detect illegal redeclarations (`function foo(x) { let x; }`).

### Alternative solution

* Add `SymbolFlags::Parameter` (or maybe existing `ScopeFlags::FunctionScopedVariable` will do).
* Track in `Traverse` if you're inside a function's parameters or not.
* When resolving a reference, if inside a function's parameters, for the first scope encountered with `ScopeFlags::Function` flag, do not match bindings unless they have a `SymbolFlags::Parameter` flag.

Actually, it's a bit more complex than that, in order to handle cases like this:

```js
const x = 1;
function foo(
    y = function bar(z = x) {
        const x = 3;
    }
) {
  const x = 2;
}
```

[...]

PR #5066, opened 2024-08-22
fix(transformer-optional-catch-binding): the `unused` binding is not in the correct scope

Dunqing:
Blocked by #5008

In the SemanticBuilder, we insert all CatchClause parameters in the BlockStatement scope, a child-scope of CatchClause

https://github.com/oxc-project/oxc/blob/858f510d59cb9ea29cca36c31fdf037b0d627f2c/crates/oxc_semantic/src/builder.rs#L709-L718

So we should do the same thing in this plugin, but because CatchClause has no parameters, SemanticBuilder doesn't create scope for `CatchClause`. This cause we cannot find the `BlockStatement` scope_id.

This PR has changed to the correct logic. Just to wait #5008 solved

### DIFF
--- a/crates/oxc_transformer/src/es2019/optional_catch_binding.rs
+++ b/crates/oxc_transformer/src/es2019/optional_catch_binding.rs
@@ -58,8 +58,13 @@ impl<'a> Traverse<'a> for OptionalCatchBinding<'a> {
         if clause.param.is_some() {
             return;
         }
-        let symbol_id =
-            ctx.generate_uid("unused", ctx.scoping.current_scope_id(), SymbolFlags::CatchVariable);
+
+        let block_scope_id = clause.body.scope_id.get().unwrap();
+        let symbol_id = ctx.generate_uid(
+            "unused",
+            block_scope_id,
+            SymbolFlags::CatchVariable | SymbolFlags::FunctionScopedVariable,
+        );
         let name = ctx.ast.atom(ctx.symbols().get_name(symbol_id));
         let binding_identifier =
             BindingIdentifier { span: SPAN, symbol_id: Cell::new(Some(symbol_id)), name };
--- a/tasks/transform_conformance/babel.snap.md
+++ b/tasks/transform_conformance/babel.snap.md
@@ -1,8 +1,9 @@
 commit: 12619ffe
 
-Passed: 338/953
+Passed: 340/953
 
 # All Passed:
+* babel-plugin-transform-optional-catch-binding
 * babel-plugin-transform-react-display-name
 * babel-plugin-transform-react-jsx-source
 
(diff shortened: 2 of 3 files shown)


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5066?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5066** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5065** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5065?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5064** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5064?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
Personally I'm not a fan of the way we move the bindings from catch param up into catch block. I understand it's to make identifying illegal declarations easier, but it's confusing/surprising.

We only have to check for illegal declarations in one place (syntax checker) whereas we have to deal with this non-standard scoping everywhere else. So on balance, in my view the downsides probably outweigh the upsides.

If we add scopes for function bodies (#5017), that'd require a mechanism in syntax checker to know when need to check parent scope for a clashing binding anyway, and could use the same mechanism for `catch` too.

Dunqing:
> Personally I'm not a fan of the way we move the bindings from catch param up into catch block. I understand it's to make identifying illegal declarations easier, but it's confusing/surprising.
> 
> We only have to check for illegal declarations in one place (syntax checker) whereas we have to deal with this non-standard scoping everywhere else. So on balance, in my view the downsides probably outweigh the upsides.
> 

Yes, I agree with your point. I am not a fan of the way as well! But so far I haven't come up with a better way to address this

> If we add scopes for function bodies (#5017), that'd require a mechanism in syntax checker to know when need to check parent scope for a clashing binding anyway, and could use the same mechanism for `catch` too.

Yes, I am looking forward we having a good mechanism to solve this. But we're stuck with the status quo for now 🥲.

overlookmotel:
#5109 unblocks this.

Dunqing:
> #5109 unblocks this.

All tests passed, Thanks!

crates/oxc_transformer/src/es2019/optional_catch_binding.rs
```diff
@@ -58,8 +58,14 @@ impl<'a> Traverse<'a> for OptionalCatchBinding<'a> {
         if clause.param.is_some() {
             return;
         }
-        let symbol_id =
-            ctx.generate_uid("unused", ctx.scoping.current_scope_id(), SymbolFlags::CatchVariable);
+
+        // Get the BlockStatement scope_id below the CatchClause scope
+        let child_scope_id = ctx.scopes().get_child_ids(clause.scope_id.get().unwrap())[0];
```
overlookmotel, line 63: I think this is simpler, and also avoid a lookup in `ctx.scopes()`:

```suggestion
        let child_scope_id = clause.body.scope_id.get().unwrap();
```

Also, I have been wondering if we can remove `child_ids` from `ScopeTree` - it's rarely used, and you can also always get child scopes by visiting from the node (like `ChildScopeCollector` does), so it's not necessary to have it. I suspect the cost of building and maintaining `child_ids` is not worth it.

overlookmotel: Also, I'd suggest renaming `child_scope_id` var here to `body_scope_id` - that's more descriptive.

overlookmotel (commented):
Great to see some of the scope tests being taken off the "fail" list. Everything I've been doing over past 2 days has been just adding more and more to that list. Now the tide is turning!

crates/oxc_transformer/src/es2019/optional_catch_binding.rs
Dunqing: Why didn't I realize I could get the scope_id from inside `clause.body`!

overlookmotel (approved):

Issue #5049, opened 2024-08-21
traverse: port `scope.push` API from Babel


Dunqing:
Babel's implementation is [here](https://github.com/babel/babel/blob/14308d35ce77b4bac30ecd37aec817046402c455/packages/babel-traverse/src/scope/index.ts#L1075-L1144)

### Why need this API?

We need to combine the variable declarators required by different plugins to be inserted at the top of the statements.

### Related tests:

* logical-assignment/null-coalescing/input.js

```diff
-var _obj$x, _obj$x2, _deep$obj, _deep$obj$x, _deep$obj2, _deep$obj2$x, _key, _obj$_key, _key2, _obj$_key2, _deep$obj3, _key3, _deep$obj3$_key, _deep$obj4, _key4, _deep$obj4$_key;
+var _obj$x, _obj$x2, _deep$obj$x, _deep$obj2$x, _obj$_key, _obj$_key2, _deep$obj3$_key, _deep$obj4$_key;
+var _deep$obj, _deep$obj2, _key, _key2, _deep$obj3, _key3, _deep$obj4, _key4;
```

Dunqing:
When we have this API, we can remove the following code. This implementation appears in many plugins
https://github.com/oxc-project/oxc/blob/200d887852e50a252276b1c2009890a987e2be5f/crates/oxc_transformer/src/es2021/logical_assignment_operators.rs#L79-L100

overlookmotel:
Yes, agreed we need this. Aside from the failing test, the repeated `exit_statements` pattern is inefficient.

I'm not sure which is best way to implement it though. Ideally it wouldn't be a method on `TraverseCtx` because it's not really related to traversal. But it's shared state between different transforms, which we're don't have a facility for at present.

By the way, Babel's implementation is inefficient:

```js
// Input
{ obj.a &&= 1; }
{ obj.b &&= 2; }
```

```js
// Output
{
  var _obj;
  (_obj = obj).a && (_obj.a = 1);
}
{
  var _obj2;
  (_obj2 = obj).b && (_obj2.b = 2);
}
```

[...]

PR #5067, opened 2024-08-22
feat(traverse): add `is_static` method to `TraverseCtx`, which is moves out from SymbolTable

Dunqing:
The [scope.isStatic](https://github.com/babel/babel/blob/419644f27c5c59deb19e71aaabd417a3bc5483ca/packages/babel-traverse/src/scope/index.ts#L557) method we port it from Babel. This method is useful for `transformer`. So we should move it out from `SymbolTable`, and it would be better to put it to `TraverseCtx`.

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -372,6 +372,35 @@ impl TraverseScoping {
         let symbol_id = reference.symbol_id();
         self.create_reference_id(ident.span, ident.name.clone(), symbol_id, flags)
     }
+
+    /// Determine whether evaluating the specific input `node` is a consequenceless reference.
+    ///
+    /// I.E evaluating it won't result in potentially arbitrary code from being ran. The following are
+    /// allowed and determined not to cause side effects:
+    ///
+    /// - `this` expressions
+    /// - `super` expressions
+    /// - Bound identifiers
+    ///
+    /// Based on Babel's `scope.isStatic` logic.
+    /// <https://github.com/babel/babel/blob/419644f27c5c59deb19e71aaabd417a3bc5483ca/packages/babel-traverse/src/scope/index.ts#L557>
+    ///
+    /// # Panics
+    /// Can only panic if [`IdentifierReference`] does not have a reference_id, which it always should.
+    #[inline]
+    pub fn is_static(&self, expr: &Expression) -> bool {
+        match expr {
+            Expression::ThisExpression(_) | Expression::Super(_) => true,
+            Expression::Identifier(ident) => self
+                .symbols
+                .get_reference(ident.reference_id.get().unwrap())
+                .symbol_id()
+                .is_some_and(|symbol_id| {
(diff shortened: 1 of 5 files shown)


Dunqing:
* **#5067** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5067?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
As @overlookmotel  said in https://github.com/oxc-project/oxc/issues/5049#issuecomment-2304138762, this method also doesn't related to `Traverse`. I don't know where we should put it is

crates/oxc_traverse/src/context/scoping.rs
```diff
@@ -382,6 +382,30 @@ impl TraverseScoping {
...
+    /// <https://github.com/babel/babel/blob/419644f27c5c59deb19e71aaabd417a3bc5483ca/packages/babel-traverse/src/scope/index.ts#L557>
+    #[inline]
+    pub fn is_static(&self, expr: &Expression) -> bool {
+        match expr {
+            Expression::ThisExpression(_) | Expression::Super(_) => true,
+            Expression::Identifier(ident) => ident.reference_id.get().is_some_and(|reference_id| {
+                self.symbols.get_reference(reference_id).symbol_id().is_some_and(|symbol_id| {
+                    self.symbols.get_resolved_references(symbol_id).all(|r| !r.is_write())
+                })
+            }),
```
overlookmotel, line 405: I think it'd be better to `unwrap` on `ident.reference_id.get()`. In the transformer, semantic has been run, so all `IdentifierReference`s should have a `ReferenceId`. Using `is_some_and` suggests that it's optional, when it's not.

overlookmotel (approved):
I have been thinking that we should create a separate `Scopes` struct which combines `ScopeTree` and `SymbolTable`. The we can move most of `TraverseCtx`'s scope/symbol-related methods to that struct, and they can all live in `oxc_semantic` crate - and be used within transformer and also any other code which has access to `Semantic`.

https://github.com/oxc-project/backlog/issues/92

But we don't have that yet, so I guess `TraverseCtx` is the best place for this method for now.

crates/oxc_traverse/src/context/scoping.rs
Dunqing: Great suggestion

crates/oxc_traverse/src/context/scoping.rs
```diff
@@ -382,6 +382,35 @@ impl TraverseScoping {
...
+    ///
+    /// - `this` expressions
+    /// - `super` expressions
+    /// - Bound identifiers
+    ///
+    /// Based on Babel's `scope.isStatic` logic.
+    /// <https://github.com/babel/babel/blob/419644f27c5c59deb19e71aaabd417a3bc5483ca/packages/babel-traverse/src/scope/index.ts#L557>
+    ///
+    /// ### Panics
+    /// Never panic because [IdentifierReference] must have a reference_id.
```
Dunqing, line 399: Should I add this?

PR #5087, opened 2024-08-22
feat(transformer): always pass in symbols and scopes

### LINKED ISSUE #3565: transformer: `traverse_mut` should error if Semantic finds errors
`traverse_mut` runs semantic to build scope tree.

If semantic finds errors, transforming should not proceed as the transformer makes the assumption that the AST is valid, and may panic or produce invalid output.

But how should `traverse_mut` behave if there are errors?

* Panic?
* Return a `Result`?
* Add errors to transformer's `errors` vec?

Boshen:
We no longer need to build semantic data inside the transformer.

The caller should be responsible for handling semantic data and its
errors.

The best way to achieve this in via `CompilerInterface`.

closes #3565

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1905,6 +1905,7 @@ dependencies = [
  "oxc_diagnostics",
  "oxc_isolated_declarations",
  "oxc_parser",
+ "oxc_semantic",
  "oxc_sourcemap",
  "oxc_span",
  "oxc_transformer",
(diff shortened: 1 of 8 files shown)


Boshen:
* **#5087** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5130, opened 2024-08-23
refactor(ast): remove `#[non_exhaustive]` attr from `AstBuilder`

overlookmotel:
Partially revert #4925. That PR's description gave no explanation of why this attribute is desirable.

### DIFF
--- a/crates/oxc_ast/src/generated/ast_builder.rs
+++ b/crates/oxc_ast/src/generated/ast_builder.rs
@@ -14,7 +14,6 @@ use crate::ast::*;
 
 /// AST builder for creating AST nodes
 #[derive(Clone, Copy)]
-#[non_exhaustive]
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#5130** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5130?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (commented):
This is needed if we want to record node/scope/symbol/reference counts during parsing. We could add a property here that stores those statistics without breaking consuming code. If we have no plans on adding fields to `AstBuilder`, then this is safe to remove.

overlookmotel:
We don't have plans currently to add fields to `AstBuilder` currently, so I think best to remove this for now. I also seem less concerned about breaking changes than you are!

PR #5134, opened 2024-08-23
fix(ci): transform conformance do not skip any errors

overlookmotel:
Snapshots generated by transform conformance were missing a few errors (e.g. `declarations/const-enum/input.ts` at bottom of diff). Make sure all errors fail the test.

PR #5135, opened 2024-08-23
refactor(ci): transform conformance snapshot include when output mismatch

overlookmotel:
Include "Output mismatch" in transform conformance snapshots when a test also fails due to scopes/symbols mismatches.

At present, many of the tests are failing on scopes/symbols mismatches, so it's hard to see from snapshots which also have an output mismatch. In particular, when working to fix scope mismatches, it's hard to notice if changes you make cause the output to be wrong. These extra messages make that visible.

### DIFF
--- a/tasks/transform_conformance/babel.snap.md
+++ b/tasks/transform_conformance/babel.snap.md
@@ -1916,6 +1916,7 @@ preset-env: unknown field `shippedProposals`, expected `targets` or `bugfixes`
 
 
 * logical-assignment/null-coalescing/input.js
+  x Output mismatch
   x Symbol reference IDs mismatch:
   | after transform: SymbolId(17): [ReferenceId(86), ReferenceId(87),
   | ReferenceId(90)]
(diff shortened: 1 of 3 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5135?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5135** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5134** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5134?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Dunqing:
Personally, if the output doesn't match we shouldn't show the semantic errors. In

overlookmotel:
@Dunqing Yes maybe that would be a better way to do it. Check that output matches first, and then only check if semantic matches if the output match test passes. You think that'd work?

Dunqing:
> @Dunqing Yes maybe that would be a better way to do it. Check that output matches first, and then only check that semantic is correct if the output match test passes. You think that'd work?

Yes, I agree we do that. From my experience development of the transformer. When the output is matched, fixing semantic errors is not too hard

Issue #3326, opened 2024-05-17
Incorrect reference flags for `x++` and `x--`


overlookmotel:
```js
let x;
x += 1;
x++;
++x;
x--;
--x;
```

`x += 1` gets correctly flagged as `Read | Write`. But all the rest (`x++` etc) are flagged as just `Write`. This appears to be incorrect.

[playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAICkgICAgICAgIC2mcpqppRb4Fw6K%2FlQ6ebCAzKgaQCmbQvUeyG5MqU9%2Fdu%2Ff3EPr4A%3D)

Dunqing:
This seems to be intentional

https://github.com/oxc-project/oxc/blob/e4b3a3c06a9d30a1fb60b1758bc9176377c48d58/crates/oxc_semantic/src/lib.rs#L273-L277

Same behavior in `no_unused_private_class_members` rule
https://github.com/oxc-project/oxc/blob/e4b3a3c06a9d30a1fb60b1758bc9176377c48d58/crates/oxc_linter/src/rules/eslint/no_unused_private_class_members.rs#L331-L337

But I think `x += 1` should be flagged as `Write` because `x` is never used for reading. I believe this variable can be safely removed by the Minifier/Compressor.

Does this make sense?

overlookmotel:
Hmm. I'm not clear what these flags are meant to represent now!

These are all (broadly) equivalent:

```js
x = x + 1; // Read + Write
x += 1; // Read | Write
x++; // Write
++x; // Write
```

This also seems to be inconsistent - it flags `x` as `Read | Write` where it could be just `Write`:

```js
let x = 1;
let y = x = 2;
```

[playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAICagICAgICAgIC2mcpqppTW%2BJ2DN6Fhf%2FaAzltG8wsRc39%2F7fiAgA%3D%3D)

And something strange going on here - the addition of `console.log` makes `x` `Read | Write`:

```js
console.log;
let x = 1;
x++;
```

[playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAICggICAgICAgICxG0qZRWtpV52IeeJbVn4naXhvLZiVERX4nm00SA6ymugfDPN%2Ff8%2BEgIA%3D)

Dunqing:
> Hmm. I'm not clear what these flags are meant to represent now!
> These are all (broadly) equivalent:
> 
[...]

PR #5158, opened 2024-08-23
fix(semantic): consider update expressions in implicit returns as reads

### LINKED ISSUE #5156: bug(semantic): implicit return in arrow function does not count as a read
Given a variable `i`:

Expected behavior
image

Actual behavior
image

This causes `no-unused-vars` to incorrectly report `i` as unused.

DonIsaac:
Closes #5156

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -82,6 +82,7 @@ pub struct SemanticBuilder<'a> {
     // and when we reach a value declaration we set it
     // to value like
     pub(crate) namespace_stack: Vec<SymbolId>,
+    pub(crate) is_in_return: bool,
     current_reference_flags: ReferenceFlags,
     pub(crate) hoisting_variables: FxHashMap<ScopeId, FxHashMap<Atom<'a>, SymbolId>>,
 
@@ -134,6 +135,7 @@ impl<'a> SemanticBuilder<'a> {
             current_scope_id,
             function_stack: vec![],
             namespace_stack: vec![],
+            is_in_return: false,
             nodes: AstNodes::default(),
             hoisting_variables: FxHashMap::default(),
             scope,
@@ -1229,7 +1231,10 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         /* cfg */
 
         let ret_kind = if let Some(arg) = &stmt.argument {
+            let in_return = self.is_in_return;
+            self.is_in_return = true;
             self.visit_expression(arg);
+            self.is_in_return = in_return;
             ReturnInstructionKind::NotImplicitUndefined
         } else {
             ReturnInstructionKind::ImplicitUndefined
@@ -1624,7 +1629,10 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
             self.visit_ts_type_annotation(return_type);
         }
         if let Some(body) = &func.body {
+            let in_return = self.is_in_return;
+            self.is_in_return = false;
             self.visit_function_body(body);
+            self.is_in_return = in_return;
         }
 
         /* cfg */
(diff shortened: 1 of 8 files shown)


DonIsaac:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5158?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5158** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5155** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5155?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_semantic/src/builder.rs
```diff
@@ -1230,7 +1232,10 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         /* cfg */
 
         let ret_kind = if let Some(arg) = &stmt.argument {
+            let in_return = self.is_in_return;
+            self.is_in_return = true;
             self.visit_expression(arg);
+            self.is_in_return = in_return;
```
Dunqing, line 1238: This may cause a bit of difficulty in maintaining. We should avoid writing any logic in the visit function.

Dunqing (changes requested):

Dunqing:
I tried another way to fix this in #5161. Thank you very much for adding so many tests! I have copied them into #5161. If we approve #5161, you can create a separate PR to update the documentation and fix the false positive for linter

DonIsaac:
@Dunqing I'll split the linter test cases into a separate PR, I think I accidentally committed them... there's a few test cases missing in this PR, which I've left as comments in yours.

crates/oxc_semantic/tests/integration/symbols.rs
```diff
@@ -355,3 +355,51 @@ fn test_ts_interface_heritage() {
...
+        .has_root_symbol("i")
+        .has_number_of_reads(1)
+        .has_number_of_writes(1)
+        .test();
+
+    SemanticTester::js("let i = 0; const x = () => (++i, 0)")
+        .has_root_symbol("i")
+        .has_number_of_reads(1)
+        .has_number_of_writes(1)
+        .test();
```
overlookmotel, line 389: Is this test correct? I don't think `i` is being read from here (but see my general comment - I'm not sure).

overlookmotel (commented):
This is quite hard to give an opinion on. Personally I don't understand why `i++` / `++i` isn't *always* considered `Read | Write` (#3326), so I don't know what the meaning of the `Read` flag is currently.

But, putting that aside for a minute and assuming that your logic is correct... in terms of implementation, is it not possible to use `current_reference_flags` for this, rather than introducing a new `is_in_return` property?

overlookmotel:
> But, putting that aside for a minute and assuming that your logic is correct... in terms of implementation, is it not possible to use `current_reference_flags` for this, rather than introducing a new `is_in_return` property?

Oh, I now see Dunqing has taken exactly this approach in #5161...

crates/oxc_semantic/tests/integration/symbols.rs
Dunqing: TBH, I am not sure what `i` ReferenceFlags should be. The `x` function return result is 0 when it is called, so `i` should be `write` only.  I think we can try to make `i` only contain `write` if no tests are broken

overlookmotel: Let's try to get to the bottom of it! #5165

DonIsaac: IMO it's not, but there are other test cases that enforce this behavior. I mentioned this to you (@Dunqing) a while ago, when you made that change, and your rationale was that "we don't know what's a read and what's not a read, since things could be proxies and/or getters/setters"

PR #5169, opened 2024-08-24
feat(transformer): add remaining options to transformer options

### LINKED ISSUE #5168: transformer: preset-env: unknown field
Need to fix all `preset-env: unknown field` in tasks/transform_conformance/babel.snap.md

Boshen:
closes #5168


Boshen:
There is one remaining option https://github.com/babel/babel/blob/7edd5082152d6b89e90ae09a533eb725b15902ba/packages/babel-plugin-transform-react-jsx/test/fixtures/react-automatic/does-not-add-source-self-automatic/options.json#L6

which should be removed from upstream.

Dunqing (approved):

PR #5170, opened 2024-08-24
feat(isolated_declaration): support sourcemap option

### LINKED ISSUE #4313: Option to return declaration file maps from isolatedDeclaration
https://www.typescriptlang.org/tsconfig/#declarationMap
> Generates a source map for .d.ts files which map back to the original .ts source file. This will allow editors such as VS Code to go to the original .ts file when using features like Go to Definition.
>
>You should strongly consider turning this on if you’re using project references.

For the JavaScript API, `oxc.transform` supports a boolean `sourcemap` option, so it might make sense to do something similar for `oxc.isolatedDeclaration`
```ts
import oxc from 'oxc-transform';

const { sourceText, errors, sourceMap } = oxc.isolatedDeclaration(
    "test.ts",
    "class A {}",
    {sourcemap: true},
);
```

shulaoda:
closes #4313

### DIFF
--- a/napi/transform/index.d.ts
+++ b/napi/transform/index.d.ts
@@ -18,11 +18,16 @@ export interface Es2015BindingOptions {
 }
 
 /** TypeScript Isolated Declarations for Standalone DTS Emit */
-export declare function isolatedDeclaration(filename: string, sourceText: string): IsolatedDeclarationsResult
+export declare function isolatedDeclaration(filename: string, sourceText: string, options: IsolatedDeclarationsOptions): IsolatedDeclarationsResult
+
+export interface IsolatedDeclarationsOptions {
+  sourcemap: boolean
+}
 
 export interface IsolatedDeclarationsResult {
-  sourceText: string
   errors: Array<string>
+  sourceText: string
+  sourceMap?: SourceMap
 }
 
 /**
--- a/napi/transform/src/isolated_declaration.rs
+++ b/napi/transform/src/isolated_declaration.rs
@@ -5,27 +5,43 @@ use oxc_codegen::CodegenReturn;
 use oxc_isolated_declarations::IsolatedDeclarations;
 use oxc_span::SourceType;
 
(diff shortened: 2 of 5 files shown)


shulaoda:
Is there any difference in the naming of `sourcemap` and `source_map`? Which one should we use in the options?

Boshen:
> Is there any difference in the naming of `sourcemap` and `source_map`? Which one should we use in the options?

It should be `sourcemap` everywhere https://rollupjs.org/configuration-options/#output-sourcemap

Dunqing:
> > Is there any difference in the naming of `sourcemap` and `source_map`? Which one should we use in the options?
> 
> It should be `sourcemap` everywhere https://rollupjs.org/configuration-options/#output-sourcemap

I found `oxc_codegen` use `source_map`
https://github.com/oxc-project/oxc/blob/66a1b9d4680f8c35a9d4d4e8d05d661a8dd6c74f/crates/oxc_codegen/src/lib.rs#L186
https://github.com/oxc-project/oxc/blob/66a1b9d4680f8c35a9d4d4e8d05d661a8dd6c74f/crates/oxc_codegen/src/lib.rs#L59

Boshen:
> > > Is there any difference in the naming of `sourcemap` and `source_map`? Which one should we use in the options?
> > 
> > 
> > It should be `sourcemap` everywhere https://rollupjs.org/configuration-options/#output-sourcemap
> 
> I found `oxc_codegen` use `source_map`
> 
> https://github.com/oxc-project/oxc/blob/66a1b9d4680f8c35a9d4d4e8d05d661a8dd6c74f/crates/oxc_codegen/src/lib.rs#L186
> 
> 
> https://github.com/oxc-project/oxc/blob/66a1b9d4680f8c35a9d4d4e8d05d661a8dd6c74f/crates/oxc_codegen/src/lib.rs#L59

I'll unify them.

PR #5192, opened 2024-08-25
perf(codegen): slightly faster `print_list`

Boshen:


crates/oxc_codegen/src/lib.rs
```diff
@@ -412,11 +412,13 @@ impl<'a> Codegen<'a> {
...
     fn print_list<T: Gen>(&mut self, items: &[T], ctx: Context) {
-        for (index, item) in items.iter().enumerate() {
-            if index != 0 {
-                self.print_comma();
-                self.print_soft_space();
-            }
+        let mut iter = items.iter();
+        if let Some(item) = iter.next() {
+            item.gen(self, ctx);
+        }
```
overlookmotel, line 418: No need to fall through to the `while` loop if `iter` is empty.

```suggestion
        let Some(item) = iter.next() else { return; };
        item.gen(self, ctx);
```

DonIsaac (approved):

DonIsaac:
I'm seeing a minor performance degradation in our benchmarks. Could this be causing `rustc` to hit a de-opt path?
image

Boshen:
> I'm seeing a minor performance degradation in our benchmarks. Could this be causing `rustc` to hit a de-opt path? image

This is just noise.

overlookmotel:
I don't think it is just noise. The benchmark time went up to 21 ms on this PR and then sat there steady across the following ~10 commits. [CodSpeed](https://codspeed.io/oxc-project/oxc/benchmarks/tasks/benchmark/benches/codegen.rs::codegen::bench_codegen::codegen%5Bchecker.ts%5D?branchId=66cb07dd8f2e0b608676753e)

From the flamegraph it looks like making this function more complex has caused it not to get inlined any more.

Maybe adding `#[inline]` would fix that.

PR #5196, opened 2024-08-25
perf: use simdutf8 to validate UTF-8 when reading files 

### LINKED ISSUE #5191: linter: use `simdutf8` for validating files
See https://github.com/rolldown/rolldown/pull/2040

Our code to change is:

https://github.com/oxc-project/oxc/blob/666282a13b172de849e251eb3b8417bfb19a79cb/crates/oxc_linter/src/service.rs#L179-L183

shulaoda:
closes #5191

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -170,6 +170,7 @@ seq-macro           = "0.3.5"
 serde               = "1.0.206"
 serde_json          = "1.0.124"
 serde-wasm-bindgen  = "0.6.5"
+simdutf8            = "0.1.4"
 similar             = "2.6.0"
 syn                 = { version = "2.0.74", default-features = false }
 tempfile            = "3.12.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -170,7 +170,7 @@ seq-macro           = "0.3.5"
 serde               = "1.0.206"
 serde_json          = "1.0.124"
 serde-wasm-bindgen  = "0.6.5"
-simdutf8            = "0.1.4"
+simdutf8            = { version = "0.1.4", features = ["aarch64_neon"] }
 similar             = "2.6.0"
 syn                 = { version = "2.0.74", default-features = false }
 tempfile            = "3.12.0"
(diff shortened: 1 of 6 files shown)


Boshen (commented):
No ... we don't need a fs crate yet.

shulaoda:
> No ... we don't need a fs crate yet.

I understand, I will continue with this work tomorrow.

Cargo.toml
```diff
@@ -170,6 +170,7 @@ seq-macro           = "0.3.5"
 serde               = "1.0.206"
 serde_json          = "1.0.124"
 serde-wasm-bindgen  = "0.6.5"
+simdutf8            = "0.1.4"
```
overlookmotel, line 173: Need to enable the `aarch64_neon` feature to get SIMD acceleration on ARM chipsets (e.g. Apple M series).

```suggestion
simdutf8            = { version = "0.1.4", features = ["aarch64_neon"] }
```

overlookmotel (commented):
Aside from the missing `aarch64_neon` feature, looks good.

There's little change on benchmarks, but that's expected - they don't read from disk within the measured bench loop.

@Boshen how do we go about measuring if it makes a real-world difference?

Cargo.toml
shulaoda: > ARM64
The SIMD implementation is only available on Rust Nightly and Rust 1.59 or later. On Rust Nightly it is now turned on automatically. To get the SIMD implementation with Rust 1.59 and 1.60 the crate feature aarch64_neon needs to be enabled. For Rust Nightly this will no longer be required (but does not hurt either). It is expected that the SIMD implementation will be enabled automatically starting with Rust 1.61.

Perhaps I misunderstood the meaning, I thought it would automatically activate.

overlookmotel: Yeah, it's quite confusing. I wasn't sure either, but then I noticed that the crate has not been updated for a long time, so I assume 1.60 was latest Rust version at the time that was written. So I think that feature does need to be manually enabled (and probably won't hurt to enable it even if it doesn't need to be).

Boshen (changes requested):
missing aarch64_neon feature

shulaoda:
> missing aarch64_neon feature

already added

Boshen:
> Aside from the missing `aarch64_neon` feature, looks good.
> 
> There's little change on benchmarks, but that's expected - they don't read from disk within the measured bench loop.
> 
> @Boshen how do we go about measuring if it makes a real-world difference?

I tested this locally, Mac M2 and M3 are both too powerful to show any differences.

Boshen (approved):

PR #5214, opened 2024-08-26
feat(linter)!: parse and display syntax errors for regular expressions

Boshen:
cc @leaysgur shipping to production!

This is marked as breaking change because there may be false positives.


leaysgur:
NOTE: As I commented somewhere (sorry, I forgot), this means that invalid regexp literal will be reported in the parser stage. 

Still some lint rules like `eslint/no-invalid-regex` can report error for regexp constructor usage, but not for literal(= never reports).

Boshen:
It conforms to the spec if the regex parser throws a syntax error for `/regex/` syntax, everything else should go into a linter 😅

leaysgur:
I also think it is natural. Early errors are good. 🙂‍↕️

PR #5221, opened 2024-08-26
perf(codegen): shorten `Codegen::print_list`

overlookmotel:
Revert #5192 and add a comment that it's not a perf gain.

This was really surprising to me, but the benchmarks do demonstrate it.

Please see the benchmarks commit-by-commit on this PR. Adding `#[inline]` to the function does give +1% gain, but it's no better than it was before #5192. So I think preferable to just revert to the simpler original.

I think likely explanation is that the compiler is already performing this optimization itself. And if it does it itself, then it understands the code better, and can then make better decisions about inlining.

https://godbolt.org/z/xzhWWeMoe seems to demonstrate this - there are 2 calls to `Item::gen` in the generated assembly, so it has split the loop into 2.

### DIFF
--- a/crates/oxc_codegen/src/lib.rs
+++ b/crates/oxc_codegen/src/lib.rs
@@ -411,13 +411,25 @@ impl<'a> Codegen<'a> {
         self.needs_semicolon = false;
     }
 
+    // We tried optimizing this to move the `index != 0` check out of the loop:
+    // ```
+    // let mut iter = items.iter();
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#5221** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5232, opened 2024-08-26
refactor(semantic): remove `ScopeTree::child_ids`

### LINKED ISSUE #5244: Remove `ScopeTree::child_ids`
`ScopeTree::child_ids` is rarely used. But it's somewhat costly to build it and keep it updated in transformer. Probably we can just remove it.

If you need to get child scopes, you can start with an AST node and do a shallow traversal, the way `ChildScopeCollector` does:

https://github.com/oxc-project/oxc/blob/088cf52591a699c5917be313402acbff4f4183bf/crates/oxc_traverse/src/generated/scopes_collector.rs

`ChildScopeCollector` is also more versatile, as you can get just the child scopes of a node which doesn't have a scope itself. e.g.:

```rs
function f() {
  function g() {}
  if (foo) (() => {})();
}
```

If you want to find just the child scopes of the `if (foo)` statement (not including `g`'s scope) then `ScopeTree::child_ids` can't give you that, but `ChildScopeCollector` can.

Boshen:
closes https://github.com/oxc-project/oxc/issues/5244

### DIFF
--- a/crates/oxc_semantic/src/post_transform_checker.rs
+++ b/crates/oxc_semantic/src/post_transform_checker.rs
@@ -420,13 +420,6 @@ impl<'s> PostTransformChecker<'s> {
                 self.errors.push_mismatch("Scope parent mismatch", scope_ids, parent_ids);
             }
 
-            // Check children match
-            let child_ids = self
-                .get_pair(scope_ids, |data, scope_id| data.scopes.get_child_ids(scope_id).to_vec());
-            if self.remap_scope_ids_sets(&child_ids).is_mismatch() {
-                self.errors.push_mismatch("Scope children mismatch", scope_ids, child_ids);
-            }
-
             // NB: Skip checking node IDs match - transformer does not set `AstNodeId`s
         }
     }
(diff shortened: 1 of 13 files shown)


crates/oxc_semantic/src/post_transform_checker.rs
```diff
@@ -421,11 +421,11 @@ impl<'s> PostTransformChecker<'s> {
...
-            let child_ids = self
-                .get_pair(scope_ids, |data, scope_id| data.scopes.get_child_ids(scope_id).to_vec());
-            if self.remap_scope_ids_sets(&child_ids).is_mismatch() {
-                self.errors.push_mismatch("Scope children mismatch", scope_ids, child_ids);
-            }
+            // let child_ids = self
+            // .get_pair(scope_ids, |data, scope_id| data.scopes.get_child_ids(scope_id).to_vec());
+            // if self.remap_scope_ids_sets(&child_ids).is_mismatch() {
+            // self.errors.push_mismatch("Scope children mismatch", scope_ids, child_ids);
+            // }
```
Boshen, line 428: @overlookmotel how do you want to change this?

crates/oxc_wasm/src/lib.rs
```diff
@@ -331,13 +331,21 @@ impl Oxc {
...
-                let next_scope_ids = semantic.scopes().get_child_ids(*scope_id);
+                let flags = scope_tree.get_flags(*scope_id);
+                let child_scope_ids = scope_tree
+                    .descendants_from_root()
+                    .filter(|id| {
+                        scope_tree
+                            .get_parent_id(*id)
+                            .is_some_and(|parent_id| parent_id == *scope_id)
+                    })
+                    .collect::<Vec<_>>();
```
Boshen, line 344: Using a linear scan, performance shouldn't be that bad ...

overlookmotel: I think this is O($n^2$).`write_scope_text` is called for every scope, and it iterates over all scopes again in `let child_scope_ids = ...` (though it does nothing for most of them).

If we can assume that scopes are in traversal order (children directly after their parent), I think it's doable in O(n). That assumption would hold true if `ScopeTree` has come straight out of semantic, but it wouldn't if it's been through transformer. Do we want to support displaying the tree post-transform?

But does it matter anyway? If I understand right, this is just for playground where typically people won't paste in huge tracts of code with loads of scopes.

crates/oxc_semantic/src/post_transform_checker.rs
overlookmotel: We can just delete this chunk entirely. No child IDs = no child IDs to check. One less thing to get wrong!

overlookmotel (commented):
Nice bump on semantic + transformer benchmarks.

PR #5234, opened 2024-08-26
refactor(parser): improve example for `byte_search!` macro usage

overlookmotel:
It's more efficient when reading 2 bytes to use `read2()` than 2 x `read()` calls. Reflect that in example for using `byte_search!` macro.


Boshen:
"I'm on public holiday"

PR #5245, opened 2024-08-26
refactor(ci): include drop in semantic benchmark

overlookmotel:
Include dropping `Semantic` in semantic benchmark measure.

If you create a `Semantic`, you have to drop it, so the time it takes to drop is part of the cost of using this API, and we should be working to reduce it. Therefore I think it should be included in the benchmark.

It'll be interesting to see what effect a PR like #5232 which removes a bunch of `Vec`s from `Semantic` has on the drop time.


overlookmotel:
* **#5245** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5255, opened 2024-08-27
fix(transformer/react): mismatch output caused by incorrect transformation ordering

### LINKED ISSUE #4767: traverse: insert new statements(If any) after calling each `walk_statement` in `walk_statements`
### Background 

Here's an example of why we need to support it
```js
// In
function A() {}

const B = hoc(() => {})

// out
function A() {}
var _s1 = A;

const B = hoc(_s2(() => {}))

function C() {}
var _s3 = C;
```

Looking at the example above, we can see that we need to add the `var _sX = functionName` statement after the function statement, and transform arrow function expression to `_sX(() => {})`.

For `FunctionDeclaration`, we can easily do it in `exit_statements`, but for `ArrowFunctionExpression`, we need to transform it in `exit_arrow_function_expression`.  Due to these exit_* functions execution order is different. The output will look like:

```js
// out
function A() {}
var _s2 = A;

const B = hoc(_s1(() => {}))

function C() {}
var _s3 = C;
```

As you can see from the output, `exit_arrow_function_expression` is executed first, then `exit_statements`.  We need to make sure that the transformation order is consistent, so we can’t use exit_statements to implement it.

### Possible Solution

1. `TraverseCtx` provides an API where we can store statements that we want to insert after `exit_statement`.
2. Mutate upwards. Provides an API that is similar to Babel's `node.insertAfter(new)`.

Dunqing:
close: #4767

### DIFF
--- a/crates/oxc_transformer/src/react/refresh.rs
+++ b/crates/oxc_transformer/src/react/refresh.rs
@@ -1,4 +1,4 @@
-use std::cell::Cell;
+use std::{cell::Cell, iter::once};
 
 use oxc_allocator::CloneIn;
 use oxc_ast::{
@@ -8,6 +8,7 @@ use oxc_semantic::{ReferenceFlags, ScopeFlags, ScopeId, SymbolFlags, SymbolId};
 use oxc_span::{Atom, GetSpan, Span, SPAN};
 use oxc_syntax::operator::AssignmentOperator;
 use oxc_traverse::{Ancestor, TraverseCtx};
+use rustc_hash::FxHashMap;
 
 use super::options::ReactRefreshOptions;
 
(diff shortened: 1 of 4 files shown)


Dunqing:
* **#5255** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/react/refresh.rs
```diff
@@ -1034,3 +966,38 @@ fn is_builtin_hook(hook_name: &str) -> bool {
...
+                if let Some(id) = func.id.as_ref() {
+                    symbol_ids.push(id.symbol_id.get().unwrap());
+                }
+            }
+        }
+        _ => {}
+    };
+
+    symbol_ids
+}
```
Dunqing, line 1003: According to https://github.com/oxc-project/oxc/issues/4767#issuecomment-2286162004. We determine where to insert new statements by `symbol_id`. If we have a Statement ID, it will be simpler to determine where we should insert

overlookmotel: Yes, very true.

overlookmotel (approved):
This is a pretty complex transform, and I can't claim to understand it fully. But this has fixed one of the few remaining output mismatch fails for this transform, so 👍.

I'll have a 2nd look at it when I start doing "second pass" through transforms later this week.

PR #5264, opened 2024-08-27
perf(wasm): generate scope text with visitor

overlookmotel:
Follow-on after #5232. `oxc_wasm` build scopes text with a single AST traversal. Previous implementation was O($n^2$).

If we can assume scopes are listed in traversal order, then we could do it a bit more efficiently just from `ScopeTree`, but this approach of using `Visit` will handle out-of-order scope IDs (which you'd get if printing a post-transform `ScopeTree`).

Also reduce creating and discarding `String`s for indentation - reuse a single string instead.

### DIFF
--- a/crates/oxc_wasm/src/lib.rs
+++ b/crates/oxc_wasm/src/lib.rs
@@ -3,17 +3,21 @@
 
 mod options;
 
-use std::{cell::RefCell, path::PathBuf, rc::Rc};
+use std::{
+    cell::{Cell, RefCell},
+    path::PathBuf,
+    rc::Rc,
+};
 
 use options::OxcOptions;
 use oxc::{
     allocator::Allocator,
-    ast::{CommentKind, Trivias},
+    ast::{ast::Program, CommentKind, Trivias, Visit},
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#5264** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5264?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
I am considering merging scope and symbol into one, based on 

https://github.com/oxc-project/oxc/blob/813e10ce507760df23dea67aef88de6791b67da5/crates/oxc_semantic/tests/main.rs#L9-L96

Because I think keeping the consistent with our semantic snapshot is easy to debug the bugs

overlookmotel:
I see your point of it being easier for debugging if the format is consistent. But I think visually it might be a bit too much information all in one place to combine the 2. It'd work if you had expand/contract buttons so you didn't have to see everything at once. Maybe that's achievable with our new fancy Playground?

PR #5256, opened 2024-08-27
feat(ast, parser): add `oxc_regular_expression` types to the parser and AST.

rzvxa:
closes #5060

PR #5278, opened 2024-08-27
feat(ast_tools): add dedicated `Derive` trait.

rzvxa:
In an effort toward the implementation of #5256, this PR allows us to have a separately generated "derive" file for each crate.
This also eliminates a bunch of boilerplate when writing new "derive" generators and generally makes it more approachable.


rzvxa:
* **#5278** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5278?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Just to say sorry I've not got to reviewing this yet. Have been going hell-for-leather on the transformer and semantic. Hope I can get to it this evening.

tasks/ast_tools/src/derives/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn template(module_path: Vec<&str>, impls: TokenStream) -> TokenStream;
+}
+
+#[derive(Debug, Clone)]
+pub struct DeriveOutput(pub Vec<(PathBuf, TokenStream)>);
+
+macro_rules! define_derive {
+    ($vis:vis struct $ident:ident $($lifetime:lifetime)? $($rest:tt)*) => {
+        $vis struct $ident $($lifetime)? $($rest)*
+        impl $($lifetime)? $crate::derives::DeriveTemplate for $ident $($lifetime)? {
```
overlookmotel, line 30: Nit: Please add a line break before this line so it's easier to read.

tasks/ast_tools/src/derives/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+                    #prelude
+
+                    #(#use_modules)*
+
+                    ///@@line_break
+                    #impls
+                }
+            }
+        }
+        impl $($lifetime)? $crate::codegen::Runner for $ident $($lifetime)? {
```
overlookmotel, line 64: Nit: Please add a line break before this line.

tasks/ast_tools/src/derives/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+                                format!("derive_{}.rs", Self::trait_name().to_case(Case::Snake)).as_str()
+                            ),
+                            Self::template(
+                                modules,
+                                streams
+                                .into_iter()
+                                .fold(TokenStream::new(), |mut acc, it| {
+                                    acc.extend(quote::quote!{
+                                        ///@@line_break
+                                    });
```
overlookmotel, line 114: Do these lines mean some `///@@line_breaks` can be removed from each of the `Derive::derive` impls?

tasks/ast_tools/src/generators/ast_builder.rs
```diff
@@ -47,8 +47,8 @@ impl Generator for AstBuilderGenerator {
                     clippy::fn_params_excessive_bools,
                 )]
 
-                ///@@line_break
-                use oxc_allocator::{Allocator, Box, IntoIn, Vec};
+                    ///@@line_break
+                    use oxc_allocator::{Allocator, Box, IntoIn, Vec};
```
overlookmotel, line 51: Looks like this extra indentation isn't required.

```suggestion
                ///@@line_break
                use oxc_allocator::{Allocator, Box, IntoIn, Vec};
```

tasks/ast_tools/src/generators/ast_builder.rs
```diff
@@ -59,14 +59,14 @@ impl Generator for AstBuilderGenerator {
                 #[derive(Clone, Copy)]
                 pub struct AstBuilder<'a> {
                     pub allocator: &'a Allocator,
-                }
+                    }
```
overlookmotel, line 62:
```suggestion
                }
```

overlookmotel (approved):
This is really good. It simplifies writing a new `#[generate_derive]` a lot. I can imagine us writing a bunch more in future - it's a really fast way to build something out which would take days and days to do by hand. The comments I've made are largely nits.

overlookmotel:
One broader thing...

The one thing that could confuse people is visibility. We're used to code generated by `#[derive]` sitting next to the struct/enum it's derived on, and having access to everything you can "see" from that position. Obviously `#[generated_derive]` doesn't work like that.

I don't think much we can do about that, but we should document that you have to make structs/enums/fields `pub(crate)` if you're using `#[generated_derive]`, so people at least understand how to use it.

Or (and very possibly this is a stupid idea), could we make it work like `#[derive]`? e.g.:

* Generate 1 file of generated code for each source file (not one for each crate).
* Introduce a `#[derive_generator]` macro.

Then you use `#[generated_derive]` as follows:

```rs
// src/foo.rs
#[derive_generator]
use oxc_allocator::CloneIn;

#[generated_derive(CloneIn)]
struct Foo { // Not pub
    blah: u64 // Not pub
}
```

`#[generator]` would expand to:

```rs
include!("generated/derive_clonein_foo.rs");
```

Then the generated code has access to everything in the source file without having to make anything `pub` (still wouldn't work for a `struct` which is defined inside a private `module {}` block, but that's a niche case).

This approach has pluses and minuses... maybe it solves one problem but creates another!

overlookmotel:
One other thing. At some point, maybe we should consider breaking `#[generated_derive]` out into a stand-alone thing, separate from `#[ast]`. We may well want to derive traits on types which are not part of the AST.

tasks/ast_tools/src/codegen.rs
```diff
@@ -15,16 +19,54 @@ pub struct AstCodegen {
...
 pub struct AstCodegenResult {
     pub schema: Schema,
-    pub outputs: Vec<(/* generator name */ &'static str, /* output */ GeneratorOutput)>,
+    pub outputs: Vec<SideEffect>,
+}
+
+pub enum SideEffect {
+    /// File write side-effect
+    Write(/* path */ PathBuf, /* output */ Vec<u8>),
+}
```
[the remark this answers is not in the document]
rzvxa: Everything is a file write at the end of the day, I was thinking about the possibility of adding higher-level files to store data from multiple passes/generators. Something like this:

```rust
pub enum SideEffect {
    Write(/* path */ PathBuf, /* output */ Vec<u8>),
    Cache(/* key */ String, /* value */ Vec<u8>),
}
```

Having one file to store multiple outputs and separating them at runtime using their length helps us by lowering the syscalls on the reader part. But I agree that it might be unnecessary to make SideEffect an enum when we only have one variant.
I'll make it a tuple struct for now.

tasks/ast_tools/src/derives/mod.rs
rzvxa: Yes, Weirdly having it without removing those didn't result in double line breaks.

rzvxa:
In regards to your last 2 comments let me think about it. TBH I like the first one more than the latter, However, I see the upsides of them both.

overlookmotel:
Since #5304 is blocked at present, would you like to split this and #5280 into a separate stack, so we can get them merged now?

Concerning my idea of using `include!()` to generate code in situ, I guess the other option is one we've discussed before - the codegen passing a serialized `TokenStream` to the macro to insert. Maybe that's a better solution (when we can get it working) - though I do like that with generated code you have it committed to the repo and it's visible, rather than being hidden away behind a macro expansion.

rzvxa:
> Since https://github.com/oxc-project/oxc/pull/5304 is blocked at present, would you like to split this and https://github.com/oxc-project/oxc/pull/5280 into a separate stack, so we can get them merged now?

Let's not do that, I don't want to mess up the Graphite stack. The down stack should be ready to merge. I'll request a review after running actions.

-----------

What I like about `include` is the fact that we can have our code outside of the src directory(I'm 90% sure, Haven't tested it), This would allow us to put these stuff in a hidden directory so it isn't accessible from our code base in any other way. If we want to hide away the process then having these files in the `generated` directory might be more confusing than helpful. It's not like accessing `oxc_crate_name/.generated_derives/` is hard.

It is worth having an issue tracking this idea in the backlog. We have to do a quick test to make sure it's practical. There might be some quirks with how crates work.

Regarding the latter idea, My personal take is having derives in a separate project is "cleaner" than having them with a conditional feature flag and bloating our core projects with bootstrapping flags. We could use a third person pitching in, Technically speaking it is possible to implement I'm not sure if it would result in a better DX.

overlookmotel:
> Regarding the latter idea, My personal take is having derives in a separate project is "cleaner" than having them with a conditional feature flag and bloating our core projects with bootstrapping flags. We could use a third person pitching in, Technically speaking it is possible to implement I'm not sure if it would result in a better DX.

I suppose the reason I thought of this is I recently added another generated file to `oxc_traverse`: #5119

This generator is pretty simple and lightweight, and it felt like little generators like this would sit quite naturally in the crate in which the generated code sits, so you have the generator and generated code next to each other.

I can imagine writing many more of these (but in Rust, not JS!), and the codebase ending up peppered with them. At that point `oxc_ast_tools` could become a bit of a monolith.

I'm not 100% pushing that we should do this, just explaining where I'm coming from.

PR #5284, opened 2024-08-28
perf(traverse): store `Ancestor` parts as SoA

overlookmotel:
WIP. This is more of an experiment than anything else, to see what effect it has storing the discriminant and payload of an `enum` separately, in order to reduce memory usage. Each `Ancestor` is stored in 10 bytes, instead of 16.

### DIFF
--- a/crates/oxc_traverse/scripts/lib/walk.mjs
+++ b/crates/oxc_traverse/scripts/lib/walk.mjs
@@ -23,14 +23,14 @@ export default function generateWalkFunctionsCode(types) {
             clippy::cast_ptr_alignment
         )]
 
-        use std::{cell::Cell, marker::PhantomData};
+        use std::cell::Cell;
 
         use oxc_allocator::Vec;
         #[allow(clippy::wildcard_imports)]
         use oxc_ast::ast::*;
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#5284** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5284?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
A 1% perf bump. Perhaps not worth it for the amount of unsafe code. But probably would have more effect for a larger `Vec` of enums like `Vec<AstKind>`.

Dunqing:
> A 1% perf bump. Perhaps not worth it for the amount of unsafe code. But probably would have more effect for a larger `Vec` of enums like `Vec<AstKind>`.

Currently, only benchmark `typescript` plugin, if benchmark all the plugins, may have more perf bump

overlookmotel:
If we were going to do this, we'd better off implementing an SoA mechanism which we use everywhere (e.g. `Semantic`'s SoA structure). So closing this and may come back to it.

PR #5298, opened 2024-08-28
fix(transformer): class property with typescript value should not be removed

Boshen:


Boshen:
* **#5298** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing (approved):

tasks/transform_conformance/oxc.snap.md
```diff
@@ -1,13 +1,19 @@
...
 
 
-# babel-plugin-transform-typescript (2/7)
+# babel-plugin-transform-typescript (2/8)
+* class-property-definition/input.ts
+  x Unresolved references mismatch:
+  | after transform: ["const"]
+  | rebuilt        : []
+
+
```
Dunqing, line 16: There seems to be an incorrect check here. The `const` is an unresolved reference, after transformation, the `const` will be removed, Do we need to remove it from `unresolved_references`? @overlookmotel cc

Boshen: I don't know :-)

overlookmotel:
```ts
class Foo {
  #handlers = [] as const;
}
```

`const` in the original TS AST is an `IdentifierReference` which has no binding, so it's recorded in `unresolved_references`. When we remove the `IdentifierReference` in transform, we should remove it from `unresolved_references` too.

PR #5300, opened 2024-08-28
perf(semantic): reduce size of `AstNode` by storing `AstKind` discriminant and pointer separately

overlookmotel:
WIP. Similar to #5284. Another experiment into reducing type sizes by storing an enum's discriminant and payload separately. `AstNode` is reduced from 32 bytes to 24 bytes, by storing `AstKind` in 2 parts.


overlookmotel:
* **#5300** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5300?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
bench

This looks worthwhile to me. But I should clean up the API to make it easier to be safe.

DonIsaac:
This is going to make using a debugger much harder 🫠

overlookmotel:
> This is going to make using a debugger much harder 🫠

Good point. When you say "using a debugger" do you mean using an actual debugger? Or that `Debug` trait prints weird stuff? The latter could be fixed with a custom `Debug` impl, but the former can't.

On balance, I think I'll close this. In my view, we should convert `AstNode` into SoA-style, and could make this optimization at that time. But still would be interested in your answer on the above.

DonIsaac:
I mean a step-through debugger. I use VSCode + LLDB.

It's difficult to know what you're looking at when a raw pointer is used, since the type is used by the debugger to display information correctly. For example, I can't see the string displayed inside of `CompactStr`

overlookmotel:
Interesting. I don't use a debugger. Is there any way to "help" the debugger to see what you want it to see? Or is it baked into debugger that it only understands Rust's built-in types?

DonIsaac:
> Interesting. I don't use a debugger. Is there any way to "help" the debugger to see what you want it to see? Or is it baked into debugger that it only understands Rust's built-in types?

I have no idea actually, I'm not familiar with LLDB 😢

overlookmotel:
Am closing this. If/when we convert `AstNode` into SoA-style, could look at this again.

overlookmotel:
The +6% perf improvement CodSpeed is showing above is I think inaccurate. We had a lot of variance in the cal.com benchmark at the time of this PR. The perf gain was more like +1%. Good, but not spectacular.

Issue #5097, opened 2024-08-22
Add `Deleted` flags for scopes, symbols, references


overlookmotel:
In the transformer, we often delete scopes, symbols and references. But it's not possible to fully delete them from `ScopeTree` and `SymbolTable`, as cannot "shuffle up" entries, as it would invalidate IDs of later scopes/symbols/references.

Instead we can signal that a scope is deleted by setting its flags to `ScopeFlags::Deleted`. Or, probably better, use an absence of any flags (`ScopeFlags::empty()`) to signify a deleted scope. Introduce a `ScopeFlags::Block` flag which will cover the scopes which can currently be `ScopeFlags::empty()`.

Same thing for `SymbolFlags` and `ReferenceFlags`.

Boshen:
I see the SoA structure as database columns, a soft delete is another `is_deleted` column.

All the getter becomes `SELECT * from symbols where is_deleted = true`.

overlookmotel:
I like that analogy. But don't you think better to re-use the existing `flags` columns, as they have spare bits, and often you'd be reading the flags anyway and so will have it in cache?

`SELECT * from symbols where flags = 'Deleted'`

but also when you're e.g. looking for functions:

`SELECT * from symbols where flags = 'FunctionScopedVariable'`

instead of:

`SELECT * from symbols where flags = 'FunctionScopedVariable' AND is_deleted = false`

Boshen:
All reads now need to access `deleted` access, this may get complicated. Let's add this in multiple phases.

overlookmotel:
I don't think it'll be so bad. If you get the `SymbolId` from an AST node (e.g. `binding_ident.symbol_id`) then there's no need to check if the symbol is deleted - you know it's not deleted as you just read it from an active `BindingIdentifier`. You only need to check if the flag says "deleted" if you are looping through *all* symbols from `SymbolTable`. We don't often have reason to do that.

Boshen:
1.5 years without needing this, close as stale.

PR #5301, opened 2024-08-28
fix(semantic): remove `check_unresolved_references`

Dunqing:
As I said in https://github.com/oxc-project/oxc/pull/5298#discussion_r1734689525. The `unresolved_references` check is incorrect. In the meantime, I found that `check_reference` contained the `unresolved_references` check, so I removed it

### DIFF
--- a/crates/oxc_semantic/src/post_transform_checker.rs
+++ b/crates/oxc_semantic/src/post_transform_checker.rs
@@ -155,7 +155,6 @@ pub fn check_semantic_after_transform(
     checker.check_scopes();
     checker.check_symbols();
     checker.check_references();
-    checker.check_unresolved_references();
 
     checker.errors.get()
 }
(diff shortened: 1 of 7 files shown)


Dunqing:
* **#5301** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
I'm afraid I disagree with you on this one.

First off, the test failure discussed in https://github.com/oxc-project/oxc/pull/5298#pullrequestreview-2266286127 is correct, as far as I can see. We have failed to delete an unresolved reference when we should have, and the checker flags that.

Secondly, `check_unresolved_references` is covering a different set of cases from `check_references`. `check_references` starts from a list of all the `IdentifierReference`s which are in the AST and checks they have the correct `SymbolId`. `check_unresolved_references` comes at it from the opposite direction - it loops through `ScopeTree::root_unresolved_references` and checks (a) there aren't any keys in that there shouldn't be and (b) they each have the right `ReferenceId`s recorded against them.

In the case of https://github.com/oxc-project/oxc/pull/5298#pullrequestreview-2266286127, the problem is that we've deleted the `IdentifierReference` from AST, but not deleted the corresponding reference from `root_unresolved_references`. So `check_references` could not catch this, as it starts from the AST, and there is nothing in the AST to start from (as it was deleted).

Dunqing:
Yes, I know we should _remove_ them. but since Semantic doesn't provide `remove` action, this checking is always a mismatch, right? Also, symbol and binding have the same problem. But we haven't checked them what unremoved.

overlookmotel:
Sorry I was rushing out of the house yesterday, so just whacked out #5305, but didn't have time to reply here.

> since Semantic doesn't provide `remove` action

It does now! #5305

> Also, symbol and scope have the same problem. But we haven't checked them what unremoved.

That is true. I left that out of the transform checker for now as currently we have no way to flag scopes, symbols and references as deleted (#5097). That'll be a bit of work, so I thought best just punt the problem for now - and it only matters if you iterate over `ScopeTree` / `SymbolTable` directly, which mostly we don't.

Whereas, for unresolved references, even though we didn't have an API for deleting them, I knew it wouldn't be hard to add one.

Dunqing:
> So, in summary... in my view we should delete unresolved references, #5305 gives us an API to do that, and the transformer checker should check it's done correctly. So I think we should close this PR.

I agree

Issue #5060, opened 2024-08-22
parser: add regex AST node


Boshen:
https://github.com/oxc-project/oxc/blob/afe728a73a34fc81f8e9de57bd8089f1eabc0f9e/crates/oxc_parser/src/js/expression.rs#L348-L351

This `pattern` node cannot be added to the AST due to not having `CloneIn` implemented. cc @rzvxa

rzvxa:
@Boshen Do we want these types to be feature-gated? If so It seems this would take more time than I initially expected. It requires me to go in and refactor stuff related to our AST layout to support conditional layouts.
It would also mean that we have to distinguish between these layouts when doing AST transfer but it isn't too hard to support, It just takes more time as we need to represent conditional fields and type definitions.

Boshen:
> @Boshen Do we want these types to be feature-gated? If so It seems this would take more time than I initially expected. It requires me to go in and refactor stuff related to our AST layout to support conditional layouts. It would also mean that we have to distinguish between these layouts when doing AST transfer but it isn't too hard to support, It just takes more time as we need to represent conditional fields and type definitions.

Welcome back ❤️ 

No feature gate is needed, all the regular expression parsing code is really lightweight.

rzvxa:
> Welcome back ❤️

Thanks Captain!⚓

> No feature gate is needed, all the regular expression parsing code is really lightweight.

That's great to hear, It Simplifies the process and doesn't prevent us from adding it later on as a follow-up PR.
If you would like to feature-gate it in the future, Let me know so I can create an issue for it in the backlog.

rzvxa:
Should we close this issue?

rzvxa:
I'll close it since #5256 addresses this. Feel free to reopen if you think the requirements for this issue aren't met yet.

PR #5304, opened 2024-08-28
feat(regular_expression): implement `Display` for `RegularExpression` type.

rzvxa:
Part of #5060

### DIFF
--- a/crates/oxc_regular_expression/src/display.rs
+++ b/crates/oxc_regular_expression/src/display.rs
@@ -1,0 +1,557 @@
+use std::{
+    fmt::{self, Display},
+    iter::Peekable,
+};
+
+#[allow(clippy::wildcard_imports)]
+use crate::ast::*;
+use crate::surroage_pair::{combine_surrogate_pair, is_lead_surrogate, is_trail_surrogate};
+
+impl<'a> Display for RegularExpression<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "/{}/{}", self.pattern, self.flags)
+    }
+}
+
+impl Display for Flags {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut flags = String::with_capacity(8);
+        macro_rules! if_true_append {
+            ($flag:ident, $char:literal) => {
+                if self.$flag {
+                    flags.push($char);
+                }
+            };
+        }
+
+        // write flags in the order they are described in the `MDN`
+        // <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#advanced_searching_with_flags>
+        if_true_append!(has_indices, 'd');
+        if_true_append!(global, 'g');
+        if_true_append!(ignore_case, 'i');
+        if_true_append!(multiline, 'm');
+        if_true_append!(dot_all, 's');
+        if_true_append!(unicode, 'u');
+        if_true_append!(unicode_sets, 'v');
+        if_true_append!(sticky, 'y');
+
+        write!(f, "{flags}")
+    }
+}
+
+impl<'a> Display for Pattern<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.body)
+    }
+}
+
+impl<'a> Display for Disjunction<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write_join(f, "|", &self.body)
+    }
+}
+
+impl<'a> Display for Alternative<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fn as_character<'a>(term: &'a Term) -> Option<&'a Character> {
+            if let Term::Character(ch) = term {
+                Some(ch)
+            } else {
+                None
+            }
+        }
+        write_join_with(f, "", &self.body, |iter| {
+            let next = iter.next()?;
+            let Some(next) = as_character(next) else { return Some(next.to_string()) };
+            let peek = iter.peek().and_then(|it| as_character(it));
+            let (result, eat) = character_to_string(next, peek);
+            if eat {
+                _ = iter.next();
+            }
+            Some(result)
+        })
+    }
+}
+
+impl<'a> Display for Term<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::BoundaryAssertion(it) => write!(f, "{it}"),
+            Self::LookAroundAssertion(it) => write!(f, "{}", it.as_ref()),
+            Self::Quantifier(it) => write!(f, "{}", it.as_ref()),
+            Self::Character(it) => write!(f, "{it}"),
+            Self::Dot(it) => write!(f, "{it}"),
+            Self::CharacterClassEscape(it) => write!(f, "{it}"),
+            Self::UnicodePropertyEscape(it) => write!(f, "{}", it.as_ref()),
+            Self::CharacterClass(it) => write!(f, "{}", it.as_ref()),
+            Self::CapturingGroup(it) => write!(f, "{}", it.as_ref()),
+            Self::IgnoreGroup(it) => write!(f, "{}", it.as_ref()),
+            Self::IndexedReference(it) => write!(f, "{it}"),
+            Self::NamedReference(it) => write!(f, "{}", it.as_ref()),
+        }
+    }
+}
+
+impl Display for BoundaryAssertion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.kind)
+    }
+}
+
+impl Display for BoundaryAssertionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Start => write!(f, "^"),
+            Self::End => write!(f, "$"),
+            Self::Boundary => write!(f, r"\b"),
+            Self::NegativeBoundary => write!(f, r"\B"),
+        }
+    }
+}
+
+impl<'a> Display for LookAroundAssertion<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "({}{})", self.kind, self.body)
+    }
+}
+
+impl Display for LookAroundAssertionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Lookahead => write!(f, "?="),
+            Self::NegativeLookahead => write!(f, "?!"),
+            Self::Lookbehind => write!(f, "?<="),
+            Self::NegativeLookbehind => write!(f, "?<!"),
+        }
+    }
+}
+
+impl<'a> Display for Quantifier<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.body)?;
+
+        match (self.min, self.max) {
+            (0, None) => write!(f, "*")?,
+            (1, None) => write!(f, "+")?,
+            (0, Some(1)) => write!(f, "?")?,
+            (min, Some(max)) if min == max => write!(f, "{{{min}}}",)?,
+            (min, max) => {
+                let max = max.map_or_else(String::default, |it| it.to_string());
+                write!(f, "{{{min},{max}}}",)?;
+            }
+        }
+
+        if !self.greedy {
+            write!(f, "?")?;
+        }
+
+        Ok(())
+    }
+}
+
+impl Display for Character {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let (string, _) = character_to_string(self, None);
+        write!(f, "{string}")
+    }
+}
+
+impl Display for Dot {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, ".")
+    }
+}
+
+impl Display for CharacterClassEscape {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.kind)
+    }
+}
+
+impl Display for CharacterClassEscapeKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::D => write!(f, r"\d"),
+            Self::NegativeD => write!(f, r"\D"),
+            Self::S => write!(f, r"\s"),
+            Self::NegativeS => write!(f, r"\S"),
+            Self::W => write!(f, r"\w"),
+            Self::NegativeW => write!(f, r"\W"),
+        }
+    }
+}
+
+impl<'a> Display for UnicodePropertyEscape<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.negative {
+            write!(f, r"\P")?;
+        } else {
+            write!(f, r"\p")?;
+        }
+
+        if let Some(value) = &self.value {
+            let name = &self.name;
+            write!(f, "{{{name}={value}}}")
+        } else {
+            write!(f, "{{{}}}", self.name)
+        }
+    }
+}
+
+impl<'a> Display for CharacterClass<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fn as_character<'a>(content: &'a CharacterClassContents) -> Option<&'a Character> {
+            if let CharacterClassContents::Character(ch) = content {
+                Some(ch)
+            } else {
+                None
+            }
+        }
+        write!(f, "[")?;
+
+        if !self.body.is_empty() {
+            if self.negative {
+                write!(f, "^")?;
+            }
+            let sep = match self.kind {
+                CharacterClassContentsKind::Union => "",
+                CharacterClassContentsKind::Subtraction => "--",
+                CharacterClassContentsKind::Intersection => "&&",
+            };
+            write_join_with(f, sep, &self.body, |iter| {
+                let next = iter.next()?;
+                let Some(next) = as_character(next) else { return Some(next.to_string()) };
+                let peek = iter.peek().and_then(|it| as_character(it));
+                let (result, eat) = character_to_string(next, peek);
+                if eat {
+                    _ = iter.next();
+                }
+                Some(result)
+            })?;
+        }
+
+        write!(f, "]")
+    }
+}
+
+impl<'a> Display for CharacterClassContents<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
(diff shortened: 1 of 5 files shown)


rzvxa:
* **#5304** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5304?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,368 @@
...
+        // NOTE: surrogated characters can't be displayed without access to the next character so we
+        // can't print them correctly via `Display` trait. Instead we print their unicode code point
+        (r"/Em🥹j/", Some(r"/Em\uD83E\uDD79j/")),
+        // TODO: currently we do **NOT** respect the `Character::kind` field so everything is
+        // lowered to their final form.
+        (r"/\n\cM\0\x41\./", Some("/\n\r\0A\\./")),
+        (r"/\n\cM\0\x41\u1234\./u", Some("/\n\r\0Aሴ\\./u")),
+        (r"/\n\cM\0\x41\u{1f600}\./u", Some("/\n\r\0A😀\\./u")),
+        // TODO: `\b` is parsed as `\u{8}` which is wrong.
+        // (r"/[\bb]/", Some(r"/[\bb]/")),
```
rzvxa, line 351: @leaysgur  Is this a bug or am I missing something and they represent the same thing?

leaysgur: RegExp `[\bb]` is:

- CharacterClass: `[`
  - `\b`: ClassEscape > b = [U+0008 (BACKSPACE)](https://tc39.es/ecma262/2024/multipage/text-processing.html#sec-patterns-static-semantics-character-value)
  - `b`: SourceCharacter
- `]`

I think you need to check `Character.kind == SingleEscape`.

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+}
+
+impl Display for BoundaryAssertionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Start => write!(f, "^"),
+            Self::End => write!(f, "$"),
+            Self::Boundary => write!(f, "\\b"),
+            Self::NegativeBoundary => write!(f, "\\B"),
```
leaysgur, line 88:
```suggestion
            Self::Boundary => write!(f, r"\b"),
            Self::NegativeBoundary => write!(f, r"\B"),
```

This is easier on the eyes...? 👀

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,368 @@
...
+            (1, None) => write!(f, "+"),
+            (0, Some(1)) => write!(f, "?"),
+            (min, Some(max)) if min == max => {
+                write!(f, "{{{min}}}",)
+            }
+            (min, max) => {
+                let max = max.map_or_else(String::default, |it| it.to_string());
+                write!(f, "{{{min},{max}}}",)
+            }
+        }
```
leaysgur, line 124: `self.greedy` is missing?

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,368 @@
...
+
+impl Display for IndexedReference {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\\{}", self.index)
+    }
+}
+
+impl<'a> Display for NamedReference<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\\k{}", self.name)
```
leaysgur, line 290:
```suggestion
        write!(f, r"\k<{}>", self.name)
```

crates/oxc_regular_expression/src/display.rs
rzvxa: Thank you Sensei!

I wasn't sure about this, I think I have to consider `Character::kind` anyway to prevent our prettier and minifier from exploding the regex size or we have to keep the raw string for a while.

I mark this PR as a draft so I can go in and apply the suggested changes and take care of the `Character` displaying.
Surrogated Unicodes bloat the printer regex by quite a few characters so I should also fix this issue by peeking the next character.

crates/oxc_regular_expression/src/display.rs
rzvxa: Yes sir; I forgot it😅 I'll take care of this along with other changes you've suggested.

crates/oxc_regular_expression/src/display.rs
leaysgur: Yes, checking `Character::kind` is crucial. Otherwise you can not distinguish `\a` and `a`. 😅 

BTW, adding `raw` property to `ast::Character` like other regex parsers may help you in this scenario?

rzvxa: Yes, You are right.
I'll add a few more test cases to better test different character kinds.

> BTW, adding raw property to ast::Character like other regex parsers may help you in this scenario?

It definitely does but I think we can do without it; In that case, it would be better as we can keep the memory footprint small. If we ever want to add a raw string it would be best to keep it for the whole pattern, We can correlate characters to this string using their spans. That way we can print the regex literals without any change.

We might want to keep the raw value for the formatter but the minifier should process regex and output the optimal version of it and wouldn't need the raw value, A custom printer that focuses on output size instead of being close to the actual input should do well for it.

Let me try to achieve this without adding any raw strings, If I fail I'll come back to you so we can figure out if and where we should add this raw value.

rzvxa:
@leaysgur A quick question. Shouldn't `Character::value` be `u16`? Isn't it always utf-16?

rzvxa:
Sorry, I totally missed the Unicode mode.

leaysgur:
Yes, all for Unicode mode.
(Feel free to ask anything!)

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,524 @@
...
+        (r"/\s+/g", None, true),
+        (r"/\t\n\v\f\r/", None, true),
+        // we lose the flags ordering
+        ("/abcd/igv", Some("/abcd/giv"), true),
+        // we lose the flags ordering
+        (r"/\d/ug", Some(r"/\d/gu"), true),
+        // NOTE: we capitalize all hex values.
+        (r"/\n\cM\0\x41\u{1f600}\./u", Some(r"/\n\cM\0\x41\u{1F600}\./u"), true),
+        // TODO: tests below do **NOT** parse with unicode mode on
+        (r"/c]/", None, false),
```
rzvxa, line 473: These cases where the last tuple item is `false` fail to parse with Unicode mode. Is it a known issue? Something I've done wrong or should we create an issue for it?

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,541 @@
...
+        (r"/(.)\1/", None, false),
+        // Identity escape from: <https://github.com/tc39/test262/blob/d62fa93c8f9ce5e687c0bbaa5d2b59670ab2ff60/test/annexB/language/literals/regexp/identity-escape.js>
+        (r"/\C/", None, false),
+        (r"/O\PQ/", None, false),
+        (r"/\8/", None, false),
+        (r"/7\89/", None, false),
+        (r"/\9/", None, false),
+        (r"/8\90/", None, false),
+        // TODO: fails to parse
+        // (r"(.)(.)(.)(.)(.)(.)(.)(.)\8\8", None, false),
```
rzvxa, line 509: This always raises an error even with Unicode mode off.

overlookmotel:
@rzvxa Boshen has added the merge label but I'm unclear if this is ready for merge or not. Not sure if [this comment](https://github.com/oxc-project/oxc/pull/5304#pullrequestreview-2269468725) needs resolution or not.

crates/oxc_regular_expression/src/display.rs
leaysgur: `/c]/` is invalid with unicode mode.

image

crates/oxc_regular_expression/src/display.rs
leaysgur:
```suggestion
        // (r"/(.)(.)(.)(.)(.)(.)(.)(.)\8\8/", None, false),
```

`/` is missing...? 🙈

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    fn test_display(allocator: &Allocator, (source, expect, unicode): Case) {
+        use crate::{Parser, ParserOptions};
+        let expect = expect.unwrap_or(source);
+        let parsed_utf16 =
+            Parser::new(allocator, source, ParserOptions::default()).parse().unwrap();
+        assert_eq!(expect, parsed_utf16.to_string());
+        if unicode {
+            let parsed_unicode =
+                Parser::new(allocator, source, ParserOptions::default().with_unicode_mode())
```
leaysgur, line 529: If you are using `(Literal)Parser`, flag options are useless. They are already handled. 😉 

https://github.com/oxc-project/oxc/blob/afb038e93e85986e5fc229d8db4a5d13154a5184/crates/oxc_regular_expression/src/literal_parser.rs#L27

Boshen:
> [@Boshen](/Boshen) [Boshen](/Boshen) added the [ merge ](/oxc-project/oxc/labels/merge) label [Aug 30, 2024](#event-14073664790) — with  [Graphite App](https://github.com/apps/graphite-app)

@[Boshen](https://github.com/Boshen) Boshen added the [merge](https://github.com/oxc-project/oxc/labels/merge) label [yesterday](https://github.com/oxc-project/oxc/pull/5304#event-14073664790) — with  [Graphite App](https://github.com/apps/graphite-app)

graphite acting on my behalf ... wtf

overlookmotel:
Did you maybe add merge label to one of the PRs above it in the stack? That automatically labels the whole stack below as "merge" too. Or is this Graphite going AWOL again?

crates/oxc_regular_expression/src/display.rs
rzvxa: Oh, I used https://regexr.com/ to test it. It doesn't show this syntax error, It's a bug on their side I guess.
I'll remove that todo comment since it is the intended behavior(not compiling with unicode flag)

crates/oxc_regular_expression/src/display.rs
rzvxa: Please correct me if I'm wrong, I thought it would use the `Parser::options` field, It would only override unicode flags if they are passed in the literal. I want to test the same literal with Unicode and utf-16 mode to ensure we are handling surrogated characters and their u32 code points the same way.

crates/oxc_regular_expression/src/display.rs
leaysgur: image 👀

crates/oxc_regular_expression/src/display.rs
leaysgur: > It would only override unicode flags if they are passed in the literal

Yes, and no.

- `Parser("/abc/u", Options::default())`: true
- `Parser("/abc/u", Options { unicode_mode: false })`: true(!?)
- `Parser("/abc/", Options { unicode_mode: true })`: true

> I want to test the same literal with Unicode and utf-16 mode

For that purpose, I recommend to use `PatternParser`.

Boshen:
@leaysgur feel free to add the `merge` label for graphite to merge.

crates/oxc_regular_expression/src/display.rs
rzvxa: I agree using `PatternParser` is probably better for testing this. Maybe `Parser` shouldn't take these options if we intend to ignore them.
I think this should be either true or panic/raise diagnostic error

> Parser("/abc/u", Options { unicode_mode: false })

leaysgur:
> @leaysgur feel free to add the merge label for graphite to merge.

🫡

@rzvxa Please let me know when you're ready for the review~!

rzvxa:
> > @leaysgur feel free to add the merge label for graphite to merge.
> 
> 🫡
> 
> @rzvxa Please let me know when you're ready for the review~!

It should be ready for review, Let me know if there are any other issues.

rzvxa:
@leaysgur Sorry I marked it as draft again, Found a panic while running conformance tests on #5256

crates/oxc_regular_expression/src/display.rs
rzvxa: That error is always there, doesn't have anything to do with the `c]` pattern.

rzvxa: Ops, I've mistaken it for the Unicode warning.

rzvxa:
@leaysgur OK it is ready, Sorry for the inconvenience.

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,586 @@
+use std::{
+    fmt::{self, Display},
+    iter::Peekable,
+};
+
+#[allow(clippy::wildcard_imports)]
+use crate::ast::*;
+use crate::body_parser::unicode::{combine_surrogate_pair, is_lead_surrogate, is_trail_surrogate};
```
leaysgur, line 8: I'd like to split these 3 into `src/surroage-pair.rs` to normalize dependency graph. (prevent parent > child direction)

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,586 @@
...
+}
+
+fn character_to_string(
+    this: &Character,
+    peek: Option<&Character>,
+) -> (/* result */ String, /* true of peek is consumed */ bool) {
+    const USED_PEEK: bool = true;
+    const UNUSED_PEEK: bool = false;
+
+    if is_trail_surrogate(this.value) {
```
leaysgur, line 360: From my understanding, surrogate pair is only concern for `kind::Symbol`, so how about like this?

```rs
fn character_to_string(
    this: &Character,
    peek: Option<&Character>,
) -> (/* result */ String, /* true of peek should be consumed */ bool) {
    let cp = this.value;

    if let CharacterKind::Symbol = this.kind {
        // Trail only
        if is_trail_surrogate(cp) {
            return (format!(r"\u{cp:X}"), false);
        }

        if is_lead_surrogate(cp) {
            if let Some(peek) = peek.filter(|peek| is_trail_surrogate(peek.value)) {
                // Lead+Trail
                let cp = combine_surrogate_pair(cp, peek.value);
                let ch = char::from_u32(cp).expect("Invalid surrogate pair `Character`!");
                return (format!("{ch}"), true);
            }

            // Lead only
            return (format!(r"\u{cp:X}"), false);
        }
    }

    let ch = char::from_u32(cp).expect("Invalid `Character`!");
    let result = match this.kind {
        CharacterKind::ControlLetter => match ch {
            '\n' => r"\cJ".to_string(),
            '\r' => r"\cM".to_string(),
            '\t' => r"\cI".to_string(),
            _ => format!(r"\c{ch}"),
        },
        CharacterKind::Identifier => {
            format!(r"\{ch}")
        }
        // Not a surrogate, like BMP, or all units in unicode mode
        CharacterKind::Symbol => format!("{ch}"),
        CharacterKind::Null => String::from(r"\0"),
        CharacterKind::UnicodeEscape => {
            // we remove the leading `0x` of our 4 digit hex number.
            let hex = &format!("{cp:#4X}")[2..];
            if hex.len() <= 4 {
                format!(r"\u{hex}")
            } else {
                format!(r"\u{{{hex}}}")
            }
        }
        CharacterKind::HexadecimalEscape => {
            // we remove the leading `0x` of our 2 digit hex number.
            let hex = &format!("{cp:#2X}")[2..];
            format!(r"\x{hex}")
        }
        CharacterKind::Octal => {
            let octal = format!("{cp:o}");
            format!(r"\{octal}")
        }
        CharacterKind::SingleEscape => match ch {
            '\n' => String::from(r"\n"),
            '\r' => String::from(r"\r"),
            '\t' => String::from(r"\t"),
            '\u{b}' => String::from(r"\v"),
            '\u{c}' => String::from(r"\f"),
            '\u{8}' => String::from(r"\b"),
            '\u{2D}' => String::from(r"\-"),
            _ => format!(r"\{ch}"),
        },
    };

    (result, false)
}
```

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,586 @@
...
+
+#[cfg(test)]
+mod test {
+    use oxc_allocator::Allocator;
+    use oxc_span::Span;
+
+    use crate::{literal_parser::parse_reg_exp_literal, FlagsParser, PatternParser};
+    type Case<'a> = (
+        &'a str,
+        /* expected display */ Option<&'a str>,
```
leaysgur, line 430:
```suggestion
        /* expected display, None means expect the same as original */ Option<&'a str>,
```

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,586 @@
...
+
+    (result, peeked)
+}
+
+#[cfg(test)]
+mod test {
+    use oxc_allocator::Allocator;
+    use oxc_span::Span;
+
+    use crate::{literal_parser::parse_reg_exp_literal, FlagsParser, PatternParser};
```
leaysgur, line 427: I also like to fix the following 2 points.

- `parse_reg_exp_literal` is being used like utils
  - It was not originally intended to be exposed to the outside
- The flags to test in unicode mode and the inclusion of the `u|v` flag in the literal seem as a double intent

So how about changing the test case just like this?

```rs
// Before
("/ab/", None, true),
(r"/\c_/", None, false),

// After
("/ab/", None),
("/ab/u", None),
(r"/\c_/", None),
```

then

```rs
    #[test]
    fn test_display() {
        for &(source, expect) in CASES {
            let expect = expect.unwrap_or(source);
            let actual = Parser::new(&allocator, source, ParserOptions::default()).parse().unwrap();
            assert_eq!(expect, actual.to_string());
        }
    }
```

Although somewhat verbose, it is more declarative and eliminates all of the above problems while also cleaning up the logic in the test code.

leaysgur (changes requested):
👀

rzvxa:
@leaysgur Made the requested changes, Please give it another go.

leaysgur (approved):
👍🏻👍🏻

leaysgur:
Sorry, I've accidentally put a merge label on it!
And CI seems to be fail now, what happens in this case...?

crates/oxc_regular_expression/src/display.rs
```diff
@@ -0,0 +1,565 @@
...
+        (r"/\w+/gu", None),
+        (r"/\s/g", None),
+        (r"/\s/gu", None),
+        (r"/\s+/g", None),
+        (r"/\s+/gu", None),
+        (r"/\t\n\v\f\r/", None),
+        (r"/\t\n\v\f\r/u", None),
+        // we lose the flags ordering
+        ("/abcd/igv", Some("/abcd/giv")),
+        ("/abcd/igvu", Some("/abcd/givu")),
```
leaysgur, line 499: `u` and `v` are not allowed to live together.
```suggestion
        ("/abcd/igu", Some("/abcd/giv")),
```

rzvxa: Yes, I changed them in bulk via a macro😅 Fixing it right now.

Boshen:
> Sorry, I've accidentally put a merge label on it! And CI seems to be fail now, what happens in this case...?

I added the merge label on the next PR. https://github.com/oxc-project/oxc/pull/5278

But CI passed?

Boshen:
@leaysgur FYI it won't get merged if CI fails.

PR #5356, opened 2024-08-30
fix(transformer/arrow-functions): reaches `unreachable` when `<this.foo>` is inside an arrow function

### LINKED ISSUE #5353: Remove `JSXMemberExpressionObject::Identifier` variant
Currently we parse all JSX member expressions as an `IdentifierReference` followed by one or more `JSXIdentifier`s.

This is the case regardless of whether the first identifier is upper case or not. i.e. `foo` in `<foo.bar>` is interpreted as an `IdentifierReference`, even though in `<foo>` it'd be a plain `IdentifierName`.

This behavior aligns with Babel: [Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=DwMQ9mB0BGCGBOACA9APgNwChQRgyAjgK4AeKG2AZrnEmlsNVLYaeVkA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.25.6&externalPlugins=&assumptions=%7B%7D)

Therefore we can remove the `JSXMemberExpressionObject::Identifier` variant - it's not possible to create one - it's always an `IdentifierReference`, another nested `JSXMemberExpression`, or `this` (#5352).

It'll be easier to do this once we've closed #5352.

Dunqing:
Fixes: https://github.com/oxc-project/oxc/issues/5353#issuecomment-2321792915

### DIFF
--- a/crates/oxc_transformer/src/es2015/arrow_functions.rs
+++ b/crates/oxc_transformer/src/es2015/arrow_functions.rs
@@ -137,22 +137,31 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
             return;
         }
 
-        let ident = match name {
-            JSXElementName::Identifier(ident) => ident,
-            JSXElementName::MemberExpression(member_expr) => {
-                member_expr.get_object_identifier_mut()
+        if let JSXElementName::Identifier(ident) = name {
+            if ident.name == "this" {
+                let mut new_ident = self.get_this_name(ctx).create_read_reference(ctx);
+                new_ident.span = ident.span;
+                *name = self.ctx.ast.jsx_element_name_from_identifier_reference(new_ident);
             }
-            JSXElementName::IdentifierReference(_) | JSXElementName::NamespacedName(_) => return,
-        };
-        if ident.name == "this" {
-            // We can't produce a proper identifier with a `ReferenceId` because `JSXIdentifier`
-            // lacks that field. https://github.com/oxc-project/oxc/issues/3528
-            // So generate a reference and just use its name.
-            // If JSX transform is enabled, that transform runs before this and will have converted
-            // this to a proper `ThisExpression`, and this visitor won't run.
-            // So only a problem if JSX transform is disabled.
-            let new_ident = self.get_this_name(ctx).create_read_reference(ctx);
-            ident.name = new_ident.name;
+        }
+    }
+
+    /// Change <this.foo></this.foo> to <_this.foo></_this.foo>, and mark it as found
+    fn enter_jsx_member_expression_object(
+        &mut self,
+        node: &mut JSXMemberExpressionObject<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) {
+        if !self.is_inside_arrow_function() {
+            return;
+        }
+
+        if let JSXMemberExpressionObject::IdentifierReference(ident) = node {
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#5357** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5356** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5356?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/es2015/arrow_functions.rs
```diff
@@ -137,22 +137,35 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
...
+        if let JSXElementName::Identifier(ident) = name {
+            if ident.name == "this" {
+                // We can't produce a proper identifier with a `ReferenceId` because `JSXIdentifier`
+                // lacks that field. https://github.com/oxc-project/oxc/issues/5352
+                // So generate a reference and just use its name.
+                // If JSX transform is enabled, that transform runs before this and will have converted
+                // this to a proper `ThisExpression`, and this visitor won't run.
+                // So only a problem if JSX transform is disabled.
+                let new_ident = self.get_this_name(ctx).create_read_reference(ctx);
+                ident.name = new_ident.name;
```
overlookmotel, line 149: We *can* now produce a proper identifier!

```suggestion
                let mut new_ident = self.get_this_name(ctx).create_read_reference(ctx);
                new_ident.span = ident.span;
                *name = JSXElementName::IdentifierReference(ctx.ast.alloc(new_ident));
```

crates/oxc_transformer/src/es2015/arrow_functions.rs
```diff
@@ -137,22 +137,35 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
...
+        ctx: &mut TraverseCtx<'a>,
+    ) {
+        if !self.is_inside_arrow_function() {
+            return;
+        }
+
+        if let JSXMemberExpressionObject::IdentifierReference(ident) = node {
+            if ident.name == "this" {
+                let new_ident = self.get_this_name(ctx).create_read_reference(ctx);
+                ident.name = new_ident.name;
```
overlookmotel, line 167: We should also set `reference_id` on the new ident, and handle `<this.foo.bar>`:

```rs
let mut object = node;
let ident = loop {
    match object {
        JSXMemberExpressionObject::IdentifierReference(ident) => {
            if ident.name == "this" {
                break ident;
            }
            return;
        }
        JSXMemberExpressionObject::MemberExpression(member_expr) => {
            object = &mut member_expr.object;
        }
        JSXMemberExpressionObject::Identifier(_) => unreachable!(),
    }
};
let new_ident = self.get_this_name(ctx).create_read_reference(ctx);
ident.name = new_ident.name;
ident.reference_id = new_ident.reference_id;
```

overlookmotel (approved):

PR #5364, opened 2024-08-31
feat(linter): add `oxc/no-async-endpoint-handlers`

DonIsaac:
Adds `no-async-endpoint-handlers` rules, which bans async functions used as endpoint handlers in Express applications. These do not get caught by Express' error handler, causing the server to crash with an unhandled process rejection error.

```js
app.use(async (req, res) => {
  const foo = await api.getFoo(req.query) // server panics if this function rejects
  return res.json(foo)
})
```

I could not find this rule implemented in any ESLint plugin, but this is a problem I see quite often and I'm tired of dealing with it. I've added it to `oxc` for now, but we should consider adding an `express` or `api` plugin in the future.

### DIFF
--- a/crates/oxc_linter/src/rules/oxc/no_async_endpoint_handlers.rs
+++ b/crates/oxc_linter/src/rules/oxc/no_async_endpoint_handlers.rs
@@ -1,0 +1,376 @@
+use std::ops::Deref;
+
+use oxc_diagnostics::{LabeledSpan, OxcDiagnostic};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{CompactStr, Span};
+use serde_json::Value;
+
+use crate::{context::LintContext, rule::Rule, utils, AstNode};
+use oxc_ast::{
+    ast::{Argument, ArrowFunctionExpression, Expression, Function},
+    AstKind,
+};
+
+#[derive(Debug, Default, Clone)]
+pub struct NoAsyncEndpointHandlers(Box<NoAsyncEndpointHandlersConfig>);
+impl Deref for NoAsyncEndpointHandlers {
+    type Target = NoAsyncEndpointHandlersConfig;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct NoAsyncEndpointHandlersConfig {
+    allowed_names: Vec<CompactStr>,
+}
+
+pub fn no_async_handlers(
+    function_span: Span,
+    registered_span: Option<Span>,
+    name: Option<&str>,
+) -> OxcDiagnostic {
+    #[allow(clippy::cast_possible_truncation)]
+    const ASYNC_LEN: u32 = "async".len() as u32;
+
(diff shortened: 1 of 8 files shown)


crates/oxc_linter/src/utils/express.rs
```diff
@@ -0,0 +1,126 @@
...
+        Expression::ArrowFunctionExpression(arrow) => &arrow.params,
+        _ => return false,
+    };
+
+    // NOTE(@DonIsaac): should we check for destructuring patterns? I don't
+    // really ever see them used in handlers, and their existence could indicate
+    // this function is not a handler.
+    if params.rest.is_some() {
+        return false;
+    }
```
camc314, line 64: i wouldn't bother, i've never seen that used 🙂

crates/oxc_linter/src/rules/oxc/no_async_endpoint_handlers.rs
```diff
@@ -0,0 +1,376 @@
...
+        self.allowed_names.binary_search_by(|allowed| allowed.as_str().cmp(name)).is_ok()
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+    use serde_json::json;
+
+    let pass = vec![
```
camc314, line 302: should we allow if there's a `try` `catch` in the stmt?

camc314 (approved):

crates/oxc_linter/src/utils/express.rs
DonIsaac: Me neither, which is why its existence could indicate that the function is not a handler

crates/oxc_linter/src/rules/oxc/no_async_endpoint_handlers.rs
DonIsaac: This could introduce a lot of difficulties with checking _what_ is being caught
```ts
async handle(req, res) {
  try {
   await couldThrow()
  } catch {}
  await couldAlsoThrow()
}
// or this example, which is never caught
async handle(req, res) {
  try {
    couldThrow.then(data => res.json(data))
  } catch {}
}
```
Making sure everything is covered by a try-catch sounds painful...

crates/oxc_linter/src/utils/express.rs
camc314: yep -probably can remove the note?

crates/oxc_linter/src/rules/oxc/no_async_endpoint_handlers.rs
camc314: hmm i see. this seems reasonable to skip checking

crates/oxc_linter/src/utils/express.rs
DonIsaac: Should we introduce this check?

camc314: i don't think so?

DonIsaac:
* **#5364** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5364?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5366, opened 2024-08-31
fix(transformer): `ArrowfunctionExpression`'s expression is true but has more than one body statement

### LINKED ISSUE #5363: transformer: `??` operator mistransformed in arrow expression
In:

```javascript
(() => a ?? b);
```

out:

```javascript
() => ;
```

expected:

```javascript
() => {
  var _a;
  return (_a = a) !== null && _a !== void 0 ? _a : b;
};
```

Dunqing:
close: #5363

We insert the new statement here, but it's broken if it a `() => x;`, we need to transform it to `function () { return x }`

https://github.com/oxc-project/oxc/blob/8d565d5b23cd22e108881970c5acd10f7c77e80c/crates/oxc_transformer/src/es2020/nullish_coalescing_operator.rs#L59-L76

I don't where we should put the fallback logic, It's useful for all plugins. We had already done the same thing in the react refresh plugin.

https://github.com/oxc-project/oxc/blob/8d565d5b23cd22e108881970c5acd10f7c77e80c/crates/oxc_transformer/src/react/refresh.rs#L578-L605


Dunqing:
* **#5366** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
I have same code in minifier as well ... maybe we need to change the AST to mitigate this behavior.

overlookmotel:
> maybe we need to change the AST to mitigate this behavior.

I agree this would be ideal.

```rs
pub struct ArrowFunctionExpression<'a> {
    body: ArrowFunctionBody<'a>,
    // other fields
    // NB: No `expression` field
}

pub enum ArrowFunctionBody<'a> {
    Expression(Expression<'a>),
    Block(FunctionBody<'a>),
}
```

But... in transformer we don't have ability to mutate up, so it's tricky. Maybe we should centralize logic for inserting `var` statements at top of a statement block (sort of like `ModuleImports` which is shared between all transforms), and then that can insert `var` statements and also handle converting arrow functions where required in an `exit_arrow_function_expression` visitor.

But for now, this PR is probably best way to handle it.

overlookmotel (commented):
Should we add a test case?

Dunqing:
> Should we add a test case?

Forgot to push the test case

Boshen:
Let's clean up this PR and merge first, for getting `monitor-oxc` to pass.

PR #5387, opened 2024-09-01
feat(transformer): support all /regex/ to `new RegExp` transforms

Dunqing:
related: #4754

The implementation port from [esbuild](https://github.com/evanw/esbuild/blob/332727499e62315cff4ecaff9fa8b86336555e46/internal/js_parser/js_parser.go#L12820-L12840). And cover all babel's regexp plugins

---

### TL;DR

Added support for transforming various RegExp features to ensure compatibility with older JavaScript environments.

### What changed?

- Implemented a new `RegExp` transformer to handle unsupported RegExp literal features
- Added options to control different RegExp transformations (e.g., sticky flag, unicode flag, dot-all flag, etc.)
- Updated the transformer to convert unsupported RegExp literals into `new RegExp()` constructor calls
- Added test cases for different RegExp transformations
- Integrated the new RegExp transformer into the existing transformation pipeline

### How to test?

1. Run the existing test suite to ensure no regressions
2. Execute the new RegExp-specific tests in the `tasks/transform_conformance/tests/esbuild-tests/test/fixtures/regexp/` directory
3. Try transforming code with various RegExp features using different target environments to verify correct transformations


Dunqing:
* **#5387** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
I saw a 2% performance regression in enabling plugins.

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+                "RegExp".into(),
+                symbol_id,
+                ReferenceFlags::read(),
+            );
+            ctx.ast.expression_from_identifier_reference(ident)
+        };
+
+        let mut arguments = ctx.ast.vec_with_capacity(2);
+        // Escape backslashes and quotes in the pattern
+        let pattern = regexp.regex.pattern.as_str().replace('\\', "\\\\").replace('\"', "\\\"");
```
Boshen, line 90: This always allocates, better to check for `\` first.

Boshen (commented):
How are these turned on or off from `targets`?

crates/oxc_transformer/src/regexp/options.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Enables plugin for transform the RegExp literal has `\p{}` and `\P{}` unicode property escapes
+    pub unicode_property_escapes: bool,
+    /// Enables plugin for transform `d` flag
+    pub match_indices: bool,
+    /// Enables plugin for transform the RegExp literal has `v` flag
+    pub set_notation: bool,
+}
+
+impl RegExpOptions {
+    pub fn with_sticky_flag(&mut self, value: bool) -> &Self {
```
Boshen, line 24: Do we need these builder builds? given all the fields are public already.

tasks/transform_conformance/tests/esbuild-tests/test/fixtures/regexp/all-regex-plugins-enabled-by-targets/output.js
```diff
@@ -0,0 +1,9 @@
+x1 = new RegExp(".", "y");
+x2 = new RegExp(".", "u");
+a1 = new RegExp("a.b", "s");
+b1 = new RegExp("(?<!x)", "");
+b2 = new RegExp("(?<=x)", "");
+c1 = new RegExp("(?<a>b)", "");
+d1 = new RegExp("\\\\\\\\p{Emoji}", "u");
+f1 = new RegExp("\\\\\\\\p{Emoji}", "u");
+g1 = new RegExp("[\\\\\\\\p{White_Space}&&\\\\\\\\p{ASCII}]", "v");
```
Boshen, line 9: Are we having too many backslashes? esbuild generates `g1 = new RegExp("[\\\\p{White_Space}&&\\\\p{ASCII}]", "v");`

Dunqing:
> How are these turned on or off from `targets`?

 Consistent with other plugins, automatically enable plugins below those versions(from parsed targets). I don't know how to turn off plugins that are enabled by `targets`.  Might want to look into how Babel does this

crates/oxc_transformer/src/regexp/options.rs
Dunqing: I haven't used them, i will remove all.

Boshen (commented):
LGTM, waiting for @overlookmotel to take a look.

Dunqing:
I just found that `RegExpPattern` AST had merged into oxc. So this plugin doesn't need to roughly parse and check `pattern` anymore. That's so great! Thanks for your work! @rzvxa @leaysgur

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+            {
+                pattern
+            }
+            RegExpPattern::Pattern(ref pattern)
+                if is_unsupported || self.has_unsupported_regular_expression_pattern(pattern) =>
+            {
+                regexp.regex.pattern.source_text(self.ctx.source_text)
+            }
+            _ => return,
+        });
```
Dunqing, line 91: Now I need to handle both `RegExpPattern::Raw` and `RegExpPattern::Pattern`, it's not ideal 🥲

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+            None::<TSTypeParameterInstantiation>,
+        );
+    }
+}
+
+impl<'a> RegExp<'a> {
+    /// Check if the regular expression contains any unsupported flags.
+    fn has_unsupported_regular_expression_flags(&mut self, flags: RegExpFlags) -> bool {
+        flags.iter().any(|f| match f {
+            RegExpFlags::G | RegExpFlags::I | RegExpFlags::M => true,
```
overlookmotel, line 135: Shouldn't this be `false`? i.e. `g`, `i` and `m` flags are always supported?

crates/oxc_transformer/src/regexp/mod.rs
overlookmotel: I'm not familiar with how the new regex parser works. How do we end up with `RegExpPattern::Raw`? Is parsing regexes optional?

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+            RegExpPattern::Raw(pattern)
+                if is_unsupported
+                    || self.has_unsupported_regex_syntax_raw(pattern, regexp.regex.flags) =>
+            {
+                pattern
+            }
+            RegExpPattern::Pattern(ref pattern)
+                if is_unsupported || self.has_unsupported_regular_expression_pattern(pattern) =>
+            {
+                regexp.regex.pattern.source_text(self.ctx.source_text)
```
overlookmotel, line 89: This is a bit problematic. It's fine for now, but what if in future we allow arbitrary user-defined transforms, and the user has altered the regex's AST to change the pattern? In that case we'd clobber their changes here. Ideally we need a `Pattern::to_string` method.

Maybe just add a comment about this for now.

overlookmotel:
I've pushed a couple of commits. As usual, feel free to revert them if you don't like them.

Concerning changing `"RegExp".into()` to `Atom::from("RegExp")`: This makes no actual difference. It's just my personal preference to be explicit, as I find it makes it easier to read the code. Where I see `.into()` I often find myself wondering "but into *what*?". `Atom::from(...)` answers removes that ambiguity.

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+                            chars.next();
+                            match chars.peek() {
+                                Some(&'!' | &'=') if self.options.look_behind_assertions => {
+                                    // (?<=) and (?<!)
+                                    return true;
+                                }
+                                _ => {
+                                    if self.options.named_capture_groups {
+                                        return chars.any(|c| c == '>');
+                                    }
```
overlookmotel, line 217: I'm not 100% sure, but is it possible this exits early when it shouldn't in some cases?

```suggestion
                                    if self.options.named_capture_groups && chars.any(|c| c == '>') {
                                        return true;
                                    }
```

Or maybe, since we can assume the RegExp is valid, skip checking for the `>` entirely:

```suggestion
                                    if self.options.named_capture_groups {
                                        return true;
                                    }
```

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+                    }
+                    paren_depth -= 1;
+                }
+                '\\' => {
+                    if self.options.unicode_property_escapes && is_unicode {
+                        if let Some(&next_char) = chars.peek() {
+                            // \p{ and \P{
+                            if (next_char == 'p' || next_char == 'P') && chars.nth(1) == Some('{') {
+                                return chars.any(|c| c == '}');
+                            }
```
overlookmotel, line 236: Same as above. Can this exit early when it shouldn't / can we remove the `chars.any(|c| c == '}')` check?

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+                        && matches!(
+                            assertion.kind,
+                            LookAroundAssertionKind::Lookbehind
+                                | LookAroundAssertionKind::NegativeLookbehind
+                        ) =>
+                {
+                    true
+                }
+                _ => false,
+            })
```
overlookmotel, line 172: Sorry for barrage of comments!

`Term::Quantifier` contains more nested `Term`s. So it could have one of the `Term`s we're looking for. So that needs some form of recursion/traversal.

`Term::CharacterClass` contains `CharacterClass` which contains `CharacterClassContents` which can contain `UnicodePropertyEscape`, so we need to check for that too.

crates/oxc_transformer/src/regexp/mod.rs
overlookmotel: Oh actually, #5304 added `Display` trait impl to `Pattern`. So `regexp.regex.pattern.to_string()` should do it.

crates/oxc_transformer/src/regexp/mod.rs
leaysgur: @rzvxa What the difference between `Raw` and `Invalid`?

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: Yes, it is a mistake was made by me

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: > I'm not familiar with how the new regex parser works. How do we end up with `RegExpPattern::Raw`? Is parsing regexes optional?

Yes, The parser has a `parse_regular_expression` option to enable it.

https://github.com/oxc-project/oxc/blob/abb1e6f2ce3fa4f28422febe76cde674be43c7ab/crates/oxc_parser/src/js/expression.rs#L347-L357

So we have to check both

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: The first one is esbuild way, so I apply it

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: > If we want to avoid that work in common case where the `Pattern` has _not_ been modified, we could add a `pattern_str: Option<Atom<'a>>` field to `RegExp`. If you change the `Pattern`, you'd set `pattern_str` to `None`.

Yes, but we may easily forget to set `pattern_str` to `None`

crates/oxc_transformer/src/regexp/mod.rs
rzvxa: > @rzvxa What was the difference between `Raw` and `Invalid`?

`Raw` is when the regex parser is off, and `Invalid` is when the regex string is invalid and we fail to parse it.

They basically contain the same piece of information just tagged differently. You can use the `RegExpPattern::as_pattern` method to get it as an option when you don't care to differentiate between `Raw` and `Invalid`.

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: No, I made a mistake here

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: I decided to change to `Pattern::to_string`

crates/oxc_transformer/src/regexp/mod.rs
rzvxa: > > I'm not familiar with how the new regex parser works. How do we end up with `RegExpPattern::Raw`? Is parsing regexes optional?
> 
> Yes, The parser has a `parse_regular_expression` option to enable it.
> 
> https://github.com/oxc-project/oxc/blob/abb1e6f2ce3fa4f28422febe76cde674be43c7ab/crates/oxc_parser/src/js/expression.rs#L347-L357
> 
> So we have to check both

We can add a helper method to the `RegExpLiteral` that would parse the pattern. Something like this:

```rust
impl<'a> RegExpLiteral<'a> {
    pub fn try_parse_pattern(
        &self,
        allocator: &'a Allocator,
    ) -> Result<Pattern<'a>, OxcDiagnostic> {
        match &self.regex.pattern {
            RegExpPattern::Raw(raw) | RegExpPattern::Invalid(raw) => {
                use oxc_regular_expression::{ParserOptions, PatternParser};
                let options = ParserOptions {
                    span_offset: self.span.start + 1, // exclude `/`
                    unicode_mode: self.regex.flags.contains(RegExpFlags::U)
                        || self.regex.flags.contains(RegExpFlags::V),
                    unicode_sets_mode: self.regex.flags.contains(RegExpFlags::V),
                };
                PatternParser::new(allocator, raw, options).parse()
            }
            RegExpPattern::Pattern(it) => Ok(it.as_ref().clone_in(allocator)),
        }
    }
}
```

If we add a span field to the `RegExp` type this method can exist on that type instead. At the moment `RegExpLiteral` is the closest place to `RegExpPattern` which contains a concrete span.

rzvxa: I just want to add that we can build upon my previous suggestion and add a lazy parsing that would take mutable reference to the AST node and parse it in place whenever we need to access it.

crates/oxc_transformer/src/regexp/mod.rs
rzvxa: @Dunqing Just beware to_string method isn't a lossless operation and using source text is preferable when the intent is to keep the output same as the input.

furthermore the `RegExpPattern:: source_text` now always uses to_string underneath if span is unset. This change will be added in #5431

rzvxa: As an example one quirk of the to_string is the fact that we lose capitalization information of hex values

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: > We can add a helper method to the `RegExpLiteral` that would parse the pattern. Something like this:
> 
> ```rust
> impl<'a> RegExpLiteral<'a> {
>     pub fn try_parse_pattern(
>         &self,
>         allocator: &'a Allocator,
>     ) -> Result<Pattern<'a>, OxcDiagnostic> { // It should return a `CowLike<Pattern>` to avoid redundant cloning when it is already parsed.
>         match &self.regex.pattern {
>             // Parse even if pattern is known to be invalid so we can reproduce the errors for the caller.
>             RegExpPattern::Raw(raw) | RegExpPattern::Invalid(raw) => {
>                 use oxc_regular_expression::{ParserOptions, PatternParser};
>                 let options = ParserOptions {
>                     span_offset: self.span.start + 1, // exclude `/`
>                     unicode_mode: self.regex.flags.contains(RegExpFlags::U)
>                         || self.regex.flags.contains(RegExpFlags::V),
>                     unicode_sets_mode: self.regex.flags.contains(RegExpFlags::V),
>                 };
>                 PatternParser::new(allocator, raw, options).parse()
>             }
>             RegExpPattern::Pattern(it) => Ok(it.as_ref().clone_in(allocator)),
>         }
>     }
> }
> ```
> 
> If we add a span field to the `RegExp` type this method can exist on that type instead. At the moment `RegExpLiteral` is the closest place to `RegExpPattern` which contains a concrete span.

Actually, Here just need to check whether the pattern contains a kind of syntax. Full parsing of regexp is too costly

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: > @Dunqing Just beware to_string method isn't a lossless operation and using source text is preferable when the intent is to keep the output same as the input.
> 
> furthermore the `RegExpPattern:: source_text` now always uses to_string underneath if span is unset. This change will be added in #5431

Ah, I just realized!

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: > Sorry for barrage of comments!
> 
> `Term::Quantifier` contains more nested `Term`s. So it could have one of the `Term`s we're looking for. So that needs some kind of recursive checking.
> 
> `Term::CharacterClass` contains `CharacterClass` which contains `CharacterClassContents` which can contain `UnicodePropertyEscape`, so we need to check for that too if `self.options.unicode_property_escapes == true`.

I’ve missed these. I need to look through regexp ast

Boshen (changes requested):

crates/oxc_transformer/src/regexp/mod.rs
overlookmotel: I *think* those are only cases we're missing. I "followed the trail" to the types we're looking for through the RegExp AST, and those are the only other paths I could find to reach them.

crates/oxc_transformer/src/regexp/mod.rs
overlookmotel: Is it necessarily a problem if we lose cosmetic information? i.e. isn't it OK to change the regex as long as we maintain its original behavior? This is a transform, which means we change stuff!

rzvxa: Well Ideally for the minifier and transformers, we should use custom printers to optimize for size and compatibility respectively, But I agree it doesn't matter in the case of transformers but other parts of the project still may need to rely on the original source to keep this information.

crates/oxc_transformer/src/regexp/mod.rs
Dunqing: Now we check against the RegExp AST. I think our test should cover it

Dunqing:
@rzvxa Can you take a look at the `transform_conformance` job? There is a panic.

Dunqing:
Well, parsing RegExp increases the performance regression from 6% to 13%。

rzvxa:
> @rzvxa Can you take a look at the `transform_conformance` job? There is a panic in
> 
> https://github.com/oxc-project/oxc/blob/470a6e3b7abe176c66544a93774724d2cf4375bc/crates/oxc_regular_expression/src/display.rs#L378

Yes, I'll take a look. I'm having a hunch as to what it might be.

leaysgur:
@rzvxa Issue is caused by these `match` cases and when `ch` is not used inside like `UnicodeEscape`. (Sorry my suggested code)

https://github.com/oxc-project/oxc/blob/470a6e3b7abe176c66544a93774724d2cf4375bc/crates/oxc_regular_expression/src/display.rs#L379

These patterns were the issue.

```rs
if raw == r"/^|\udf06/ug"
  || raw == r"/\udf06/"
  || raw == r"/\udf06/u"
  || raw == r"/^|\udf06/g"
```

rzvxa:
@leaysgur Yes, that was my theory as well. Thanks for getting to the bottom of this.

@Dunqing I've submitted a fix for it.

Dunqing:
Ah, This PR was closed, and cannot reopen it 🥲

Boshen:
> @[graphite-app](https://github.com/apps/graphite-app) graphite-app bot closed this in https://github.com/oxc-project/oxc/commit/9b984b31bd75824088466b4fe1dd72722c37a2e1 [16 minutes ago](https://github.com/oxc-project/oxc/pull/5387#event-14141223948)

wtf

Boshen:
> @[Boshen](https://github.com/Boshen) Boshen reopened this [now](https://github.com/oxc-project/oxc/pull/5387#event-14141441224)

It can only be reopened from graphite.

Boshen (approved):

overlookmotel:
I didn't mean to add the merge label! I added it to upstack PR and didn't notice it would also merge this. Am reading through the updated code now.

overlookmotel:
Well Graphite merged it anyway! I'll do some follow-up PRs.

PR #5389, opened 2024-09-01
feat(linter/import): add no-dynamic-require rule

jelly:
Rule Detail:

[link](https://github.com/import-js/eslint-plugin-import/blob/v2.29.1/docs/rules/no-dynamic-require.md)


crates/oxc_linter/src/rules/import/no_dynamic_require.rs
```diff
@@ -0,0 +1,127 @@
...
+    esmodule: bool,
+}
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// The require method from CommonJS is used to import modules from different files. Unlike the
+    /// ES6 import syntax, it can be given expressions that will be resolved at runtime. While this
+    /// is sometimes necessary and useful, in most cases it isn't. Using expressions (for instance,
+    /// concatenating a path and variable) as the argument makes it harder for tools to do static
```
DonIsaac, line 25: Please follow the standard docs template.

DonIsaac: This is not resolved. Your example sections do not match the template. Please try again.

DonIsaac: Reminder that we use these comments to generate documentation pages on our website. This is directly user facing and should be high quality.

jelly: Do you mean the missing space between the Bad and Good example?

DonIsaac: https://github.com/oxc-project/oxc/blob/1d3e973f101fc2d03f705f75c7de9e9cb04ae50a/tasks/rulegen/template.txt#L24

jelly: Alright sorry that I missed that, the import rules have no rule generator so I copied one with the old template. Seems there are quite some rules which should be converted.

PR #5399, opened 2024-09-02
feat(transformer/typescript): support `rewrite_import_extensions` option

### LINKED ISSUE #5395: transform: remove ".ts" extension in output
Given that this is typically used for library/bundler transformations, the generated output shouldn't contain extensions in the output.

I suppose the jury can still be out in regards to JS extensions, but TS extensions should definitely not be there.

***Input***

```ts
// src/walk.ts
import { join } from 'node:path';
import * as walk from './walk.ts';
import type { Options } from './walk.ts';

export type { Options };

export function up(name: string, options?: Options): string | undefined {
  // ...
}
```

***Output***

```ts
// output/find.mjs
import { join } from "node:path";
import * as walk from "./walk.ts";
export function up(name, options) {
  // ...
}

// output/find.d.mts
import type { Options } from "./walk.ts";
export type { Options };
export declare function up(name: string, options?: Options): string | undefined;
```

> Note that **both** the generate `dts` and JS still contains the `walk.ts` reference.

***Desired***

```ts
// output/find.mjs
import { join } from "node:path";
import * as walk from "./walk";
export function up(name, options) {
  // ...
}

// output/find.d.mts
import type { Options } from "./walk";
export type { Options };
export declare function up(name: string, options?: Options): string | undefined;
```

Because this particular case is a library, `./walk` will be re-resolved to the correct `"exports"` condition when needed. 

And arguably, any bundler (or even runtime) contexts that ingest an extension-less import will still be able to resolve the local file sibling.

Dunqing:
close: #5395

Babel only supports `rewrite`, we also support `remove`

### DIFF
--- a/crates/oxc_transformer/src/typescript/options.rs
+++ b/crates/oxc_transformer/src/typescript/options.rs
@@ -1,6 +1,9 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, fmt};
 
-use serde::Deserialize;
+use serde::{
+    de::{self, Visitor},
+    Deserialize, Deserializer,
+};
 
 use crate::context::TransformCtx;
 
@@ -45,6 +48,16 @@ pub struct TypeScriptOptions {
 
     /// Unused.
     pub optimize_const_enums: bool,
+
+    // Preset options
+    /// Modifies extensions in import and export declarations.
+    ///
+    /// This option, when used together with TypeScript's [`allowImportingTsExtension`](https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions) option,
+    /// allows to write complete relative specifiers in import declarations while using the same extension used by the source files.
+    ///
+    /// When set to `true`, same as [`RewriteExtensionsMode::Rewrite`]. Defaults to `false` (do nothing).
+    #[serde(deserialize_with = "deserialize_rewrite_import_extensions")]
+    pub rewrite_import_extensions: Option<RewriteExtensionsMode>,
 }
 
 impl TypeScriptOptions {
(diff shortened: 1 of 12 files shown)


Dunqing:
* **#5399** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/typescript/options.rs
```diff
@@ -45,6 +48,11 @@ pub struct TypeScriptOptions {
 
     /// Unused.
     pub optimize_const_enums: bool,
+
+    // Preset options
+    /// Rewrite the import extensions to js extensions
```
Boshen, line 53: Can you copy over some of the documentation from https://babel.dev/docs/babel-preset-typescript#rewriteimportextensions?

Boshen (commented):
Can you also update napi?

crates/oxc_transformer/src/typescript/rewrite_extensions.rs
```diff
@@ -0,0 +1,78 @@
+use std::marker::PhantomData;
+
+use oxc_ast::ast::{
+    ExportAllDeclaration, ExportNamedDeclaration, ImportDeclaration, StringLiteral,
+};
+use oxc_traverse::{Traverse, TraverseCtx};
+
+use super::options::RewriteExtensionsMode;
+
+pub struct TypeScriptRewriteExtensions<'a> {
```
Boshen, line 10: Can you reference the plugin code from babel?

crates/oxc_transformer/src/typescript/rewrite_extensions.rs
```diff
@@ -0,0 +1,78 @@
...
+impl<'a> TypeScriptRewriteExtensions<'a> {
+    pub fn new(mode: RewriteExtensionsMode) -> Self {
+        Self { mode, _marker: PhantomData }
+    }
+
+    #[allow(clippy::case_sensitive_file_extension_comparisons)]
+    pub fn rewrite_extensions(&self, source: &mut StringLiteral<'a>, ctx: &mut TraverseCtx<'a>) {
+        let value = source.value.as_str();
+        if value.contains('/') || value.contains('\\') {
+            source.value = if value.ends_with(".ts") {
```
Boshen, line 24: I think you can match on `value.get(value.len() - 3..value.len())` 🤣

crates/oxc_transformer/src/typescript/options.rs
Dunqing: Updated, made some changes

crates/oxc_transformer/src/typescript/rewrite_extensions.rs
Dunqing: Ahh, Yes! But should be `value.get(..value.len() - 3)` ?

Boshen (approved):
LGTM, feel free to merge when you're ready.

Dunqing:
@Boshen Do you mind adding a napi feature gate in the transformer? So that we can directly use such `RewriteExtensionsMode` and `ReactRuntime` options in `napi/transform`

Boshen:
> @Boshen Do you mind adding a napi feature gate in the transformer? So that we can directly use such `RewriteExtensionsMode` and `ReactRuntime` options in `napi/transform`

I'll rework all the options for wasm and napi later in the future.

crates/oxc_transformer/src/typescript/rewrite_extensions.rs
```diff
@@ -0,0 +1,97 @@
...
```
Boshen, line 31: I'll do a quick clean up.

Boshen:
@Dunqing I made some simplifications, PTAL.

Dunqing:
Looks like allocating is unavoidable in this way, But now it looks more intuitive

PR #5402, opened 2024-09-02
perf(linter/react): add should_run conditions for react rules

jelly:
Only run the React rules on files containing JSX elements.


camc314 (commented):
hmm what about `createElement` calls?

these could be called inside `ts`/`js` files

jelly:
Updated the PR to only include `should_run` for when there is no React.createElement or Rules of hooks as they work for classes as well

DonIsaac (approved):

PR #5407, opened 2024-09-02
fix(traverse): invalid variable name generated by `generate_uid_based_on_node`

### LINKED ISSUE #5371: transformer: incorrect temporary variable name
in:

```
output.headers["x-amzn-requestid"] ?? output.headers["x-amzn-request-id"]
```

out:

```
var _output$headers$x-amzn-requestid;
(_output$headers$x-amzn-requestid = output.headers["x-amzn-requestid"]) !== null && _output$headers$x-amzn-requestid !== void 0 ? _output$headers$x-amzn-requestid : output.headers["x-amzn-request-id"];
```

expected

```
var _output$headers$xAmz;
(_output$headers$xAmz = output.headers["x-amzn-requestid"]) !== null && _output$headers$xAmz !== void 0 ? _output$headers$xAmz : output.headers["x-amzn-request-id"];
```

https://github.com/oxc-project/monitor-oxc/actions/runs/10646542195/job/29513518443

Note there's also other cases such as `options['grpc.enable_http_proxy'] ?? 1`

Dunqing:
close: #5371

### DIFF
--- a/crates/oxc_traverse/src/context/identifier.rs
+++ b/crates/oxc_traverse/src/context/identifier.rs
@@ -1,0 +1,58 @@
+use std::borrow::Cow;
+
+use oxc_syntax::identifier::{is_identifier_name, is_identifier_part, is_identifier_start};
+
+/// Convert a str to a valid identifier name.
+///
+/// Based on Babel's [`toIdentifier`](https://github.com/babel/babel/blob/3bcfee232506a4cebe410f02042fb0f0adeeb0b1/packages/babel-types/src/converters/toIdentifier.ts#L4-L26) function.
+pub fn to_identifier(input: &str) -> Cow<str> {
+    if is_identifier_name(input) {
+        return Cow::Borrowed(input);
+    }
+
+    let mut name = String::with_capacity(input.len());
+
+    let mut capitalize_next = false;
+
+    let mut chars = input.chars();
+    if let Some(first) = chars.next() {
+        if is_identifier_start(first) {
+            name.push(first);
+        } else {
+            capitalize_next = true;
+        }
+    }
+
+    for c in chars {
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#5407** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
Another viable approach with less impact on performance is to handle it in the `GatherNodeParts` trait, but it needs to do more work, I will try it later.

crates/oxc_traverse/src/context/identifier.rs
Boshen, line 1: It feels like this should be moved to `oxc_syntax`, add a `oxc_syntax/src/string.rs`?

Dunqing: I think no other places will use this because it is a transformer-only logic that ports from `Babel`.

overlookmotel (approved):
I pushed a commit to make the test case a bit clearer. At first I was like "this transform isn't correct" because I didn't notice the difference between `x-amzn-requestid` and `x-amzn-request-id`. I think making the two more obviously different is an improvement, but feel free to revert that commit if you disagree.

We could make this more efficient by sanitizing strings in `GatherNodeParts`. Most common case is probably where all the parts are `identifierReference`s and `IdentifierName`s, which we know contain only valid characters, so then we could skip sanitization on them. But we can do that later on.

NB I know this is a direct port of Babel's algorithm. It's just that Babel isn't always as optimized as it could be!

https://github.com/oxc-project/backlog/issues/105

@Dunqing Please go ahead and merge. I'm only not merging myself to give you chance to revert my extra commit if you don't like it.

PR #5420, opened 2024-09-03
feat(linter/react): Implement `no-danger-with-children` rule

camchenry:
- Part of https://github.com/oxc-project/oxc/issues/1022

This implements the [recommended rule `no-danger-with-children`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-danger-with-children.md) from the `eslint-plugin-react` package.

This rule proved to be a little bit trickier to implement than I anticipated, as it required searching up through all of the scopes recursively in order to resolve object properties. This would be easier with something like a type-check system, but this will work for now.

(Sidenote: this is my first real attempt at Rust programming in ~5 years, so any feedback on making this more idiomatic is welcome.)


crates/oxc_linter/src/rules/react/no_danger_with_children.rs
```diff
@@ -0,0 +1,357 @@
...
+    node: &AstNode,
+    name: &str,
+) -> Option<&'c AstNode<'c>> {
+    ctx.scopes()
+        .find_binding(node.scope_id(), name)
+        .map(|symbol_id| ctx.semantic().symbol_declaration(symbol_id))
+}
+
+/// Returns whether a given object has a property with the given name.
+fn is_object_with_prop_name(
```
camchenry, line 344: This function shares a decent chunk of code with part of `does_object_var_have_prop_name` but I was struggling to figure out how to refactor it cleanly. Thoughts appreciated 🙏

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
```diff
@@ -0,0 +1,351 @@
...
+
+                // Only accept calls like `.createElement(...)`
+                if callee.property.name != "createElement" {
+                    return;
+                }
+
+                let Some(props) = call_expr
+                    .arguments
+                    .get(1)
+                    .expect("Could not get props argument for createElement call")
```
DonIsaac, line 83: Don't panic here. This could reasonably happen in real code; it's best to just return.

Consider using this instead:
```rust
call_expr
  .arguments
  .get(1)
  .and_then(Argument::as_expression)
```

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
DonIsaac: You could attept to resolve the variable, then call this function if you find an object? I.e. starting on like 302.

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
camchenry: Ah good call 👍 Added in https://github.com/oxc-project/oxc/pull/5420/commits/9cd32f688a245a9f89a1e3d7234794c90aba23c0

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
camchenry: I might need an example here, I'm just having trouble seeing it 😅

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
```diff
@@ -339,13 +337,13 @@ fn is_object_with_prop_name(
...
-                return false;
-            };
-            key == prop_name
-        }
-        ObjectPropertyKind::SpreadProperty(_) => false,
+    obj_props.iter().any(|prop| {
+        let ObjectPropertyKind::ObjectProperty(obj_prop) = prop else {
+            return false;
+        };
+        let Some(key) = obj_prop.key.static_name() else {
```
DonIsaac, line 344:
```suggestion
        obj_prop.key.static_name().is_some_and(|key| key == prop_name)
```

crates/oxc_linter/src/rules/react/no_danger_with_children.rs
DonIsaac: I'll provide one. My comment is non-blocking, this level of duplication is not problematic.

DonIsaac (approved):

DonIsaac:
I keep updating this and forgetting to come back and merge it... my bad haha

PR #5425, opened 2024-09-03
perf(syntax): optimize `is_identifier_name`

overlookmotel:
Optimize `oxc_syntax::identifier::is_identifier_name`. Add a fast path for ASCII, which will be the common case. Only fall back to iterating over `char`s and using the more expensive test functions e.g. `is_identifier_start_unicode` if non-ASCII chars are found.


overlookmotel:
* **#5425** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5425?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
But it didn't win us back the -3% on transformer benchmark, which is what I was hoping for!

Issue #5283, opened 2024-07-16
ast: Derive `Hash` and `PartialEq` on AST node types


overlookmotel:
Continuation from https://github.com/oxc-project/oxc/issues/3819

We should derive `Hash` and `PartialEq` on all AST node types.

## Why?

The reasons we need to codegen them are (copied from https://github.com/oxc-project/oxc/issues/3819#issuecomment-2215593269):

### `Hash`

We want `Hash` to skip `span` fields, and also `symbol_id` etc. Skipping `Hash` is currently achieved by a dummy `Hash` impl on `Span` type:

https://github.com/oxc-project/oxc/blob/1b91d40da26a59c353873e93483aac58c4ac9f39/crates/oxc_span/src/span.rs#L309-L313

I saw this as a bit of a hack, and thought it would be more "proper" solution to implement `Hash` on all AST types with the `span` field explicitly skipped. That would be laborious and error-prone to do by hand, but much easier with codegen.

### `PartialEq`

The reason why we are implementing `Hash` on AST types is for (if I remember right) comparing expressions in a lint rule to identify pointless `if` branches e.g. `if (x) {} else if (x) { /* unreachable */ }`. But actually `Hash` is not ideal for this use - `PartialEq` would be more efficient.

But it's hard to implement `PartialEq` because, again, we need to skip `span` fields, and the no-op hash hack isn't viable here, so we can't use `#[derive(PartialEq)]`.

So I thought we could do that with codegen.

## Use different traits?

Instead of implementing `Hash` and `PartialEq`, should we be more explicit about what these trait implementations do, and call them `HashContent` and `PartialEqContent`?

`PartialEq` could then also be implemented on AST nodes for when you want to `==` check if 2 AST nodes are the same *including* their spans. And when you compare 2 nodes without spans, you have to do `node1.eq_content(&node2)` so it's explicit that spans *aren't* included in the comparison.

I'm not sure if this is a good idea or not.

## Other useful info

rzvxa provided some useful background on how Rust's internal derive macros work: https://github.com/oxc-project/oxc/issues/3819#issuecomment-2215497901

Boshen:
Once this is done, we can replace this hash comparison of expressions to use `PartialEq` instead.

https://github.com/oxc-project/oxc/blob/23e84564e7ac04b476c45da90f4a713ce24e799e/crates/oxc_linter/src/ast_util.rs#L10-L14

overlookmotel:
[...]

PR #5327, opened 2024-08-29
refactor(ast): remove `IdentifierName` from `TSThisParameter`

overlookmotel:
`TSThisParameter` does not need to include an `IdentifierName` which is always "this". Just storing the `Span` is sufficient.

PR #5427, opened 2024-09-03
feat(ast): add `ContentEq` trait.

rzvxa:
Part of #5283

### DIFF
--- a/crates/oxc_span/src/cmp.rs
+++ b/crates/oxc_span/src/cmp.rs
@@ -1,0 +1,158 @@
+/// This trait works similarly to [PartialEq] but it gives the liberty of checking the equality of the
+/// content loosely. This would mean the implementor can skip some parts of the content while doing
+/// equality checks.
+/// As an example, In AST types we ignore fields such as [crate::Span].
+///
+/// One should always prefer using the [PartialEq] over this since implementations of this trait
+/// inherently are slower or in the best-case scenario as fast as the [PartialEq] comparison.
+pub trait ContentEq<Rhs: ?Sized = Self> {
+    /// This method tests for contents of `self` and `other` to be equal.
+    #[must_use]
+    fn content_eq(&self, other: &Rhs) -> bool;
+
+    /// This method tests for contents of `self` and `other` not to be equal.
+    /// The default implementation is almost always
+    /// sufficient, and should not be overridden without very good reason.
+    #[inline]
+    #[must_use]
+    fn content_ne(&self, other: &Rhs) -> bool {
+        !self.content_eq(other)
+    }
+}
+
+impl ContentEq for () {
+    #[inline]
+    fn content_eq(&self, _other: &()) -> bool {
+        true
+    }
+
+    #[inline]
+    fn content_ne(&self, _other: &()) -> bool {
+        false
+    }
+}
+
+/// Blanket implementation for references
+impl<A: ?Sized, B: ?Sized> ContentEq<&B> for &A
+where
+    A: ContentEq<B>,
+{
+    #[inline]
+    fn content_eq(&self, other: &&B) -> bool {
+        ContentEq::content_eq(*self, *other)
+    }
+    #[inline]
+    fn content_ne(&self, other: &&B) -> bool {
+        ContentEq::content_ne(*self, *other)
+    }
+}
+
+/// Blanket implementation for mutable references
+impl<A: ?Sized, B: ?Sized> ContentEq<&mut B> for &mut A
+where
+    A: ContentEq<B>,
+{
+    #[inline]
+    fn content_eq(&self, other: &&mut B) -> bool {
+        ContentEq::content_eq(*self, *other)
+    }
+    #[inline]
+    fn content_ne(&self, other: &&mut B) -> bool {
+        ContentEq::content_ne(*self, *other)
+    }
+}
+
+/// Blanket implementation for mixed references
+impl<A: ?Sized, B: ?Sized> ContentEq<&mut B> for &A
+where
+    A: ContentEq<B>,
+{
+    #[inline]
+    fn content_eq(&self, other: &&mut B) -> bool {
+        ContentEq::content_eq(*self, *other)
+    }
+    #[inline]
+    fn content_ne(&self, other: &&mut B) -> bool {
+        ContentEq::content_ne(*self, *other)
+    }
+}
+
+/// Blanket implementation for mixed references
+impl<A: ?Sized, B: ?Sized> ContentEq<&B> for &mut A
+where
+    A: ContentEq<B>,
+{
+    #[inline]
+    fn content_eq(&self, other: &&B) -> bool {
+        ContentEq::content_eq(*self, *other)
+    }
+    #[inline]
+    fn content_ne(&self, other: &&B) -> bool {
+        ContentEq::content_ne(*self, *other)
+    }
+}
+
+/// Blanket implementation for [Option] types
+impl<T: ContentEq> ContentEq for Option<T> {
+    #[inline]
+    fn content_eq(&self, other: &Self) -> bool {
+        // NOTE: based on the standard library
+        // Spelling out the cases explicitly optimizes better than
+        // `_ => false`
+        #[allow(clippy::match_same_arms)]
+        match (self, other) {
+            (Some(lhs), Some(rhs)) => lhs.content_eq(rhs),
+            (Some(_), None) => false,
+            (None, Some(_)) => false,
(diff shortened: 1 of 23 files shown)


Boshen (approved):

crates/oxc_span/src/cmp.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.as_ref().content_eq(other.as_ref())
+    }
+}
+
+/// Blanket implementation for [oxc_allocator::Vec] types
+///
+/// # WARNING
+///
+/// This implementation is slow compared to `[PartialEq]`, Consider comparing the 2 vectors using
+/// that when one is implemented. This implementation takes triple the time the `PartialEq` would take.
```
overlookmotel, line 125: Why is `ContentEq` so much slower than `PartialEq`?

crates/oxc_span/src/cmp.rs
```diff
@@ -0,0 +1,158 @@
...
+{
+    #[inline]
+    fn content_eq(&self, other: &&B) -> bool {
+        ContentEq::content_eq(*self, *other)
+    }
+    #[inline]
+    fn content_ne(&self, other: &&B) -> bool {
+        ContentEq::content_ne(*self, *other)
+    }
+}
```
overlookmotel, line 93: Are all these impls necessary? Doesn't auto-deref handle this for us without explicit impls? (that's a question, not a statement)

crates/oxc_span/src/cmp.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// Blanket implementation for [oxc_allocator::Box] types
+impl<'a, T: ContentEq> ContentEq for oxc_allocator::Box<'a, T> {
+    #[inline]
+    fn content_eq(&self, other: &Self) -> bool {
+        self.as_ref().content_eq(other.as_ref())
+    }
+}
```
overlookmotel, line 118: Is this required? Doesn't auto-deref handle this?

e.g. even without this blanket impl, doesn't this work?

```rs
let box1 = Box::new_in(NullLiteral { span: Span::new(0, 1) }, allocator);
let box2 = Box::new_in(NullLiteral { span: Span::new(1, 2) }, allocator);
let equal = box1.content_eq(&box2);
```

overlookmotel (commented):
I'm late to the party and this has already been merged, but...

As well as specific comments below, a broader thought:

Currently `ContentEq` is defined as `pub trait ContentEq<Rhs: ?Sized = Self>`. So it allows comparing one type to another type.

I wonder if this makes it broader than it needs to be?

The original problem we needed to solve was something specific: Compare AST nodes while missing out certain fields (`span`, `scope_id` etc). So I had imagined the scope of `ContentEq` being:

1. Comparing AST nodes.
2. Comparing an AST node to another node *of the same type*.

In this use case, "content equality" means the nodes have equal "core content", ignoring "context" - "context" being where it sits in AST or source text.

But in this implementation, you've made `ContentEq` a much broader thing that could be applied to any type, and allows comparing to any type (including *other* types). The meaning of "content equality" is implementation-defined, and can be whatever the implementation wants it to.

There are positives and negatives to this:

* Positive: More flexible.
* Negative: Potentially confusing. When it's this broad, what does "content equality" even mean?

To judge the balance between the pluses and minuses, some questions:

1. What other uses of "content equality" can we conceive of needing?
2. Would those equality relations best be considered "content equality", or would they be better modelled as some *other* equality relation? e.g. "scope equality" which would have a separate `ScopeEq` trait (I have no idea what "scope equality" would mean, just a spurious example)

In my personal opinion, if we don't have good answers to these questions, making `ContentEq` broader than the original use case might be considered "premature abstraction".

We need people to understand the difference between `==` and `content_eq`. But that's harder to do if we don't have a solid definition of "content equality".

What do you think @rzvxa?

tasks/ast_tools/src/derives/content_eq.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub struct DeriveContentEq;
+}
+
+const IGNORE_FIELDS: [(/* field name */ &str, /* field type */ &str); 5] = [
+    ("span", "Span"),
+    ("trailing_comma", "Span"),
+    ("scope_id", "ScopeId"),
+    ("symbol_id", "SymbolId"),
+    ("reference_id", "ReferenceId"),
+];
```
overlookmotel, line 23: Could we make this a list of types to ignore, rather than field name + type pairs?

I assume we should be ignoring *any* `Span` (regardless of what field it appears in). If we add any new fields which contain a `Span` to AST types in future, we'll have to remember to add them here too.

Boshen had to remember to do that in his amends to #5327 (and he did remember), but if this PR wasn't fresh in the mind, it would have been very easy to forget.

crates/oxc_span/src/cmp.rs
rzvxa: This is the extreme case, Most of the time it doesn't perform that much worse however this is a thing.

I should update this comment to better explain it, Inherintly `ContentEq` does more work. So when a type implements `PartialEq` we should always prefer it as it would more or less optimize down to something similar to a `memcmp` + a length check.

Otherwise, for simple types, these 2 would result in almost identical assembly when optimization is on.

https://godbolt.org/z/KzaTxTd7M

crates/oxc_span/src/cmp.rs
rzvxa: Yes Sir! It was a mistake on my part, As it turns out Rust doesn't support auto deref on operators and that's why the standard library has to implement reference blanket implementations for operators. I'll make sure to remove these.

crates/oxc_span/src/cmp.rs
rzvxa: This one is required to allow this syntax:

```rust
let box1 = &Box::new_in(NullLiteral { span: Span::new(0, 1) }, allocator);
let box2 = &Box::new_in(NullLiteral { span: Span::new(1, 2) }, allocator);
let equal = ContentEq::content_eq(box1, box2);
```

This is really common in enums so it is worth having it around.

We are moving towards this syntax to avoid accidental hiccups in refactors. #5462

tasks/ast_tools/src/derives/content_eq.rs
rzvxa: I wasn't sure about it since our types aren't context-aware so for less common type names you can have code like this:

```rust
// locally used type name.
type ReferenceId = usize;
#[ast]
#[generate_derive(ContentEq)]
struct X {
    span: Span,
    foo: ReferenceId,
}
```

Something like `ReferenceId` isn't that crazy to be used by someone who isn't aware of this and would mess up their equality checks for good(it is always true in this example).

What if we ignore the `Span` type name but use the `name + type` combo for the rest? Or we can actually look at the use statements and resolve the type names, I think it would also be a welcomed change in the `oxc_regular_expression` crate, `Atom` in regex means something else, and our atom is usually used there as `use oxc_span::Atom as SpanAtom`, But I had to change it back to `Atom` in the ast file to make the codegen recognize it.

rzvxa:
I'll submit a follow-up PR addressing these issues and simplifying the implementation.

crates/oxc_span/src/cmp.rs
overlookmotel: Thanks for explaining. That makes sense.

crates/oxc_span/src/cmp.rs
overlookmotel: These were removed in #5468.

overlookmotel:
#5468 narrows the scope of `ContentEq` as discussed [above](https://github.com/oxc-project/oxc/pull/5427#pullrequestreview-2281036577).

crates/oxc_span/src/cmp.rs
overlookmotel: Interesting. Yes you're right - for native primitive types, `PartialEq` is faster.

But if the type being compared is even a native primitive wrapped in a struct, then `ContentEq` and `PartialEq` seem to produce the same assembly: https://godbolt.org/z/54on5sMWc

NB: Even adding `#[derive(Clone, Copy, Eq)]` on `WrappedU32` doesn't make it kick in the memcmp optimization.

I guess `std` has specialized impls for `PartialEq` of slices of primitives types, but they don't generalize to slices of `Copy` types in general. I don't think we could do that even if we wanted to, because specialization requires nightly features, if I understand right.

Luckily for us, the AST doesn't contain any `Vec<u32>`s (or similar), so for our purposes I think `ContentEq` and `PartialEq` are going to be equally fast.

PR #5430, opened 2024-09-03
feat(linter/eslint): implement `sort-vars` rule

jelly:


crates/oxc_linter/src/rules/eslint/sort_vars.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// ### Why is this bad?
+    ///
+    /// Unsorted variable declarations can make the code harder to read and maintain.
+    ///
+    /// ### Examples
+    ///
+    /// Examples of **incorrect** code for this rule:
+    /// ```js
+    /// var b, a;
+    /// var a, B, c;
```
DonIsaac, line 32: Isn't this correct? If not, why is `a, B, d` correct?

crates/oxc_linter/src/rules/eslint/sort_vars.rs
```diff
@@ -0,0 +1,226 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        let AstKind::VariableDeclaration(var_decl) = node.kind() else {
+            return;
+        };
+
+        if var_decl.declarations.len() <= 1 {
+            return;
+        };
+
+        let binding_declarations: Vec<&VariableDeclarator> = var_decl
```
DonIsaac, line 65: Can we iterate without collecting?

crates/oxc_linter/src/rules/eslint/sort_vars.rs
```diff
@@ -0,0 +1,226 @@
...
+                    OxcDiagnostic::warn("Variable declarations should be sorted")
+                        .with_label(curr.span),
+                );
+            }
+        }
+    }
+}
+
+impl SortVars {
+    fn get_sortable_name(&self, decl: &VariableDeclarator) -> String {
```
DonIsaac, line 91: Consider returning a Cow so we don't allocate when we don't need to.

jelly: Do you mean returning this [Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html)? 

I'm afraid I don't understand what you are proposing, can you give an example of somewhere where this is used in oxlint?

crates/oxc_linter/src/rules/eslint/sort_vars.rs
jelly: We can, I grabbed some inspiration from sort_imports. It still uses filter() in theory that can be refactored out as well.

crates/oxc_linter/src/rules/eslint/sort_vars.rs
DonIsaac:
```rs
fn get_sortable_name<'a>(&self, decl: &VariableDeclarator<'a>) -> Cow<'a, str> {
      let BindingPatternKind::BindingIdentifier(ident) = &decl.id.kind else {
            unreachable!();
        };

        if self.ignore_case {
            return Cow::Owned(ident.name.to_lowercase());
        }

        Cow::Borrowed(ident.name.as_str())
    }
}
```

better yet, just pass in the thing you want to compare as a parameter and use
`.eq_ignore_ascii_case()`

DonIsaac:
This is good work. Ping me when you're ready for another review.

crates/oxc_linter/src/rules/eslint/sort_vars.rs
jelly: Thanks! We could pass both decl's in theory but we need to test comparison not equality so not sure how `.eq_ignore_ascii_case()` would help

jelly: Ah, or is that an optimization for the case where it is equal?

crates/oxc_linter/src/rules/eslint/sort_vars.rs
```diff
@@ -0,0 +1,223 @@
...
+            .declarations
+            .iter()
+            .filter(|decl| matches!(decl.id.kind, BindingPatternKind::BindingIdentifier(_)))
+        {
+            if let Some(previous) = previous {
+                if self.get_sortable_name(previous).cmp(&self.get_sortable_name(current))
+                    == Ordering::Greater
+                {
+                    ctx.diagnostic(
+                        OxcDiagnostic::warn("Variable declarations should be sorted")
```
DonIsaac, line 76: Create a diagnostic factory function please

jelly: Sure!

DonIsaac (approved):

PR #5431, opened 2024-09-03
fix(ast, codegen, linter): panics in fixers.

### LINKED ISSUE #5434: Panic `byte index 18 is out of bounds of` in `crates/oxc_span/src/span/mod.rs`
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
switch(n){case'':/s/}
```

command
```
timeout -v 100 oxlint -D all -D nursery --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin TEST___FILE.js --fix
```

cause this
```
thread '<unnamed>' panicked at crates/oxc_span/src/span/mod.rs:315:21:
byte index 18 is out of bounds of ``
stack backtrace:

##### Automatic Fuzzer note, output status "None", output signal "Some(6)"
```

Looks like quite new panic, introduced ~1/2 day ago

[compressed.zip](https://github.com/user-attachments/files/16862063/compressed.zip)

rzvxa:
Closes #5434

https://github.com/oxc-project/oxlint-ecosystem-ci/actions/runs/10685877915/job/29619941099

### DIFF
--- a/crates/oxc_ast/src/ast_impl/literal.rs
+++ b/crates/oxc_ast/src/ast_impl/literal.rs
@@ -6,6 +6,7 @@
 use crate::ast::*;
 
 use std::{
+    borrow::Cow,
     fmt,
     hash::{Hash, Hasher},
 };
@@ -132,10 +133,11 @@ impl<'a> RegExpPattern<'a> {
         self.len() == 0
     }
 
-    pub fn source_text(&self, source_text: &'a str) -> &'a str {
+    pub fn source_text(&self, source_text: &'a str) -> Cow<str> {
         match self {
-            Self::Raw(raw) | Self::Invalid(raw) => raw,
-            Self::Pattern(pat) => pat.span.source_text(source_text),
+            Self::Raw(raw) | Self::Invalid(raw) => Cow::Borrowed(raw),
+            Self::Pattern(pat) if pat.span.is_unspanned() => Cow::Owned(pat.to_string()),
+            Self::Pattern(pat) => Cow::Borrowed(pat.span.source_text(source_text)),
         }
     }
 
--- a/crates/oxc_codegen/tests/integration/unit.rs
+++ b/crates/oxc_codegen/tests/integration/unit.rs
@@ -1,4 +1,4 @@
-use crate::tester::{test, test_minify};
+use crate::tester::{test, test_minify, test_without_source};
 
 #[test]
 fn module_decl() {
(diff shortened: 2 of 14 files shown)


rzvxa:
* **#5431** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camc314:
does this fix all of the panics in that action?

Boshen (changes requested):
Oh wait I forgot to say that we should add some tests. I intend to change how codegen deal with source_text at some point.

crates/oxc_ast/src/ast_impl/literal.rs
Boshen, line 1: how do I block the merge queue lol

Boshen (dismissed):
block!

rzvxa:
> Oh wait I forgot to say that we should add some tests. I intend to change how codegen deals with source_text at some point.

I am converting this to draft while I'm trying to find a good way to test it.

Do you have any specific approach for testing this in mind? What part should I test? The reason behind the panic was that I assumed `source_text` in the codegen would always be set similarly to formatter, That's why we encountered out-of-bound issues when trying to generate regexes. Now that fixers are operating on a codegen with source_text, This error is resolved.

The `pattern.source_text` method is a dead simple function and `to_string`/`Display` has its own set of tests.

We use source_text when available because the regex pattern loses some of the information from the original input so subsequently in some cases, we can't recreate the exact original input with the `Display` trait. We might even want to do the same thing for regex flags because as of now we change their order(However this is a small issue compared to the pattern itself which is more annoying if it changes unintentionally).

Boshen:
The linter rule tests should have at least 1 fixer test.

We should also have one integration test in apps/oxlint/src/lint/mod.rs with the fixture in apps/oxlint/fixtures/linter/debugger.js

crates/oxc_codegen/tests/integration/unit.rs
```diff
@@ -63,6 +63,33 @@ fn unicode_escape() {
     test("console.log('🧑‍🤝‍🧑');", "console.log(\"🧑‍🤝‍🧑\");\n");
 }
 
+#[test]
+fn regex() {
```
rzvxa, line 67: Added some tests here, It would've caught the issue we had.
I'm not sure exactly how to write the oxlint integration test, Maybe we can write a temp file and apply a bunch of fixes to it to make sure the whole lint cycle is working as expected but I feel that should be its own PR.

rzvxa: There are also tests available for rules with fixers, The only reason that it was uncaught was that none of our regex fixers were using codegen, The root of the issue was that codegen was panicking when the source wasn't available to it.
These tests make sure we generate the correct regexes with or without the source. Now having the original source in the linter becomes a decision rather than a bug and that's why I have difficulties imagining a useful unit test for it.

PR #5441, opened 2024-09-04
fix(semantic): `IdentifierReference` within `TSPropertySignature` cannot reference type-only import binding

### LINKED ISSUE #5435: semantic: imported type is used but a warning message appears
When you import a type and then use this type as the key of an interface, a warning message will appear saying that this type is not used and it is recommended to remove it.

The sample code is as follows:
```ts
import type a from 'b'

export interface A {
  [a]: 'test' | '10'
}
```
I created a repro [link](https://playground.oxc.rs/#eNo9jb0OwjAMhF/F8pKlA6zZeA7KEIKDIjU/clypVcm7k4DSyXe+8+cDLWr0IScWkD0TGHCcAqinmuMcafslPgqxM5bgBsccAe7moUEJFVHwAXW9tHbtBzhhQn0gr7GPskcxG2rhlSZcGmfoYlOm0+zhmZbhhE0sLnFA7cxSqE6YDRfiRmy6U4Y+q2Nh04ve1H83E3z0zv+jWr/8NVGr) with the above code, and you can see the warning message when you open it.

Dunqing:
close: #5435

The behavior of `IdentifierReference` in `TSPropertySignature` is the same as in `TSTypeQuery`, both allow only reference value bindings and type-only import bindings.

I still use `ReferenceFlags::TSTypeQuery` here because I want to avoid producing many changes unrelated to the bug in this PR. I will refactor it in the follow-up PR soon

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1830,6 +1830,14 @@ impl<'a> SemanticBuilder<'a> {
             AstKind::TSInterfaceHeritage(_) => {
                 self.current_reference_flags = ReferenceFlags::Type;
             }
+            AstKind::TSPropertySignature(signature) => {
+                if signature.key.is_expression() {
+                    // interface A { [prop]: string }
+                    //               ^^^^^ The property can reference value or [`SymbolFlags::TypeImport`] symbol
+                    self.current_reference_flags =
+                        ReferenceFlags::Read | ReferenceFlags::TSTypeQuery; // TODO: Should use another flag
+                }
+            }
             AstKind::TSTypeQuery(_) => {
                 // type A = typeof a;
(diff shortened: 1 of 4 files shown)


Dunqing:
* **#5441** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (approved):

DonIsaac:
Should we add a test case to semantic's integration tests? maybe in `symbols.rs`?

Dunqing:
> Should we add a test case to semantic's integration tests? maybe in `symbols.rs`?

I added a fixture for testing, and as you can see from the snapshot the results are as expected，So I think that's enough.

PR #5444, opened 2024-09-04
refactor(semantic): change the reference flag to `ReferenceFlags::Type` if it is used within a `TSTypeQuery`

Dunqing:
So far, the `ReferenceFlags::TSTypeQuery` only used indicates it is referenced by `TSTypeQuery` that we can confirm the reference should be regarded as a type reference, namely `ReferenceFlags::Type`.

This PR adds a `ReferenceFlags::ValueAsType` instead of `ReferenceFlags::TSTypeQuery`.  The new flag has the same behavior as the previous one. But it looks more general and is not only used in `TSTypeQuery`. But now it is a temporary flag. We use it to resolve the symbol correctly and replace `ReferenceFlags::ValueAsTyoe` with `ReferenceFlags::Type` after resolved. 

Also, this change eliminates the inconsistency in behavior between the `Reference::is_type` and `ReferenceFlags::is_type` methods.

### DIFF
--- a/crates/oxc_semantic/src/reference.rs
+++ b/crates/oxc_semantic/src/reference.rs
@@ -117,6 +117,6 @@ impl Reference {
     /// Returns `true` if this reference is used in a type context.
     #[inline]
     pub fn is_type(&self) -> bool {
-        self.flags.is_type() || self.flags.is_ts_type_query()
+        self.flags.is_type()
     }
 }
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -479,23 +479,20 @@ impl<'a> SemanticBuilder<'a> {
                     let flags = reference.flags();
                     if flags.is_type() && symbol_flags.can_be_referenced_by_type()
                         || flags.is_value() && symbol_flags.can_be_referenced_by_value()
-                        || flags.is_ts_type_query() && symbol_flags.is_import()
+                        || flags.is_value_as_type()
+                            && (symbol_flags.can_be_referenced_by_value()
+                                || symbol_flags.is_type_import())
                     {
-                        // The non type-only ExportSpecifier can reference a type/value symbol,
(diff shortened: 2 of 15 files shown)


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5444?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5444** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5441** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_semantic/src/reference.rs
```diff
@@ -117,6 +117,6 @@ impl Reference {
     /// Returns `true` if this reference is used in a type context.
     #[inline]
     pub fn is_type(&self) -> bool {
-        self.flags.is_type() || self.flags.is_ts_type_query()
+        self.flags.is_type()
```
DonIsaac, line 120: this behavior is depended upon by `no-unused-vars`. If we want to change what `is_type` does, we need to re-add this API under a different name.

DonIsaac: maybe we should have an `is_used_as_type` method?

Dunqing: The behavior has not changed

 `ReferenceFlags` that used to be `ReferenceFlags::Read` | `ReferenceFlags::TSTypeQuery` are now `ReferenceFlags::Type`.

DonIsaac: Nothing fails = happy Don

PR #5445, opened 2024-09-04
refactor(linter): simplify `eslint/radix` rule

overlookmotel:
Refactor to remove duplicate logic.

PR #5446, opened 2024-09-04
fix(linter): `eslint/radix` rule correctly check for unbound symbols

overlookmotel:
`SymbolTable::get_symbol_id_from_name(name).is_none()` is not always correct, because it will return `false` if there is a binding *anywhere* in the AST with that name, whereas what we actually want to know is whether *this* `IdentifierReference` is referring to a global or not.

Instead, look up whether this reference is resolved or not using `SymbolTable::is_global_reference`.

The 3 test cases added were not working prior to this change.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/radix.rs
+++ b/crates/oxc_linter/src/rules/eslint/radix.rs
@@ -70,7 +70,7 @@ impl Rule for Radix {
             match call_expr.callee.without_parenthesized() {
                 Expression::Identifier(ident) => {
                     if ident.name == "parseInt"
-                        && ctx.symbols().get_symbol_id_from_name("parseInt").is_none()
+                        && ctx.symbols().is_global_reference(ident.reference_id().unwrap())
                     {
                         Self::check_arguments(self, call_expr, ctx);
                     }
@@ -81,7 +81,7 @@ impl Rule for Radix {
                     {
                         if ident.name == "Number"
                             && member_expr.property.name == "parseInt"
-                            && ctx.symbols().get_symbol_id_from_name("Number").is_none()
+                            && ctx.symbols().is_global_reference(ident.reference_id().unwrap())
                         {
                             Self::check_arguments(self, call_expr, ctx);
                         }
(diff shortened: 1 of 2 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5446?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5447** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5446** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5445** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5445?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_linter/src/rules/eslint/radix.rs
```diff
@@ -70,7 +70,7 @@ impl Rule for Radix {
             match call_expr.callee.without_parenthesized() {
                 Expression::Identifier(ident) => {
                     if ident.name == "parseInt"
-                        && ctx.symbols().get_symbol_id_from_name("parseInt").is_none()
+                        && ctx.symbols().is_global_reference(ident.reference_id().unwrap())
```
DonIsaac, line 73: Is there ever a case where `reference_id()` would return `None`?

overlookmotel: They are all `None` before semantic runs, and then all are always `Some` after semantic. Linter runs after semantic, so this `unwrap()` can never fail.

I would like to replace all these `Option<Id>`s with just plain `Id`s with 0 as a "sentinel value" equivalent to `None`, so we can avoid all this unwrapping (`ScopeId`, `SymbolId`, `ReferenceId`). But that's another issue...

camc314 (approved):

PR #5223, opened 2024-08-26
feat(ast)!: add `JSXElementName::IdentifierReference` and `JSXMemberExpressionObject::IdentifierReference`

Dunqing:
close: #3528

part of #4746

PR #5450, opened 2024-09-04
fix(linter): `tree_shaking/no_side_effects_in_initialization` handle JSX correctly

overlookmotel:
#5223 altered `JSXElementName` so `JSXElementName::Identifier` is used only for non-reference JSX names (e.g. `<div>`). `JSXElementName::IdentifierReference` is used where the name is a reference (e.g. `<Foo>`). Similarly `JSXMemberExpressionObject`'s `object` is always an `IdentifierReference` now.

So, the net result is that `JSXIdentifier` is now never a reference, it's just the JSX equivalent of `IdentifierName`. So I don't think `JSXIdentifier` can ever have side-effects.

This PR:

1. Removes `impl ListenerMap for JSXIdentifier`
2. Adds `impl ListenerMap for JSXMemberExpression` and makes sure the root `IdentifierReference` (`Foo` in `<Foo.bar.qux>`) is visited.

### DIFF
--- a/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/listener_map.rs
+++ b/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/listener_map.rs
@@ -7,11 +7,11 @@ use oxc_ast::{
         ConditionalExpression, Declaration, ExportSpecifier, Expression, ForStatementInit,
         FormalParameter, Function, IdentifierReference, JSXAttribute, JSXAttributeItem,
         JSXAttributeValue, JSXChild, JSXElement, JSXElementName, JSXExpression,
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#5450** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5450?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac:
I unfortunately have no context into this, but I can handle it if @camc314 is unable to.

overlookmotel:
We only merged because we wanted to get the upstack PR merged. The question of whether the error is correct or not is still open.

@Boshen Git blame says you wrote this code. Sounds like DonIsaac is not familiar with this lint rule, so it's hard for him to answer my query. Can you please take a look?

Boshen:
@mysteryven Can you double check whether the code is correct or not?

PR #5451, opened 2024-09-04
feat(ast): add `ContentHash` trait; remove noop `Hash` implementation from `Span`

### LINKED ISSUE #5283: ast: Derive `Hash` and `PartialEq` on AST node types

rzvxa:
closes #5283

Also removes the noop Hash implementation on `Span` in favor of a real implementation.

### DIFF
--- a/.github/.generated_ast_watch_list.yml
+++ b/.github/.generated_ast_watch_list.yml
@@ -24,4 +24,7 @@ src:
   - 'crates/oxc_ast/src/generated/derive_content_eq.rs'
   - 'crates/oxc_regular_expression/src/generated/derive_content_eq.rs'
   - 'crates/oxc_syntax/src/generated/derive_content_eq.rs'
+  - 'crates/oxc_ast/src/generated/derive_content_hash.rs'
+  - 'crates/oxc_regular_expression/src/generated/derive_content_hash.rs'
+  - 'crates/oxc_syntax/src/generated/derive_content_hash.rs'
   - 'tasks/ast_codegen/src/**'
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -9,7 +9,7 @@ use std::cell::Cell;
 
 use oxc_allocator::{Box, CloneIn, Vec};
 use oxc_ast_macros::ast;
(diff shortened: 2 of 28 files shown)


Boshen (approved):

Boshen:
I'll test this in Rolldown before merging.

PR #5467, opened 2024-09-05
fix(linter/tree-shaking): detect the correct export symbol resolution

### LINKED ISSUE #5455: Incorrect symbol resolution in `tree_shaking/no_side_effects_in_initialization`?
I am fairly sure that symbol resolution in this rule is not correct for `ExportSpecifier`:

https://github.com/oxc-project/oxc/blob/1d3e973f101fc2d03f705f75c7de9e9cb04ae50a/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/listener_map.rs#L193-L202

### Background

I would like to remove the method `SymbolTable::get_symbol_id_from_name`, as in my opinion it's a footgun (#5456). We use it to get the `SymbolId` for a reference, but actually that's not what it's doing - it will give you the `SymbolId` of the *first* binding with this variable name *anywhere* in the AST.

```js
{
    let foo;
}
let bar = foo;
```

#5446 removes a usage of `get_symbol_id_from_name` and adds tests which demonstrate it was incorrect.

### This rule

I'm pretty sure the usage of `get_symbol_id_from_name` in this rule is wrong too. Isn't the question we should be asking in this code "is the `local` a bound identifier?", not "is the `exported` a bound identifier?"

In the case of `export {foo}`, they're the same, but in case of `export {foo as bar}`, they're not.

However, I must admit this rule rather bends my brain! Can someone more familiar with this code advise?

@Boshen git "blames" you for this code. Can you help? But maybe @DonIsaac or @camc314 also may have an idea what this rule is about?

mysteryven:
fixes: #5455

### DIFF
--- a/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/listener_map.rs
+++ b/crates/oxc_linter/src/rules/tree_shaking/no_side_effects_in_initialization/listener_map.rs
@@ -8,10 +8,10 @@ use oxc_ast::{
         FormalParameter, Function, IdentifierReference, JSXAttribute, JSXAttributeItem,
         JSXAttributeValue, JSXChild, JSXElement, JSXElementName, JSXExpression,
         JSXExpressionContainer, JSXFragment, JSXMemberExpression, JSXMemberExpressionObject,
-        JSXOpeningElement, LogicalExpression, MemberExpression, NewExpression, ObjectExpression,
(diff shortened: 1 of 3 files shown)


mysteryven:
* **#5467** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Thanks!

PR #5470, opened 2024-09-05
fix(linter): Don't mark binding rest elements as unused in TS function overloads

### LINKED ISSUE #5406: linter: `no-unused-vars` false positive w/TS overload
`no-unused-vars` warns on the following code:

```ts
export function log(message: string, ...interpolations: unknown[]): void;
export function log(message: string, ...interpolations: unknown[]): void {
  console.log(message, interpolations);
}
```

It claims `interpolations` in the overload signature is unused. However, it only makes sense to check for unused vars in functions with bodies. Here's the output:

```
❯ npx oxlint@latest test.ts

  ⚠ eslint(no-unused-vars): Variable 'interpolations' is declared but never used.
   ╭─[test.ts:1:38]
 1 │ export function log(message: string, ...interpolations: unknown[]): void;
   ·                                      ──────────────┬─────────────
   ·                                                    ╰── 'interpolations' is declared here
 2 │ export function log(message: string, ...interpolations: unknown[]): void {
   ╰────

Finished in 84ms on 1 file with 93 rules using 3 threads.
Found 1 warning and 0 errors.
```

camchenry:
- Fixes https://github.com/oxc-project/oxc/issues/5406

This implements a fix for the `BindingRestElement` symbol, which is currently unhandled and gets automatically marked as unused. If we happen to find that it is a child of declaration, then we will automatically allow the binding rest element.

The code for this was based on what we currently do in `is_allowed_param_because_of_method`: https://github.com/oxc-project/oxc/blob/5187f384cb38b1ba69acc2cb9b677d72ef175e3e/crates/oxc_linter/src/rules/eslint/no_unused_vars/allowed.rs#L258

I opted not to refactor this to re-use the same code though, as I think the duplication is still incidental and the implementations could diverge in the future.


camc314 (commented):
can't we just check if the funciton is TS syntax https://github.com/oxc-project/oxc/blob/f81e8a126e83cb74cf03cd430121a310ecc218a0/crates/oxc_ast/src/ast_impl/js.rs#L1069-L1075 and skip checking it?

camchenry:
@camc314 Thanks for the tip. Since we are looking at whether a given symbol is allowed, I changed it to check if the binding rest element is a child of a `Function`. If so, if `is_typescript_syntax()` is true, then we will allow it, otherwise it fails. I think this makes it a lot simpler, let me know if this makes sense.

crates/oxc_linter/src/rules/eslint/no_unused_vars/allowed.rs
```diff
@@ -298,4 +298,21 @@ impl NoUnusedVars {
...
+    /// reported). Currently, this handles the case where a rest element is part
+    /// of a TS function declaration.
+    pub(super) fn is_allowed_binding_rest_element<'a>(&self, symbol: &Symbol<'_, 'a>) -> bool {
+        for parent in symbol.iter_parents() {
+            // If this is a binding rest element that is part of a TS function parameter,
+            // for example: `function foo(...messages: string[]) {}`, then we will allow it.
+            if let AstKind::Function(f) = parent.kind() {
+                if f.is_typescript_syntax() {
+                    return true;
+                }
```
camc314, line 312: i think we can just return f.is_typescript_syntax.

no need to continue up the tree if we've found a function

camc314:
one last comment then this looks good to me.

thanks for contributing!

crates/oxc_linter/src/rules/eslint/no_unused_vars/allowed.rs
DonIsaac: ^

DonIsaac: This also doesn't apply only to rest elements, it should apply to all parameters IMO.

DonIsaac (dismissed):
See: Cam's comment thread

DonIsaac:
@camchenry run `just fix` to auto-fix clippy/rustfmt/etc issues.

camc314:
interestingly enough @DonIsaac this isn't an error 🤔 

Screenshot 2024-09-05 at 20 40 55

crates/oxc_linter/src/rules/eslint/no_unused_vars/allowed.rs
camc314: cool, thanks!

camc314 (approved):

camc314 (approved):
Thank you!

PR #5469, opened 2024-09-05
fix(regex): panic on displaying surrogated `UnicodeEscape` characters.

rzvxa:
fixes https://github.com/oxc-project/oxc/pull/5387#issuecomment-2330534180

PR #5466, opened 2024-09-05
feat(ast)!: add `ThisExpression` variants to `JSXElementName` and `JSXMemberExpressionObject`

overlookmotel:
Close #5352.

Add to AST:

* `JSXElementName::ThisExpression` (`<this>`)
* `JSXMemberExpressionObject::ThisExpression` (`<this.foo>`, `<this.foo.bar>`)

PR #5462, opened 2024-09-04
refactor(ast, ast_tools): use full method path for generated derives trait calls.

rzvxa:
As of now if we remove the implementation of a trait for a type and implement the method on that type directly it wouldn't break while it isn't the original trait anymore so that method might do something entirely different.
This change is more explicit on trait calls so we hit compile errors on these kinds of changes.

PR #5471, opened 2024-09-05
feat(ast)!: add `ThisExpression` variants to `JSXElementName` and `JSXMemberExpressionObject` (#5466)

### LINKED ISSUE #5283: ast: Derive `Hash` and `PartialEq` on AST node types

### LINKED ISSUE #5352: Add `ThisExpression` variants to `JSXElementName` and `JSXMemberExpressionObject`
#5223 did the bulk of the work of revamping `JSXElementName` (#3528). Only thing that's missing is to add:

* `JSXElementName::ThisExpression`
* `JSXMemberExpressionObject::ThisExpression`

A couple of notes:

#### Arrow functions transform

This code is currently broken:

https://github.com/oxc-project/oxc/blob/f052a6d6664b4bfa3fae14445ce7558f8186e0c1/crates/oxc_transformer/src/es2015/arrow_functions.rs#L140-L156

We can fix after we have the `ThisExpression` variants.

#### Linter

Once this is done, we can remove these lines from linter.

https://github.com/oxc-project/oxc/blob/f052a6d6664b4bfa3fae14445ce7558f8186e0c1/crates/oxc_linter/src/rules/react/jsx_no_undef.rs#L60-L64

### LINKED ISSUE #5434: Panic `byte index 18 is out of bounds of` in `crates/oxc_span/src/span/mod.rs`

### LINKED ISSUE #5455: Incorrect symbol resolution in `tree_shaking/no_side_effects_in_initialization`?

Dunqing:
feat(ast)!: add `ThisExpression` variants to `JSXElementName` and `JSXMemberExpressionObject` (#5466)

Close #5352.

Add to AST:

* `JSXElementName::ThisExpression` (`<this>`)
* `JSXMemberExpressionObject::ThisExpression` (`<this.foo>`, `<this.foo.bar>`)

fix(linter/tree-shaking): detect the correct export symbol resolution (#5467)

fixes: #5455

feat(ast): add `ContentHash` trait; remove noop `Hash` implementation from `Span` (#5451)

closes #5283

Also removes the noop Hash implementation on `Span` in favor of a real implementation.

refactor(ast, ast_tools): use full method path for generated derives trait calls. (#5462)

As of now if we remove the implementation of a trait for a type and implement the method on that type directly it wouldn't break while it isn't the original trait anymore so that method might do something entirely different.
This change is more explicit on trait calls so we hit compile errors on these kinds of changes.

refactor(ast): simplify `ContentEq` trait definition. (#5468)

Addresses the concerns brought up in #5427

fix(ast, codegen, linter): panics in fixers. (#5431)

Closes #5434

https://github.com/oxc-project/oxlint-ecosystem-ci/actions/runs/10685877915/job/29619941099

fix(regex): panic on displaying surrogated `UnicodeEscape` characters. (#5469)

fixes https://github.com/oxc-project/oxc/pull/5387#issuecomment-2330534180

Merge remote-tracking branch 'origin/main' into 09-05-feat_transformer_support_all__regex__to_new_regexp_transforms

update snapshot


Dunqing:
* **#5471** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5471?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5480, opened 2024-09-05
refactor(semantic)!: remove `SymbolTable::get_symbol_id_from_name` and `SymbolTable::get_scope_id_from_name`

### LINKED ISSUE #5456: semantic!: Remove `SymbolTable::get_symbol_id_from_name` and `SymbolTable::get_scope_id_from_name`
These methods are in my opinion a footgun because they don't do what you might think they do. Worse still, minimal test cases often won't reveal that.

We use `get_symbol_id_from_name` to get the `SymbolId` for a reference, but actually that's not what it's doing - it will give you the `SymbolId` of the *first* binding with this variable name *anywhere* in the AST.

```js
{
    let foo;
}
let bar = foo;
```

Instead we should use `symbol_table.is_global_reference(ident.reference_id().unwrap())`.

As well as being correct, this is also much more performant, as it doesn't iterate through every single symbol in the entire AST the way `get_symbol_id_from_name` does.

#### `get_symbol_id_from_name`

* #5446 removes a usage and adds tests which demonstrate it was incorrect.
* #5450 removed another one.
* The only remaining usage of is in the `tree_shaking/no_side_effects_in_initialization` linter rule, and I believe it's incorrect there too (#5455).

#### `get_scope_id_from_name`

This method is not used anywhere in Oxc.

overlookmotel:
Close #5456.

### DIFF
--- a/crates/oxc_semantic/src/symbol.rs
+++ b/crates/oxc_semantic/src/symbol.rs
@@ -71,16 +71,6 @@ impl SymbolTable {
             .find_map(|(symbol, &inner_span)| if inner_span == span { Some(symbol) } else { None })
     }
 
-    pub fn get_symbol_id_from_name(&self, name: &str) -> Option<SymbolId> {
-        self.names.iter_enumerated().find_map(|(symbol, inner_name)| {
-            if inner_name.as_str() == name {
-                Some(symbol)
-            } else {
-                None
-            }
-        })
-    }
-
     #[inline]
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#5480** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5491, opened 2024-09-05
feat(xtask): impl `as_ast_kind` method for each variant

IWANABETHATGUY:


Boshen:
What's the usage?

IWANABETHATGUY:
> What's the usage?

e.g.
https://github.com/oxc-project/oxc/blob/eda3d101c052a8fa3afa2b7cc9bc8477f9df5089/crates/oxc_linter/src/rules/eslint/no_this_before_super.rs#L136-L150

could be rewritten as: 
```rs
    fn is_wanted_node(node: &AstNode, ctx: &LintContext<'_>) -> Option<bool> {
        let parent = ctx.nodes().parent_node(node.id())?;
        if let AstKind::MethodDefinition(mdef) = parent.kind() {
            if matches!(mdef.kind, MethodDefinitionKind::Constructor) {
                let parent_2 = ctx.nodes().parent_node(parent.id())?;
                let parent_3 = ctx.nodes().parent_node(parent_2.id())?;
                let c = parent_3.kind().as_class()?;
                let super_class = c.super_class.as_ref()?;
                return Some(!matches!(super_class, Expression::NullLiteral(_)));
            }
        }

        Some(false)
    }


```

IWANABETHATGUY:
This could avoid written many nested levels if let #ty pattern match: 
```if let AstKind::xxx() = ```

**Comment by IWANABETHATGUY** [comment]
Also, this makes the code more easily combine with `Option` combinator

**Comment by IWANABETHATGUY** [comment]
feel free to close it if you are not satisfied since it is not written by hand, :smile:

**Comment by IWANABETHATGUY** [comment]
It is used in many place in our codebase: 
image

File: tasks/ast_tools/src/generators/ast_kind.rs [hunk 1]
```diff
@@ -154,6 +155,23 @@ impl Generator for AstKindGenerator {
             .map(|(ident, _)| parse_quote!(Self :: #ident(it) => it.span()))
             .collect_vec();
 
+        let as_ast_kind_impls: Vec<ImplItemFn> = have_kinds
+            .iter()
+            .map(|(ident, typ)| {
+                let snake_case_name =
+                    format_ident!("as_{}", ident.to_string().to_case(Case::Snake));
+                parse_quote!(
+                    pub fn #snake_case_name(&self) -> Option<&'a #typ> {
```
overlookmotel, line 164: Please add `///@@line_break` before this line to insert line breaks into output, making the generated code easier to read.

(`///@@line_break` is a workaround for that `parse_quote!` and `quote!` macros do not take any notice of whitespace)

rzvxa (approved):
LGTM!

I'm not sure if we need to return the inner values with the lifetime of `a`; But other than this thought everything looks good.

IWANABETHATGUY:
> LGTM!
> 
> I'm not sure if we need to return the inner values with the lifetime of `a`; But other than this thought everything looks good.

Ok, I omit `'a` lifetime

rzvxa:
> > LGTM!
> > I'm not sure if we need to return the inner values with the lifetime of `a`; But other than this thought everything looks good.
> 
> Ok, I omit the `'a` lifetime

Thanks, That wasn't really a change request. I just wanted to put my concern out there to see what you guys think about it.

So please feel free to revert it if you think having the lifetime can be helpful(I'm not sure how many of these pattern matchings are looking for an inner value with the lifetime so it might very well be necessary to simplify many of these).

Boshen (approved):

IWANABETHATGUY:
> > > LGTM!
> > > I'm not sure if we need to return the inner values with the lifetime of `a`; But other than this thought everything looks good.
> > 
> > 
> > Ok, I omit the `'a` lifetime
> 
> Thanks, That wasn't really a change request. I just wanted to put my concern out there to see what you guys think about it.
> 
> So please feel free to revert it if you think having the lifetime can be helpful(I'm not sure how many of these pattern matchings are looking for an inner value with the lifetime so it might very well be necessary to simplify many of these).

I don't know it either, I think people could add it back if meets some scenario when found it is not convenient to use.

Boshen:
AST codegen is getting too powerful 🦸

shulaoda:
> This could avoid written many nested levels if let #ty pattern match: `if let AstKind::xxx() = `

But it seems that we can't stop the behavior of `if let Some() =`. Does this help with the following writing style?
https://github.com/oxc-project/oxc/blob/cedf7a4daa4084523050d9ae1b4555e5174c6347/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs#L69-L72

IWANABETHATGUY:
> > This could avoid written many nested levels if let #ty pattern match: `if let AstKind::xxx() = `
> 
> But it seems that we can't stop the behavior of `if let Some() =`. Does this help with the following writing style?
> 
> https://github.com/oxc-project/oxc/blob/cedf7a4daa4084523050d9ae1b4555e5174c6347/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs#L69-L72

Yeah you could just: 
```diff
diff --git a/crates/oxc_ast/src/generated/ast_kind.rs b/crates/oxc_ast/src/generated/ast_kind.rs
index 2c68cb958..e512d0034 100644
--- a/crates/oxc_ast/src/generated/ast_kind.rs
+++ b/crates/oxc_ast/src/generated/ast_kind.rs
@@ -696,7 +696,7 @@ impl<'a> AstKind<'a> {
         }
     }
 
-    pub fn as_call_expression(&self) -> Option<&CallExpression<'a>> {
+    pub fn as_call_expression(&self) -> Option<&'a CallExpression<'a>> {
         if let Self::CallExpression(v) = self {
             Some(*v)
         } else {
@@ -936,7 +936,7 @@ impl<'a> AstKind<'a> {
         }
     }
 
-    pub fn as_if_statement(&self) -> Option<&IfStatement<'a>> {
+    pub fn as_if_statement(&self) -> Option<&'a IfStatement<'a>> {
         if let Self::IfStatement(v) = self {
             Some(*v)
         } else {
diff --git a/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs b/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs
index 17be91c88..ce9be1882 100644
--- a/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs
+++ b/crates/oxc_linter/src/rules/vitest/no_conditional_tests.rs
@@ -66,30 +66,27 @@ impl Rule for NoConditionalTests {
     }
 }
 
-fn run<'a>(possible_jest_node: &PossibleJestNode<'a, '_>, ctx: &LintContext<'a>) {
+fn run<'a>(possible_jest_node: &PossibleJestNode<'a, '_>, ctx: &LintContext<'a>) -> Option<()> {
     let node = possible_jest_node.node;
-    if let AstKind::CallExpression(call_expr) = node.kind() {
-        if is_type_of_jest_fn_call(
-            call_expr,
-            possible_jest_node,
-            ctx,
-            &[
-                JestFnKind::General(JestGeneralFnKind::Describe),
-                JestFnKind::General(JestGeneralFnKind::Test),
-            ],
-        ) {
-            let if_statement_node = ctx
-                .nodes()
-                .iter_parents(node.id())
-                .find(|node| matches!(node.kind(), AstKind::IfStatement(_)));
+    let call_expr = node.kind().as_call_expression()?;
+    if is_type_of_jest_fn_call(
+        call_expr,
+        possible_jest_node,
+        ctx,
+        &[
+            JestFnKind::General(JestGeneralFnKind::Describe),
+            JestFnKind::General(JestGeneralFnKind::Test),
+        ],
+    ) {
+        let if_statement_node = ctx
+            .nodes()
+            .iter_parents(node.id())
+            .find(|node| matches!(node.kind(), AstKind::IfStatement(_)))?;
 
-            let Some(node) = if_statement_node else { return };
-
-            if let AstKind::IfStatement(if_statement) = node.kind() {
-                ctx.diagnostic(no_conditional_tests(if_statement.span));
-            }
-        }
+        let if_stmt = if_statement_node.kind().as_if_statement()?;
+        ctx.diagnostic(no_conditional_tests(if_stmt.span));
     }
+    None
 }
 
 #[test]
```

IWANABETHATGUY:
I need to add the `'a` back , or the code not gonna compile, :cry: @rzvxa

rzvxa:
> I need to add the `'a` back , or the code not gonna compile, 😢 @rzvxa

Yes, it seems like it😓

I like this implementation(#5506) better than the one with the reference, What are your thoughts on it?

PR #5492, opened 2024-09-05
refactor(transformer): RegExp transform do not take ownership of `Pattern` then reallocate it

overlookmotel:
Previously `try_parse_pattern` took ownership of an existing pattern and then reallocated it back into arena, even if it wasn't changed.

Rust's borrow-checker makes it quite hard to avoid this. Only way I could find is to move some of the logic out of `try_parse_pattern`.

PR #5494, opened 2024-09-05
fix(transformer): RegExp transform do not transform invalid regexps

overlookmotel:
Treat invalid regexps the same regardless of whether they have unsupported flags or not - don't transform them.

### DIFF
--- a/crates/oxc_transformer/src/regexp/mod.rs
+++ b/crates/oxc_transformer/src/regexp/mod.rs
@@ -150,8 +150,9 @@ impl<'a> Traverse<'a> for RegExp<'a> {
         }
 
         let pattern_source: Cow<'_, str> = match &regexp.regex.pattern {
-            RegExpPattern::Raw(raw) | RegExpPattern::Invalid(raw) => Cow::Borrowed(raw),
+            RegExpPattern::Raw(raw) => Cow::Borrowed(raw),
             RegExpPattern::Pattern(p) => Cow::Owned(p.to_string()),
+            RegExpPattern::Invalid(_) => return,
         };
 
         let callee = {


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5494?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5494** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5492** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
I *think* this is the right

overlookmotel:
This PR makes the behavior consistent regardless of whether the regex has unsupported flags or not - if the regexp is invalid, don't transform it.

But... is this the right way around?

I *think* so - if it's invalid, then that's an early syntax error and we shouldn't transform it as it'll convert it to a runtime error, which makes that error harder to find.

But the opposing argument is that if the regexp *hasn't* been parsed, it may or may not be invalid, and if it is invalid, then we are disguising that with the transform.

@Dunqing What do you think?

overlookmotel:
Maybe #5496 helps make this OK...

Dunqing:
> Maybe #5496 helps make this OK...

Yes

Dunqing (approved):

PR #5495, opened 2024-09-05
feat(transformer): add `TransformOptions::enable_all` method

Boshen:
Make it **really** explicit about which transformer options are being turned on. I also need this in monitor-oxc.


Boshen:
Converting to draft and trying to find the broken code ...

Dunqing (approved):

overlookmotel (commented):
Great to have this explicit for benchmarks.

Personally, I think it might be best to leave the example as it was, as that's more typical real world use.

PR #5501, opened 2024-09-05
fix(transformer): RegExp transform handle `Term::Quantifier`

overlookmotel:
`Term::Quantifier` contains a nested `Term` so we need to recurse into that `Term` to check if it contains any unsupported syntax.


overlookmotel:
* **#5501** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5501?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@Dunqing I mentioned this in https://github.com/oxc-project/oxc/pull/5387#pullrequestreview-2278003275 but you chose not to implement it. Is there a reason why which I'm not seeing? (that's a genuine question, not having a dig)

Dunqing:
> @Dunqing I mentioned this in https://github.com/oxc-project/oxc/pull/5387#pullrequestreview-2278003275 but you chose not to implement it. Is there a reason why which I'm not seeing? (that's a genuine question, not criticism)
> 

I just realized I forgot to submit this code

overlookmotel:
I tried, but I can't work it out. My starting point was to add a test which should fail, so I could make sure it was running. But it doesn't fail.

Here's what I did: #5509

Please can you explain me the error of my ways?

Dunqing:
> I tried, but I can't work it out. My starting point was to add a test which should fail, so I could make sure it was running. But it doesn't run the test.
> 
> Here's what I did: #5509
> 
> Please can you explain me the error of my ways?


The way we add tests is the same as Babel. You added a test in `tasks/transform_conformance/tests/regexp/nested-terms`. From the path. we regard `regexp` as a plugin name, but there is another requirement that all tests should be under the `plugin_name/tests/fixtures`, so this would work if you move your test to `tasks/transform_conformance/tests/regexp/test/fixtures/nested-terms`


For example, if we need to add a typescript test, we will add in

Babel: https://github.com/babel/babel/tree/main/packages/babel-plugin-transform-typescript/test/fixtures
Oxc: https://github.com/oxc-project/oxc/tree/main/tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures

Since the RegExp plugin doesn't port Babel's implementation, I have added all related tests for regexp in
https://github.com/oxc-project/oxc/tree/main/tasks/transform_conformance/tests/esbuild-tests/test/fixtures/regexp

Surely you can add a test under the `regexp-plugin-name/test/fixtures` as well

I hope I explained clearly!

Dunqing (approved):

PR #5503, opened 2024-09-05
chore: run `cargo +nightly fmt` to sort imports

Boshen:
They are never going to be stable are they ... https://github.com/oxc-project/oxc/blob/cedf7a4daa4084523050d9ae1b4555e5174c6347/.rustfmt.toml#L8-L16


DonIsaac:
Nightly rust is the new stable rust 👀

PR #5504, opened 2024-09-05
refactor(transformer): RegExp transform unbox early

overlookmotel:
Unbox the `RegExpLiteral` early to avoid all lookups thereafter going through the indirection of a `Box`.

### DIFF
--- a/crates/oxc_transformer/src/regexp/mod.rs
+++ b/crates/oxc_transformer/src/regexp/mod.rs
@@ -113,9 +113,10 @@ impl<'a> Traverse<'a> for RegExp<'a> {
         expr: &mut Expression<'a>,
         ctx: &mut oxc_traverse::TraverseCtx<'a>,
     ) {
-        let Expression::RegExpLiteral(ref mut regexp) = expr else {
+        let Expression::RegExpLiteral(regexp) = expr else {
             return;
         };
+        let regexp = regexp.as_mut();
 
         let flags = regexp.regex.flags;
         let has_unsupported_flags = flags.intersects(self.unsupported_flags);


overlookmotel:
This is an experiment. I'm not sure if it'll make any difference.

overlookmotel:
No difference in benchmarks, but it's probably not a bad idea anyway.

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -113,9 +113,10 @@ impl<'a> Traverse<'a> for RegExp<'a> {
         expr: &mut Expression<'a>,
         ctx: &mut oxc_traverse::TraverseCtx<'a>,
     ) {
-        let Expression::RegExpLiteral(ref mut regexp) = expr else {
+        let Expression::RegExpLiteral(regexp) = expr else {
             return;
         };
+        let regexp = &mut **regexp;
```
rzvxa, line 119: Maybe we can use `DerefMut` here, I always find too many symbols to be confusing compared to this:

```rust
let regexp = regexp.deref_mut();
```

It's just a personal opinion, Though I may as well share it in case you find it tasteful.

overlookmotel: I agree! `&mut **` is rather confusing. But clippy seems to disagree (always ruining the fun, clippy!)

deref

I think what we need is `Box::as_mut`. #5515

overlookmotel: It's now `regexp.as_mut()`. Much better.

Dunqing (approved):

PR #5512, opened 2024-09-05
style(linter): introduce the writing style from PR #5491 and reduce the if nesting

shulaoda:
Related to #5491


Boshen:
Thank you for spending the time on this.

PR #5517, opened 2024-09-05
fix(linter): handle loops in `getter-return` rule

### LINKED ISSUE #3935: linter: getter-return false positive
```javascript
  get steps() {
    const steps = [];

    const gradient = colorGradient('#3D68FF', '#8845FF', this.maxStepsNumber);

    for (let i = 0; i < this.maxStepsNumber; i++) {
      steps.push({
        stepnum: i + 1,
        status: i <= this.currentStepIndex ? 'active' : '',
        background: htmlSafe(`background: ${gradient[i]};`),
      });
    }

    return steps;
  }
```

https://github.com/1024pix/pix/blob/1352bd8d7f6070f1ff8da79867f543c1c1926e59/mon-pix/app/components/progress-bar.js#L29-L43

camchenry:
- Fixes https://github.com/oxc-project/oxc/issues/3935

Prior to this, it didn't seem like we were correctly continuing to search the graph after we encountered a backedge (loop). Now, when encountering a cycle, we will not automatically break or prune the search.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/getter_return.rs
+++ b/crates/oxc_linter/src/rules/eslint/getter_return.rs
@@ -257,6 +257,7 @@ impl GetterReturn {
                                 e.weight(),
                                 EdgeType::Jump
                                     | EdgeType::Normal
+                                    | EdgeType::Backedge
                                     | EdgeType::Error(ErrorEdgeKind::Explicit)
                             )
                         }) {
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/eslint/getter_return.rs
```diff
@@ -426,6 +478,31 @@ fn test() {
...
+            None,
+        ),
+        (
+            "
+            var foo = {
+                get bar() {
+                    let i = 0;
+                    while (i < 10) {
+                        return i;
+                    }
```
camchenry, line 500: Although this might actually return in all cases practically, in general it would be impossible to know if this actually returns or not (halting problem). So, we just assume that it won't and require it to be explicit

rzvxa:
Thank you for your contribution, Funnily enough, earlier this day I was looking into this issue; I'm so glad you submitted this PR as early as you did otherwise we would've lost some time working on the same thing.

crates/oxc_linter/src/rules/eslint/getter_return.rs
rzvxa: Right, In most cases we can't detect it, However, We have `ControlFlowGraph::is_infinite_loop_start` which does a super simple constant evaluation that detects things like `while(true)` and `for (;;)`; We may want to enhance this rule in the future to pass things like this:

```js
class foo {
  get bar() {
    while(true) { return 0 }
  }
}
```

While linting code like:

```js
class foo {
  get bar() {
    while(false) { return 0 }
  }
}
```

However, it isn't necessary for now and seems out of the scope of this PR.

rzvxa (changes requested):

crates/oxc_linter/src/rules/eslint/getter_return.rs
rzvxa: You are right, We have means to detect infinite loops like `while(true)` and `for(;;)` via constant evaluation but supporting it here might be out of scope.

rzvxa (approved):

rzvxa:
Please update snapshots so the tests pass.
I personally use `cargo insta test` to regenerate all snapshots.

camchenry:
@rzvxa Thanks, just updated it now!

PR #5519, opened 2024-09-06
refactor(ast_tools): add some minimal logging.

rzvxa:


rzvxa:
* **#5519** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rzvxa:
@Boshen For some reason Graphite is stuck in the merge, So I'll use this chance to do a small edit, Marking it as draft in the meanwhile.

tasks/ast_tools/src/main.rs
```diff
@@ -128,5 +137,41 @@ fn write_ci_filter(
...
+
+    /// Shouldn't be called after first use of the logger macros.
+    pub(super) unsafe fn quiet() {
+        LOG = false;
+    }
+
+    pub(super) fn __internal_log_enable() -> bool {
+        // SAFETY:`LOG` doesn't change from the moment we start using it.
+        unsafe { LOG }
+    }
```
overlookmotel, line 158: Once again, I'm late to the party!

Do we need to resort to unsafe code for this? Could we use something like a `OnceLock` or `AtomicBool` instead?

I know that'd be a bit less efficient, but I imagine it'd be a drop in the ocean in the context of the whole codegen process, and performance is not critical in AST tools anyway.

overlookmotel:
Side question: Does AST tools exit with a non-zero error-code if something fails? I guess it should for [`ast_changes` CI job](https://github.com/oxc-project/oxc/blob/dc924892ccd6a6297ebb02e9e9a9a538a910df0f/.github/workflows/ci.yml#L235-L258).

tasks/ast_tools/src/main.rs
rzvxa: You are correct, I'll submit a PR for it.

rzvxa:
> Side question: Does AST tools exit with a non-zero error code if something fails? I guess it should for [`ast_changes` CI job](https://github.com/oxc-project/oxc/blob/dc924892ccd6a6297ebb02e9e9a9a538a910df0f/.github/workflows/ci.yml#L235-L258).

Yes, If it doesn't then it is a bug. main returns a result for this exact reason.

tasks/ast_tools/src/main.rs
rzvxa: #5574 will fix this

overlookmotel:
OK cool. Just checking!

Issue #4891, opened 2024-08-14
oxlint -v return 'dev' instead of '0.7.1' 


GiladShoham:
When running `oxlint -V` it returns `dev` instead of `0.7.1`

rzvxa:
It was caused by a mistake in our CI, @Boshen fixed it in a direct push(no PR so I can't link)
https://github.com/oxc-project/oxc/commit/0813af9d3053118d4b2d0985dda014df4c063896

It should report the correct version in our upcoming release.

PR #5531, opened 2024-09-06
ci: Passthrough OXC_VERSION and JEMALLOC_SYS_WITH_LG_PAGE to build container

valeneiko:
Fixes:
- https://github.com/oxc-project/oxc/issues/4891

Currently only Mac OS and Windows versions return correct version. Linux versions still returns `dev`. This is caused by Mac OS and Windows building on the host ([logs](https://github.com/oxc-project/oxc/actions/runs/10667173657/job/29564299213#step:6:16)), while Linux is building inside a container. Container builds do not have access to host env variables by default.

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,2 +1,6 @@
 [build.env]
-passthrough = ["CI"] # for crates/oxc_traverse/build.rs
+passthrough = [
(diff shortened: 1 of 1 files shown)


Boshen:
Thank you, nice catch.

PR #5535, opened 2024-09-06
feat(linter/eslint): implement no-alert

edwinlzs:
Contributing to #479 !

Rule Details: [link](https://eslint.org/docs/latest/rules/no-alert)

Am new to the internals of parsers and linters so I kept the implementation similar to ESLint's source code

Commented out the `globalThis` test cases in the pass array meant for pre-ES2020 before `globalThis` was introduced, having clarified in the Discord that it's ok to treat the built-in `globalThis` as always present.


DonIsaac (approved):
Thanks for contributing to Oxc! This PR looks good to me, let's get it in.

PR #5547, opened 2024-09-06
refactor(ast): inline `AstKind::as_*` methods

overlookmotel:
One would hope compiler will inline these trivial methods anyway, but mark them `#[inline]` to make sure.


overlookmotel:
Very little difference on benchmarks (I believe the +2% on one semantic benchmark is spurious - those benchmarks have a lot of variance). But I think a good idea regardless, especially for external consumers who don't enable LTO.

PR #5548, opened 2024-09-06
fix(linter/tree-shaking): align JSXMemberExpression's report

mysteryven:
Follow up #5450.

[eslint-plugin-tree-shaking](https://github.com/lukastaegert/eslint-plugin-tree-shaking/blob/463fa1f0bef7caa2b231a38b9c3557051f506c92/src/rules/no-side-effects-in-initialization.ts#L673-L677) just report on `JSXMemberExpression.property`

CleanShot 2024-09-06 at 22 34 55@2x


overlookmotel:
Thanks very much for following up on this.

mysteryven:
> Thanks very much for following up on this.

Any issue on this plugin, just @me!

PR #5368, opened 2024-08-31
feat(transformer): support transforming static import and export to commonjs

7086cmd:
Related issue
* #4050

@DonIsaac, I believe this PR cannot close the issue as it lacks support for `interop` and `esModule`. I have an idea to support these features, but it may contain excessive code for review.

This PR only transforms imports and exports, and it is not yet ready for production. Additionally, the tests have not been passed.

PR #5566, opened 2024-09-07
feat(transformer/commonjs): support dynamic import and `esModule` flag.

7086cmd:
Based on the issue #5368, since the size of the PR is too large, I opt to implement additional features using other PRs (including this) or align the behavior with Babel.


7086cmd:
stale

PR #5568, opened 2024-09-07
feat(oxc_cfg): add implicit return instruction

IWANABETHATGUY:
1. Adding a new InstructionKind `ImplicitReturn`
2. Adding a new `ImplicitReturn` instruction for each function body, just like biome
image

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -10,7 +10,7 @@ use rustc_hash::FxHashMap;
 
 use oxc_ast::{ast::*, AstKind, Visit};
 use oxc_cfg::{
-    ControlFlowGraphBuilder, CtxCursor, CtxFlags, EdgeType, ErrorEdgeKind,
+    ControlFlowGraphBuilder, CtxCursor, CtxFlags, EdgeType, ErrorEdgeKind, InstructionKind,
     IterationInstructionKind, ReturnInstructionKind,
 };
 use oxc_diagnostics::OxcDiagnostic;
@@ -1286,7 +1286,7 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
 
         /* cfg */
         control_flow!(self, |cfg| {
-            cfg.push_return(ret_kind, node_id);
+            cfg.push_return(ret_kind, Some(node_id));
             cfg.append_unreachable();
         });
         /* cfg */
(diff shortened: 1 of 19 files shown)


crates/oxc_semantic/src/builder.rs
```diff
@@ -1633,6 +1633,15 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
             self.visit_function_body(body);
         }
 
+        control_flow!(self, |cfg| {
```
rzvxa, line 1636: Let's merge these 2 `control_flow!` calls.

crates/oxc_semantic/src/builder.rs
```diff
@@ -1633,6 +1633,15 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
...
+            let c = cfg.current_basic_block();
+            if !matches!(
```
rzvxa, line 1638: Shall we use `if cfg.current_basic_block().unreachable` to detect more complex cases where somewhere up the graph we have returned?

Example:

```js
function f() {
  try {
    return g(x);
  } catch {
    return f();
  }
}
```

<details>
<summary>Before:</summary>
image
</details>

<details>
<summary>After:</summary>
image
</details>

Biome has the redundant return block but then again they always add a return block at the end, If I'm not mistaken we want to have a return and don't care for a dedicated block for it(as long as we encounter a return instruction in our path to the end of the function).

Please let me know what you think.

rzvxa (dismissed):
We should also do the same thing for the `visit_arrow_function_expression` method.

crates/oxc_semantic/src/builder.rs
IWANABETHATGUY: > Shall we use if cfg.current_basic_block().unreachable to detect more complex cases where somewhere up the graph we have returned?

I don't think we need to add a `unreachable` after a `implicit return`, because a `implicit return` must be last instruction of a function other than an `explicit return`

IWANABETHATGUY: what about  rename `ImplicitReturn` to `DummyReturn`  and don't insert a extra `unreachable`,

Boshen:
Waiting for @rzvxa to take a look.

crates/oxc_semantic/src/builder.rs
rzvxa: Now that I'm thinking about this, Maybe we can remove the unreachable instruction altogether, It was kept from the old implementation but I've added this field later on to cache some parts of the analysis from the build phase.

PR #5571, opened 2024-09-07
feat(oxc_wasm):  add `controlFlowDotString` to `oxc_wasm` 

IWANABETHATGUY:


crates/oxc_wasm/src/lib.rs
```diff
@@ -40,6 +40,9 @@ pub struct Oxc {
     #[wasm_bindgen(readonly, skip_typescript)]
     pub ir: String,
 
+    #[wasm_bindgen(readonly, skip_typescript, js_name = "debugDot")]
+    pub debug_dot: String,
```
Boshen, line 44: What is `debug_dot`? It should be something like `control_flow_graph`.

IWANABETHATGUY: It comes from the function call name, https://github.com/oxc-project/oxc/pull/5571/files#diff-2c0c75d87e0ee51363fff8c9b90b94b8bc5305ff7a8acfb29d3c51cd599b92daR203, it is just means the cfg dot representation

IWANABETHATGUY:
Merged since this pr just added another field, and I fixed the naming issue ,we could polish in the future.

Issue #5567, opened 2024-09-07
rust: enable `clippy::ref_as_ptr`


Boshen:
https://github.com/oxc-project/oxc/blob/0ac420d6f91f82b756e1395cee3dfe06e9eb9d64/Cargo.toml#L67

PR #5577, opened 2024-09-07
refactor: enable clippy::ref_as_ptr 

daflyinbed:
#5567 
`clippy::borrow_as_ptr` is also enabled in `oxc_traverse/src/generated/walk.rs`.


Boshen:
Thank you!

PR #5580, opened 2024-09-07
test(linter): add a passing case to no_undef

### LINKED ISSUE #2093: eslint(no-undef) doesn’t understand TypeScript generics
Given the following TypeScript code, it throws an error:

```ts
⚠ eslint(no-undef): Disallow the use of undeclared variables
function resolve<T>(path: string): T {
  return { path } as T;            ─
}

help: 'T' is not defined.
```

I understand this may just be a conflict between enabled rules; if so, maybe just adding a note in doc about recommended TS settings?

---

Also as a similar-but-slightly-different error, what’s the recommended way to handle builtin libraries?

```ts
⚠ eslint(no-undef): Disallow the use of undeclared variables

const root = new URL("../", import.meta.url);
                 ───
help: 'URL' is not defined.
```

I understand throwing an error for “magically-loaded” globals like test runner `expect()` calls. But for built-in DOM libraries, is there a recommended setup to allow this?

---

For both, I realize it may just require documentation, but wanted to ask for my own understanding. Thanks for such an amazing project!

Boshen:
closes #2093

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_undef.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_undef.rs
@@ -157,6 +157,7 @@ fn test() {
         "class C { static { function a() {} a; } }",
         "class C { static { a; function a() {} } }",
         "String;Array;Boolean;",
+        "function resolve<T>(path: string): T { return { path } as T; }"
     ];
 
     let fail = vec![


Boshen:
* **#5580** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5581, opened 2024-09-07
fix: try perf no plusplus

IWANABETHATGUY:
try perf the no plus plus rules, and see if there is still has regression


IWANABETHATGUY:
Ok as I expected, so then

PR #5590, opened 2024-09-07
feat(transformer): `async-to-generator` plugin.

7086cmd:
Tests are still not passed. A lot need to do yet.


7086cmd:
It appears that the iteration of the ast tree incurs performance degradation. However, it necessitates the duplication of statements for replacement. Are there alternative approaches that could be considered?

Boshen:
@7086cmd Can you hold onto this and let us take a look first? We may end up implementing this differently with architectural changes.

7086cmd:
> @7086cmd Can you hold onto this and let us take a look first? We may end up implementing this differently with architectural changes.

Ok. It does need some refactor, especially when it comes some polyfills here, we'd better do as what babel does: from js text to ast, which is visually more vibrant and more convenient. Replacing to multiple statements is also a handful problem.

7086cmd:
For this PR, I would prefer what `esbuild` does: more visible and no need for refactor for multiple statements here: https://esbuild.github.io/try/#dAAwLjIzLjEALS10YXJnZXQ9Y2hyb21lNTAAYXN5bmMgZnVuY3Rpb24gYSgpIHsKICAgIGF3YWl0IDE7Cn0KCmFzeW5jIChkKSA9PiB7CiAgICBhd2FpdCAyOwogICAgYXdhaXQgZCgpCn0K.

overlookmotel:
I've not dug into this in full, but at first glance, it looks like this is using ESBuild's technique, rather than Babel's.

I can see why you've chosen to do that, but I'm afraid we need to follow [@babel/plugin-transform-async-to-generator](https://babel.dev/docs/babel-plugin-transform-async-to-generator) exactly.

There's a couple of reasons for this:

### 1. Runtime performance

ESBuild's technique is easier to implement, and maybe faster, but the code it generates will be slower at runtime.

Input:

```js
async function foo() {
  await bar();
}
```

Babel output:

```js
var foo = _asyncToGenerator(function*() {
  yield bar();
});
```

ESBuild output:

```js
function foo() {
  return __async(this, null, function*() {
    yield bar();
  });
}
```

Babel's version creates a single function `foo`. But ESBuild's version has the cost of generating a new function `function*()`  every time you call `foo`.

Basically, Babel's version is better.

### 2. Testing

Our testing infra depends on passing Babel's tests. Sometimes it's annoying that we have to produce exactly same output as Babel, even in cosmetic things like variable naming. But the advantage is that we don't have to write loads of tests ourselves. I don't know about you, but I don't particularly enjoy writing tests!

Once we have the transformer passing all Babel's tests, our next step will be run the transformer on all source files for the top 3000 npm packages, like [monitor-oxc](https://github.com/oxc-project/monitor-oxc) does. We'll compare Oxc's output to Babel's output transforming the same files. Bombarding our transformer with so many test cases will no doubt reveal many bugs, and once we fix them we can have extremely high confidence that the transformer is behaving correctly. But for this to work, we need Oxc's transformer to emulate Babel exactly.

Only once we've got all that passing, then we'll bootstrap ourselves, and we can start making Oxc produce *better* output than Babel. But we haven't got to that stage yet.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -0,0 +1,171 @@
...
+impl<'a> Traverse<'a> for AsyncToGenerator<'a> {
+    fn exit_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
+        if let Expression::AwaitExpression(await_expr) = expr {
+            // Do not transform top-level await
+            if ctx.ancestry.ancestors().any(|ancestor| {
+                matches!(
+                    ancestor,
+                    Ancestor::FunctionBody(_) | Ancestor::ArrowFunctionExpressionBody(_)
+                )
+            }) {
```
overlookmotel, line 82: This should be faster:

```rs
if !self.ctx.source_type.is_module()
    || ctx.ancestor_scopes().any(|scope_id| {
        let scope_flags = ctx.scopes().get_flags(scope_id);
        scope_flags.contains(ScopeFlags::Function)
    })
{
```

* Can skip this check if it's a script, as scripts don't support top level await.
* There are less scopes in scope chain than nodes in ancestors, so iterating over scopes should get the answer with less iterations.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -0,0 +1,171 @@
...
+                let yield_expression = YieldExpression {
+                    span: SPAN,
+                    delegate: false,
+                    argument: Some(await_expr.argument.clone_in(ctx.ast.allocator)),
```
overlookmotel, line 86: I'm guessing this line is where some of the poor performance is coming from. There's no need to clone.

```rs
argument: Some(ctx.ast.move_expression(&mut await_expr.argument)),
```

`move_expression` just takes the `Expression`, without cloning it.

tasks/transform_conformance/snapshots/babel.snap.md
```diff
@@ -1651,6 +1651,56 @@ x Output mismatch
...
 
+# babel-plugin-transform-async-to-generator (1/17)
+* assumption-ignoreFunctionLength-true/basic/input.mjs
+x Output mismatch
+
+* assumption-ignoreFunctionLength-true/export-default-function/input.mjs
+x Output mismatch
+
+* assumption-noNewArrows-false/basic/input.js
+x Output mismatch
```
7086cmd, line 1662: It is a option: `noNewArrows`, which is not included in this.

tasks/transform_conformance/snapshots/babel.snap.md
```diff
@@ -1651,6 +1651,56 @@ x Output mismatch
...
+
+* bluebird-coroutines/expression/input.js
+x Output mismatch
+
+* bluebird-coroutines/named-expression/input.js
+x Output mismatch
+
+* bluebird-coroutines/statement/input.js
+x Output mismatch
```
7086cmd, line 1677: Also, a option

tasks/transform_conformance/snapshots/babel.snap.md
```diff
@@ -1651,6 +1651,56 @@ x Output mismatch
...
+
+* regression/gh-6923/input.js
+x Output mismatch
+
+* regression/in-uncompiled-class-fields/input.js
+x Output mismatch
+
+* regression/regression-2765/input.js
+x Output mismatch
```
7086cmd, line 1701: I'll handle these these days.

tasks/transform_conformance/snapshots/babel_exec.snap.md
```diff
@@ -71,6 +74,17 @@ exec failed
...
 
+# babel-plugin-transform-async-to-generator (3/6)
+* regression/15978/exec.js
+exec failed
+
+* regression/8783/exec.js
+exec failed
+
+* regression/T6882/exec.js
+exec failed
```
7086cmd, line 85: Because babel helper is not founded.

tasks/transform_conformance/src/test_case.rs
```diff
@@ -298,7 +299,14 @@ impl TestCase for ConformanceTestCase {
                 String::default,
                 |output| {
                     // Get expected code by parsing the source text, so we can get the same code generated result.
-                    let ret = Parser::new(&allocator, &output, source_type).parse();
+                    let ret = Parser::new(&allocator, &output, source_type)
+                        .with_options(ParseOptions {
+                            // Related: async to generator, regression
+                            allow_return_outside_function: true,
```
7086cmd, line 305: Some requires `return` on top.

7086cmd (commented):
some descs.

overlookmotel:
Wow! The recent changes you've made took this from a -70% perf regression to only -1%/-2% (and of course we should expect some perf hit from adding a new transform - nothing comes for free).

7086cmd:
Hello @overlookmotel and @Dunqing,

It appears that in babel, these functions share a common `wrap_function` helper. Given the complexity of the scenarios, I propose that I address several common situations in this PR and handle the remaining cases in _subsequent_ PRs, and this approach ensures that the `async_to_generator` and `async_generator_function` share a unified `wrap_function` helper.

Dunqing:
Thanks for your work! I am going to merge this PR and then complete this plugin later

7086cmd:
> Thanks for your work! I am going to merge this PR and then complete this plugin later

Thanks! Just a heads-up: this PR is from my personal forked repository, so we can't merge it using Graphite. If that's inconvenient, I'm willing to reopen it from the `origin` repository.

Dunqing:
> Thanks! Just a heads-up: this PR is from my personal forked repository, so we can't merge it using Graphite. If that's inconvenient, I'm willing to reopen it from the `origin` repository.

It doesn't matter, I have been rebasing the main branch

Dunqing (approved):
LGTM

PR #5615, opened 2024-09-08
fix(linter): `react/no_set_state` + `react/no_string_refs` rules find correct parent

overlookmotel:
`get_parent_es6_component` was finding *any* binding which is inside a class component, rather than parent of current node, leading to false positives.

The added test cases were not behaving correctly previously.


overlookmotel:
* **#5615** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5643, opened 2024-09-09
perf(linter): use `cow_replace` instead of `replace`

shulaoda:
Related to #5586


crates/oxc_linter/src/rules/eslint/no_loss_of_precision.rs
```diff
@@ -184,7 +185,7 @@ impl<'a> RawNum<'a> {
 impl NoLossOfPrecision {
     fn get_raw<'a>(node: &'a NumericLiteral) -> Cow<'a, str> {
         if node.raw.contains('_') {
```
Boshen, line 187: I think we no longer need to do this kind of checks.

Issue #5601, opened 2024-09-08
perf(ast): reduce AST memory usage


Boshen:
```
cargo binstall top-type-sizes
RUSTFLAGS=-Zprint-type-sizes cargo +nightly build -j 1 -p oxc > type-sizes.txt
top-type-sizes -s -w  -f oxc < type-sizes.txt
```

Structs / enums larget than 64 bytes should be reduced:

<details>

```
168 oxc_regular_expression::PatternParser<'_> align=8
     96 reader
     40 state
     16 source_text
      8 allocator
      4 span_factory

160 oxc_ast::ast::Class<'_> align=8
     32 decorators align=8
     32 id
     32 implements
     16 super_class
      8 span align=4
      8 type_parameters
      8 super_type_parameters
      8 body
      4 scope_id align=4
      1 type
      1 abstract
      1 declare

152 oxc_diagnostics::OxcDiagnosticInner align=8
     48 code
     24 message
     24 labels
     24 help
     24 url
      1 severity

152 oxc_ast::ast::ExportNamedDeclaration<'_> align=8
     64 with_clause align=8
     32 specifiers
     24 source
     16 declaration
      8 span
      1 export_kind

144 oxc_ast::ast::ExportAllDeclaration<'_> align=8
     64 with_clause
     40 exported
     24 source
      8 span
      1 export_kind

136 oxc_ast::ast::ImportDeclaration<'_> align=8
     64 with_clause
     32 specifiers
     24 source
      8 span
      1 import_kind

120 oxc_ast::ast::TSImportType<'_> align=8
     64 attributes
     16 parameter align=8
     16 qualifier
      8 span
      8 type_parameters
      1 is_type_of

112 std::result::Result<oxc_ast::ast::Program<'_>, oxc_diagnostics::OxcDiagnostic> align=8
    112 variant Ok
      8 variant Err

112 std::array::IntoIter<oxc_diagnostics::LabeledSpan, 2> align=8
     96 data
     16 alive

112 oxc_ast::ast::Program<'_> align=8
     32 directives
     32 body
     24 hashbang align=8
      8 span
      4 scope_id
      3 source_type

104 oxc_ast::ast::TaggedTemplateExpression<'_> align=8
     72 quasi
     16 tag
      8 span
      8 type_parameters

104 oxc_ast::ast::PropertyDefinition<'_> align=8
     32 decorators align=8
     16 key
     16 value
      8 span align=4
      8 type_annotation align=8
      1 type
      1 computed
[...]

PR #5645, opened 2024-09-09
perf(oxc_ast): reduce ast memory usage

IWANABETHATGUY:
part of #5601

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -597,7 +597,7 @@ pub struct PrivateFieldExpression<'a> {
 pub struct CallExpression<'a> {
     #[serde(flatten)]
     pub span: Span,
-    pub callee: Expression<'a>,
+    pub callee: Box<'a, Expression<'a>>,
     pub type_parameters: Option<Box<'a, TSTypeParameterInstantiation<'a>>>,
     pub arguments: Vec<'a, Argument<'a>>,
     pub optional: bool, // for optional chaining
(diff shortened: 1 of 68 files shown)


crates/oxc_ast/src/ast/js.rs
```diff
@@ -597,7 +597,7 @@ pub struct PrivateFieldExpression<'a> {
 pub struct CallExpression<'a> {
     #[serde(flatten)]
     pub span: Span,
-    pub callee: Expression<'a>,
+    pub callee: Box<'a, Expression<'a>>,
```
overlookmotel, line 600: This change doesn't make sense to me. It introduces indirection to save 8 bytes. And it's not really saving any bytes anyway as the `Expression` is still stored in arena, just in a separate allocation - in fact overall it *costs* an extra 8 bytes for the additional pointer.

overlookmotel (commented):
I'm afraid these changes so far don't make sense to me. The indirection of `Box` has a cost, so generally speaking should be avoided unless necessary.

In my opinion, the fields which we *should* consider boxing are either:

1. Large and wrapped in a `Option`, which is often `None`.
2. `Vec`s which are commonly empty.

e.g.:

* `Class::decorators` (most classes aren't decorated, so this `Vec` is usually empty).
* `FormalParameter::decorators` (ditto)
* `FunctionBody::directives` (most functions don't have any directives)
* `ArrayAssignmentTarget::rest` (`[a, b] = arr` is more common than `[a, ...b] = arr`, so `rest` is usually `None`)

In particular, I imagine that the majority of code that Oxc will run against will be plain JS (when bundling an application, the volume of code in `node_modules` will likely dwarf the application code written in TS). Therefore I think we should optimize the AST to be as memory-efficient as possible for JS, by boxing large TS-only fields.

Side note: We can expect `Vec` to be reduced to 16 bytes in future, which will reduce the size of a lot of types. https://github.com/oxc-project/backlog/issues/18

PR #5646, opened 2024-09-09
perf(formatter): use cow and put cow_ascii on clippy

### LINKED ISSUE #5639: perf(formatter): can `cow_to_ascii_lowercase` be used here?
[https://github.com/oxc-project/oxc/blob/main/crates/oxc_prettier/src/format/mod.rs#L1406-L1409](https://github.com/oxc-project/oxc/blob/main/crates/oxc_prettier/src/format/mod.rs#L1406-L1409)

related examples: #5637 

I tried it and it seems to work. Should this be done?

heygsc:
close: #5639 

part of: #5586

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,6 +1,6 @@
 ignore-interior-mutability = ["oxc_linter::rule::RuleWithSeverity"]
 
 disallowed-methods = [
-  "str::to_ascii_uppercase",
-  "str::to_ascii_lowercase",
+  { path = "str::to_ascii_lowercase", reason = "avoid a memory allocation from `to_ascii_lowercase`" },
+  { path = "str::to_ascii_uppercase", reason = "avoid a memory allocation from `to_ascii_uppercase`" },
 ]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,6 @@
 ignore-interior-mutability = ["oxc_linter::rule::RuleWithSeverity"]
+
+disallowed-methods = [
+  "str::to_ascii_uppercase",
+  "str::to_ascii_lowercase",
+]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,6 +1,6 @@
 ignore-interior-mutability = ["oxc_linter::rule::RuleWithSeverity"]
 
 disallowed-methods = [
-  { path = "str::to_ascii_lowercase", reason = "avoid a memory allocation from `to_ascii_lowercase`" },
-  { path = "str::to_ascii_uppercase", reason = "avoid a memory allocation from `to_ascii_uppercase`" },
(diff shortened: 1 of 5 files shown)


crates/oxc_prettier/src/format/mod.rs
```diff
@@ -1341,7 +1342,7 @@ impl<'a> Format<'a> for NumericLiteral<'a> {
             let mut string = Cow::Borrowed(raw);
 
             if string.contains(|c: char| c.is_ascii_uppercase()) {
```
Boshen, line 1344: Why was this `is_ascii_uppercase` but the code is `cow_to_ascii_lowercase` 🤣 

But these `contains` check can be removed.

heygsc: remove all or remove this?

crates/oxc_prettier/src/format/mod.rs
```diff
@@ -1405,7 +1402,7 @@ impl<'a> Format<'a> for BigIntLiteral<'a> {
         let text = self.span.source_text(p.source_text);
         // Perf: avoid a memory allocation from `to_ascii_lowercase`.
         if text.contains(|c: char| c.is_lowercase()) {
```
Boshen, line 1404: Same here?

.clippy.toml
```diff
@@ -1 +1,6 @@
 ignore-interior-mutability = ["oxc_linter::rule::RuleWithSeverity"]
+
+disallowed-methods = [
+  { path = "str::to_ascii_lowercase", reason = "avoid a memory allocation from `to_ascii_lowercase`" },
```
Boshen, line 4: Avoid memory allocation. Use `cow_utils::CowUtils::cow_to_ascii_uppercase` instead.

crates/oxc_prettier/src/format/mod.rs
heygsc: Then the formatter doesn't need cow at the moment🤣

heygsc: `if contain` condition is deleted, will the contents still be retained? 

This `fn format` seems there's nothing left.

crates/oxc_prettier/src/format/mod.rs
```diff
@@ -1340,10 +1340,6 @@ impl<'a> Format<'a> for NumericLiteral<'a> {
             let raw = self.span.source_text(p.source_text);
             let mut string = Cow::Borrowed(raw);
 
-            if string.contains(|c: char| c.is_ascii_uppercase()) {
-                string = Cow::Owned(string.to_ascii_lowercase());
```
Boshen, line 1344: oh ... this should be changed to `string = string.to_ascii_lowercase()`?

heygsc: Yes. I misunderstood "remove contains check".

PR #5649, opened 2024-09-09
refactor(linter): use `std::ptr::eq`

overlookmotel:
Follow-on after #5577.

`!std::ptr::eq(x, y)` is more idiomatic than `std::ptr::from_ref(x) != std::ptr::from_ref(y)`.

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_useless_switch_case.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_useless_switch_case.rs
@@ -66,7 +66,7 @@ impl Rule for NoUselessSwitchCase {
         let default_case = default_cases[0];
 
         // Check if the `default` case is the last case
-        if std::ptr::from_ref(default_case) != std::ptr::from_ref(cases.last().unwrap()) {
+        if !std::ptr::eq(default_case, cases.last().unwrap()) {
             return;
         }
 
--- a/crates/oxc_linter/src/rules/unicorn/prefer_event_target.rs
+++ b/crates/oxc_linter/src/rules/unicorn/prefer_event_target.rs
@@ -63,7 +63,7 @@ impl Rule for PreferEventTarget {
                     return;
                 };
 
-                if std::ptr::from_ref(ident) != std::ptr::addr_of!(**callee_ident) {
+                if !std::ptr::eq(ident, callee_ident.as_ref()) {
                     return;
                 }
             }
--- a/crates/oxc_linter/src/rules/unicorn/prefer_regexp_test.rs
+++ b/crates/oxc_linter/src/rules/unicorn/prefer_regexp_test.rs
@@ -87,7 +87,7 @@ impl Rule for PreferRegexpTest {
                 };
 
                 // Check if the `test` of the for statement is the same node as the call expression.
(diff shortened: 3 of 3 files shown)


overlookmotel:
* **#5649** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5649?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/rules/unicorn/no_useless_switch_case.rs
```diff
@@ -66,7 +66,7 @@ impl Rule for NoUselessSwitchCase {
         let default_case = default_cases[0];
 
         // Check if the `default` case is the last case
-        if std::ptr::from_ref(default_case) != std::ptr::from_ref(cases.last().unwrap()) {
+        if !std::ptr::eq(default_case, cases.last().unwrap()) {
```
camc314, line 69: not sure why i didn't write this originally, but i htink comparing spans is much clearer than checking if the pointers point to the same thing

overlookmotel: In general we shouldn't use `Span`s as unique identifiers. In the linter, it's probably OK, as the AST has come straight from the parser. But if for some reason we wanted to lint code that's come out of the transformer, `Span`s can no longer be trusted to be unique. e.g. `x ?? y` is transformed to `x !== null && x !== void 0 ? x : y` (and all those `x`s have the same `Span`).

Comparing `Span`s is also more error-prone than comparing pointers, as `ptr::eq` will error if the 2 pointers don't have the same type, whereas any 2 `Span`s can be compared.

So personally I'm not against comparing pointers, it's purely the verbose syntax `std::ptr::from_ref(x) != std::ptr::from_ref(y)` that I'm not keen on.

Boshen: Should these be replaced with `ContentEq`? I did similar stuff just now https://github.com/oxc-project/oxc/pull/5648/files

camc314: i'm not really against either, but from a developer point of view, i think comparing spans is a little more user friendly?
to be clear, don't let this block merging this

camc314 (approved):

crates/oxc_linter/src/rules/unicorn/no_useless_switch_case.rs
Boshen: `NoDuplicateCase` has to compare whether expressions are exactly the same.

> to be clear, don't let this block merging this. 🙂

Agree.

camc314: does `NoDuplicateCase` have different spans?

camc314: i implemented https://github.com/oxc-project/oxc/blob/59abf27d9563165883491e7ccaa89f0d78433559/crates/oxc_linter/src/utils/unicorn.rs#L150 a while ago to allow for catching stuff like is `foo as bar` the same as `foo`. in mose cases we would want to use that?

Issue #3483, opened 2024-05-31
`AstBuilder::copy` is unsound


overlookmotel:
https://github.com/oxc-project/oxc/blob/7f7b5ea9e83f37ff666d6f33fda7a9e27ea2aa07/crates/oxc_ast/src/ast_builder.rs#L71-L79

This method *can* be used correctly when a node is being removed from AST anyway, but there are multiple ways in which it can lead to various unsound situations:

Aliasing violation:

```rs
fn do_bad_stuff(stmt: &Statement, ast: &AstBuilder) {
  if Statement::BlockStatement(block) = stmt {
    let box1 = ast.copy(block);
    let box2 = ast.copy(block);
    // We now have two `Box<BlockStatement>` pointing to same `BlockStatement`
  }
}
```

Mutation through immutable ref:

```rs
fn do_bad_stuff2(stmt: &Statement, ast: &AstBuilder) {
  if Statement::BlockStatement(block) = stmt {
    let block = ast.copy(block);
    // We just mutated via an immutable ref
    block.body.clear();
  }
}
```

The double-free which #3481 fixes was likely caused by using `copy`.

We could make the method `unsafe` and caller will have to ensure it's only used when safe to do so. However, personally I think we should remove the API entirely because (1) it is difficult to reason about when it's safe to use and is therefore a footgun and (2) the performance hit from using safe alternatives is very minimal.

Unfortunately the alternatives are likely `take` or `mem::replace` which are not so ergonomic.

Boshen:
@Dunqing Please don't use this API from now on 😁 .

Use `move_xxx` APIs instead.

overlookmotel:
I don't think it's necessarily high priority to fix this. I just wanted to request that we don't introduce *more* uses of it now, as it will make it more work to remove it later.

rzvxa:
@overlookmotel Can we use the `CloneIn` trait over this?

overlookmotel:
I think `CloneIn` is something different. Usually `ast.copy` is used when we want to extract a node from AST, and insert it somewhere else. Cloning would be very expensive if the node is the tip of a deep tree, and isn't necessary.

rzvxa:
> I think `CloneIn` is something different. Usually `ast.copy` is used when we want to extract a node from AST, and insert it somewhere else. Cloning would be very expensive if the node is the tip of a deep tree, and isn't necessary.

[...]

PR #5652, opened 2024-09-09
refactor(traverse): revert changes to `walk.rs`

overlookmotel:
Revert the changes to `oxc_traverse` made in #5577.

Those changes may well have been good, but...

The soundness of `Traverse` is a finely balanced thing, and the pointer gymnastics are subtle, so I think it's best to be extremely conservative about changes. I last ran Miri on it with the original formulation (`as *mut _` etc) and it passed. Currently it's not possible to usefully run Miri on the transformer because it contains known unsound code (#3483). So until we're able to check soundness with Miri, I think it's best to avoid changes, as it'd be easy to trigger UB unintentionally.

### DIFF
--- a/crates/oxc_traverse/scripts/lib/walk.mjs
+++ b/crates/oxc_traverse/scripts/lib/walk.mjs
@@ -18,6 +18,9 @@ export default function generateWalkFunctionsCode(types) {
       clippy::missing_panics_doc,
       clippy::undocumented_unsafe_blocks,
       clippy::semicolon_if_nothing_returned,
+      clippy::ptr_as_ptr,
+      clippy::ref_as_ptr,
+      clippy::borrow_as_ptr,
       clippy::cast_ptr_alignment
     )]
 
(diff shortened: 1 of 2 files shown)


Boshen:
Wait ... look at this!

https://doc.rust-lang.org/1.81.0/src/core/ptr/mod.rs.html#792-804

```
/// Convert a mutable reference to a raw pointer.
///
/// This is equivalent to `r as *mut T`, but is a bit safer since it will never silently change
/// type or mutability, in particular if the code is refactored.
#[inline(always)]
#[must_use]
#[stable(feature = "ptr_from_ref", since = "1.76.0")]
#[rustc_const_stable(feature = "ptr_from_ref", since = "1.76.0")]
#[rustc_allow_const_fn_unstable(const_mut_refs)]
#[rustc_never_returns_null_ptr]
pub const fn from_mut<T: ?Sized>(r: &mut T) -> *mut T {
    r
}
```

> This is equivalent to r as *mut T, but is a bit safer since it will never silently change type or mutability, in particular if the code is refactored.

Are there misinformation or missing context?

overlookmotel:
I totally agree `Traverse` should use `ptr::from_mut`, `ptr::cast` and `ptr::addr_of` etc in place of the existing `as` casting. As you say, according to the docs, it'd be a definite improvement.

It's just that I would prefer to be ultra-conservative about any changes to `Traverse` until we can run Miri on the transformer. If (as I fear) Miri *does* find problems, it will be easier to figure them out and fix them if little has changed since I last ran Miri on it.

Background is that when I originally wrote `Traverse`, it took me about a week of fiddling with tiny changes before I could get Miri to give it a pass. And when I was reading about the "rules" of what's UB and what's not, it was surprising how often you get to a point where even the experts say "I can't say if that's sound or not. It depends on what we decide the memory model is. Don't know.". There are subtleties beyond what the docs say, and I certainly don't understand it all.

So, after that experience, I'm pretty wary, and consider making *any* changes without Miri to check it kind of "flying blind".

On the positive side, there's only a handful of `AstBuilder::copy` calls left in the transformer, and I'd hope to have removed them all within next few weeks as I go though doing a "second pass" over transformer. So my preference would be to delay making these changes to `Traverse` until that's complete and we can get Miri running again.

Does that sound OK to you?

PR #5656, opened 2024-09-09
fix(transformer/react): refresh plugin has incorrect reference flags

Dunqing:
As you see the diff


Boshen:
Let's merge and review fix later.

PR #5661, opened 2024-09-10
fix(regular_expression): Report more MayContainStrings error in (nested)class

### LINKED ISSUE #5632: regular_expression: Uncaught `MayContainStrings` in `/[^[\q{abc}]]/v`
Should be reported as error, but it's not.

image

Ultimaterly, `[^[[[[[[[[\q{abc}]]]]]]]]]` should be reported too.

leaysgur:
Fixes #5632

### DIFF
--- a/crates/oxc_regular_expression/src/ast.rs
+++ b/crates/oxc_regular_expression/src/ast.rs
@@ -238,6 +238,7 @@ pub struct CharacterClass<'a> {
     pub span: Span,
     pub negative: bool,
     pub kind: CharacterClassContentsKind,
+    pub strings: bool,
     pub body: Vec<'a, CharacterClassContents<'a>>,
 }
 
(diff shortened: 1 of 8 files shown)


crates/oxc_regular_expression/src/ast.rs
```diff
@@ -238,6 +238,7 @@ pub struct CharacterClass<'a> {
     pub span: Span,
     pub negative: bool,
     pub kind: CharacterClassContentsKind,
+    pub strings: bool,
```
Boshen, line 241: We probably need some documentation around these fields, I don't know what they mean 😢

leaysgur: OK, I'll review the AST documentation in another PR. 🏃🏻

PR #5662, opened 2024-09-10
refactor(linter): forbid string allocation methods such as `to_lowercase` and `replace`

Boshen:
closes #5586

Feel free to use the PR as the base and make changes.

PR #5664, opened 2024-09-10
perf(rust): use `cow_utils` instead

shulaoda:
Related to #5586 and #5662


Boshen:
Thank you! Nice work!

PR #5674, opened 2024-09-10
refactor(semantic): simplify getting `AstNodeId`s for CFG

overlookmotel:
Remove `ast_node_records`, and the functions that push and pop it, from `SemanticBuilder`.

The purpose of that apparatus is to do something quite simple - get the `AstNodeId` of an `Expression` from its parent, for the control flow graph (CFG). As nodes are created in order, a simpler way to do this is to get what the next `AstNodeId` will be *before* visiting the `Expression`.

This implementation has to work around a current shortcoming with `IndexVec` - converting `Nodes::len()` to an `AstNodeId` involves a panicking bounds-check, which adds overhead even when CFG is disabled. So for now `get_next_node_id` returns the "id" as a `u32`, and we convert it to `AstNodeId` only after the node has been created, which guarantees that the `u32` is a valid `AstNodeId` (not out of bounds), because `IndexVec::push` also performs that same bounds check.

If we implement https://github.com/oxc-project/backlog/issues/109, `get_next_node_index` could become `get_next_node_id` and return an `AstNodeId`, and the unsafe code could be removed.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -106,8 +106,6 @@ pub struct SemanticBuilder<'a> {
     pub(crate) cfg: Option<ControlFlowGraphBuilder<'a>>,
 
     pub(crate) class_table_builder: ClassTableBuilder,
-
-    ast_node_records: Vec<NodeId>,
 }
 
 /// Data returned by [`SemanticBuilder::build`].
@@ -152,7 +150,6 @@ impl<'a> SemanticBuilder<'a> {
             check_syntax_error: false,
             cfg: None,
             class_table_builder: ClassTableBuilder::new(),
-            ast_node_records: Vec::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#5674** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5674?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This is showing on benchmarks as a 1% perf regression. This is surprising, as the code is doing less, not more. Possibly reducing the amount of code in some functions has caused compiler to make different decisions about inlining, and those decisions aren't as good as before. 🤷

The benchmarks have CFG disabled. I strongly suspect this would be a speed-up when CFG is enabled.

Boshen:
Close as stale.

overlookmotel:
Reopening just to check benchmarks of this rebased on latest main. I found it surprising originally that this did not have a positive effect.

overlookmotel:
Nope. Still -1% on all semantic benchamrks.

PR #5686, opened 2024-09-10
refactor(semantic): compare nodes by pointer equality

overlookmotel:
Follow up after #5673.

It's safer to compare nodes by pointer equality than by equality of `Span`s. Transformer can create multiple nodes of same type with same span.

Once nodes have node IDs, we can use them for comparison instead.

That said, `ptr::eq` is much cheaper, as it doesn't have to follow the pointers to check equality. https://godbolt.org/z/hsqaeKr5M

### DIFF
--- a/crates/oxc_semantic/src/binder.rs
+++ b/crates/oxc_semantic/src/binder.rs
@@ -1,6 +1,6 @@
 //! Declare symbol for `BindingIdentifier`s
 
-use std::borrow::Cow;
+use std::{borrow::Cow, ptr};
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#5686** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camc314 (approved):

PR #5692, opened 2024-09-11
fix(regular_expression): Keep LegacyOctalEscape raw digits for `to_string`

### LINKED ISSUE #5690: regular_expression: `to_string()` missmatch for LegacyOctalEscape
Found in https://github.com/oxc-project/oxc/pull/5676#discussion_r1751905218

- - -

Failed regexp is `/\0003/`.

This is parsed as:

- Character(Octal): `\000`
- Character(Symbol): `3`

But once `to_string`ed, this will be `/\03/`.

This is parsed as:

- Character(Octal): `\03`

Because we remove leading `0`s for now.

This also happens:

- `/\0111/` => `/\111/`
- `/\0022/` => `/\22/`

Is it possible to restore these without having extra `raw` field in the AST?

But both `\000` and `\00` are the same Character(Octal)...

leaysgur:
Fixes #5690 

- Update `CharacterKind` enum from `Octal` to `Octal1`, `Octal2` and `Octal3`
- Stylistic refactoring for `impl Display`


Boshen (approved):
NICE!

PR #5694, opened 2024-09-11
feat(prettier): add assignment/sequence

heygsc:
test: https://github.com/prettier/prettier/blob/main/tests/format/js/assignment/sequence.js

snap:https://github.com/prettier/prettier/blob/main/tests/format/js/assignment/__snapshots__/format.test.js.snap#L870-L888

code: https://github.com/prettier/prettier/blob/main/src/language-js/needs-parens.js#L790-L796


heygsc:
Not merged, it seems that fork is not supported...

PR #5703, opened 2024-09-11
perf(semantic): avoid counting nodes

overlookmotel:
Instead of visiting AST to count AST nodes in `SemanticBuilder`, just make high estimates of counts for nodes, scopes, symbols and references, based on source text length. These estimates are the maximum number of items theoretically possible for a given source text length. Almost always they'll be vast over-estimates, but should never be an under-estimate.

This ensures `AstNodes`, `ScopeTree` and `SymbolTable`'s allocations will not grow during their construction, which is a major performance hit, but avoids the complete AST traversal we were using previously to get accurate counts.

This "estimate high" approach is a viable strategy on 64-bit systems with virtual memory, where:

1. The memory space is so large that it's almost impossible to exhaust.
2. Allocating memory only consumes virtual memory, not physical memory.

So use this faster "estimate high" method as standard, and the slower visitor-based counter on 32-bit systems and WASM.

Note: ~~I tried using `shrink_to_fit` on the `Vec`s once semantic run is complete, to free up the excess memory. But this caused the `Vec`s to all reallocate, which negated the performance gains.~~ It was an artefact of the custom global allocator we use in our benchmarking setup. Shrinking allocations does not cause reallocation.

### DIFF
--- a/crates/oxc/src/compiler.rs
+++ b/crates/oxc/src/compiler.rs
@@ -156,12 +156,13 @@ pub trait CompilerInterface {
         /* Compress */
 
         if let Some(options) = self.compress_options() {
-            self.compress(&allocator, &mut program, options);
+            self.compress(&allocator, &mut program, source_text, options);
         }
 
         /* Mangler */
 
-        let mangler = self.mangle_options().map(|options| self.mangle(&mut program, options));
+        let mangler =
+            self.mangle_options().map(|options| self.mangle(&mut program, source_text, options));
 
         /* Codegen */
 
@@ -214,13 +215,19 @@ pub trait CompilerInterface {
         &self,
         allocator: &'a Allocator,
         program: &mut Program<'a>,
+        source_text: &str,
         options: CompressOptions,
     ) {
-        Compressor::new(allocator, options).build(program);
+        Compressor::new(allocator, options).build(program, source_text);
     }
 
-    fn mangle(&self, program: &mut Program<'_>, options: MangleOptions) -> Mangler {
-        Mangler::new().with_options(options).build(program)
+    fn mangle(
+        &self,
+        program: &mut Program<'_>,
+        source_text: &str,
+        options: MangleOptions,
+    ) -> Mangler {
+        Mangler::new().with_options(options).build(program, source_text)
     }
 
     fn codegen<'a>(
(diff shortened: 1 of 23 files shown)


overlookmotel:
* **#5703** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
TODO: Allow passing existing node counts into `SemanticBuilder` instead of basing counts on source text length.

This is necessary when re-running semantic on an AST which has been transformed/minified. Nodes / scopes / symbols / references counts may have increased in transformation, but the source text has not changed, so estimating based on source text length will be inaccurate. Additionally, the actual counts will be accurate, unlike estimating from source text length.

At present all test pass but this is an accident because the over-estimates are so large that they obscure this problem. It would be possible to produce panics with the right input.

overlookmotel:
As an experiment, I hard-coded the exact counts into `SemanticBuilder` for our benchmark fixtures, so it doesn't over-allocate.

This has little effect on most of semantic benchmarks, but it does noticeably speed things up for `semantic[RadixUIAdoptionSection.jsx]` (+4%).

So, while over-allocating is not *too* bad, it's not as good as accurate counts for small files.

The over-allocating version is *massively* over-allocating. For `RadixUIAdoptionSection.jsx`:

|            | Actual | Estimated |
|------------|--------|-----------|
| nodes      | 392    | 5041      |
| scopes     | 2      | 1261      |
| symbols    | 11     | 1259      |
| references | 28     | 1261      |

So likely the reason for the slowdown of over-allocating is that the much larger estimated counts push the allocations up a size class in the system allocator. I believe system allocator serves small allocations from a thread-local pool, rather than taking from the main cross-thread pool. The former is faster as it doesn't involve any synchronisation.

I feel that small files are the most representative of the code Oxc will be processing in the "real world", so the `RadixUIAdoptionSection.jsx` benchmark is probably the most important one.

Conclusion: Getting accurate counts in the parser will be a better solution if it doesn't cost much in the parser. According to @DavidHancu [on Discord](https://discord.com/channels/1079625926024900739/1281761615833993279) cost of compiling counts in parser is about -0.3%. That's pretty cheap.

Boshen:
Close as stale.

overlookmotel:
Archived on [overlookmotel/semantic-avoid-counting-nodes](https://github.com/oxc-project/oxc/tree/overlookmotel/semantic-avoid-counting-nodes) branch.

PR #5715, opened 2024-09-12
perf(ast)!: revert: reduce byte size of `TaggedTemplateExpression::quasi` by `Boxing` it (#5679)

overlookmotel:
As discussed in #5601, there is little benefit to reducing type sizes for the sake of it. In this case, the `quasi` field in not in an `Option`, so putting it in a `Box` does not save data, only moves where that data is stored, and introduces pointer-indirection.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -461,7 +461,7 @@ pub struct TaggedTemplateExpression<'a> {
     #[serde(flatten)]
     pub span: Span,
     pub tag: Expression<'a>,
-    pub quasi: Box<'a, TemplateLiteral<'a>>,
+    pub quasi: TemplateLiteral<'a>,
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#5715** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5715?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
I believe the benchmark changes to be noise. In particular the `transformer[cal.com.tsx]` benchmark change is clearly spurious - the negative all appears to be due to a hashmap reallocating at the start of transform.

Boshen:
> As discussed in #5601, there is little benefit to reducing type sizes for the sake of it. In this case, the `quasi` field in not in an `Option`, so putting it in a `Box` does not save data, only moves where that data is stored, and introduces pointer-indirection.

We need a "specification" of what to box and what not.

PR #5728, opened 2024-09-12
refactor(semantic): fix dead code warning in release mode

overlookmotel:
`Counts::assert_accurate` is only used in debug mode. Silence the dead code warning in release mode.


overlookmotel:
Cosmetic change, so merging without review.

PR #5729, opened 2024-09-12
feat(eslint/jsx_a11y): implement anchor_ambiguous_text

jelly:
There is a remaining clippy issue with passing `COW<_, str>` which I simply can't resolve, but maybe there is more to fix first in my approach.

Currently this PR might do a bit too much allocations:

* trim() - this sadly has no cowutils replacement but might not be needed. I kept the `trim()` in there as a "maybe optimisation" to skip the whole string modification shenanigans
* replace() can't be used so I used retain (which also is a bit faster iirc?), but that converts COW<String> => String. So maybe this should chain `cow_replace`?
* I'm also not sure if `if normalized_str.contains(char::is_whitespace) {` is worth it? But my gut feeling is that it is, as it deconstructs and constructs a whole String<>

### DIFF
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -379,6 +379,7 @@ mod unicorn {
 
 mod jsx_a11y {
     pub mod alt_text;
+    pub mod anchor_ambiguous_text;
     pub mod anchor_has_content;
     pub mod anchor_is_valid;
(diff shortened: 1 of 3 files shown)


Boshen:
@camc314 can you help and get this PR merged.

camc314 (approved):
thank you!

PR #5731, opened 2024-09-12
feat(oxc_cfg):  better control flow graph dot dot repr

IWANABETHATGUY:
1. add basic block index for each basicblock, so there should no basic block display empty string
2. use `shape` box, since it is basic **block**
## Before
image

## After
image


rzvxa (approved):
This is really nice😍

rzvxa:
On another kind of related note; It will be super cool if we add 2 checkboxes to the playground for adding some extra information to the graph and show/hide implicit error paths.

I sometimes use this example to get extra debug information while I'm tracking something. It's a bit noisy but it has helped me quite a few times and might be useful to others as well.

https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/examples/cfg.rs

IWANABETHATGUY:
> On another kind of related note; It will be super cool if we add 2 checkboxes to the playground for adding some extra information to the graph and show/hide implicit error paths.
> 
> I sometimes use this example to get extra debug information while I'm tracking something. It's a bit noisy but it has helped me quite a few times and might be useful to others as well.
> 
> https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/examples/cfg.rs

Sure, I will have a look.

PR #5733, opened 2024-09-12
feat(index)!: remove ability to index `IndexVec` with `usize`

overlookmotel:
`IndexVec` should only be indexed into with its associated `Idx` type. The implementation taken from `index_vec` crate however also includes the ability to index into the `Vec` with a `usize` "as an ergonomic concession, it's too painful without":

https://docs.rs/index_vec/0.1.4/src/index_vec/indexing.rs.html#112-134

I think we want to be stricter than this, so remove this ability.


overlookmotel:
* **#5733** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Issue #5204, opened 2024-08-25
Avoid multiple iterations over `nodes` and `symbols` in linter


overlookmotel:
Various linter rules are implemented as a `Rule::run_once` method which iterates over `nodes` or `symbols`.

Usually this is because they work in 2 passes. The first pass collects info from the AST/symbols and then the 2nd processes that info.

This results in `nodes` or `symbols` being unnecessarily iterated over multiple times, when `run` and `run_on_symbol` methods are already provided for doing this.

I imagine it'd be faster to:

* Add a `Rule::run_once_at_end` method which executes after `run` and `run_on_symbol` have executed on all nodes/symbols.
* Make `Rule::run` and `Rule::run_on_symbol` take a `&mut self` instead of `&self`.
* Reimplement these rules:
  * To store their intermediate data in `self`.
  * Do the main loop over `nodes` or `symbols` as a `run` / `run_on_symbol` method.
  * Do the 2nd pass in `run_once_at_end`.

It's iterating over `nodes` repeatedly which would likely have the largest perf impact. But there are so many rules which iterate over `symbols` in `run_once` that it might also be having a noticeable impact in aggregate.

## Which rules?

Rules which use `run_once` in this way are:

### Iterating over `nodes`

* `eslint/func_names`
* `eslint/no_this_before_super`
* `import/no_named_as_default_member`
* `jest/no_large_snapshots`
* `jest/prefer_hooks_in_order`
* `jsdoc/require_returns`
* `tree_shaking/no_side_effects_in_initialization` (which does a partial tree traversal)

### Iterating over `symbols`

* `eslint/no_shadow_restricted_name`
* `jest/consistent_test_it`
* `jest/expect_expect`
* `jest/max_expects`
* `jest/max_nested_describes`
* `jest/no_alias_methods`
* `jest/no_conditional_expect`
* `jest/no_confusing_set_timeout`
* `jest/no_disabled_tests`
* `jest/no_done_callback`
* Loads more jest rules + vitest rules via `collect_possible_jest_call_node`

@DonIsaac What do you think?

DonIsaac:
I like where this is going. I need to think on this for a bit before I give an answer.

Boshen:
This is not a priority, none of the rules are enabled by default.

overlookmotel:
#5201 was an example of fixing a rule which was unnecessarily iterating over symbols.

Boshen:
cc @mysteryven if you're intrested.

[...]

PR #5741, opened 2024-09-13
refactor(linter): move `run_once` to the end 

shulaoda:
Related to #5204

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -124,10 +124,6 @@ impl Linter {
             .map(|rule| (rule, self.ctx_for_rule(&ctx, rule)))
             .collect::<Vec<_>>();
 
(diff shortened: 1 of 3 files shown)


Boshen:
It feels like moving `run_once` to the end should make things work.

shulaoda:
> It feels like moving `run_once` to the end should make things work.

Good idea!

Boshen:
We can defer merging this until we figure out how to store rule states.

DonIsaac:
> We can defer merging this until we figure out how to store rule states.

How would you feel about a cache object for each rule stored in `LintContext`?

Boshen:
> > We can defer merging this until we figure out how to store rule states.
> 
> How would you feel about a cache object for each rule stored in `LintContext`?

Can give this a try.

PR #5745, opened 2024-09-13
fix(transformer/react): isStaticChildren should be false when there is only one child

Dunqing:
We found a warning report [here](https://github.com/facebook/react/blob/206df66e70652e85711c3177ce1a0459609a7771/packages/react/src/jsx/ReactJSXElement.js#L614-L632). It caused by we transform incorrectly

### DIFF
--- a/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/static-children/input.jsx
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/static-children/input.jsx
@@ -1,0 +1,7 @@
+<div>
+  { /*
+      There is only one child, so the isStaticChildren option should be false.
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#5745** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5745?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/static-children/input.jsx
```diff
@@ -0,0 +1 @@
+<div><div></div></div>
```
Boshen, line 1: Can you add some comments here? Otherwise this is a really plain confusing test.

Dunqing: Added

PR #5747, opened 2024-09-13
feat(minifier): add `MinimizeConditions` pass

Boshen:
I expect small performance regression.

But managed to improve the following case from react.developmement.js

```
oxc  main ❯ diff before.js after.js
670c670
< 		if (!(dispatcher !== null)) throw Error("Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:\n1. You might have mismatching versions of React and the renderer (such as React DOM)\n2. You might be breaking the Rules of Hooks\n3. You might have more than one copy of React in the same app\nSee https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.");
---
> 		if (dispatcher === null) throw Error("Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:\n1. You might have mismatching versions of React and the renderer (such as React DOM)\n2. You might be breaking the Rules of Hooks\n3. You might have more than one copy of React in the same app\nSee https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.");
```

### DIFF
--- a/crates/oxc_minifier/Cargo.toml
+++ b/crates/oxc_minifier/Cargo.toml
@@ -39,5 +39,6 @@ num-traits = { workspace = true }
 [dev-dependencies]
 oxc_parser = { workspace = true }
 
+cow-utils = { workspace = true }
 insta = { workspace = true }
 pico-args = { workspace = true }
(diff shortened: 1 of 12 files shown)


Boshen:
* **#5747** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5751, opened 2024-09-13
perf(transformer): call `transform_jsx` in `exit_expression` rather than `enter_expression`

Dunqing:
### Difference

In `enter_expression`: Recursive transform JSX
In `exit_expression`: Deep first transform

After the change,  `transform_jsx` still has a lot of room for improvement.


overlookmotel (approved):
I'll be honest, I don't quite understand how this improves perf - I think I don't get how JSX and arrow function transforms interact. But I'm not going to argue with 7% perf increase. Nice one!

PR #5760, opened 2024-09-13
refactor(syntax): replace `nonmax` crate with zero-cost non-max types

overlookmotel:
Replace `nonmax` crate with own `NonMax*` types.

`nonmax`'s `NonMax*` types work by storing the value internally as a `NonZero*`, and inverting the bits on every read/write (`native_u32 = nonmax_u32 as u32 ^ u32::MAX`). This reserves only a single illegal value, but has (small) runtime cost on every read/write.

The `NonMax*` types provided here make a different trade-off.

`NonMaxU8` wraps an enum with a single illegal value (255). All the other types store the value as a number of `u8`s + a single `NonMaxU8` as highest byte. i.e. they reserve all values which have 255 as the highest byte.

* `NonMaxU8` can represent the values `0` to `254`.
* `NonMaxU16` can represent the values `0` to `(255 << 8) - 1`.
* `NonMaxU32` can represent the values `0` to `(255 << 24) - 1`.
* `NonMaxU64` can represent the values `0` to `(255 << 56) - 1`.
* `NonMaxU128` can represent the values `0` to `(255 << 120) - 1`.

We trade approx 0.4% of the legal range being unavailable, in return for zero runtime cost when reading or writing the values.

Linter snapshot change is because `PreferHooksInOrder` rule uses a hash map keyed by `ScopeId`. The hash of `NonMaxU32` is now same as hash of `u32`, which alters the hash map iteration order. This does not change what errors are output, only the order they're output in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1308,12 +1308,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "nonmax"
-version = "0.5.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "610a5acd306ec67f907abe5567859a3c693fb9886eb1f012ab8f2a47bef3db51"
-
-[[package]]
 name = "nu-ansi-term"
 version = "0.46.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1530,7 +1524,6 @@ dependencies = [
  "cow-utils",
  "daachorse",
  "insta",
- "nonmax",
  "once_cell",
  "oxc_allocator",
  "oxc_ast",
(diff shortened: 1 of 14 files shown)


overlookmotel:
* **#5760** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5760?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
-3% on compressor benchmarks. Zero cost my arse!

Boshen:
Close as stale.

overlookmotel:
Archived on [overlookmotel/non-max-id-types](https://github.com/oxc-project/oxc/tree/overlookmotel/non-max-id-types) branch.

PR #5767, opened 2024-09-13
refactor(semantic)!: rename `AstNode` to `Node` and `AstNodes` to `Nodes`

overlookmotel:
#5740 renamed `AstNodeId` to `NodeId`. To match this, also rename:

* `AstNode` -> `Node`
* `AstNodes` -> `Nodes`

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -1,5 +1,5 @@
 use oxc_ast::{ast::BindingIdentifier, AstKind};
-use oxc_semantic::{AstNode, IsGlobalReference, NodeId, SymbolId};
+use oxc_semantic::{IsGlobalReference, Node, NodeId, SymbolId};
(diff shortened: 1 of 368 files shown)


overlookmotel:
* **#5767** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
@Boshen I think it makes sense for consistency, but is it a step too far?

overlookmotel:
I changed my mind. I don't think this is a good idea. We also have `AstKind` and `AstType`, which we don't want to rename to just `Kind` and `Type`, so `Ast*` is going to remain a thing.

We plan to remove the `AstNode` type anyway as part of #5689. Once that's done, we could rename `AstNodes` to `Nodes`.

PR #5770, opened 2024-09-14
feat(linter): support shared rule state in `ctx`

shulaoda:
Related to #5204

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1631,6 +1631,7 @@ dependencies = [
  "convert_case",
  "cow-utils",
  "dashmap 6.0.1",
+ "dyn-clone",
  "globset",
  "insta",
  "itertools",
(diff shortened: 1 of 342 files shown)


shulaoda:
How about this?  @Boshen 
I have tested replacing all rules in the local commit and didn't find any errors.

Boshen (commented):
:thinking: I'm not sure if this is going to work, can you continue prototyping by collecting state in `run`, and then inspect the state in `run_once` moved to the end? We probably need to implement this prototype with two rules.

I'd imagine the states need to be isolated, rule A cannot read data from rule B.

shulaoda:
Let me reconsider, it seems like it's not feasible and a bit difficult. 😩

DonIsaac:
I've got some ideas on how to implement this. I'll take a closer look tomorrow.

Issue #4753, opened 2024-08-08
transformer: helper library


Boshen:
It's either `@babel/helpers` or tslib ...

* https://www.npmjs.com/package/@babel/helpers
* https://www.npmjs.com/package/tslib

overlookmotel:
I vote Babel, since we're primarily modelling our transforms on Babel's implementations.

rzvxa:
I also think using Babel helpers is easier but it's quite chunky, Is there any possibility of forking it in the future so we can strip it down to its bare bones?

Boshen:
The next task is to figure out how to get our conformance suite to work with this ...

Boshen:
As it turns out we don't need to inline the helper functions for the test suites

https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-class-properties/test/fixtures/private/call/output.js

```javascript
var _foo = /*#__PURE__*/new WeakMap();
let Foo = /*#__PURE__*/function () {
  "use strict";

  function Foo() {
    babelHelpers.classCallCheck(this, Foo);
    babelHelpers.classPrivateFieldInitSpec(this, _foo, function () {
      return this;
    });
  }
  return babelHelpers.createClass(Foo, [{
    key: "test",
    value: function test(other) {
      var _other$obj;
      babelHelpers.classPrivateFieldGet2(_foo, this).call(this);
      babelHelpers.classPrivateFieldGet2(_foo, _other$obj = other.obj).call(_other$obj);
    }
  }]);
}();
```

The relevant functionality we need to port is https://github.com/babel/babel/blob/7edd5082152d6b89e90ae09a533eb725b15902ba/packages/babel-core/src/transformation/file/file.ts#L153-L200

overlookmotel:
OK great!

I guess we need to add a `helpers` struct to `TransformCtx` to centralize this logic.

Shall we put in a basic version for now which always returns a `babelHelpers.<helper_name>(...)` call, so we can implement plugins and get tests passing? Then we can make it work properly with inline helpers/imports later on. Something like:

```rs
struct Helpers<'a>(PhantomData<&'a ()>);

impl<'a> Helpers<'a> {
    fn call(helper_name: Atom<'a>, arguments: Vec<'a, Argument<'a>>, ctx: &mut TraverseCtx<'a>) -> CallExpression<'a>;
[...]

PR #5771, opened 2024-09-14
chore(transformer): a small `ast_helper` via `oxc_parser`.

7086cmd:
Related: https://github.com/oxc-project/backlog/issues/184.

I recently conceived the concept of a transitional phase. In my opinion, the current plan for `ast! { let Foo = ${a} }` syntax is overly complex for the time being. However, it appears that the Babel helper is somewhat indispensable for handling `commonjs` and `async/await`.

### DIFF
--- a/crates/oxc_transformer/src/helpers/ast_helper.rs
+++ b/crates/oxc_transformer/src/helpers/ast_helper.rs
@@ -1,0 +1,41 @@
+#[macro_export]
+macro_rules! ast_helper {
+    ($allocator: expr, $source: ident) => {{
+        let source_type = oxc_span::SourceType::mjs();
+        let ret = oxc_parser::Parser::new($allocator, &$source, source_type).parse();
+        if !ret.errors.is_empty() {
+            // Since the code is only used in `oxc_transformer` crate, we can use `panic!` here.
+            panic!("Parser Errors: {:?} in using ast_helper", ret.errors);
+        }
+        ret
+    }};
+}
+
+#[cfg(test)]
+mod tests {
+    use oxc_allocator::Allocator;
+
+    #[test]
+    fn test_ast_helper() {
+        let allocator = Allocator::default();
+        let source = format!("const a = {}", 0);
+        let ast = ast_helper!(&allocator, source);
+        assert_eq!(ast.program.body.len(), 1);
+    }
+    
+    #[test]
+    fn test_more_complex_ast_helper() {
+        let allocator = Allocator::default();
+        let key = "key";
+        let src = format!(r#"Object.defineProperty(exports, {key}, {{
+    enumerable: true,
+    writable: true,
+    value: function () {{
+        return 0;
+    }}
+}})
+"#);
+        let ast = ast_helper!(&allocator, src);
+        assert_eq!(ast.program.body.len(), 1);
+    }
+}
--- a/crates/oxc_transformer/src/helpers/ast_helper.rs
+++ b/crates/oxc_transformer/src/helpers/ast_helper.rs
@@ -22,19 +22,21 @@ mod tests {
         let ast = ast_helper!(&allocator, source);
         assert_eq!(ast.program.body.len(), 1);
     }
-    
+
     #[test]
     fn test_more_complex_ast_helper() {
         let allocator = Allocator::default();
         let key = "key";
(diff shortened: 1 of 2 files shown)


overlookmotel:
I'm sorry to say, I am not a fan of this approach. You are right that it is a challenge to create an `ast!` macro which does parsing at compile time. But parsing snippets at runtime is going to be unacceptably slow, in my opinion.

Also, various nodes need `ScopeId`s, `SymbolId`s and `ReferenceId`s correctly initialized, which the parser cannot do. So to get it right, and pass our transformer checker tests, you'd likely have to monkey about with the generated code to fix it all up, which negates the purpose of the `ast!` macro, which is to allow a less verbose code style.

7086cmd:
> 

I see. Let’s implement these tasks via the `ast!` prototype after we’ve implemented it :-)

For now, I think the macros should be used to generate some related functions, for "polyfill", which is definitly unable to use AstBuilder in the code (otherwise it may contains thousands of LoC only for a static string function. Suppose we have a function `__async` which is the function to transform the async function (from esbuild's code):

```js
var __async = (__this, __arguments, generator) => {
  return new Promise((resolve, reject) => {
    var fulfilled = (value) => {
      try {
        step(generator.next(value));
      } catch (e) {
        reject(e);
      }
    };
    var rejected = (value) => {
      try {
        step(generator.throw(value));
      } catch (e) {
        reject(e);
      }
    };
    var step = (x) => x.done ? resolve(x.value) : Promise.resolve(x.value).then(fulfilled, rejected);
    step((generator = generator.apply(__this, __arguments)).next());
  });
};
```

If we use AstBuilder, it contains about 12800 chars in the code, and I think these code can make the size much bigger. So let's say the priority of these tasks, and I will wait the day the `ast!` comes :-).

crates/oxc_transformer/src/helpers/ast_helper.rs
```diff
@@ -0,0 +1,43 @@
+#[macro_export]
+macro_rules! ast_helper {
+    ($allocator: expr, $source: ident) => {{
```
DonIsaac, line 3: does `$source` need to be an `ident`? couldn't it be a `'static str`?

crates/oxc_transformer/src/helpers/ast_helper.rs
```diff
@@ -0,0 +1,43 @@
+#[macro_export]
+macro_rules! ast_helper {
```
DonIsaac, line 2: I like this macro. That being said, `helpers` sets off some red flags for me. If we want to support AST-related macros, we should use domain-specific folders and macro names.

overlookmotel:
For the `__async` function you mention, we'd get that from a helper library (or publish our own helper library for any helper functions we need which are not available elsewhere). #4753 

Then the transform would add `import __async from 'helper-lib/async';` to the output. One advantage of this approach over inlining is that a bundler will be able to de-duplicate helper functions which are used in multiple files, and put the function in a shared chunk.

So the `ast!` macro is not intended for such large chunks of code.

It would be great if we had a proper `ast!` macro which parses at compile time, but as you note, it's not a trivial task. So unfortunately in the meantime we have to write by hand using `AstBuilder` calls, annoying though it is.

Personally, I am opposed to merging a runtime-parsing `ast!` macro as a "stopgap", as (a) it will just encourage people who are in a hurry to use it, and land us with technical debt and (b) it can't/won't handle symbol IDs etc, so it won't work properly anyway. So, all in all, I'm afraid I am not in favour of merging this PR.

7086cmd:
Got it.

crates/oxc_transformer/src/helpers/ast_helper.rs
7086cmd: Sorry to see it after closing :-), but actually the `helpers` does not apply directly in the AST. [Here](https://github.com/oxc-project/oxc/issues/4753#issuecomment-2354216972) I think it should be a better solution, and I am trying to work on it.

crates/oxc_transformer/src/helpers/ast_helper.rs
7086cmd: I may think it can't be for dynamic strings, e.g. for defining property.

PR #5774, opened 2024-09-14
fix(ast_tools): fix miscalculation of enum layouts.

rzvxa:
Fixes a bug with how we were calculating the size of enums when there are different padding on variants and there aren't enough niches. Now it aligns the largest variant size to the largest alignment before consuming niches.

### DIFF
--- a/tasks/ast_tools/src/passes/calc_layout.rs
+++ b/tasks/ast_tools/src/passes/calc_layout.rs
@@ -122,9 +122,14 @@ fn calc_enum_layout(ty: &mut Enum, ctx: &EarlyCtx) -> Result<PlatformLayout> {
         acc
     }
 
-    let with_tag = |mut acc: KnownLayout| -> KnownLayout {
-        acc.consume_niches(ty.item.variants.len() as u128, true);
-        acc
+    let with_tag = |layout: KnownLayout| -> Result<KnownLayout> {
+        let niches = layout.niches();
+        let layout = std::alloc::Layout::from_size_align(layout.size(), layout.align())
+            .normalize()?
+            .pad_to_align();
+        let mut layout = KnownLayout::new(layout.size(), layout.align(), niches);
+        layout.consume_niches(ty.item.variants.len() as u128, true);
+        Ok(layout)
     };
 
     let layouts = collect_variant_layouts(ty, ctx)?;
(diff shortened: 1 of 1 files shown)


rzvxa:
* **#5774** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (commented):
I am not familiar with the layout calculation logic in `ast_tools`, so I could be reading it wrong, but it looks like this is taking the niches of the variants' fields into account for niche calculation of the enum.

I don't think this is correct. e.g.: [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9637bea24752dc29bc2b1ccfb57cc6f):

* `StartNiche` has an unused discriminant at start of the `u8` range (0).
* `EndNiche` has an unused discriminant at end of the `u8` range (255).
* `NoNiche` has no unused discriminants at either start or end of `u8` range.

* `Option<StartNiche>` and `Option<EndNiche>` are 2 bytes (discriminant's niche used as `None`).
* `Option<NoNiche>` is 3 bytes (indicating `NoNiche` has no niche).

A couple of conclusions we can draw from this:

1. The number of variants is not relevant, only whether there are spare discriminant values at start or end of range (not in the middle).
2. Whether the fields have niches or not is also irrelevant. Compiler could have noticed that every variant in `NoNiche` has a field with 254 illegal values, and those illegal values overlap for all fields. It could have used one of those illegal values as a niche (`Option<NoNiche>::None` could be stored as 2 bytes `[0, 2]`). But it doesn't! An enum only has niches if there are available discriminant values.

You have other PRs stacked on top of this, so feel free to merge this, and then change the niche logic in a follow-up if you like.

overlookmotel:
Am going to merge this, and then address the issues raised in my [comment above](https://github.com/oxc-project/oxc/pull/5774#pullrequestreview-2306334340) in a follow-on.

overlookmotel:
### Merge activity

* **Oct 21, 8:18 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
> Am going to merge this, and then address the issues raised in my [comment above](https://github.com/oxc-project/oxc/pull/5774#pullrequestreview-2306334340) in a follow-on.

Actually, I can't understand this at all. To calculate niches correctly, need to know the value of the variant discriminants, which isn't easily available at this stage. I think to make it exactly correct, will need quite a large refactoring. In any case, for AST transfer, we need to know not just *how many* niches, but their *positions and values*. So will need a rework for that anyway. I'll return to this later.

PR #5790, opened 2024-09-15
feat(prettier): support typescript enums

Sysix:
The last failed test is only because of the comments. the oxc prettier output is:

```
enum Direction {
  Up = 1,
  Down,
  Left,
  Right,
}

enum FileAccess {
  None,
  Read = // constant members
  1 << 1,
  Write = 1 << 2,
  ReadWrite = Read | Write,
  G = // computed member
  "123".length,
}

enum Empty {}

const enum Enum {
  A = 1,
  B = A * 2,
}
```
Expected output: https://github.com/prettier/prettier/blob/aa3853b7765645b3f3d8a76e41cf6d70b93c01fd/tests/format/typescript/enum/__snapshots__/format.test.js.snap#L91-L99

Hope you can tell more why this happens :)


Boshen:
Because of we haven't figured out how to print comments :-/

PR #5795, opened 2024-09-15
perf(linter): move shared context info to `ContextHost`

DonIsaac:
> Related to #5770

## What This PR Does

Moves state that is constant over a linted file out of `LintContext` and into a shared `ContextHost` struct, turning `LintContext` into a [flyweight](https://en.wikipedia.org/wiki/Flyweight_pattern).

This brings `LintContext` from 144 bytes down to 96. `Linter::run` iterates over `(rule, ctx)` pairs in a very tight loop, and each rule instance gets its own clone of `ctx`. A smaller `LintContext` means better cache locality and a smaller heap allocation for this vec. I'm hoping to eventually get it small enough to fit in a single cache line.

I made a PR a while ago that did something similar to this one, but instead of using an `Rc`, each `LintContext` stored a read-only reference. This added an extra lifetime parameter, making it slightly unwieldy, but I saw up to 15%/25% perf improvements on local benchmarks. I'll dig around for it and add a link shortly.

### Architecture
image
----
image

### DIFF
--- a/crates/oxc_linter/src/context/host.rs
+++ b/crates/oxc_linter/src/context/host.rs
@@ -1,0 +1,177 @@
+use oxc_semantic::Semantic;
+use oxc_span::SourceType;
+use std::{cell::RefCell, path::Path, rc::Rc, sync::Arc};
+
+use crate::{
+    config::LintConfig,
+    disable_directives::{DisableDirectives, DisableDirectivesBuilder},
+    fixer::FixKind,
+    frameworks,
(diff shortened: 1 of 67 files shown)


DonIsaac:
* **#5795** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5795?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #5793, opened 2024-09-15
refactor(ast): move functions to top level in `ast` macro

overlookmotel:
Pure refactor. Convert nested functions to top level, to make it easier to read.

PR #5791, opened 2024-09-15
refactor(ast): split `ast` macro into multiple files

overlookmotel:
Pure refactor. Split implementation of `#[ast]` macro into multiple files. This means each file works with a single version of `TokenStream`, rather than having `proc_macro::TokenStream` and `proc_macro2::TokenStream` both in use in a single file.

Issue #4112, opened 2024-05-31
Shortcut for generating AST fragments


overlookmotel:
This is the code we have to write in transformer to generate `let Foo;`:

```rs
let kind = VariableDeclarationKind::Let;
let declarations = {
    let ident = BindingIdentifier::new(SPAN, name.clone());
    let pattern_kind = self.ctx.ast.binding_pattern_identifier(ident);
    let binding = self.ctx.ast.binding_pattern(pattern_kind, None, false);
    let decl = self.ctx.ast.variable_declarator(SPAN, kind, binding, None, false);
    self.ctx.ast.new_vec_single(decl)
};
Declaration::VariableDeclaration(self.ctx.ast.variable_declaration(
    SPAN,
    kind,
    declarations,
    Modifiers::empty(),
))
```

Seems a bit bananas! Very long to write, and non-trivial to understand at first glance what the code is doing.

Surely we can find some way to reduce this.

Like: `ast!{ let #name; }`

The obvious problem is compile time impact of macros. Can we find some way to precompile macro expansion so we can enjoy a simpler code style without paying the compile-time price?

overlookmotel:
Me, @Boshen, @rzvxa and @Dunqing discussed this today on a video call.

### How SWC does this

We noted that SWC does something similar to the `ast!` macro mentioned above. Their version is:

```rs
let stmt = quote!{
    "let $name = true;" as Stmt,
    name = BindingIdentifier { span: SPAN, name: "Foo" }
};
```

We decided that we'd prefer to have the template insertions outside of the macro like this:

```rs
let name = BindingIdentifier { span: SPAN, name: "Foo" };
let stmt = ast_stmt!{ let #name = 123; };
```

### Our plan

We agreed that:

* Ideal for the user is the `ast!` macro approach (compared to the other options of AST snippets, or adding loads of complicated methods to AST builder).
* We should try to do that.
* Parsing the JS code fragment inside the macro will happen at compile time.
* Macro will expand to `Statement::VariableDeclaration( /* ...etc... */ )`.
* Initial implementation will hurt compile times due to the macros.
* In a 2nd phase, we will improve compile time performance using some form of caching (as per rzvxa's comment above).
* rzvxa is going to take the lead on this.

I believe that's an accurate representation of our discussion. If not, please let me know!

Boshen:
nit: `#name` is a private identifier. We'll let the implementor to decide what marker to use.

overlookmotel:
[...]

PR #5796, opened 2024-09-15
feat(ast_tools): pre-compile `#[ast]` macro output

overlookmotel:
Work in progress.

Generate code for `#[ast]` macro with an `oxc_ast_tools` generator.

`oxc_ast_macros` crate now has no dependencies (no `syn` etc).

The guts of it is the `code!` macro. This is the same as `quote!`, except it evaluates to a `TokenStream` of code to *create* the `TokenStream` which `quote!` evaluates to.

Unfortunately I'm not seeing an improvement in compile times, because other dependencies of `oxc_ast` also use `syn`, so it doesn't drop from the dependency graph. ~~The generated code is 3000 lines, and could likely be shrunk down, which also could help.~~ The generated code is now 600 lines, and now there is a 5% reduction in debug build time for `oxc_ast` crate - but that's still a disappointingly small improvement.

Am opening this PR not necessarily with the intent of merging it, but to discuss whether this is maybe a useful approach. @rzvxa would appreciate your feedback.

Note: This technique would allow `#[generate_derive]` to inject the derived `impl` next to the type def, same as `#[derive]` does, which would solve the problem of types/fields needing to be `pub(crate)`.


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5796?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5796** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5793** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5793?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5792** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5792?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5791** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5791?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

tasks/ast_tools/src/to_code.rs
```diff
@@ -0,0 +1,189 @@
...
+/// The value in `@{...}` must be a `TokenStream`.
+/// e.g. `code!( foo( @{args} ) )` evaluates to a `TokenStream` containing:
+///
+/// ```
+/// TokenStream::from_iter([
+///     TokenTree::Ident(Ident::new("foo", Span::call_site())),
+///     TokenTree::Group(Group::new(Delimiter::Parenthesis, args))
+/// ].into_iter())
+/// ```
+macro_rules! code {
```
rzvxa, line 29: This is super smart! I like what I'm seeing👍

overlookmotel: Thanks. But unfortunately it's not moving the needle on performance.

I suspect (but don't know) that creating `TokenTree`s and `TokenStream`s is inherently expensive due to how proc macros interface with the compiler. There may just be no way to make proc macros cheap.

Or maybe the `#[ast]` macro just isn't that costly in the first place, so optimizing it is never going to pay dividends.

Thanks for taking a look. I'm going to pause on this though for now as am planning to spend this week focused on transformer. Must... not... get... distracted...

By the way, these 2 articles were the inspiration. They do a very good job of explaining how proc macros work internally. I always thought it involved serialization, but it turns out not:
* https://blog.jetbrains.com/rust/2022/03/18/procedural-macros-under-the-hood-part-i/
* https://blog.jetbrains.com/rust/2022/07/07/procedural-macros-under-the-hood-part-ii/

rzvxa: > Or maybe the #[ast] macro just isn't that costly in the first place, so optimizing it is never going to pay dividends.

I think this is more likely; No matter how much proc macro overhead costs us, If we do enough work in a macro eventually caching it would be faster. I suspect we have a better chance of getting some juice out of this approach in #4112 However it might be a wild goose chase.

And thanks for the links, They seem really interesting especially the second part(gave them a cursory look), I'll give them a read.

PR #5799, opened 2024-09-16
[RFC] feat(linter): support persisted rule state

DonIsaac:
Provides a way for rules to maintain state across invocations of `run`, `run_on_symbol`, and `run_once` without forcing those functions to take `&mut self`.

### Note
I really do not like this approach. It's way too loose w.r.t Rust's type checker, and solutions requiring `Any` are red flags to me. I'd much prefer rules to create state ad-hoc inside of `run_once` and then iterate over nodes themselves.

I also didn't test this code at all. For now, this PR is meant to be a conversation starter.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1297,6 +1297,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "650eef8c711430f1a879fdd01d4745a7deea475becfb90269c06775983bbf086"
 
 [[package]]
+name = "nohash-hasher"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bf50223579dc7cdcfb3bfcacf7069ff68243f8c363f62ffa99cf000a6b9c451"
+
+[[package]]
 name = "nom"
 version = "7.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1640,6 +1646,7 @@ dependencies = [
  "markdown",
  "memchr",
  "mime_guess",
+ "nohash-hasher",
  "once_cell",
  "oxc_allocator",
  "oxc_ast",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,6 +147,7 @@ memoffset = "0.9.1"
 miette = { version = "7.2.0", features = ["fancy-no-syscall"] }
 mimalloc = "0.1.43"
 mime_guess = "2.0.5"
+nohash-hasher = "0.2.0"
 nonmax = "0.5.5"
 num-bigint = "0.4.6"
 num-traits = "0.2.19"
--- a/crates/oxc_linter/Cargo.toml
+++ b/crates/oxc_linter/Cargo.toml
@@ -45,6 +45,7 @@ language-tags = { workspace = true }
 lazy_static = { workspace = true }
 memchr = { workspace = true }
 mime_guess = { workspace = true }
+nohash-hasher = { workspace = true }
 once_cell = { workspace = true }
 phf = { workspace = true, features = ["macros"] }
 rayon = { workspace = true }
--- a/crates/oxc_linter/src/context/host.rs
+++ b/crates/oxc_linter/src/context/host.rs
@@ -11,7 +11,7 @@ use crate::{
     utils, FrameworkFlags, RuleWithSeverity,
 };
 
-use super::{plugin_name_to_prefix, LintContext};
+use super::{plugin_name_to_prefix, LintContext, RuleState};
 
 /// Stores shared information about a file being linted.
 ///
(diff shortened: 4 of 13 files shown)


DonIsaac:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5799?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5799** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5799?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5795** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5795?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#5797](https://github.com/oxc-project/oxc/pull/5797) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5797?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

overlookmotel:
I haven't delved fully into this but...

> I really do not like this approach.

Don't do it then! You are the chief of the linter, it's your call.

Also, it seems my original issue #5204 which started all this was based on a fundamental misunderstanding (I didn't realize that `Rule`s are shared across threads). So it may be that what I suggested was a terrible idea! Please don't bend over backwards doing stuff you're not comfortable with just to satisfy my nonsense.

DonIsaac:
> I haven't delved fully into this but...
> 
> > I really do not like this approach.
> 
> Don't do it then! You are the chief of the linter, it's your call.
> 
> Also, it seems my original issue #5204 which started this all was based on a fundamental misunderstanding (I didn't realize that `Rule`s are shared across threads). So it may be that what I suggested was just a terrible idea! Please don't bend over backwards doing stuff you're not comfortable with just to satisfy my ill-informed nonsense.

Apologies for never responding with my thoughts. I do not enjoy writing and my thoughts about it could be quite long :P I'll get them down for you though.

> Side note: What's with the "no hash hasher" crate? This seems quite a weird idea. Maybe I'm missing something, but isn't it a recipe for hashmap collisions?
> 
> `FxHash` is surprisingly cheap for hashing integers. Check out how simple it is:
> 
> https://github.com/rust-lang/rustc-hash/blob/eb049a8209f58003957c34477a2d8d2729f6b633/src/lib.rs#L96-L101 https://github.com/rust-lang/rustc-hash/blob/eb049a8209f58003957c34477a2d8d2729f6b633/src/lib.rs#L167-L190
> 
> That is just 3 CPU ops to hash a `u32` or `u64`.

Its a crate that hashes small primitives using the identity function. It's supposed to replace a sparse array. There's a chance for collision if your have duplicate keys. In this case, all rules are guaranteed to have a unique ID, so this won't be a problem.

overlookmotel:
> I do not enjoy writing and my thoughts about it could be quite long :P I'll get them down for you though.

You don't need to explain yourself for me. I just wanted to make sure I hadn't pushed you in a direction you didn't really want to go in.

> Its a crate that hashes small primitives using the identity function. It's supposed to replace a sparse array. There's a chance for collision if your have duplicate keys. In this case, all rules are guaranteed to have a unique ID, so this won't be a problem.

As far as I can see, that's not *quite* true...

`nohash_hasher::IntMap` is a wrapper around `std::collections::HashMap` (which is itself a wrapper around `hashbrown`).

If the keys are `u32`, for example, it doesn't allocate `u32::MAX + 1` buckets (4 billion!). It'll instead allocate `1 << power` buckets, and keys are converted to bucket index with `key % (1 << power)` (i.e. take `power` bottom bits). If `power` is 8, then capacity is 256, so keys 10 and 266 will share the same bucket (collision).

So the question isn't "are the keys unique?" but "are the *bottom `power` bits* of keys unique?"

In addition, `hashbrown` uses the top 7 bits of hash to speed up searches, so you want those bits to have good entropy. If your keys don't exceed `1 << (32 - 7)` (33 million), and you use an identity hash function, then the top 7 bits of every hash will be 0. So this hobbles part of what makes `hashbrown` fast.

There will probably be some use cases where the pattern of keys being used may fit well with this scheme (you'd hope so, that crate has 13 million downloads!). But in many cases, it won't - the cost of hashing with a cheap hash function like `FxHash` will be smaller than the cost of frequent collisions.

I see the repo for `nohash_hasher` was archived in 2022. Perhaps identity hash function made sense before `std` switched to using `hashbrown` as its underlying hash table implementation a few years ago, but now it's less advantageous. Perhaps.

TLDR: I can't say for sure whether the pattern of keys here is suitable for `nohash_hasher`, but quite possibly not. I'd recommend at least testing `FxHashMap` and benchmarking the difference.

PR #5806, opened 2024-09-16
perf(linter): make all rules share a diagnostics vec

DonIsaac:
## What This PR Does

Each time a lint rule is run on a file, it will now store diagnostics in a
shared vec instead of having its own list. This is done by moving `diagnostics`
from `LintContext` to `ContextHost`.

The motivating line of code can be found in `Linter::run` in
`oxc_linter/src/lib.rs#145`

```rust
rules.into_iter().flat_map(|(_, ctx)| ctx.into_message()).collect::<Vec<_>>()
```

Each `LintContext` allocates a new vec, and pushes diagnostics into it. After
all run-related methods have been run, a new vec is created and those
diagnostics are copied into it. This is `O(n+1)` allocations and `O(n)` copies,
not to mention that allocations for the original vecs cannot be re-used since
those vecs aren't dropped until after the new one is created.

### DIFF
--- a/crates/oxc_linter/src/context/host.rs
+++ b/crates/oxc_linter/src/context/host.rs
@@ -5,7 +5,7 @@ use std::{cell::RefCell, path::Path, rc::Rc, sync::Arc};
 use crate::{
     config::LintConfig,
     disable_directives::{DisableDirectives, DisableDirectivesBuilder},
-    fixer::FixKind,
+    fixer::{FixKind, Message},
     frameworks,
     options::{LintOptions, LintPlugins},
     utils, FrameworkFlags, RuleWithSeverity,
@@ -33,9 +33,13 @@ use super::{plugin_name_to_prefix, LintContext};
 /// - [Flyweight Pattern](https://en.wikipedia.org/wiki/Flyweight_pattern)
 #[must_use]
 #[non_exhaustive]
-pub struct ContextHost<'a> {
+pub(crate) struct ContextHost<'a> {
     pub(super) semantic: Rc<Semantic<'a>>,
     pub(super) disable_directives: DisableDirectives<'a>,
+    /// Diagnostics reported by the linter.
+    ///
+    /// Contains diagnostics for all rules across a single file.
+    diagnostics: RefCell<Vec<Message<'a>>>,
     /// Whether or not to apply code fixes during linting. Defaults to
     /// [`FixKind::None`] (no fixing).
     ///
@@ -56,6 +60,8 @@ impl<'a> ContextHost<'a> {
         semantic: Rc<Semantic<'a>>,
         options: LintOptions,
     ) -> Self {
+        const DIAGNOSTICS_INITIAL_CAPACITY: usize = 512;
+
         // We should always check for `semantic.cfg()` being `Some` since we depend on it and it is
         // unwrapped without any runtime checks after construction.
         assert!(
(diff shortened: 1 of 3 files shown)


DonIsaac:
* **#5806** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5806?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Just to check my understanding of the linter is correct:

* For each individual file, all rules run on the same thread.
* Each file may run on a different thread.

Do I have that right? If so, this change does make some sense.

But... benchmarks seem to show this actually making things slightly slower rather than faster (-0.2% to -0.5% change). It's hard to tell if that's just noise in benchmarks. You may want to add a dummy commit to re-run the benchmarks, and then revert the commit to run them again. If the perf drop shows consistently across 3 runs, then it's probably real.

0.5% is not much, so it's no big deal either way. But it seems counter-productive to make a change specifically to improve perf, if it actually does the opposite.

If the perf drop *is* real, explanation may be that the additional cost of `RefCell` cloning is outweighing the gain of reducing memory copies.

Presumably in real world usage, most rules will report 0 errors on 99% of files, because the programmer will fix (or disable) any lint errors pretty quickly. And if there are no errors, then there's no memory to copy, so this change doesn't gain anything, but on negative side it incurs the cost of cloning lots of `RefCell`s.

(We could have our cake and eat it by replacing `RefCell` with zero-cost `GhostCell`, but that's a bit of an undertaking due to the weird lifetimes.)

DonIsaac:
> Just to check my understanding of the linter is correct:
> 
> * For each individual file, all rules run on the same thread.
> * Each file may run on a different thread.
> 
> Do I have that right? If so, this change does make some sense.

Yes

> But... benchmarks seem to show this actually making things slightly slower rather than faster (-0.2% to -0.5% change). It's hard to tell if that's just noise in benchmarks. You may want to add a dummy commit to re-run the benchmarks, and then revert the commit to run them again. If the perf drop shows consistently across 3 runs, then it's probably real.
> 
> 0.5% is not much, so it's no big deal either way. But it seems counter-productive to make a change specifically to improve perf, if it actually does the opposite.

I'll investigate those regressions. Locally, I was seeing a 7% perf improvement on the RadixUIAdoptionSection benchmark. Maybe it has to do with the number of rules?

> PS: Unrelated, but I also wonder if initial capacity of 128 is too high. Do files in general produce that many errors? If I'm right that 0 is the most common number of errors, it might be better not to allocate any initial capacity at all.

I'll play around with the numbers and see what works best

DonIsaac:
...And now it's even worse in CI... that's odd, I'll keep looking into it.

overlookmotel:
@DonIsaac I'm not sure if you intended this to be merged, given that you were still investigating.

> Locally, I was seeing a 7% perf improvement on the RadixUIAdoptionSection benchmark.

This is really odd, and worth investigating. If our CodSpeed benchmarks are that far off reality, we should try to fix that (or at least understand in what circumstances it happens).

When you ran it locally, did you run with the same rules enabled as on CI? (i.e. was it exact like-for-like comparison?)

How many errors does running all the rules on RadixUIAdoptionSection produce?

If my assumption is correct that in real world most commonly the linter will produce 0 errors, and our benchmark fixtures produce loads, then perhaps we need new benchmark fixtures!

PR #5823, opened 2024-09-17
feat(codegen): have `with_source_text` reserve memory for code buffer

DonIsaac:
`Codegen::with_source_text` gives us enough information to use `Codegen::with_capacity`. This change makes the API cleaner as users have to provide less redundant information.

### DIFF
--- a/crates/oxc_codegen/src/lib.rs
+++ b/crates/oxc_codegen/src/lib.rs
@@ -41,9 +41,13 @@ pub type CodeGenerator<'a> = Codegen<'a>;
 #[derive(Default, Clone, Copy)]
 pub struct CodegenOptions {
     /// Use single quotes instead of double quotes.
+    ///
+    /// Default is `false`.
     pub single_quote: bool,
 
     /// Remove whitespace.
+    ///
+    /// Default is `false`.
     pub minify: bool,
 }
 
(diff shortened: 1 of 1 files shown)


crates/oxc_codegen/src/lib.rs
```diff
@@ -158,7 +167,9 @@ impl<'a> Codegen<'a> {
     #[must_use]
     pub fn with_capacity(mut self, source_text_len: usize) -> Self {
         let capacity = if self.options.minify { source_text_len / 2 } else { source_text_len };
-        self.code = Vec::with_capacity(capacity);
+        // ensure space for at least `capacity` additional bytes without clobbering existing
+        // allocations.
+        self.code.reserve(capacity);
```
overlookmotel, line 172: I must say, I don't understand `Codegen`'s API. I'd have expected it'd be impossible to add to `self.code` - I'd have expected you just call `build`, and that consumes the `Codegen`.

But no! We have public `print_char`, `print_str` and `print_expression` methods which push to `self.code`. I'm not sure if that's a mistake or not.

But as the API stands now, it looks like `self.code.reserve(capacity)` is correct, so as not to clobber any existing data in `self.code`.

DonIsaac: I agree, its API is quite confusing. For example, there are 3 methods that take source_text. 

The linter uses Codegen in fixers, where providing source_text and an entire AST is not desireable. Maybe we should separate out expression printing and provide two different APIs for those two different use cases?

overlookmotel: > Maybe we should separate out expression printing and provide two different APIs for those two different use cases?

I would say that's a good idea, yes.

PR #5835, opened 2024-09-17
feat(prettier): improve ts class output 

Sysix:
please look first at  #5834 :)

### DIFF
--- a/crates/oxc_prettier/src/format/mod.rs
+++ b/crates/oxc_prettier/src/format/mod.rs
@@ -628,8 +628,17 @@ impl<'a> Format<'a> for TSTypeAliasDeclaration<'a> {
     fn format(&self, p: &mut Prettier<'a>) -> Doc<'a> {
         let mut parts = p.vec();
 
+        if self.declare {
+            parts.push(ss!("delcare "));
+        }
+
         parts.push(ss!("type "));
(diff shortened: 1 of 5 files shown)


crates/oxc_prettier/src/format/mod.rs
```diff
@@ -860,7 +869,45 @@ impl<'a> Format<'a> for TSLiteralType<'a> {
...
+
+        if let Some(name_type) = &self.name_type {
+            parts.push(ss!(" as "));
+            parts.push(name_type.format(p));
+        }
+
+        parts.push(ss!("]"));
+
+        match self.optional {
+            TSMappedTypeModifierOperator::Plus => parts.push(ss!("+?")),
```
DonIsaac, line 892: Could we add an `as_str` method to `TSMappedTypeModifierOutput` in `oxc_ast`?

Sysix: Should I use for  `TSMappedTypeModifierOperator::None`  an empty string? 
Is it okay to push empty string or should I check?

DonIsaac: Good find, ignore this comment then!

DonIsaac (approved):

PR #5838, opened 2024-09-17
feat(prettier): class improvements part 2

Sysix:


crates/oxc_prettier/src/format/mod.rs
```diff
@@ -1296,10 +1352,6 @@ impl<'a> Format<'a> for TSTypeParameterDeclaration<'a> {
         let mut parts = p.vec();
         let mut print_comma = false;
 
-        if self.params.len() == 0 {
```
Sysix, line 1299: because of this: https://github.com/prettier/prettier/blob/aa3853b7765645b3f3d8a76e41cf6d70b93c01fd/tests/format/typescript/class/__snapshots__/format.test.js.snap#L56-L58 :/

Boshen:
NICE!

PR #5847, opened 2024-09-18
fix(semantic): analyze `ReferenceFlags` incorrectly when there are nested `AssignmentTarget`

Dunqing:
close: #5801

PR #5850, opened 2024-09-18
perf(transformer): arrow function transform: calculate whether `this` is in arrow function lazily

Dunqing:
I initially added `inside_arrow_function_stack` as we haven't got `TraverseCtx` in `Transformer`. Now we can use `ctx.current_scope_flags().is_arrow()` directly instead

### DIFF
--- a/crates/oxc_transformer/src/es2015/arrow_functions.rs
+++ b/crates/oxc_transformer/src/es2015/arrow_functions.rs
@@ -73,7 +73,10 @@
 use oxc_allocator::Vec;
 use oxc_ast::{ast::*, NONE};
 use oxc_span::SPAN;
-use oxc_syntax::{scope::ScopeFlags, symbol::SymbolFlags};
+use oxc_syntax::{
+    scope::{ScopeFlags, ScopeId},
+    symbol::SymbolFlags,
+};
 use oxc_traverse::{Ancestor, Traverse, TraverseCtx};
 use serde::Deserialize;
 
@@ -93,8 +96,6 @@ pub struct ArrowFunctions<'a> {
     ctx: Ctx<'a>,
     _options: ArrowFunctionsOptions,
     this_var_stack: std::vec::Vec<Option<BoundIdentifier<'a>>>,
-    /// Stack to keep track of whether we are inside an arrow function or not.
-    inside_arrow_function_stack: std::vec::Vec<bool>,
 }
 
 impl<'a> ArrowFunctions<'a> {
@@ -102,9 +103,8 @@ impl<'a> ArrowFunctions<'a> {
         Self {
             ctx,
             _options: options,
-            // Initial entries for `Program` scope
+            // Initial entry for `Program` scope
             this_var_stack: vec![None],
-            inside_arrow_function_stack: vec![false],
         }
     }
 }
@@ -115,8 +115,6 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
 
     /// Insert `var _this = this;` for the global scope.
     fn exit_program(&mut self, program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
-        debug_assert!(self.inside_arrow_function_stack.len() == 1);
-
         assert!(self.this_var_stack.len() == 1);
         let this_var = self.this_var_stack.pop().unwrap();
         if let Some(this_var) = this_var {
@@ -127,7 +125,6 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
     fn enter_function(&mut self, func: &mut Function<'a>, _ctx: &mut TraverseCtx<'a>) {
         if func.body.is_some() {
             self.this_var_stack.push(None);
-            self.inside_arrow_function_stack.push(false);
         }
     }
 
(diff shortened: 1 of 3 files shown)


overlookmotel (changes requested):
In addition to the "are we in an arrow function?" problem noted below, I've also pushed a commit that removes unused methods from `src/lib.rs`.

It's an unfortunate consequence of making all transforms implement `Traverse` that the compiler doesn't spot these things for us now, because `Traverse` provides empty default methods. I didn't think of that when I suggested that change.

overlookmotel:
It'll be interesting to see if this improves performance or not once #5851 is merged.

overlookmotel:
Does this PR depend on #5847 or #5846? If not, you might want to move this PR into its own stack, based on latest main with my recent changes to arrow functions transform.

Dunqing:
> Does this PR depend on #5847 or #5846? If not, you might want to move this PR into its own stack, based on latest main with my recent changes to arrow functions transform.

Thank you point this out, It doesn't depend.

overlookmotel:
Something has gone wrong with rebase. Commits from #5846 and #5847 are still included in this PR. Graphite doesn't make this easy!

Dunqing:
> Something has gone wrong with rebase. Commits from #5846 and #5847 are still included in this PR. Graphite doesn't make this easy!

I've found that once this PR contains extra commits, the Graphite will break when use `move` command

crates/oxc_transformer/src/es2015/arrow_functions.rs
```diff
@@ -174,7 +142,7 @@ impl<'a> Traverse<'a> for ArrowFunctions<'a> {
         ctx: &mut TraverseCtx<'a>,
     ) {
         if let JSXElementName::ThisExpression(this) = element_name {
-            if !self.is_inside_arrow_function() {
+            if !ctx.current_scope_flags().is_arrow() {
```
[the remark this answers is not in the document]
Dunqing: I updated the output.js where you added the test to match Babel's output and they all passed. Do you think the new output makes sense for you?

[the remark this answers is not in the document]
overlookmotel: This had become a bit of a mess with recent changes from main getting clobbered when merging this PR. That's what was causing the class static block test to fail. I've taken the liberty of squashing all the commits, rebasing on main, and fixing the problems from the faulty merge. Now this PR only contains the intended changes, and it's easier to see what's going on.

The new test I added `use-this-inside-blocks-in-arrow` is failing, and it shouldn't be.

I believe the problem is this:

`ScopeFlags` for a scope doesn't inherit the `ArrowFunction` flag from parent scope.

[Oxc playground](https://playground.oxc.rs/#eNo9jVEKAyEMRK8S8tXCnsDSnqQ/1sZWcOOiEbos3r0Rcb9m3kyYHOjQoK/sJCSGVIXy5QrHkwEiCXi4g/L9MSKYMlrRVr6h3EbYujQFNbhgQnNgrtyl7Cz2h0ZypQVjYJm+uLTRCfv6SnGSZMvFp7yi8TYWagtuNhfKuqi+r0x/ns7ApTd9qP9WWAMHH0bV2h/EjVDh)

The scope which `this` is in (scope 3) is a nested block, so has flags `ScopeFlags::empty()`. So `ctx.current_scope_flags().is_arrow()` returns `false`.

To correctly determine if `this` is in an arrow function would require tracing up the scope chain until hit either an arrow function (yes) or a non-arrow function or static block (no) - similar to the logic in `get_this_identifier`.

overlookmotel:
@Dunqing I know you're working on isolated declarations, but is it possible to finish this off? I have a few more changes to arrow functions transform, but I don't want to do them until this is merged, to avoid having merge conflicts mess again.

Or, if you're busy, would you like me to finish it up? I think I know what needs to be done.

Dunqing:
> @Dunqing I know you're working on isolated declarations, but is it possible to finish this off? I have a few more changes to arrow functions transform, but I don't want to do them until this is merged, to avoid having merge conflicts mess again.
> 
> Or, if you're busy, would you like me to finish it up? I think I know what needs to be done.

Thank you! Feel free to edit this PR or create a new one instead!

overlookmotel (approved):

crates/oxc_transformer/src/es2015/arrow_functions.rs
```diff
@@ -324,7 +246,76 @@ impl<'a> ArrowFunctions<'a> {
...
+        //     static { this }
+        // }
+        // ```
+        //
+        // So in this loop, we only exit when we encounter one of the above.
+        for ancestor in ctx.ancestors() {
+            match ancestor {
+                // Top level
+                Ancestor::ProgramBody(_)
+                // Function (includes class method body)
```
Dunqing, line 293: Is it possible to implement it by iterating over ancestors scopes?

overlookmotel: Unfortunately not, in order to pass these 2 tests (which Babel didn't have, but I added):

* [this-in-class-property/input.js](https://github.com/oxc-project/oxc/blob/6e0c12f5c3a552fece9773be95b56245d05389b1/tasks/transform_conformance/tests/babel-plugin-transform-arrow-functions/test/fixtures/this-in-class-property/input.js)
* [this-in-class-property-key/input.js](https://github.com/oxc-project/oxc/blob/6e0c12f5c3a552fece9773be95b56245d05389b1/tasks/transform_conformance/tests/babel-plugin-transform-arrow-functions/test/fixtures/this-in-class-property-key/input.js)

To get both of them right, we have to figure out whether `this` is in:

1. Class property value - `class { x = this; }` or
2. Class property key - `class { [this] = 1; }`

There is no `ScopeFlags::ClassProperty`, so only way to tell the difference is to go up the ancestor chain.

crates/oxc_transformer/src/es2015/arrow_functions.rs
```diff
@@ -324,7 +246,76 @@ impl<'a> ArrowFunctions<'a> {
...
+                }
+                Ancestor::ArrowFunctionExpressionParams(func) => {
+                    return Some(func.scope_id().get().unwrap())
+                }
+                Ancestor::ArrowFunctionExpressionReturnType(func) => {
+                    return Some(func.scope_id().get().unwrap())
+                }
+                Ancestor::ArrowFunctionExpressionBody(func) => {
+                    return Some(func.scope_id().get().unwrap())
+                }
```
overlookmotel, line 314: @Dunqing Do we need to consider the TS fields here? Is it possible to have a `this` expression in any of these:

* `Function::type_parameters`
* `Function::this_param`
* `Function::return_type`
* `ArrowFunctionExpression::type_parameters`
* `ArrowFunctionExpression::return_type`

?

(again, my poor knowledge of TypeScript is hindering me)

Dunqing: No, We don't care about TS fields. In TypeScript, it has a `TSThisType` rather than `ThisExpression`

crates/oxc_transformer/src/es2015/arrow_functions.rs
Dunqing: It sounds like we can combine these two ways. Let's try it in another PR if we want.

PR #5857, opened 2024-09-18
refactor(syntax)!: remove `SymbolFlags::ArrowFunction`

overlookmotel:
`SymbolFlags::ArrowFunction` is an oddity, as whether a symbol is an arrow function is not statically knowable. In the following cases, `f` symbol did not have `ArrowFunction` flag set:

```js
const {f} = {f: () => {}};
```

```js
let f = 123;
f = () => {};
```

`SymbolFlags::ArrowFunction` is therefore not particularly useful, and possibly misleading. Having it complicates the transformer, and it's not used anywhere in Oxc.

This PR removes it.

### DIFF
--- a/crates/oxc_semantic/src/binder.rs
+++ b/crates/oxc_semantic/src/binder.rs
@@ -23,7 +23,7 @@ pub(crate) trait Binder<'a> {
 
 impl<'a> Binder<'a> for VariableDeclarator<'a> {
     fn bind(&self, builder: &mut SemanticBuilder<'a>) {
-        let (mut includes, excludes) = match self.kind {
+        let (includes, excludes) = match self.kind {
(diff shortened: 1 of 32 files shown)


overlookmotel:
@DonIsaac Have requested your review as I think you added this flag. But I'm not clear why - it's not used in the linter (or anywhere else in Oxc).

Boshen (approved):

Dunqing (approved):

DonIsaac (approved):

DonIsaac:
> @DonIsaac Have requested your review as I think you added this flag. But I'm not clear why - it's not used in the linter (or anywhere else in Oxc).

I did add it for the linter, but I did not refactor existing rules to use it. Removing this should be fine, as we can achieve what we need without it.

PR #5867, opened 2024-09-18
feat(linter): implement `no-extend-native` rule

camchenry:
- part of https://github.com/oxc-project/oxc/issues/479

This implements the `no-extend-native` rule from the ESLint core ruleset. This was translated based on the rule source code from the ESLint codebase. That means some oddity, such as treating any capitalized identifier (like `Math`) as a global object, is preserved. So even though `Math.prototype` doesn't exist in reality, we still report it as an error. This is probably okay given that such code doesn't make sense anyway.


crates/oxc_linter/src/rules/eslint/no_extend_native.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+}
+
+/// If this usage of `*.prototype` is a `Object.defineProperty` or `Object.defineProperties` call,
+/// then this function returns the `CallExpression` node.
+fn get_define_property_call<'a>(
+    ctx: &'a LintContext,
+    node: &AstNode<'a>,
+) -> Option<&'a AstNode<'a>> {
+    let mut ancestor = ctx.nodes().parent_node(node.id())?;
```
DonIsaac, line 145: Consider iterating over parent nodes instead. Use `nodes().iter_parents().skip(1)`.

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+}
+
+/// Get an assignment to the property of the given node.
+/// Example: `*.prop = 0` where `*.prop` is the given node.
+fn get_property_assignment<'a>(
+    ctx: &'a LintContext,
+    node: &AstNode<'a>,
+) -> Option<&'a AstNode<'a>> {
+    let mut parent = ctx.nodes().parent_node(node.id())?;
```
DonIsaac, line 187: Same here

DonIsaac (commented):
Looking pretty good, just a few nitpicks and we should be good to go!

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
camchenry: https://github.com/oxc-project/oxc/pull/5867/commits/ea19b08aa3550d365544109d01a026471652e7b3

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
camchenry: https://github.com/oxc-project/oxc/pull/5867/commits/ea19b08aa3550d365544109d01a026471652e7b3

DonIsaac (approved):

PR #5869, opened 2024-09-19
perf(linter): Perform fewer allocation in `jest/no-deprecated-functions`

camchenry:
hypothesis: flamegraphs show that this rule is copying and allocating quite a bit of the time. I wanted to try and see if using a single String would be more efficient.


DonIsaac:
2% bump on cal.com.tsx benchmarks. Nice!

camchenry:
@DonIsaac I also see a -2% decrease on the checker.ts benchmark, so it seems like it cancels out? Still worth it you think?

DonIsaac:
Hmmm it could just be noise... What do you see when you run benchmarks locally?

PR #5874, opened 2024-09-19
fix(isolated_declarations): Don't require private method arguments to have type annotations

MichaelMitchell-at:
In https://github.com/oxc-project/oxc/pull/4934 I introduced a regression which caused the following code to begin having an isolated declarations error where it shouldn't:
```ts
export class Foo {
  private foo(a): void {}
}
```
[TS playground](https://www.typescriptlang.org/play/?isolatedDeclarations=true#code/KYDwDg9gTgLgBAYwDYEMDOa4DEITgbwCg44B6UuAARjQFpQxgEZ6oppi4woBLANxQxgcAGa4AFCgCUALjh8IPACYEAvoVVA)

### DIFF
--- a/crates/oxc_isolated_declarations/src/class.rs
+++ b/crates/oxc_isolated_declarations/src/class.rs
@@ -376,6 +376,8 @@ impl<'a> IsolatedDeclarations<'a> {
                         continue;
                     }
 
+                    let errors_length = self.errors.borrow().len();
+
                     let inferred_accessor_types = self.collect_inferred_accessor_types(decl);
                     let function = &method.value;
                     let params = if method.kind.is_set() {
@@ -404,6 +406,7 @@ impl<'a> IsolatedDeclarations<'a> {
                     }
 
                     if method.accessibility.is_some_and(TSAccessibility::is_private) {
+                        self.errors.borrow_mut().truncate(errors_length);
                         elements.push(self.transform_private_modifier_method(method));
                         continue;
                     }
--- a/crates/oxc_isolated_declarations/tests/fixtures/function-parameters.ts
+++ b/crates/oxc_isolated_declarations/tests/fixtures/function-parameters.ts
@@ -14,6 +14,10 @@ const x = 42;
 const y = '';
 export function fooGood3({a = x, b: [{c = y}]}: object): void {}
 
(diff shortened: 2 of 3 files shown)


crates/oxc_isolated_declarations/src/class.rs
```diff
@@ -404,6 +406,7 @@ impl<'a> IsolatedDeclarations<'a> {
                     }
 
                     if method.accessibility.is_some_and(TSAccessibility::is_private) {
+                        self.errors.borrow_mut().truncate(errors_length);
```
MichaelMitchell-at, line 409: This is a little bit hacky. I think a less hacky way might be to decouple methods like `transform_formal_parameters` from generating errors and then move the check for errors after this block, but I wanted to get some feedback before starting to attempt something like that.

Dunqing:
Sorry, I didn't see any errors in your case. 

The private method will turn into a property without type annotations. The output is

```
export class Foo {
  private foo;
}
```

MichaelMitchell-at:
> Sorry, I didn't see any errors in your case.
> 
> The private method will turn into a property without type annotations.

That's the correct behavior, but currently there is an error when using oxc_isolated_declarations. This PR removes the error. If you pull this branch locally and undo the change I made to `class.rs`, the snapshot test will begin to fail.

Dunqing:
> That's the correct behavior, but currently there is an error when using oxc_isolated_declarations. This PR removes the error. If you pull this branch locally and undo the change I made to `class.rs`, the snapshot test will begin to fail.

Ah, you are right. I found our example that doesn't report any isolated declaration errors. Fixes in #5918

crates/oxc_isolated_declarations/src/class.rs
Dunqing: I think we can handle the private method early, can you try it?

Dunqing:
Fixed in #5964

We are about to release a new version that includes fixes for many incorrect outputs in `oxc-isolated-declarations`, as well as support for the `strip_internal` option. Therefore, I will fix all known bugs before the next release. Thank you for trying to fix it.

MichaelMitchell-at:
Thank you!

PR #5887, opened 2024-09-19
feat(transformer): sync `Program::source_type` after transform

### LINKED ISSUE #5552: transformer: change `program.source_type`
Need to update 

https://github.com/oxc-project/oxc/blob/81a394d11d732ad58c7a4a5e3589e1e9a41313fd/crates/oxc_ast/src/ast/js.rs#L41

to keep things in sync.

Boshen:
closes #5552

### DIFF
--- a/crates/oxc_transformer/src/typescript/mod.rs
+++ b/crates/oxc_transformer/src/typescript/mod.rs
@@ -77,6 +77,7 @@ impl<'a> Traverse<'a> for TypeScript<'a> {
             program.hashbang = None;
             program.body.clear();
         } else {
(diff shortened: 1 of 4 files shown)


crates/oxc_transformer/src/typescript/mod.rs
```diff
@@ -77,6 +77,7 @@ impl<'a> Traverse<'a> for TypeScript<'a> {
             program.hashbang = None;
             program.body.clear();
         } else {
+            program.source_type = program.source_type.with_javascript(false);
```
overlookmotel, line 80: Should be `true` I think.

PR #5892, opened 2024-09-19
perf(linter): use `aho-corasick` instead of `regex` for string matching in `jsx-a11y/img-redundant-alt`

camchenry:
hypothesis: profiling shows that Regex creation takes a decent amount of time. the `regex` crate uses `aho-corasick` internally for string matching, which is all we need in some cases. in theory, we could save time by using the lib directly and not needing the full regex syntax.

PR #5901, opened 2024-09-19
perf(linter): Use `aho-corasick` instead of `regex` for default comment patterns in `no-fallthrough`

camchenry:
theory: using Regex is slow in most cases, since we have a fixed number of patterns that we need to search for. but we do still need to support it since the user can pass a custom regex pattern. however, we can at least try to use a simpler `aho-corasick` matcher for the default strings, which should likely be the majority of cases.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -1,5 +1,6 @@
 use std::ops::Range;
 
+use aho_corasick::AhoCorasick;
 use itertools::Itertools;
 use oxc_ast::{
     ast::{Statement, SwitchCase, SwitchStatement},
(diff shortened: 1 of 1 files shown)


camchenry:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5901?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5901** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5901?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5892** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
```diff
@@ -53,9 +56,18 @@ impl NoFallthrough {
...
+            None
+        };
+
         Self(Box::new(Config {
-            comment_pattern: Regex::new(format!("(?iu){comment_pattern}").as_str()).unwrap(),
+            custom_comment_pattern,
+            default_comment_pattern: AhoCorasick::builder()
+                .ascii_case_insensitive(true)
+                .build(DEFAULT_FALLTHROUGH_COMMENT_PATTERNS)
+                .expect("Could not build AhoCorasick for default comment patterns"),
```
DonIsaac, line 70: Can we provide the name of the rule so users know where and how to fix their configs?

camchenry:
I tested this locally and it seems like this might not actually be much of a perf win. Seems like it is 2-3% slower on the large benchmarks like checker.ts and cal.com, but faster in small files (~6-7%), but not at a statistically significant level.

DonIsaac:
This PR could still be useful if it lets us remove our dependency on `regex`

camchenry:
@DonIsaac I think we still have to depend on regex still unfortunately, because the custom comment pattern config is allowed to be any pattern.

PR #5904, opened 2024-09-20
refactor(ast): remove `.iter()` where not needed

camchenry:
this is just an experiment, I noticed while profiling these code paths came up a number of times:
Screenshot 2024-09-19 at 8 32 51 PM


camchenry:
just trying some stuff, going to close this

Boshen:
This seems like a acceptable improvement. Any improvement that reduces something is 👍 from me.

camchenry:
@Boshen I didn't think about it like that, but makes sense to me. If you're amenable to it then, I'll re-open

PR #5907, opened 2024-09-20
feat(span)!: expose `SourceType` fields; change builder pattern to setters

Boshen:
Hated the `with_xxx(&self, yes: bool)` thing, its behavior was too confusing.

### DIFF
--- a/crates/oxc_semantic/tests/integration/util/mod.rs
+++ b/crates/oxc_semantic/tests/integration/util/mod.rs
@@ -45,7 +45,7 @@ impl<'a> SemanticTester<'a> {
     ///
     /// Use [`SemanticTester::ts`] for TypeScript test cases without JSX.
     pub fn tsx(source_text: &'static str) -> Self {
-        Self::ts(source_text).with_jsx(true)
+        Self::ts(source_text).set_jsx()
     }
 
     /// Create a new tester for a JavaScript test case.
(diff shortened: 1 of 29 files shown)


Boshen:
* **#5907** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5907?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/tests/integration/util/mod.rs
```diff
@@ -71,22 +71,27 @@ impl<'a> SemanticTester<'a> {
...
-        self.source_type = self.source_type.with_jsx(yes);
+    pub fn set_jsx(mut self) -> Self {
+        self.source_type = self.source_type.set_jsx();
         self
     }
 
     /// Mark the [`SourceType`] as an ESM module.
-    pub fn with_module(mut self, yes: bool) -> Self {
-        self.source_type = self.source_type.with_module(yes);
+    pub fn set_module(mut self) -> Self {
```
DonIsaac, line 88: Issue (nonblocking): These appear to still be builder methods since they don't take an `&mut self`. Would we keep the name as `with_*`?

crates/oxc_span/src/source_type/mod.rs
```diff
@@ -269,68 +265,50 @@ impl SourceType {
...
+        self.module_kind = ModuleKind::Unambiguous;
         self
     }
 
     #[must_use]
-    pub const fn with_javascript(mut self, yes: bool) -> Self {
-        if yes {
-            self.language = Language::JavaScript;
-        }
+    pub const fn set_js(mut self) -> Self {
```
DonIsaac, line 286: praise: I much prefer the intent being in the name over passing a Boolean parameter

crates/oxc_span/src/source_type/mod.rs
```diff
@@ -20,13 +20,13 @@ pub use error::UnknownExtension;
 #[serde(rename_all = "camelCase")]
 pub struct SourceType {
     /// JavaScript or TypeScript, default JavaScript
-    pub(super) language: Language,
+    pub language: Language,
```
DonIsaac, line 23: Issue: can we avoid making these public? Barring that, we should make it non_exhaustive

crates/oxc_span/src/source_type/mod.rs
```diff
@@ -269,68 +265,50 @@ impl SourceType {
     }
 
     #[must_use]
-    pub const fn with_script(mut self, yes: bool) -> Self {
```
DonIsaac, line 272: Question: should we support both setter and builder methods? Source type is a user-facing API, and users may want one or the other depending on their use case

Boshen:
Re builder pattern: let's try and remove this pattern from oxc altogether, I find it more and more unintuitive. It is a neat pattern when you discover it, but gets confusing once you add more fields and methods to it.

The builder pattern should be only used for structs that are called a builder, e.g. https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#

Boshen:
I think I need to find a better contention for these setters.

PR #5909, opened 2024-09-20
refactor(isolated-declarations): pre-filter statements that do not need to be transformed

Dunqing:
We only transform `Declaration` and `ModuleDeclaration` in `IsolatedDeclaration`. We pre-filter statements that need to transform and then use `clone_in` which will avoid producing overhead for clone `Statements`.

### DIFF
--- a/crates/oxc_isolated_declarations/src/lib.rs
+++ b/crates/oxc_isolated_declarations/src/lib.rs
@@ -22,7 +22,7 @@ mod types;
 use std::{cell::RefCell, collections::VecDeque, mem};
 
 use diagnostics::function_with_assigning_properties;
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#5909** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5909?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
I found that performance regression due to the use of `clone_in` replace with `ast.copy `

Dunqing:
FYI: This PR changes handling way from fold pattern to mutating cloned AST

PR #5911, opened 2024-09-20
feat(linter): Implement `no-unexpected-multiline` rule

camchenry:
- part of https://github.com/oxc-project/oxc/issues/479

The bulk of this rule is closely based on the logic from the original ESLint rule. One major difference between this implementation and the original though is the lack of a tokenizer. ESLint uses a proper tokenizer to find identifers, parentheses, brackets, and other tokens. For this rule, I opted to just manually search for the characters we might expect to find. I'm not sure how this will hold up in the real world, I expect it could lead to some missing cases potentially, but it at least works for all of the given test cases.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
@@ -1,0 +1,383 @@
+use memchr::{memchr, memrchr};
+use oxc_ast::{ast::BinaryOperator, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+#[derive(Debug, Default, Clone)]
+pub struct NoUnexpectedMultiline;
+
+enum DiagnosticKind {
+    FunctionCall { open_paren_span: Span },
+    PropertyAccess { open_bracket_span: Span },
+    TaggedTemplate { backtick_span: Span },
+    Division { slash_span: Span },
+}
+
+fn no_unexpected_multiline_diagnostic(kind: &DiagnosticKind) -> OxcDiagnostic {
+    match kind {
+        DiagnosticKind::FunctionCall { open_paren_span } => OxcDiagnostic::warn(
+            "Unexpected newline between function name and open parenthesis of function call",
+        )
+        .with_label(
+            open_paren_span.label("this is parsed as a function call, which may be unintentional"),
+        )
+        .with_help(
+            "If you did not intend to make a function call, insert ';' before the parenthesis",
+        ),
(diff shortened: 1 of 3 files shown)


camchenry:
I still need to polish this, add more docs, and add better diagnostics, then will put up for review

crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,369 @@
...
+                };
+                let Some(newline) = memchr(b'\n', src) else {
+                    return;
+                };
+                if newline < open_paren {
+                    let paren_span = Span::new(
+                        span.start + u32::try_from(open_paren).unwrap(),
+                        span.start + u32::try_from(open_paren + 1).unwrap(),
+                    );
+                    let diagnostic = OxcDiagnostic::warn(
```
DonIsaac, line 84: Please move diagnostic building to a separate function (at the top of the file). It makes quickly seeing what kind of messages this lint rule warns about much easier

crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,369 @@
...
+                let src = ctx.source_range(span).as_bytes();
+                let Some(open_bracket) = memchr(b'[', src) else {
+                    return;
+                };
+                let Some(newline) = memchr(b'\n', src) else {
+                    return;
+                };
+                if newline < open_bracket {
+                    let bracket_span = Span::new(
+                        span.start + u32::try_from(open_bracket).unwrap(),
```
DonIsaac, line 114: you may find `Span::sized` helpful

crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
```diff
@@ -0,0 +1,369 @@
...
+                    return;
+                }
+
+                // find index of first non-identifier character after the second slash
+                let ident_name: &str = if let Some(offset) = src[(second_slash + 1)..]
+                    .as_bytes()
+                    .iter()
+                    .position(|&b| !b.is_ascii_alphanumeric())
+                {
+                    &src[second_slash + 1..second_slash + 1 + offset]
```
DonIsaac, line 200: issue: I've seen ranged index on `source_text` cause rules to panic in the past (e.g. weird unicode characters cause this to end on a character boundary). Is it possible to refactor this logic to use `chars()` instead?

camchenry: good idea, I wish we had a good way to quickly tokenize some text here, but I refactored this to at least do a good enough job at finding identifier-like strings

crates/oxc_linter/src/rules/eslint/no_unexpected_multiline.rs
camchenry: good to know!

DonIsaac (approved):

PR #5923, opened 2024-09-20
chore(oxc): add `.idea` files to gitignore

radu2147:
Given that RustRover now has a community license, I think people might be interested in using for open-source contributions, so I think it can be a good idea to add `.idea/` directory to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ target/
 
 /editors/vscode/out/
 /editors/vscode/*.vsix
+.idea/
 
 npm/cli-*
 


Boshen:
https://stackoverflow.com/a/7335487

overlookmotel:
Just to expand on what Boshen said, our feeling is that the place for this is in user's local `.gitignore` file, not the repo's. `.gitignore` in this repo also does not include other environment-specific files e.g. `.DS_Store`.

That said @Boshen do you think we should change that policy? This is the 3rd time in a few months that someone has made a PR like this (I was one of them). I believe current policy is "correct", but if it eases entry for contributors, maybe that's more important than being "by the book"?

Issue #5416, opened 2024-09-02
Migrate regex linter rules to use parsed patterns


rzvxa:
Merging #5256 provides support for optional parsed regex literals, We can enhance our regex-related linter rules to use this parsed pattern whenever available.

Here is the list of the rules operating on regex patterns which are written before the introduction of regex parser.

- [x] no_div_regex(#5417)
- [x] no_control_regex
- [x] no_empty_character_class
- [x] no_regex_spaces
- [x] no_hex_escape
- [x] no_useless_escape
- [x] prefer_string_replace_all
- [x] prefer_string_starts_ends_with

Boshen:
I'll add "good first issue" when the required PRs are merged.

PR #5951, opened 2024-09-21
feat(regular_expression): add terms visitor to pattern

camchenry:
To make it easier to iterate over the nodes in the regex AST, I attempted to add a pseudo-visitor method for patterns

PR #5952, opened 2024-09-21
refactor(linter): use regex parser in `eslint/no-regex-spaces`

camchenry:
- part of https://github.com/oxc-project/oxc/issues/5416

Use the `oxc_regular_expression` parser to make these checks more robust. a few snapshots are updated because they now output more accurate diagnostics based on the regex AST. for example, `/   ?/` now correctly only highlights two spaces rather than three (because the last one is part of a quantifier)

### DIFF
--- a/crates/oxc_linter/src/snapshots/no_regex_spaces.snap
+++ b/crates/oxc_linter/src/snapshots/no_regex_spaces.snap
@@ -2,6 +2,13 @@
 source: crates/oxc_linter/src/tester.rs
 ---
   ⚠ eslint(no-regex-spaces): Spaces are hard to count.
+   ╭─[no_regex_spaces.tsx:1:12]
+ 1 │ var foo = /  /;
+   ·            ──
+   ╰────
+  help: Use a quantifier, e.g. {2}
+
+  ⚠ eslint(no-regex-spaces): Spaces are hard to count.
    ╭─[no_regex_spaces.tsx:1:15]
  1 │ var foo = /bar  baz/;
    ·               ──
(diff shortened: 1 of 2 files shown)


camchenry:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5952?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#5952** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5951** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_linter/src/snapshots/no_regex_spaces.snap
```diff
@@ -46,28 +53,28 @@ source: crates/oxc_linter/src/tester.rs
...
  1 │ var foo = /bar    ?baz/;
-   ·               ────
+   ·               ───
    ╰────
   help: Use a quantifier, e.g. {2}
 
   ⚠ eslint(no-regex-spaces): Spaces are hard to count.
    ╭─[no_regex_spaces.tsx:1:26]
  1 │ var foo = new RegExp('bar   *baz')
-   ·                          ───
```
DonIsaac, line 63: Are these spans incorrect? This smells like a bug in the regex parser.

camchenry: I think it's the opposite: these are actually more accurate now thanks to the parser. `/   ?/` is not equivalent to `/ {3}/`, as the diagnostic might suggest, since the last space is part of the quantifier. it is actually more like `/ {2,3}/` but that is beyond the scope of what the original lint rule targets.

camchenry:
I moved the visitor method code to be specific to this rule, to avoid the conversation of what a generic regex AST visitor might look like.

DonIsaac (approved):

PR #5968, opened 2024-09-22
perf(isolated-declarations): Combine type/value bindings and type/value references into one

Dunqing:


Dunqing:
* **#5968** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5968?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_isolated_declarations/src/scope.rs
```diff
@@ -51,48 +53,36 @@ impl<'a> ScopeTree<'a> {
...
-
-    fn add_value_binding(&mut self, ident: Atom<'a>) {
-        let scope = self.levels.last_mut().unwrap();
-        scope.value_bindings.insert(ident);
-    }
-
-    fn add_type_binding(&mut self, ident: Atom<'a>) {
-        let scope = self.levels.last_mut().unwrap();
-        scope.type_bindings.insert(ident);
+        let scope = self.levels.last().unwrap();
```
DonIsaac, line 56: What happens when two symbols with the same name are declared and merged?

```ts
interface Foo {}
namespace Foo {}
```

crates/oxc_isolated_declarations/src/scope.rs
```diff
@@ -8,27 +9,28 @@ use oxc_ast::AstBuilder;
 use oxc_ast::{visit::walk::*, Visit};
 use oxc_span::Atom;
 use oxc_syntax::scope::{ScopeFlags, ScopeId};
-use rustc_hash::FxHashSet;
+use rustc_hash::FxHashMap;
+
+bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct KindFlags: u8 {
```
DonIsaac, line 16: Could we use `ReferenceFlags`?

crates/oxc_isolated_declarations/src/scope.rs
Dunqing: In declarations, they don't merge, all of it will print. See https://www.typescriptlang.org/play/?target=99#code/JYOwLgpgTgZghgYwgAgGIHt3IN4F8BQIcAthAM4AOiKGWe++EAHhelGDmpsrkA

crates/oxc_isolated_declarations/src/scope.rs
Dunqing: Yes, but here we don't care about `read` and `write` so add this for understanding easily

PR #5976, opened 2024-09-22
test(semantic): add comprehensive regression test suite

DonIsaac:
# What This PR Does

Enhance's `oxc_semantic`'s integration tests with a regression test suite that ensures semantic's contract guarantees hold over all test cases in typescript-eslint's scope snapshot tests. Each test case checks a separate assumption and runs independently from other test cases.

This PR sets up the code infrastructure for this test suite and adds two test cases to start us off:
1. Reflexivity tests for `IdentifierReference` and `Reference`
2. Symbol declaration reflexivity tests between declarations in `SymbolTable` and their corresponding node in the AST.

Please refer to the doc comments for each of these tests for an in-depth explanation.

## Aren't our existing tests sufficient?
`oxc_semantic` is currently tested directly via
1. scope snapshot tests, ported from `typescript-eslint`
2. Hand-written tests using `SemanticTester` in `tests/integration`

And indirectly via

3. Conformance test suite over Test262/TypeScript/Babel
4. Linter snapshot tests

Shouldn't this be sufficient? I argue not, for two reasons:

## 1. Clarify Contract Ambiguity

When using `Semantic`, I often find myself asking these questions?
* Does `semantic.symbols().get_declaration(id)` point to a `BindingIdentifer`/`BindingPattern` or the declaration that holds an identifier/pattern?
* Will a `Reference`'s `node_id` point me to an `IdentifierReference` or the expression/statement that is holding an `IdentifierReference`?
* When will `BindingIdentifier`'s `symbol_id` get populated? can we guarantee that after semantic analysis it will never be `None`?
* What actually _is_ the node covered by `semantic.symbols().get_span(id)`? This one really messed me up, and resulted in me creating #4739.
* What scope does `Function::scope_id` point to? The one where the function is declared? The one created by its body? The one created by the type annotations but before the function body? Or something else entirely?

**These test cases are meant to answer such questions and guarnatee those answers as test cases**. No other existing testing solution currently upholds such promises: they only tell us if code expecting one answer or another produces an unexpected result. However, those parts of the codebase could always be adjusted to conform to new `Semantic` behavior, meaning no contract guarantees are actually upheld.

## 2. Existing Tests Do Not Test The Same Behavior

I'll cover each above listed test case one-by-one:

1. For starters, these tests only cover scopes. Additionally, they only tell us **how behavior has changed**, not that **behavior is now incorrect**.
2. These _do_ generally cover the same behaviors, but **are not comprehensive and are difficult to maintain**. These are unit tests that should be used hand-in-hand with this new test suite.
3. The most relevant tests here are for the parser. However, these tests **only tell us if a syntax/parse error was produced**, and tell us nothing about the validity of `Semantic`.
4. Relying on lint rule's output is a a mediiocre proxy of `Semantic`'s behavior at best. They can tell us if changes to `Semantic` break assumptions made by lint rules, but they do not tell us if **those assumptions are the ones we want to uphold to external crates consuming `Semantic`.

### DIFF
--- a/crates/oxc_semantic/tests/conformance/test_symbol_declaration.rs
+++ b/crates/oxc_semantic/tests/conformance/test_symbol_declaration.rs
@@ -1,0 +1,130 @@
+use oxc_ast::ast::BindingPattern;
+use oxc_ast::{ast::BindingIdentifier, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_span::{GetSpan, Span};
+use oxc_syntax::symbol::SymbolId;
+
+use super::{ConformanceTest, TestResult};
+use crate::Semantic;
+
+/// Verifies that symbol binding relationships between the SymbolTable and AST nodes are reflexive.
+///
+/// What does this mean?
+/// 1. [`SymbolTable`] stores the AST node id of the node declaring a symbol.
+/// 2. That symbol should _always_ be a declaration-like node containing either a
+///    [`BindingIdentifier`] or a [`BindingPattern`].
+/// 3. The binding pattern or identifier in that node should be populated (e.g. not [`None`]) and
+///    contain the symbol id.
+///
+/// [`SymbolTable`]: oxc_semantic::SymbolTable
+#[derive(Debug, Clone, Default)]
+pub struct SymbolDeclarationTest;
+
+/// The binding pattern or identifier contained in the declaration node is [`None`].
+///
+/// See: [`BindingIdentifier::symbol_id`]
+fn bound_to_statement_with_no_binding_identifier(
+    symbol_id: SymbolId,
+    span: Span,
+    statement_kind: &str,
+) -> TestResult {
+    OxcDiagnostic::error(format!(
+        "Symbol {symbol_id:?} got bound to a {statement_kind} with no BindingIdentifier"
+    ))
+    .with_label(span.label("Symbol was declared here"))
+    .into()
+}
+
+/// [`BindingIdentifier::symbol_id`] contained [`Some`] value, but it was not the [`SymbolId`] used
+/// to find it in the [`SymbolTable`].
+fn symbol_declaration_not_in_ast_node(
+    expected_id: SymbolId,
+    binding: &BindingIdentifier,
+) -> TestResult {
+    let bound_id = binding.symbol_id.get();
+    OxcDiagnostic::error(format!(
+        "Expected binding to be bound to {expected_id:?} but it was bound to {bound_id:?}"
+    ))
+    .with_label(binding.span())
+    .into()
+}
+
+/// Found a non-destructuring [`BindingPattern`] that did not contain a [`BindingIdentifier`].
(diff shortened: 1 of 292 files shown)


DonIsaac:
* **#5976** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen (commented):
I'll let @overlookmotel to decide on this one.

overlookmotel:
Thanks very much for writing up the rationale for this change so clearly. I totally agree with the need for a more structured approach to testing, for all the reasons you have given. And I too want to know what `semantic.symbols().get_declaration(id)` should be pointing to!

What I'm not so keen on is storing conformance test results in each snapshot. I am worried that if details of test fails are hidden away there, they may never get fixed. I can imagine a scenario like this: someone makes a PR that breaks a test. They say "I'll fix that test in a follow-up". But they forget to do it, and because it's hidden away in one `.snap` file of 290, no-one notices and the bug lives forever.

So personally, I think it'd be preferable if all tests fails were listed in a single `.snap.md` file.

crates/oxc_semantic/tests/conformance/test_symbol_declaration.rs
```diff
@@ -0,0 +1,108 @@
...
+            AstKind::Function(func) => match func.id.as_ref() {
+                Some(id) => check_binding(symbol_id, id),
+                None => bound_to_statement_with_no_binding_identifier(symbol_id, span, "Function"),
+            },
+            AstKind::Class(class) => match class.id.as_ref() {
+                Some(id) => check_binding(symbol_id, id),
+                None => bound_to_statement_with_no_binding_identifier(symbol_id, span, "Class"),
+            },
+            AstKind::FormalParameter(param) => check_binding_pattern(symbol_id, &param.pattern),
+            _ => TestResult::Pass,
```
overlookmotel, line 85: Shouldn't this be a test fail not pass?

DonIsaac: There are other declarations I didn't get around to. I'll add them.

DonIsaac:
> So personally, I think it'd be preferable if all tests fails were listed in a single `.snap.md` file.

I've looked into this, and making this change would require modifying how scope snapshot tests are run. I'd rather make changes to those tests in a separate PR. Is this a hard blocker, or can it be deferred to an up-stack PR?

overlookmotel:
> Is this a hard blocker, or can it be deferred to an up-stack PR?

Not a blocker to merging, as far as I'm concerned. But personally I would like to see it happen!

DonIsaac:
> > Is this a hard blocker, or can it be deferred to an up-stack PR?
> 
> Not a blocker to merging, as far as I'm concerned. But personally I would like to see it happen!

Ok, I'll definitely add this in a follow-up PR. Are there any other blockers I should address before we can merge this PR?

PR #5993, opened 2024-09-23
perf(linter, prettier, diagnostics): use `FxHashMap` instead of `std::collections::HashMap`

camchenry:
Using `FxHashMap` is faster than `HashMap` in many cases, especially for hashing-heavy workloads. This change improves the performance of the linter, prettier, and diagnostics crates by using `FxHashMap` instead of `std::collections::HashMap`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1572,6 +1572,7 @@ version = "0.30.0"
 dependencies = [
  "miette",
  "owo-colors",
+ "rustc-hash",
  "textwrap",
  "unicode-width",
 ]
@@ -1796,6 +1797,7 @@ dependencies = [
  "oxc_span",
  "oxc_syntax",
  "pico-args",
+ "rustc-hash",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


camchenry:
* **#5993** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/5993?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camchenry:
The -1% changes appear to be noise related to `ModuleRecord`. However, the +1% in `transformer` appears to be genuine: 
image

The +1% in linter seems uncertain, but doesn't look like it hurts performance at all. I think this is worth it, just for the +1% in transformer.

Boshen:
Oh my! How did these go under the radar? Would you like to try banning these in https://github.com/oxc-project/oxc/blob/f5eee723fa4c6becb4e170b1a99f49813f302ff8/.clippy.toml#L3?

camchenry:
@Boshen yeah, was going to ask about that! I can do that

Issue #5982, opened 2024-09-22
parser: `?` after a type should produce a parse error


DonIsaac:
This snippet should produce a parse error, but it instead produces a `TSJsdocType`
* [Oxc Playground](https://playground.oxc.rs/#eNpFjtEKwyAMRX9F8rRBv0Do9rb/sG0cgiYl2q5S/Pcpw+4p9yTh3nvCDBpSXlEdalQ3y6wVbWFCed7V+FA7uwUGYNAnyEZtxEzJHKCTbDiAd5S6jjOveEEOE/tOSQxFyxJAW+MjlgFWIxGlOlbdXLq+Xvti5gXf2LIrBEfOuv+JkrB/ef60ajvKxLFW+GWU8gUIDU3q)
* [TypeScript Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAHlC8UAUAzA9mgXFAdgVwFsAjCAJwH4BKBAPigDc0BLAEyA)
```ts
type x = (foo: number?) => void
```

IWANABETHATGUY:
this case is also invalid, 
```
type x = (foo: number!) => void
```
[ts](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAHlC8UAUAzA9mgXFAdgVwFsAjCAJwEIBKBAPigDc0BLAEwChRIoBBBXQkqQDcHcNAAqfANowA-Nm4BdIUA)
[oxc](https://playground.oxc.rs/#eNpFjkEOgzAMBL+S+tRKvCASPfYVvQRwqkiJjZxAQSh/b6IqcPKObe3uASNoSPuMalO9ultmrWgJA8rtofqnWtlNb4IOGPQBslAdcadkNtBJFuzAO0pNx5FnPGEPA/tGSQxFyxJAW+Mj5g5mIxGlOBZdXZo+X9ti5Ak/WLMLBEfOuutESdi/PH9rtRVl4Fgq/DNy/gGcY06W)

Boshen:
The AST is correct. Expect we need to produce an error:

```
    function checkJSDocTypeIsInJsFile(node: Node): void {
        if (!isInJSFile(node)) {
            if (isJSDocNonNullableType(node) || isJSDocNullableType(node)) {
                const token = tokenToString(isJSDocNonNullableType(node) ? SyntaxKind.ExclamationToken : SyntaxKind.QuestionToken);
                const diagnostic = node.postfix
[...]

PR #5994, opened 2024-09-23
fix(oxc_parser): only allow question after type when in none named `tupleElement` context

IWANABETHATGUY:
Part of #5982

### DIFF
--- a/crates/oxc_parser/src/ts/types.rs
+++ b/crates/oxc_parser/src/ts/types.rs
@@ -1,3 +1,4 @@
+use bitflags::Flags;
 use oxc_allocator::{Box, Vec};
 use oxc_ast::{ast::*, NONE};
 use oxc_diagnostics::Result;
@@ -277,6 +278,7 @@ impl<'a> ParserImpl<'a> {
     fn parse_postfix_type_or_higher(&mut self) -> Result<TSType<'a>> {
         let span = self.start_span();
         let mut ty = self.parse_non_array_type()?;
+        // dbg!(&ty);
 
         while !self.cur_token().is_on_new_line {
             match self.cur_kind() {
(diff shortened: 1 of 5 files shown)


crates/oxc_parser/src/ts/types.rs
```diff
@@ -353,7 +353,9 @@ impl<'a> ParserImpl<'a> {
             // // falls through
             // case SyntaxKind.FunctionKeyword:
             // return parseJSDocFunctionType();
-            Kind::Question => self.parse_js_doc_unknown_or_nullable_type(),
+            Kind::Question if self.ctx.contains(Context::AllowQuestionMarkAfterType) => {
```
DonIsaac, line 356: Could we parse the question mark and report an error instead of panicking? It would be nice to display a descriptive error message. [Example](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKlC8UDaAPAXFAdgVwLYCMIAnAfgF0BuIA) from TypeScript:

> A labeled tuple element is declared as optional with a question mark after the name and before the colon, rather than after the type. (5086)

Boshen (commented):
I don't think this is the correct fix, let me check tsc.

tasks/coverage/snapshots/parser_typescript.snap
```diff
@@ -10204,6 +10201,13 @@ Expect to Parse: tasks/coverage/typescript/tests/cases/conformance/salsa/private
    ╰────
   help: Try insert a semicolon here
 
+  × Unexpected token
+   ╭─[typescript/tests/cases/compiler/parseInvalidNullableTypes.ts:1:30]
+ 1 │ function f1(a: string): a is ?string {
```
DonIsaac, line 10206: Regarding the above comment, [TypeScript provides this error message](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMAjACgIYC5EGcoBOMYA5gJQ4aIy6ID8+RpiA3gFCKeIECmUIBJIRA82AXyA):

> '?' at the start of a type is not valid TypeScript syntax. Did you mean to write 'string | null | undefined'? (17020)

PR #6003, opened 2024-09-23
feat(parser)!: use `BindingIdentifier` for `namespace` declaration names

DonIsaac:
Use `BindingIdentifier` instead of `IdentifierName` so that AST visitors can access the bound symbol id for the namespace's name. This makes namespace consistent with other named declarations, such as `Class`, `Function`, and `TSInterfaceDeclaration`.

We should consider modifying `TSModuleDeclarationName::StringLiteral(...)` to also store a `symbol_id`, since one gets bound in semantic for it as well.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -1391,7 +1391,7 @@ pub enum TSModuleDeclarationKind {
 #[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
 #[serde(untagged)]
 pub enum TSModuleDeclarationName<'a> {
-    Identifier(IdentifierName<'a>) = 0,
+    Identifier(BindingIdentifier<'a>) = 0,
     StringLiteral(StringLiteral<'a>) = 1,
 }
 
--- a/crates/oxc_ast/src/generated/assert_layouts.rs
+++ b/crates/oxc_ast/src/generated/assert_layouts.rs
@@ -1091,19 +1091,19 @@ const _: () = {
     assert!(size_of::<TSTypePredicateName>() == 16usize);
     assert!(align_of::<TSTypePredicateName>() == 8usize);
 
-    assert!(size_of::<TSModuleDeclaration>() == 64usize);
+    assert!(size_of::<TSModuleDeclaration>() == 72usize);
     assert!(align_of::<TSModuleDeclaration>() == 8usize);
     assert!(offset_of!(TSModuleDeclaration, span) == 0usize);
     assert!(offset_of!(TSModuleDeclaration, id) == 8usize);
-    assert!(offset_of!(TSModuleDeclaration, body) == 40usize);
-    assert!(offset_of!(TSModuleDeclaration, kind) == 56usize);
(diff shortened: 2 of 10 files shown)


DonIsaac:
* **#6004** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6004?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6003** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6003?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
Let me check the spec.

Boshen:
This is incorrect, it's not a binding identifier. Although the "global" fix is correct in the next PR.

DonIsaac:
How should we make a namespace's symbol is available from the AST?

Boshen:
> How should we make a namespace's symbol is available from the AST?

Can you elaborate?

DonIsaac:
> > How should we make a namespace's symbol is available from the AST?
> 
> Can you elaborate?

`BindingIdentifier` has a `symbol_id` property, but `IdentifierName` does not. In this snippet, a symbol `Foo` is bound and put into `SymbolTable`
```ts
namespace Foo {}
```
but since `TSModuleDeclarationName` is an `IdentifierReference`, `Foo`'s symbol id is not available in the AST.

Boshen:
> > > How should we make a namespace's symbol is available from the AST?
> > 
> > 
> > Can you elaborate?
> 
> `BindingIdentifier` has a `symbol_id` property, but `IdentifierName` does not. In this snippet, a symbol `Foo` is bound and put into `SymbolTable`
> 
> ```ts
> namespace Foo {}
> ```
> 
> but since `TSModuleDeclarationName` is an `IdentifierReference`, `Foo`'s symbol id is not available in the AST.

Let me double check the outdated tsc reference...

overlookmotel:
In my view, regardless of TS's spec, in our scheme, anything which has a record in the `SymbolTable` *must* have a `symbol_id` field. Without that field, our transform checker is not able to check the correctness of `SymbolTable`. I was planning to open a PR doing exactly this when I come to fix the scopes/symbols manipulation in TS transform.

I suppose we *could* add `symbol_id` field to `TSModuleDeclaration` instead, but in my view it's much simpler to make the `id` a `BindingIdentifier`, as this PR does.

So am re-opening this and, if you're willing @Boshen, would like to merge it.

Boshen:
I need proof and reference from tsc to confirm this is a BindingIdentifier.

DonIsaac:
@Boshen from section `10.1 Module Declarations` in the TypeScript v1.4 spec:

```
ModuleDeclaration:
  module IdentifierPath { ModuleBody }
IdentifierPath:
  Identifier
  IdentifierPath . Identifier
```

The TypeScript spec uses `Identifier` for all other declarations. For example, here is the grammar for `ClassDeclaration` in section `8.1 Class Declarations`:

```
ClassDeclaration:
  class Identifier TypeParametersopt ClassHeritage { ClassBody }
```

Our `Class` node uses a `BindingIdentifier`, meaning `Identifier` in TypeScript is the same as `BindingIdentifier` in oxc.

PR #6008, opened 2024-09-23
perf(linter): `no-fallthrough`: Use string matching instead of Regex for default comment pattern

camchenry:
Profiling has shown this rule to be a consistent slow point, and in particular, the Regex construction is slow.

Screenshot 2024-09-23 at 7 12 58 PM

This PR improves the situation in two ways:

1. A `Regex` is only constructed when there is a custom comment pattern (which is likely the minority of cases)
2. For the default comment pattern (when no custom pattern is passed), we now use a simple string matcher function, instead of a full-blown regex matcher. I believe this should be faster since it involves much less work, though can still allocate a `String`.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -1,5 +1,6 @@
 use std::ops::Range;
 
+use cow_utils::CowUtils;
 use itertools::Itertools;
 use oxc_ast::{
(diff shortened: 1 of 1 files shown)


camchenry:
* **#6008** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6008?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

PR #6009, opened 2024-09-24
perf(linter): `jest/no-commented-out-tests`: lazily initialize comment Regex

camchenry:
Profiling shows that this rule spends essentially all of its time just constructing the Regex and doing matching. The hypothesis here is that if we can return early and lazily initialize the Regex, then in most cases this rule will be able to do almost nothing.

Screenshot 2024-09-23 at 8 27 11 PM

### DIFF
--- a/crates/oxc_linter/src/rules/jest/no_commented_out_tests.rs
+++ b/crates/oxc_linter/src/rules/jest/no_commented_out_tests.rs
@@ -1,4 +1,5 @@
 use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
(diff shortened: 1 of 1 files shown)


camchenry:
* **#6009** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac:
Benchmarks won't reflect any improvements since jest rules only run on relevant files (based on their path). Local benchmarks face the same issue. Can you compare performance on a local repository using Jest?

camchenry:
doesn't look like this was worth it in reality, there was no significant difference when benchmarking. I'm guessing that the lazy initialization was still initialized often enough it didn't really make any difference (maybe even slower?).

image

PR #6030, opened 2024-09-24
perf(linter): `jest/prefer-hooks-in-order`: rewrite rule to allocate less and iterate fewer times

camchenry:
Profiling this in a single-threaded run showed this rule to be relatively slow compared to other rules:

image

The profile shows that a lot of time is spent doing pushes on `Vec`: 
image

I believe this is because we were essentially duplicating the entirety of `ctx.nodes()`, and then iterating over it again for the actual rule.

I addressed this by:

- We no longer store any nodes (we only store the previous hook span + order), which saves many allocations.
- We only iterate over the nodes once and store the previous hook function in a per-scope hash map. This should be faster since we now only do `n` iterations, instead of `2n` iterations.

Benchmarking on the `microsoft/vscode` repository shows that this is probably faster (by up to ~3%), and probably not any slower:

Screenshot 2024-09-24 at 12 06 32 PM

The snapshot change in this PR is due to slightly different ordering, I think, but it appears to be the same diagnostics overall.

### DIFF
--- a/crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
+++ b/crates/oxc_linter/src/rules/jest/prefer_hooks_in_order.rs
@@ -1,7 +1,7 @@
-use oxc_ast::{ast::CallExpression, AstKind};
+use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_semantic::{AstNode, ScopeId};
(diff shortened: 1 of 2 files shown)


camchenry:
* **#6030** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

PR #6033, opened 2024-09-24
perf(linter): `no-magic-numbers` remove redudant checks in `is_array_index`

Sysix:
- there can not be an `UnaryExpression` as a `parent_kind`, when the `ast_kind` is  `NumericLiteral`,  
  `InternConfig::from` prevents it from happending
- `NumericLiteral` can not be negative or else they would be passed as `UnaryExpression`


camchenry (approved):

camchenry:
Seems like a pretty straightforward change to me!

DonIsaac (approved):

PR #6040, opened 2024-09-24
fix(semantic): do not create a `global` symbol for `declare global {}`

DonIsaac:
Re-creation of #6004 to unblock it from down-stack PRs. @Boshen has already reviewed the previous PR and determined this to be correct behavior.


Dunqing (approved):
LGTM

PR #6055, opened 2024-09-25
feat(regular_expression): implement visitor pattern trait for regex AST

### LINKED ISSUE #5977: regular_expression: implement visitor pattern for regex AST
With the [regex parser](https://github.com/oxc-project/oxc/issues/1164) completed, we have now started using `oxc_regular_expression` in the linter for parsing patterns instead of handwritten pattern matching and parsing code.

However, it is difficult to traverse the regex AST and visit each node (term/disjunction/alternative/etc.). We should implement the visitor pattern for the regex AST, so it is easier to write some code that visits every node.

Ideally, we should be able to replace code like this:

https://github.com/oxc-project/oxc/blob/74d8714d5a53ac60f46f0144044dbe6959172d6e/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs#L119-L146

https://github.com/oxc-project/oxc/blob/74d8714d5a53ac60f46f0144044dbe6959172d6e/crates/oxc_linter/src/rules/eslint/no_regex_spaces.rs#L171-L206

camchenry:
- resolves https://github.com/oxc-project/oxc/issues/5977
- supersedes https://github.com/oxc-project/oxc/pull/5951

To facilitate easier traversal of the Regex AST, this PR defines a `Visit` trait with default implementations that will walk the entirety of the Regex AST. Methods in the `Visit` trait can be overridden with custom implementations to do things like analyzing only certain nodes in a regular expression, which will be useful for regex-related `oxc_linter` rules.

In the future, we should consider automatically generating this code as it is very repetitive, but for now a handwritten visitor is sufficient.

### DIFF
--- a/crates/oxc_regular_expression/src/visit.rs
+++ b/crates/oxc_regular_expression/src/visit.rs
@@ -1,0 +1,389 @@
+#![allow(unused_variables, clippy::wildcard_imports)]
+use walk::walk_pattern;
+
+use crate::ast::{
+    Alternative, BoundaryAssertion, CapturingGroup, Character, CharacterClass,
+    CharacterClassContents, CharacterClassEscape, CharacterClassRange, ClassString,
+    ClassStringDisjunction, Disjunction, Dot, IgnoreGroup, IndexedReference, LookAroundAssertion,
+    NamedReference, Pattern, Quantifier, Term, UnicodePropertyEscape,
+};
+use walk::*;
+
+#[derive(Copy, Clone, Debug)]
+pub enum RegExpAstKind<'a> {
+    Pattern(&'a Pattern<'a>),
+    Disjunction(&'a Disjunction<'a>),
+    Alternative(&'a Alternative<'a>),
+    Term(&'a Term<'a>),
+    LookAroundAssertion(&'a LookAroundAssertion<'a>),
+    Quantifier(&'a Quantifier<'a>),
+    CapturingGroup(&'a CapturingGroup<'a>),
+    IgnoreGroup(&'a IgnoreGroup<'a>),
+    BoundaryAssertion(&'a BoundaryAssertion),
+    Character(&'a Character),
+    Dot(&'a Dot),
+    CharacterClassEscape(&'a CharacterClassEscape),
+    UnicodePropertyEscape(&'a UnicodePropertyEscape<'a>),
+    CharacterClass(&'a CharacterClass<'a>),
+    CharacterClassContents(&'a CharacterClassContents<'a>),
+    CharacterClassRange(&'a CharacterClassRange),
+    CharacterClassStringDisjunction(&'a ClassStringDisjunction<'a>),
+    CharacterClassString(&'a ClassString<'a>),
+    IndexedReference(&'a IndexedReference),
+    NamedReference(&'a NamedReference<'a>),
+}
+
+/// RegEx syntax tree traversal
+pub trait Visit<'a>: Sized {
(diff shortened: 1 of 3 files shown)


camchenry:
* **#6055** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6055?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_regular_expression/src/visit.rs
```diff
@@ -0,0 +1,391 @@
...
+use crate::ast::{
+    Alternative, BoundaryAssertion, CapturingGroup, Character, CharacterClass,
+    CharacterClassContents, CharacterClassEscape, CharacterClassRange, ClassString,
+    ClassStringDisjunction, Disjunction, Dot, IgnoreGroup, IndexedReference, LookAroundAssertion,
+    NamedReference, Pattern, Quantifier, Term, UnicodePropertyEscape,
+};
+use walk::*;
+
+#[derive(Copy, Clone, Debug)]
+pub enum RegexAstKind<'a> {
```
leaysgur, line 13: IMO: I feel it might be better to avoid using the term `Regex`. It brings to mind the `regex` crate. 👀 

Although it's a bit verbose, how about `RegularExpressionAstKind`?

crates/oxc_regular_expression/src/visit.rs
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    #[inline]
+    fn visit_named_reference(&mut self, it: &NamedReference<'a>) {
+        walk_named_reference(self, it);
+    }
+}
+
+pub mod walk {
+    use crate::ast::UnicodePropertyEscape;
```
leaysgur, line 153: Seems unnecessary?

Boshen (approved):

crates/oxc_regular_expression/src/visit.rs
camchenry: hm, I'm not strongly opposed to `RegularExpressionAstKind`, but it is kind of long. maybe to differentiate from the `regex` crate, we could use the term `RegExp` instead? so it would be `RegExpAstKind` instead. That is more similar to how the ECMA262 spec refers to it.

leaysgur: Personally, I still prefer `RegularExpressionAstKind` (since it's the `AstKind` of `oxc_regular_expression`), but I think `RegExpAstKind` would also be fine. 👍🏻 

---

BTW, (feel free to ignore this if I'm wrong, 😓) is it possible to alias an export with the same name in Rust?

```rs
use oxc_ast::AstKind;
use oxc_regular_expression::AstKind as RegExpAstKind; // 👈🏻 
```

If it's possible, I’m starting to think that simply leaving it as `AstKind` might be a good option.

leaysgur (approved):
Thank you! 🙌🏻

crates/oxc_regular_expression/src/visit.rs
camchenry: I think that's definitely possible, though it's less ideal for the linter codebase since it's pretty common to import AstKind too. Let's stick with this name for now and we can always change it later?

leaysgur: > that's definitely possible

Thanks, I learned something new. 😃 

> Let's stick with this name for now

🙆🏻

PR #6083, opened 2024-09-26
Draft: Quick and dirty node counting in the parser

DavidHancu:
This is a quick and dirty implementation for counting nodes in the parser instead of a separate `oxc_semantic` pass. Currently, this draft PR is not in a production-ready state, introducing printing statements in the AST builder to debug the node incrementing and reconciliation.

The debugging statements have this form:

```<function>: nodes<++|~~>```

- The `<function>` tag represents the function name in the builder.
- If the node count is being incremented, `nodes` will be followed by `++`.
- If the node count is not being incremented (being synced due to the parser rewinding or throwing tokens away), `nodes` will be followed by `~~` indicating a reconciliation.


overlookmotel:
I've had a look. It's quite hard to see what's going on in this PR as there seem to be a bunch of unrelated changes (related to comments in codegen, and stuff to do with isolated declarations). Has something gone wrong while updating to sit on top of latest main?

But a few comments based on what I can see:

1. I don't think `Stats` should need to be passed in to `Parser::new`. Instead parser can generate a `Stats` object internally, and return it as a property of `ParserReturn`.
2. I don't think it's necessary for `Stats`'s fields to be `Cell`s. In parser, it can just be owned by `AstBuilder`. `AstBuilder` is owned by `ParserImpl`, which parser methods receive as `&mut self`. So `AstBuilder`'s methods can require a `&mut self` and then mutate `Stats`'s fields without using `Cell`.
3. `AstBuilder` will have to cease to be `Clone` and `Copy`, as it's now stateful.
4. `Stats` can be `Copy`. It's only 16 bytes.

Since you submitted this PR, there's been changes to transformer, so it now only holds a single instance of `AstBuilder` (#6209). This should enable `AstBuilder` being stateful (without `Cell`s) in the transformer too now.

I know you're busy, so if you don't have time for this, please feel free to say so and I can take this over when I have time. If you have time to at least clean up the (what I assume is) extraneous changes which have got in via some git accident, that'd be a big help - you know better than me what changes were intentional, and which accidental.

DavidHancu:
I'll rewrite it using the newest changes and also clean the PR

PR #6118, opened 2024-09-27
refactor(transformer): share `TransformCtx` with ref not `Rc`

overlookmotel:
Many transforms share `TransformCtx`. Currently it's shared with `Rc`, which has a cost as the `Rc` has to be cloned many times, and it also makes dropping `Transformer` more expensive.

The PR changes that to share it as a normal reference `&TransformCtx` instead.

This requires adding an inner `TransformerImpl`. `Transformer` is now just a facade which creates the `TransformCtx` and stores options. `Transformer::build_with_symbols_and_scopes` constructs `TransformerImpl` and runs the visitor on it.

Unlikely to have any perf impact on larger files, but for small files where setup/teardown is a larger % of the overall workload, it may help a little.

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -30,7 +30,7 @@ mod helpers {
     pub mod stack;
 }
 
-use std::{path::Path, rc::Rc};
+use std::path::Path;
 
 use es2016::ES2016;
 use es2018::ES2018;
@@ -53,12 +53,7 @@ pub use crate::{
     react::{ReactJsxRuntime, ReactOptions, ReactRefreshOptions},
     typescript::{RewriteExtensionsMode, TypeScriptOptions},
 };
-use crate::{
-    context::{Ctx, TransformCtx},
-    es2015::ES2015,
-    react::React,
-    typescript::TypeScript,
-};
+use crate::{context::TransformCtx, es2015::ES2015, react::React, typescript::TypeScript};
 
 pub struct TransformerReturn {
     pub errors: std::vec::Vec<OxcDiagnostic>,
@@ -67,17 +62,8 @@ pub struct TransformerReturn {
 }
 
 pub struct Transformer<'a> {
-    ctx: Ctx<'a>,
-    // NOTE: all callbacks must run in order.
-    x0_typescript: TypeScript<'a>,
-    x1_react: React<'a>,
-    x2_es2021: ES2021<'a>,
-    x2_es2020: ES2020<'a>,
-    x2_es2019: ES2019,
-    x2_es2018: ES2018,
-    x2_es2016: ES2016<'a>,
-    x3_es2015: ES2015<'a>,
-    x4_regexp: RegExp<'a>,
+    ctx: TransformCtx<'a>,
+    options: TransformOptions,
 }
 
 impl<'a> Transformer<'a> {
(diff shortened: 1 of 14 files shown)


overlookmotel:
* **#6118** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6118?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/lib.rs
```diff
@@ -89,41 +75,50 @@ impl<'a> Transformer<'a> {
...
+            x4_regexp: RegExp::new(self.options.regexp, &self.ctx),
+        };
+
+        let (symbols, scopes) = traverse_mut(&mut transformer, allocator, program, symbols, scopes);
         TransformerReturn { errors: self.ctx.take_errors(), symbols, scopes }
     }
 }
 
-impl<'a> Traverse<'a> for Transformer<'a> {
+struct TransformerImpl<'a, 'c> {
```
Boshen, line 108: 🤮 Should we use `'ast` and `'ctx` just like some other hardcore projects?

overlookmotel: I've pushed a commit to change `'c` to `'ctx`.

I'm less sure about swapping `'a` to `'ast`. We have a convention that `'a` means AST, and `'a` lifetime is used everywhere. I assume we don't want to change that to `'ast` everywhere, right?

overlookmotel: The extra lifetime is unfortunate.

But from what I can see it is having a significant effect on the `RadixUIAdoptionSection.jsx` benchmark. `Transformer::new` is x2.5 faster.

The odd thing is that other changes offset that improvement. From the flamegraph, it looks like `visit_expression` doesn't get inlined any more, and that's what causes the slowdown. But I think that's a random fluke, and likely we could reverse that by just putting `#[inline]` on it.

About 10-15% of the total time on that benchmark is in setup and teardown, and I think it's worthwhile to chop that down.

So... the question is... are the lifetimes worth it for the perf improvement? Personally, I think yes. But you may have a different view.

overlookmotel: Actually, I *think* that if we changed the API so you pass `program` and `allocator` into `build_with_symbols_and_scopes`, instead of into `new`, then we'd be borrowing everything at the same point in time, and we'd only need one lifetime. We could get rid of `'ctx` again.

I know you were thinking of getting rid of the builder pattern, but I don't know where you go to in your thinking on that.

Boshen: > I don't know where you go to in your thinking on that

I'm on the edge of hating Rust :-)

overlookmotel: > I'm on the edge of hating Rust :-)

When you start hating rust, god help us all! You need some chill time, good man.

overlookmotel: Just to say, the perf boost on this PR was all down to adding one `#[inline]` on `visit_expression`, nothing to do with getting rid of the `Rc`s! Adding `#[inline]` was just an experiment, I would have done it in a separate PR so could see what getting rid of the `Rc`s did in isolation.

overlookmotel:
Just for the record, the large perf boost in this PR is not really related to the main thrust of the PR. It's all down to an unrelated change which got into this PR - adding `#[inline]` to `enter_expression`.

The change of replacing `Rc` with `&` doesn't affect any of the benchmarks for large files, but it looks like it gave a 1%-2% speed-up on the small `RadixUIAdoptionSection.jsx` benchmark - because setup and teardown is a larger proportion of the overall workload for small files.

PR #6121, opened 2024-09-27
refactor(transformer): share `TypeScriptOptions` with ref not `Rc`

overlookmotel:
Similar to #6118.

Share `TypeScriptOptions` between transforms as a plain `&` reference, rather than an `Rc`, to reduce setup/teardown time.

The code to parse pragmas from comments is moved into `Transformer::new`. This isn't the ideal place for it, but it means `TypeScriptOptions` can be shared with the existing `'ctx` lifetime, rather than having to have a 3rd lifetime `TypeScript<'a, 'ctx, 'options>`.

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -73,10 +73,13 @@ impl<'a> Transformer<'a> {
         source_type: SourceType,
         source_text: &'a str,
         trivias: Trivias,
-        options: TransformOptions,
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#6121** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6121?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6130, opened 2024-09-27
refactor(linter): consolidate file loading logic

DonIsaac:
# Human Description
Low on time, so this one is short.
- consolidate source file and partial loader logic into `loader` module. I have more plans for this.
- ~LSP no longer uses `VALID_EXTENSIONS`, so now `.d.ts` files (and the like) will be linted as well~ LSP does not respect `.gitignore` files, so this change was reverted.

# AI Description
## Refactor Loader and Partial Loader

This PR refactors the loader and partial loader functionality in the oxc_linter crate:

* Introduce a new `Loader` struct with methods for checking if a file can be loaded and loading file contents
* Move `partial_loader` module to `loader/partial_loader`
* Rename `JavaScriptSource` to `source.rs` and move it to the `loader` module
* Update `JavaScriptSource` to use `u32` for `start` offset instead of `usize`
* Refactor `IsolatedLintHandler` to use the new `Loader`
* Update imports and module references throughout the codebase

This change improves the organization of the loader-related code and provides a more unified interface for loading different file types.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1621,6 +1621,7 @@ dependencies = [
  "oxc_semantic",
  "oxc_span",
  "ropey",
+ "rustc-hash",
  "serde",
  "serde_json",
  "tokio",
--- a/apps/oxlint/src/lint/mod.rs
+++ b/apps/oxlint/src/lint/mod.rs
@@ -3,8 +3,8 @@ use std::{env, io::BufWriter, time::Instant};
 use ignore::gitignore::Gitignore;
 use oxc_diagnostics::{DiagnosticService, GraphicalReportHandler};
 use oxc_linter::{
-    partial_loader::LINT_PARTIAL_LOADER_EXT, AllowWarnDeny, InvalidFilterKind, LintFilter,
-    LintService, LintServiceOptions, Linter, OxlintOptions,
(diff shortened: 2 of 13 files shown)


DonIsaac:
* **#6130** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6130?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac:
@Boshen I want to use `oxc_resolver`'s FS abstraction so that we can plug the linter right into VSCode's file API. This will also help with integrating with Monaco. This cant be done unless `FileSystemOs` is exposed publically.

Boshen (commented):
> so now .d.ts files (and the like) will be linted as well

We should never lint `.d.ts` files from my experience, they are never hand written.

qweered:
I heard that some companies write them by hand

DonIsaac:
@Boshen this change (linting `.d.ts`) files only affects the LSP, not oxlint. If a user is opening a file in their editor, and it's not already ignored by version control, it's more than likely to be a source file.

Now I need to go back and check that `oxc_language_server` respects .gitignores...

DonIsaac:
Update: The LSP does _not_ respect gitignores, so I've re-added the `VALID_EXTENSIONS` check. Note that it looks like the LSP also doesn't support .eslintignore files. We should add this in the future.

PR #6135, opened 2024-09-28
refactor(transformer): TS transforms only store options they need

overlookmotel:
All TS transforms only need a subset of the options in `TypeScriptOptions` - mostly just a single `bool`. Store these specific options in the transforms, not the whole `TypeScriptOptions` object.

This makes the transformer types smaller, and also checking the option is cheaper as it doesn't involve going through a double-reference.

e.g. Accessing `only_remove_type_imports` option with Rust's deref sugar made explicit:

Before: `*(&*self.options).only_remove_type_imports`
After: `*self.only_remove_type_imports`


overlookmotel:
* **#6135** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6144, opened 2024-09-28
feat(linter): implement `no-throw-literal`

shulaoda:
Related to #479


camc314 (approved):

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,197 @@
...
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_throw_literal_diagnostic(span: Span, is_undef: bool) -> OxcDiagnostic {
+    let message =
+        if is_undef { "Do not throw undefined." } else { "Expected an error object to be thrown." };
+
+    OxcDiagnostic::warn("Disallow throwing literals as exceptions")
```
DonIsaac, line 15: Your help message should be the warning message. Please refer to the diagnostics message part of the "adding a lint rule" guide

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,197 @@
...
+        } else if matches!(&stmt.argument, Expression::Identifier(id) if id.name == "undefined") {
+            ctx.diagnostic(no_throw_literal_diagnostic(stmt.span, true));
+        };
+    }
+}
+
+impl NoThrowLiteral {
+    fn could_be_error(expr: &Expression) -> bool {
+        match expr {
+            Expression::Identifier(_)
```
DonIsaac, line 90: Can we attempt to resolve this reference and see if it's really an error object?

crates/oxc_linter/src/snapshots/no_throw_literal.snap
```diff
@@ -0,0 +1,121 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint(no-throw-literal): Disallow throwing literals as exceptions
+   ╭─[no_throw_literal.tsx:1:1]
+ 1 │ throw 'error';
```
DonIsaac, line 6: Please make the span only cover the thrown expression

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,197 @@
...
+            ctx.diagnostic(no_throw_literal_diagnostic(stmt.span, false));
+        } else if matches!(&stmt.argument, Expression::Identifier(id) if id.name == "undefined") {
+            ctx.diagnostic(no_throw_literal_diagnostic(stmt.span, true));
+        };
+    }
+}
+
+impl NoThrowLiteral {
+    fn could_be_error(expr: &Expression) -> bool {
+        match expr {
```
DonIsaac, line 89: Consider using `get_inner_expr`. That might not actually be the name, I'm on my phone and cannot check

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,197 @@
...
+        "throw 'literal' && new Error();",
+        "throw new Error() || 'literal';",
+        "throw foo ? new Error() : 'literal';",
+        "throw foo ? 'literal' : new Error();",
+        "throw tag `${foo}`;", // { "ecmaVersion": 6 },
+        "function* foo() { var index = 0; throw yield index++; }", // { "ecmaVersion": 6 },
+        "async function foo() { throw await bar; }", // { "ecmaVersion": 8 },
+        "throw obj?.foo",      // { "ecmaVersion": 2020 },
+        "throw obj?.foo()",    // { "ecmaVersion": 2020 }
+    ];
```
DonIsaac, line 174: Please add some typescript examples to both pass and fail tests.

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// throw e;
+    ///
+    /// try {
+    ///     throw new Error("error");
+    /// } catch (e) {
+    ///     throw e;
+    /// }
+    /// ```
+    NoThrowLiteral,
+    correctness,
```
DonIsaac, line 70: todo: @camc314 we could add a conditional_suggestion when a string is being thrown

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
DonIsaac: https://oxc.rs/docs/contribute/linter/adding-rules.html#diagnostics

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
shulaoda: This seems a bit tricky. The only case worth discussing is in assignment scenarios, where certain things are not considered `Error` objects.

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,219 @@
...
+                        .get_expression()
+                        .map_or(true, |expr| Self::could_be_error(expr))
+                        || Self::could_be_error(&expr.right);
+                }
+
+                false
+            }
+            Expression::SequenceExpression(expr) => {
+                if expr.expressions.len() > 0 {
+                    return Self::could_be_error(expr.expressions.last().unwrap());
```
DonIsaac, line 137: Don't unwrap. Use `is_some_and`

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
```diff
@@ -0,0 +1,215 @@
...
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_throw_literal_diagnostic(span: Span, is_undef: bool) -> OxcDiagnostic {
+    let message =
+        if is_undef { "Do not throw undefined" } else { "Expected an error object to be thrown" };
+
+    OxcDiagnostic::warn(message).with_label(span)
```
camchenry, line 15: Could we add help text for this? Like:

- ```
  Wrap the value with an error object: `throw new Error(...)`
  ```
- or if `undefined`:
  ```
  throw an error object instead: `throw new Error()`
  ```

shulaoda: Currently, there are many rules that do not provide help, only help prompts when fixable.

> Wrap the value with an error object: `throw new Error(...)`

For fixable types like `string` and `template`, they automatically include help suggestions. We've already implemented fixes for some of them, and the remaining ones are largely unfixable, so this prompt might not be necessary.

> throw an error object instead: `throw new Error()`

For `undefined`, this should be a very intuitive error prompt in itself, and perhaps `help` is not necessary.


Is there a necessary requirement for adding `help`? Under what circumstances should we add it? @DonIsaac

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
DonIsaac: I've added it myself.

crates/oxc_linter/src/rules/eslint/no_throw_literal.rs
camchenry: consider this comment as non-blocking. I'm of the opinion that every warning/error should include at least _some_ indication of how to fix it. I think a static message here would be fine too:

```rust
.with_help("Throw an Error object instead: `throw new Error()`")
```

Some diagnostics don't contain help messages, I would consider that something we should fix. From our [Linter Rule principles](https://oxc.rs/docs/contribute/linter/adding-rules.html#diagnostics):

> - The `message` should be an imperative statement about what is wrong, not a description of what the rule does.
> - The `help` message should be a command-like statement that tells the user how to fix the issue.

PR #6148, opened 2024-09-28
refactor(codegen): add `CodeBuffer` to fix soundness hole

### LINKED ISSUE #6147: Soundness hole in `Codegen`
`Codegen::print_char`, a public API, can be used to construct invalid UTF-8 in safe code. When `Codegen::into_source_text` gets called, `String::from_utf8_unchecked`'s invariants will not be upheld, producing undefined behavior.

I do not think this bug is causing any problem in internal oxc crates. However, `Codegen` and both problematic methods are publically available to anyone using `oxc_codegen`, and this _could_ introduce unsound code into their projects.

Example:
```rust
use oxc::codegen::Codegen;

let mut code = Codegen::new();
code.push_char(0xFF); // past ASCII char boundary for single-byte UTF-8 code points
let invalid = code.into_source_string();
```

DonIsaac:
# What This PR Does

Adds `CodeBuffer`, a simple wrapper over a `Vec<u8>` with a protective and reduced API for upholding UTF-8 validity guarantees. Closes #6147.

Note that this struct is actually quite small. Most of the added lines are doc comments.

### DIFF
--- a/crates/oxc_codegen/src/code_buffer.rs
+++ b/crates/oxc_codegen/src/code_buffer.rs
@@ -1,0 +1,427 @@
+/// A string builder for constructing source code.
+///
+///
+/// `CodeBuffer` provides safe abstractions over a byte array, allowing for
+/// a compact byte-array representation without soundness holes.
+///
+/// Use one of the various `print_*` methods to add text into a buffer. When you
+/// are done, call [`take_source_text`] to extract the final [`String`].
+///
+/// # Examples
+/// ```
+/// use oxc_codegen::CodeBuffer;
+/// let mut code = CodeBuffer::new();
+///
+/// // mock settings
+/// let is_public = true;
+///
+/// if is_public {
+///     code.print_str("export ")
+/// }
+/// code.print_str("function foo() {\n");
+/// code.print_str("    console.log('Hello, world!');\n");
+/// code.print_str("}\n");
+///
+/// let source = code.take_source_text();
+/// ```
+///
+/// [`take_source_text`]: CodeBuffer::take_source_text
+#[derive(Debug, Default, Clone)]
+pub struct CodeBuffer {
+    /// INVARIANT: `buf` is a valid UTF-8 string.
+    buf: Vec<u8>,
+}
+
+impl CodeBuffer {
+    /// Create a new empty `CodeBuffer`.
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    ///
+    /// // use `code` to build new source text
+    /// code.print_str("fn main() { println!(\"Hello, world!\"); }");
+    /// let source_text = code.take_source_text();
+    /// ```
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Create a new, empty `CodeBuffer` with the specified capacity.
+    ///
+    /// The buffer will be able to hold at least `capacity` bytes without
+    /// reallocating. This method is allowed to allocate for more bytes than
+    /// `capacity`. If `capacity` is 0, the buffer will not allocate.
+    ///
+    /// It is important to note that although the returned buffer has the
+    /// minimum *capacity* specified, the buffer will have a zero *length*.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the new capacity exceeds `isize::MAX` _bytes_.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self { buf: Vec::with_capacity(capacity) }
+    }
+
+    /// Returns the number of bytes in this buffer.
+    ///
+    /// This is _not_ the same as the number of characters in the buffer, since
+    /// non-ASCII characters require multiple bytes.
+    pub fn len(&self) -> usize {
+        self.buf.len()
+    }
+
+    /// Returns `true` if this buffer contains no characters.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    /// assert!(code.is_empty());
+    ///
+    /// code.push_char('c');
+    /// assert!(!code.is_empty());
+    /// ```
+    pub fn is_empty(&self) -> bool {
+        self.buf.is_empty()
+    }
+
+    /// Reserves capacity for at least `additional` more characters in the given
+    /// `CodeBuffer`. The buffer may reserve more space to speculatively avoid
+    /// frequent reallocations. After calling `reserve`, capacity will be
+    /// greater than or equal to `self.len() + additional`. Does nothing if
+    /// capacity is already sufficient.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the new capacity exceeds `isize::MAX` _bytes_.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let mut code = CodeBuffer::default();
+    /// code.reserve(10);
+    /// ```
+    #[inline]
+    pub fn reserve(&mut self, additional: usize) {
+        self.buf.reserve(additional);
+    }
+
+    /// Peek the `n`th character from the end of the buffer.
+    /// When `n` is zero, the last character is returned. Returns [`None`] if
+    /// `n` exceeds the length of the buffer.
+    ///
+    /// ## Examples
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    /// code.print_str("foo");
+    ///
+    /// assert_eq!(code.peek_nth_back(0), Some('o'));
+    /// assert_eq!(code.peek_nth_back(2), Some('f'));
+    /// assert_eq!(code.peek_nth_back(3), None);
+    /// ```
+    #[inline]
+    #[must_use = "Peeking is pointless if the peeked char isn't used"]
+    pub fn peek_nth_back(&self, n: usize) -> Option<char> {
+        // SAFETY: `buf` is a valid UTF-8 string because of invariants upheld by CodeBuffer
+        unsafe { std::str::from_utf8_unchecked(&self.buf) }.chars().nth_back(n)
+    }
+
+    /// Push a single ASCII character into the buffer
+    ///
+    /// # Panics
+    /// If `ch` is not a valid UTF-8 code point in the ASCII range (`0 - 0x7F`).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    /// code.print_ascii_byte('f');
+    /// code.print_ascii_byte('o');
+    /// code.print_ascii_byte('o');
+    ///
+    /// let source = code.take_source_text();
+    /// assert_eq!(source, "foo");
+    /// ```
+    #[inline]
+    pub fn print_ascii_byte(&mut self, b: u8) {
+        // NOTE: since this method is inlined, this assertion should get
+        // optimized away by the compiler when the value of `b` is known,
+        // e.g. when printing a constant.
+        assert!(b.is_ascii(), "byte {b} is not ASCII");
+        self.buf.push(b);
+    }
+
+    /// Print a byte without checking that this buffer still represents a valid
+    /// UTF-8 string.
+    ///
+    /// If you are looking to print a byte you know is valid ASCII, prefer
+    /// [`print_ascii_byte`]. If you are not certain, you may use [`print_char`]
+    /// as a safe alternative.
+    ///
+    /// # Safety
+    /// The caller must ensure that, after 1 or more sequential calls, this
+    /// buffer represents a valid UTF-8 string.
+    ///
+    /// It is safe for a single call to temporarily result in invalid UTF-8, as
+    /// long as UTF-8 integrity is restored before calls to any other `print`
+    /// method or [`take_source_text`]. This lets you, for example, print an
+    /// 8-byte code point using 4 separate calls to this method.
+    ///
+    /// If you find yourself in such a scenario, consider using
+    /// [`print_unchecked`] instead.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    /// // Safe: 'a' is a valid ASCII character. Its UTF-8 representation only
+    /// // requires a single byte.
+    /// unsafe { code.print_byte_unsafe(b'a') };
+    ///
+    /// let not_ascii = '⚓';
+    /// let as_bytes = not_ascii.to_string().into_bytes();
+    /// // Safe: after this loop completes, `code` returns to a valid state.
+    /// for byte in as_bytes {
+    ///     unsafe { code.print_byte_unsafe(byte) };
+    /// }
+    ///
+    /// // NOT SAFE: `ch` exceeds the ASCII segment range. `code` is no longer
+    /// valid UTF-8
+    /// unsafe { code.print_byte_unsafe(0xFF) };
+    /// ```
+    ///
+    /// [`print_ascii_byte`]: CodeBuffer::print_ascii_byte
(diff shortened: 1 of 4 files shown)


DonIsaac:
* **#6148** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+///
+/// `CodeBuffer` provides safe abstractions over a byte array, allowing for
+/// a compact byte-array representation without soundness holes.
+///
+/// Use one of the various `print_*` methods to add text into a buffer. When you
+/// are done, call [`take_source_text`] to extract the final [`String`].
+///
+/// # Examples
+/// ```
+/// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 12:
```suggestion
/// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    buf: Vec<u8>,
+}
+
+impl CodeBuffer {
+    /// Create a new empty `CodeBuffer`.
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 41:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    ///
+    /// assert_eq!(code.peek_nth(0), Some('o'));
+    /// assert_eq!(code.peek_nth(2), Some('f'));
+    /// assert_eq!(code.peek_nth(3), None);
+    /// ```
+    #[inline]
+    #[must_use = "Peeking is pointless if the peeked char isn't used"]
+    pub fn peek_nth(&self, n: usize) -> Option<char> {
+        // SAFETY: `buf` is a valid UTF-8 string because of invariants upheld by
+        // CodeBuffer
```
Dunqing, line 131:
```suggestion
        // SAFETY: `buf` is a valid UTF-8 string because of invariants upheld by CodeBuffer
```
It looks like better to keep one line

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+
+    /// Push a single ASCII character into the buffer
+    ///
+    /// # Panics
+    /// If `ch` is not a valid UTF-8 code point in the ASCII range (`0 - 0x7F`).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 143:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    /// method or [`take_source_text`]. This lets you, for example, print an
+    /// 8-byte code point using 4 separate calls to this method.
+    ///
+    /// If you find yourself in such a scenario, consider using
+    /// [`print_unchecked`] instead.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 183:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    ///
+    /// When pushing multiple characters, consider choosing [`print_str`] over
+    /// this method since it's much more efficient. If you really want to insert
+    /// only a single character and you're certain it's ASCII, consider using
+    /// [`print_ascii_char`].
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 220:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    ///
+    /// 1. Both the buffer and the byte sequence are valid UTF-8,
+    /// 2. The buffer became invalid after a call to [`print_byte`] and `bytes`
+    ///    completes any incomplete code points, returning the buffer to a valid
+    ///    state.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 294:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+    /// It is safe to re-use a buffer after calling this method. Its contents
+    /// will be emptied out, but all memory resources are retained and in a
+    /// valid state. You may use [`String::from`] if you don't intend on
+    /// re-using the buffer. It simply calls this method and drops the
+    /// `CodeBuffer` afterwards.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 325:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+        let mut b = [0; 4];
+        self.buf.extend(ch.encode_utf8(&mut b).as_bytes());
+    }
+
+    /// Push a string into this the buffer.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 243:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,422 @@
...
+
+    /// Push a sequence of ASCII characters into the buffer.
+    ///
+    /// # Panics
+    /// If any byte in the iterator is not valid ASCII.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use oxc_codegen::CodeBuffer;
```
Dunqing, line 260:
```suggestion
    /// use oxc_codegen::CodeBuffer;
```

Dunqing (dismissed):
Overall looks good

Boshen:
Is this just `String`? @overlookmotel said it could just be a string but I didn't spend the time converting it.

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+impl AsRef<[u8]> for CodeBuffer {
+    fn as_ref(&self) -> &[u8] {
+        &self.buf
+    }
+}
+impl From<CodeBuffer> for String {
+    fn from(mut printer: CodeBuffer) -> Self {
+        printer.take_source_text()
+    }
+}
```
overlookmotel, line 359: I'm not sure if this should delegate to `take_source_text`. Unlike in `take_source_text`, here you're consuming the `CodeBuffer`, so no need to `take` its contents.

If compiler inlines `take_source_text`, then it will *probably* see that it can elide the `take` operation. But no guarantee that it will inline, so I think better to play it safe and duplicate the code. And in release mode, this function is a no-op (just a type system cast) so all the code will disappear anyway.

This method should be marked `#[inline]` because it's a no-op.

```suggestion
impl From<CodeBuffer> for String {
    #[inline]
    fn from(printer: CodeBuffer) -> Self {
        #[cfg(debug_assertions)]
        {
            String::from_utf8(printer.buf).unwrap()
        }
        #[cfg(not(debug_assertions))]
        {
            // SAFETY: `buf` is valid UTF-8 because of invariants upheld by `CodeBuffer`
            unsafe { String::from_utf8_unchecked(printer.buf) }
        }
    }
}
```

crates/oxc_codegen/src/lib.rs
```diff
@@ -569,13 +570,13 @@ impl<'a> Codegen<'a> {
 
     fn add_source_mapping(&mut self, position: u32) {
         if let Some(sourcemap_builder) = self.sourcemap_builder.as_mut() {
-            sourcemap_builder.add_source_mapping(&self.code, position, None);
+            sourcemap_builder.add_source_mapping(self.code.as_ref(), position, None);
```
overlookmotel, line 573: Maybe add an `as_bytes` method to do the same? `code.as_ref()` is a slightly unintuitive/mysterious (what kind of ref?)

overlookmotel (commented):
I really like this. As you say, we were providing a safe API which allowed doing unsafe things. But, beyond that, the `str::from_utf8_unchecked` calls in codegen gave me the willies, as without reviewing every line of the entire crate, there was no way to know if it was sound or not. Any code anywhere could sneak in a dodgy byte and violate the safety constraints.

`CodeBuffer` encapsulates the unsafe code and provides a clear contract for what's allowed and what's not. Nice!

Most of my comments below are just nits about naming. I think it's useful to be clear about the difference between bytes and `char`s.

overlookmotel:
> Is this just `String`? @overlookmotel said it could just be a string but I didn't spend the time converting it.

Yes, in one sense it's a re-implementation of `String`. This could be a wrapper around a `String` with some added unsafe methods to push bytes to it. But I think the implementation in this PR wrapping a `Vec<u8>` is fine too.

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// valid UTF-8
+    /// unsafe { code.print_byte(0xFF) };
+    /// ```
+    ///
+    /// [`print_ascii_char`]: CodeBuffer::print_ascii_char
+    /// [`print_char`]: CodeBuffer::print_char
+    /// [`take_source_text`]: CodeBuffer::take_source_text
+    /// [`print_unchecked`]: CodeBuffer::print_unchecked
+    #[inline]
+    pub unsafe fn print_byte(&mut self, ch: u8) {
```
overlookmotel, line 205: I think this should be called `print_byte_unchecked` to reflect that it's an unsafe method.

overlookmotel:
If you want to remove some of the unsafe calls, you could also copy this: https://doc.rust-lang.org/std/ascii/enum.Char.html

Then to push a space, you'd do:

```rs
self.code.print_ascii(AsciiByte::Space);
```

`AsciiByte` can *only* be an ASCII byte, so no need for an assertion that it is. i.e. Implementation of `print_ascii` would be just `self.buf.push(ascii_byte)` (same as unsafe method `print_byte`), but it can be a safe method.

Also:

```rs
pub(crate) fn print_asciis<I>(&mut self, bytes: I)
    where I: IntoIterator<Item = AsciiByte>,
{
    self.buf.extend(bytes.map(|b| b as u8));
}
```

```rs
// Previously
unsafe { self.code.print_unchecked(std::iter::repeat(b'\t').take(self.indent as usize)); }
// Now
self.code_print_asciis(std::iter::repeat(AsciiByte::Tab).take(self.indent as usize));
```

If you did want to do that, I'd suggest doing it in a follow-up PR.

overlookmotel:
One last thing... I was looking at the assembly for the various print methods, and noticed something odd:

`print_char(' ')` and `print_str(" ")` seem to produce tighter assembly than `print_ascii_byte(b' ')`.

https://godbolt.org/z/Kv8sEoEed (ignore all the `alloc::raw_vec::*` stuff at the top - that's all handling `Vec` growth)

`print_ascii_byte` has the "this Vec must grow" branch as the default. That's not ideal because growth strategy is doubling, so e.g. when the `Vec` has just grown from 1024 bytes to 2048 bytes, it won't need to grow again until another 1024 bytes have been pushed. "Needs to grow" is a *very* rare occurrence.

Also, it pushes 3 registers to the stack before it does anything else, and it's quite hefty overall. Not good!

`print_ascii_byte_v2` is my own version which pushes all the "Vec needs to grow" logic into a cold path, and only uses 2 registers.

The question is why does `std::vec::Vec` have this "bad" behavior? This seems odd, as `std` is highly optimized. It's implausible it wouldn't be as it is without a good reason.

My best guess is that `Vec` is optimized for a general use case, and it's not uncommon for a `Vec` to be empty and then have a single item pushed. In which case it only hits the "needs to grow" path, and never hits the "no growth necessary". So my `print_ascii_byte_v2` would penalize this common case.

But for our use case here, we are always pushing a ton of bytes. So it may be that the `print_ascii_byte_v2` would work better for this specific case.

We may also be able to replicate what [`Stack` does](https://github.com/oxc-project/oxc/blob/1cee2070506ec58a8dea86ab9ff3ca3258a98ac3/crates/oxc_transformer/src/helpers/stack/standard.rs) to make all these operations even cheaper. This `push` is 1 operation shorter than `print_ascii_byte_v2`, and uses 1 less register: https://godbolt.org/z/sna4K84G1

This is all micro-optimization stuff, but as pushing bytes to output is such a large component of what `oxc_codegen` does, and its such a hot path, it might be worth giving it a go.

(again, that'd be best done in a follow-up PR)

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use oxc_codegen::CodeBuffer;
+    /// let mut code = CodeBuffer::new();
+    ///
+    /// code.print_ascii(['f', 'o', 'o'].iter().copied());
+    /// assert_eq!(String::from(code), "foo");
+    /// ```
+    pub fn print_ascii<I>(&mut self, chars: I)
```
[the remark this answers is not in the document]
overlookmotel: Don't worry about my "fancy" suggestion. That can be done in a follow-up if we think this function is a perf bottleneck.

overlookmotel:
@DonIsaac Just to check - are any of my comments above blockers for you?

I think this PR is a great improvement, and would very much like to get it merged. Removing potential UB is always a high priority, in my opinion.

If you're busy and would like me to finish it off, please say so.

DonIsaac:
> @DonIsaac Just to check - are any of my comments above blockers for you?
> 
> I think this PR is a great improvement, and would very much like to get it merged. Removing potential UB is always a high priority, in my opinion.
> 
> If you're busy and would like me to finish it off, please say so.

None are blockers, I'm just working towards it slowly. I've had a lot going on these past few days 😄

crates/oxc_codegen/src/code_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// let mut code = CodeBuffer::new();
+    /// code.print_str("foo");
+    ///
+    /// assert_eq!(code.peek_nth(0), Some('o'));
+    /// assert_eq!(code.peek_nth(2), Some('f'));
+    /// assert_eq!(code.peek_nth(3), None);
+    /// ```
+    #[inline]
+    #[must_use = "Peeking is pointless if the peeked char isn't used"]
+    pub fn peek_nth(&self, n: usize) -> Option<char> {
```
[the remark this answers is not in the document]
overlookmotel: Don has made this change.

crates/oxc_codegen/src/code_buffer.rs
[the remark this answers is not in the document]
overlookmotel: I've changed the var name in an additional commit. This method is never called, so no point optimizing it.

crates/oxc_codegen/src/lib.rs
overlookmotel: I've made this change in an additional commit.

Boshen:
### Merge activity

* **Oct 13, 5:14 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6151, opened 2024-09-28
perf(transformer): faster parsing JSX pragmas from comments

overlookmotel:
Parse JSX pragmas faster by:

1. Finding patterns with byte-by-byte iteration, with fast path for ASCII.
2. Checking for the `@jsx` prefix first before parsing the rest of the pragma.

These routines could be speeded up further by using unchecked slicing (unsafe code). But I've avoided that as the routines at present are quite complex, so hard to verify. The exception is `trim_end`, which is trivial to verify, so I have used unchecked slicing there.

The perf gain here is only ~0.5%, but we should split these routines out into a separate common crate and further optimize them, so that they can be used elsewhere in the codebase where we do string operations, which are often quite expensive.

### DIFF
--- a/crates/oxc_transformer/src/react/comments.rs
+++ b/crates/oxc_transformer/src/react/comments.rs
@@ -1,0 +1,621 @@
+use oxc_ast::{Comment, CommentKind};
+use oxc_syntax::identifier::is_irregular_whitespace;
+
+use crate::{ReactJsxRuntime, ReactOptions, TransformCtx};
+
+/// Scan through all comments and find the following pragmas:
+///
+/// * @jsxRuntime classic / automatic
+/// * @jsxImportSource custom-jsx-library
+/// * @jsxFrag Preact.Fragment
+/// * @jsx Preact.h
+///
+/// The comment does not need to be a JSDoc comment,
+/// otherwise `JSDoc` could be used instead.
+///
+/// This behavior is aligned with Babel.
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#6151** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/react/comments.rs
```diff
@@ -0,0 +1,621 @@
...
+/// * @jsxImportSource custom-jsx-library
+/// * @jsxFrag Preact.Fragment
+/// * @jsx Preact.h
+///
+/// The comment does not need to be a JSDoc comment,
+/// otherwise `JSDoc` could be used instead.
+///
+/// This behavior is aligned with Babel.
+pub(crate) fn update_options_with_comments(options: &mut ReactOptions, ctx: &TransformCtx) {
+    for comment in ctx.trivias.comments() {
```
Boshen, line 18: With the introduction of leading comments, I think iterating through the first set of leading comments should suffice. There are no specifications but I think we can iterate through less comments.

Boshen: Can experiment in another PR.

overlookmotel: Ah OK great!

Issue #4655, opened 2024-08-05
☂️ eslint-plugin-promise


Boshen:

Goldziher:
I'll work on prefer-await-to-callbacks

GeoffreyParrier:
Based on #13233 `import/no-native` should have `🚫 = No need to implement` if I'm not mistaken

taearls:
All the rules for `eslint-plugin-promise` appear to have been implemented. @camc314 should this issue be closed?

PR #6157, opened 2024-09-29
feat(linter): implement `no-callback-in-promise`

shulaoda:
Related to #4655

### DIFF
--- a/crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
+++ b/crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
@@ -4,14 +4,14 @@ use oxc_ast::{
 };
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_span::{GetSpan, Span};
+use oxc_span::{CompactStr, GetSpan, Span};
 
 use crate::{context::LintContext, rule::Rule, AstNode};
 
-static CALLBACKS: [&str; 4] = ["done", "cb", "callback", "next"];
-
 fn no_callback_in_promise_diagnostic(span: Span) -> OxcDiagnostic {
-    OxcDiagnostic::warn("Avoid calling back inside of a promise").with_label(span)
+    OxcDiagnostic::warn("Avoid calling back inside of a promise")
+        .with_help("Use `then` and `catch` directly, avoid callbacks")
+        .with_label(span)
 }
 
 #[derive(Debug, Default, Clone)]
@@ -19,7 +19,7 @@ pub struct NoCallbackInPromise(Box<NoCallbackInPromiseConfig>);
 
 #[derive(Debug, Default, Clone)]
 pub struct NoCallbackInPromiseConfig {
-    exceptions: Vec<String>,
+    callbacks: Vec<CompactStr>,
 }
 
 impl std::ops::Deref for NoCallbackInPromise {
@@ -60,8 +60,8 @@ declare_oxc_lint!(
     /// Examples of **correct** code for this rule:
     /// ```js
     /// Promise.resolve()
-    ///   .then(() => setTimeout(() => callback(null, 'data'), 0))
-    ///   .catch((err) => setTimeout(() => callback(err.message, null), 0))
+    ///   .then((data) => { console.log(data) })
+    ///   .catch((err) => { console.error(err) })
     /// ```
     NoCallbackInPromise,
     correctness,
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,193 @@
...
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+static CALLBACKS: [&str; 4] = ["done", "cb", "callback", "next"];
+
+fn no_callback_in_promise_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Avoid calling back inside of a promise").with_label(span)
```
camchenry, line 14: Could we add some help text here to tell users how to fix this problem? Like using `setTimeout` to wrap the callback

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn from_configuration(value: serde_json::Value) -> Self {
+        Self(Box::new(NoCallbackInPromiseConfig {
+            exceptions: value
+                .get(0)
+                .and_then(|v| v.get("exceptions"))
+                .and_then(serde_json::Value::as_array)
+                .map(|v| {
+                    v.iter()
+                        .filter_map(serde_json::Value::as_str)
+                        .map(ToString::to_string)
```
camchenry, line 80: Do we need to allocate on the heap here? Could we use `CompactStr` to store these exceptions instead? For example: https://github.com/oxc-project/oxc/blob/40bd9198496dbf2dbcbff1b1cb58869aa3443e92/crates/oxc_linter/src/rules/eslint/no_extend_native.rs#L76-L83

shulaoda: Let me try to attempt it

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
shulaoda: I found that they haven't achieved this since 2021. Recently, there is [a new PR](https://github.com/eslint-community/eslint-plugin-promise/pull/514/files) working on this, and I will try to fix it.
```
  // TODO: support safe callbacks (see #220)
  // 'whatever.then((err) => { process.nextTick(() => cb()) })',
  // 'whatever.then((err) => { setImmediate(() => cb())) })',
  // 'whatever.then((err) => setImmediate(() => cb()))',
  // 'whatever.then((err) => process.nextTick(() => cb()))',
  // 'whatever.then((err) => process.nextTick(cb))',
  // 'whatever.then((err) => setImmediate(cb))',
```

shulaoda: There are also some issues with this PR, let's first implement the current feature.

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,192 @@
...
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{CompactStr, GetSpan, Span};
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_callback_in_promise_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Avoid calling back inside of a promise")
+        .with_help("Use `then` and `catch` directly, avoid callbacks")
```
DonIsaac, line 13:
```suggestion
        .with_help("Use `then` and `catch` directly")
```

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,192 @@
...
+    ///   .catch((err) => { console.error(err) })
+    /// ```
+    NoCallbackInPromise,
+    correctness,
+);
+
+impl Rule for NoCallbackInPromise {
+    fn from_configuration(value: serde_json::Value) -> Self {
+        let default_callbacks: Vec<CompactStr> =
+            ["done", "cb", "callback", "next"].map(Into::into).to_vec();
```
DonIsaac, line 73: These should be in the `Default` impl

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,192 @@
...
+        Self(Box::new(NoCallbackInPromiseConfig {
+            callbacks: default_callbacks
+                .into_iter()
+                .filter(|item| !exceptions.contains(&item.to_string()))
+                .collect(),
+        }))
+    }
+
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if !self.is_callback(node) {
```
DonIsaac, line 93: You can move the code in this function into this rule, then reference the call_expr without needing to unwrap later

crates/oxc_linter/src/rules/promise/no_callback_in_promise.rs
```diff
@@ -0,0 +1,192 @@
...
+                };
+
+                let Some(id) = call_expr.arguments.first().and_then(|arg| {
+                    arg.as_expression().and_then(Expression::get_identifier_reference)
+                }) else {
+                    return;
+                };
+
+                let name = id.name.as_str();
+                if self.callbacks.contains(&name.into()) {
```
DonIsaac, line 106: Consider binary searching using Ord to avoid string allocations. You'll need to sort this list when creating the rule from configuration.

DonIsaac (approved):

PR #6165, opened 2024-09-29
feat(linter/node): implement no-new-require

jelly:


camc314 (approved):

crates/oxc_linter/src/rules/node/no_new_require.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+
+    let pass = vec![
+        "var appHeader = require('app-header')",
+        "var AppHeader = new (require('app-header'))",
```
DonIsaac, line 62: Why should this be allowed?

crates/oxc_linter/src/rules/node/no_new_require.rs
```diff
@@ -0,0 +1,73 @@
...
+        "var AppHeader = new (require('headers').appHeader)",
+        "var AppHeader = require('app-header'); var appHeader = new AppHeader();",
```
DonIsaac, line 64: Can you please add a pass case for `new require('foo')()`

crates/oxc_linter/src/rules/node/no_new_require.rs
```diff
@@ -0,0 +1,73 @@
...
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+use crate::{context::LintContext, rule::Rule, AstNode};
+
+fn no_new_require(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Unexpected use of new with require")
+        .with_label(span)
+        .with_help("Initialise the constructor separate from the import statement")
```
DonIsaac, line 11: This message is a little confusing

crates/oxc_linter/src/rules/node/no_new_require.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct NoNewRequire;
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Warn about calling `new` on `require`.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// The `require` function is used to include modules and might return a constructor. As this
```
DonIsaac, line 24: We should mention that `require` is a non-newable function, so if they're not using a custom function also called `require` then they'll get a runtime error.

crates/oxc_linter/src/rules/node/no_new_require.rs
```diff
@@ -0,0 +1,73 @@
...
+fn no_new_require(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Unexpected use of new with require")
+        .with_label(span)
+        .with_help("Initialise the constructor separate from the import statement")
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct NoNewRequire;
+
+declare_oxc_lint!(
```
DonIsaac, line 17: Prefer curly braces for `declare_oxc_lint!`

crates/oxc_linter/src/rules/node/no_new_require.rs
jelly: That's what the original rule allows as well, if we want to be more strict I would be happy to implement that behaviour.

crates/oxc_linter/src/rules/node/no_new_require.rs
jelly: test.js
```javascript
const x = new require("./lol.js");
console.log(x);
```

foo.js
```javascript
class Foo {
};

module.exports = Foo;
```

This does not throw a runtime error with nodejs 22.

jelly: Similar example here https://nodejs.org/docs/v20.17.0/api/modules.html#modules-commonjs-modules

DonIsaac: You make a good point. This is pathological, but valid

crates/oxc_linter/src/rules/node/no_new_require.rs
nrayburn-tech: Parenthesis are optional when constructing an instance that doesn't take any arguments.  Confusing but valid.

PR #6169, opened 2024-09-29
fix(transformer): `NonEmptyStack::push` write value before updating cursor

overlookmotel:
`NonEmptyStack::push` was previously updating `cursor` before writing the value to stack. This could be UB if the type `T` is `Drop`, and writing the value panicked - as then `drop` would attempt to drop an uninitialized `T`.

I think `ptr::write` is infallible (can't panic), so it shouldn't matter. But *maybe* in debug mode it can panic if some invariant is broken (e.g. pointer not aligned). So play it safe and write the value first and update the cursor after.


overlookmotel:
* **#6169** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6173, opened 2024-09-30
refactor(transformer): use `AstBuilder` instance from `TraverseCtx`

### LINKED ISSUE #6172: Refactor transformer to use the instance of `AstBuilder` which is in `TraverseCtx` everywhere

For example:

Should become:

```rs
```

shulaoda:
closes #6172

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -66,6 +66,7 @@ pub struct TransformerReturn {
 pub struct Transformer<'a> {
     ctx: TransformCtx<'a>,
     options: TransformOptions,
+    allocator: &'a Allocator,
 }
 
 impl<'a> Transformer<'a> {
@@ -79,7 +80,7 @@ impl<'a> Transformer<'a> {
     ) -> Self {
         let ctx =
             TransformCtx::new(allocator, source_path, source_type, source_text, trivias, &options);
-        Self { ctx, options }
+        Self { ctx, options, allocator }
     }
 
     pub fn build_with_symbols_and_scopes(
@@ -87,9 +88,8 @@ impl<'a> Transformer<'a> {
         symbols: SymbolTable,
         scopes: ScopeTree,
         program: &mut Program<'a>,
-        allocator: &'a Allocator,
     ) -> TransformerReturn {
-        let traverse_ctx = TraverseCtx::new(scopes, symbols, allocator);
+        let traverse_ctx = TraverseCtx::new(scopes, symbols, self.allocator);
         let mut transformer = TransformerImpl {
             x0_typescript: TypeScript::new(self.options.typescript, &self.ctx),
             x1_react: React::new(self.options.react, &self.ctx, &traverse_ctx),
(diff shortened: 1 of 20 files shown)


crates/oxc_transformer/src/lib.rs
```diff
@@ -85,12 +85,12 @@ impl<'a> Transformer<'a> {
         symbols: SymbolTable,
         scopes: ScopeTree,
         program: &mut Program<'a>,
+        allocator: &'a Allocator,
```
Boshen, line 88: You can save it in Transformer::new instead of passing it in again.

crates/oxc_traverse/src/lib.rs
```diff
@@ -144,12 +143,9 @@ mod compile_fail_tests;
 
 pub fn traverse_mut<'a, Tr: Traverse<'a>>(
     traverser: &mut Tr,
-    allocator: &'a Allocator,
     program: &mut Program<'a>,
-    symbols: SymbolTable,
-    scopes: ScopeTree,
+    mut ctx: TraverseCtx<'a>,
 ) -> (SymbolTable, ScopeTree) {
-    let mut ctx = TraverseCtx::new(scopes, symbols, allocator);
```
Boshen, line 152: I'll let @overlookmotel decide on this change.

Dunqing:
Sorry, I tagged wrong

crates/oxc_traverse/src/lib.rs
overlookmotel: Please could you revert this to as it was?

`Traverse` is stacked full of unsafe code, and the scheme which maintains soundness is quite finely balanced. It'd be easy to accidentally open the door to UB. The problem is that it's a while since I wrote it, and I didn't document it enough (bad bad me!), so I'd need to retrace my steps and scratch my head a lot to figure out if it's safe to expose `TraverseCtx` or not.

As this change is not essential, it'd be preferable just to leave it as was if you don't mind please.

overlookmotel (commented):
Thank you so much for doing this.

Apart from the comment above about reverting `traverse_mut` to how it was, looks great.

Please could make that change and ping me, so I can do a final review and merge it ASAP before we get merge conflicts?

crates/oxc_traverse/src/lib.rs
shulaoda: In fact, this was a last resort, because prior to `traverse_mut`, the `x1_react` in `TransformerImpl` requires the `ast` from `traverse_ctx`.

overlookmotel:
@shulaoda This has merge conflicts and I was worried that if I requested changes, there'd be more merge conflicts again by the time I got to reviewing your changes (probably we are in different time zones, and this PR touches a lot of files in the transformer, which is currently under heavy development).

So I have taken the liberty of rebasing this on latest main, fixing the problem discussed above, and merging it.

Sorry if it feels like I've taken over your work. I just really didn't want to waste your time going back-and-forwards endlessly dealing with merge conflicts. *Really* appreciate your work on this.

PR #6176, opened 2024-09-30
refactor(transformer/react): don't follow the Babel's imports order

Dunqing:
The `ImportDeclaration` orders are not matter, they are always hoisted. Previously to 100% match Babel's output to pass the tests, we did some weird logic in our implementation. In this PR I removed them, and always inserted the imports at the top of the statements. The reason for doing this is I am moving `ModuleImports` to https://github.com/oxc-project/oxc/blob/5b5aad856213df18f09bf1972f6e3a13a5302a15/crates/oxc_transformer/src/common/mod.rs#L14-16

After being moved to `common`,  the `NamedImports` will only insert imports once in `common` instead of inserting in each plugin that used it.

### DIFF
--- a/crates/oxc_transformer/src/helpers/module_imports.rs
+++ b/crates/oxc_transformer/src/helpers/module_imports.rs
@@ -59,16 +59,12 @@ impl<'a> ModuleImports<'a> {
     }
 
     /// Add `var named_import from 'source'`
-    pub fn add_require(&self, source: Atom<'a>, import: NamedImport<'a>, front: bool) {
-        let len = self.imports.borrow().len();
+    pub fn add_require(&self, source: Atom<'a>, import: NamedImport<'a>) {
         self.imports
             .borrow_mut()
             .entry(ImportType::new(ImportKind::Require, source))
             .or_default()
             .push(import);
-        if front {
-            self.imports.borrow_mut().move_index(len, 0);
-        }
     }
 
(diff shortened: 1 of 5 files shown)


Dunqing:
* **#6176** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Personally I am not keen on diverging from Babel at this stage.

My feeling is that Babel's transformer tests which we are using for conformance are not enough. To *really* test the transformer we need to throw a lot more code at it. The best way I can see to do that is to do something like this:

* Crawl all the JS/TS files from the top 1000 npm packages.
* Run each of them through both Babel and Oxc transformer.
* Compare the outputs.

When we have no mismatches all all, we can have high confidence that our transformer is bug-free (or at least has no more bugs than Babel). But to run those tests without spurious differences in the diffs we need our transformer's output to match Babel's exactly - even in cosmetic aspects like statement order.

I'll look into this further and try to find if we can achieve what we're trying to do (move `ModuleImports` into `common`) without failing any tests. I *think* it should be possible.

Dunqing:
@overlookmotel Feel free to take over this stack, I am on holiday now. 🏖️

overlookmotel:
#6183 negates the need for this.

PR #6182, opened 2024-09-30
chore(linter): add category to completed rules

shulaoda:
I noticed that some rules have been completed, but their category is still set to `nursery`.


Boshen:
I don't think these rules are completed in a sense that they are usable? They haven't been test properly on actual files yet.

DonIsaac:
I agree with Boshen. These should not be moved out of nursery right now.

PR #6195, opened 2024-09-30
feat(allocator): add `Vec::into_boxed_slice`

### LINKED ISSUE #6193: allocator: `Vec::into_boxed_slice` does not return our `Box`
`Vec::into_boxed_slice` returns an `allocator_api2::stable::boxed::Box`, which is not assignable to `oxc_allocator::boxed::Box`

Add this test case to `vec.rs` to reproduce.
```rust
#[test]
fn test_vec_to_boxed_slice() {
    // file: oxc_allocator/src/vec.rs
    use crate::boxed::Box;

    let v = Vec::from_iter_in([1,2,3], &allocator);
    let b: Box<[i32]> = v.into_boxed_slice(); // <- compiler error
    assert_eq!(&*b, &[1, 2, 3]);
}
```

DonIsaac:
Note that this PR does not implement the inverse operation (`Box::to_vec` for `[T]`).

### DIFF
--- a/crates/oxc_allocator/src/boxed.rs
+++ b/crates/oxc_allocator/src/boxed.rs
@@ -76,6 +76,25 @@ impl<'alloc, T> Box<'alloc, T> {
     }
 }
 
+impl<'alloc, T: ?Sized> Box<'alloc, T> {
+    /// Create a [`Box`] from a raw pointer to a value.
+    ///
+    /// The [`Box`] takes ownership of the data pointed to by `ptr`.
+    ///
+    /// # SAFETY
+    /// Data pointed to by `ptr` must live as long as `'alloc`.
+    /// This requirement is met if the pointer was obtained from other data in the arena.
+    ///
+    /// Data pointed to by `ptr` must *only* be used for this `Box`. i.e. it must be unique,
+    /// with no other aliases. You must not, for example, create 2 `Box`es from the same pointer.
+    ///
+    /// `ptr` must have been created from a `*mut T` or `&mut T` (not a `*const T` / `&T`).
+    #[inline]
+    pub(crate) const unsafe fn from_non_null(ptr: NonNull<T>) -> Self {
+        Self(ptr, PhantomData)
+    }
+}
+
 impl<'alloc, T: ?Sized> ops::Deref for Box<'alloc, T> {
     type Target = T;
 
--- a/crates/oxc_allocator/src/vec.rs
+++ b/crates/oxc_allocator/src/vec.rs
@@ -7,6 +7,7 @@ use std::{
     fmt::Debug,
     hash::{Hash, Hasher},
     ops,
+    ptr::NonNull,
 };
 
 use allocator_api2::vec;
@@ -14,7 +15,7 @@ use bumpalo::Bump;
 #[cfg(any(feature = "serialize", test))]
 use serde::{ser::SerializeSeq, Serialize, Serializer};
 
-use crate::Allocator;
+use crate::{Allocator, Box};
 
 /// Bumpalo Vec
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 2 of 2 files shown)


DonIsaac:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6195?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6195** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6194** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/oxc_allocator/src/vec.rs
```diff
@@ -101,6 +101,33 @@ impl<'alloc, T> Vec<'alloc, T> {
...
+    ///
+    /// assert_eq!(&*b, &[1, 2, 3]);
+    /// ```
+    pub fn into_boxed_slice(self) -> Box<'alloc, [T]> {
+        let b = self.0.into_boxed_slice();
+        let (ptr, _) = allocator_api2::boxed::Box::into_non_null(b);
+        // SAFETY: this is effectively a transmute from allocator_api2's box to
+        // our own representation. Ptr is non-null and points to initialized, well-aligned memory.
+        unsafe { Box::from_non_null(ptr) }
+    }
```
overlookmotel, line 130: Hmm. I'm not sure about this.

I'm not that familiar with `allocator_api2`'s implementation of `into_boxed_slice`, but it sounds like it can reallocate (copy all the data), which we definitely don't want.

That makes sense for `allocator_api2` as it's generic over any kind of allocator, but in our case we're using a bump allocator which cannot recover excess capacity, so it's pointless to copy the data.

I don't know if `bumpalo::Bump` would automatically ignore the request to reallocate or not. Would you like to look into that?

Alternatively, we could get the pointer and length of the `Vec` and construct a `NonNull<[T]>` manually. That's definitely do what we want. (Note: would need to wrap the `Vec` in a `ManuallyDrop` first, as `allocator_api2`'s `Vec` is `Drop`)

This all requires quite a bit of care. I'm not completely up to speed on the safety constraints. Sometimes there are non-obvious subtleties to avoiding UB, and I usually end up having to spend a few hours boning up on arcane details before I'm sure (well at least reasonably sure!) that I've not committed some UB crime!

Do you have a concrete use case for wanting a boxed slice to data in the arena?

overlookmotel: By the way, we're planning to replace this `Vec` implementation with a hand-written one with length and capacity fields as `u32`s, to bring size of `Vec` down from 32 bytes to 24 bytes (https://github.com/oxc-project/backlog/issues/18). Dunqing was going to look at doing that fairly soon.

So at that point, we wouldn't be able to use `allocator_api2`'s `into_boxed_slice` anyway.

DonIsaac: https://github.com/zakarumych/allocator-api2/blob/f937b9650e10137f782698bd4eb4f8b36f1817a7/src/stable/vec/mod.rs#L1046

I don't think it creates a new allocation. I want this because it cuts the size of a dynamically sized vec in half (32 -> 16 bytes) at the cost of not being re-sizeable.  I'm working on a prototype where this is useful.

overlookmotel: Actually, this works:

```rs
pub fn into_boxed_slice(self) -> Box<'alloc, [T]> {
    let slice = self.0.leak();
    let ptr = NonNull::from(slice);
    // SAFETY: `ptr` points to a valid slice `[T]`.
    // Lifetime of returned `Box<'alloc, [T]>` is same as lifetime of consumed `Vec<'alloc, T>`,
    // so data in the `Box` must be valid for its lifetime.
    unsafe { Box::from_non_null(ptr) }
}
```

Then you can remove the comment about "If the vector has excess capacity, its items will be moved into a newly-allocated buffer with exactly the right capacity."

crates/oxc_allocator/src/boxed.rs
```diff
@@ -76,6 +76,18 @@ impl<'alloc, T> Box<'alloc, T> {
     }
 }
 
+impl<'alloc, T: ?Sized> Box<'alloc, T> {
+    /// Create a [`Box`] from a raw pointer to a value in the arena.
+    ///
+    /// # SAFETY
+    /// It is only safe to create a `Box` this way with pointers taken out of
+    /// another `Box` in the same arena.
```
overlookmotel, line 84: This comment should be updated. It doesn't require that the pointer is from *another `Box`*. If that was true, then `Vec:: into_boxed_slice` which you're adding here would violate the contract, as it's not a `Box`!

The requirement is that pointer needs to point to data in same arena (could be from a `Vec`, a `String`, anything else).

Actually, not even that is quite right. We *want* it to be date from the same arena, but the safety constraint I think is  a bit looser - it can point to *any* data as long as that data outlives `'alloc`.

There's also a safety requirement about no aliasing.

This may seem like nitpicking, but it's vitally important that documentation for unsafe methods is accurate about what the contract is, otherwise it's impossible to verify the method is being used correctly.

DonIsaac:
@overlookmotel does it make sense to just abandon this PR?

overlookmotel:
No, I think we can do it. I just think better to use this version:

```rs
pub fn into_boxed_slice(self) -> Box<'alloc, [T]> {
    let slice = self.0.leak();
    let ptr = NonNull::from(slice);
    // SAFETY: `ptr` points to a valid slice `[T]`.
    // Lifetime of returned `Box<'alloc, [T]>` is same as lifetime of consumed `Vec<'alloc, T>`,
    // so data in the `Box` will be valid for its lifetime.
    unsafe { Box::from_non_null(ptr) }
}
```

It's safe, and won't copy memory.

crates/oxc_allocator/src/boxed.rs
overlookmotel: I've implemented this change in an extra commit.

crates/oxc_allocator/src/vec.rs
overlookmotel: I've implemented this change in an extra commit.

overlookmotel:
@DonIsaac Hope you don't mind that I've stepped in to finish this off. Just wanted to close a few open PRs which are close to the finish line. Unless you have any objections, I plan to merge this with the additional commits I've added.

Boshen:
### Merge activity

* **Oct 12, 12:29 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6196, opened 2024-09-30
fix(linter): get correct source offsets for astro files

camchenry:
Previously, we reported the `source_text` of an Astro file correctly, but the `start` offset was incorrectly set to before the `---` of the frontmatter. The start value now correctly points to the end of the split.

### DIFF
--- a/crates/oxc_linter/src/loader/partial_loader/astro.rs
+++ b/crates/oxc_linter/src/loader/partial_loader/astro.rs
@@ -42,8 +42,9 @@ impl<'a> AstroPartialLoader<'a> {
             return None;
         };
 
-        let js_code =
-            Span::new(start + ASTRO_SPLIT.len() as u32, end).source_text(self.source_text);
+        // move start to the end of the ASTRO_SPLIT
+        let start = start + ASTRO_SPLIT.len() as u32;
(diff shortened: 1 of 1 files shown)


camchenry:
* **#6196** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/loader/partial_loader/astro.rs
```diff
@@ -42,8 +42,9 @@ impl<'a> AstroPartialLoader<'a> {
             return None;
         };
 
-        let js_code =
-            Span::new(start + ASTRO_SPLIT.len() as u32, end).source_text(self.source_text);
+        // move start to the end of the ASTRO_SPLIT
+        let start = start + ASTRO_SPLIT.len() as u32;
```
DonIsaac, line 46: This logic looks identical. Am I missing something?

camchenry: yep, the issue here is that the `start` used for getting the source text had the length added to it, but then the `start` without the addition was stored as the start offset

PR #6197, opened 2024-10-01
feat(linter): allow fixing in files with source offsets

### LINKED ISSUE #5913: linter: `no-useless-spread` fixer clobbers code before span in Vue script block
```vue
<template>
    <div>Hello world</div>
</template>

<script lang="js">
const isCondition = true

const bar = {
    ...(isCondition ? { a: 1 } : { b: 2 })
}

export default {
    data() {
        bar
    }
}
</script>
```

```
  ! ]8;;https://oxc.rs/docs/guide/usage/linter/rules/unicorn/no-useless-spread.html\eslint-plugin-unicorn(no-useless-spread)]8;;\: Using a spread operator here creates a new object unnecessarily.
   ,-[src\renderer\App.vue:5:5]
 4 | const bar = {
 5 |     ...(isCondition ? { a: 1 } : { b: 2 })
   :     ^^^
 6 | }
   `----
  help: `isCondition ? { a: 1 } : { b: 2 }` returns a new object. Spreading it into an object expression to create a new object is redundant.
```

And here is the fix:

https://github.com/user-attachments/assets/ebd3752f-a791-481c-8baa-7ec8ae022822

camchenry:
- fixes https://github.com/oxc-project/oxc/issues/5913

This PR fixes the calculation of spans when dealing with files that have multiple sources and non-zero source start offsets. This is almost always the case for `.vue`, `.astro`, and `.svelte` files. This enables us to correctly apply fixes for these file types both in the CLI with `oxlint` and also in editors like VS Code.

https://github.com/user-attachments/assets/2836c8bd-09be-4e59-801d-7c95f8c2491f

I'm open to ideas on how to improve testing in this area, as I don't think that we currently have any tests for fixing files end-to-end (beyond what the linter rules check). I did run this locally on the gitlab repository (which is written in Vue) and all of the fixes appeared to be applied correctly.

### DIFF
--- a/crates/oxc_linter/src/service.rs
+++ b/crates/oxc_linter/src/service.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Cow,
     ffi::OsStr,
     fs,
     path::{Path, PathBuf},
@@ -226,6 +227,9 @@ impl Runtime {
         })
     }
 
+    // clippy: the source field is checked and assumed to be less than 4GB, and
+    // we assume that the fix offset will not exceed 2GB in either direction
+    #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
     fn process_path(&self, path: &Path, tx_error: &DiagnosticSender) {
         if self.init_cache_state(path) {
             return;
@@ -250,9 +254,7 @@ impl Runtime {
             }
         };
 
-        let sources = PartialLoader::parse(ext, &source_text);
-        let is_processed_by_partial_loader = sources.is_some();
-        let sources = sources
+        let sources = PartialLoader::parse(ext, &source_text)
             .unwrap_or_else(|| vec![JavaScriptSource::partial(&source_text, source_type, 0)]);
 
         if sources.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/service.rs
```diff
@@ -250,26 +251,44 @@ impl Runtime {
...
                 if fix_result.fixed {
-                    fs::write(path, fix_result.fixed_code.as_bytes()).unwrap();
+                    // write to file, replacing only the changed part
+                    let start = source.start.saturating_add_signed(fix_offset) as usize;
+                    let end = start + source.source_text.len();
+                    new_source_text.to_mut().replace_range(start..end, &fix_result.fixed_code);
+                    let old_code_len = u32::try_from(source.source_text.len()).unwrap();
+                    let new_code_len = u32::try_from(fix_result.fixed_code.len()).unwrap();
+                    fix_offset += i32::try_from(new_code_len).unwrap();
+                    fix_offset -= i32::try_from(old_code_len).unwrap();
```
camchenry, line 291: I would love any ideas on how to avoid all of this number casting here. the main issue is that `start..end` has to be a range of `usize`, but `source.start` is `u32`, while `fix_offset` needs to potentially be negative (as source files can be smaller after applying fixes), so it becomes `i32`.

Boshen: Just cast them with `as`, we assume source length is under u32.

Boshen: https://github.com/oxc-project/oxc/blob/51a78d594640a069db53940bdb838fb420ea698c/crates/oxc_parser/src/lib.rs#L22-L33

camchenry: right, I am often forgetting about that, I'll update this to just cast and ignore truncation

DonIsaac:
I like this. I'm currently working on a full re-write of source files and writing, but this is a good in-between. Excellent stuff @camchenry

DonIsaac (approved):

DonIsaac:
I'd like feedback from @camc314. After that, feel free to merge when you're ready

PR #6201, opened 2024-10-01
feat(minifier): minify unary plus negation.

7086cmd:
It is uncertain whether this applies to other situations, so let us eliminate it in `negation` first, and after I have conducted a comprehensive investigation of the Closure Compiler, we can make a definitive decision.

### DIFF
--- a/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
+++ b/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
@@ -160,13 +160,17 @@ impl<'a> PeepholeFoldConstants {
             UnaryOperator::UnaryPlus => {
                 if expr.argument.is_number() {
                     Some(ctx.ast.move_expression(&mut expr.argument))
-                } else if let Expression::UnaryExpression(un_expr ) = &mut expr.argument {
-                    // handle +- 
+                } else if let Expression::UnaryExpression(un_expr) = &mut expr.argument {
+                    // handle +-
                     if un_expr.operator == UnaryOperator::UnaryNegation {
                         println!("Here it is!");
                         Some(ctx.ast.move_expression(&mut expr.argument))
-                    } else { None }
-                } else { None }
+                    } else {
+                        None
+                    }
+                } else {
+                    None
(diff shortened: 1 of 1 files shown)


crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
```diff
@@ -157,8 +157,19 @@ impl<'a> PeepholeFoldConstants {
...
             }
             // `+1` -> `1`
-            UnaryOperator::UnaryPlus if expr.argument.is_number() => {
-                Some(ctx.ast.move_expression(&mut expr.argument))
+            UnaryOperator::UnaryPlus => {
+                if expr.argument.is_number() {
+                    Some(ctx.ast.move_expression(&mut expr.argument))
+                } else if let Expression::UnaryExpression(un_expr) = &mut expr.argument {
+                    // handle +-
+                    if un_expr.operator == UnaryOperator::UnaryNegation {
```
Boshen, line 165: You can try the https://doc.rust-lang.org/std/primitive.bool.html#method.then method

crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
```diff
@@ -157,8 +157,19 @@ impl<'a> PeepholeFoldConstants {
                 Some(ctx.ast.move_expression(&mut expr.argument))
             }
             // `+1` -> `1`
-            UnaryOperator::UnaryPlus if expr.argument.is_number() => {
-                Some(ctx.ast.move_expression(&mut expr.argument))
+            UnaryOperator::UnaryPlus => {
+                if expr.argument.is_number() {
```
Boshen, line 161: Try using match if there are many if elses, we may have more branches in the future.

match xxx {
  _ if expr.argument.is_number() => {}
}

crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
7086cmd: Refactored.

PR #6209, opened 2024-10-01
refactor(transformer): use AstBuilder instance from TraverseCtx

### LINKED ISSUE #6172: Refactor transformer to use the instance of `AstBuilder` which is in `TraverseCtx` everywhere

overlookmotel:
A small modification of #6173. Closes #6172.

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -8,6 +8,8 @@
 //! * <https://babel.dev/docs/presets>
 //! * <https://github.com/microsoft/TypeScript/blob/main/src/compiler/transformer.ts>
 
+use oxc_ast::AstBuilder;
+
 // Core
 mod common;
 mod compiler_assumptions;
@@ -68,6 +70,7 @@ pub struct TransformerReturn {
 pub struct Transformer<'a> {
     ctx: TransformCtx<'a>,
     options: TransformOptions,
+    allocator: &'a Allocator,
 }
 
 impl<'a> Transformer<'a> {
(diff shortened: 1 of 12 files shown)


overlookmotel:
* **#6209** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This is basically #6173 with a few tweaks and rebased on main to solve bug fixes. I've reviewed in full, so I don't think it needs further review. So merging.

crates/oxc_transformer/src/lib.rs
```diff
@@ -90,11 +92,12 @@ impl<'a> Transformer<'a> {
...
         program: &mut Program<'a>,
     ) -> TransformerReturn {
-        let allocator = self.ctx.ast.allocator;
+        let allocator = self.allocator;
+        let ast_builder = AstBuilder::new(allocator);
 
         let mut transformer = TransformerImpl {
             x0_typescript: TypeScript::new(self.options.typescript, &self.ctx),
-            x1_react: React::new(self.options.react, &self.ctx),
+            x1_react: React::new(self.options.react, ast_builder, &self.ctx),
```
overlookmotel, line 100: @shulaoda This is how I solved the problem. This is pretty hacky, but it works.

We are intending to revamp the interface to `Transformer` soon, so hopefully we can come up with a more ergonomic solution then.

shulaoda: 👍

PR #6205, opened 2024-10-01
feat(minifier): support bitwise unary folding.

7086cmd:
Related expr: https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540compilation_level%2520SIMPLE_OPTIMIZATIONS%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Ax%2520%253D%2520~434.2543%250Ax%2520%253D%2520~-0x3%250Ax%2520%253D%2520~9.

Don't support folding non-digit numbers, which is opt-in and throws an warning in closure compiler. Let's do it in a seprate one.

PR #6214, opened 2024-10-01
feat(minifier): support subtraction assignment.

### LINKED ISSUE #6172: Refactor transformer to use the instance of `AstBuilder` which is in `TraverseCtx` everywhere

7086cmd:
Due to the potential for string concatenation when using the `+=` operator, we should only handle the scenario when using the `-=` operator.

### DIFF
--- a/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
+++ b/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
@@ -1,3 +1,14 @@
+use std::cmp::Ordering;
+
+use num_bigint::BigInt;
+use oxc_ast::ast::*;
+use oxc_span::{GetSpan, Span, SPAN};
+use oxc_syntax::{
+    number::{NumberBase, ToJsInt32},
+    operator::{BinaryOperator, LogicalOperator, UnaryOperator},
+};
+use oxc_traverse::{Ancestor, Traverse, TraverseCtx};
+
 use crate::{
     node_util::{
         is_exact_int64, IsLiteralValue, MayHaveSideEffects, NodeUtil, NumberValue, ValueType,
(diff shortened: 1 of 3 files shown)


crates/oxc_minifier/src/ast_passes/peephole_substitute_alternate_syntax.rs
```diff
@@ -260,6 +267,37 @@ impl<'a> PeepholeSubstituteAlternateSyntax {
...
     }
+
+    fn try_compress_assignment_expression(
+        &mut self,
+        expr: &mut AssignmentExpression<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) -> Option<Expression<'a>> {
+        let Some(target) = expr.left.as_simple_assignment_target() else { return None };
+        // TODO avoid using `clone_in`, as it is performance-intensive
+        let target = target.clone_in(ctx.ast.allocator);
```
Boshen, line 278: I think you should only clone when necessary, this is currently clone on every assignment expression?

7086cmd: The clone only appears in the following situations:
- Is the target situation (from `-= 1` or `-= -1`)
- Is a simple assignment target
I am also investigating whether we can use `mem::replace`.

7086cmd: I used the transform, and it is solved.

crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
```diff
@@ -1,3 +1,14 @@
...
+
+use num_bigint::BigInt;
+use oxc_ast::ast::*;
+use oxc_span::{GetSpan, Span, SPAN};
+use oxc_syntax::{
+    number::{NumberBase, ToJsInt32},
+    operator::{BinaryOperator, LogicalOperator, UnaryOperator},
+};
+use oxc_traverse::{Ancestor, Traverse, TraverseCtx};
+
```
7086cmd, line 11: This resolves a prev problem in https://github.com/oxc-project/oxc/pull/6205#discussion_r1782490192.

PR #6251, opened 2024-10-02
refactor(transformer)!: remove `source_type` param from `Transformer::new`

### LINKED ISSUE #6248: Transformer don't take `source_type` param
Currently `Transformer::new` takes a `source_type` param:

https://github.com/oxc-project/oxc/blob/bc757c89b4974c051fbd660a843e867d864736f1/crates/oxc_transformer/src/lib.rs#L76-L86

Can we get rid of it? `source_type` is present in `Program` and parser updates it from "unambiguous" to either "script" or "module".

It seems to make more sense to use the value of `source_type` present in `Program`, than allow user to over-ride it.

I can't see any circumstance in which that's a useful thing to do, and it's a potential footgun. It's confusing that in transformer it's possible that `self.ctx.source_type.is_script() != !self.ctx.source_type.is_module()`, because it can also be "unambiguous". That's not theoretical - I hit this problem while trying to pass some TypeScript tests.

If for some obscure reason, the user *does* want to override it, they can always alter the value of `Program::source_type` before passing `Program` to transformer.

overlookmotel:
Closes #6248.

### DIFF
--- a/crates/oxc/src/compiler.rs
+++ b/crates/oxc/src/compiler.rs
@@ -137,7 +137,6 @@ pub trait CompilerInterface {
                 &mut program,
                 source_path,
                 source_text,
-                source_type,
                 &trivias,
                 symbols,
                 scopes,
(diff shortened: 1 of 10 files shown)


overlookmotel:
* **#6251** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6251?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6258, opened 2024-10-02
perf(linter): use fastcmp for string comparisons

DavidHancu:
This PR introduces a fixed fork of `fastcmp`. It is used for faster string comparisons for strings under 256 bytes.

See https://github.com/oxc-project/backlog/issues/46#issuecomment-2386434062.


overlookmotel:
Damn! Can I ask, why did you have to fork `fastcmp` rather than just using the original crate? The devil is sometimes in the small details on these things.

overlookmotel:
See discussion on Discord: https://discord.com/channels/1079625926024900739/1281761615833993279

Turns out `fastcmp` is a pretty dodgy library which does unaligned reads - which is UB. The forked version in this PR fixed that by using `read_unaligned`. But then it's not fast any more.

PR #6270, opened 2024-10-03
fix(semantic): enter scope after `check_type` for `TSConditionalType`

DonIsaac:
When visiting a `TSConditionalType`, enter a new scope after visiting the `check_type`, and exit before entering the `false_type`.

Fixes this case, where `B` should have a reference. I can't find the issue that was reported with this, if someone finds it please link it to this PR.
```typescript
type S<A> = A extends /* enter scope */ (infer B extends number ? string : never)
  ? B 
  : /* exit scope before false branch */ false;
```

Ideally we'd update `ast_tools` to let us configure the order of enter/leave scopes. CC: @overlookmotel.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -345,12 +345,36 @@ pub struct TSConditionalType<'a> {
     #[serde(flatten)]
     pub span: Span,
     /// The type before `extends` in the test expression.
+    ///
+    /// ## Example
+    /// ```ts
+    /// type F<T> = T extends string ? string : number;
+    /// //          ^
+    /// ```
     pub check_type: TSType<'a>,
     /// The type `check_type` is being tested against.
+    ///
+    /// ## Example
+    /// ```ts
+    /// type F<T> = T extends string ? string : number;
+    /// //                    ^^^^^^
+    /// ```
     pub extends_type: TSType<'a>,
     /// The type evaluated to if the test is true.
+    ///
+    /// ## Example
+    /// ```ts
+    /// type F<T> = T extends string ? string : number;
+    /// //                             ^^^^^^
+    /// ```
     pub true_type: TSType<'a>,
     /// The type evaluated to if the test is false.
+    ///
+    /// ## Example
+    /// ```ts
+    /// type F<T> = T extends string ? string : number;
+    /// //                                      ^^^^^^
+    /// ```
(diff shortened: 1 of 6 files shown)


DonIsaac:
* **#6270** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6270?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
> Ideally we'd update `ast_tools` to let us configure the order of enter/leave scopes.

You can (and should). Add `#[scope(enter_before)]` to the `extends_type` field. Same as it is in `TSEnumDeclaration`:

https://github.com/oxc-project/oxc/blob/50a0029a97c4f9d5f33faafce29f3271b6f2f5e9/crates/oxc_ast/src/ast/ts.rs#L85-L90

If you do that, `ast_tools` will update `Visit` and `VisitMut` automatically, and you don't need to add `visit_ts_conditional_type` method to `impl<'a> Visit<'a> for SemanticBuilder<'a>`.

Note: Run `just ast` to run `ast_tools` codegen after adding the attribute.

DonIsaac:
> > Ideally we'd update `ast_tools` to let us configure the order of enter/leave scopes.
> 
> You can (and should). Add `#[scope(enter_before)]` to the `extends_type` field. Same as it is in `TSEnumDeclaration`:
> 
> https://github.com/oxc-project/oxc/blob/50a0029a97c4f9d5f33faafce29f3271b6f2f5e9/crates/oxc_ast/src/ast/ts.rs#L85-L90
> 
> If you do that, `ast_tools` will update `Visit` and `VisitMut` automatically, and you don't need to add `visit_ts_conditional_type` method to `impl<'a> Visit<'a> for SemanticBuilder<'a>`. It'll also make `Traverse` enter the scope at the right point.
> 
> Note: Run `just ast` to run `ast_tools` codegen after adding the attribute.

Is there also an `#[scope(exit_after)]`?

overlookmotel:
> Is there also an `#[scope(exit_after)]`?

No. We haven't found a use for that. But it wouldn't be hard to add if we need it.

DonIsaac:
We need that to exit the scope before the false branch, since inferred type parameters (e.g. `infer R`) are only in scope within the true branch.

DonIsaac:
Closed in favor of #6351

PR #6276, opened 2024-10-04
perf(linter): do not concat vec in `no-useless-length-check`

camchenry:
Probably a very small win, but rather than making lots of small vecs and concatenating, we should just use one vec and then push to it as we visit nodes.


camchenry:
* **#6276** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6276?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6283, opened 2024-10-04
fix(minifier): correct the reference link

shulaoda:
These are minor changes, and recently I’ve been learning and trying to implement some features of `oxc_minifier`, which feels a bit complex.

By the way, I’m wondering whether we should gradually add test cases during the feature implementation process or just copy all the corresponding tests directly? Perhaps we could implement something like `rulegen` similar to `linter`?


Boshen:
By the way, I’m wondering whether we should gradually add test cases during the feature implementation process or just copy all the corresponding tests directly?

I should've copied all the tests and ignore them first 😅 

Perhaps we could implement something like rulegen similar to linter?

There's not many test files, we can just copy them by hand.

PR #6313, opened 2024-10-06
fix(transformer): exponentiation transform: do not replace object when private property

overlookmotel:
Fix exponentiation operator transform to bail out early if a private class property.

We can't transform this:

```js
class C {
    #p;
    method() {
        this.#p **= 2;
    }
}
```

But we should at least leave it alone. Previously `get_obj_ref` called `ast.move_expression(expr)` on the member expression's object before bailing out, so example above was transformed to:

```js
class C {
    #p;
    method() {
        null.#p **= 2; // <-- `null`
    }
}
```

This PR makes it spot this case early and bail out *before* calling `ast.move_expression(expr)`.

### DIFF
--- a/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
+++ b/crates/oxc_transformer/src/es2016/exponentiation_operator.rs
@@ -171,10 +171,10 @@ impl<'a, 'ctx> ExponentiationOperator<'a, 'ctx> {
                 (ident, obj)
             }
             match_member_expression!(AssignmentTarget) => {
-                let obj = self.get_obj_ref(node, nodes, ctx).unwrap();
+                let obj = self.get_obj_ref(node, nodes, ctx)?;
                 let member_expr = node.to_member_expression_mut();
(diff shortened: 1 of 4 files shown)


overlookmotel:
* **#6313** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
I thought this logic is part of that super weird secondary visit we discovered in babel with @Dunqing

overlookmotel:
> I thought this logic is part of that super weird secondary "environment visitor" we discovered in babel with @Dunqing

I wasn't completely sure what that "environment visitor" thing was, but my impression is that it's just a perf optimization.

What's meant to happen is it throws an error if you've not also enabled the class private properties transform. But we can't do that at the moment, as we've not implemented class transforms yet.

But what we definitely *shouldn't* be doing is raising no error and silently converting `obj.#p` into `null.#p`! That's what this PR deals with.

PR #6328, opened 2024-10-07
docs(traverse): remove erroneous doc comment

overlookmotel:
`TraverseCtx::generate_uid_in_current_scope_based_on_node` no longer delegates to a method on `TraverseScoping`, so this comment was outdated.


overlookmotel:
Docs only change, so merging without review.

PR #6342, opened 2024-10-07
feat(linter): implement `react/style-prop-object`

nekosaur:
Implements not recommended rule `react/style-prop-object` (https://github.com/oxc-project/oxc/issues/1022)


crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,428 @@
...
+    };
+
+    match expression {
+        Expression::StringLiteral(_) | Expression::BooleanLiteral(_) => true,
+        Expression::Identifier(ident) => {
+            let Some(node) = get_declaration_of_variable(ident, ctx) else {
+                return false;
+            };
+
+            let AstKind::VariableDeclarator(var_decl) = node.kind() else {
```
DonIsaac, line 91: Consider checking `FormalParameter`s as well

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,428 @@
...
+                return false;
+            };
+
+            return is_invalid_expression(var_decl.init.as_ref(), ctx);
```
DonIsaac, line 95: You may be able to check type annotations here for simple cases. See `no-throw-literal` for an example.

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,428 @@
...
+    ast_util::get_declaration_of_variable,
+    context::{ContextHost, LintContext},
+    rule::Rule,
+    utils::is_create_element_call,
+    AstNode,
+};
+
+fn style_prop_object_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Style prop value must be an object")
+        .with_help("Make sure the 'style' prop value is an object")
```
DonIsaac, line 22: This help message is redundant and adds no extra information that the diagnostic message does not already provide. Consider removing it or providing more information (e.g. what type we think `style` is, if not an object)

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+declare_oxc_lint!(
+    /// ### What it does
+    /// Require that the value of the prop style be an object or a variable that is an object.
+    ///
```
DonIsaac, line 45: Missing "Why is this bad?" section.

crates/oxc_linter/src/snapshots/style_prop_object.snap
```diff
@@ -0,0 +1,64 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint-plugin-react(style-prop-object): Style prop value must be an object
+   ╭─[style_prop_object.tsx:1:6]
+ 1 │ <div style="color: 'red'" />
```
DonIsaac, line 6: Shrink the span to only cover the attribute's value

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,428 @@
...
+    fn from_configuration(value: serde_json::Value) -> Self {
+        Self(Box::new(StylePropObjectConfig {
+            allow: value
+                .get(0)
+                .and_then(|v| v.get("allow"))
+                .and_then(serde_json::Value::as_array)
+                .map(|v| {
+                    v.iter()
+                        .filter_map(serde_json::Value::as_str)
+                        .map(ToString::to_string)
```
DonIsaac, line 131: use `CompactStr` instead of `String`s

DonIsaac: Also, if you sort this, you can use a binary search for `O(log(n))` containment checks (instead of `.contains`, which is `O(n)`.

DonIsaac:
Thanks for your interest in Oxc! This PR will be a great contribution. Just a few small fixes and we should be able to merge this.

crates/oxc_linter/src/rules/react/style_prop_object.rs
nekosaur: I don't have a great grasp of the AST yet, but can FormalParameters also be something other than function parameters?

crates/oxc_linter/src/rules/react/style_prop_object.rs
nekosaur: Do you mean as a cheaper way instead of checking `.init`?

crates/oxc_linter/src/rules/react/style_prop_object.rs
DonIsaac: They're used in type parameters, but I don't think that's relevant here.

crates/oxc_linter/src/rules/react/style_prop_object.rs
DonIsaac:
```ts
function SomeComponent(props) {
  let foo: string | undefined
  if (bar) foo = 'foo'
  return <div style={foo} />
}
```

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,497 @@
...
+        TSType::TSNumberKeyword(_) | TSType::TSStringKeyword(_) | TSType::TSBooleanKeyword(_) => {
+            true
+        }
+        TSType::TSUnionType(union) => union.types.iter().any(is_invalid_type),
+        TSType::TSIntersectionType(intersect) => intersect.types.iter().any(is_invalid_type),
+        _ => false,
+    }
+}
+
+fn is_invalid_expression(expression: Option<&Expression>, ctx: &LintContext<'_>) -> bool {
```
DonIsaac, line 91:
```suggestion
fn is_invalid_expression<'a>(expression: Option<&Expression<'a>>, ctx: &LintContext<'a>) -> bool {
```

crates/oxc_linter/src/rules/react/style_prop_object.rs
```diff
@@ -0,0 +1,497 @@
...
+                return is_invalid_type(&asd.type_annotation);
+            };
+
+            return is_invalid_expression(var_decl.init.as_ref(), ctx);
+        }
+        _ => false,
+    }
+}
+
+fn is_invalid_style_attribute(attribute: &JSXAttribute, ctx: &LintContext<'_>) -> bool {
```
DonIsaac, line 119:
```suggestion
fn is_invalid_style_attribute<'a>(attribute: &JSXAttribute<'a>, ctx: &LintContext<'a>) -> bool {
```

PR #6348, opened 2024-10-07
fix(linter): `no-else-return` fixer fails when `else` has no trailing whitespace

DonIsaac:
Fixes a bug in `eslint/no_else_return`'s fixer where fixes were not being
property applied when `else` had no whitespace immediately after it. For
example:
```js
if(x){ return x }else{ return y }
```

I also refactored the rule's fixer to avoid string allocations as much as
possible.


DonIsaac:
* **#6348** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6348?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camc314 (approved):

PR #6350, opened 2024-10-07
feat(ast_tools): support `#[scope(exit_before)]`

### LINKED ISSUE #6311: Support exiting scope before last field `#[scope(exit_before)]`
`oxc_ast_tools` recognises `#[scope(enter_before)]` to mark a field as the start of the scope in a type. We also need `#[scope(exit_before)]` for #6270.

Note: I think `#[scope(exit_before)]` is better than `#[scope(exit_after)]` because attributes go before fields, so it reads more clearly:

```rs
pub struct TSConditionalType<'a> {
    pub span: Span,
    pub check_type: TSType<'a>,
    #[scope(enter_before)]
    pub extends_type: TSType<'a>,
    pub true_type: TSType<'a>,
    #[scope(exit_before)]
    pub false_type: TSType<'a>,
    pub scope_id: Cell<Option<ScopeId>>,
}
```

We will need to build support for this into both:

1. `oxc_ast_tools` and the `Visit` / `VisitMut` generator.
2. `oxc_traverse`'s codegen.

DonIsaac:
Closes #6311.

Adds support for `#[scope(exit_before)]`, which is the opposite of `#[scope(enter_before)]`

### DIFF
--- a/crates/oxc_traverse/scripts/lib/walk.mjs
+++ b/crates/oxc_traverse/scripts/lib/walk.mjs
@@ -75,7 +75,9 @@ function generateWalkForStruct(type, types) {
 
   const { scopeArgs } = type;
   /** @type {Field | undefined} */
-  let scopeEnterField;
+  let scopeEnterField,
+    /** @type {Field | undefined} */
+    scopeExitField;
   let enterScopeCode = '', exitScopeCode = '';
 
   if (scopeArgs && scopeIdField) {
@@ -90,6 +92,17 @@ function generateWalkForStruct(type, types) {
       );
     } else {
       scopeEnterField = visitedFields[0];
+    }
+
+    // Get field to exit scope before
+    const exitFieldName = scopeArgs.exitScopeBefore;
+    if (exitFieldName) {
+      scopeExitField = visitedFields.find(field => field.name === exitFieldName);
+      assert(
+        scopeExitField,
+        `\`ast\` attr says to exit scope before field '${exitFieldName}' ` +
+          `in '${type.name}', but that field is not visited`,
+      );
     }
 
     // TODO: Maybe this isn't quite right. `scope_id` fields are `Cell<Option<ScopeId>>`,
@@ -107,7 +120,11 @@ function generateWalkForStruct(type, types) {
   const fieldsCodes = visitedFields.map((field, index) => {
     const fieldWalkName = `walk_${camelToSnake(field.innerTypeName)}`,
       fieldCamelName = snakeToCamel(field.name);
-    const scopeCode = field === scopeEnterField ? enterScopeCode : '';
+    const scopeCode = field === scopeEnterField
+      ? enterScopeCode
+      : field === scopeExitField
+      ? exitScopeCode
+      : '';
 
     let tagCode = '', retagCode = '';
     if (index === 0) {
@@ -212,7 +229,7 @@ function generateWalkForStruct(type, types) {
     ) {
       traverser.enter_${typeSnakeName}(&mut *node, ctx);
       ${fieldsCodes.join('\n')}
-      ${exitScopeCode}
+      ${scopeExitField ? '' : exitScopeCode}
       traverser.exit_${typeSnakeName}(&mut *node, ctx);
     }
   `.replace(/\n\s*\n+/g, '\n');
--- a/tasks/ast_tools/src/generators/visit.rs
+++ b/tasks/ast_tools/src/generators/visit.rs
@@ -464,6 +464,7 @@ impl<'a> VisitBuilder<'a> {
         };
 
         let mut enter_scope_at = 0;
+        let mut exit_scope_at: Option<usize> = None;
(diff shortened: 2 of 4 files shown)


DonIsaac:
* **#6350** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing:
From the description of #6311, we should be supporting `#[scope(exit_before)]` rather than `#[scope(exit_after]`. Let me know if I got it wrong.

overlookmotel (changes requested):
As mentioned in #6311, I think `#[scope(exit_before)]` is better than `#[scope(exit_after)]`, because then when you read the type def, the `#[scope(exit_before)]` attr is at the point where the scope is exited.

e.g.:

```rs
pub struct TSConditionalType<'a> {
    pub span: Span,
    pub check_type: TSType<'a>,
    #[scope(enter_before)]
    pub extends_type: TSType<'a>,
    pub true_type: TSType<'a>,
    #[scope(exit_before)]
    pub false_type: TSType<'a>,
    pub scope_id: Cell<Option<ScopeId>>,
}
```

The fields that are inside the scope are `extends_type` and `true_type`, and they are bounded by the `#[scope(...)]` attrs on each side.

Is there a reason why you decided against that?

overlookmotel (dismissed):
`Traverse` codegen also needs to support the new exit attr.

https://github.com/oxc-project/oxc/tree/main/crates/oxc_traverse/scripts

You can see in #6351 that `walk.rs` is not exiting the scope when it should: https://github.com/oxc-project/oxc/pull/6351/files#diff-f70f8cd02df83ed234c9b27dcb7ea2c948b2ab6110952e87eadaf0ddafbd6b9a

tasks/ast_tools/src/markers.rs
```diff
@@ -64,7 +64,10 @@ pub struct VisitMarkers {
 /// A struct representing `#[scope(...)]` markers
 #[derive(Default, Debug)]
 pub struct ScopeMarkers {
+    /// `#[scope(enter_before)]`
```
Boshen, line 67: Unintentional commented out code?

overlookmotel: I believe this is intentional. It's a doc comment explaining what attribute `ScopeMarkers::enter_before` represents.

DonIsaac:
> `Traverse` codegen also needs to support the new exit attr.
> 
> https://github.com/oxc-project/oxc/tree/main/crates/oxc_traverse/scripts
> 
> You can see in #6351 that `walk.rs` is not exiting the scope when it should: https://github.com/oxc-project/oxc/pull/6351/files#diff-f70f8cd02df83ed234c9b27dcb7ea2c948b2ab6110952e87eadaf0ddafbd6b9a

Fixed. Ready for another review.

crates/oxc_traverse/scripts/lib/walk.mjs
```diff
@@ -92,6 +94,20 @@ function generateWalkForStruct(type, types) {
...
+      exitScopeField = visitedFields.find(field => field.name === exitFieldName);
+      assert(
+        exitScopeField,
+        `\`ast\` attr says to exit scope before field '${exitFieldName}' ` +
+          `in '${type.name}', but that field is not visited`,
+      );
+    }
+    // else {
+    //   exitScopeField = visitedFields[visitedFields.length - 1];
+    // }
```
overlookmotel, line 109: Remove commented-out code.

```suggestion
```

crates/oxc_traverse/scripts/lib/walk.mjs
```diff
@@ -75,7 +75,9 @@ function generateWalkForStruct(type, types) {
 
   const { scopeArgs } = type;
   /** @type {Field | undefined} */
-  let scopeEnterField;
+  let scopeEnterField,
+    /** @type {Field | undefined} */
+    exitScopeField;
```
overlookmotel, line 80: Could you name this var `scopeExitField` for consistency with `scopeEnterField`?

tasks/ast_tools/src/generators/visit.rs
```diff
@@ -525,6 +527,18 @@ impl<'a> VisitBuilder<'a> {
...
+                if have_exit_scope {
+                    assert!(
+                        exit_scope_at.is_none(),
+                        "Scopes cannot be exited more than once. Remove the extra `#[scope(exit_before)]` attribute(s)."
+                    );
+                    let scope_leave = &scope_events.1;
+                    result = quote! {
+                        #scope_leave
+                        #result
+                    };
```
overlookmotel, line 539: Rename var for consistency.

```suggestion
                    let scope_exit = &scope_events.1;
                    result = quote! {
                        #scope_exit
                        #result
                    };
```

overlookmotel (commented):
Sorry for the nits...

PR #6376, opened 2024-10-08
feat(linter): add suggestions for `no-plusplus`

camchenry:
Implements naive suggestions for the `no-plusplus` rule. This is not implemented as a fix, because there isn't an exact 1:1 equivalence between `++` and `+= 1`. But most cases should be autofixable by this, but I've marked it as dangerous suggestion due to the semantics change

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_plusplus.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_plusplus.rs
@@ -78,7 +78,10 @@ declare_oxc_lint!(
     /// ```
     NoPlusplus,
     restriction,
-    pending
+    // This is not guaranteed to rewrite the code in a way that is equivalent.
+    // For example, `++i` and `i++` will be rewritten as `i += 1` even though they are not the same.
+    // If the code depends on the order of evaluation, then this might break it.
+    conditional_suggestion,
 );
 
 impl Rule for NoPlusplus {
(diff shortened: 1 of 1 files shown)


camchenry:
* **#6376** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/rules/eslint/no_plusplus.rs
```diff
@@ -78,7 +78,10 @@ declare_oxc_lint!(
     /// ```
     NoPlusplus,
     restriction,
-    pending
+    // This is "dangerous" because this will rewrite the code in a way that is not equivalent.
+    // For example, `++i` and `i++` will be rewritten as `i += 1` even though they are not the same.
+    // If the code depends on the order of evaluation, then this might break it.
+    dangerous_suggestion,
```
DonIsaac, line 84:
```suggestion
    conditional_dangerous_suggestion,
```

DonIsaac: We could even make this `conditional_suggestion`, since this is unlikely to create semantically/syntactically invalid JavaScript.

camchenry: fair, there is part of me that fears this creating some really strange regressions, but I guess labeling it as a "suggestion" already implies it's not a guaranteed fix

DonIsaac (approved):

PR #6388, opened 2024-10-09
fix(semantic): correctly resolve binding for return type of functions

### LINKED ISSUE #6387: Return type of function incorrectly resolved
```ts
export type T = number;

function foo(n: number): T {
  type T = string;
  return n;
}
```

`T` in return type of `foo` should resolve to top level `export type T`, but instead gets resolved to `type T` inside the function body. [Oxc Playground](https://playground.oxc.rs/#eNpFTjtuwzAMvYrAqQWMHkBF18wdOnqRXSoQIJECSaUxDN+9chrXE/n4flxhBg94ryzmbKnovtyHo1YmlPeRRoqNZktMLjK/kH9Sr77r1pHc6VGTRNfucU7QmpCjDraRYAAGv4I02ocuZOEO3qThADmRgY8hawc6c8WD0aVMnA9kEkgjS3mKtwFqEEV5RHKTGS8pI4XSA8BQ7c20N1fBLrrhZxAkPc1778Pc9//s4zDzN15x/7aDkijFdFJkwvmS+WdvvqFMrL3yL3fbfgG1h3ai)

TypeScript resolves it to the top-level `T`. [TS Playground](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcAqcC8cB2BXAWwCNgoBuAKAoDM8cBjGASwhzmoggAocAuXQiSgBKfhgDeFOAmSoM2AM4woTHAHNK0qMBh4obHJQC+FIA). If it was resolving to `type T`, the code would fail type check.

overlookmotel:
Fixes #6387.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1655,6 +1655,18 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         if let Some(return_type) = &func.return_type {
             self.visit_ts_type_annotation(return_type);
         }
+
+        if func.params.has_parameter() || func.return_type.is_some() {
+            // `function foo({bar: identifier_reference}) {}`
+            //                     ^^^^^^^^^^^^^^^^^^^^
+            // `function foo<SomeType>(v: SomeType): SomeType { return v; }`
+            //                            ^^^^^^^^   ^^^^^^^^
+            // Parameter initializers must be resolved after all parameters have been declared.
+            // Param types and return type must be resolved after type parameters have been declared.
+            // In both cases, need to avoid binding to variables/types declared inside the function body.
+            self.resolve_references_for_current_scope();
+        }
+
         if let Some(body) = &func.body {
             self.visit_function_body(body);
         }
@@ -1717,6 +1729,17 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
 
         if let Some(return_type) = &expr.return_type {
             self.visit_ts_type_annotation(return_type);
+        }
+
+        if expr.params.has_parameter() || expr.return_type.is_some() {
+            // `let foo = ({bar: identifier_reference}) => {};`
+            //                   ^^^^^^^^^^^^^^^^^^^^
+            // `let foo = <SomeType>(v: SomeType): SomeType => v;`
+            //                          ^^^^^^^^   ^^^^^^^^
+            // Parameter initializers must be resolved after all parameters have been declared.
+            // Param types and return type must be resolved after type parameters have been declared.
+            // In both cases, need to avoid binding to variables/types declared inside the function body.
+            self.resolve_references_for_current_scope();
         }
 
         self.visit_function_body(&expr.body);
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#6388** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6388?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
...or at least I *think* it's a fix. @Dunqing could you please help me with this one?

Questions:

1. There are various TS types which contain `FormalParameters`:

* `TSCallSignatureDeclaration`
* `TSMethodSignature`
* `TSConstructSignatureDeclaration`
* `TSFunctionType`
* `TSConstructorType`

None of these function types have a body, so I assume we're OK to skip the extra call to `resolve_references_for_current_scope` for them, right?

2. Is the `func.params.kind != FormalParameterKind::Signature` check required for `Function` and `ArrowFunctionExpression`? Or can their params never have `kind: FormalParameterKind::Signature` anyway?

3. How/where would I add tests for this?

Dunqing:
> Questions:
> 
> 1. There are various TS types which contain `FormalParameters`:
> 
> * `TSCallSignatureDeclaration`
> * `TSMethodSignature`
> * `TSConstructSignatureDeclaration`
> * `TSFunctionType`
> * `TSConstructorType`
> 
> None of these function types have a body, so I assume we're OK to skip the extra call to `resolve_references_for_current_scope` for them, right?
>
 
Right

> 2. Is the `func.params.kind != FormalParameterKind::Signature` check required for `Function` and `ArrowFunctionExpression`? Or can their params never have `kind: FormalParameterKind::Signature` anyway?

Skipping `func.params.kind != FormalParameterKind::Signature` is ok

> 3. How/where would I add tests for this?

￼I think as long as there are no test failures, it’s fine, so we don’t need to add tests for it.

crates/oxc_semantic/src/builder.rs
```diff
@@ -1726,6 +1738,17 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
             self.visit_ts_type_annotation(return_type);
         }
 
+        if expr.params.kind != FormalParameterKind::Signature && expr.params.has_parameter() {
```
Dunqing, line 1741: If I remember correctly, the `FormalParameterKind` of `ArrowFunctionExpression` and `Function` never going to be  `FormalParameterKind::Signature`, so if you want to call `resolve_references_for_current_scope` here, I think you don't need to check anymore.

Dunqing (changes requested):

overlookmotel:
> > 2. Is the `func.params.kind != FormalParameterKind::Signature` check required for `Function` and `ArrowFunctionExpression`? Or can their params never have `kind: FormalParameterKind::Signature` anyway?
> 
> Skipping `func.params.kind != FormalParameterKind::Signature` is ok

Great. I've removed that.

> > 3. How/where would I add tests for this?
> 
> ￼I think as long as there are no test failures, it’s fine, so we don’t need to add tests for it.

But #6387 was a bug, and this PR fixes it. So there should have been a test failing before, but there wasn't. So I think we should add one. I just don't know where to put it. Can you point me in right direction please?

Dunqing:
> But #6387 was a bug, and this PR fixes it. So there should have been a test failing before, but there wasn't. So I think we should add one. I just don't know where to put it. Can you point me in right direction please?

Oh, sorry, I didn't notice that. We have two ways to write semantic tests:

1. The snapshot-based tests, you can add a test [here](https://github.com/oxc-project/oxc/blob/93575cd09e626a783c780d244996a447099e5782/crates/oxc_semantic/tests/fixtures/oxc), and then run https://github.com/oxc-project/oxc/blob/c822b48d4f7eedf4db803d158cfe7bcde24c25ea/crates/oxc_semantic/tests/main.rs#L158-L172

2. The integration test, you can add [here](https://github.com/oxc-project/oxc/blob/c822b48d4f7eedf4db803d158cfe7bcde24c25ea/crates/oxc_semantic/tests/integration/symbols.rs#L493)

Either two ways are ok. I prefer the first way because I can see what scopes/symbols have in the test code

Dunqing:
Let me add tests to it

Dunqing (approved):

Boshen:
### Merge activity

* **Nov 26, 10:41 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
Thank you very much for cleaning up the mess I left!

PR #6391, opened 2024-10-09
refactor(ast): type params field before params in TS function declaration types

overlookmotel:
Our convention is that AST type fields are ordered in order they appear in source. Move `type_parameters` fields in TS function declaration types to before `this_param` and formal parameters.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -1147,10 +1147,10 @@ pub struct TSIndexSignature<'a> {
 pub struct TSCallSignatureDeclaration<'a> {
     #[serde(flatten)]
     pub span: Span,
+    pub type_parameters: Option<Box<'a, TSTypeParameterDeclaration<'a>>>,
     pub this_param: Option<TSThisParameter<'a>>,
     pub params: Box<'a, FormalParameters<'a>>,
     pub return_type: Option<Box<'a, TSTypeAnnotation<'a>>>,
-    pub type_parameters: Option<Box<'a, TSTypeParameterDeclaration<'a>>>,
 }
 
 #[ast]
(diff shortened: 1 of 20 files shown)


overlookmotel:
What is odd is that, because prior to #6388, we resolved references in `FormalParameters` early, I imagine `U` in the function params and return type in this example would not have been resolved correctly:

```ts
type T = <U>(x: U) => U;
```

But, if I'm right about that, our tests did not catch this. Am I wrong? Or do we not have enough test coverage?

@Dunqing calling on your TS expertise!

overlookmotel:
PS: I am *slowly* learning TypeScript!

Boshen:
> Our convention is that AST type fields are ordered in order they appear in source. 

In source or runtime order? This should be documented.

Dunqing:
> What is odd is that, because prior to #6388, we resolved references in `FormalParameters` early, I imagine `U` in the function params and return type in this example would not have been resolved correctly:
> 
> ```ts
> type T = <U>(x: U) => U;
> ```
> 
> But, if I'm right about that, our tests did not catch this. Am I wrong? Or do we not have enough test coverage?
> 
> @Dunqing calling on your TS expertise!

Well, good question, we have enough tests to cover these ASTs. This case is correct before, check out the [Playground](https://playground.oxc.rs/#eNpFjkEOwjAMBL8S+QRSXxBoj7yA3rikxUWRUrtyUmhU5e8kQikn79jW7u4wgoYQF1R31apr3502rfqzajvVXx4EDTDoHWSlMnykYDbQQVZswFkKVfuRFzwgzgO7SkEM+YllBj0Z5zE1sBjxKNkx6+JS9fFaFyM/8YUlO8NsyU72f6Ig7G6OP6XaG2Vgnyv8MlL6AqCcSv4=)

### Why resolve correctly even if the order of the fields is incorrect

It is related to our resolving algorithms. Our process is collecting bindings/references first, and then resolving references in `exit_scope`.

So in `type T = <U>(x: U) => U;`, the resolving steps are as follows.

1. enter_scope (TypeAlias)
2. declare reference `x` (FormalParameter) 
3. declare reference `U` (FormalParameter's TSTypeAnnotation) 
4. declare reference `U` (ReturnType) 
5. bind `U` (TSTypeParameter)
6. exit_scope  (calling `resolve_references_for_current_scope` here)

Based on the above steps, we know that we have known all the bindings before resolving so that the references are resolved correctly no matter what the order of fields. 

But attention if there is a nested scope, the order is crucial.

overlookmotel:
> In source or runtime order? This should be documented.

I've opened #6453 to discuss this question. But in this particular case, the concept of "runtime order" is a bit slippery - TS types *don't even exist* at runtime - they're a compile-time concept only.

So I think regardless of the answer to #6453, the change this PR makes is good. Either way, `type_parameters` being last is definitely *not* correct.

Dunqing (approved):
LGTM

overlookmotel:
### Merge activity

* **Oct 11, 5:56 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
> LGTM

Thanks Dunqing!

PR #6393, opened 2024-10-09
feat(editors/vscode): change default config path from .eslintrc to .oxlintrc.json

### LINKED ISSUE #5802: VSCode extension: default `oxc_language_server.configPath` seems wrong
Hi 👋

While trying to move part of my linting config from eslint to oxlint, I did not understand why removing the `no-console` rule from `.eslintrc` & adding it to `.oxlintrc.json` did not trigger an error in my VSCode editor while it was triggering it properly running the oxlint CLI.

After a short investigating, I discovered that the default config path used by the OXC VSCode language server was [`.eslintrc`](https://github.com/oxc-project/oxc/blob/bb95306efda4af7f49fb9c393e834c299c6142da/crates/oxc_language_server/src/main.rs#L54).

It means that the oxlint language server was using my legacy eslint config instead of the new appropriate one. Should we change the default config path to be the filename documented [here](https://oxc.rs/docs/guide/usage/linter/config.html) or at least explain that the configuration file used can be changed using VSCode `settings.json` file? Many thanks!

pumano:
closes #5802

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -51,7 +51,7 @@ struct Options {
 
 impl Default for Options {
     fn default() -> Self {
-        Self { enable: true, run: Run::default(), config_path: ".eslintrc".into() }
+        Self { enable: true, run: Run::default(), config_path: ".oxlintrc.json".into() }
     }
 }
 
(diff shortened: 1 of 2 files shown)


DonIsaac:
Conversation thread is on #5802

Issue #6347, opened 2024-10-04
Implement `serde::Serialize` on AST types via `#[generate_derive]`


overlookmotel:
We currently use `serde`'s derive macros to implement `Serialize` on AST types.

We could use `#[generate_derive]` to generate these impls instead.

### 1. Reduce compile time

`serde`'s macro is pretty expensive at compile time for the NAPI build. We can remove it.

### 2. Reduce boilerplate

`serde`'s derive macro is less powerful than `ast_tools`. Because `Serialize` is a macro, all it knows about is the type that `#[derive(Serialize)]` is on. Whereas `ast_tools` builds a schema of the entire AST, so it knows not just about the type it's deriving impl for, but also all the other types too, and how they link to each other.

Currently we have to put `#[serde]` attributes everywhere:

```rs
#[ast]
#[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
#[serde(tag = "type")]
pub struct ClassBody<'a> {
    #[serde(flatten)]
    pub span: Span,
    pub body: Vec<'a, ClassElement<'a>>,
}

#[ast]
#[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
#[serde(tag = "type")]
pub struct PrivateIdentifier<'a> {
    #[serde(flatten)]
    pub span: Span,
    pub name: Atom<'a>,
}

#[ast]
#[cfg_attr(feature = "serialize", derive(Serialize, Tsify))]
pub struct Span {
    start: u32,
    end: u32,
}
```

Instead, we can use `ast_tools` in 2 ways to remove this boilerplate:

1. Make things that we implement on every type the defaults, so they don't need to be stated over and over.
2. Use `ast_tools`'s knowledge of the whole AST to move the instruction to flatten `Span` onto `Span` type itself. "flatten this" instruction does not need to be repeated on every type that contains `Span`.

```rs
#[ast]
#[generate_derive(ESTree)]
pub struct ClassBody<'a> { // <-- no `#[serde(tag = "type")]` attr
    pub span: Span, // <-- no `#[serde(flatten)]` attr
    pub body: Vec<'a, ClassElement<'a>>,
}

#[ast]
#[generate_derive(ESTree)]
pub struct PrivateIdentifier<'a> { // <-- no `#[serde(tag = "type")]` attr
    pub span: Span, // <-- no `#[serde(flatten)]` attr
    pub name: Atom<'a>,
}

#[ast]
#[generate_derive(ESTree)]
[...]

PR #6404, opened 2024-10-09
feat(parser): Generate `Serialize` impls in ast_tools

ottomated:
Beginning of #6347. Instead of using serde-derive, we generate `Serialize` impls manually.


Boshen:
Nice, I believe this'll make compilation faster as well, because serde no longer runs `syn`.

ottomated:
Tested on `checker.ts`. Diff below. The only changes are in regex serialization which is incorrect anyway so should be an issue.

```diff
--- a/expected.json
+++ b/actual.json
@@ -20457,11 +20457,9 @@
             "type": "RegExpLiteral",
             "start": 29848,
             "end": 29856,
-            "value": null,
+            "value": {},
             "regex": {
-              "pattern": {
-                "Raw": "^\".+\"$"
-              },
+              "pattern": "^\".+\"$",
               "flags": ""
             }
           },
@@ -247915,11 +247913,9 @@
                                                                     "type": "RegExpLiteral",
                                                                     "start": 404540,
                                                                     "end": 404545,
-                                                                    "value": null,
+                                                                    "value": {},
                                                                     "regex": {
-                                                                      "pattern": {
-                                                                        "Raw": "\\n"
-                                                                      },
+                                                                      "pattern": "\\n",
                                                                       "flags": "g"
                                                                     }
                                                                   },
@@ -256454,11 +256450,9 @@
                                                   "type": "RegExpLiteral",
                                                   "start": 417226,
                                                   "end": 417238,
-                                                  "value": null,
+                                                  "value": {},
                                                   "regex": {
-                                                    "pattern": {
-                                                      "Raw": "\\r\\n|\\n|\\r"
-                                                    },
+                                                    "pattern": "\\r\\n|\\n|\\r",
                                                     "flags": ""
                                                   }
                                                 }
@@ -256546,11 +256540,9 @@
                                                           "type": "RegExpLiteral",
                                                           "start": 417265,
                                                           "end": 417271,
-                                                          "value": null,
+                                                          "value": {},
                                                           "regex": {
-                                                            "pattern": {
-                                                              "Raw": "^\\s+"
-                                                            },
+                                                            "pattern": "^\\s+",
                                                             "flags": ""
                                                           }
                                                         },
@@ -287745,11 +287737,9 @@
                                                         "type": "RegExpLiteral",
                                                         "start": 470730,
                                                         "end": 470736,
-                                                        "value": null,
+                                                        "value": {},
                                                         "regex": {
-                                                          "pattern": {
-                                                            "Raw": "\\\\."
-                                                          },
+                                                          "pattern": "\\\\.",
                                                           "flags": "g"
                                                         }
                                                       },
@@ -321794,11 +321784,9 @@
                                                               "type": "RegExpLiteral",
                                                               "start": 536344,
                                                               "end": 536349,
-                                                              "value": null,
+                                                              "value": {},
                                                               "regex": {
-                                                                "pattern": {
-                                                                  "Raw": "\\n"
-                                                                },
+                                                                "pattern": "\\n",
                                                                 "flags": "g"
                                                               }
                                                             },
@@ -358021,11 +358009,9 @@
                                             "type": "RegExpLiteral",
                                             "start": 606654,
                                             "end": 606667,
-                                            "value": null,
+                                            "value": {},
                                             "regex": {
-                                              "pattern": {
-                                                "Raw": "[^a-z0-9]"
-                                              },
+                                              "pattern": "[^a-z0-9]",
                                               "flags": "gi"
                                             }
                                           },
@@ -1212733,11 +1212719,9 @@
                                             "type": "RegExpLiteral",
                                             "start": 2037983,
                                             "end": 2038033,
-                                            "value": null,
+                                            "value": {},
                                             "regex": {
-                                              "pattern": {
-                                                "Raw": "^(?:EventTarget|Node|(?:HTML[a-zA-Z]*)?Element)$"
-                                              },
+                                              "pattern": "^(?:EventTarget|Node|(?:HTML[a-zA-Z]*)?Element)$",
                                               "flags": ""
                                             }
                                           },
```

ottomated:
Important changes:
- `#[derive(Serialize)]` -> `#[generate_derive(ESTree)]`
- `#[serde(...)]` -> `#[estree(...)]`
  - Supported on fields: `skip`, `flatten`, `rename`
  - Supported on enums: `rename_all`, `untagged` (in the future, I propose renaming this to something like `#[estree(transparent)]` for untagged enums, and maybe `#[estree(type_enum)]` for enums that shouldn't be camelCase)
  - Supported on structs: `type = "Identifier"`, `no_type` (for things that aren't really AST nodes and don't need a `"type"` field on the JSON
- By default, structs get the "type" field from the following (prioritized in this order):
1. `#[estree(type = "X")]` / `#[estree(no_type)]`
2. If the struct has an `r#type` field, it'll use that as the value
3. The name of the struct
Therefore, no need for `tag = "type"`.
- Struct fields are converted to camelCase by default. I verified that any struct missing `rename_all = "camelCase"` already had all its fields in camelCase format so this should not change anything.
- New `oxc_estree` crate that currently only contains an empty trait. I envision this being used for custom serialization and type generation in the future, but maybe a different approach is warranted.

crates/oxc_estree/src/lib.rs
Boshen, line 1: What's the rational for having a new crate versus feature gating this in `oxc_ast`?

Boshen (commented):
I like the direction this is going, good job 👍

crates/oxc_estree/src/lib.rs
ottomated: It needs to be accessible from oxc_span, oxc_regular_expression, and oxc_syntax as well as oxc_ast. Open to a better way of accomplishing this

overlookmotel:
Sorry it's taken me a while to get to reviewing this.

Excellent work. *Really* pleased to see this moving forwards. Thank you.

However, we do have one problem: The TS type defs generated for WASM are now inaccurate, because `Tsify` derive macro relies on `#[serde]` attrs, and they've been replaced with `#[estree]`, which `Tsify` doesn't understand.

Old TS def:

```ts
export interface NumericLiteral extends Span {
    type: "NumericLiteral";
    value: number;
    raw: string;
}
```

After this PR:

```ts
export interface NumericLiteral {
    span: Span;
    value: number;
    raw: string;
    base: NumberBase;
}
```

To see the generated TS types:

```sh
cd wasm/parser
pnpm install
pnpm run build
```

Type defs are generated at `npm/parser-wasm/node/oxc_parser_wasm.d.ts`.

I can see 2 ways to solve this:

### 1. Go back to `#[serde]`

Scale back the scope of this PR a little. Revert `#[estree]` to `#[serde]`.

Unfortunately this would also involve reverting the way `type` is dealt with, which is definitely an improvement, but is not compatible with `Tsify`.

We'd then try to replace `Tsify` as the next step, and *then* bring in `#[estree]` again.

### 2. Replace `Tsify` as well now

Replace `Tsify` and generate type defs in `oxc_ast_tools` instead.

I'm not sure how much work this would be. It's definitely not trivial, but maybe it's less hard than it initially looks.

`Tsify` derive macro produces a lot of code, but as we don't use `wasm-bindgen` for transferring these types to JS, I think the only part of `Tsify`'s output that we actually need is:

```rs
#[wasm_bindgen(typescript_custom_section)]
const TS_APPEND_CONTENT: &'static str =
    "export interface NumericLiteral extends Span {\n    value: number;\n    raw: string;\n}";
```

### Which?

Usually we try to keep PRs small and work incrementally, rather than long-lived branches which tend to acquire merge conflicts, and are hard to review. So that would point more towards approach (1) and that'd be my preference. However, @ottomated if you feel that 2nd approach is less work and you'd prefer to go that way, happy to go with your preference.

What do you think?

ottomated:
@overlookmotel yep, already noticed this and started work on replacing tsify. I was wondering, however, if anything's actually depending on the wasm types right now — are they published, could they be temporarily disabled? I should have the new type generation work done this week.

overlookmotel:
They are published publicly, so probably *someone* depends on them.

We try hard to keep main branch in a state where we can cut a release from main at any time (particularly useful for urgent bug fixes), so it's not really workable to merge this as is when we know it breaks things.

If you want to take the approach of replacing `Tsify` too, and think it's doable in that kind of timescale, I suggest we hold off merging this PR until that's done, and merge it all in one go. Hopefully there won't be too many merge conflicts in the meantime. Sound OK to you?

ottomated:
Sounds good. @overlookmotel what do you think about publishing a separate @types/oxc package that both wasm and napi depend on, and disabling wasm-bindgen type generation?

overlookmotel:
I'm not sure if it's necessary to publish types in a separate package, but yes end goal is that we generate our own types for the AST without the assistance of `Tsify` and use them for both the NAPI and WASM packages use.

For the `oxc-parser` NPM package (NAPI) I assumed we'd publish types as part of that package. For `@oxc-parser/wasm`, we need to somehow combine the AST types with the other types (`export function parseSync` etc), so easiest way may be to still go via wasm-bindgen.

But this area isn't my strong suit. I'm not that familiar with how wasm-bindgen works, so there may be a better way.

ottomated:
The reason I would push for a separate types package is for people who want to build something with the AST types that doesn't involve installing oxc. I can think of a couple uses for that. Also, if both `oxc-parser` and `@oxc-parser/wasm` are installed, it would be duplicated.

Boshen:
oxc-parser and @oxc-parser/wasm should publish their own types, for convenience.

ottomated:
> oxc-parser and @oxc-parser/wasm should publish their own types, for convenience.

Why would them re-exporting types from a common package be less convenient?

Boshen:
> > oxc-parser and @oxc-parser/wasm should publish their own types, for convenience.
> 
> Why would them re-exporting types from a common package be less convenient?

I meant convenient for the user, no need to add and import `@types/oxc`.

ottomated:
> I meant convenient for the user, no need to add and import `@types/oxc`.

Ah my proposal is simply that both packages declare a dependency on @types/oxc and `export *` from it. To the end user it would be entirely transparent but internally it means one source of truth

overlookmotel:
As I mentioned above, I am not that knowledgable in this area, so I can't really evaluate the arguments on either side (unfashionable opinion I know, but I don't like TS and don't use it!)

But what I *do* know is it'd be unfortunate if this PR's progress was stalled by a side-argument. So my suggestion is that initially we put the types in both `oxc-parser` and `@oxc-parser/wasm` packages, and then open a separate issue to discuss a shared `@types/oxc` package. We don't have to solve *all* the problems at the same time!

ottomated:
Sure, happy to keep the scope low. I don't think I can make a scoped PR right now though.

ottomated:
* **#6404** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ottomated:
@overlookmotel Merged in the tsify work. Pretty simple, just added a new `#[estree(custom_serialize)]` which skips generating Serialize but still generates typedefs.

overlookmotel (approved):
I've pushed a few more commits. @ottomated I hope you don't mind. It took me most of yesterday to review this PR, so I didn't want to request changes and end up going through the whole diff again to re-review. If anything you disagree with, feel free to open additional PRs to revert my changes.

Most are cosmetic. The ones which aren't are:

* fix: `type` field in TS defs on structs with `#[estree(custom_serialize)]` attr

TS type defs were missing the `type` field for `ArrayAssignmentTarget` etc.

Note: I checked all the type defs. These were the only significant errors I found (comparing to the old `.d.ts`). Will submit separate PRs to fix another couple of less important differences.

* fix: restore `features = ["derive"]` for `serde` in `oxc_ast` crate

We still have some `#[derive(Serialize)]` usage in `oxc_ast/src/serialize.rs`. Surprised Rust didn't complain. I guess it worked because `oxc_span` still had the feature enabled and `oxc_ast` depends on `oxc_span`.

* refactor: remove unused `TypescriptGenerator`
* refactor: revert `GeneratorOutput`

I removed these to reduce the diff. I'll resubmit those changes as separate PRs.

overlookmotel:
### Merge activity

* **Oct 19, 4:26 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6412, opened 2024-10-10
chore(editors/vscode): remove the usage of prettier

shulaoda:
Because we use `dprint`, there are some conflicts between them.


Boshen:
Good catch!

PR #6227, opened 2024-10-01
test(minifier): add variable collapsing test cases

DonIsaac:
Add test cases for the variable declaration collapsing AST pass

PR #6421, opened 2024-10-10
test(minifier): add all test cases for `collapse_variable_declarations`

shulaoda:
closes #6227

I noticed that the implementation of `collapse-variable-declarations` seems incomplete and appears quite simple. Therefore, I would like to quickly add all the test cases to work on this. 

Previously, I saw that `DonIsaac` had also submitted a related PR, but he might have been too busy to respond for a long time. I want to apologize for closing his PR.

### DIFF
--- a/crates/oxc_minifier/src/ast_passes/collapse_variable_declarations.rs
+++ b/crates/oxc_minifier/src/ast_passes/collapse_variable_declarations.rs
@@ -129,4 +129,145 @@ mod test {
     ",
         );
     }
+
+    #[test]
+    #[ignore]
+    fn test_collapsing() {
+        // Basic collapsing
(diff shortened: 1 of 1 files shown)


shulaoda:
BTW, I would like to understand why there are some discrepancies between our `collapse_variable_declarations` and `closure-compiler`. Is there a specific reason behind the decision to not join require calls for `cjs-module-lexer`?
https://github.com/oxc-project/oxc/blob/eacde0553e6c0b5094d8cc213e4fed6b9da01564/crates/oxc_minifier/src/ast_passes/collapse_variable_declarations.rs#L120-L131

Boshen:
> cjs-module-lexer

it doesn't pick up the exports if you collapse them.

PR #6461, opened 2024-10-11
refactor(linter): make advertised fix kinds consistent

Sysix:
- fix_dangerous is already used by `eslint/no-eq-null` and `eslint/no-unexpected-multiline`
- conditional_fix_suggestion is already used by `eslint/func-names`

You can see the doubles here: https://github.com/oxc-project/eslint-plugin-oxlint/blob/main/src/rules-by-category.ts#L477-L495


DonIsaac:
These are the same thing, so this is purely a stylistic change. Fine by me.

DonIsaac (approved):

PR #6226, opened 2024-10-01
fix(minifier): collapse variables when `join_vars` option is `true`

DonIsaac:
The variable collapse AST pass was only collapsing declarations when `CompressOptions::join_vars` was `true`, and skipping when `false`. This does not follow the behavior documented for this option. This PR fixes this.

PR #6464, opened 2024-10-12
feat(minifier): implement `collapse-variable-declarations`

shulaoda:


shulaoda:
It appears that the test262 breakage is caused by the `void x` declaration. However, surprisingly, this pass does not address the `void x` issue at all!

_Originally posted by @7086cmd in https://github.com/oxc-project/oxc/issues/6226#issuecomment-2397157338_

`CollapseVariableDeclarations` was previously disabled, and enabling it caused issues, but in fact, it has nothing to do with the problem!

7086cmd:
> `CollapseVariableDeclarations` was previously disabled, and enabling it caused issues, but in fact, it has nothing to do with the problem!

Indeed, it is strange. While I intend to conduct an investigation, I currently lack the necessary bandwidth to do so. If you are interested, you can check the reasons for committing breaking tests.

tasks/coverage/snapshots/minifier_test262.snap
```diff
@@ -2,4 +2,9 @@ commit: d62fa93c
 
 minifier_test262 Summary:
 AST Parsed     : 43914/43914 (100.00%)
-Positive Passed: 43914/43914 (100.00%)
+Positive Passed: 43909/43914 (99.99%)
+Compress: tasks/coverage/test262/test/language/directive-prologue/10.1.1-6-s.js
```
Boshen, line 6: 👀

shulaoda: This is actually unrelated to `CollapseVariableDeclarations` because I checked the relevant cases, and it didn't make any modifications to them. I need to investigate the cause further.

shulaoda:
```rust
Mismatch("Compress", "var x = 0;\nvar x = {};\n", "var x = 0, x = {};\n")
```

I believe this is due to the fact that during the first run, we executed the removal of unnecessary code first, without executing `CollapseVariableDeclarations`. @7086cmd cc

https://github.com/oxc-project/oxc/blob/702c049ebc3bb873d4c2be8eeb247f34bb551962/crates/oxc/src/compiler.rs#L186-L187

https://github.com/oxc-project/oxc/blob/702c049ebc3bb873d4c2be8eeb247f34bb551962/crates/oxc_minifier/src/compressor.rs#L40-L46

https://github.com/oxc-project/oxc/blob/702c049ebc3bb873d4c2be8eeb247f34bb551962/tasks/coverage/src/driver.rs#L126-L141

Boshen: Ok, I'll come back to this PR once I'm done with the other ones.

7086cmd: Yes, it seems. In this scenario, actually the test runner joins the var (actually we don't want them to join!).

7086cmd: But I havn't eliminate this dead code right now! 🤯

tasks/minsize/minsize.snap
```diff
@@ -1,26 +1,26 @@
...
+72.14 kB   | 24.12 kB   | 23.70 kB   | 8.62 kB    | 8.54 kB    | react.development.js
 
-173.90 kB  | 61.68 kB   | 59.82 kB   | 19.53 kB   | 19.33 kB   | moment.js 
+173.90 kB  | 61.67 kB   | 59.82 kB   | 19.53 kB   | 19.33 kB   | moment.js 
 
-287.63 kB  | 92.83 kB   | 90.07 kB   | 32.29 kB   | 31.95 kB   | jquery.js 
+287.63 kB  | 92.70 kB   | 90.07 kB   | 32.27 kB   | 31.95 kB   | jquery.js 
 
-342.15 kB  | 124.11 kB  | 118.14 kB  | 44.80 kB   | 44.37 kB   | vue.js    
+342.15 kB  | 121.95 kB  | 118.14 kB  | 44.59 kB   | 44.37 kB   | vue.js
```
DonIsaac, line 9: Interesting that minified output is larger?

shulaoda: Isn't this smaller? 🤔

Boshen:
### Merge activity

* **Oct 13, 10:34 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6467, opened 2024-10-12
perf(linter): get fewer parent nodes in `unicorn/prefer-dom-node-text-content`

camchenry:
Profiling showed that this rule was one of the slower ones, partially due to the fact that it was constantly calling `ctx.nodes()` and `ctx.nodes().parent_node()`. I've tried to fix this by reordering the logic so that we only fetch the parent nodes right before we need it. If we can early return, such as the identifier name not being `innerText`, or the node kind not being right, then we can skip fetching the parent.

Before:

Screenshot 2024-10-12 at 3 05 05 AM

After (-0.4% reduction in number of samples spent in this rule):

image

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/prefer_dom_node_text_content.rs
+++ b/crates/oxc_linter/src/rules/unicorn/prefer_dom_node_text_content.rs
@@ -44,58 +44,68 @@ declare_oxc_lint!(
 
 impl Rule for PreferDomNodeTextContent {
     fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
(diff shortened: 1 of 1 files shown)


camchenry:
* **#6467** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
### Merge activity

* **Oct 13, 10:16 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6540, opened 2024-10-13
chore(benchmark): Add small JSX file to linter benchmarks

camchenry:
We don't currently have any good representation for small files in the linter benchmarks. We already have benchmarks for lexer, parser, semantic, etc. for `RadixUIAdoptionSection.tsx`, so I think it makes sense to enable it for `linter` as well.

This will help us measure any fixed costs that we pay for some lint rules, such as initializing data structures, etc.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -152,6 +152,7 @@ jobs:
         fixture:
           - 0
           - 1
+          - 2
 
     steps:
       - name: Checkout Branch


camchenry:
* **#6540** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6540?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Boshen:
Queueing for jobs is our current CI bottleneck, I'll try and balance this.

DonIsaac:
I think graphite offers some optimizations we don't currently use. May be worth looking in to

Boshen:
> I think graphite offers some optimizations we don't currently use. May be worth looking in to

Been there done that 😅 They create total chaos.

camchenry:
@Boshen Do you have any ideas on how you wanted to address that? I haven't personally noticed CI queueing to be too much of a bottleneck. Some potential options we could explore:

- Cancel jobs when new commits are pushed
- Skip some jobs when PR is a draft
- Reducing time it takes to run each job (using smaller images, downloading less, etc.)

overlookmotel:
> * Cancel jobs when new commits are pushed

This one already happens, I believe.

I *think* adding this small benchmark should not be too bad, as the linter benchmark is built in a separate job, so it's only *running* the linter which is extra here. The extra job took 1m05s on this PR.

camchenry:
@Boshen would it be more worth it to add more small files 

> > * Cancel jobs when new commits are pushed
> 
> This one already happens, I believe.
> 
> I _think_ adding this small benchmark should not be too bad, as the linter benchmark is built in a separate job, so it's only _running_ the linter which is extra here. The extra job took 1m05s on this PR.

Yeah, it does look like it happens on most (all?) jobs actually. Not sure how to improve job bottleneck situation here, apart from making the linter or compiling like 10x faster 😅

overlookmotel:
https://github.com/oxc-project/backlog/issues/3 would help a lot. I got it *almost* working, but unlikely to get time to return to it and finish it off until we've finished transformer milestone 2.

PR #6551, opened 2024-10-14
fix(minifier): Preserve init variable declarations when removing `for` statements during DCE

### LINKED ISSUE #6547: Wrong full-elimination on `for` stmt with `var` decl.
Given the code:

```js
for (var { c, x: [d] } = {}; 0;);
```

In reality, our actual behavior is to eliminate the entire code, which is incorrect. We need to eliminate it to:

```js
var { c: a, x: [c] } = {};
```

Also given:

```js
for (var se = [1, 2]; false;);
```

need we handle it to:

```js
var se=[1,2];
```

magic-akari:
- Closes: #6547

### DIFF
--- a/crates/oxc_ast/src/ast_builder_impl.rs
+++ b/crates/oxc_ast/src/ast_builder_impl.rs
@@ -113,6 +113,17 @@ impl<'a> AstBuilder<'a> {
     }
 
     #[inline]
+    pub fn move_variable_declaration(self, decl: &mut VariableDeclaration<'a>) -> VariableDeclaration<'a> {
+        let empty_decl = self.variable_declaration(
+            Span::default(),
+            VariableDeclarationKind::Var,
+            self.vec(),
+            false,
+        );
+        mem::replace(decl, empty_decl)
+    }
+
+    #[inline]
     pub fn move_vec<T>(self, vec: &mut Vec<'a, T>) -> Vec<'a, T> {
         mem::replace(vec, self.vec())
     }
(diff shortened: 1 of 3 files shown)


crates/oxc_ast/src/ast_builder_impl.rs
```diff
@@ -112,6 +112,20 @@ impl<'a> AstBuilder<'a> {
...
+        decl: &mut VariableDeclaration<'a>,
+    ) -> VariableDeclaration<'a> {
+        let empty_decl = self.variable_declaration(
+            Span::default(),
+            VariableDeclarationKind::Var,
+            self.vec(),
+            false,
+        );
+        mem::replace(decl, empty_decl)
+    }
```
7086cmd, line 127: Will the `var` raise some effects on `KeepVar`?

magic-akari: Not sure what you mean, but you can review the added test cases.

magic-akari: Do you mean this case?

```JavaScript
for(let i = 0; i < 10; i++) {
    var a = 0;
}
```

In this case, we should not count the variable `a` because it is in block scope.

camc314: Screenshot 2024-10-14 at 13 12 59

why do we drop `a` ?

7086cmd: Well, I got misunderstood. You can ignore this review :joy:

7086cmd (approved):
LGTM

PR #6573, opened 2024-10-15
feat(transformer): add `async_generator_functions` plugin

Dunqing:
Passed 15/19 tests. The remaining 4 failed tests related to `this` expression, the problem same as I mentioned in #6658. I will fix them in follow-up PRs.

### DIFF
--- a/crates/oxc_transformer/src/es2018/async_generator_functions/for_await.rs
+++ b/crates/oxc_transformer/src/es2018/async_generator_functions/for_await.rs
@@ -1,0 +1,426 @@
+//! This module is responsible for transforming `for await` to `for` statement
+
+use oxc_allocator::Vec;
+use oxc_ast::{ast::*, NONE};
+use oxc_semantic::{ScopeFlags, ScopeId, SymbolFlags};
+use oxc_span::SPAN;
+use oxc_traverse::{BoundIdentifier, TraverseCtx};
+
+use super::AsyncGeneratorFunctions;
+use crate::{common::helper_loader::Helper, es2017::AsyncGeneratorExecutor};
+
+impl<'a, 'ctx> AsyncGeneratorFunctions<'a, 'ctx> {
+    pub(crate) fn transform_for_of_statement(
+        &mut self,
+        stmt: &mut ForOfStatement<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) -> Vec<'a, Statement<'a>> {
+        let step_key =
+            ctx.generate_uid("step", ctx.current_scope_id(), SymbolFlags::FunctionScopedVariable);
+        // step.value
+        let step_value = ctx.ast.expression_member(ctx.ast.member_expression_static(
+            SPAN,
+            step_key.create_read_expression(ctx),
+            ctx.ast.identifier_name(SPAN, "value"),
+            false,
+        ));
+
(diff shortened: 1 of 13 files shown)


Dunqing:
* **#6573** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_transformer/src/es2018/async_generator_functions/for_await.rs
```diff
@@ -0,0 +1,340 @@
...
+            body,
+            ctx.current_scope_id(),
+            ctx,
+        )
+    }
+}
+
+/// Constructs AST based on AstBuilder
+/// <https://github.com/babel/babel/blob/d20b314c14533ab86351ecf6ca6b7296b66a57b3/packages/babel-plugin-transform-async-generator-functions/src/for-await.ts#L3-L30>
+fn build_for_await<'a>(
```
overlookmotel, line 79: Holy CHRIST! This function is *enormous*. We really do need the `ast!` macro.

Probably you were going to do this anyway, but could you include what code it produces in the comment (written out, rather than just link to Babel's impl)?

overlookmotel: Also, I'm not sure the 3 x `VariableDeclaration`s are actually needed. Doesn't Babel just discard them?

I had assumed they were only in Babel's template to make the `template` function resolve the identifiers correctly (I could be wrong).

Dunqing: I think so, but anyway, we need to follow Babel's implementation 😢

overlookmotel (approved):
I've not reviewed the whole transform, but nice one getting the semantic tests to pass!

overlookmotel:
### Merge activity

* **Oct 30, 8:03 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

crates/oxc_transformer/src/es2018/async_generator_functions/for_await.rs
overlookmotel: We need to follow Babel's *output*, but I don't think we need follow its *implementation* necessarily.

Anyway, I don't fully understand this transform, so I may be wrong in what I said above anyway.

overlookmotel:
@Dunqing Looks like CI has got stuck. I see it do this sometimes, don't know why. Solution I've found is:

* Make a random change (add a dummy comment, whatever) and `gt m -a`.
* Revert that change and `gt m -a` again.
* `gt submit`

Or if branch has got behind main, then just `gt sync` + `gt submit` also works.

Dunqing:
> @Dunqing Looks like CI has got stuck. I see it do this sometimes, don't know why. Solution I've found is:
> 
> * Make a random change (add a dummy comment, whatever) and `gt m -a`.
> * Revert that change and `gt m -a` again.
> * `gt submit`
> 
> Or if branch has got behind main, then just `gt sync` + `gt submit` also works.

Thanks! That works, just an empty commit so I think it's fine if we don't revert it

PR #4742, opened 2024-08-07
feat(oxlint): implement `--print-config` cli option

nicksrandall:
I've attempted to implement [`--print-config`](https://eslint.org/docs/latest/use/command-line-interface#--print-config) for oxlint. This implementation works but I'm almost certain there is a better way to do this and I'm open to feedback.

See discussion: #4279

PR #6579, opened 2024-10-15
feat(linter): support `--print-config` to print config file for project

mysteryven:
Continue work on #4742.

### DIFF
--- a/apps/oxlint/src/command/mod.rs
+++ b/apps/oxlint/src/command/mod.rs
@@ -25,6 +25,11 @@ pub struct MiscOptions {
     /// Number of threads to use. Set to 1 for using only 1 CPU core
     #[bpaf(argument("INT"), hide_usage)]
     pub threads: Option<usize>,
+
+    /// This option outputs the configuration to be used.
+    /// When present, no linting is performed and only config-related options are valid.
+    #[bpaf(switch, hide_usage)]
+    pub print_config: bool,
 }
 
 #[allow(clippy::ptr_arg)]
--- a/apps/oxlint/fixtures/print_config/ban_rules/eslintrc.json
+++ b/apps/oxlint/fixtures/print_config/ban_rules/eslintrc.json
@@ -1,0 +1,9 @@
+{
+  "rules": {
+    "no-debugger": "warn",
+    "eqeqeq": [
+      "warn",
+      "always"
+    ]
+  }
+}
--- a/apps/oxlint/fixtures/print_config/ban_rules/expect.json
+++ b/apps/oxlint/fixtures/print_config/ban_rules/expect.json
@@ -1,0 +1,44 @@
+{
+  "plugins": [
+    "react",
+    "unicorn",
+    "typescript",
+    "oxc"
+  ],
+  "categories": {},
+  "rules": {
+    "eqeqeq": [
+      "deny",
+      [
+        "always"
+      ]
+    ]
+  },
+  "settings": {
+    "jsx-a11y": {
+      "polymorphicPropName": null,
+      "components": {}
+    },
+    "next": {
+      "rootDir": []
+    },
+    "react": {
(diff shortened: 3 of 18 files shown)


mysteryven:
* **#6579** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pumano:
very useful feature for debugging!

apps/oxlint/src/lint/mod.rs
```diff
@@ -116,6 +116,14 @@ impl Runner for LintRunner {
         };
 
         enable_plugins.apply_overrides(&mut oxlintrc.plugins);
+
+        // TODO: add support for printing the configuration for a specific path when overrides supported.
+        let oxlintrc_for_print = if let Some(PrintConfig::All) = &misc_options.print_config {
```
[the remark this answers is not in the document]
mysteryven: > even advertise it as a CLI option

I wanted `--print-config`/`--print-config all` to represent 'all', but I didn't find a proper way in bpaf, to avoid breaking change, maybe warn for now?

apps/oxlint/src/command/mod.rs
```diff
@@ -25,6 +25,31 @@ pub struct MiscOptions {
...
+
+#[derive(Debug, Clone, Default)]
+pub enum PrintConfig {
+    /// Print the configuration for all files
+    #[default]
+    All,
+    Path(PathBuf),
+}
+
+impl std::str::FromStr for PrintConfig {
```
pacak, line 43: `PathBuf` can handle non-utf8 inputs since they are valid in Windows/Linux, `FromStr` can handle only valid utf8 so doing this you break a bit of compatibility. In practice this is not a big problem, but I wouldn't do it unless there's a reason.

I'm assuming you want to treat `--print-config` as "let's check all the configs" and `--print-config foo` as "let's check the config in `foo`.

You want two parsers for each case, both parsers should produce `PrintConfig` type. IT is perfectly valid to have multiple parsers with the same name, here - `--print-config`, you want to try to run that parses a file name first. Easiest way would be this is something like this:

```
#[derive(Debug, Clone, Default, Bpaf)]
pub enum PrintConfig {
    Path(
        /// help if you want
        #[bpaf(long("print-config"))] PathBuf),

    /// Print the configuration for all files
    #[default]
    #[bpaf(long("print-config"))]
    All,
    
}
```

DonIsaac, line 43: Why not `From<&str>` if this always returns `Ok`?

apps/oxlint/src/command/mod.rs
```diff
@@ -25,6 +25,31 @@ pub struct MiscOptions {
...
+    type Err = String;
```
DonIsaac, line 44: If you do decide to stick with `FromStr`, consider using `Infallible`

pacak: `bpaf` constraints arbitrary arguments to be a `FromStr`. One way would be to consume `String` or `OsString` and convert it into `PrintConfig` with `map`, second way is to parse into `enum` directly as I described.

apps/oxlint/src/command/mod.rs
mysteryven: Sorry, it's auto generated by editor, I didn’t notice this.

DonIsaac (approved):

DonIsaac:
LGMT

DonIsaac:
### Merge activity

* **Oct 22, 1:55 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6590, opened 2024-10-15
perf(linter/no-unused-vars): do not construct `Regex` for default ignore pattern

camchenry:
Profiling shows that constructing the `Regex` for `no-unused-vars` takes up a good chunk of time for small files. Since the default pattern for ignored variables is just `^_`, we can more simply implement this with `.starts_with('_')`.

image

I decided to create a new `Option`-like enum that has three states: `Some` and `None` which are like the typical `Option` states, and another `Default` state which is a special case. The `Default` state represents that matching code should use the default pattern (if applicable). I decided not to keep using `Option`, because I didn't want to overload the `None` state by confusing whether or not it was the default state, or the user explicitly overrided it as `None`.

The theory here is that most users are not going to pass a custom ignore pattern, so we can avoid needing to create a `Regex` in most cases, which greatly speeds up the initialization for this rule.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_unused_vars/options.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_unused_vars/options.rs
@@ -34,7 +34,7 @@ pub struct NoUnusedVarsOptions {
     /// var b = 10;
     /// console.log(b);
     /// ```
-    pub vars_ignore_pattern: Option<Regex>,
+    pub vars_ignore_pattern: IgnorePattern<Regex>,
 
     /// Controls how unused arguments are checked.
     ///
@@ -65,7 +65,7 @@ pub struct NoUnusedVarsOptions {
     /// }
     /// foo(1, 2);
     /// ```
-    pub args_ignore_pattern: Option<Regex>,
+    pub args_ignore_pattern: IgnorePattern<Regex>,
 
     /// Using a Rest property it is possible to "omit" properties from an
     /// object, but by default the sibling properties are marked as "unused".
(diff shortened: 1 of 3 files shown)


camchenry:
* **#6590** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camchenry:
codspeed shows +0% but it's closer to ~1% locally:

```
hyperfine --warmup 100 --shell=none --runs 9000 './oxlint-main --silent ./RadixUIAdoptionSection.jsx' './oxlint-unused-regex --silent ./RadixUIAdoptionSection.jsx'
Benchmark 1: ./oxlint-main --silent ./RadixUIAdoptionSection.jsx
  Time (mean ± σ):       3.4 ms ±   0.1 ms    [User: 1.1 ms, System: 1.3 ms]
  Range (min … max):     3.2 ms …   3.9 ms    9000 runs

Benchmark 2: ./oxlint-unused-regex --silent ./RadixUIAdoptionSection.jsx
  Time (mean ± σ):       3.4 ms ±   0.1 ms    [User: 1.1 ms, System: 1.3 ms]
  Range (min … max):     3.0 ms …   4.2 ms    9000 runs

Summary
  ./oxlint-unused-regex --silent ./RadixUIAdoptionSection.jsx ran
    1.01 ± 0.05 times faster than ./oxlint-main --silent ./RadixUIAdoptionSection.jsx
```

crates/oxc_linter/src/rules/eslint/no_unused_vars/options.rs
```diff
@@ -207,18 +207,92 @@ pub struct NoUnusedVarsOptions {
     pub report_used_ignore_pattern: bool,
 }
 
+/// Represents an `Option<Regex>` with an additional `Default` variant,
+/// which represents the default ignore pattern for when no pattern is
+/// explicitly provided.
+#[derive(Debug, Clone, Copy)]
+pub enum IgnorePattern<R> {
```
DonIsaac, line 214: Do we need the type parameter?

DonIsaac (approved):

PR #6594, opened 2024-10-15
perf(linter/no-unescaped-entities): add fast path to check if char should be replaced

camchenry:
Flamegraph shows a lot of time spent hashing here, so I added a faster check to see if it is a char that we need to replace.

Issue #6038, opened 2024-09-24
linter: duplicate work in Jest/Vitest rules


DonIsaac:
This is a big one. I haven't taken the time to write this down until now.

## Problem

Most (if not all) `eslint-plugin-jest` rules use `collect_possible_jest_call_node` in a `run_once` block, then pass those nodes to an internal "run" implementation. For example,

```rust
// crates/oxc_linter/src/rules/jest/no_hooks.rs
impl Rule for NoHooks {
    fn from_configuration(value: serde_json::Value) -> Self {
        let allow = value
            .get(0)
            .and_then(|config| config.get("allow"))
            .and_then(serde_json::Value::as_array)
            .map(|v| {
                v.iter().filter_map(serde_json::Value::as_str).map(ToString::to_string).collect()
            })
            .unwrap_or_default();

        Self(Box::new(NoHooksConfig { allow }))
    }

    fn run_once(&self, ctx: &LintContext) {
        for possible_jest_node in collect_possible_jest_call_node(ctx) {
            self.run(&possible_jest_node, ctx);
        }
    }
}
```

This is problematic. Each jest rule repeats the same work, which wastes cycles. To make things worse, `collect_possible_jest_call_node` allocates nodes it discovers onto the heap.

## (Possible) Solution
1. Update the `Rule` trait, adding a `run_on_jest_node` method.
2. Update `Linter::run` to use `collect_possible_jest_call_node` and pass a reference to them to `rule.run_on_jest_node`
3. Remove all `run_once` impls from jest/vitest rules.

## Acceptance Criteria
1. `Linter` should skip jest node collection if neither the jest or vitest plugins are enabled
2. Same as above, but non-test files should also be skipped, regardless if jest/vitest plugins are enabled or not

## Extra Credit
Bonus points if you can refactor `collect_possible_jest_node` to return an `impl Iterator` instead of a `Vec`.

CC: @camchenry, who has actively been making performance improvements to the linter.

camchenry:
I attempted to do this, but I had an absolute heck of a time getting the lifetimes to work out properly. I'm not experienced enough with Rust yet to know exactly where the mistake is, so help appreciated: https://github.com/oxc-project/oxc/pull/6049

camchenry:
Closing this as done with https://github.com/oxc-project/oxc/pull/6722, we even got the extra credit 😄

PR #6600, opened 2024-10-15
perf(linter): apply small file optimization, up to 30% faster

camchenry:
Theory: iterating over the rules three times has slightly worse cache locality, because the prior iterations have pushed `rule` out of the cache by the time we iterate over it again. By iterating over each rule only once, we improve cache performance (hopefully). We also don't need to collect rules to a Vec, so it saves some CPU/memory there too.

In practice: the behavior here actually depends on the number of AST nodes that are in the program. If the number of nodes is large, then it's better to iterate over the nodes only once and iterate the rules multiple times. But if the number of nodes is small, then it's better to iterate over nodes multiple times and only iterate over the rules once. See this comment for more context: https://github.com/oxc-project/oxc/pull/6600#issuecomment-2427837715, as well as the comment inside the PR: https://github.com/oxc-project/oxc/pull/6600/files#diff-207225884c5e031ffd802bb99e4fbacbd8364b1343a1cec5485bf50f29186300R131-R143.

In practice, this can make linting a file 1-45% faster, depending on the size of the file, number of AST nodes, number of files, CPU cache size, etc. To accommodate large and small files better, we have an explicit threshold of 200,000 AST nodes, which is an arbitrary number picked based on some benchmarks on my laptop. For large files, the linter behavior doesn't change. For small files, we switch to iterating over nodes in the inner loop and iterating over rules once in the outer loop.

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -121,30 +121,74 @@ impl Linter {
             .rules
             .iter()
             .filter(|rule| rule.should_run(&ctx_host))
-            .map(|rule| (rule, Rc::clone(&ctx_host).spawn(rule)))
-            .collect::<Vec<_>>();
-
-        for (rule, ctx) in &rules {
-            rule.run_once(ctx);
-        }
+            .map(|rule| (rule, Rc::clone(&ctx_host).spawn(rule)));
 
         let semantic = ctx_host.semantic();
-        for symbol in semantic.symbols().symbol_ids() {
-            for (rule, ctx) in &rules {
-                rule.run_on_symbol(symbol, ctx);
-            }
-        }
 
-        for node in semantic.nodes() {
-            for (rule, ctx) in &rules {
-                rule.run(node, ctx);
-            }
-        }
+        let should_run_on_jest_node =
+            self.options.plugins.has_test() && ctx_host.frameworks().is_test();
 
-        if ctx_host.frameworks().is_test() && self.options.plugins.has_test() {
-            for jest_node in iter_possible_jest_call_node(semantic) {
+        // IMPORTANT: We have two branches here for performance reasons:
+        //
+        // 1) Branch where we iterate over each node, then each rule
+        // 2) Branch where we iterate over each rule, then each node
+        //
+        // When the number of nodes is relatively small, most of them can fit
+        // in the cache and we can save iterating over the rules multiple times.
+        // But for large files, the number of nodes can be so large that it
+        // starts to not fit into the cache and pushes out other data, like the rules.
+        // So we end up thrashing the cache with each rule iteration. In this case,
+        // it's better to put rules in the inner loop, as the rules data is smaller
+        // and is more likely to fit in the cache.
+        //
+        // The threshold here is chosen to balance between performance improvement
+        // from not iterating over rules multiple times, but also ensuring that we
+        // don't thrash the cache too much. Feel free to tweak based on benchmarking.
+        //
+        // See https://github.com/oxc-project/oxc/pull/6600 for more context.
+        if semantic.stats().nodes > 200_000 {
+            // Collect rules into a Vec so that we can iterate over the rules multiple times
+            let rules = rules.collect::<Vec<_>>();
+
+            for (rule, ctx) in &rules {
+                rule.run_once(ctx);
+            }
+
+            for symbol in semantic.symbols().symbol_ids() {
                 for (rule, ctx) in &rules {
-                    rule.run_on_jest_node(&jest_node, ctx);
+                    rule.run_on_symbol(symbol, ctx);
+                }
+            }
+
+            for node in semantic.nodes() {
+                for (rule, ctx) in &rules {
+                    rule.run(node, ctx);
+                }
+            }
+
+            if should_run_on_jest_node {
+                for jest_node in iter_possible_jest_call_node(semantic) {
+                    for (rule, ctx) in &rules {
+                        rule.run_on_jest_node(&jest_node, ctx);
+                    }
+                }
+            }
+        } else {
+            for (rule, ref ctx) in rules {
+                rule.run_once(ctx);
+
+                for symbol in semantic.symbols().symbol_ids() {
+                    rule.run_on_symbol(symbol, ctx);
+                }
+
+                for node in semantic.nodes() {
+                    rule.run(node, ctx);
+                }
+
+                if should_run_on_jest_node {
+                    for jest_node in iter_possible_jest_call_node(semantic) {
+                        rule.run_on_jest_node(&jest_node, ctx);
+                    }
                 }
             }
         }


camchenry:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6600?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6600** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6600?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6590** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6590?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#6594](https://github.com/oxc-project/oxc/pull/6594) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6594?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

overlookmotel:
Benchmark results are wild! I'll take a look at this tomorrow.

overlookmotel:
This is rather mysterious! Personally I think unlikely this effect is due to cache locality. It seems unlikely that'd have such a large effect. More likely it's just doing a lot less work turning the loops this way around.

The big question is: what's the big difference between `cal.com.tsx` and `checker.ts`?

I suggest running before/after with a proper flamegraph (better than what CodSpeed gives) and see what that reveals.

Also, you might want to remove the intermediate `Vec` of rules and see if that makes any difference. Just loop over the iterator instead of `.collect::<Vec<_>>()` first.

camchenry:
> Also, you might want to remove the intermediate `Vec` of rules and see if that makes any difference. Just loop over the iterator instead of `.collect::<Vec<_>>()` first.

I did try this first, I had just re-added it to see if it made any difference. It didn't improve the regression, so I'm going to leave it out.



> This is rather mysterious! Personally I think unlikely this effect is due to cache locality. I could be wrong, but it seems unlikely to me that'd have such a large effect. More likely it's just doing a lot less work turning the loops this way around.
> 
> The big question is: what's the big difference between `cal.com.tsx` and `checker.ts`?
> 
> I suggest running before/after with a proper flamegraph (better than what CodSpeed gives) and see what that reveals.

Yeah, I'm going to work on profiling this to see what's up.

camchenry:
Samply didn't help too much in profiling, I'm still a bit confused. The only thing that sticks out is this branch doesn't have a big call to `bumpalo::dealloc` for some reason. But I feel like the debug symbols are not quite attributed correctly and I wonder if that contributes to some of the weird behavior.

`main`:
Screenshot 2024-10-15 at 11 19 27 PM

This branch:
Screenshot 2024-10-15 at 11 18 25 PM

The call tree shows some strange differences like `oxc_parser::js::module::::parse_export_named_declaration` dropping out from the top calls, but that seems unrelated?

`main`:
Screenshot 2024-10-15 at 11 28 13 PM

This branch:
Screenshot 2024-10-15 at 11 28 04 PM

Boshen:
This is a benchmark fallacy where all rules are turn on.

For the two loops, that's 420 rules * total number of AST nodes, so performance will obviously be different for different file sizes.

For real usages, it's around 100 rules so performance will be different as well.

Boshen:
What about applying a strategy where we switch the loop when the total number of ast nodes is smaller than a threshold?

camchenry:
I'm not in a rush to merge this, so I'm going to continue investigating as best as I can. However, the performance gains here are definitely real and very statistically significant. With default settings used for `oxlint`, I see ~5-10% performance improvement on my M1 laptop. With all rules/plugins enabled, the performance improvement goes up to ~22%.

```
hyperfine -i -N --warmup 10 --runs 100 './oxlint-rules-iter --silent ./checker.ts' './oxlint-main --silent ./checker.ts'Benchmark 1: ./oxlint-rules-iter --silent ./checker.ts
  Time (mean ± σ):     172.7 ms ±   0.8 ms    [User: 162.4 ms, System: 9.2 ms]
  Range (min … max):   171.0 ms … 174.5 ms    100 runs

Benchmark 2: ./oxlint-main --silent ./checker.ts
  Time (mean ± σ):     192.0 ms ±   0.6 ms    [User: 181.8 ms, System: 9.1 ms]
  Range (min … max):   188.7 ms … 193.5 ms    100 runs

Summary
  ./oxlint-rules-iter --silent ./checker.ts ran
    1.11 ± 0.01 times faster than ./oxlint-main --silent ./checker.ts
```

```
hyperfine -i -N --warmup 10 --runs 100 './oxlint-rules-iter --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts' './oxlint-main --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts'
Benchmark 1: ./oxlint-rules-iter --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts
  Time (mean ± σ):     612.6 ms ±   2.0 ms    [User: 596.0 ms, System: 15.3 ms]
  Range (min … max):   608.1 ms … 617.4 ms    100 runs

Benchmark 2: ./oxlint-main --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts
  Time (mean ± σ):     753.6 ms ±   6.7 ms    [User: 735.7 ms, System: 16.3 ms]
  Range (min … max):   733.5 ms … 795.0 ms    100 runs

Summary
  ./oxlint-rules-iter --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts ran
    1.23 ± 0.01 times faster than ./oxlint-main --deny=all --import-plugin --jsdoc-plugin --jest-plugin --vitest-plugin --jsx-a11y-plugin --nextjs-plugin --react-perf-plugin --promise-plugin --node-plugin --threads 1 --silent ./checker.ts
```

I did some debugging in Xcode and did find that this new version actually has _more_ L1 data cache misses than the code on `main` which is interesting, but doesn't explain the performance difference quite yet. The main thing that seems to explain the performance is that this branch simply executes fewer cycles.

camchenry:
So I noticed an interesting pattern here between `main` and this PR. In `main`, there are a bunch of L1 data cache misses at the beginning, which then go away which I assume means that we are using the cache a lot more for the remainder of the process. There are also ~2M L1 cache misses.

image

However, for this PR, the L1 data cache misses are evenly distributed over the duration of the process, and there are ~20x more cache misses with ~45M L1 cache misses:

image

For some reason, ~35M of these cache misses are being attributed to the `jest/valid-expect` rule:
image

And specifically, it looks like it's a lot of time being spent getting these possible jest nodes: 
image

So I'm wondering if by changing the data access pattern here we've actually made the caching _worse_ because now the Jest nodes are constantly getting pushed out of the cache and then put back into the cache. It might be worth working on https://github.com/oxc-project/oxc/issues/6038 and then revisiting this to see if it has improved. It's possible that the performance regression in CodSpeed could maybe be explained by different caching behavior versus my M1 laptop?

crates/oxc_linter/src/lib.rs
```diff
@@ -121,30 +121,70 @@ impl Linter {
...
             for (rule, ctx) in &rules {
-                rule.run_on_symbol(symbol, ctx);
+                rule.run_once(ctx);
             }
-        }
 
-        for node in semantic.nodes() {
-            for (rule, ctx) in &rules {
-                rule.run(node, ctx);
+            let semantic = ctx_host.semantic();
```
overlookmotel, line 152: Nit: Do you need to get `semantic` twice?

camchenry:
I think I finally figured this one out. Re-profiling with the latest code changes shows a much healthier usage of L1 cache, where we have a bunch of misses initially but it falls off to a smaller amount.


image


I think the reason why we were encountering issues in the `checker.ts` benchmark file is because `checker.ts` is the largest of the three benchmarked files by far. That means `checker.ts` also has the most `AstNode` to iterate over in `semantic.nodes().iter()`. When the number of nodes is relatively small (<200K), the nodes seem more able to fit in the cache with each iteration over the rules. However, when the number of nodes is large, it is better to have rules in the inner loop so that the nodes stay in the cache. Otherwise, the number of nodes completely negates the benefit of only having to iterate over the rules once.

So I've taken @Boshen's [suggestion from earlier](https://github.com/oxc-project/oxc/pull/6600#issuecomment-2416959423) to have a threshold for when we apply this strategy. This makes this essentially a "small file" optimization where smaller files are linted in a slightly different order but large files are unaffected. This does complicate the linter running code, but I believe it's worth the benefit. Benchmarking shows this can speed up the linter by 10-40%, depending on the number of rules and plugins, but generally the higher the number of rules the better this performs. I think the extra complication here is well worth the slightly duplicated implementations.

crates/oxc_linter/src/lib.rs
camchenry: Good catch, that was unintentional.

overlookmotel:
Great that the perf drop on `checker.ts` has gone. Spectacular!

Out of interest, how many nodes do the various benchmarks have? Clearly `checker.ts` has more than 200,000. What about the others?

But... (and I'm sorry to be the guy to piss on your parade when you've achieved such a stellar result)... are you sure you're not over-fitting the data? i.e. optimizing for our specific benchmarks, rather than the general case?

And did you ever get to the bottom of what's going on with the expensive `drop` calls in `parse_export_named_declaration`?

Just to be clear... despite me asking questions, what I'm *not* questioning is that this is excellent work. Bravo!

overlookmotel:
> I think the extra complication here is well worth the slightly duplicated implementations.

Totally agree. It's not a huge complication, and you've made a comment which clearly explains the rationale.

camchenry:
> Out of interest, how many nodes do the various benchmarks have? Clearly `checker.ts` has more than 200,000. What about the others?

Good question, the AST node counts are as follows:

- `checker.ts`: 303,463 nodes
- `RadixUIAdoptionSection.jsx`: 392 nodes
- `cal.com.tsx`: 153,029 nodes

And for some other targets:

- `target/antd.js`: 571,672 nodes
- `target/pdf.mjs`: 108,379 nodes


> But... (and I'm sorry to be the guy to piss on your parade when you've achieved such a stellar result)... are you sure you're not over-fitting the data? i.e. optimizing for our specific benchmarks, rather than the general case?

Not at all, and thank you for your help on this as well! I had written out a much longer response but I accidentally closed the tab and deleted it 😅 It's definitely possible that I've overfitted this to these particular benchmarks and to my personal machine. I've tried to mitigate this somewhat by running this against not just single files but larger codebases that have files of many different shapes and sizes. But I haven't tried this on other computers. **We should continue to tweak this threshold.**

Even still, I am able to replicate the CodSpeed results across whole codebases not just single files. For example:

Linting `supabase` (3,748 files) is only ~3% faster:

Screenshot 2024-10-21 at 6 37 45 PM


Linting `gitlab` (12,598 files) is ~16% faster:

Screenshot 2024-10-21 at 6 43 02 PM


Linting `vscode` (5,107 files) is ~33% faster:

Screenshot 2024-10-21 at 5 51 43 PM

camchenry:
> And did you ever get to the bottom of what's going on with the expensive `drop` calls in `parse_export_named_declaration`?

I might have misspoken earlier, when I said "dropped out" I think I just meant it didn't appear in the profiler call tree. But I think there was some strange debug symbol attribution going on there, I didn't see this happening again when profiling in Xcode Instruments. I didn't see anything that pointed to this method being a big issue currently.

overlookmotel:
Thanks very much for giving the real-world results on various codebases. That's pretty convincing.

Perhaps we could finesse the 200,000 "tipping point", but it seems to me it's clearly good enough. In the real world who (apart from the maniac authors of TS) writes code in files anywhere near as large as that? We *might* encounter files that big in transformer - large libraries pre-bundled into a single file in `node_modules`. But that's not relevant for the linter - you don't lint your `node_modules` folder!

> I might have misspoken earlier, when I said "dropped out" I think I just meant it didn't appear in the profiler call tree. But I think there was some strange debug symbol attribution going on there, I didn't see this happening again when profiling in Xcode Instruments. I didn't see anything that pointed to this method being a big issue currently.

Can I ask you a favour? Would you mind just checking if benchmark result on `checker.ts` is still the same as it was before if you disable the "large file" branch, so `checker.ts` goes through same branch as the rest? I think very likely your cache lines explanation of what's going on is correct, but I would just like to double-check that the `drop` calls were not significant. It's *possible* that it was significant, but was solved by #6623, and that would be another explanation of why it's now disappeared from the flamegraph after that PR was merged.

Beyond that, LGTM!

crates/oxc_linter/src/lib.rs
```diff
@@ -121,30 +121,69 @@ impl Linter {
...
+        //
+        // This is done because when the number of nodes is relatively small,
+        // we can fit most of them in the cache and save iterating over the rules
+        // Vec multiple times. But for large files, we get penalized because there
+        // is too much data to fit into the cache and actually end up using the
+        // cache less often.
+        //
+        // The threshold here is chosen to balance between performance improvement
+        // from not iterating over rules multiple times, but also ensuring that we
+        // don't thrash the cache too much. Feel free to tweak based on benchmarking.
```
overlookmotel, line 143: I think it'd be useful to include a link to this PR at the end of this comment. The instrumentation screen-grabs and your comments here are very useful background for anyone who revisits this.

camchenry:
> Can I ask you a favour? Would you mind just checking if benchmark result on `checker.ts` is still the same as it was before if you disable the "large file" branch, so `checker.ts` goes through same branch as the rest? I think very likely your cache lines explanation of what's going on is correct, but I would just like to double-check that the `drop` calls were not significant. It's _possible_ that it was significant, but was solved by #6623, and that would be another explanation of why it's now disappeared from the flamegraph after that PR was merged.

@overlookmotel I think I did benchmark this on `main` after https://github.com/oxc-project/oxc/pull/6623 was merged (and rebased this PR on top of it). Just want to make sure I understand what you're asking. Should I comment out the large branch case and benchmark it, using the latest changes from main? Or did you want me to benchmark against an older commit?

crates/oxc_linter/src/lib.rs
camchenry: Good idea 👍

crates/oxc_linter/src/lib.rs
```diff
@@ -121,30 +121,74 @@ impl Linter {
...
+        // So we end up thrashing the cache with each rule iteration. In this case,
+        // it's better to put rules in the inner loop, as the rules data is smaller
+        // and is more likely to fit in the cache.
+        //
+        // The threshold here is chosen to balance between performance improvement
+        // from not iterating over rules multiple times, but also ensuring that we
+        // don't thrash the cache too much. Feel free to tweak based on benchmarking.
+        //
+        // See https://github.com/oxc-project/oxc/pull/6600 for more context.
+        if semantic.stats().nodes > 200_000 {
```
Boshen, line 149: I may try and move these two branches to two functions, to squeeze in a bit of CPU instructions cache 😅

Boshen:
The codspeed benchmark has all the benchmark turned on. Is your hyperfine benchmark running all rules or just the default? I would optimize for the default case, where most people run ~100 rules, not 430.

camchenry:
> The codspeed benchmark has all the benchmark turned on. Is your hyperfine benchmark running all rules or just the default? I would optimize for the default case, where most people run ~100 rules, not 430.

[This benchmarking](https://github.com/oxc-project/oxc/pull/6600#issuecomment-2427882223) was done with the default rules and plugins, so it was 98 rules in total.

Boshen:
I'm super happy with the results, eager to merge!

crates/oxc_linter/src/lib.rs
camchenry: @Boshen shall we do this in a follow-up PR? I don't want to conflate the results in this PR too much with any potential inlining changes we make here. It might be easier to do that separately.

Boshen: Sure!

Boshen:
### Merge activity

* **Oct 21, 10:23 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6596, opened 2024-10-15
refactor(transformer): `HelperLoader` common transform: construct string directly in arena

overlookmotel:
Small optimization. Avoid allocating a temporary `String` which then gets copied into the arena, by constructing the string directly in the arena.

We should probably have a dedicated API for doing this, but for now just do it here.

PR #6595, opened 2024-10-15
refactor(transformer): `HelperLoader` common transform: `Helper` enum

overlookmotel:
`HelperLoaderStore::load` take a `Helper` instead of a string.

`Helper` has 2 purposes:

1. Constrain what helpers you can add to a fixed list (prevent typos).
2. Makes the hashmap lookup cheaper - hashing a single-byte `enum` (`u8`) is cheaper than hashing a string.

PR #6601, opened 2024-10-15
refactor(transformer): `HelperLoader`: add import immediately

overlookmotel:
Helper loader does not need to store details for import statements and then add them in an `exit_program` visitor. It can add them immediately.

That means `HelperLoader` transform can be removed, and its public methods are implemented directly on `TransformCtx`.

### DIFF
--- a/crates/oxc_transformer/src/common/helper_loader.rs
+++ b/crates/oxc_transformer/src/common/helper_loader.rs
@@ -1,21 +1,21 @@
-//! Utility transform to load helper functions.
+//! Utility to load helper functions.
 //!
 //! This module provides functionality to load helper functions in different modes.
 //! It supports runtime, external, and inline (not yet implemented) modes for loading helper functions.
 //!
 //! ## Usage
 //!
-//! You can call [`HelperLoaderStore::load`] to load a helper function and use it in your CallExpression.
+//! You can call [`TransformCtx::helper_load`] to load a helper function and use it in your CallExpression.
 //!
 //! ```rs
-//! let callee = self.ctx.helper_loader.load("helperName");
+//! let callee = self.ctx.helper_load("helperName");
 //! let call = self.ctx.ast.call_expression(callee, ...arguments);
 //! ```
 //!
-//! And also you can call [`HelperLoaderStore::call`] directly to load and call a helper function.
+//! And also you can call [`TransformCtx::helper_call`] directly to load and call a helper function.
 //!
 //! ```rs
-//! let call_expression = self.ctx.helper_loader.call("helperName", ...arguments);
+//! let call_expression = self.ctx.helper_call("helperName", ...arguments);
 //! ```
 //!
 //! ## Modes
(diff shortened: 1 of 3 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6601?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6601** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6599** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6599?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6596** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6595** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel:
I think this is *probably* an improvement, but not 100% sure. Please let me know if you think it was better before.

If we do merge this, then I'll do a follow-up to move the implementation out of the "common" transforms directory and into `context`, since it's no longer a transform.

crates/oxc_transformer/src/common/helper_loader.rs
```diff
@@ -182,16 +164,19 @@ impl<'a> HelperLoaderStore<'a> {
             loaded_helpers: RefCell::new(FxHashMap::default()),
         }
     }
+}
 
+// Public methods implemented directly on `TransformCtx`, as they need access to `TransformCtx::module_imports`.
+impl<'a> TransformCtx<'a> {
```
Dunqing, line 170: Do you think this is a good place to support helper-related methods for `TransformCtx`

Dunqing: Just saw your [comment](https://github.com/oxc-project/oxc/pull/6601#issuecomment-2414554481).

Dunqing:
> I think this is _probably_ an improvement, but not 100% sure. Please let me know if you think it was better before.
> 
> If we do merge this, then I'll do a follow-up to move the implementation out of the "common" transforms directory and into `context`, since it's no longer a transform.

Yes, this is indeed a performance improvement. I am just a little confused, here added some methods for `TransformerCtx`.  We can merge this. I am looking forward to your follow-up PRs

overlookmotel:
### Merge activity

* **Oct 16, 9:01 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
> I am just a little confused, here added some methods for `TransformerCtx`.

The idea is:

Previously, `HelperLoaderStore` stored the details of imports internally, and then at the very end of traversal, in `exit_program`, it called `ModuleImportsStore` to add those imports.

I realized that there was no need to wait until the end to pass the details of imports to `ModuleImportsStore`. It could just do it immediately when you request a helper.

But to do that, `load` needs access *both* `HelperLoaderStore` and `ModuleImportsStore`, which are both properties of `TransformCtx`.

So I moved `load` method to be a method of `TransformCtx`, so it can access both.

I don't know if that makes any more sense...

PR #6615, opened 2024-10-15
feat(linter): Implement node/no-new-require rule

nrayburn-tech:
Rust is still pretty new to me, so please let me know if something seems incorrect.

I think the rule might be fixable.  We can be pretty confident that the code is wrong and propose one of the three valid alternatives.  Unfortunately, I have no idea how that would be implemented.

Ref #493.


DonIsaac:
Duplicate of #6165

DonIsaac:
Thanks for your contribution, but I think this is redundant as there's already a PR out for it. Maybe try another rule?

PR #6622, opened 2024-10-16
fix(isolated_declarations): fix potential memory leak

overlookmotel:
`Scope` contains 2 x `FxHashMap`s, which own data outside of the arena. So this data will not be dropped when the allocator is dropped.

The scope for this becoming a memory leak in practice is limited for 2 reasons:

1. All `Scope`s except the root one are popped from the stack by the end of traversal. That last scope's hashmaps are always empty, unless there are unresolved references (references to globals).

2. `oxc_allocator::Vec` is currently `Drop`.

However, `oxc_allocator::Vec` will cease to be `Drop` in future, at which point this would become a real memory leak.

Additionally, it doesn't make sense to store temporary data in the arena, as the arena is intended to hold data that needs to live as long as the AST, which temporary data doesn't.

### DIFF
--- a/crates/oxc_isolated_declarations/src/lib.rs
+++ b/crates/oxc_isolated_declarations/src/lib.rs
@@ -71,7 +71,7 @@ impl<'a> IsolatedDeclarations<'a> {
             ast: AstBuilder::new(allocator),
             strip_internal,
             internal_annotations: FxHashSet::default(),
-            scope: ScopeTree::new(allocator),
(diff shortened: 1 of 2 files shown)


Dunqing:
### Merge activity

* **Oct 16, 11:05 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6626, opened 2024-10-16
fix(allocator): prevent putting drop types into arena

overlookmotel:
Should not be merged.

Is a quick and dirty check that we do not put any drop types into either `oxc_allocator::Box` or `oxc_allocator::Vec`.

Unfortunately, this is probably not reliable enough to use. From [`needs_drop`'s docs](https://doc.rust-lang.org/std/mem/fn.needs_drop.html):

> May be implemented conservatively: it may return `true` for types that don’t actually need to be dropped. As such always returning `true` would be a valid implementation of this function. However if this function actually returns `false`, then you can be certain dropping `T` has no side effect.

So the fact that tests and conformance pass on this PR does prove that no code path that they exercise attempts to put `Drop` types into `Box` or `Vec`. So this demonstrates that #6623 does not introduce any memory leaks. Good!

However, as the docs say, we cannot rely on `needs_drop` not producing false positives, so this could start producing erroneous compilation errors in a future version of Rust.

### DIFF
--- a/crates/oxc_allocator/src/boxed.rs
+++ b/crates/oxc_allocator/src/boxed.rs
@@ -7,6 +7,7 @@ use std::{
     fmt::{self, Debug, Formatter},
     hash::{Hash, Hasher},
     marker::PhantomData,
+    mem::needs_drop,
     ops::{self, Deref},
     ptr::{self, NonNull},
 };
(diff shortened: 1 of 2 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6626?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6626** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6626?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6623** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6623?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6622** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6622?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

PR #6639, opened 2024-10-17
fix(transformer): Correctly trim JSX

### LINKED ISSUE #6638: Transform: Tab `\t` in JSX should be trimmed
[OXC playground](https://oxc-project.github.io/oxc/playground/?code=3YCAAIDcgYCAgICAgICyHorESipoTXg5tGjCQfrkttQjp%2BeydgrtwPxXWnKjfGJOpOItqQuFiwzy7SfXue0uloeaWYLNd3DtKjmvp1cYlWCbw2DxU%2FrgLsrAO4E8Wiu0Nr02kKm0lOQJh%2BQ%2BShYXYge9zaPLAXv3xvOc3pdePtmwzid4Ii1mEQo0ZC6KamShhMEBcTav8ibwZKjs8tv5uIYGW6zJcNe7ORvrlGsYw6fbkj6hfjm6e70V14jYFUgS7yNKWtTAEJAqF4fcYQHvxWSMnR5WeF9oYVBlHBJCd%2BRFJ0UleeHATGEDqHr0974%3D)

magic-akari:
- Closes: #6638

### DIFF
--- a/crates/oxc_syntax/src/identifier.rs
+++ b/crates/oxc_syntax/src/identifier.rs
@@ -28,15 +28,45 @@ pub const VT: char = '\u{b}';
 /// U+000C FORM FEED, abbreviated `<FF>`.
 pub const FF: char = '\u{c}';
 
+/// U+0020 SPACE, abbreviated `<SP>`.
+pub const SP: char = '\u{20}';
+
 /// U+00A0 NON-BREAKING SPACE, abbreviated `<NBSP>`.
 pub const NBSP: char = '\u{a0}';
+
+// U+0085 NEXT LINE, abbreviated `<NEL>`.
+const NEL: char = '\u{85}';
+
+const OGHAM_SPACE_MARK: char = '\u{1680}';
+
+const EN_QUAD: char = '\u{2000}';
+
+// U+200B ZERO WIDTH SPACE, abbreviated `<ZWSP>`.
+const ZWSP: char = '\u{200b}';
+
+// Narrow NO-BREAK SPACE, abbreviated `<NNBSP>`.
+const NNBSP: char = '\u{202f}';
+
+// U+205F MEDIUM MATHEMATICAL SPACE, abbreviated `<MMSP>`.
+const MMSP: char = '\u{205f}';
+
+const IDEOGRAPHIC_SPACE: char = '\u{3000}';
 
 pub fn is_irregular_whitespace(c: char) -> bool {
     matches!(
         c,
-        VT | FF | NBSP | ZWNBSP | '\u{85}' | '\u{1680}' | '\u{2000}'
-            ..='\u{200a}' | '\u{202f}' | '\u{205f}' | '\u{3000}'
+        VT | FF | NBSP | ZWNBSP | NEL | OGHAM_SPACE_MARK | EN_QUAD
+            ..ZWSP | NNBSP | MMSP | IDEOGRAPHIC_SPACE
     )
+}
+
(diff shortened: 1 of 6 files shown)


crates/oxc_syntax/src/identifier.rs
```diff
@@ -39,6 +42,26 @@ pub fn is_irregular_whitespace(c: char) -> bool {
...
+// Medium Mathematical Space
+const MMSP: char = '\u{205f}';
+const IDEOGRAPHIC_SPACE: char = '\u{3000}';
+
+// https://github.com/microsoft/TypeScript/blob/b8e4ed8aeb0b228f544c5736908c31f136a9f7e3/src/compiler/scanner.ts#L556
+pub fn is_white_space_single_line(c: char) -> bool {
+    // Note: nextLine is in the Zs space, and should be considered to be a whitespace.
+    // It is explicitly not a line-break as it isn't in the exact set specified by EcmaScript.
+    matches!(c, |SP| TAB | VT | FF | NBSP | NEL | OGHAM_SPACE_MARK | EN_QUAD
+        ..=ZWSP | NNBSP | MMSP | IDEOGRAPHIC_SPACE | ZWNBSP)
```
overlookmotel, line 62: Style nit (remove the `|` before `SP`, and then rustfmt does the rest):

```suggestion
    matches!(
        c,
        SP | TAB | VT | FF | NBSP | NEL | OGHAM_SPACE_MARK | EN_QUAD
            ..=ZWSP | NNBSP | MMSP | IDEOGRAPHIC_SPACE | ZWNBSP
    )
```

overlookmotel:
I wonder if we should be doing this in parser not transformer (see https://github.com/oxc-project/oxc/issues/6638#issuecomment-2419133045). But, even if that's the case, it'd be good to get this merged as the output at present is terrible!

One question:

It looks like `is_white_space_single_line` should be equivalent to:

```rs
fn is_white_space_single_line(c: char) -> bool {
    matches!(c, SP | TAB) || is_irregular_whitespace(c)
}
```

That'd be preferable, as it's easier to understand.

But... why does `is_white_space_single_line` return `true` for `ZWSP` (`\u{200b}`), and `is_irregular_whitespace` returns `false`? Is `is_irregular_whitespace` wrong, or is `is_white_space_single_line` wrong? The discrepancy is:

* `is_irregular_whitespace` matches `'\u{2000}'..='\u{200a}'`
* `is_white_space_single_line` matches `EN_QUAD..=ZWSP`, which is equivalent to `'\u{2000}'..='\u{200b}'`

magic-akari:
> It looks like `is_white_space_single_line` should be equivalent to:
> 
> ```rust
> fn is_white_space_single_line(c: char) -> bool {
>     matches!(c, SP | TAB) || is_irregular_whitespace(c)
> }
> ```

This is how I initially wrote it. I also noticed the difference with ZWSP.
I attempted to compare it with the ECMAScript specification at https://tc39.es/ecma262/#sec-white-space ,
but neither `is_irregular_whitespace` nor `is_white_space_single_line` matched it. For example, NEL (U+0085) isn't listed. 
So I simply, inherited TypeScript's logic.

And `is_irregular_whitespace` was kept because it's used elsewhere.

overlookmotel:
@Boshen I think we need your expertise on this. Why does `is_irregular_whitespace` not match `ZWSP` (`\u{200b}`)?

overlookmotel (approved):

overlookmotel:
### Merge activity

* **Oct 17, 10:40 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
Thanks! I've opened #6646 on the ZWSP question.

PR #6632, opened 2024-10-17
refactor(transformer/async-to-generator): move internal methods below entry points

Dunqing:

PR #6640, opened 2024-10-17
feat(transformer/async-to-generator): handle arrow-function correctly

Dunqing:


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6640?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6640** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6640?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6634** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6634?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6633** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6632** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

overlookmotel (approved):
I have some suggestions to shorten and simplify this transform. But let's merge first, and I'll write those suggestions up shortly.

overlookmotel:
### Merge activity

* **Oct 17, 4:41 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
My suggestions for follow-ups (which you are free to ignore!):

### Small things

* `exit_expression`: Style: Convert the 3-branch `if` / `else if` into a `match`.
* `exit_statement`: Style: Reduce indentation by starting function with `let Statement::FunctionDeclaration(func) = stmt else { return };`
* `exit_statement`: `Function::id` field is always `Some` for function declarations. Replace `if let Some(id) = func.id.take() {` with `let id = func.id.take().unwrap()`, and remove the `else` branch.
* `exit_expression`: No need to check `body.r#async()`. `await` cannot be used in a non-async function, so `*body.r#async()` is always `true` here.

### Refactor

Instead of `ctx.ancestry.ancestors().find_map()` and then `if in_async_function ...`, use a loop and `return` to bail out if nothing to be done. e.g.:

```rs
// Exit if this `await` is top level, or in an async generator function
for ancestor in ctx.ancestors() {
    match ancestor {
        Ancestor::FunctionBody(body) => {
            if *body.generator() {
                return;
            }
            break;
        }
        // Arrow functions are never generators
        Ancestor::ArrowFunctionExpressionBody(_) => break,
        Ancestor::ProgramBody(_) => return,
        _ => {}
    }
}

// Convert the expression to yield
*expr = ctx.ast.expression_yield(
    SPAN,
    false,
    Some(ctx.ast.move_expression(&mut await_expr.argument)),
);
```

Personally I think this is a bit clearer, though it's a matter of taste. If you prefer how it is, feel free to ignore me.

### Avoid constructing a new `Function`

`transform_function` doesn't need to construct a new `Function`, it can mutate the existing one.

* Take ownership of the `Expresson` / `Statement` in `exit_expression` and `exit_statement` via `ctx.ast.move_expression` / `move_statement`.
* Get the owned `Box<Function>` via [this pattern](https://github.com/oxc-project/oxc/blob/f3451d7ec49dc6b6af7913954aa7f6701f1dae38/crates/oxc_transformer/src/es2016/exponentiation_operator.rs#L108-L114).
* Pass `Box<'a, Function<'a>>` into `transform_function`.

Getting ownership by matching twice on `expr` being an `Expression::Function` is an ugly pattern, but the upside is (a) less operations and allocations and (b) simplify `transform_function` because only one `move_*` operation, instead of loads of little ones.

Note: I'm suggesting passing `Box<Function>` (not `Function`) into `transform_function`, because:

1. `Box` is only 8 bytes. You can avoid having a `Function` (104 bytes) on the stack.
2. Then you don't have to allocate the `Function` back into arena. `Box<Function>` already has it in arena, and you can mutate it in place.

Dunqing:
@overlookmotel Good suggestion, I will keep it in mind!

PR #6488, opened 2024-10-13
test: reduce size of vec

Dunqing:
related: https://github.com/oxc-project/oxc/issues/9706

TEST ONLY!

Everything is hard-coded, just for testing

PR #6644, opened 2024-10-17
test(traverse): add an api to get the index of the current node in Vec<'a, Statements<'a>>.

Dunqing:
Try to get the index of the current node in `Vec<'a, Statements<'a>>`.

### DIFF
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -354,6 +354,7 @@ impl<'a, 'ctx> Traverse<'a> for TransformerImpl<'a, 'ctx> {
     }
 
     fn enter_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        println!("index {:?}", ctx.ancestor_index(stmt));
         self.x0_typescript.enter_statement(stmt, ctx);
     }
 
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#6644** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6644?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Yes, agree this API would be good. But I'm afraid at present I think no way to do it without undefined behavior.

Problem is that `walk_statements` has a live mutable reference to the `Vec<Statement>`:

https://github.com/oxc-project/oxc/blob/92812349e5266b074d2cb050cc7a6f200e4ac914/crates/oxc_traverse/src/generated/walk.rs#L5646-L5656

So it's illegal to get a `&` ref to `Vec<Statement>` to get its start pointer.

But... once we take control of `Vec` (#6488), we can add APIs to allow iterating over the `Vec` while staying purely in pointer-land, without creating any `&mut` reference to it. And *then* we should be able to do this.

*Maybe* I can find a way to do it prior to implementing our own `Vec`, but it's tricky. Is there an urgent need for this in one of the transforms?

Dunqing:
@overlookmotel check your discord DM

Dunqing:
Closing as we have decided using `Address` to identifier `Statement`

PR #6658, opened 2024-10-18
feat(transformer): complete the async-to-generator plugin

Dunqing:
In this PR, most of the async functions have transformed correctly. But the async arrow functions don't fully transform correctly yet, it is related to we need to transform the arrow function to the generator function. For example:

Input:
```js
function declaration() {
  const asy = async () => {
  console.log(this.name)
 }
}
```

Output:
```js
function declaration() {
  const asy = babelHelpers.asyncToGenerator(function* () {
     console.log(this.name);
  });
}
```

Expected Output:

```js
function declaration() {
  var _this = this;
  const asy = /*#__PURE__*/function () {
    var _ref = babelHelpers.asyncToGenerator(function* () {
      console.log(_this.name);
    });
    return function asy() {
      return _ref.apply(this, arguments);
    };
  }();
}
```

From the expected output, we haven't handled `this` correctly, which means even if the `arrow-function` plugin doesn't enable, we still need to handle this correctly as the `arrow-function` plugin does, and further question if `arrow-function` plugin is enabled, how to avoid these making conflict? 

I thought we may move out the implementation of `arrow-function` and as a common helper, this way every plugin can handle this well

### DIFF
--- a/crates/oxc_transformer/src/common/statement_injector.rs
+++ b/crates/oxc_transformer/src/common/statement_injector.rs
@@ -78,7 +78,6 @@ impl<'a> StatementInjectorStore<'a> {
     }
 
     /// Add a statement to be inserted immediately after the target statement.
-    #[expect(dead_code)]
     pub fn insert_after(&self, target: &Statement<'a>, stmt: Statement<'a>) {
         let mut insertions = self.insertions.borrow_mut();
         let adjacent_stmts = insertions.entry(target.address()).or_default();
--- a/crates/oxc_transformer/src/es2017/async_to_generator.rs
+++ b/crates/oxc_transformer/src/es2017/async_to_generator.rs
@@ -1,6 +1,7 @@
-//! ES2017: Async / Await \[WIP\]
+//! ES2017: Async / Await
 //!
-//! This plugin transforms async functions to generator functions.
+//! This plugin transforms async functions to generator functions
+//! and wraps them with `asyncToGenerator` helper function.
 //!
 //! ## Example
 //!
@@ -17,16 +18,28 @@
 //! }
 //! ```
 //!
-//! Output (Currently):
+//! Output:
 //! ```js
 //! function foo() {
-//!   return _asyncToGenerator(function* () {
-//!     yield bar();
-//!   })
+//!   return _foo.apply(this, arguments);
 //! }
-//! const foo2 = () => _asyncToGenerator(function* () {
+//! function _foo() {
+//!   _foo = babelHelpers.asyncToGenerator(function* () {
+//!           yield bar();
+//!   });
+//!   return _foo.apply(this, arguments);
+//! }
+//! const foo2 = function() {
+//!   var _ref = babelHelpers.asyncToGenerator(function* () {
+//!           yield bar();
+//!   });
+//!   return function foo2() {
+//!      return _ref.apply(this, arguments);
+//!   };
+//! }();
+//! babelHelpers.asyncToGenerator(function* () {
 //!   yield bar();
-//! }
+//! });
 //! ```
 //!
 //! ## Implementation
@@ -35,21 +48,16 @@
 //!
 //! Reference:
 //! * Babel docs: <https://babeljs.io/docs/en/babel-plugin-transform-async-to-generator>
-//! * Esbuild implementation: <https://github.com/evanw/esbuild/blob/main/internal/js_parser/js_parser_lower.go#L392>
 //! * Babel implementation: <https://github.com/babel/babel/blob/main/packages/babel-plugin-transform-async-to-generator>
-//! * Babel helper implementation: <https://github.com/babel/babel/blob/main/packages/babel-helper-remap-async-to-generator>
 //! * Async / Await TC39 proposal: <https://github.com/tc39/proposal-async-await>
 
-use oxc_ast::{
-    ast::{
-        ArrowFunctionExpression, Expression, Function, FunctionType, Statement,
-        VariableDeclarationKind,
-    },
-    NONE,
-};
-use oxc_semantic::ScopeFlags;
-use oxc_span::{GetSpan, SPAN};
-use oxc_traverse::{Ancestor, Traverse, TraverseCtx};
+use std::mem;
+
+use oxc_allocator::Box;
+use oxc_ast::{ast::*, Visit, NONE};
+use oxc_semantic::{ReferenceFlags, ScopeFlags, ScopeId, SymbolFlags};
+use oxc_span::{Atom, GetSpan, SPAN};
+use oxc_traverse::{Ancestor, BoundIdentifier, Traverse, TraverseCtx};
 
 use crate::{common::helper_loader::Helper, TransformCtx};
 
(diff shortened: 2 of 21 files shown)


Dunqing:
* **#6658** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_semantic/src/scope.rs
```diff
@@ -173,6 +173,16 @@ impl ScopeTree {
         }
     }
 
+    pub fn change_parent_id(&mut self, scope_id: ScopeId, new_parent_id: Option<ScopeId>) {
+        let old_parent_id = self.parent_ids[scope_id];
+        if self.build_child_ids {
+            if let Some(old_parent_id) = old_parent_id {
+                self.child_ids[old_parent_id].retain(|&child_id| child_id != scope_id);
+            }
+        }
```
overlookmotel, line 182: I think need to record this scope as child of its new parent too.

```suggestion
        if self.build_child_ids {
            let old_parent_id = self.parent_ids[scope_id];
            if let Some(old_parent_id) = old_parent_id {
                self.child_ids[old_parent_id].retain(|&child_id| child_id != scope_id);
            }
            if let Some(new_parent_id) = new_parent_id {
                self.child_ids[new_parent_id].push(scope_id);
            }
        }
```

Dunqing: Yes, it was handled in `self.set_parent_id`. But it causes two repeated branch checks for `self.build_child_ids`. I will refactor it so that it doesn't call `self.set_parent_id` anymore

crates/oxc_transformer/src/common/statement_injector.rs
```diff
@@ -131,6 +130,8 @@ impl<'a> StatementInjectorStore<'a> {
                     new_statements.push(stmt);
                     new_statements.extend(adjacent_stmts.into_iter().map(|s| s.stmt));
                 }
+            } else {
+                new_statements.push(stmt);
```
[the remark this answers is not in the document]
Dunqing: Good catch! I didn't realize this

overlookmotel:
Oh I'm sorry. #6856 or #6858 has caused merge conflicts and likely broken things. I should have reviewed and merged this PR first.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -17,16 +17,28 @@
...
 //! ```
 //!
-//! Output (Currently):
+//! Output:
 //! ```js
 //! function foo() {
-//!   return _asyncToGenerator(function* () {
-//!     yield bar();
-//!   })
+//!        return _foo.apply(this, arguments);
```
overlookmotel, line 23: Nit: The indentation in these examples seems a bit wacky/excessive.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -1,6 +1,6 @@
 //! ES2017: Async / Await \[WIP\]
 //!
-//! This plugin transforms async functions to generator functions.
+//! This plugin transforms async functions to generator functions and wraps them in the asyncToGenerator helper function.
```
overlookmotel, line 3: Style nit!

```suggestion
//! This plugin transforms async functions to generator functions
//! and wraps them with `asyncToGenerator` helper function.
```

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -35,21 +47,18 @@
...
-};
-use oxc_semantic::ScopeFlags;
-use oxc_span::{GetSpan, SPAN};
-use oxc_traverse::{Ancestor, Traverse, TraverseCtx};
+use std::mem;
+
+use oxc_allocator::{Box, GetAddress};
+use oxc_ast::ast::*;
+use oxc_ast::Visit;
+use oxc_ast::NONE;
```
overlookmotel, line 58: Another style nit. Don't we usually go with this style?

```suggestion
use oxc_ast::{ast::*, Visit, NONE};
```

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -65,138 +74,318 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
 
 impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
-    fn transform_function(
+    /// Transforms `await` expressions to `yield` expressions.
+    /// Ignores top-level await expressions.
+    #[allow(clippy::unused_self)]
+    fn transform_await_expression(
         &self,
-        func: &mut Function<'a>,
+        expr: &mut Box<'a, AwaitExpression<'a>>,
```
overlookmotel, line 133: `&mut Box<'a, AwaitExpression<'a>>` is a double-reference (`&mut` is a reference, and so is `Box`). It might make more sense to take a `&mut AwaitExpression<'a>`.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -65,138 +74,318 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
+    /// class A {
+    /// foo() {
+    ///     return babelHelpers.asyncToGenerator(function* () {
+    ///         yield bar();
+    ///     })();
+    /// }
+    /// ```
+    fn transform_function_for_method_definition(
+        &self,
+        func: &mut Box<'a, Function<'a>>,
```
overlookmotel, line 168: Same as above. Probably better to take a `&mut Function<'a>`.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -65,138 +74,318 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
+        func.r#async = false;
+        func.body = Some(ctx.ast.alloc_function_body(SPAN, ctx.ast.vec(), ctx.ast.vec1(statement)));
+        func.scope_id.set(Some(wrapper_scope_id));
+    }
+
+    /// Transforms [`Function`] whose type is [`FunctionType::FunctionExpression`] to a generator function
+    /// and wraps it in asyncToGenerator helper function.
+    fn transform_function_expression(
+        &self,
+        wrapper_function: &mut Box<'a, Function<'a>>,
```
overlookmotel, line 208: Same as above. Probably better to take a `&mut Function<'a>`.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -65,138 +74,318 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
+
+        // Construct the IIFE
+        let callee = ctx.ast.expression_from_function(ctx.ast.move_function(wrapper_function));
+        Some(ctx.ast.expression_call(SPAN, callee, NONE, ctx.ast.vec(), false))
+    }
+
+    /// Transforms async function declarations into generator functions wrapped in the asyncToGenerator helper.
+    fn transform_function_declaration(
+        &self,
+        wrapper_function: &mut Box<'a, Function<'a>>,
```
overlookmotel, line 313: Same as above. Probably better to take a `&mut Function<'a>`.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -208,24 +397,306 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
+            let params = Self::create_empty_params(ctx);
+            let wrapper_function = Self::create_function(None, params, body, wrapper_scope_id, ctx);
+            // Construct the IIFE
+            let callee = ctx.ast.expression_from_function(wrapper_function);
+            Some(ctx.ast.expression_call(SPAN, callee, NONE, ctx.ast.vec(), false))
+        }
+    }
+
+    /// Infers the function id from [`Ancestor::VariableDeclaratorInit`].
+    #[inline]
```
overlookmotel, line 453: What's the reason for `#[inline]` here?

Generally it's better to leave decisions on whether to inline or not to the compiler, except when:

1. the function is trivial (only a couple of assembly instructions) or
2. there's some specific circumstances where you know more than the compiler, or want to "tell" the compiler something.

An example of (2) is when the function needs to be inlined for the compiler to prove an `unwrap()` will never panic and that it can remove it:

```rs
fn caller(expr: &Expression) {
    if matches!(expr, Expression::BinaryExpression(_) {
        // `expr` is definitely a `BinaryExpression`
        callee(expr);
    }
}

// `#[inline]` so that compiler can prove that `expr` is always a `BinaryExpression`
// and remove the `unreachable!()` branch
#[inline]
fn callee(expr: &Expression) {
    // If this function isn't inlined, the compiler knows nothing about the type of `expr`
    // because each function is optimized separately.
    // But if this function is inlined into `caller`, then compiler sees all the code together.
    // It knows `expr` is a `BinaryExpression` because of the `if` statement in `caller`,
    // so it will remove the `else { unreachable!() }` branch here.
    let Expression::BinaryExpression(bin_expr) = expr else { unreachable!() };
}
```

If you're doing something like this, include a comment to explain why `#[inline]` (like the comment above `callee` in example above).

But my rule of thumb is that if I'm not sure whether to inline or not, don't. I trust the compiler more than myself!

Same for most of the other functions below.

crates/oxc_transformer/src/es2017/async_to_generator.rs
```diff
@@ -208,24 +397,306 @@ impl<'a, 'ctx> AsyncToGenerator<'a, 'ctx> {
...
+        };
+        let Some(id) = declarator.id().get_binding_identifier() else { unreachable!() };
+        Some(
+            ctx.generate_binding(id.name.clone(), scope_id, SymbolFlags::FunctionScopedVariable)
+                .create_binding_identifier(ctx),
+        )
+    }
+
+    /// Creates a [`Function`] with the specified params, body and scope_id.
+    #[inline]
```
overlookmotel, line 469: `#[inline]` seems sensible here. This function just constructs a type, and inlining it makes it more likely that compiler will see it can construct the type directly in arena, rather than constructing on stack and then moving to the arena.

overlookmotel (approved):
I've not reviewed the logic of this transform, as I don't know it well enough. But this PR makes the tests pass so 👍!

All my comments are style nits, and some notes about use of `#[inline]`. I hope that's useful, and not annoying!

crates/oxc_transformer/src/es2017/async_to_generator.rs
Dunqing: Thanks for explaining a lot, I understand now!

Dunqing:
Mering this!

Dunqing:
### Merge activity

* **Oct 24, 11:18 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6665, opened 2024-10-18
fix(linter/no-cond-assign): false positive when assignment is in body statement

### LINKED ISSUE #6656: linter: no-cond-assign: false positive

0.9.10

### What command did you run?

oxlint -c .oxlintrc.json -- path/to/example.ts

{
  "rules": {
    "no-cond-assign": ["error", "always"]
  }
}

### What happened?

Same-line conditional blocks that contain an assignment falsely trigger the `no-cond-assign` rule, when they do not in ESLint:

```ts
const value: string = 'a';
let newValue = '';
if (['a', 'b', 'c', 'd'].includes(value)) newValue = value;
else newValue = 'default';
console.log(newValue);
```

Lines 3 and 4 both produce a rule violation:

```
  × eslint(no-cond-assign): Expected a conditional expression and instead saw an assignment
   ╭─[path/to/example.ts:3:52]
 2 │ let newValue = '';
 3 │ if (['a', 'b', 'c', 'd'].includes(value)) newValue = value;
   ·                                                    ─
 4 │ else newValue = 'default';
   ╰────
  help: Consider wrapping the assignment in additional parentheses

  × eslint(no-cond-assign): Expected a conditional expression and instead saw an assignment
   ╭─[path/to/example.ts:4:15]
 3 │ if (['a', 'b', 'c', 'd'].includes(value)) newValue = value;
 4 │ else newValue = 'default';
   ·               ─
 5 │ console.log(newValue);
```

The ESLint rule is intended to catch things like `if (foo = 'bar')`, where an assignment happens _inside_ a conditional block; this is catching things _after_ that block.

camchenry:
- fixes https://github.com/oxc-project/oxc/issues/6656

rather than reporting any assignment inside of `if`, `while`, etc. when the `always` option is enabled, we only check if the assignment resides in specific areas that assignment should not be. for example, inside the test of an `if`, or the test of a `for` loop.

incidentally, this also fixes an issue (seen in snapshot file) where the same assignment was reported twice.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_cond_assign.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_cond_assign.rs
@@ -78,21 +78,46 @@ impl Rule for NoCondAssign {
                 self.check_expression(ctx, expr.test.get_inner_expression());
             }
             AstKind::AssignmentExpression(expr) if self.config == NoCondAssignConfig::Always => {
-                for node_id in ctx.nodes().ancestors(node.id()).skip(1) {
-                    match ctx.nodes().kind(node_id) {
-                        AstKind::IfStatement(_)
-                        | AstKind::WhileStatement(_)
-                        | AstKind::DoWhileStatement(_)
-                        | AstKind::ForStatement(_)
-                        | AstKind::ConditionalExpression(_) => {
-                            Self::emit_diagnostic(ctx, expr);
+                let mut spans = vec![];
+                for ancestor in ctx.nodes().iter_parents(node.id()).skip(1) {
+                    match ancestor.kind() {
+                        AstKind::IfStatement(if_stmt) => {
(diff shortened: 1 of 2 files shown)


camchenry:
* **#6665** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6665?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

DonIsaac (approved):

DonIsaac:
### Merge activity

* **Oct 18, 11:16 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6699, opened 2024-10-20
feat(linter): Added fixer for duplicate prefix in valid title jest rule

tapanprakasht:
Supported fixer for duplicate prefix in jest title rule.

### DIFF
--- a/crates/oxc_linter/src/rules/jest/valid_title.rs
+++ b/crates/oxc_linter/src/rules/jest/valid_title.rs
@@ -316,7 +316,7 @@ fn validate_title(
         let (error, help) = Message::DuplicatePrefix.detail();
         ctx.diagnostic_with_fix(valid_title_diagnostic(error, help, span), |fixer| {
             let replaced_title = title[first_word.len()..].trim().to_string();
-            fixer.replace(Span::new(span.start + 1, span.end - 1), replaced_title)
+            fixer.replace(span.shrink_left(1).shrink_right(1), replaced_title)
         });
         return;
     }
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/jest/valid_title.rs
```diff
@@ -312,9 +312,12 @@ fn validate_title(
         return;
     };
 
-    // TODO: support fixer
     if first_word == un_prefixed_name {
-        Message::DuplicatePrefix.diagnostic(ctx, span);
+        let (error, help) = Message::DuplicatePrefix.detail();
+        ctx.diagnostic_with_fix(valid_title_diagnostic(error, help, span), |fixer| {
+            let replaced_title = title[first_word.len()..].trim().to_string();
```
DonIsaac, line 318: I think to_string is unnecessary

crates/oxc_linter/src/rules/jest/valid_title.rs
```diff
@@ -312,9 +312,12 @@ fn validate_title(
...
+            fixer.replace(Span::new(span.start + 1, span.end - 1), replaced_title)
```
DonIsaac, line 319: Consider shrink_left and shrink_right

tapanprakasht: Done. Moved to shrink methods.

crates/oxc_linter/src/rules/jest/valid_title.rs
tapanprakasht: Tried removing the to_string(). Getting lifetime issues.

Issue #5816, opened 2024-09-16
feat(semantic): add `SymbolId` and `SymbolFlags` to `ModuleRecord`


DonIsaac:
This allows exporting modules to provide symbol information to modules importing them.

I'm currently thinking we could add it to `NameSpan`, kind of like this:

```rust
pub struct NameSpan {
  name: CompactStr,
  span: Span,
  symbol_id: OnceLock<SymbolId>, // NameSpan must be Send + Sync
  symbol_flags: OnceLock<SymbolFlags>,
}
```

Binding this data to `NameSpan` will require a change to when/how `ModuleRecordBuilder` is used. it currently runs before `SemanticBuilder::build`. Should we:
1. Bind symbol info during SemanticBuilder's traversal? Pros: fewer AST iterations. Cons: var hoisting and visit order of export declarations (e.g. `export const = 5`) may make this challenging
2. Run `ModuleRecordBuilder` _after_ `Semantic` has been built. Pros: much simpler. Cons: requires 2 AST traversals.

### Other Questions
- Is there a better place to put this information?
- Should we include anything else besides id and flags?

overlookmotel:
### API change

We would lose one thing with this change: Currently you can build `ModuleRecord` without also building `Semantic`, but after this change you won't be able to, as it needs to run semantic analysis to generate `SymbolId`s.

But is that an important ability to preserve? (I doubt it, `oxc_module_lexer` doesn't use `oxc_semantic`). @Boshen?

We'd need to change API to something like one of these:

```rs
let semantic = SemanticBuilder::new(&source_text)
        .with_module_record(path)
        .build(&program);
let module_record = &semantic.module_record;
```

```rs
let semantic = SemanticBuilder::new(&source_text).build(&program);
let module_record = semantic.build_module_record(path, &program);
```

```rs
let semantic = SemanticBuilder::new(&source_text).build(&program);
let module_record = ModuleRecordBuilder::new(path, &semantic).build(&program);
```

Personally, I think any of these is an improvement. I find it unintuitive that `SemanticBuilder` uses the `with_*` pattern for everything else, but `build_module_record` behaves differently.

### Implementation

> 2. Run `ModuleRecordBuilder` *after* `Semantic` has been built. Pros: much simpler. Cons: requires 2 AST traversals.

`ModuleRecordBuilder` doesn't do a full traversal, does it? It just iterates through the `Vec` of top-level statements. So a 2nd pass is not super-costly.

[...]

PR #6701, opened 2024-10-20
fix(semantic): get_symbol_id_from_span not searching symbol redeclarations

### LINKED ISSUE #6689: Variable and Type declarations are merged into a single symbol in oxc_semantic
## Summary

A common pattern to define enum in TS is:
```ts
export const Foo = [
  'bar',
  'baz'
] as const;

export type Foo = typeof Foo[number];
```

There are 2 distinct declarations here with the same name, but oxc_semantic symbol table merges them into one. 

## Actual result

`const Foo` is recorded in the symbol table. But `type Foo` is entirely missing.
Instead the `const Foo` declaration gains a `TypeAlias` flag.

[OXC Playground](https://playground.oxc.rs/#eNpFjksOgzAMRK8SecOGE1B1yyWARaCmQkrsyAktFOXuJWpCV37jz3gOmKAB3BxLUBOTD6plVnfV9aRUNWqp6kyfqqdBaf9bu/XUU74Lu8N8lpDnJDpa7Ygy3KAGhuYAWSkVv1PQGzRBVqzBLBQK+4kdXmK3I5uigmjyM4uFZtbGY6zBafEop+PJyaXwtVoaEz/wien3KexCy7z8RxSETWv4naK9UEb2OUKM8QtSumPE)

```json
[
  {
    "name": "Foo",
    "span": {
      "start": 13,
      "end": 16
    },
    "nodeId": 4,
    "scopeId": 0,
    "flags": "BlockScopedVariable | ConstVariable | Export | TypeAlias",
    "references": [
      {
        "nodeId": 24,
        "flags": "Type"
      }
    ]
  }
]
```

## Expected result

`const Foo` and `type Foo` should be treated as 2 distinct symbols: the same way they would be if they were named differently.

[OXC Playground](https://playground.oxc.rs/#eNpFjjEOgzAMRa8SeWHhBKCunKAbMARqKqRgR05ooSh3b6ISOvl929/+B4xQAW6WxauRyXnVMKubajtSqhi0FOVJn6KjXmn3W6s76uj0+d1ist2jLzFPSbW0LgNKX0MJDNUBslIqbievN6i8rFiCmclndiNbvMS+DGyy8qLJTSwLVJM2DkMJVotDiRcjpyuZr9XcGPmBT0y/o1hmmqf5PyIvbBrD7xTthTKwOyOEEL6xGWQY)

```json
[
  {
    "name": "Foo",
    "span": {
      "start": 13,
      "end": 16
    },
    "nodeId": 4,
    "scopeId": 0,
    "flags": "BlockScopedVariable | ConstVariable | Export",
    "references": [
      {
        "nodeId": 24,
        "flags": "Type"
      }
    ]
  },
  {
    "name": "Foo",
    "span": {
      "start": 63,
      "end": 66
    },
    "nodeId": 19,
    "scopeId": 0,
    "flags": "Export | TypeAlias",
    "references": []
  }
]
```

This would also match what TypeScript does ([TS AST Viewer](https://ts-ast-viewer.com/#code/KYDwDg9gTgLgBAYwgOwM7wGIQnAvHAbQCg44ByAIwEMoyAaE86gLzKIF04rVEV0BuIkVCRYcGAE8wwOFhz5J0iADNZ2AsgCuAWwrAo7fkA))
ts-symbol-definition

valeneiko:
Fixes #6689

### DIFF
--- a/crates/oxc_semantic/src/symbol.rs
+++ b/crates/oxc_semantic/src/symbol.rs
@@ -92,6 +92,19 @@ impl SymbolTable {
         self.spans
             .iter_enumerated()
             .find_map(|(symbol, &inner_span)| if inner_span == span { Some(symbol) } else { None })
+            .or_else(|| self.get_symbol_id_from_redeclaration_span(span))
+    }
+
+    fn get_symbol_id_from_redeclaration_span(&self, span: Span) -> Option<SymbolId> {
+        self.redeclarations
+            .iter_enumerated()
+            .filter(|(_, &redeclaration)| redeclaration.is_some())
+            .flat_map(|(symbol, &redeclaration)| {
+                self.redeclaration_spans[redeclaration.unwrap()]
+                    .iter()
+                    .map(move |&inner_span| (symbol, inner_span))
+            })
+            .find_map(|(symbol, inner_span)| if inner_span == span { Some(symbol) } else { None })
     }
 
     /// Get the [`Span`] of the [`AstNode`] declaring a symbol.
--- a/crates/oxc_semantic/tests/integration/symbols.rs
+++ b/crates/oxc_semantic/tests/integration/symbols.rs
@@ -1,6 +1,9 @@
-use oxc_semantic::{Reference, SymbolFlags};
+use std::rc::Rc;
 
-use crate::util::SemanticTester;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_semantic::{Reference, Semantic, SymbolFlags, SymbolId};
+
+use crate::util::{Expect, SemanticTester};
 
 #[test]
 fn test_class_simple() {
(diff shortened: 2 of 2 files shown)


Dunqing:
Are there any situations where we need to use this method?

The method looks a little complicated and not very efficient. I have checked our codebase only three places used this method and it looks like we can replace them.

To remove this method, we need to add `SymbolId` to` ModuleRecord`'s ImportEntry struct. Is it possible? @Boshen

valeneiko:
@overlookmotel 
I am building a whole-program dependency graph between symbols E.g.:
- `const a = b+c;` - Symbol `a` depends on `b` and `c`
- `function foo() { return bar(); }` - Symbol `foo` depends on `bar`

I am doing it by iterating over symbol references, for each reference finding the ancestor declaration in root scope and resolving that declaration's AstNode into a Symbol. 

`get_symbol_id_from_span` only works with spans of names already. So I could just do `span.source_text(src)` and use `get_symbol_by_name`. Or if you know of a better way to do all this, I am open to suggestions.

For now the goal was to just explore what can be done with such a graph.


@Dunqing as for the perf of this specific PR. I think it is fine, because symbol redeclarations are rare, so in reality that whole method will most of the time be an empty iterator and occasionally iterate over very few items. As for the `get_symbol_id_from_span` itself, yes it could be improved, or removed if a better alternative exists.

Dunqing:
@valeneiko Sorry for the response too late

Each reference has stored `symbol_id`, I don't think we need to get `symbol_id` by `span`.

Dunqing:
> To remove this method, we need to add `SymbolId` to` ModuleRecord`'s ImportEntry struct. Is it possible? @Boshen

related to https://github.com/oxc-project/oxc/issues/5816#issue-2529431945

valeneiko:
I'm happpy to close the PR. For my use case I instead used `semantic.scopes().get_root_binding(name)`, where `name` comes from `decl.id.name` (instead of previous `decl.id.span`)

Dunqing:
> I'm happpy to close the PR. For my use case I instead used `semantic.scopes().get_root_binding(name)`, where `name` comes from `decl.id.name` (instead of previous `decl.id.span`)

Oh no! I think you should use `decl.id.symbol`!

valeneiko:
Wow, that's even beter! Didn't realize this existed.

Dunqing:
> Wow, that's even beter! Didn't realize this existed.

Very glad to help you! Closing as we are consider to remove the `get_symbol_id_from_span` method.

PR #6703, opened 2024-10-20
refactor(linter): `jest/valid-title` fixer to use `Span::shrink` method

tapanprakasht:
Used shrink apis from span


DonIsaac:
I forgot that I added `Span::shrink` a while ago... consider using that instead.

tapanprakasht:
> I forgot that I added `Span::shrink` a while ago... consider using that instead.

Done

camchenry (approved):

PR #6705, opened 2024-10-20
ci(linter): skip linter benchmarks on unrelated file changes

DonIsaac:
Skip building/running linter benchmarks on PRs that only change the transformer, ID, formatter, or the minifier.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -124,16 +124,33 @@ jobs:
   build-linter:
     name: Build Linter Benchmark
     runs-on: ubuntu-latest
+    outputs:
+      src: ${{ steps.filter.outputs.src }}
     steps:
       - name: Checkout Branch
         uses: taiki-e/checkout-action@v1
 
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            src:
+              - '!crates/oxc_isolated_declarations/**'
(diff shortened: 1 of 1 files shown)


camchenry (approved):

overlookmotel:
This is a really good idea. The linter benchmarks are always the slowest to complete.

I'm not familiar with `dorny/paths-filter`, so can I just check the filter is the right way around? i.e. it will run linter benchmarks if:

* PR changes at least one files, not including files in `oxc_transformer` etc.

And not:

* PR does not change any files in `oxc_transformer` etc.

It's possible a PR might make changes to *both* the linter *and* the transformer, and we'd want linter benchmarks to run in that case.

DonIsaac:
> This is a really good idea. The linter benchmarks are always the slowest to complete.
> 
> I'm not familiar with `dorny/paths-filter`, so can I just check the filter is the right way around? i.e. it will run linter benchmarks if:
> 
> * PR changes at least one file, not including files in `oxc_transformer` etc.
> 
> And not:
> 
> * PR does not change any files in `oxc_transformer` etc.
> 
> It's possible a PR might make changes to _both_ the linter _and_ the transformer, and we'd want linter benchmarks to run in that case.

I believe so. I followed the example in `CI / Conformance`. I'd like a review from @Boshen before merging just to make sure this works as I think it does.

PR #6731, opened 2024-10-21
docs(ast): fix comment of `ClassElement::r#static`

kermanx:
The comment says: `Returns true if this ClassElement is a static block or has a static modifier.`

### DIFF
--- a/crates/oxc_ast/src/ast_impl/js.rs
+++ b/crates/oxc_ast/src/ast_impl/js.rs
@@ -1303,15 +1303,14 @@ impl<'a> ClassElement<'a> {
         matches!(self, Self::StaticBlock(_))
     }
 
-    /// Returns `true` if this [`ClassElement`] is a static block or has a
+    /// Returns `true` if this [`ClassElement`] is a property and has a
     /// static modifier.
     pub fn r#static(&self) -> bool {
         match self {
-            Self::StaticBlock(_) => true,
+            Self::TSIndexSignature(_) | Self::StaticBlock(_) => false,
             Self::MethodDefinition(def) => def.r#static,
             Self::PropertyDefinition(def) => def.r#static,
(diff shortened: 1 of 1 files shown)


Dunqing (dismissed):
Thanks!

Dunqing:
Wait, I just realized this comment was added later and it's wrong.

DonIsaac:
Yeah it returns true for static properties. Static blocks aren't properties

KermanX:
I've reverted the code change and updated the comment :)

crates/oxc_ast/src/ast_impl/js.rs
```diff
@@ -1303,7 +1303,7 @@ impl<'a> ClassElement<'a> {
         matches!(self, Self::StaticBlock(_))
     }
 
-    /// Returns `true` if this [`ClassElement`] is a static block or has a
+    /// Returns `true` if this [`ClassElement`] is a property and has a
     /// static modifier.
```
overlookmotel, line 1307: Sorry to be a pain, but I don't think this is quite right either, because this suggests it has to be a property, not a method.

I suggest let's clear this up for once and for all!

```rs
/// Returns `true` if this [`ClassElement`] has a static modifier.
/// 
/// Note: Class static blocks do not have a "modifier", as there is no non-static equivalent.
/// Therefore, returns `false` for static blocks.
///
/// The following all return `true`:
/// ```ts
/// class {
///   static prop = 1;
///   static method() {}
///   static #private = 2;
///   static #privateMethod() {}
///   static accessor accessorProp = 3;
///   static accessor #privateAccessorProp = 4;
/// }
/// ```
```

overlookmotel: I'm too late. Merged while I was writing! Will make a follow-up PR.

PR #6741, opened 2024-10-21
fix(regular_expression)!: Migrate to new regexp parser API

leaysgur:
Follow up #6635 

- [x] Remove old APIs
- [x] Update linter usage
- [x] Update parser usage
- [x] Update transformer usage


crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -133,10 +133,11 @@ impl<'a, 'ctx> Traverse<'a> for RegExp<'a, 'ctx> {
             }
 
             let span = regexp.span;
+            let _flags_text = flags.to_string(); // TODO: How to add `&'a str` lifetime?
```
leaysgur, line 136: Now I am having trouble to resolve this lifetime constraint due to my lack of Rust experience...

crates/oxc_transformer/src/regexp/mod.rs
```diff
@@ -242,10 +242,12 @@ fn try_parse_pattern<'a>(
...
 ) -> Result<Pattern<'a>> {
-    use oxc_regular_expression::{Parser, ParserOptions};
-
-    let options = ParserOptions::default()
-        .with_span_offset(span.start + 1) // exclude `/`
-        .with_flags(&flags.to_string());
-    Parser::new(ctx.ast.allocator, raw, options).parse()
+    use oxc_regular_expression::{LiteralParser, Options};
+
+    let _flags_text = &flags.to_string(); // TODO: How to add `&'a str` lifetime?
```
leaysgur, line 247: Now I am having trouble to resolve this lifetime constraint due to my lack of Rust experience...

tasks/coverage/snapshots/parser_test262.snap
```diff
@@ -21017,13 +21024,27 @@ Expect Syntax Error: tasks/coverage/test262/test/language/import/import-attribut
   × Unexpected flag G in regular expression literal
     ╭─[test262/test/language/literals/regexp/early-err-bad-flag.js:19:4]
  18 │ 
+ 19 │ /./G;
+    ·    ─
+    ╰────
+
```
leaysgur, line 21030: Because new regexp parser also validates flags, it seems invalid flags are reported twice. 

Should I do something?

leaysgur:
Now escape sequences are printed correctly. ✨ 

image

tasks/coverage/snapshots/parser_test262.snap
Boshen: Interesting. We need to turn this error off in `oxc_regular_expression` if it is called from the parser.

leaysgur: Between those two, the latter feels more natural to me.

But I'd like to avoid depending error message strings to check like this...

```diff
         {
             Ok(regular_expression) => Some(self.ast.alloc(regular_expression)),
             Err(diagnostic) => {
-                self.error(diagnostic);
+                if !diagnostic.message.contains("flags") {
+                    self.error(diagnostic);
+                }
                 None
             }
         }
```

Do you have any ideas?

What about

- add the flag that the lexer detected an invalid regexp flags
- propagate it and then not parsing the regular expression in the first place

```diff
     pub(crate) fn parse_literal_regexp(&mut self) -> Result<RegExpLiteral<'a>> {
         let span = self.start_span();
         // split out pattern
-        let (pattern_end, flags) = self.read_regex()?;
+        let (pattern_end, flags, flags_invalid) = self.read_regex()?;
         let pattern_start = self.cur_token().start + 1; // +1 to exclude left `/`
         let pattern_text = &self.source_text[pattern_start as usize..pattern_end as usize];
         let flags_start = pattern_end + 1; // +1 to include right `/`
         let flags_text = &self.source_text[flags_start as usize..self.cur_token().end as usize];
         self.bump_any();
-        let pattern = self
-            .options
-            .parse_regular_expression
+        let pattern = (self.options.parse_regular_expression && !flags_invalid)
             .then_some(())
             .map(|()| {
                 self.parse_regex_pattern(pattern_start, pattern_text, flags_start, flags_text)

```

Boshen: * add the flag that the lexer detected an invalid regexp flags
* propagate it and then not parsing the regular expression in the first place

let's try this

crates/oxc_linter/src/rules/eslint/no_control_regex.rs
```diff
@@ -279,7 +267,6 @@ mod tests {
             vec![
                 r"let r = /\u{0}/u",
                 r"let r = new RegExp('\\u{0}', 'u');",
-                r"let r = new RegExp('\\u{0}', `u`);",
```
camchenry, line 282: Should we still support this? It's perfectly valid as far as I can tell, not sure what the spec says.

tasks/coverage/snapshots/parser_test262.snap
leaysgur: 👉🏻  843d6af5269c4dd0eb3087ef3686e51879c4c5b3

crates/oxc_linter/src/rules/eslint/no_control_regex.rs
leaysgur: String literals and Template literals may look similar but they have different syntax.

- https://tc39.es/ecma262/#sec-literals-string-literals
- https://tc39.es/ecma262/#sec-template-literal-lexical-components

Support only "static"(no substitution) template literals is feasible but I'm not sure we should.

Boshen: https://tc39.es/ecma262/#sec-regexpinitialize

4. Else, let F be ? [ToString](https://github.com/oxc-project/oxc/pull/6741#sec-tostring)(flags).

`ToString` is called, so this test case is valid. You can use this trait https://github.com/oxc-project/oxc/blob/main/crates/oxc_ecmascript/src/to_string.rs

leaysgur: I think it's OK to  use this `ToString` implementation and supports template literals in the linter side.

But then again, wouldn't that cause problems with `Span` position?

Boshen: 🤯 Maybe it's easiest to just allow this extra case.

leaysgur: My final proposals are...

---

1️⃣ Do not support template literals

- ESLint also appears not to support
- https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-control-regex.js
- [REPL](https://eslint.org/play/#eyJ0ZXh0IjoibmV3IFJlZ0V4cChgYWJjezIsMX1gLCBgT01HYClcbm5ldyBSZWdFeHAoXCJhYmN7MiwxfVwiLCBcIk9NR1wiKVxuIiwib3B0aW9ucyI6eyJydWxlcyI6eyJjb25zdHJ1Y3Rvci1zdXBlciI6WyJlcnJvciJdLCJmb3ItZGlyZWN0aW9uIjpbImVycm9yIl0sImdldHRlci1yZXR1cm4iOlsiZXJyb3IiXSwibm8tYXN5bmMtcHJvbWlzZS1leGVjdXRvciI6WyJlcnJvciJdLCJuby1jYXNlLWRlY2xhcmF0aW9ucyI6WyJlcnJvciJdLCJuby1jbGFzcy1hc3NpZ24iOlsiZXJyb3IiXSwibm8tY29tcGFyZS1uZWctemVybyI6WyJlcnJvciJdLCJuby1jb25kLWFzc2lnbiI6WyJlcnJvciJdLCJuby1jb25zdC1hc3NpZ24iOlsiZXJyb3IiXSwibm8tY29uc3RhbnQtYmluYXJ5LWV4cHJlc3Npb24iOlsiZXJyb3IiXSwibm8tY29uc3RhbnQtY29uZGl0aW9uIjpbImVycm9yIl0sIm5vLWNvbnRyb2wtcmVnZXgiOlsiZXJyb3IiXSwibm8tZGVidWdnZXIiOlsiZXJyb3IiXSwibm8tZGVsZXRlLXZhciI6WyJlcnJvciJdLCJuby1kdXBlLWFyZ3MiOlsiZXJyb3IiXSwibm8tZHVwZS1jbGFzcy1tZW1iZXJzIjpbImVycm9yIl0sIm5vLWR1cGUtZWxzZS1pZiI6WyJlcnJvciJdLCJuby1kdXBlLWtleXMiOlsiZXJyb3IiXSwibm8tZHVwbGljYXRlLWNhc2UiOlsiZXJyb3IiXSwibm8tZW1wdHkiOlsiZXJyb3IiXSwibm8tZW1wdHktY2hhcmFjdGVyLWNsYXNzIjpbImVycm9yIl0sIm5vLWVtcHR5LXBhdHRlcm4iOlsiZXJyb3IiXSwibm8tZW1wdHktc3RhdGljLWJsb2NrIjpbImVycm9yIl0sIm5vLWV4LWFzc2lnbiI6WyJlcnJvciJdLCJuby1leHRyYS1ib29sZWFuLWNhc3QiOlsiZXJyb3IiXSwibm8tZmFsbHRocm91Z2giOlsiZXJyb3IiXSwibm8tZnVuYy1hc3NpZ24iOlsiZXJyb3IiXSwibm8tZ2xvYmFsLWFzc2lnbiI6WyJlcnJvciJdLCJuby1pbXBvcnQtYXNzaWduIjpbImVycm9yIl0sIm5vLWludmFsaWQtcmVnZXhwIjpbImVycm9yIl0sIm5vLWlycmVndWxhci13aGl0ZXNwYWNlIjpbImVycm9yIl0sIm5vLWxvc3Mtb2YtcHJlY2lzaW9uIjpbImVycm9yIl0sIm5vLW1pc2xlYWRpbmctY2hhcmFjdGVyLWNsYXNzIjpbImVycm9yIl0sIm5vLW5ldy1uYXRpdmUtbm9uY29uc3RydWN0b3IiOlsiZXJyb3IiXSwibm8tbm9ub2N0YWwtZGVjaW1hbC1lc2NhcGUiOlsiZXJyb3IiXSwibm8tb2JqLWNhbGxzIjpbImVycm9yIl0sIm5vLW9jdGFsIjpbImVycm9yIl0sIm5vLXByb3RvdHlwZS1idWlsdGlucyI6WyJlcnJvciJdLCJuby1yZWRlY2xhcmUiOlsiZXJyb3IiXSwibm8tcmVnZXgtc3BhY2VzIjpbImVycm9yIl0sIm5vLXNlbGYtYXNzaWduIjpbImVycm9yIl0sIm5vLXNldHRlci1yZXR1cm4iOlsiZXJyb3IiXSwibm8tc2hhZG93LXJlc3RyaWN0ZWQtbmFtZXMiOlsiZXJyb3IiXSwibm8tc3BhcnNlLWFycmF5cyI6WyJlcnJvciJdLCJuby10aGlzLWJlZm9yZS1zdXBlciI6WyJlcnJvciJdLCJuby11bmRlZiI6WyJlcnJvciJdLCJuby11bmV4cGVjdGVkLW11bHRpbGluZSI6WyJlcnJvciJdLCJuby11bnJlYWNoYWJsZSI6WyJlcnJvciJdLCJuby11bnNhZmUtZmluYWxseSI6WyJlcnJvciJdLCJuby11bnNhZmUtbmVnYXRpb24iOlsiZXJyb3IiXSwibm8tdW5zYWZlLW9wdGlvbmFsLWNoYWluaW5nIjpbImVycm9yIl0sIm5vLXVudXNlZC1sYWJlbHMiOlsiZXJyb3IiXSwibm8tdW51c2VkLXByaXZhdGUtY2xhc3MtbWVtYmVycyI6WyJlcnJvciJdLCJuby11bnVzZWQtdmFycyI6WyJlcnJvciJdLCJuby11c2VsZXNzLWJhY2tyZWZlcmVuY2UiOlsiZXJyb3IiXSwibm8tdXNlbGVzcy1jYXRjaCI6WyJlcnJvciJdLCJuby11c2VsZXNzLWVzY2FwZSI6WyJlcnJvciJdLCJuby13aXRoIjpbImVycm9yIl0sInJlcXVpcmUteWllbGQiOlsiZXJyb3IiXSwidXNlLWlzbmFuIjpbImVycm9yIl0sInZhbGlkLXR5cGVvZiI6WyJlcnJvciJdfSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=)

---

2️⃣ Use `ToString` in the `oxc_linter` and accept shifted `Span` position if error

- May be applied to pattern text, not only flags text

---

3️⃣ Implement Template literal parser in `oxc_regular_expression`

- Not a complete parser
  - If substitution found, just emits error

---

I think 1️⃣ is reasonable enough, how about you?

Boshen: Ok, let's go for 1️⃣ Do not support template literals

Boshen:
### Merge activity

* **Oct 22, 1:33 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6748, opened 2024-10-21
fix(linter): correct false positive in `no-duplicates`

### LINKED ISSUE #6736: linter: `import/no-duplicates` false positive

0.10.1

### What command did you run?

oxlint -c oxlintrc.json --quiet

```json
{
  "plugins": ["import"],
  "rules": {
    "import/no-duplicates": "error"
  }
}
```

### What happened?

Looks like some regression after #6694 but that not catched, because we don't have some valid test cases for it.

I provide valid test case for reproduction:
```rs
        (
            r"import { ReactiveFormsModule } from '@angular/forms';
            import { RouterModule, Routes } from '@angular/router';",
            None,
        ),
```

shulaoda:
closes #6736 

There may be a better solution. If there is a good way to fix it, please feel free to close it. I will fix the clippy problem later because I went back to the dormitory to sleep and my computer was in the studio.


crates/oxc_linter/src/snapshots/no_duplicates.snap
```diff
@@ -540,14 +540,6 @@ source: crates/oxc_linter/src/tester.rs
    ╰────
   help: Merge these imports into a single import statement
 
-  ⚠ eslint-plugin-import(no-duplicates): Module './foo' is imported more than once in this file
-   ╭─[index.ts:1:38]
- 1 │ import {AValue, type x, BValue} from './foo'; import {type y} from './foo'
```
DonIsaac, line 545: What happened here?

camchenry: It looks like maybe we were reporting the same line twice? It looks to be the same message as the one below?

crates/oxc_linter/src/rules/import/no_duplicates.rs
```diff
@@ -131,7 +133,11 @@ impl Rule for NoDuplicates {
                             _ => 0,
                         }
                     };
-                    import_entries_maps.entry(key).or_default().push(requested_module);
+
+                    if flags[key as usize] {
+                        flags[key as usize] = false;
```
camchenry, line 138: Could we convert `key` to be `u8` instead to avoid sign loss here?

```diff
-ImportImportName::Name(_) => i8::from(!self.prefer_inline),
+ImportImportName::Name(_) => u8::from(!self.prefer_inline),
```

camchenry (approved):

camchenry (approved):

cheald:
I just upgraded to oxlint 0.10.2 and this seems to have broken things for me, unless I misunderstand the rule. I have a huge number of reported breaks in my project now, like:

```
app/javascript/src/pages/Accounts/useAccountContext.tsx:1:39: Module '@apollo/client' is imported more than once in this file [Error]
app/javascript/src/pages/Accounts/useAccountContext.tsx:8:8: Modules should not be imported multiple times in the same file [Error]
```

My full import declaration for that file is:

```tsx
import { QueryResult, useQuery } from "@apollo/client";
import {
  AccountRight,
  GetAccountDocument,
  GetAccountQuery,
  GetAccountQueryVariables,
  GrantRole,
} from "@app/graphql-codegen";
import { notEmpty } from "@app/utils";
import { Context } from "react";
import { useHistory } from "react-router-dom";
```

Downgrading to 0.10.0 (but not 0.10.1) causes the errors to disappear.

PR #6751, opened 2024-10-21
feat(linter): add `oxc/no-map-spread`

DonIsaac:
Adds a new performance rule that disallows using object and array spreads in `map` and `flatMap` to modify/merge arrays and objects.
I've currently added it to `nursery` so we can battle-harden it before making it generally available as a `perf` rule.

```ts
// needless O(n) spread on `x`, making this an O(n^2) operation
const x = arr.map(x => {
  return { ...x, foo: getFoo(x.id) }
})

// Object.assign is better here
const x = arr.map(x => {
  return Object.assign(x, { foo: getFoo(x.id) })
})
```

### DIFF
--- a/crates/oxc_linter/src/rules/oxc/no_map_spread.rs
+++ b/crates/oxc_linter/src/rules/oxc/no_map_spread.rs
@@ -1,0 +1,863 @@
+use oxc_semantic::{Reference, ReferenceId, ScopeId, SymbolId};
+use serde::{Deserialize, Serialize};
+use std::ops::Deref;
+
+use oxc_ast::{
+    ast::{
+        ArrayExpression, ArrayExpressionElement, CallExpression, Expression, ObjectExpression,
+        ObjectPropertyKind,
+    },
+    AstKind, Visit,
+};
+use oxc_diagnostics::{LabeledSpan, OxcDiagnostic};
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::{
+    ast_util::{is_method_call, leftmost_identifier_reference},
+    context::LintContext,
+    fixer::{RuleFix, RuleFixer},
+    rule::Rule,
+    utils::default_true,
+    AstNode,
+};
+
+fn no_map_spread_diagnostic(
+    map_call: Span,
+    spread: &Spread<'_, '_>,
+    returned_span: Option<Span>,
+) -> OxcDiagnostic {
+    let spans = spread.spread_spans();
+    assert!(!spans.is_empty());
+    let mut spread_labels = spread.spread_spans().into_iter();
+    let first_message = if spans.len() == 1 {
+        "It should be mutated in place"
+    } else {
+        "They should be mutated in place"
+    };
+    let first = spread_labels.next().unwrap().label(first_message);
+    let others = spread_labels.map(LabeledSpan::from);
+
+    let returned_label = returned_span
+        .filter(|span| !span.contains_inclusive(spread.span()))
+        .map(|span| span.label("Map returns the spread here"));
+
+    let diagnostic =
+        match spread {
+            // Obj
+            Spread::Object(_) => OxcDiagnostic::warn(
+                "Spreading to modify object properties in `map` calls is inefficient",
+            )
+            .with_labels([map_call.label("This map call spreads an object"), first])
+            .with_help("Consider using `Object.assign` instead"),
+            // Array
+            Spread::Array(_) => OxcDiagnostic::warn(
+                "Spreading to modify array elements in `map` calls is inefficient",
+            )
+            .with_labels([map_call.label("This map call spreads an array"), first])
+            .with_help("Consider using `Array.prototype.concat` or `Array.prototype.push` instead"),
+        };
+
+    diagnostic.and_labels(others).and_labels(returned_label)
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NoMapSpreadConfig {
+    /// Ignore mapped arrays that are re-read after the `map` call.
+    ///
+    /// Re-used arrays may rely on shallow copying behavior to avoid mutations.
+    /// In these cases, `Object.assign` is not really more performant than spreads.
+    ///
+    /// Default: `true`
+    #[serde(default = "default_true")]
+    ignore_rereads: bool,
+    /// Ignore maps on arrays passed as parameters to a function.
+    ///
+    /// This option is enabled by default to better avoid false positives. It
+    /// comes at the cost of potentially missing spreads that are inefficient.
+    /// We recommend turning this off in your `.oxlintrc.json` files.
+    ///
+    /// ### Example
+    ///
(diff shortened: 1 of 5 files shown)


DonIsaac:
* **#6751** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

camc314 (commented):
this looks great.

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,545 @@
...
+                    return this.#arr.map(x => ({ ...x, id: this.id }))
+                }
+            }",
+            None,
+        ),
+        (
+            // Spread does not occur within `map`
+            "let extras = { ...foo, ...bar }; let a = b.map(x => extras)",
+            None,
+        ),
```
camc314, line 525: i would suggest adding
```ts
function foo(arr) {
    return arr.map(x => ({ ...x }))
}
```

i don't know whether this would be reported currently, but it shuold not be as we don't know whether the fn arg could be used at the calcite

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,545 @@
...
+//         }
+//         _ => None,
+//     }
+// }
+fn get_map_callback<'a, 'b>(call_expr: &'b CallExpression<'a>) -> Option<&'b Expression<'a>> {
+    if !is_method_call(call_expr, None, Some(&["map", "flatMap"]), Some(1), Some(1)) {
+        return None;
+    }
+
+    let arg = call_expr.arguments.first()?.as_expression()?;
```
camc314, line 359: use without_parenthes/get_inner_expr ?

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,545 @@
...
+        let AstKind::CallExpression(call_expr) = node.kind() else {
+            return;
+        };
+        let Some(mapper) = get_map_callback(call_expr) else {
+            return;
+        };
+        // let Some((is_expression, body)) = get_map_callback(call_expr) else {
+        //     return;
+        // };
+
```
camc314, line 258:
```suggestion
```

camc314 (approved):
this looks good to me - there's some dead code that could be removed.

i also wonder whether it's worth either:
a. `SpreadInReturnVisitor::<'a, '_>::iter_spreads` converting this to follow the same pat as no_accumulating_spread to fine spreads
b. converting no_accumulating spread to use `SpreadInReturnVisitro`

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,578 @@
...
+    rule::Rule,
+    utils::default_true,
+    AstNode,
+};
+
+fn no_map_spread_diagnostic(map_call: Span, spread: &Spread<'_, '_>) -> OxcDiagnostic {
+    let spans = spread.spread_spans();
+    assert!(!spans.is_empty());
+    let mut spread_labels = spread.spread_spans().into_iter(); // .map(LabeledSpan::from);
+    let first_message = if spans.len() == 0 {
```
camchenry, line 28: This seems contradictory: we are asserting `spans` is not empty, but then checking if spans length is zero? Do we need to assert here? (If so, could we use `debug_assert!`?)

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,578 @@
...
+        .symbols()
+        // skip the reference within the spread itself
+        .get_resolved_reference_ids(symbol_id)
+        .iter()
+        .filter(|id| **id != reference_id)
+        .map(|id| ctx.semantic().symbols().get_reference(*id))
+        // we don't care if the symbol is overwritten
+        .filter(|r| r.is_read())
+        // Find where the symbol was declared
+        .map(|r| ctx.semantic().nodes().get_node(r.node_id())) //
```
camchenry, line 340:
```suggestion
        .map(|r| ctx.semantic().nodes().get_node(r.node_id()))
```

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
```diff
@@ -0,0 +1,578 @@
...
+                    return this.#arr.map(x => ({ ...x, id: this.id }))
+                }
+            }",
+            None,
+        ),
+        (
+            // Spread does not occur within `map`
+            "let extras = { ...foo, ...bar }; let a = b.map(x => extras)",
+            None,
+        ),
```
camchenry, line 552: Would probably be a good idea to add some "close-but-not-quite" cases like if `map` is a local function call (I think we already handle this, but always good to make sure)

```
let x = {}
function map(f) {}
map(x => ({ ...x }))
```

or map just returns an object:

```
a.map(x => ({ foo: x }))
```

crates/oxc_linter/src/snapshots/no_map_spread.snap
```diff
@@ -0,0 +1,144 @@
...
+   ·            │                                            ╰── They should be mutated in place
+   ·            ╰── This map call spreads an object
+   ╰────
+  help: Consider using `Object.assign` instead
+
+  × Expected `,` but found `}`
+   ╭─[no_map_spread.tsx:1:49]
+ 1 │ let a = b.map(({ x, y }) => ({ ...(cond ? x : y })
+   ·                                                 ┬
+   ·                                                 ╰── `,` expected
```
camchenry, line 116: Looks like this is missing a paren

mysteryven (commented):
LGTM

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
DonIsaac: That should be 1, good catch.

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
DonIsaac: Added cases for custom map definitions

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
camc314: sounds good to me

mysteryven (approved):

crates/oxc_linter/src/rules/oxc/no_map_spread.rs
DonIsaac: I've added `ignoreArgs`. It defaults to `true`.

DonIsaac:
### Merge activity

* **Oct 26, 1:40 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6686, opened 2024-10-19
feat(ast_tools): ability to generate text files

overlookmotel:
This was originally in #6404 but I pulled it out to reduce that PR's diff.

Add ability (currently unused) to output arbitrary text files from `oxc_ast_tools` generators.

PR #6683, opened 2024-10-19
fix(wasm): remove type defs for `ArrayExpressionElement` and `Elision`

overlookmotel:
Follow-on after #6404.

`ArrayExpressionElement` and `Elision` are not used in the TS types, because `ArrayExpression` has an override for the field it uses.

https://github.com/oxc-project/oxc/blob/002289b4b1b0d6fac080c4b12f4939c8f455272f/crates/oxc_ast/src/ast/js.rs#L293-L302

Prevent these TS type defs being emitted by introducing a new `#[estree(custom_ts_def)]` attr, to go with `#[estree(custom_serialize)]`.

PR #6755, opened 2024-10-21
feat(ast_tools): output typescript to a separate package

ottomated:
Part of #6347.

Moves typescript logic from derive_estree into a new ast_tools generator.

### DIFF
--- a/crates/oxc_ast/src/ast/types.d.ts
+++ b/crates/oxc_ast/src/ast/types.d.ts
@@ -1,0 +1,32 @@
+export interface FormalParameterRest extends Span {
+  type: 'RestElement';
+  argument: BindingPatternKind;
+  typeAnnotation: TSTypeAnnotation | null;
+  optional: boolean;
+}
+
+export type RegExpFlags = {
+  /** Global flag */
+  G: 1;
+  /** Ignore case flag */
+  I: 2;
+  /** Multiline flag */
+  M: 4;
+  /** DotAll flag */
+  S: 8;
+  /** Unicode flag */
+  U: 16;
+  /** Sticky flag */
+  Y: 32;
+  /** Indices flag */
+  D: 64;
+  /** Unicode sets flag */
+  V: 128;
+};
+
+export type JSXElementName =
+  | JSXIdentifier
+  | JSXNamespacedName
+  | JSXMemberExpression;
+
+export type JSXMemberExpressionObject = JSXIdentifier | JSXMemberExpression;
--- a/tasks/ast_tools/src/generators/typescript.rs
+++ b/tasks/ast_tools/src/generators/typescript.rs
@@ -1,0 +1,170 @@
+use convert_case::{Case, Casing};
+use itertools::Itertools;
+use std::{
+    io::Write,
+    process::{Command, Stdio},
+};
+
+use super::define_generator;
+use crate::{
+    codegen::LateCtx,
+    output,
+    schema::{
+        serialize::{enum_variant_name, get_type_tag},
+        EnumDef, GetIdent, StructDef, TypeDef, TypeName,
+    },
+    Generator, GeneratorOutput,
+};
+
+const CUSTOM_TYPESCRIPT: &str = include_str!("../../../../crates/oxc_ast/src/ast/types.d.ts");
+
+define_generator! {
+    pub struct TypescriptGenerator;
+}
+
+impl Generator for TypescriptGenerator {
+    fn generate(&mut self, ctx: &LateCtx) -> GeneratorOutput {
+        let file = file!().replace('\\', "/");
+        let mut content = format!(
+            "\
+        		// To edit this generated file you have to edit `{file}`\n\
+        		// Auto-generated code, DO NOT EDIT DIRECTLY!\n\n\
+						{CUSTOM_TYPESCRIPT}\n\
+						"
+        );
+
+        for def in ctx.schema() {
+            if !def.generates_derive("ESTree") {
+                continue;
+            }
+            let ts_type_def = match def {
+                TypeDef::Struct(it) => Some(typescript_struct(it)),
+                TypeDef::Enum(it) => typescript_enum(it),
+            };
+            let Some(ts_type_def) = ts_type_def else { continue };
+
+            content.push_str(&ts_type_def);
+            content.push_str("\n\n");
+        }
+        GeneratorOutput::Text {
+            path: output(crate::TYPESCRIPT_PACKAGE, "types.d.ts"),
+            content: format_typescript(&content),
+        }
+    }
+}
+
+// Untagged enums: "type Expression = BooleanLiteral | NullLiteral"
+// Tagged enums: "type PropertyKind = 'init' | 'get' | 'set'"
+fn typescript_enum(def: &EnumDef) -> Option<String> {
+    if def.markers.estree.custom_ts_def {
+        return None;
+    }
+
+    let union = if def.markers.estree.untagged {
+        def.all_variants().map(|var| type_to_string(var.fields[0].typ.name())).join(" | ")
+    } else {
+        def.all_variants().map(|var| format!("'{}'", enum_variant_name(var, def))).join(" | ")
+    };
+    let ident = def.ident();
+    Some(format!("export type {ident} = {union};"))
+}
+
+fn typescript_struct(def: &StructDef) -> String {
+    let ident = def.ident();
+    let mut fields = String::new();
+    let mut extends = vec![];
+
+    if let Some(type_tag) = get_type_tag(def) {
+        fields.push_str(&format!("\n\ttype: '{type_tag}';"));
+    }
+
+    for field in &def.fields {
+        if field.markers.derive_attributes.estree.skip {
+            continue;
+        }
+        let ty = match &field.markers.derive_attributes.tsify_type {
+            Some(ty) => ty.clone(),
+            None => type_to_string(field.typ.name()),
+        };
+
+        if field.markers.derive_attributes.estree.flatten {
+            extends.push(ty);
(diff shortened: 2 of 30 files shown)


ottomated:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6755?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6755** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6687** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6686** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6685** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6685?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6683** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Boshen:
👍 nice work

Boshen:
Our ts types came from typescript-eslint, I would model them as such https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/expression/ArrayExpression/spec.ts

but this is for another topic.

overlookmotel:
I have taken the liberty of combining the changes from #6687 into this PR, since neither PR stands alone, and it was confusing trying to review each in isolation. I've preserved the separate commits from both PRs.

I've added a commit to remove unused dependencies, which unblocks the "autofix" CI job.

@ottomated You will need to force pull the branch to your local machine, and then run `gt track` to get Graphite back in sync.

This is great, but there are a few problems:

#### Autofix

The "autofix" CI job automatically reformats all `.d.ts` files in the repo with `dprint`. But the "AST changes" CI job checks that generated files have not been altered manually. So it's freaking out because it sees the changes "autofix" made.

Longer term we should run `dprint` on the `.d.ts` in `ast_tools`, so it's formatted in standard style. But that can come later. Simplest way to solve the problem for now is to add the generated `.d.ts` file to autofix's ignore list:

https://github.com/oxc-project/oxc/blob/d2b8f505f136bbad0b3abffa3a8664a9439b305a/.github/workflows/autofix.yml#L4-L8

#### WASM types

This PR generates the `.d.ts` file, but doesn't link it into `wasm-bindgen`. So the types are now gone from the WASM package. This breaks that package, so this PR can't be merged as it stands (CI always needs to be green).

I suggest in this PR:

* Remove the new `oxc-types` package.
* Instead output the combined types in an `.rs` file as a single `const TS_APPEND_CONTENT` statement.
* Import that module into `oxc_ast/src/serialize.rs`.

That'll get this PR into a state in which it can be merged. Leave figuring out how to share the same type defs between `oxc-types` and WASM package to the next PR. *At that point*, re-introduce the `oxc-types` package.

#### Missing custom types

The generated `.d.ts` file is missing type defs for `FormalParameterRest`, `JSXElementName` etc which are currently defined in `const TS_APPEND_CONTENT` statements. I know you're aware of this, but it needs to be resolved before we can switch over to the generated `.d.ts`.

Sorry if this all feels like bureaucracy! It's annoying sometimes, but each individual PR (even in a stack) should not leave any crates / packages in a broken state. The advantage of stacked PRs is you can move fast (no need to wait for previous PR to get merged before starting on the next one), but the trade-off is that you need to consider the order you do things in, to keep CI green on every PR.

ottomated:
@overlookmotel thanks for the PR juggling, I'm well aware that this PR is not done and have marked it as draft for clarity.

ottomated:
@overlookmotel everything should be addressed now!

tasks/ast_tools/src/generators/typescript.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl Generator for TypescriptGenerator {
+    fn generate(&mut self, ctx: &LateCtx) -> GeneratorOutput {
+        let file = file!().replace('\\', "/");
+        let mut content = format!(
+            "\
+        		// To edit this generated file you have to edit `{file}`\n\
+        		// Auto-generated code, DO NOT EDIT DIRECTLY!\n\n\
+						{CUSTOM_TYPESCRIPT}\n\
+						"
```
overlookmotel, line 33: Nit: Weird indentation.

tasks/ast_tools/src/generators/typescript.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+        GeneratorOutput::Text {
+            path: output(crate::TYPESCRIPT_PACKAGE, "types.d.ts"),
+            content: format_typescript(&content),
+        }
+    }
+}
+
+// Untagged enums: "type Expression = BooleanLiteral | NullLiteral"
+// Tagged enums: "type PropertyKind = 'init' | 'get' | 'set'"
```
overlookmotel, line 57: Nit.

```suggestion
// Untagged enums: `type Expression = BooleanLiteral | NullLiteral;`
// Tagged enums: `type PropertyKind = 'init' | 'get' | 'set';`
```

tasks/ast_tools/src/generators/typescript.rs
```diff
@@ -0,0 +1,170 @@
...
+//         &allocator,
+//         PrettierOptions {
+//             semi: true,
+//             trailing_comma: TrailingComma::All,
+//             single_quote: true,
+//             ..PrettierOptions::default()
+//         },
+//     )
+//     .build(&ret.program)
+// }
```
overlookmotel, line 170: Delete this dead code. Dprint is probably what we want to use anyway, since that's the tool we use to format all other JS/TS code in the repo.

overlookmotel (commented):
All looks good. Very nice! My comments below are just nits.

The one remaining question is the name of the types package. We don't control the `@oxc` npm scope unfortunately, and Oxc's NPM packages are scattered around a bit.

We could use either a scope-less package name `oxc-types` or `@oxc-project/types`.

@Boshen?

tasks/ast_tools/src/generators/typescript.rs
ottomated: Only problem with dprint is it doesn't turn `(foo) | null` into `foo | null`.

overlookmotel:
@Boshen I've reviewed all the Rust code, but could you please take a look at the package publishing + github workflow stuff?

tasks/ast_tools/src/generators/typescript.rs
overlookmotel: I was going to ask. Can we just output `foo | null` to start with?

I imagine in some cases `foo` *needs* wrapping in brackets, but I don't know TS well enough to know in what circumstances. Could we use a quick-and-dirty approach of wrapping only if it contains `{` or something?

ottomated: Yeah I went through and it doesn't seem to cause any issues here. If we were using prettier it would just be a no-brainer to include it but since dprint can't handle it it makes sense to remove the parentheses.

ottomated: I accidentally committed this to #6796, still getting used to graphite, sorry.

tasks/ast_tools/src/generators/typescript.rs
overlookmotel: Ah I think it's just a tabs-vs-spaces thing. Probably some tabs have crept in, so it renders weirdly if you have your IDE set to render tabs as 4 spaces. Our convention is to use spaces instead of tabs.

tasks/ast_tools/src/generators/typescript.rs
overlookmotel: No problem. It's just a nit anyway!

Boshen:
> @Boshen I've reviewed all the Rust code, but could you please take a look at the package publishing + github workflow stuff?
> 
> Also, see above my question about package name for the types. `oxc-types` or `@oxc-project/types`?

I'll address these after this PR is merged.

Boshen:
Let me make a new release for `oxc` before merging this.

ottomated:
Could also be @oxc-ast/types or something along those lines

crates/oxc_ast/src/ast/types.d.ts
Boshen, line 1: This file seems out of place.

ottomated: Yeah this holds the manual implementation of any types that require it. @overlookmotel suggested putting it in the oxc_ast crate but it could go in ast_tools as well.

overlookmotel: I do think it should go in `oxc_ast`, so it's co-located with the rest of the AST serialization logic, rather than hidden away in `ast_tools`.

Perhaps it'd be better if we moved `serialize.rs` into a separate folder, and put this `types.d.ts` in that folder too. But I suggest we tackle that in a follow-up PR.

The intent is that this file is only a temporary measure anyway. Once `#[estree]` gets a bit more fully-featured and can generate types for these "special cases" itself, this file will be removed again.

overlookmotel (approved):
Boshen said he'd tackle the package release stuff after this is merged. Oxc 0.33.0 has been released. So I think this is good to merge now. Ditto #6796 and #6836.

@Boshen I'll leave it to you to merge, in case I've missing something.

Issue #6750, opened 2024-10-21
linter: oxc for vscode, document.createElement("iframe") - missing a sandbox attribute


lborgman:
### What version of Oxlint are you using?

0.10.1

### What command did you run?

oxc vscode extension

### What does your `.oxlint.json` config file look like?

As far as I know there is no .oxlint.json

### What happened?

I have this line in a .js file:
```javascript
const i = document.createElement("iframe");
```

I get this from oxc:

> An iframe element is missing a sandbox attribute
> help: Add a `sandbox` attribute to the `iframe` element.oxc[eslint-plugin-react(iframe-missing-sandbox)]> 
> (https://oxc.rs/docs/guide/usage/linter/rules/eslint-plugin-react/iframe-missing-sandbox)

I am not using React. And I have no idea about the plugin mentioned. (And the link above gives 404.)

PR #6765, opened 2024-10-22
fix(oxc_language_server): correct diagnostic link

shulaoda:
Related to #6750 

The current correction may not be very good. I suggest that we unify the linter rule link of the website.
For example, changing `https://oxc.rs/docs/guide/usage/linter/rules/jsx_a11y/alt-text.html` to `https://oxc.rs/docs/guide/usage/linter/rules/eslint-plugin-jsx-a11y/alt-text.html`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1629,6 +1629,7 @@ dependencies = [
 name = "oxc_language_server"
 version = "0.0.1"
 dependencies = [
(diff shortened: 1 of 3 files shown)


shulaoda:
@DonIsaac @camchenry cc

DonIsaac (approved):

PR #6777, opened 2024-10-22
chore(tasks/lint_rules): sync rules with ESLint v9

camchenry:
The previous version of the `react-perf` ESLint plugin was holding us back from updating this to ESLint V9. I've manually specified the ESLint version here, because we want to make sure we are running the latest major version of ESLint which changes a lot of things.

Also changed it so that rules which are not supported and also implemented do not get counted twice in the total count. For example, `eslint/no-with` is marked as not supported but it is still implemented. This was previously counted for both, which made it look like we had an additional rule implemented.

### DIFF
--- a/tasks/lint_rules/package.json
+++ b/tasks/lint_rules/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@next/eslint-plugin-next": "latest",
     "@typescript-eslint/eslint-plugin": "latest",
-    "eslint": "latest",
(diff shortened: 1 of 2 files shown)


camchenry:
* **#6777** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6777?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #6782, opened 2024-10-22
feat(linter): warn unmatched rule names

### LINKED ISSUE #6763: linter: warn unmatched rule names
Given 

```
  "rules": {
    "for-direction": "error",
    "no-async-promise-executor": "error",
    "no-caller": "error",
    "no-case-declarations": "error",
    "no-class-assign": "error",
    "no-compare-neg-zero": "error",
...
```

I don't know which rule is applied or not, we should warn the ones that are incorrect.

tapanprakasht:
Fixes https://github.com/oxc-project/oxc/issues/6763

### DIFF
--- a/crates/oxc_linter/src/config/rules.rs
+++ b/crates/oxc_linter/src/config/rules.rs
@@ -53,7 +53,7 @@ impl IntoIterator for OxlintRules {
 }
 
 impl OxlintRules {
-    #[allow(clippy::option_if_let_else)]
+    #[allow(clippy::option_if_let_else, clippy::print_stdout)]
     pub(crate) fn override_rules(&self, rules_for_override: &mut RuleSet, all_rules: &[RuleEnum]) {
         use itertools::Itertools;
         let mut rules_to_replace: Vec<RuleWithSeverity> = vec![];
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/config/rules.rs
```diff
@@ -121,6 +126,13 @@ impl OxlintRules {
         for rule in rules_to_replace {
             rules_for_override.replace(rule);
         }
+
+        if !rules_not_matched.is_empty() {
+            println!("The following rules do not match the currently supported rules:");
```
DonIsaac, line 131: Consider using `OxcDiagnostic::warn` and the reporter instead of directly printing to stdout.

DonIsaac (dismissed):

Boshen (changes requested):
Consider using OxcDiagnostic::warn and the reporter instead of directly printing to stdout.

crates/oxc_linter/src/config/rules.rs
tapanprakasht: What is the reporter here? any example from code base?

Boshen: Oh ...  surfacing the warning is actually harder than I thought.

Boshen (approved):
Let's merge and iterate.

PR #6805, opened 2024-10-23
feat(traverse): add `generate_binding` and `generate_binding_current_scope` APIs in context

Dunqing:
These two APIs are used to create a symbol with the provided symbol name, The difference from `generate_uid` is that we don't need to create a unique name for the symbol.

If you have a better method name for this, Feel free to edit this PR directly

### DIFF
--- a/crates/oxc_traverse/src/context/mod.rs
+++ b/crates/oxc_traverse/src/context/mod.rs
@@ -301,6 +301,36 @@ impl<'a> TraverseCtx<'a> {
         self.scoping.remove_scope_for_expression(scope_id, expr);
     }
 
+    /// Generate binding.
+    ///
+    /// Creates a symbol with the provided name and flags and adds it to the specified scope.
+    pub fn generate_binding(
+        &mut self,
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#6806** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6806?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6805** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6805?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel (approved):
I think it'd be preferable if `generate_binding` took an `Atom` instead of a `CompactStr`. If it's an existing var name (which it must be, otherwise we'd be using `generate_uid`), an `Atom` will already exist in the arena for this symbol name, so we'd be better off reusing it. I'll submit that change in a follow-on PR.

This business with juggling `CompactStr`s and `Atom`s is a real pain!

PR #6809, opened 2024-10-23
refactor(transformer): shorten code

overlookmotel:
Use `BoundIdentifier::create_binding_pattern`, rather than writing out the code to create a `BindingPattern` manually each time.

### DIFF
--- a/crates/oxc_transformer/src/common/var_declarations.rs
+++ b/crates/oxc_transformer/src/common/var_declarations.rs
@@ -15,7 +15,7 @@
 use std::cell::RefCell;
 
 use oxc_allocator::Vec;
-use oxc_ast::{ast::*, NONE};
+use oxc_ast::ast::*;
 use oxc_data_structures::stack::SparseStack;
 use oxc_span::SPAN;
 use oxc_traverse::{Ancestor, BoundIdentifier, Traverse, TraverseCtx};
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#6809** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing (approved):
I have to say the `BoundIdentifier` is so great!

Dunqing:
### Merge activity

* **Oct 23, 9:11 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
> I have to say the `BoundIdentifier` is so great!

I still think we can improve the API. I would prefer methods on the AST types e.g. `BindingIdentifier::from_binding(binding, ctx)`, `BindingPattern::from_binding(binding, ctx)`. But at least `BoundIdentifier` removes boilerplate code for things we do commonly.

overlookmotel:
> Yes, that would be better, but unfortunately, we can't get the `Atom` from `SymbolTable::get_name`, this may incur additional performance overhead.

Yeah we need to switch `Semantic` from storing names as `CompactStr`s to `Atom`s. That'd be a significant perf boost, and make lots of other things simpler. The blocker on that is that it makes `SymbolTable` need a lifetime `SymbolTable<'a>`, and apparently that causes problems for Rolldown (though I don't understand why).

PR #6819, opened 2024-10-23
fix(linter): false positive in `iframe-missing-sandbox`

### LINKED ISSUE #6750: linter: oxc for vscode, document.createElement("iframe") - missing a sandbox attribute

shulaoda:
closes #6750 

According to `eslint-plugin-react`, it seems that we should also search for `// @jsx pragma` in the source code.


crates/oxc_linter/src/utils/react.rs
```diff
@@ -12,16 +12,13 @@ use oxc_semantic::AstNode;
...
-        }
-        Expression::ComputedMemberExpression(member_expr) => {
-            member_expr.static_property_name().is_some_and(|name| name == "createElement")
-        }
         Expression::Identifier(ident) => ident.name == "createElement",
-        _ => false,
+        expr => expr.is_specific_member_access(
+            ctx.settings().react.pragma.as_ref().map_or("React", |pragma| pragma.as_str()),
+            "createElement",
+        ),
```
mysteryven, line 21: There are also some false positive:

```js
const createElement = document.createElement
createElement('iframe')

const React = myReact
React.createElement('iframe')
```


Can we port  [`isDestructuredFromPragmaImport`](https://github.com/jsx-eslint/eslint-plugin-react/blob/d1556a3d2f95f56cfb6e7c0a341af16742f64033/lib/util/isCreateElement.js#L28-L29) here? we can ignore the `require` part for this issue.

crates/oxc_linter/src/config/settings/react.rs
```diff
@@ -8,6 +8,8 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Deserialize, Default, Serialize, JsonSchema)]
 #[cfg_attr(test, derive(PartialEq))]
 pub struct ReactPluginSettings {
+    pub pragma: Option<CompactStr>,
```
mysteryven, line 11: The `pragma` are in `react.pragma`: https://github.com/jsx-eslint/eslint-plugin-react#configuration-legacy-eslintrc-

We can also set the default serde value to remove `Option`

apps/oxlint/fixtures/print_config/ban_rules/expect.json
```diff
@@ -23,6 +23,7 @@
       "rootDir": []
     },
     "react": {
+      "pragma": "React",
```
mysteryven, line 26: Why not keep same with ESLint config file? https://github.com/oxc-project/oxc/blob/3923e63b6100104a9985dec6fa27780471e88d78/crates/oxc_linter/src/config/oxlintrc.rs#L16

crates/oxc_linter/src/utils/react.rs
mysteryven: Or don't add `ctx.settings().react.pragma` for this issue:

https://github.com/oxc-project/oxc/blob/6f2253886ddb461063acc45e7647c2805adc4d65/crates/oxc_linter/src/utils/react.rs#L147

apps/oxlint/fixtures/print_config/ban_rules/expect.json
shulaoda: `form_components` and `link_components` don't seem to be in `react` but we designed them this way from the beginning.
```rust
// <https://github.com/jsx-eslint/eslint-plugin-react#configuration-legacy-eslintrc->
#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
#[cfg_attr(test, derive(PartialEq))]
pub struct ReactPluginSettings {
    #[serde(default = "default_pragma")]
    pub pragma: CompactStr,

    #[serde(default)]
    #[serde(rename = "formComponents")]
    form_components: Vec<CustomComponent>,

    #[serde(default)]
    #[serde(rename = "linkComponents")]
    link_components: Vec<CustomComponent>,
    // TODO: More properties should be added
}
```

shulaoda: https://github.com/jsx-eslint/eslint-plugin-react#configuration-legacy-eslintrc

mysteryven:
Close in favor of #6872

apps/oxlint/fixtures/print_config/ban_rules/expect.json
mysteryven: My fault, I missed this. Seems a simple way to fix this issue has merged.

PR #6822, opened 2024-10-23
feat(linter): Add rule eslint/no-dup-args and eslint/no-octal.

Orenbek:
Implemented_ last two eslint recommend rules. First time contributing to oxc project (and also my first rust project contribution).


Orenbek:
I don't know the difference between `run_on_symbol` and `run` function. What does symbol mean?

Boshen:
These two rules weren't implemented because they are strict mode errors. They should've been marked as completed 😢

Orenbek:
Why shouldn't we consider strict mode errors? Is there any documentation explaining this?

crates/oxc_linter/src/rules/eslint/no_dupe_args.rs
```diff
@@ -0,0 +1,106 @@
...
+                no_dupe_rags(&func.params.items, ctx);
+            }
+            AstKind::ArrowFunctionExpression(func) => no_dupe_rags(&func.params.items, ctx),
+            _ => return,
+        }
+    }
+}
+
+fn no_dupe_rags<'a>(items: &'a [FormalParameter<'a>], ctx: &LintContext<'a>) {
+    let mut param_name_set: HashSet<CompactStr> = HashSet::new();
```
DonIsaac, line 69: use `FxHashSet`

crates/oxc_linter/src/rules/eslint/no_octal.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// var num  = "071";
+    /// ```
+    NoOctal,
+    correctness,
+);
+
+impl Rule for NoOctal {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::NumericLiteral(numeric) = node.kind() {
+            let re = Regex::new(r"^0\d").unwrap();
```
DonIsaac, line 45: Use starts-with and manual char matching instead of a regular expression. These are _extremely_ costly.

DonIsaac (changes requested):

DonIsaac:
Thanks for the PR, and thanks for contributing to Oxc! Just a few perf-related issues, then we should be good to merge.

crates/oxc_linter/src/rules/eslint/no_dupe_args.rs
Orenbek: fixed

crates/oxc_linter/src/rules/eslint/no_octal.rs
Orenbek: I changed it to char matching.

DonIsaac:
@Orenbek what Boshen is saying is that duplicate args produces an error in `Semantic` when syntax checking is enabled, making the lint rule redundant. Syntax checks are enabled when linting.

Orenbek:
> @Orenbek what Boshen is saying is that duplicate args produces an error in `Semantic` when syntax checking is enabled, making the lint rule redundant. Syntax checks are enabled when linting.

oh, I see. Then, I remove this no-dup-args rule.

PR #6889, opened 2024-10-25
fix(transformer): only run typescript plugin for typescript source

### LINKED ISSUE #6865: transformer: turn off typescript and jsx plugin by option and sourcetype
typescript and jsx plugins should be 

* turned off by default when the source type is not ts / jsx.
* can be turned off by an option

Change proposal:

```rust
struct TransformerImpl<'a, 'ctx> {
    x0_typescript: TypeScript<'a, 'ctx>,
    x1_react: React<'a, 'ctx>,
```

```rust
pub struct TransformOptions {
    pub typescript: TypeScriptOptions,
    pub react: JsxOptions,
}
```

Make these two optional.

In `oxc_transformer/src/lib.rs`,

change code to

```rust
        self.x0_typescript.map(|plugin| plugin.enter_program(program, ctx));
```

Boshen:
closes #6865

TypeScript plugin changes import / export statements so it needs to be turned off for non-typescript files. This should also give a little performance boost for non-typescript files.

### DIFF
--- a/crates/oxc/src/napi/transform.rs
+++ b/crates/oxc/src/napi/transform.rs
@@ -96,7 +96,10 @@ impl From<TransformOptions> for oxc_transformer::TransformOptions {
     fn from(options: TransformOptions) -> Self {
         Self {
             cwd: options.cwd.map(PathBuf::from).unwrap_or_default(),
-            typescript: options.typescript.map(Into::into).unwrap_or_default(),
+            typescript: options
+                .typescript
+                .map(oxc_transformer::TypeScriptOptions::from)
+                .unwrap_or_default(),
             jsx: options.jsx.map(Into::into).unwrap_or_default(),
             es2015: options.es2015.map(Into::into).unwrap_or_default(),
             ..Self::default()
--- a/crates/oxc_transformer/src/lib.rs
+++ b/crates/oxc_transformer/src/lib.rs
@@ -92,7 +92,10 @@ impl<'a> Transformer<'a> {
         jsx::update_options_with_comments(&program.comments, &mut self.options, &self.ctx);
 
         let mut transformer = TransformerImpl {
-            x0_typescript: TypeScript::new(&self.options.typescript, &self.ctx),
+            x0_typescript: program
+                .source_type
+                .is_typescript()
+                .then(|| TypeScript::new(&self.options.typescript, &self.ctx)),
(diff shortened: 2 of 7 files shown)


Boshen:
* **#6889** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6889?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
> ⚡	`transformer[RadixUIAdoptionSection.jsx]`	157.6 µs	152.8 µs	+3.1%
> ❌	`transformer[checker.ts]`	18.7 ms	19.3 ms	-3.18%

I assume the regression for `checker.ts` is the cost of the extra "is TS enabled" checks. We may be able to make those checks cheaper by using an `is_ts_enabled` boolean flag instead. But, as you said in https://github.com/oxc-project/oxc/issues/6865#issuecomment-2436626895, we also want to avoid constructing the TS transformer if it's disabled - that setup cost will be significant for small JS files.

We should probably look at how transforms are constructed, and how options are stored, for all the transforms together once we've completed the rest of the ES6 transforms. We can probably gain some perf by refactoring it.

overlookmotel (approved):
> ⚡	`transformer[RadixUIAdoptionSection.jsx]`	157.6 µs	152.8 µs	+3.1%
> ❌	`transformer[checker.ts]`	18.7 ms	19.3 ms	-3.18%

I assume the regression for `checker.ts` is the cost of the extra "is TS enabled?" checks. We may be able to make those checks cheaper by using an `is_ts_enabled` boolean flag instead. But, as you said in https://github.com/oxc-project/oxc/issues/6865#issuecomment-2436626895, we also want to avoid constructing the TS transformer if it's disabled - that setup cost will be significant for small JS files.

We should probably look at how transforms are constructed, and how options are stored, for all the transforms together once we've completed the rest of the ES6 transforms. We can probably gain some perf by refactoring it, one way or another.

But for now I say merge it! It speeds up all the benchmarks except `checker.ts`, and fixes loads of tests.

Boshen:
### Merge activity

* **Oct 25, 9:59 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #6919, opened 2024-10-26
perf(parser): build comments vec directly in arena

overlookmotel:
Parser build `Vec<Comment>` directly in the arena, rather than building on heap and then copying contents into arena at the end of parsing.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -86,7 +86,7 @@ pub struct Lexer<'a> {
 
     context: LexerContext,
 
-    pub(crate) trivia_builder: TriviaBuilder,
+    pub(crate) trivia_builder: TriviaBuilder<'a>,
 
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#6919** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
Ha! I hate it when I label a PR "perf" and then it does the opposite!

Slight perf drop evident on lexer benchmarks. Looks like `add_comment` doesn't get inlined any more with the change to `oxc_allocator::Vec`.

Looking into the cause of that, I noticed that `allocator_api::vec::Vec` has its "vec is at capacity" branch as the hot branch in `push`, and inlines the code for that branch. This seems like a really bad choice - `std::vec::Vec` does the opposite, which makes much more sense.

So closing this for now, and we should address the root cause in `oxc_allocator::Vec`. If we fix that, it'll probably speed up other parts of the parser too.

PR #6965, opened 2024-10-27
fix(editor): Publish diagnostics on open for all run modes except disabled

nrayburn-tech:


nrayburn-tech:
Should the commit scope for this be `editor` or `language_server`?  I tried `language_server` with https://github.com/oxc-project/oxc/pull/6968, but it tagged the area as editor still.  So, not really sure which this should be.

PR #6967, opened 2024-10-27
fix(ecmascript): allow getting PropName for object methods

camchenry:
Working on some lint rules and noticed that PropName didn't seem to be working as intended. According to the spec, we should definitely allow get/set/methods to have prop names:

https://tc39.es/ecma262/#sec-static-semantics-propname

image

However, we still need to retain this logic for checking `__proto__` because it has special rules.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1603,7 +1603,9 @@ version = "0.34.0"
 dependencies = [
  "num-bigint",
  "num-traits",
+ "oxc_allocator",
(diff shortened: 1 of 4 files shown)


camchenry:
* **#6967** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Issue #6966, opened 2024-10-27
Editor: Support quick fixes to disable rule for line or file


nrayburn-tech:
Should add quick fixes to disable a rule for a single line or for the entire file.

Eslint screenshot
![image](https://github.com/user-attachments/assets/8cc869db-33df-403b-b05b-f1f0840dacd4)

nrayburn-tech:
#6968 is a rough implementation of this, but requires some cleanup if anybody familiar with rust wants to take that on.

PR #6968, opened 2024-10-27
feat(language_server): Add code actions to disable rules for the current line or entire file

nrayburn-tech:
This definitely needs some cleanup by somebody more familiar with Rust, but I think it contains most of the important pieces.  I'll leave a review with the different pieces that I know need cleanup.

Ref #6966.

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -303,7 +303,7 @@ impl LanguageServer for Backend {
                             changes: Some(std::collections::HashMap::from([(
                                 uri.clone(),
                                 vec![TextEdit {
-                                    range: fixed_content.range.clone(),
+                                    range: fixed_content.range,
                                     new_text: fixed_content.code.clone(),
                                 }],
                             )])),
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -2,6 +2,7 @@ mod linter;
 
 use std::{fmt::Debug, path::PathBuf, str::FromStr};
 
+use crate::linter::{DiagnosticReport, ServerLinter};
 use dashmap::DashMap;
 use futures::future::join_all;
 use globset::Glob;
@@ -10,6 +11,7 @@ use log::{debug, error, info};
 use oxc_linter::{FixKind, LinterBuilder, Oxlintrc};
 use serde::{Deserialize, Serialize};
 use tokio::sync::{Mutex, OnceCell, RwLock, SetError};
+use tower_lsp::lsp_types::{NumberOrString, Position, Range};
 use tower_lsp::{
     jsonrpc::{Error, ErrorCode, Result},
     lsp_types::{
@@ -24,8 +26,6 @@ use tower_lsp::{
     },
     Client, LanguageServer, LspService, Server,
 };
-
-use crate::linter::{DiagnosticReport, ServerLinter};
 
 struct Backend {
     client: Client,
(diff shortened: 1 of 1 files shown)


crates/oxc_language_server/src/main.rs
```diff
@@ -240,34 +240,92 @@ impl LanguageServer for Backend {
                 .iter()
                 .find(|r| r.diagnostic.range == params.range && r.fixed_content.is_some())
             {
-                let title =
-                    report.diagnostic.message.split(':').next().map_or_else(
-                        || "Fix this problem".into(),
-                        |s| format!("Fix this {s} problem"),
-                    );
-
+                let rule_name: Option<String> = None;
```
nrayburn-tech, line 243: Needs to get the exact rule name.  Is used as part of the disable comment like `eslint-disable-next-line no-debugger`.

crates/oxc_language_server/src/main.rs
```diff
@@ -240,34 +240,92 @@ impl LanguageServer for Backend {
...
+                            #[expect(clippy::disallowed_types)]
+                            changes: Some(std::collections::HashMap::from([(
+                                uri.clone(),
+                                vec![TextEdit {
+                                    range: Range {
+                                        start: fixed_content.range.start,
+                                        end: fixed_content.range.start,
+                                    },
+                                    new_text: rule_name.clone().map_or_else(
+                                        || "// eslint-disable-next-line\n".into(),
```
nrayburn-tech, line 287: Should this be `eslint-disable` or is there an oxlint equivalent available?

nrayburn-tech, line 287: Probably needs a system dependent line separator instead of `\n`.

crates/oxc_language_server/src/main.rs
```diff
@@ -240,34 +240,92 @@ impl LanguageServer for Backend {
...
+                            |s| format!("Disable {s} for this line"),
+                        ),
+                        kind: Some(CodeActionKind::QUICKFIX),
+                        is_preferred: Some(false),
+                        edit: Some(WorkspaceEdit {
+                            #[expect(clippy::disallowed_types)]
+                            changes: Some(std::collections::HashMap::from([(
+                                uri.clone(),
+                                vec![TextEdit {
+                                    range: Range {
```
nrayburn-tech, line 282: If there is already an existing `eslint-disable-next-line` comment, then it should probably be updated instead of adding a completely new line.  Otherwise you end up with something that doesn't work.

This is what would be created today.
```
// eslint-disable-next-line some-rule
// eslint-disable-next-line no-var
var something = 'abc';
```

I believe this is what we need to create (untested).
```
// eslint-disable-next-line some-rule no-var
var something = 'abc';
```

crates/oxc_language_server/src/main.rs
camchenry: We do support "oxlint-disable", but I'd be partial to keeping it as eslint-disable just for the sake of compatibility personally

nrayburn-tech:
Made some updates to this and added some TODOs in the code.  I'm not sure the TODOs are something that can easily be resolved in this PR.  I think they require larger changes.

So, this could probably be merged in as is (somebody may want to cleanup the code, but I think it hits the functional parts) or leave this as a draft and make some other PRs so that the TODOs can be resolved and then merge this in.

Boshen:
@Sysix can you help and get this merged.

Sysix (commented):
there is no quick fix option when no oxlint auto fix is available:
grafik

I expect as an user, to disable the diagnostic even when no auto fix is available

crates/oxc_language_server/src/main.rs
Sysix: In PHPStorm under Windows I can change the line separator. I even prefer setting it globally to "\n". 
But at the end there can be an `.editorconfig` or some other project settings which will force one style.

https://www.jetbrains.com/help/phpstorm/configuring-line-endings-and-line-separators.html

=> I would keep it how it is

nrayburn-tech: Keep it how it is, meaning we need to detect what the file has and use that line separator? Any suggestions on how to do that here, I don’t think we currently have access to the file contents without reading from the filesystem again.

Boshen:
Do we have a consensus on this PR?

nrayburn-tech:
I think so. I’ve got one bug to fix in it and I need to get it rebased.

Hopefully I’ll get to it by end of day tomorrow.

Boshen:
> I think so. I’ve got one bug to fix in it and I need to get it rebased.
> 
> Hopefully I’ll get to it by end of day tomorrow.

Thank you.

nrayburn-tech:
Should be ready now, including the bug fix for the issue mentioned by @Sysix.

Sysix:
Thanks you @nrayburn-tech, will look tomorrow on it :)

Sysix (approved):
tested it and works great <3

PR #6985, opened 2024-10-28
fix(parser): do not double-parse modifiers of ts index signature in class

wan9chi:
Modifiers of ts index signatures in classes are parsed twice. The first one is ignored and the second one is always empty, causing the incorrect `readonly` and `span` in ast. https://oxc-project.github.io/oxc/playground/?code=3YCAAICvgICAgICAgICxG4jI43W9aqTWr3Wzy1UcIvgVm0uDVLP%2FHYReSN%2FJ0Mhppm0XIo%2F7DW7cd39qCwCA

### DIFF
--- a/crates/oxc_parser/src/lib.rs
+++ b/crates/oxc_parser/src/lib.rs
@@ -696,6 +696,23 @@ mod test {
         assert_eq!(ret.errors.first().unwrap().to_string(), "Source length exceeds 4 GiB limit");
     }
 
+    #[test]
+    fn class_index_signature_modifiers() {
+        use oxc_ast::ast::*;
+        let allocator = Allocator::default();
+        let source_type = SourceType::ts();
+        let source = "class Foo { readonly [a: string]: string }";
+        let ret = Parser::new(&allocator, source, source_type).parse();
+        let Statement::ClassDeclaration(class_decl) = &ret.program.body[0] else {
+            panic!("expected Statement::ClassDeclaration");
+        };
+        let ClassElement::TSIndexSignature(index_sig) = &class_decl.body.body[0] else {
+            panic!("expected ClassElement::TSIndexSignature");
+        };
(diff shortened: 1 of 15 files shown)


crates/oxc_parser/src/lib.rs
```diff
@@ -696,6 +696,23 @@ mod test {
         assert_eq!(ret.errors.first().unwrap().to_string(), "Source length exceeds 4 GiB limit");
     }
 
+    #[test]
```
Boshen, line 699: This file is for integration tests; parser tests should go into https://github.com/oxc-project/oxc/tree/main/tasks/coverage/misc/pass

Boshen:
I'm curious what project you are working on with `oxc` 😄

crates/oxc_parser/src/lib.rs
branchseer: I think this fix can only be verified by examining the ast. It can't be done in there, can it?

branchseer:
> I'm curious what project you are working on with `oxc` 😄

Just experimenting something 😄. Don't wanna spoil it now because I'm not sure if it's even viable. But if it is, I will surely let you guys know.

tasks/coverage/snapshots/parser_typescript.snap
```diff
@@ -15102,38 +15148,6 @@ Expect to Parse: tasks/coverage/typescript/tests/cases/conformance/salsa/private
  9 │ }
    ╰────
 
-  × TS(1071): 'static' modifier cannot appear on an index signature.
-    ╭─[typescript/tests/cases/conformance/classes/staticIndexSignature/staticIndexSignature4.ts:12:5]
- 11 │ interface IB {
- 12 │     static [s: string]: number;
```
DonIsaac, line 15108: Looks like you forgot to check if `static` is being used within an interface

PR #6986, opened 2024-10-28
[POC] perf(ast): use pointer instead of slice for identifier names

DonIsaac:
> This is a very rough proof-of-concept

Many different AST nodes store both a span and an `Atom<'a>`, which has redundant information. `Atom<'a>` is a fat pointer storing a length, but we already have that information in the form of `Span::offset`.

This POC shows how we can shave a byte off all AST nodes that use both a `Span` and an `Atom<'a>` (i.e. all identifiers) by using a thin pointer over an `Atom<'a>`, then using `Span::offset` to get an `&str` slice as needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1489,6 +1489,7 @@ dependencies = [
  "cow-utils",
  "itertools",
  "lazy_static",
+ "oxc-miette",
  "prettyplease",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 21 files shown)


DonIsaac:
* **#6986** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/6986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
This would work great for an AST that's fresh out of the parser. However, once you get into transform/minify, I don't think it works unfortunately. Transformer/minifier may (1) alter identifiers from the original or (2) insert new identifiers which have a dummy `Span { start: 0, end: 0 }`.

However, your idea made me realize that we *could* get rid of the `Atom` entirely if we created `SymbolId`s in the parser, and stocked the `SymbolTable` at that time too. The identifier's name is already stored in `SymbolTable`, so storing it again in the AST is redundant.

Boshen:
Unfortunately we are not building a runtime. We sometimes need duplicated information for ease of use.

PR #6995, opened 2024-10-29
feat(linter): implement `jsx-no-script-url`

radu2147:
https://github.com/oxc-project/oxc/issues/1022


crates/oxc_linter/src/rules/react/jsx_no_script_url.rs
```diff
@@ -0,0 +1,266 @@
...
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{CompactStr, GetSpan, Span};
+use regex::Regex;
+use rustc_hash::FxHashMap;
+use serde_json::Value;
+
+fn jsx_no_script_url_diagnostic(span: Span) -> OxcDiagnostic {
+    // See <https://oxc.rs/docs/contribute/linter/adding-rules.html#diagnostics> for details
+    OxcDiagnostic::warn("A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML, try using dangerouslySetInnerHTML instead.")
```
DonIsaac, line 14: Move all sentences but the first to a help message

crates/oxc_linter/src/rules/react/jsx_no_script_url.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+declare_oxc_lint!(
+    /// ### What it does
+    ///
+    /// Disallow usage of `javascript:` URLs
+    ///
+    /// ### Why is this bad?
+    ///
+    /// In React 16.9 any URLs starting with javascript: scheme log a warning. React considers the pattern as a dangerous attack surface, see details. In a future major release, React will throw an error if it encounters a javascript: URL.
```
DonIsaac, line 44: What details?

radu2147: Copy paste issue, will change

crates/oxc_linter/src/rules/react/jsx_no_script_url.rs
```diff
@@ -0,0 +1,272 @@
...
+                return;
+            };
+            if let Some(link_props) = self.components.get(component_name.as_str()) {
+                for jsx_attribute in &element.attributes {
+                    if let JSXAttributeItem::Attribute(attr) = jsx_attribute {
+                        let Some(prop_value) = &attr.value else {
+                            return;
+                        };
+                        if prop_value.as_string_literal().is_some_and(|val| {
+                            let re = Regex::new(IS_JAVA_SCRIPT_PROTOCOL).unwrap();
```
IWANABETHATGUY, line 99: you should use `lazy_static` to wrap it to make sure it is only initialized once. https://github.com/oxc-project/oxc/blob/a786dc2449c46112cd3a5d3f99ea6087e39c3075/crates/oxc_linter/src/rules/eslint/no_nonoctal_decimal_escape.rs#L87-L90

IWANABETHATGUY: here is the reason, https://github.com/rust-lang/regex?tab=readme-ov-file#usage-avoid-compiling-the-same-regex-in-a-loop, I recommended using `lazy_static` since it is widely used in `linter` crate

crates/oxc_linter/src/rules/react/jsx_no_script_url.rs
```diff
@@ -0,0 +1,274 @@
...
+        ("<a href />", None, None),
+        (
+            r#"<Foo href="javascript:"></Foo>"#,
+            None,
+            Some(
+                serde_json::json!({ "settings": {"react": {"linkComponents": [{ "name": "Foo", "linkAttribute": ["to", "href"] }]} } }),
+            ),
+        ),
+        (
+            r#"<Foo other="javascript:"></Foo>"#,
```
radu2147, line 187: FYI, I added an extra test to `pass`,  as it checks for the link attributes not to have the `javascript:` value.(It seems that the original implementation of the rule does that, it just doesn't have a test case for it). I can remove it though if deemed as not necessary.

Issue #7009, opened 2024-10-29
linter: using typescript alias rules with its native eslint rule, deactives both


Sysix:
### What version of Oxlint are you using?

0.10.3

### What command did you run?

oxlint -c .oxlint.json

### What does your `.oxlint.json` config file look like?

```
{
  "categories": {
    "correctness": "off"
  },
  "rules": {
    "no-magic-numbers": "warn",
    "@typescript-eslint/no-magic-numbers": "warn"
  }
}
```

and 

```

{
  "categories": {
    "correctness": "off"
  },
  "rules": {
    "no-magic-numbers": "warn"
  }
}
```

### What happened?

First config output:

```
Finished in 19ms on 21 files with 0 rules using 24 threads.
Found 0 warnings and 0 errors.
```

Second config output:

```

Finished in 15ms on 21 files with 1 rules using 24 threads.
Found 53 warnings and 0 errors.
```

PR #7027, opened 2024-10-30
feat(linter)!: report unmatched rules with error exit code

### LINKED ISSUE #6988: linter: bail unmatched rules
https://github.com/oxc-project/oxc/blob/1691cab507cd22e0e32f80d6fcb2911df82669f1/crates/oxc_linter/src/config/rules.rs#L129-L142

Should we bail unmatched rules?

### LINKED ISSUE #7025: linter: rule incorrectly reported as unmatched
I wonder why I got this with config: `"no-default-export": "allow"`,

> oxlint Stderr:   ! The following rules do not match the currently supported rules:
>   | no-default-export

when `oxlint --rules` have this rule in the list under Restriction category (on `0.10.3`)

originally reported by `@ariu` on discord

camchenry:
- closes https://github.com/oxc-project/oxc/issues/6988

we now return an error exit code when there are unmatched rules. previously, we would print an error to stderr and continue running. however, this masked errors in some tests that actually had unmatched rules in them. these test cases now trigger a panic (in tests only, not at runtime), and help ensure that we are reporting an error message to the user for unknown rules, which we did not have any tests cases for before.

- fixes https://github.com/oxc-project/oxc/issues/7025

this also fixes https://github.com/oxc-project/oxc/issues/7025, where we were reporting rules as unmatched simply because they had been disabled prior to being configured. similar to https://github.com/oxc-project/oxc/issues/7009.

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -1,10 +1,10 @@
 use std::{env, io::BufWriter, time::Instant};
 
 use ignore::gitignore::Gitignore;
-use oxc_diagnostics::{DiagnosticService, GraphicalReportHandler};
+use oxc_diagnostics::{DiagnosticService, Error, GraphicalReportHandler, OxcDiagnostic};
 use oxc_linter::{
     loader::LINT_PARTIAL_LOADER_EXT, AllowWarnDeny, InvalidFilterKind, LintFilter, LintService,
-    LintServiceOptions, Linter, LinterBuilder, Oxlintrc,
+    LintServiceOptions, Linter, LinterBuilder, LinterBuilderError, Oxlintrc,
 };
 use oxc_span::VALID_EXTENSIONS;
 
@@ -119,9 +119,24 @@ impl Runner for LintRunner {
 
         let oxlintrc_for_print =
             if misc_options.print_config { Some(oxlintrc.clone()) } else { None };
-        let builder = LinterBuilder::from_oxlintrc(false, oxlintrc)
-            .with_filters(filter)
-            .with_fix(fix_options.fix_kind());
+        let builder = LinterBuilder::from_oxlintrc(false, oxlintrc);
+        // Gracefully report any linter builder errors as CLI errors
+        let builder = match builder {
+            Ok(builder) => builder,
+            Err(err) => match err {
+                LinterBuilderError::UnknownRules { rules } => {
+                    let rules = rules.iter().map(|r| r.full_name()).collect::<Vec<_>>().join("\n");
+                    let error = Error::from(
+                        OxcDiagnostic::warn(format!(
+                        "The following rules do not match the currently supported rules:\n{rules}"
+                    ))
+                        .with_help("Check that the plugin that contains this rule is enabled."),
+                    );
+                    return CliRunResult::LintError { error: format!("{error:?}") };
+                }
+            },
+        };
+        let builder = builder.with_filters(filter).with_fix(fix_options.fix_kind());
 
         if let Some(basic_config_file) = oxlintrc_for_print {
             return CliRunResult::PrintConfigResult {
@@ -244,6 +259,9 @@ mod test {
         let options = lint_command().run_inner(new_args.as_slice()).unwrap();
         match LintRunner::new(options).run() {
             CliRunResult::LintResult(lint_result) => lint_result,
+            CliRunResult::LintError { error } => {
+                panic!("{error}")
+            }
             other => panic!("{other:?}"),
         }
     }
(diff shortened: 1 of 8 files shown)


camchenry:
* **#7027** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7027?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/oxc_linter/src/builder.rs
```diff
@@ -21,6 +21,13 @@ pub struct LinterBuilder {
     cache: RulesCache,
 }
 
+/// An error that can occur while building a [`Linter`] from an [`Oxlintrc`].
+#[derive(Debug, Clone)]
```
DonIsaac, line 25: Should implement std::error::Error for easy conversion to miette errors

DonIsaac (commented):
I need to review this more thoroughly. I'll keep looking when I have time (tonight or tomorrow)

crates/oxc_language_server/src/main.rs
```diff
@@ -359,6 +359,8 @@ impl Backend {
                     Oxlintrc::from_file(&config_path)
                         .expect("should have initialized linter with new options"),
                 )
+                // FIXME: Handle this error more gracefully and report it properly
```
Boshen, line 362: Let's not crash the IDE 😅 shall we silence the error before the fix?

apps/oxlint/src/lint.rs
```diff
@@ -483,18 +501,20 @@ mod test {
         assert_eq!(result.number_of_errors, 0);
     }
 
+    // Previously, this test would pass and the unmatched rule would be ignored, but now we report that
+    // there was unmatched rule, because the typescript plugin has been disabled and we are trying to configure it.
     #[test]
+    #[should_panic(
```
Boshen, line 507: Instead of testing for a panic that's coming out of the `test` function, shall we add another `test_failure` function instead.

apps/oxlint/src/lint.rs
```diff
@@ -119,9 +119,24 @@ impl Runner for LintRunner {
...
+        let builder = match builder {
+            Ok(builder) => builder,
+            Err(err) => match err {
+                LinterBuilderError::UnknownRules { rules } => {
+                    let rules = rules.iter().map(|r| r.full_name()).collect::<Vec<_>>().join("\n");
+                    let error = Error::from(
+                        OxcDiagnostic::warn(format!(
+                        "The following rules do not match the currently supported rules:\n{rules}"
+                    ))
+                        .with_help("Check that the plugin that contains this rule is enabled."),
```
Boshen, line 133: This help message is somewhat confusing - it's either:

* we don't support the rule
* the rule is contained in a disabled plugin

camchenry: maybe we should just skip this for now - I foresee in the future us being able to provide more helpful text here like finding which plugin contains this rule. for now though, we can just keep it simple

Boshen: 👍

crates/oxc_language_server/src/main.rs
camchenry: I think it's probably okay for now, since we already have two other unwrap/expect calls in this function? I need more expertise on the editor side though, I feel like the "right" way to handle this is to create an error popup, but maybe that happens automatically?

Boshen: Yeah the correct way is to show a popup, I don't know what happens when it panics.

camchenry: I can test it out and see what happens

nrayburn-tech: From memory, I believe this is the behavior for VS Code and IntelliJ.
The language server will crash. If it crashes multiple times, a pop up will display indicating it has been disabled. You can restart your IDE to reenable it. You can check your logs to see what might have happened. (Oxc channel in VS Code or idea.log for IntelliJ)

If you want a popup, that needs to be communicated by the language server with a message to the IDE. I don’t know if a standard message exists, or if this would be a custom one.

For now, if you just catch the error and log it to stderr then I can fix the editor support after.

Boshen:
### Merge activity

* **Oct 31, 11:26 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Boshen:
https://github.com/oxc-project/oxlint-ecosystem-ci/actions/runs/11623823150/job/32371532541

We need to fix preact with this change 😅

PR #7040, opened 2024-10-31
ci(transformer): enable unfinished plugins in benchmark

Dunqing:
Even the plugins are unfinished, we still want to enable all of them to track the performance changes during the development.

### DIFF
--- a/crates/oxc_transformer/src/options/env.rs
+++ b/crates/oxc_transformer/src/options/env.rs
@@ -1,8 +1,14 @@
 use serde::Deserialize;
 
 use crate::{
-    es2015::ES2015Options, es2016::ES2016Options, es2017::ES2017Options, es2018::ES2018Options,
-    es2019::ES2019Options, es2020::ES2020Options, es2021::ES2021Options, es2022::ES2022Options,
+    es2015::{ArrowFunctionsOptions, ES2015Options},
+    es2016::ES2016Options,
+    es2017::ES2017Options,
+    es2018::{ES2018Options, ObjectRestSpreadOptions},
+    es2019::ES2019Options,
+    es2020::ES2020Options,
+    es2021::ES2021Options,
+    es2022::{ClassPropertiesOptions, ES2022Options},
     regexp::RegExpOptions,
 };
 
@@ -32,7 +38,10 @@ pub struct EnvOptions {
 
 impl EnvOptions {
     /// Explicitly enable all plugins that are ready, mainly for testing purposes.
-    pub fn enable_all() -> Self {
+    ///
+    /// NOTE: for internal use only
+    #[doc(hidden)]
+    pub fn enable_all(include_unfinished_plugins: bool) -> Self {
         Self {
             regexp: RegExpOptions {
                 sticky_flag: true,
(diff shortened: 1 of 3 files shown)


Dunqing:
* **#7040** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

overlookmotel:
I agree that it's good to measure all our transforms, even the unfinished ones, in benchmarks.

However, `enable_all` is a public API (like Babel's "force all transforms" option), so in my opinion it seems a bit odd to offer an "include unfinished plugins" option, as that's a bit like offering a "break all your code" option!

I think it'd be preferable to leave `enable_all` as it is, and manually enable the unfinished transforms in [tasks/benchmark/benches/transformer.rs](https://github.com/oxc-project/oxc/pull/7040/files#diff-98d2c590e9cca0d9ce223338036ae6cf808a0f3283132995c9e1e986e67165f0).

Or, if you want to locate this logic in `oxc_transformer` without exposing it to the user:

```rs
impl EnvOptions {
    pub fn enable_all() -> Self {
        Self::enable_all_impl(false)
    }

    /// For internal use only
    #[doc(hidden)]
    pub fn enable_all_impl(include_unfinished_plugins: bool) -> Self {
        // Same as impl in this PR
    }
}
```

Benchmarks can use `enable_all_impl`, but we don't "officially" expose ability to run our shonky half-finished transforms to the user.

Dunqing:
> However, `enable_all` is a public API (like Babel's "force all transforms" option), so in my opinion it seems a bit odd to offer an "include unfinished plugins" option, as that's a bit like offering a "break all your code" option!
> 
> I think it'd be preferable to leave `enable_all` as it is, and manually enable the unfinished transforms in [tasks/benchmark/benches/transformer.rs](https://github.com/oxc-project/oxc/pull/7040/files#diff-98d2c590e9cca0d9ce223338036ae6cf808a0f3283132995c9e1e986e67165f0).

We tend to forget to enable unfinished plugins in the benchmark easily. So I am not a fan of this way

> 
> Or, if you want to locate this logic in `oxc_transformer` without exposing it to the user:


> ```rust
> impl EnvOptions {
>     pub fn enable_all() -> Self {
>         Self::enable_all_impl(false)
>     }
> 
>     /// For internal use only
>     #[doc(hidden)]
>     pub fn enable_all_impl(include_unfinished_plugins: bool) -> Self {
>         // Same as impl in this PR
>     }
> }
> ```
> 
> Benchmarks can use `enable_all_impl`, but we don't "officially" expose ability to run our shonky half-finished transforms to the user.


IMO, `enable_all` is also an internal-only API. So we can add  
```
/// For internal use only
#[doc(hidden)]
```
to `enable_all` method. Tell me what's your thought about this

Boshen:
> @Boshen Is `enable_all` meant to be a public method or not? I thought maybe it is (like Babel's "force all transforms" option).

This is not a public API.

overlookmotel:
OK great. @Dunqing I had misunderstood. I had thought `enable_all` was a public API.

Since it's not, this PR with this added to `enable_all` (as you suggested), seems good to me:

```rs
/// For internal use only
#[doc(hidden)]
```

Sorry I wasted time!

Boshen:
### Merge activity

* **Nov 2, 2:17 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7067, opened 2024-11-01
fix(linter): check is_reference_to_global_variable in `no-array-constructor` rule

azihsoyn:
Hello,

I am currently reading through the code as I work on contributing by creating some ESLint rules.
Along the way, I noticed an incompatibility in certain cases within the `no-array-constructor` rule.

In this PR, I added test cases and modified the code to include a check for is_reference_to_global_variable.

The relevant ESLint behavior can be verified in the following playground.

https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tYXJyYXktY29uc3RydWN0b3I6IFwiZXJyb3JcIiovXG5cbi8vIGlmIGNvbW1lbnQgb3V0IGJlbG93LCBuZXh0IGxpbmUgYXMgZXJyb3JcbnZhciBBcnJheTsgbmV3IEFycmF5KCk7XG5cbm5ldyBBcnJheSgpOyIsIm9wdGlvbnMiOnsicnVsZXMiOnt9LCJsYW5ndWFnZU9wdGlvbnMiOnsic291cmNlVHlwZSI6Im1vZHVsZSIsInBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX19fX0=

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
@@ -2,6 +2,7 @@ use oxc_ast::ast::Expression;
 use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
+use oxc_semantic::IsGlobalReference;
 use oxc_span::Span;
 
 use crate::{context::LintContext, rule::Rule, AstNode};
@@ -68,7 +69,11 @@ impl Rule for NoArrayConstructor {
             }
         };
 
-        if callee.is_global_reference_name("Array", ctx.symbols())
+        let Expression::Identifier(ident) = &callee else {
+            return;
+        };
+
+        if ident.is_global_reference_name("Array", ctx.symbols())
             && arguments.len() != 1
             && type_parameters.is_none()
             && !optional
--- a/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
@@ -68,12 +68,7 @@ impl Rule for NoArrayConstructor {
             }
         };
 
-        let Expression::Identifier(ident) = &callee else {
-            return;
-        };
-
-        if ctx.semantic().is_reference_to_global_variable(ident)
-            && callee.is_specific_id("Array")
+        if callee.is_global_reference_name("Array", ctx.symbols())
             && arguments.len() != 1
             && type_parameters.is_none()
             && !optional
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
```diff
@@ -62,10 +63,17 @@ impl Rule for NoArrayConstructor {
...
+            }
         };
 
-        if callee.is_specific_id("Array")
+        let Expression::Identifier(ident) = &callee else {
+            return;
+        };
+
+        if ctx.semantic().is_reference_to_global_variable(ident)
+            && callee.is_specific_id("Array")
```
camchenry, line 76: two small notes: 1) we should do the global check last (after all other checks), as it is more expensive, 2) we can use the `Expression.is_global_reference_name` method helper to make this a bit shorter:

```suggestion
        if callee.is_global_reference_name("Array", ctx.symbols())
```

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
```diff
@@ -104,6 +112,7 @@ fn test() {
         ("Array?.<Foo>();", None),
         ("Array?.(0, 1, 2);", None),
         ("Array?.(x, y);", None),
+        ("var Array; new Array;", None),
```
camchenry, line 115: if you are interested in contributing another PR, porting more test cases for this rule would be very helpful 🙏

camchenry (approved):

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
azihsoyn: Thanks!

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
azihsoyn: Of course! I'd be happy to help.

camc314 (approved):

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
azihsoyn: Since callee cannot call `is_global_reference_name` directly, I made a slight modification. [ac35515](https://github.com/oxc-project/oxc/pull/7067/commits/ac35515d7c8a210919cb5720cc3cc3431a9b15a3)

camc314 (approved):

camchenry (approved):

PR #7085, opened 2024-11-02
feat(editor/vscode): Support window/showMessage event

nrayburn-tech:
Whenever the language server sends a `window/showMessage` event, VS Code will listen and display a notification or log a message based on the severity level of the message.

The language server doesn't currently send this event.  I'm going to follow up with a separate PR for that.

Ref https://github.com/oxc-project/oxc/issues/6988

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -22,7 +22,7 @@
     "oxc"
   ],
   "engines": {
-    "vscode": "^1.90.0"
+    "vscode": "^1.95.0"
   },
   "sponsor": {
     "url": "https://github.com/sponsors/boshen"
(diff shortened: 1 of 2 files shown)


editors/vscode/package.json
```diff
@@ -22,7 +22,7 @@
     "oxc"
   ],
   "engines": {
-    "vscode": "^1.90.0"
+    "vscode": "^1.95.0"
```
nrayburn-tech, line 25: Without this change packaging fails.

```
> vsce package --no-dependencies -o oxc_language_server.vsix

 ERROR  @types/vscode 1.95.0 greater than engines.vscode ^1.90.0. Either upgrade engines.vscode or use an older @types/vscode version
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 1.
```

PR #7087, opened 2024-11-02
feat(oxc_language_server): Publish a message when there are UnknownRules errors building the LinterBuilder

nrayburn-tech:
I think this retains the current CLI behavior of exiting on unknown rules (that's still being discussed, but I figured I should just keep it the same here).  This modifies the language server to emit a message to editors which then display a notification within the editor.

Ref https://github.com/oxc-project/oxc/issues/6988

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -119,20 +119,20 @@ impl Runner for LintRunner {
 
         let oxlintrc_for_print =
             if misc_options.print_config { Some(oxlintrc.clone()) } else { None };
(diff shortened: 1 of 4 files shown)


nrayburn-tech:
The unknown rules check was pulled out of main for the 0.11 release. This needs to wait until that is re-added.

If it isn’t going to be added back (or added back differently), then this can be closed.

nrayburn-tech:
Closing for now since this probably won't be enabled again for a while. https://github.com/oxc-project/oxc/pull/7257

PR #7092, opened 2024-11-02
feat(linter): Add support for ignorePatterns property within config file

### LINKED ISSUE #7032: linter: add support "ignores" to config

0.10.3

### What command did you run?

_No response_

proposed example:

```json
{
  "ignores": [
      "packages/*",
      ".nx/",
      "dist/"
  ],
  "plugins": ["import", "unicorn"],
  "rules": {
    "oxc/no-barrel-file": "warn"
  }
  ```

### What happened?

When I use eslint flat config with eslint 9, eslint spam me warning about not use `.eslintignore` which is only way to ignore files and folders supported by `oxlint`. 

I suggest to support `"ignores"` in config from eslint flat config which way to set paths for ignoring. Link to `ignores` in eslint website: https://eslint.org/docs/latest/use/configure/ignore

After it implemented people from eslint 9 can use `oxlint` without annoying warnings.

### PR DESCRIPTION
**Pull request by nrayburn-tech**
This could probably use some tests, but I'm not really sure what exactly should be tested.

Will leave a review with a few comments on things that might need a different approach.

Closes #7032.

### CODE REVIEW

File: apps/oxlint/src/lint.rs [hunk 1]
```diff
@@ -115,6 +108,18 @@ impl Runner for LintRunner {
             Oxlintrc::default()
         };
 
+        let cwd = std::env::current_dir().unwrap();
+
+        let paths = Walk::new(
+            &paths,
+            &ignore_options,
+            &oxlintrc.ignores.iter().map(|value| cwd.join(value)).collect::<Vec<_>>(),
```
nrayburn-tech, line 116: This joins with the current working directory, but maybe it should be based on the config file path?

crates/oxc_language_server/src/main.rs
```diff
@@ -353,16 +353,21 @@ impl Backend {
         }
         if let Some(config_path) = config_path {
             let mut linter = self.server_linter.write().await;
+            let oxlintrc = Oxlintrc::from_file(&config_path)
+                .expect("should have initialized linter with new options");
+            let mut gitignore_builder = GitignoreBuilder::new(config_path.parent().unwrap());
+            for path in &oxlintrc.ignores {
+                gitignore_builder
+                    .add_line(None, path)
+                    .expect("Invalid glob in config \"ignores\" property");
```
nrayburn-tech, line 362: Pretty sure this will crash the language server if an invalid glob is supplied.  It would probably be better to just ignore that glob and display a message for the user, but I'm not sure.

crates/oxc_language_server/src/main.rs
```diff
@@ -353,16 +353,21 @@ impl Backend {
...
+            }
+            self.gitignore_glob.lock().await.push(gitignore_builder.build().unwrap());
```
nrayburn-tech, line 364: This modifies `self.gitignore_glob` after it has been created in `init_ignore_glob`.  This might be confusing and things could probably be refactored so that it isn't modified outside of `init_ignore_glob`.

crates/oxc_language_server/src/main.rs
pumano: Good idea

apps/oxlint/src/lint.rs
nrayburn-tech: Using the config file path is also problematic because you might store all your config in a configs folder, so then all your globs would need to traverse up to get out of it.

Maybe the current working directory is the easiest to explain right now. When thinking about nested config files I feel like it becomes less clear though.

pumano: Usually use case:
1) simple project: eslint config in root folder
2) Monorepo setup: each folder can contain eslint config extended from base root config file and then ovverride or add own rules and etc

crates/oxc_language_server/src/main.rs
nrayburn-tech: Modified this to publish a message instead of panic.  Example of the error message below.  It could probably be cleaned up some, but I couldn't come up with anything that was an improvement that I was also able to figure out how to implement in Rust.

image

pumano:
@Boshen Could you please a look at this PR?

apps/oxlint/src/lint.rs
camchenry: I recently added a `path` property to the `Oxlintrc` struct which stores a path to the configuration file. Feel free to use this, as I'm sure we'll probably need to do it anyway once we support nested configurations. It at least gets us one degree of abstraction away from "current working directory" as the only notable directory.

camchenry (commented):
I think this approach is okay for now, we'll likely need to adjust it when we support nested configuration. The main thing we should add in this PR is some test cases in `apps/oxlint/fixtures`. I would recommend adding a JSON config file that includes some common `ignores` patterns, and then add a test in `apps/oxlint/src/lint.rs` that the number of files is correct.

Sysix:
Do we want `ignores`? The name is for eslint v9, eslint v8 uses `ignorePatterns`:
https://eslint.org/docs/latest/use/configure/ignore-deprecated

nrayburn-tech:
I’m hoping to get to the feedback here in the next few days.

I’m also fine with just waiting until nested configs are supported and then creating a solution based on that if that would be preferred.

pumano:
> Do we want `ignores`? The name is for eslint v9, eslint v8 uses `ignorePatterns`: https://eslint.org/docs/latest/use/configure/ignore-deprecated

I'm really waiting any of it. If team want to support v8 first, it's nice to have `ignorePatterns` or maybe both as alias? It's cover v9 too

PR #7095, opened 2024-11-03
fix(linter): fix `is_method_call` with parentheses and chain expression

tbashiyy:
I fixed `is_method_call` behavior in specific case below.

For example, when we use `is_method_call(call_expr, None, Some(&["apply"]), Some(1), Some(1))` for [`(a?.b.apply)(args)`](https://playground.oxc.rs/#eNpFjkEKg0AMRe+SVQviAdx057alN4gaZWBMJJmxFfHunVG0q+Qnn//fCi1UcMNH2ZQ4TX6531AHu0MBAtUKGjkPWzjgF6qgkQrwjsO5WysTXWIZG/GnCopsvegIVY/eaCtgQjXSnIjey+dNISo/YzDXUR25DU5S3+5OZqVknumFSmxHasrI7XtG2q+K89BKRwNl5iRGx653/xcHFV+n3gwwkzZiCf1g27YfdH1fTA==), we cannot check correctly.

This is because we don't consider `call_expr` has parentheses and chain expression.

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -332,8 +332,14 @@ pub fn is_method_call<'a>(
         }
     }
 
-    let Some(member_expr) = call_expr.callee.without_parentheses().as_member_expression() else {
-        return false;
+    let callee_without_parentheses = call_expr.callee.without_parentheses();
+    let member_expr = match callee_without_parentheses {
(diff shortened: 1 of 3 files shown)


Boshen (changes requested):
We need a supporting test case.

Boshen (approved):

PR #7105, opened 2024-11-04
refactor(transformer/async-to-generator): move handling of `MethodDefinition`'s value to `exit_function`

Dunqing:
Part of #7074

We need to handle this before the `arrow_function` plugin inserts `_this = this` because, in the `async-to-generator` plugin, we will move the body to an inner generator function. However, we don't want `_this = this` moved to the inner generator function as well. So as long as we move first, and then insert, we can fix this problem.

The new semantic error is related to another tricky problem, I will fix it in another PR

PR #7117, opened 2024-11-04
fix(transformer/async_to_generator): fix checking if function is class method

overlookmotel:
Follow-on after #7105.

Correctly identify when a function is a method definition. `ctx.parent().is_method_definition()` would return `true` for this weird case where the function is the property key of the method, not the method itself:

```js
class C {
    [async function() {}]() {}
}
```

`matches!(ctx.parent(), Ancestor::MethodDefinitionValue(_))` only returns `true` if the function *is* a method definition. Of course, no-one would write such ridiculous code, but we may as well handle whatever is thrown at us.

It's also slightly more performant to check for one specific ancestor type, rather than:

https://github.com/oxc-project/oxc/blob/c2802e63fc0a43a7be3a4caf30d179db9eca6e5e/crates/oxc_traverse/src/generated/ancestor.rs#L1319-L1326

### DIFF
--- a/crates/oxc_transformer/src/es2017/async_to_generator.rs
+++ b/crates/oxc_transformer/src/es2017/async_to_generator.rs
@@ -131,7 +131,7 @@ impl<'a, 'ctx> Traverse<'a> for AsyncToGenerator<'a, 'ctx> {
     }
 
     fn exit_function(&mut self, func: &mut Function<'a>, ctx: &mut TraverseCtx<'a>) {
-        if func.r#async && ctx.parent().is_method_definition() {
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#7117** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7117?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Dunqing (approved):

Dunqing:
### Merge activity

* **Nov 4, 10:45 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7122, opened 2024-11-04
test(transformer): script to amend Babel fixtures

overlookmotel:
Add a NodeJS script which amends Babel's fixtures in place to remove transform plugins which we don't support from `options.json` files. Where options are changed from the original, the script runs Babel transform with the new options to regenerate the fixture `output.js` files.

Currently limited to transforming the fixtures for `babel-plugin-transform-class-properties` transform, but we can also enable it for other plugins if we wish in future, to get additional test coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,6 +241,10 @@ jobs:
           save-cache: ${{ github.ref_name == 'main' }}
           tools: just
 
+      - name: Update transformer fixtures
+        if: steps.filter.outputs.src == 'true'
+        run: node tasks/transform_conformance/update_fixtures.js
+
       - name: Check Conformance
         if: steps.filter.outputs.src == 'true'
         run: |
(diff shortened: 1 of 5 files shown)


overlookmotel:
* **#7122** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/workflows/ci.yml
```diff
@@ -240,6 +240,10 @@ jobs:
           save-cache: ${{ github.ref_name == 'main' }}
           tools: just
 
+      - name: Update transformer fixtures
+        if: steps.filter.outputs.src == 'true'
+        run: bun tasks/transform_conformance/update_fixtures.js
```
Boshen, line 245: I removed bun from the main branch, we should be using node.

overlookmotel: Now using NodeJS.

PR #7177, opened 2024-11-06
feat(transformer/modules): Add support for module transformation.

magic-akari:


Boshen:
We'll work on this after https://github.com/oxc-project/oxc/milestone/6

PR #7194, opened 2024-11-07
test(coverage): enable Babel es2022 tests

JLHwung:
Enabled the Babel es2022 tests. Previously they were skipped probably because ES2022 was not yet supported at that time.


Boshen:
Thank you Babel maintainer!

PR #7203, opened 2024-11-08
refactor(transformer/arrow-function-coverter): rename function name and add some comments to explain confusing parts.

Dunqing:
close: #7174

PR #7205, opened 2024-11-08
feat(transformer):  add ast changed hint in `InjectGlobalVariables`

IWANABETHATGUY:
here is a usage, https://github.com/rolldown/rolldown/blob/01197ec5548a091f8f1c1b6f112c1d30e883b0d9/crates/rolldown/src/utils/pre_process_ecma_ast.rs#L100-L106, 
this hint could avoid recreating semantic info, if the ast is not changed.

### DIFF
--- a/crates/oxc_transformer/src/plugins/inject_global_variables.rs
+++ b/crates/oxc_transformer/src/plugins/inject_global_variables.rs
@@ -107,6 +107,7 @@ impl<'a> From<&InjectImport> for DotDefineState<'a> {
 pub struct InjectGlobalVariablesReturn {
     pub symbols: SymbolTable,
     pub scopes: ScopeTree,
+    pub changed: bool,
 }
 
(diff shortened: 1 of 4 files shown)


Boshen (commented):
Need some tests.

IWANABETHATGUY:
For every test, I add an extra assert to make sure if the ast changed, the `XXXXReturn.changed` is also needs to be `true`

IWANABETHATGUY:
Seems noise, same as https://github.com/oxc-project/oxc/pull/7203#issuecomment-2463925786

Dunqing:
It seems the plugin is not very complex, I suggest syncing the semantic rather than adding a state to indicate it's no change.

IWANABETHATGUY:
> It seems the plugin is not very complex, I suggest syncing the semantic rather than adding a state to indicate it's no change.

Not sure this should do it with

PR #7211, opened 2024-11-09
feat(ast): serialize StringLiterals to ESTree

ottomated:


ottomated:
* **#7211** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ottomated:
@overlookmotel needs some work. There's a lot of places where string literals are constructed and don't have a `raw` value, and I'm not sure what to do there.

overlookmotel:
Hmm, yes. Tricky. Let me think about that a little... am just catching up now after being away.

overlookmotel:
Hmm, yes, tricky.

I've opened #7254 for the wider problem of these `raw` fields.

I now realize my suggestion to add a `raw` field to `StringLiteral` was unwise. Sorry.

I suggest for now just skip the `raw` field entirely in JSON AST, until we resolve #7254. If we decide to go the route of getting `raw` values via methods, we'll need a way to get the source text into the serializer. But, as we've discussed in #6347, we need to find a way to get external state into the serializer anyway, and we have a few options.

overlookmotel:
Implemented in #7263 instead.

PR #7234, opened 2024-11-11
fix(transformer/async-to-generator): `arguments` isn't correct after transformation

Dunqing:
Fix due to this plugin transforming the async method and async arrow function, it caused arguments no longer point the original function. 

For example:

Before 

```js
class Cls {
  async method() {
    () => {
    	console.log(arguments)
    }
  }
}
```

After:

```js
class Cls {
  method() {
    var _arguments = arguments;
    return babelHelpers.asyncToGenerator(function* () {
      () => {
        console.log(_arguments);
      };
    })();
  }
}
```

In this way, the `_arguments` is its original function's arguments

### For performance regression

It seems we need to check the IdentifierReference and BindingIdentifier if it's an `arguments`, that causes a significant regression, we may need a cheap way to do checking

### DIFF
--- a/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/assign/input.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/assign/input.js
@@ -1,0 +1,4 @@
+const ArgumentsAssignment = async () => {
+  let arguments = arguments;
+  console.log(arguments);
+};
--- a/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/nested-block/input.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/nested-block/input.js
@@ -1,0 +1,7 @@
+const ArrowFunction = async () => {
+  {
+    var arguments = arguments;
+    console.log(arguments);
+  }
+  console.log(arguments);
+};
--- a/crates/oxc_semantic/src/reference.rs
+++ b/crates/oxc_semantic/src/reference.rs
@@ -81,7 +81,7 @@ impl Reference {
     }
 
     #[inline]
-    pub(crate) fn set_symbol_id(&mut self, symbol_id: SymbolId) {
+    pub fn set_symbol_id(&mut self, symbol_id: SymbolId) {
         self.symbol_id = Some(symbol_id);
     }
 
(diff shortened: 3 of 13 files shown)


Dunqing:
* **#7234** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/assign/input.js
```diff
@@ -0,0 +1,4 @@
+const ArgumentsAssignment = async () => {
+  let arguments = arguments;
+  console.log(arguments);
+};
```
overlookmotel, line 4: This is quite a weird example. It'd be an immediate error to call this function because `let arguments = arguments;` is a TDZ violation.

Is it possible to make a more reasonable example which still tests what it's meant to test?

tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/nested-block/input.js
```diff
@@ -0,0 +1,7 @@
+const ArrowFunction = async () => {
+  {
+    var arguments = arguments;
+    console.log(arguments);
+  }
+  console.log(arguments);
+};
```
overlookmotel, line 7: This one is not a TDZ violation, but `var arguments = arguments;` is still very weird code.

overlookmotel (commented):
I've squashed the 3 commits and rebased on main, updating the code to take into account #7300. I caused that problem by delaying in reviewing this, so I've cleaned up the mess I made!

I can't say I really understand the logic of async function transform, so can't give a review on the "meat" of this PR. I've pushed a few commits with small optimizations and style nits. Feel free to revert any of them that you don't like.

I assume the perf impact is due to string comparisons on every single `IdentifierReference` and `BindingIdentifier`. Not much we can do about that at present I don't think.

I do think the fixtures should be altered, but @Dunqing feel free to merge and then do that in follow-up PR if you prefer.

tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/arguments/assign/input.js
Dunqing: Oh, yeah, I just realized this.

Dunqing:
### Merge activity

* **Nov 17, 12:01 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7245, opened 2024-11-12
fix(semantic): resolve function return type to function parent scope

### LINKED ISSUE #6387: Return type of function incorrectly resolved

Boshen:
closes #6387

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -1671,6 +1671,8 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         self.visit_formal_parameters(&func.params);
         if let Some(return_type) = &func.return_type {
             self.visit_ts_type_annotation(return_type);
+            // Resolve identifier references in return type.
+            self.resolve_references_for_current_scope();
         }
         if let Some(body) = &func.body {
             self.visit_function_body(body);
@@ -1724,6 +1726,8 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
 
         if let Some(return_type) = &expr.return_type {
             self.visit_ts_type_annotation(return_type);
+            // Resolve identifier references in return type.
+            self.resolve_references_for_current_scope();
         }
 
         self.visit_function_body(&expr.body);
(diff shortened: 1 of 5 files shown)


crates/oxc_semantic/src/builder.rs
```diff
@@ -1671,6 +1671,8 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         self.visit_formal_parameters(&func.params);
         if let Some(return_type) = &func.return_type {
             self.visit_ts_type_annotation(return_type);
+            // Resolve identifier references in return type.
+            self.resolve_references_for_current_scope();
```
Dunqing, line 1675: So far, we don't have extra logic in `Visitor` to resolve, so I think you should use the way https://github.com/oxc-project/oxc/pull/6388/ does

Dunqing (changes requested):

crates/oxc_semantic/src/builder.rs
overlookmotel: I'm sorry. I forgot about #6388. I *think* it fixes the issue, I just needed to add some tests, and I didn't get around to doing that.

PR #7256, opened 2024-11-12
feat(language_server): auto detect oxlint config

### LINKED ISSUE #5802: VSCode extension: default `oxc_language_server.configPath` seems wrong

Sysix:
closes #5802


crates/oxc_language_server/src/main.rs
```diff
@@ -67,11 +67,42 @@ impl Options {
...
+                ".oxlintrc.json",
+                ".oxlint.json",
+                ".oxlintrc",
+                ".eslintrc",
+            ];
+
+            for search_config in search_configs {
+                let config_path = root_path.join(search_config);
+
+                if config_path.exists() {
```
nrayburn-tech, line 85: This returns the first config file that exists, but there might be more than one match.  Should that be communicated somehow?

`Using oxlintrc.json, but found oxlint.json too. Specify oxlint.json in your settings to use it` or something like that as a log message.  It has some additional overhead, but could be useful for users.

I think just logging which config is being used could be helpful too. `Using oxlintrc.json` or `Automatically detected oxlintrc.json`.

crates/oxc_language_server/src/main.rs
```diff
@@ -67,11 +67,42 @@ impl Options {
...
         } else {
-            Some(PathBuf::from(&self.config_path))
+            // provided config path is not empty
+            let config_path = root_path.join(config_path);
+
+            if config_path.exists() {
+                return Some(config_path);
+            }
+
+            warn!("could not found oxlint config in {:?}", config_path);
```
nrayburn-tech, line 104:
```suggestion
            warn!("could not find oxlint config file {:?}", config_path);
```

crates/oxc_language_server/src/main.rs
```diff
@@ -67,11 +67,42 @@ impl Options {
...
+            None
```
nrayburn-tech, line 105: The CLI defaults to `oxlintrc.json`, should that be done here as well instead of nothing if a config isn't found?

Might also be worthwhile to enhance the CLI with the autodetection ability.

crates/oxc_language_server/src/main.rs
```diff
@@ -67,11 +67,42 @@ impl Options {
...
+    fn get_config_path(&self, root_path: &Path) -> Option<PathBuf> {
+        let Some(config_path) = &self.config_path else {
+            // no config file is provided, let search in the root_path for one
+            let search_configs = vec![
+                "oxlintrc.json",
+                "oxlint.json",
+                ".oxlintrc.json",
+                ".oxlint.json",
+                ".oxlintrc",
+                ".eslintrc",
```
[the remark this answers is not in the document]
Sysix: The old config  checked only for `.eslintrc`, I only included the line for non breaking change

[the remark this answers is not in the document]
Sysix: added `.eslintrc.json`

crates/oxc_language_server/src/main.rs
Sysix: added `info!("Automatically detected config {config_path:?}");`

crates/oxc_language_server/src/main.rs
Sysix: In this branch the user defines manually a config path, I would not try to find a config

nrayburn-tech: I think I left this at the wrong spot, I meant for it to be with the searching for different config files.

It wouldn’t make sense there either though because you would have already looked for this file and failed before returning none.

nrayburn-tech: I don’t think I have access to resolve this comment, but I’m good here.

Sysix: `oxlintrc.json` is the first file to be searched.   

I do not know if I want report an not found config (in search branch),  
because some projects have no config and using CLI only.

Please explain more what you want to see changed :)

camc314 (approved):

nrayburn-tech:
https://github.com/oxc-project/oxc/issues/4600 is an issue for the CLI equivalent and https://github.com/oxc-project/oxc/pull/4924 is the relevant PR for that.

Should probably review merging the CLI PR in if this goes in.

Sysix:
good point, I tried to add it into CLI https://github.com/oxc-project/oxc/pull/7348. This PR depends on another one :) 

Back to "Draft" until CLI is merged

Sysix:
Because in #7348 we decided to only check for `.oxlintrc.json`, I will close this PR and change the default value on the client side: #7442

PR #7283, opened 2024-11-14
fix(tasks): cargo clippy error on rust 1.82.0

shulaoda:
I mentioned this clippy issue in the discord help channel. The problem seems to be related to using the stable rust toolchain, but when I switched to version `1.81.0`, the issue was resolved. 

However, some of the clippy suggestions might still be worth considering and could be adopted.


Boshen:
We are stuck upgrading https://github.com/oxc-project/oxc/pull/6649

How come your rustup didn't pick up our rust-toolchain.toml?

shulaoda:
> How come your rustup didn't pick up our rust-toolchain.toml?

I contribute to many rust projects and use the stable toolchain by default. I recently cloned the `oxc` repository again, but I don’t know why the `oxc` toolchain wasn't automatically adopted. I later discovered this issue by chance.

PR #7290, opened 2024-11-15
fix(just): make submodules work on windows

### LINKED ISSUE #7282: `just submodules` does not work in Windows
Since powershell 5.x does not support "&&" and "||", just submodules does not work

ShuiRuTian:
Fix #7282 
- Change default shell to powershell 7(pwsh) to supports "&&"
- change url from git protocol to https protocol. This does not need a public token(I do not know what it is...).
- specify different submodule init command for windows

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,6 +1,6 @@
 #!/usr/bin/env -S just --justfile
 
-set windows-shell := ["powershell"]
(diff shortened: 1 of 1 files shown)


shulaoda:
It seems that this PR has not fully resolved the issue. Have you tested it on `windows`?

PR #7302, opened 2024-11-15
fix(editor): reload workspace configuration after change

### LINKED ISSUE #7308: editor: Sync configuration with running LSP without restarting
## Story
As a developer, I want config changes to take immediate effect so that I don't need to restart VSCode every time I modify them.

Sysix:
after changing `settings.json`, the changes whould not to updated internally. 

With this PR the updated values get passed to the language-server. 🥳 

probably closes https://github.com/oxc-project/oxc/issues/7308 because we also support the oxlint config

### DIFF
--- a/editors/vscode/client/config.ts
+++ b/editors/vscode/client/config.ts
@@ -4,11 +4,11 @@ import { IDisposable } from './types';
 export class ConfigService implements Config, IDisposable {
   private static readonly _namespace = 'oxc';
   private readonly _disposables: IDisposable[] = [];
-  private _inner: WorkspaceConfiguration;
-  private _runTrigger: Trigger;
-  private _enable: boolean;
-  private _trace: TraceLevel;
-  private _configPath: string;
+  private _inner!: WorkspaceConfiguration;
+  private _runTrigger!: Trigger;
+  private _enable!: boolean;
+  private _trace!: TraceLevel;
+  private _configPath!: string;
   private _binPath: string | undefined;
 
   public onConfigChange:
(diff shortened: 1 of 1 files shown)


editors/vscode/client/config.ts
```diff
@@ -4,24 +4,19 @@ import { IDisposable } from './types';
...
-  private _inner: WorkspaceConfiguration;
-  private _runTrigger: Trigger;
-  private _enable: boolean;
-  private _trace: TraceLevel;
-  private _configPath: string;
+  private _inner!: WorkspaceConfiguration;
+  private _runTrigger!: Trigger;
+  private _enable!: boolean;
+  private _trace!: TraceLevel;
+  private _configPath!: string;
```
nrayburn-tech, line 11: I don't like the non-null assertions here, but I understand the why.  A possible alternative would be to change the class from `implements Config` to contain a property `private _config: Config` and use that instead which wouldn't require non-null assertions.

Previously `ConfigService` was passed to `JSON.stringify` and that was used, but that was replaced with the `toLanguageServerConfig` method.  This is why `implements Config` isn't important anymore.

nrayburn-tech: I believe this is a summary of the changes you would need for my suggestion above.
- Replace all the `Config` related properties with a single `_config: Config` property
- Instead of `setSettingsFromWorkspace` modifying the config properties directly, make it `getSettingsFromWorkspace` which returns a `Config`. Update the `_config` property with this in the two places you need it.
- Modify the getters and setters to use the new `_config` property.

nrayburn-tech:
Is this actually a fix or just a refactor?  I'm not seeing where any behavior would actually be changed.

Sysix:
> Is this actually a fix or just a refactor? I'm not seeing where any behavior would actually be changed.

`this._inner = workspace.getConfiguration(ConfigService._namespace);` is set again when `onVscodeConfigChange` is called

editors/vscode/client/config.ts
nrayburn-tech: Added bonus to this is that if we only have a single config property, bugs like the one you are fixing here become less likely.  There won't be two possible ways to retrieve a property, it will always be `_config.<property>` instead of `_inner.<property>` or `_<property>`.

PR #6974, opened 2024-10-28
feat(linter): support `overrides` config field

DonIsaac:
Part of #5653

PR #7303, opened 2024-11-15
fix(linter)!: override plugins array when passed in config file

### LINKED ISSUE #6896: linter: default plugins are always enabled in config file

latest

### What command did you run?

npx oxlint -c .oxlint.json

{
  "plugins": ["typescript"]
}

### What happened?

because the [default values](https://oxc.rs/docs/guide/usage/linter/config.html#plugins) for `plugins` is `react`, `unicorn`, `typescript` and `oxc`,   
I expect to run `oxlint -c .oxlint.json` with less rules.

`npx oxlint -c .oxlint.json` output:

```
Finished in 16ms on 21 files with 98 rules using 24 threads.
```

`npx oxlint` output:

```
Finished in 16ms on 21 files with 98 rules using 24 threads.
```

camchenry:
- fixes https://github.com/oxc-project/oxc/issues/6896

When the `plugins` config property is specified, it will overwrite the default plugins array. This allows the plugins list to be easily customized and allows for disabling default plugins at the same time as enabling non-default ones.

- example: `{ "plugins": [] }` will enable no plugins.
- example: `{ }` will enable default plugins.
- example: `{ "plugins": ["typescript", "import"] }` will enable only the import and typescript plugins.

### DIFF
--- a/crates/oxc_linter/src/config/oxlintrc.rs
+++ b/crates/oxc_linter/src/config/oxlintrc.rs
@@ -142,29 +142,28 @@ mod test {
     #[test]
     fn test_oxlintrc_de_plugins_empty_array() {
         let config: Oxlintrc = serde_json::from_value(json!({ "plugins": [] })).unwrap();
+        assert_eq!(config.plugins, LintPlugins::empty());
+    }
+
+    #[test]
+    fn test_oxlintrc_empty_config_plugins() {
+        let config: Oxlintrc = serde_json::from_str(r"{}").unwrap();
         assert_eq!(config.plugins, LintPlugins::default());
     }
 
     #[test]
-    fn test_oxlintrc_de_plugins_enabled_by_default() {
-        // NOTE(@DonIsaac): creating a Value with `json!` then deserializing it with serde_json::from_value
-        // Errs with "invalid type: string \"eslint\", expected a borrowed string" and I can't
-        // figure out why. This seems to work. Why???
-        let configs = [
-            r#"{ "plugins": ["eslint"] }"#,
-            r#"{ "plugins": ["oxc"] }"#,
-            r#"{ "plugins": ["deepscan"] }"#, // alias for oxc
-        ];
-        // ^ these plugins are enabled by default already
-        for oxlintrc in configs {
-            let config: Oxlintrc = serde_json::from_str(oxlintrc).unwrap();
-            assert_eq!(config.plugins, LintPlugins::default());
-        }
-    }
+    fn test_oxlintrc_specifying_plugins_will_override() {
+        let config: Oxlintrc = serde_json::from_str(r#"{ "plugins": ["react", "oxc"] }"#).unwrap();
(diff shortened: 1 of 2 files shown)


Sysix:
Thanks on working on this. ❤️ With the new [overrides](https://github.com/oxc-project/oxc/pull/6974/files) property there is a new discussion to make :) 

```
{
  plugins: ["typescript"],
  overrides: [
    {
        files: ["*.tsx"],
        plugins: ["react"]
    }
  ]
}
```

with this config I expect to enable ts and react plugin for `.tsx` files. Maybe someone other expect it differently :)

Sysix:
That is a also a topic 👍 Let not block this PR, perrsonally im fine at the moment.

> if i specify an empty array. but have a rule in the rules list that uses a plugin does it get reported as an unknown rule?

existing rules in the project should always be known and can be enabled/disabled in the rules block.
plugins + categories enables the base rules, rules block should adjust the last needs.

That what I expect as a user, don't know what the core team says about it :)

camchenry:
> Thanks for working on this. ❤️ With the new [overrides](https://github.com/oxc-project/oxc/pull/6974/files) property there is a new discussion to make :)
> 
> ```
> {
>   plugins: ["typescript"],
>   overrides: [
>     {
>         files: ["*.tsx"],
>         plugins: ["react"]
>     }
>   ]
> }
> ```
> 
> with this config I expect to enable ts and react plugin for `.tsx` files. Maybe someone other expect it differently :)

Good point, I forgot to consider the `overrides` property here. I think that making the plugins array append when inside of overrides feels the most intuitive to me? I can update the tests and behavior accordingly.

> if i specify an empty array. but have a rule in the rules list that uses a plugin does it get reported as an unknown rule?

Yes, I think this should report as an unknown/disabled rule warning. I can imagine a case where you remove a plugin that is no longer needed and want to be warned of config that is no longer useful. Or, like in the case above with overrides, you need to explicitly enable the plugin to use the rule.

Sysix:
> I think that making the plugins array append when inside of overrides feels the most intuitive to me? 

Feels for me too, but I do not know how it is currently with `categories`. 
When `overrides` **_really overrides_** and not append. I think my expection is wrong and we should keep it consistent.

camchenry:
I checked the behavior with the legacy `.eslintrc` setup in ESLint, and it this config will append so that `.ts` files have the react and typescript plugins enabled:

```json
{
  "plugins": ["@typescript-eslint"],
  "overrides": [
    {
      "files": ["*.ts", "*.tsx"],
      "plugins": ["react"]
    }
  ]
}
```

So, I think this is the way to go for compatibility. And specifying an empty array here does not disable all of the plugins.

Boshen:
### Merge activity

* **Nov 16, 12:41 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7307, opened 2024-11-16
feat(linter): add `no-promise-in-callback`

no-yan:
related: #4655

This PR implements a rule to detect Promises inside error-first callbacks, preventing the mixed usage of callbacks and Promises.

Example of problematic code:
```javascript
a(function(err) { doThing().then(a) });
                  ^^^^^^^^^^^^^^
```

[Original implementation](https://github.com/eslint-community/eslint-plugin-promise/blob/266ddbb03076c05c362a6daecb9382b80cdd7108/rules/no-promise-in-callback.js)

### DIFF
--- a/crates/oxc_linter/src/rules/promise/no_promise_in_callback.rs
+++ b/crates/oxc_linter/src/rules/promise/no_promise_in_callback.rs
@@ -1,0 +1,160 @@
+use oxc_ast::ast::FormalParameters;
+use oxc_ast::AstKind;
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::utils::is_promise;
+use crate::{context::LintContext, rule::Rule, AstNode};
+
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/promise/no_promise_in_callback.rs
```diff
@@ -0,0 +1,168 @@
...
+        };
+
+        let mut ancestors = ctx.nodes().ancestors(node.id());
+        if ancestors.any(|node| is_callback_function(node, ctx)) {
+            ctx.diagnostic(no_promise_in_callback_diagnostic(call_expr.callee.span()));
+        }
+    }
+}
+
+fn is_callback_function<'a>(node: &AstNode<'a>, ctx: &LintContext<'a>) -> bool {
```
no-yan, line 74: In the original ESLint implementation, some functions were located in the `lib` directory.
I moved them into this rule file because they are not used by other rules.

camchenry (approved):
Nice work!

camc314 (approved):

PR #7315, opened 2024-11-16
feat(linter): add `typescript/no-require-imports` rule

pumano:


pumano:
add test case for covering require in function https://github.com/oxc-project/oxc/pull/7316#discussion_r1845228949 original rule has same behavior

PR #7324, opened 2024-11-17
docs(website): Link to specific ref when generating website docs

nrayburn-tech:
Currently the website links to main which is subject to change without an update to the website.  This updates the website to link to the specific commit that was used when the website was published.

Feel free to cleanup anything in this PR.

### DIFF
--- a/justfile
+++ b/justfile
@@ -207,6 +207,6 @@ clone-submodule dir url sha:
   cd {{dir}} ; git fetch --depth=1 origin {{sha}} ; git reset --hard {{sha}}
 
 website path:
-  cargo run -p website -- linter-rules --table {{path}}/src/docs/guide/usage/linter/generated-rules.md --rule-docs {{path}}/src/docs/guide/usage/linter/rules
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -207,6 +207,6 @@ clone-submodule dir url sha:
   cd {{dir}} ; git fetch --depth=1 origin {{sha}} ; git reset --hard {{sha}}
 
 website path:
-  cargo run -p website -- linter-rules --table {{path}}/src/docs/guide/usage/linter/generated-rules.md --rule-docs {{path}}/src/docs/guide/usage/linter/rules
+  cargo run -p website -- linter-rules --table {{path}}/src/docs/guide/usage/linter/generated-rules.md --rule-docs {{path}}/src/docs/guide/usage/linter/rules --git-ref $(git rev-parse HEAD)
```
nrayburn-tech, line 210: Using `$(...)` probably breaks Windows compatibility here.  I'm not sure of an alternative other than separating based on OS like `clone-submodule` does.

nrayburn-tech, line 210: This uses a commit hash, but tags and branches are also supported here.  Might be worth using a tag for the website releases, but default to a commit hash for development.

PR #7329, opened 2024-11-17
perf(transformer/arrow-functions): inline hot visitors

overlookmotel:
Follow-on after #7322. Add more `#[inline]` attrs to make absolutely sure these methods are inlined.


overlookmotel:
Not clear this is an improvement.

PR #7330, opened 2024-11-17
fix(transformer): remove typescript symbols after transformation

Boshen:


crates/oxc_transformer/src/typescript/annotations.rs
```diff
@@ -65,10 +66,18 @@ impl<'a, 'ctx> Traverse<'a> for TypeScriptAnnotations<'a, 'ctx> {
...
+        let scope_id = ctx.current_scope_id();
         program.body.retain_mut(|stmt| {
             let need_retain = match stmt {
                 Statement::ExportNamedDeclaration(decl) if decl.declaration.is_some() => {
-                    decl.declaration.as_ref().is_some_and(|decl| !decl.is_typescript_syntax())
+                    let keep =
+                        decl.declaration.as_ref().is_some_and(|decl| !decl.is_typescript_syntax());
+                    if !keep {
+                        decl.bound_names(&mut |ident| {
+                            ctx.scopes_mut().remove_binding(scope_id, &ident.name.clone().into());
```
Boshen, line 77: This is slow, need to batch the call.

PR #7335, opened 2024-11-18
feat(linter): implement `unicorn/prefer-string-raw` lint rule

ryan-m-walker:
There are a few [places in the original implementation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/prefer-string-raw.js#L46-L50) where the parent value node is compared against the actual node. I couldn't find a similar API to use here so instead opted to test if the span values matched which seemed reasonable to me but please let me know if I missed a better way to handle this. 

Implements [unicorn/refer-string-raw](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-raw.md) for #684


crates/oxc_linter/src/rules/unicorn/prefer_string_raw.rs
```diff
@@ -0,0 +1,282 @@
...
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        let raw = ctx.source_range(string_literal.span);
+
+        let last_char_index = raw.len() - 2;
+        if let Some('\\') = raw.char_at(Some(last_char_index as f64)) {
```
camc314, line 154: mind explaining to me what this line does? i can't work it out

like i know its checking to see if the last char in the string is a `\` but why?

camc314 (commented):
this looks great, thank you!

crates/oxc_linter/src/rules/unicorn/prefer_string_raw.rs
```diff
@@ -0,0 +1,282 @@
...
+            return;
+        };
+
+        if !raw.contains("\\\\") || raw.contains('`') || raw.contains("${") {
```
camc314, line 158: `"\\\\"` might be better as r#... ?

crates/oxc_linter/src/rules/unicorn/prefer_string_raw.rs
camc314, line 154: i don't think this needs to be an `if-let`, if can just be a direct comparison

camc314: ah i see this in the original impl, any ideas why?

ryan-m-walker: Apparently you can still escape backticks in String.raw so if the raw value end's in `\` it will result in an unterminated string.

image

PR #7351, opened 2024-11-18
perf(syntax): reorder operator enum variants

overlookmotel:
Re-order enum variants of `AssignmentOperator`, `BinaryOperator` and `UnaryOperator`.

* `Exponential` moved to after `Remainder` (so with the rest of the arithmetic operators).
* `Shift*` operators follow arithmetic operators.
* `AssignmentOperator::Bitwise*` ops moved to before `Logical*` ops (so all ops which correspond to `BinaryOperator`s are together).
* `*Or` always before `*And`.
* Plus/Addition always before Minus/Subtraction.

The purpose is to make the various methods on these types maximally efficient:

1. Group together variants so that `AssignmentOperator::is_*` methods can be executed with the minimum number of operations (essentially `variant - min <= max`).
2. Align the variants of `AssignmentOperator` and `BinaryOperator` so that conversion methods added in #7350 become very cheap too (essentially `if variant - min <= max { Some(variant + offset) } else { None }`).

### DIFF
--- a/crates/oxc_syntax/src/generated/derive_clone_in.rs
+++ b/crates/oxc_syntax/src/generated/derive_clone_in.rs
@@ -44,16 +44,16 @@ impl<'alloc> CloneIn<'alloc> for AssignmentOperator {
             Self::Multiplication => AssignmentOperator::Multiplication,
             Self::Division => AssignmentOperator::Division,
             Self::Remainder => AssignmentOperator::Remainder,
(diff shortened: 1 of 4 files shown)


Boshen:
### Merge activity

* **Nov 18, 8:08 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7352, opened 2024-11-19
feat(oxlint): add `cwd` property to `LintRunner`

Sysix:
allows to use multiple `LintRunner` in one file targeting multiple directory.
The current problem is for the test in #7348 we can not change `env::current_dir`,  
so as a workaround I introduce a new property :) 

See more info here: https://discord.com/channels/1079625926024900739/1117322804291964931/1308179827576016897

### DIFF
--- a/apps/oxlint/src/command/lint.rs
+++ b/apps/oxlint/src/command/lint.rs
@@ -68,6 +68,11 @@ pub struct BasicOptions {
     /// TypeScript `tsconfig.json` path for reading path alias and project references for import plugin
     #[bpaf(argument("./tsconfig.json"), hide_usage)]
     pub tsconfig: Option<PathBuf>,
+    
+    /// The working directory where oxlint should be executed
+    /// defaults to `env::current_dir()`
+    #[bpaf(long("working-dir"), hide_usage)]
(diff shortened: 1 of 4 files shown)


Boshen:
This is not a eslint CLI option, https://eslint.org/docs/latest/use/command-line-interface

> allows to use multiple LintRunner in multiple directory with only using one command / file.

Where and how is this going to be used? Should this be made internal without a CLI?

Sysix:
> Where and how is this going to be used? Should this be made internal without a CLI?

```
npx oxlint --working-dir packages/oxlint
npx oxlint --working-dir packages/vscode
```

instead of:
```
cd packages/oxlint && npx oxlint
cd ../..
cd packages/vscode && npx oxlint
```

and you can use two `LintRunner` instances with different working-dirs in one file/command.
This last feature is needed for the referenced PR :) 

> This is not a eslint CLI option,

I know :/ Maybe there is a better solution for my problem?

Sysix:
Updated the PR.

The remap of the paths is needed because else the path would be filtered out in line 100 (`Walk::...`).  
Maybe this can get even cleaner :)

PR #7337, opened 2024-11-18
chore(docs): Replace TAB with SPACE to render markdown correctly

leaysgur:
Possibly fixes #7336

PR #7362, opened 2024-11-19
chore: add tab and url check to changed files

### LINKED ISSUE #7299: feat: avoid referenced url in github to be master/main
Although #7298 rewrite the urls, it's still feasible to check in url that is realted to master/main branch. 

I suppose 2 options to avoid this in future:

1. git hooks, husky and some script. Check the committed files.
2. write a test(rust or js, whatever). Find the diff files between HEAD and main, and search the content.

For option 1, 

Pros: easy to understand 

Cons: 
- not forced, could be avoid manually, 
- make git operation slower

For option 2,

Pros: 
- must be runned, could not be skipped

Cons:
- Not common, RA is using this strategy
- The test case is not stable, related to branch and Head, this might be confusing.

ShuiRuTian:
fix #7299

follow #7337

This PR tried to add a way to check the changed files
- If it's a git url, it should not refer to master/main branch
- There should be no tab used in the repo

And also copy .editorconfig from RA, to help to unify the code style

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,2 +1,10 @@
+# https://EditorConfig.org
+root = true
+
+[*]
+charset = utf-8
 trim_trailing_whitespace = true
+end_of_line = lf
 insert_final_newline = true
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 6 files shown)


ShuiRuTian:
However, this is a draft PR for now, since I am not sure how we want to run the validations.

Seems I have 2 choices,
- run them as test cases, just like current implemention. This might be kind of confusing, since they are not really traditional test cases.
- run them through cli manually, maybe thorugh xtask or something like it. 

@Boshen @overlookmotel would you mind to share some opinions here?

Boshen:
I think this is an overkill. Add it to .github/workflows/link-check.yml if it's easy to do, otherwise don't bother too much ...

PR #7368, opened 2024-11-20
docs(semantic): document the meaning of `ReferenceFlags::Read` and `Write`

### LINKED ISSUE #5165: What does `ReferenceFlags::Read` mean?
This question came up previously in #3326, and has reared its head again in #5158.

`++x` currently gets flagged as `Write` only.

This doesn't make sense to me as it is equivalent to `x += 1` which is `Read | Write`.

The tests indicate this is intentional, but the reason why is not documented anywhere.

@Dunqing's [guess was](https://github.com/oxc-project/oxc/issues/3326#issuecomment-2118527489) "I believe this variable can be safely removed by the Minifier/Compressor."

But [as I pointed out](https://github.com/oxc-project/oxc/issues/3326#issuecomment-2118858837), `++x` can have side effects if `x` is an object with a `valueOf` method:

```js
let x = {
    valueOf() { console.log("side effect!"); }
};

++x; // Logs "side effect!"
```

So that suggests that `++x` cannot be removed by minifier in all cases.

Whatever the rationale is, can we please hash it out and document it?

Boshen:
closes #5165

### DIFF
--- a/crates/oxc_syntax/src/reference.rs
+++ b/crates/oxc_syntax/src/reference.rs
@@ -76,9 +76,31 @@ bitflags! {
     #[cfg_attr(feature = "serialize", derive(Serialize))]
     pub struct ReferenceFlags: u8 {
         const None = 0;
-        /// A symbol is being read as a Value
+        /// A symbol is being read as a Value.
+        ///
+        /// This value can be derived from the spec:
+        ///
+        /// Under `Runtime Semantics: Evaluation`, when [`GetValue`](https://tc39.es/ecma262/#sec-getvalue) is called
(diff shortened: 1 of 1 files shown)


Boshen:
* **#7368** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

PR #7388, opened 2024-11-21
fix(semantic)!: correct all `ReferenceFlags::Write` according to the spec

### LINKED ISSUE #7323: semantic: `ReferenceFlag::Write` should only be assigned for assignment expressions
```
var a, b;
z = (a, b = _o, x);
        ^ "flags": "Read | Write"
```

`b` should be "Write" only.

Dunqing:
close #7323

According to the specification re-design the JavaScript-part ReferenceFlags inferring approach.

* https://tc39.es/ecma262/#sec-assignment-operators-runtime-semantics-evaluation
* https://tc39.es/ecma262/#sec-postfix-increment-operator-runtime-semantics-evaluation
* https://tc39.es/ecma262/#sec-runtime-semantics-restdestructuringassignmentevaluation
* ... See references of https://tc39.es/ecma262/#sec-putvalue

### Changes

1. The left-hand of `AssignmentExpression` is always `ReferenceFlags::Write`
```js
let a = 0;
console.log(a = 0);
            ^ Write only
```

2. The `argument` of `UpdateExpression` is always `ReferenceFlags::Read | Write`
```js
let a = 0;
a++;
^ Read and Write
```

￼This change might cause some trouble for `Minfier` to remove this code, because ‘a’ is not used elsewhere. I have taken a look at `esbuild` and `Terser`. Only the `Terser` can remove this code.


Dunqing:
I prefer to fix transformer-related semantic errors in follow-up PR, but if you think should be fixed in this PR I am okay

Boshen (changes requested):
Can you add the test case to no-const-assign?

```
const FOO = 1;

({
  files = FOO,
} = arg1);
```

crates/oxc_semantic/src/builder.rs
```diff
@@ -890,6 +890,13 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
 
         let kind = AstKind::AssignmentExpression(self.alloc(expr));
         self.enter_node(kind);
+
+        // Only the operator is not an `=` operator, the left-hand side is both read and write.
```
Boshen, line 894:
```suggestion
        // Only when the operator is not an `=` operator, the left-hand side is both read and write.
```

Boshen (approved):

Boshen:
Can 

> Can you add the test case to no-const-assign?
> 
> ```
> const FOO = 1;
> 
> ({
>   files = FOO,
> } = arg1);
> ```

Can you add all the assignment cases to no object assign?

Dunqing:
> Can you add the test case to no-const-assign?
> 
> ```
> const FOO = 1;
> 
> ({
>   files = FOO,
> } = arg1);
> ```

Added

> Can you add all the assignment cases to no object assign?

I have already added a lot of assignment-related tests in semantic.

Boshen:
### Merge activity

* **Nov 22, 1:05 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
@Dunqing Is `x` in `x++` now `Read | Write`? I can't see tests for this (if I remember right, previously it was `Write` only which is incorrect).

Dunqing:
> @Dunqing Is `x` in `x++` now `Read | Write`? I can't see tests for this (if I remember right, previously it was `Write` only which is incorrect).

Yes, you are right. The test is added in https://app.graphite.dev/github/pr/oxc-project/oxc/7495.

PR #7390, opened 2024-11-21
perf(transformer/optional-chaining): mark `enter_expression` as inline

Dunqing:


Dunqing:
* **#7390** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7390?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7390?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

overlookmotel (commented):
I suggest to move more code out of `enter_expression`. To make sure `enter_expression` gets inlined, you want the function to be as small as possible. Only include logic which is "unless [cheap check], do nothing".

The point is that in most cases `expr` is *not* a `ChainExpression` or a `UnaryExpression` with `delete` operator, so want to make `enter_expression` as optimized as possible for the common "nothing to do here" path.

In practice, this means to move these two blocks out into separate functions:

```rs
if self.is_inside_function_parameter {
    // To insert the temp binding in the correct scope, we wrap the expression with
    // an arrow function. During the chain expression transformation, the temp binding
    // will be inserted into the arrow function's body.
    Self::wrap_arrow_function(expr, ctx)
} else {
    self.transform_chain_expression(false, expr, ctx)
}
```

```rs
if self.is_inside_function_parameter {
    // Same as the above explanation
    Self::wrap_arrow_function(expr, ctx)
} else {
    self.transform_chain_expression(true, &mut unary_expr.argument, ctx)
}
```

Dunqing:
> I suggest to move more code out of `enter_expression`. To make sure `enter_expression` gets inlined, you want the function to be as small as possible. Only include logic which is "unless [cheap check], do nothing".
> 
> The point is that in most cases `expr` is _not_ a `ChainExpression` or a `UnaryExpression` with `delete` operator, so want to make `enter_expression` as optimized as possible for the common "nothing to do here" path.
> 
> In practice, this means to move these two blocks out into separate functions:
> 
> ```rust
> if self.is_inside_function_parameter {
>     // To insert the temp binding in the correct scope, we wrap the expression with
>     // an arrow function. During the chain expression transformation, the temp binding
>     // will be inserted into the arrow function's body.
>     Self::wrap_arrow_function(expr, ctx)
> } else {
>     self.transform_chain_expression(false, expr, ctx)
> }
> ```
> 
> ```rust
> if self.is_inside_function_parameter {
>     // Same as the above explanation
>     Self::wrap_arrow_function(expr, ctx)
> } else {
>     self.transform_chain_expression(true, &mut unary_expr.argument, ctx)
> }
> ```

Thank you for your long explanation, I think I understand now!

Boshen:
No change 🤔 Maybe it was already inlined.

overlookmotel:
I'm going to try some stuff... marking as draft for now.

Boshen:
### Merge activity

* **Nov 25, 3:36 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
No change in benchmarks. But I still think this change is good. The smaller we can make all `enter_expression` / `exit_expression` methods, the more likely `enter_expression` / `exit_expression` in *other* transforms can be inlined too.

PR #7402, opened 2024-11-21
fix(semantic): function is value

ShuiRuTian:
`Function` should be `Value`


ShuiRuTian:
And a stupid question, just out of curious...

The implementation seems to refer typescript's design. Then why it is missed?

Boshen:
This breaks many tests 🤔

ShuiRuTian:
> This breaks many tests 🤔

Yeah, strange, investigating...

Dunqing:
The `SymbolFlags` is port from `TypeScript` and did some adjustments to support `non-strict` checks. I have changed this in #7479 as well

PR #7407, opened 2024-11-21
fix(linter): make `overrides` globs relative to config path

### LINKED ISSUE #7365: linter: overrides path does not seem to work
Following the doc, https://eslint.org/docs/v8.x/use/configure/configuration-files#how-do-overrides-work

The `overrides` configuration `"files": ["bin/*.js", "lib/*.js"]` does not work unless I prefix it with a `**`: `**/bin/*.js`. I think this is due to the glob trying to match a full path in our code.

camchenry:
- fixes https://github.com/oxc-project/oxc/issues/7365

currently, we are matching globs on the absolute path to the file, which means that in order to properly match, all globs would need to start with `**/`. this is not consistent with the behavior in ESLint, which is defined here:  https://eslint.org/docs/v8.x/use/configure/configuration-files#how-do-overrides-work

> **The patterns are applied against the file path relative to the directory of the config file.** For example, if your config file has the path `/Users/john/workspace/any-project/.eslintrc.js` and the file you want to lint has the path `/Users/john/workspace/any-project/lib/util.js`, then the pattern provided in `.eslintrc.js` is executed against the relative path `lib/util.js`.

This PR adds the ability to store the path to the configuration file along with the configuration data, so that we can later use it to resolve the relative paths of the files in the overrides section.

I'm not exactly sure if this will still make sense with nested configuration files, but I think it makes sense to store a path to each configuration file, alongside where the overrides are stored.

### DIFF
--- a/crates/oxc_linter/src/config/oxlintrc.rs
+++ b/crates/oxc_linter/src/config/oxlintrc.rs
@@ -1,4 +1,4 @@
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use oxc_diagnostics::OxcDiagnostic;
 use schemars::JsonSchema;
@@ -85,6 +85,9 @@ pub struct Oxlintrc {
     /// Add, remove, or otherwise reconfigure rules for specific files or groups of files.
     #[serde(skip_serializing_if = "OxlintOverrides::is_empty")]
     pub overrides: OxlintOverrides,
+    /// Absolute path to the configuration file.
+    #[serde(skip)]
+    pub path: PathBuf,
 }
 
 impl Oxlintrc {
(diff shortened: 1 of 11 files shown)


camchenry:
* **#7407** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

crates/oxc_linter/src/config/oxlintrc.rs
```diff
@@ -116,10 +119,15 @@ impl Oxlintrc {
             OxcDiagnostic::error(format!("Failed to parse eslint config {path:?}.\n{err}"))
         })?;
 
-        let config = Self::deserialize(&json).map_err(|err| {
+        let mut config = Self::deserialize(&json).map_err(|err| {
             OxcDiagnostic::error(format!("Failed to parse config with error {err:?}"))
         })?;
 
+        // Get absolute path from `path`
+        let absolute_path = path.canonicalize().unwrap_or_else(|_| path.to_path_buf());
```
Boshen, line 127: nit: I wonder whether this heavy sys call can be avoided.

PR #7423, opened 2024-11-22
feat(traverse): add `generate_uid_in_current_hoist_scope` method

Dunqing:
Add an API to handle variable hoisting.

### DIFF
--- a/crates/oxc_traverse/scripts/lib/walk.mjs
+++ b/crates/oxc_traverse/scripts/lib/walk.mjs
@@ -109,11 +109,26 @@ function generateWalkForStruct(type, types) {
     // but we don't take that into account.
     // Visitor should not do that though, so maybe it's OK.
     // In final version, we should not make `scope_id` fields `Cell`s to prevent this.
+
     enterScopeCode = `
       let previous_scope_id = ctx.current_scope_id();
-      ctx.set_current_scope_id((*(${makeFieldCode(scopeIdField)})).get().unwrap());
+      let current_scope_id = (*(${makeFieldCode(scopeIdField)})).get().unwrap();
+      ctx.set_current_scope_id(current_scope_id);
     `;
-    exitScopeCode = `ctx.set_current_scope_id(previous_scope_id);`;
+
+    exitScopeCode = 'ctx.set_current_scope_id(previous_scope_id);';
+
+    // const Var = Self::Top.bits() | Self::Function.bits() | Self::ClassStaticBlock.bits() | Self::TsModuleBlock.bits();
+    // `Function` type is a special case as its flags are set dynamically depending on the parent.
+    let isVarHoistingScope = type.name == 'Function' ||
+      ['Top', 'Function', 'ClassStaticBlock', 'TsModuleBlock'].some(flag => scopeArgs.flags.includes(flag));
+    if (isVarHoistingScope) {
+      enterScopeCode += `
+        let previous_hoist_scope_id = ctx.current_hoist_scope_id();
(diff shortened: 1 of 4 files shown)


Dunqing:
* **#7423** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7423?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7423?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

overlookmotel:
I've squashed the auto-fix commit and updated conformance results, so can see what's going on better here. Reviewing now...

overlookmotel (commented):
Some suggestions:

1. Add `current_var_hoisting_scope_id` method to `TraverseCtx` too.
2. Instead of `set_current_scope_id` taking 2 params, add a 2nd function `set_current_hoist_scope_id` instead.
  * Only generate calls to `set_current_hoist_scope_id` for the scopes which are "hoist scopes" (functions etc).
  * This should be faster for the majority of scopes which are not "hoist scopes".
3. Rather than all binding creation going via `add_binding` which has the cost of looking up scope flags and a branch, should we have `generate_uid_in_current_scope` and `generate_uid_in_current_hoist_scope`?
  * When generating a `let` binding, there's no reason to do any work checking flags.

These could be done in follow-up PRs if you prefer.

Longer-term we should probably (as you suggested previously) integrate creating the binding into `VarDeclarationsStore::insert_var` and `insert_let`. So then those 2 functions will already know what scope the binding should be in, as they know if it's a `var` or `let`.

Dunqing:
> 3\. Rather than all binding creation going via `add_binding` which has the cost of looking up scope flags and a branch, should we have `generate_uid_in_current_scope` and `generate_uid_in_current_hoist_scope`?

Yes, we even do not need to pass `SymbolFlags` in this API

overlookmotel:
> Yes, we even do not need to pass `SymbolFlags` in this API

Oh yes! I didn't even think of that.

Dunqing:
> > Yes, we even do not need to pass `SymbolFlags` in this API
> 
> Oh yes! I didn't even think of that.

Haha. I am trying to avoid adding new APIs (because we have many generate binding APIs), but it is worth doing it.

Dunqing:
> 1. Add `current_var_hoisting_scope_id` method to `TraverseCtx` too.

We have no real case that needs to be inserted to var hoisting scope id manually, so I think it doesn't need to be added to `TraverseCtx` for now. What are your thoughts on this? Please let me know.

overlookmotel:
We currently alias all methods of `TraverseScoping` (and `TraverseAncestry`) on to `TraverseCtx`, so personally I think it's preferable to be consistent. Generated code for `walk_*` methods can also use the shorter version.

Dunqing:
Seems we still need an API to check whether to use pass-in `scope_id` or `current_hoist_scope_id`, Please see https://github.com/oxc-project/oxc/pull/7469

overlookmotel (approved):
I pushed a few commits. Just nits, really.

Merging this now. @Dunqing if you don't like any of my changes, feel free to alter them in follow-ups.

overlookmotel:
### Merge activity

* **Nov 26, 6:17 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7444, opened 2024-11-23
feat(language_server)!: add capability `diagnosticProvider`

Sysix:
The current implementation checks on onChange / onSave. There they will respect the `run` option. 
This is not optimal because the client is sending events, which will be ignored by the server.

This implementation uses the feature, that the client requests diagnostics from the server. 
> Diagnostics are currently published by the server to the client using a notification. This model has the advantage that for workspace wide diagnostics the server has the freedom to compute them at a server preferred point in time. On the other hand the approach has the disadvantage that the server can’t prioritize the computation for the file in which the user types or which are visible in the editor. Inferring the client’s UI state from the [textDocument/didOpen](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen) and [textDocument/didChange](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didChange) notifications might lead to false positives since these notifications are ownership transfer notifications.
> 
> The specification therefore introduces the concept of diagnostic pull requests to give a client more control over the documents for which diagnostics should be computed and at which point in time.

_https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics_

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -17,12 +17,12 @@ use tower_lsp::{
         CodeActionProviderCapability, CodeActionResponse, ConfigurationItem, Diagnostic,
         DiagnosticOptions, DiagnosticServerCapabilities, DidChangeConfigurationParams,
         DidChangeTextDocumentParams, DidChangeWatchedFilesParams, DidCloseTextDocumentParams,
-        DidOpenTextDocumentParams, DidSaveTextDocumentParams, DocumentDiagnosticParams,
-        DocumentDiagnosticReport, DocumentDiagnosticReportResult, FullDocumentDiagnosticReport,
-        InitializeParams, InitializeResult, InitializedParams, OneOf,
-        RelatedFullDocumentDiagnosticReport, ServerCapabilities, ServerInfo,
-        TextDocumentSyncCapability, TextDocumentSyncKind, TextEdit, Url, WorkDoneProgressOptions,
-        WorkspaceEdit, WorkspaceFoldersServerCapabilities, WorkspaceServerCapabilities,
+        DidOpenTextDocumentParams, DocumentDiagnosticParams, DocumentDiagnosticReport,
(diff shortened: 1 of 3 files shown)


crates/oxc_language_server/src/main.rs
```diff
@@ -31,43 +33,26 @@ struct Backend {
     client: Client,
     root_uri: OnceCell<Option<Url>>,
     server_linter: RwLock<ServerLinter>,
+    document_content_cache: DashMap<String, String>,
```
Sysix, line 36: I do know now if this could lead to performance problems. We need to keep all content of any open file in the memory.

Boshen:
I've invited you and @nrayburn-tech as collaborators to help with review.

nrayburn-tech:
I don't think IntelliJ currently supports pulling diagnostics, so this would be a pretty big blocker for the IntelliJ plugin.  Either the IntelliJ plugin would have to implement this functionality or wait until it is added to IntelliJ.
https://plugins.jetbrains.com/docs/intellij/language-server-protocol.html#supported-features - Currently supported features.
https://youtrack.jetbrains.com/issues?q=%22textDocument%2Fdiagnostic%22 - Not finding any open issues for this feature.

I don't think it's worth merging until IntelliJ has support for this.  I didn't review the actual changes much given the lack of IntelliJ support.

nrayburn-tech:
I don’t have specifics, but my guess is that if we didn’t use the language server then the effort involved to maintain the JetBrains plugin would significantly increase. There might be community alternatives available that do something similar, but I don’t know off hand.

@Boshen is it worth discussing using something other than the official language server support in JetBrains IDEs?

FYI @IWANABETHATGUY since you started the plugin.

Sysix:
for the afterworld:
https://youtrack.jetbrains.com/issue/IJPL-172853/LSP-Add-support-for-textDocument-diagnostic-and-workspace-diagnostic

PR #7481, opened 2024-11-25
fix(transformer/class-properties): transform `this` in static prop initializers

overlookmotel:
Convert `this` in static property initializers to reference to class name.

`class C { static prop = this; }` -> `class C {}; C.prop = C;`


overlookmotel:
I think I know this transform better than anyone else at present, so merging without review to avoid wasting anyone else's time.

PR #7529, opened 2024-11-28
feat(napi/parser): introduce experimental `MagicString`

Boshen:
Hold magic string instance on the Rust side for utf8 string manipulation.


yyx990803:
I like this - having this would actually allow Vue's `compiler-sfc` to replace `@babel/parser` and `MagicString` with just `oxc-parser`.

- Performance-wise, instead of making napi calls on every op, we should probably buffer the operations on the JS side and lazy-apply them only when `toString` is actually called.

- With the buffering, in Rolldown we can also expose this API (maybe with the AST already available via hook arguments) and allow plugin `transform` hooks to return an oxc `ParserBuilder` instance instead of strings so we don't even need to pass the finalized string to JS ever.

overlookmotel:
I believe the API this PR provides is:

```js
const code = 'const s: String = "测试"';
const oxc = new ParserBuilder(code);
const {program} = oxc.parseSync({ sourceFilename: 'test.ts' });
const stringLiteral = program.body[0].declarations[0].init;
// Magic string manipulation
oxc.remove(stringLiteral.start + 1, stringLiteral.end - 1);
// Get altered code
const alteredCode = oxc.toString();
```

How about returning the magic string from `parseSync` instead of passing in a `ParserBuilder`?

```js
const code = 'const s: String = "测试"';
const {program, magicString} = parseSync(code, { sourceFilename: 'test.ts', magicString: true });
const stringLiteral = program.body[0].declarations[0].init;
// Magic string manipulation
magicString.remove(stringLiteral.start + 1, stringLiteral.end - 1);
// Get altered code
const alteredCode = magicString.toString();
```

Builder patterns are quite rusty, but personally I think this is more idiomatic in JS.

Boshen:
> I believe the API this PR provides is:
> 
> ```js
> const code = 'const s: String = "测试"';
> const oxc = new ParserBuilder(code);
> const {program} = oxc.parseSync({ sourceFilename: 'test.ts' });
> const stringLiteral = program.body[0].declarations[0].init;
> // Magic string manipulation
> oxc.remove(stringLiteral.start + 1, stringLiteral.end - 1);
> // Get altered code
> const alteredCode = oxc.toString();
> ```
> 
> How about returning the magic string from `parseSync` instead of passing in a `ParserBuilder`?
> 
> ```js
> const code = 'const s: String = "测试"';
> const {program, magicString} = parseSync(code, { sourceFilename: 'test.ts', magicString: true });
> const stringLiteral = program.body[0].declarations[0].init;
> // Magic string manipulation
> magicString.remove(stringLiteral.start + 1, stringLiteral.end - 1);
> // Get altered code
> const alteredCode = magicString.toString();
> ```
> 
> i.e. You choose whether you want a `MagicString` to be returned from `parseSync` or not with an option, following same pattern as how you choose if you want a source map or not.
> 
> Builder patterns are quite rusty, but personally I think this is more idiomatic for a JS API.

I can't make your suggestion work without cloning the string

```
#[napi]
pub struct ParserBuilder {
    cell: ParserBuilderImpl,
}

self_cell!(
    struct ParserBuilderImpl {
        owner: String,

        #[covariant]
        dependent: MagicString,
    }
);
```

overlookmotel:
Ah OK. I don't know the internals. If we *could* make my suggestion work, would you prefer it as an API?

napi/parser/src/types.rs
```diff
@@ -22,13 +25,38 @@ pub struct ParserOptions {
...
+        mem::take(&mut self.comments)
+    }
+
+    pub fn errors(&mut self) -> Vec<Error> {
+        mem::take(&mut self.errors)
+    }
+
+    pub fn magic_string(&mut self) -> MagicString {
+        MagicString::new(mem::take(&mut self.source_text))
+    }
```
Boshen, line 59: @Brooooooklyn is there any better approach to return this magic string instance without changing `ParseResult` from a `#[napi(object)]` to `#[napi]` class?

Boshen: Perhaps changing this to a class is better, I can add methods for getting line column numbers, and other string utility methods.

I'll just need to change this to `ParserReturn` with a result object.

Brooooooklyn: We can add `#[napi(getter)]` to the `ParserReturn` fields rather than the `methods`:

```rust
#[napi]
impl ParseResult {
    #[napi(getter, ts_return_type = "import(\"@oxc-project/types\").Program")]
    pub fn get_ast(&mut self) -> String {
        mem::take(&mut self.program)
    }
}
```

Generated .d.ts should be:

```ts
export class PasreResult {
  get ast(): import("@oxc-project/types").Program
}
```

PR #7547, opened 2024-11-29
refactor(ast)!: change 'raw' from &str to Option<Atom>

### LINKED ISSUE #7254: Make `raw` field on literal AST types optional, or remove them
### The problem

#7211 brought up a problem. Most of our literal types (`NumericLiteral` etc) have `raw` field.

This makes sense when the AST has come from the parser, but little sense when the node is generated (e.g. in transformer), because the node has no "raw" representation.

Having to provide `"0"` here feels like a hack to me, for example:

https://github.com/oxc-project/oxc/blob/44375a5662ee3e6451f1a5335c75d6379d1878a6/crates/oxc_ast/src/ast_builder_impl.rs#L209-L211

The exception is `StringLiteral` which *doesn't* have a `raw` field. This feels like an omission - if the other literal types have a `raw` field, it should have one too. But adding one would be a pain, as there are a lot of places we generate strings in transformer etc.

### Prior art

Babel's types have the `raw` field under `extra` which is optional:

```ts
export interface StringLiteral extends BaseNode {
  type: "StringLiteral";
  value: string;
}

interface BaseNode {
  type: Node["type"];
  // ...
  extra?: Record<string, unknown>;
}
```

[StringLiteral](https://github.com/babel/babel/blob/38d26cd5eeb66b697671cfb8c78f963f02992073/packages/babel-types/src/ast-types/generated/index.ts#L497-L500) [BaseNode](https://github.com/babel/babel/blob/38d26cd5eeb66b697671cfb8c78f963f02992073/packages/babel-types/src/ast-types/generated/index.ts#L37-L47)

Acorn's types also have the `raw` field as optional:

```ts
export interface Literal extends Node {
  type: "Literal"
  value?: string | boolean | null | number | RegExp | bigint
  raw?: string
  regex?: {
    pattern: string
    flags: string
  }
  bigint?: string
}
```

[source](https://github.com/acornjs/acorn/blob/3c6a5a98e8c6dcefef391e6114cc7e64657979d3/acorn/src/acorn.d.ts#L27-L36)

### Possible solutions

I can see 2 options:

#### Make `raw` fields optional

* Change `raw` field on `NumericLiteral` etc to `Option<&'a str>`.
* Add a `raw: Option<&'a str>` field to `StringLiteral`.

#### Remove all `raw` fields

Where an AST node has a non-empty span, the raw value can be obtained by slicing source text. So remove all the `raw` fields, and use methods to get raw value where required:

```rs
impl NumericLiteral {
    pub fn raw<'a>(&self, source_text: &'a str) -> Option<&'a str> {
        if self.span.start == 0 && self.span.end == 0 {
            None
        } else {
            Some(&source_text[self.span.start..self.span.end])
        }
    }
}
```

(as suggested in #5522)

### Which?

Personally I prefer the 2nd. I don't think the `raw` fields are used much, so they're bloating the AST for little value. The methods to get raw value from `Span` are pretty trivial and cheap.

ShuiRuTian:
Fix #7254 

Changed all "raw" propperty of literal types(if it has this property) to Option<Atom>

### DIFF
--- a/crates/oxc_ast/src/ast_impl/literal.rs
+++ b/crates/oxc_ast/src/ast_impl/literal.rs
@@ -82,20 +82,20 @@ impl ContentHash for NumericLiteral<'_> {
 
 impl fmt::Display for NumericLiteral<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.raw.fmt(f)
+        self.raw.as_ref().unwrap().fmt(f)
     }
 }
 
 impl BigIntLiteral<'_> {
     /// Is this BigInt literal zero? (`0n`).
     pub fn is_zero(&self) -> bool {
-        self.raw == "0n"
+        matches!(&self.raw, Some(x) if x == "0n")
     }
 }
 
 impl fmt::Display for BigIntLiteral<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.raw.fmt(f)
+        self.raw.as_ref().unwrap().fmt(f)
     }
 }
 
--- a/crates/oxc_ast/src/ast_impl/literal.rs
+++ b/crates/oxc_ast/src/ast_impl/literal.rs
@@ -71,6 +71,23 @@ impl NumericLiteral<'_> {
             int32bit as i32
         }
     }
+
+    /// Provide a fallback for converting NumericLiteral's value to a string if `raw` is None
+    pub fn raw_str(&self) -> Cow<str> {
+        match self.raw.as_ref() {
+            Some(raw) => return Cow::Borrowed(raw),
+            None => {
+                let value = self.value;
+                let s = match self.base {
+                    NumberBase::Binary => format!("{:#b}", value as i64),
+                    NumberBase::Decimal | NumberBase::Float =>  format!("{}", value),
+                    NumberBase::Hex => format!("{:#x}", value as i64),
+                    NumberBase::Octal => format!("{:#o}", value as i64),
+                };
+                Cow::Owned(s)
+            }
+        }
+    }
 }
 
 impl ContentHash for NumericLiteral<'_> {
@@ -82,20 +99,24 @@ impl ContentHash for NumericLiteral<'_> {
 
 impl fmt::Display for NumericLiteral<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.raw.as_ref().unwrap().fmt(f)
+        // There are basically 2 choices:
+        // 1. If raw.is_none(), then use the NumericLiteral's number field. or
+        // 2. Always use the NumericLiteral's number field.
+        // For now, the first approach is chosen, since to_string is only used in linter, in which raw does matter
+        self.raw_str().fmt(f)
     }
 }
 
 impl BigIntLiteral<'_> {
     /// Is this BigInt literal zero? (`0n`).
     pub fn is_zero(&self) -> bool {
-        matches!(&self.raw, Some(x) if x == "0n")
+        self.raw == "0n"
     }
 }
 
 impl fmt::Display for BigIntLiteral<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.raw.as_ref().unwrap().fmt(f)
+        self.raw.fmt(f)
     }
 }
 
--- a/crates/oxc_ast/src/ast_impl/literal.rs
+++ b/crates/oxc_ast/src/ast_impl/literal.rs
@@ -72,7 +72,8 @@ impl NumericLiteral<'_> {
         }
     }
 
-    /// Provide a fallback for converting NumericLiteral's value to a string if `raw` is None
+    /// Return raw source code for `NumericLiteral`.
+    /// If `raw` is `None` (node is generated, not parsed from source), fallback to formatting `value`.
     pub fn raw_str(&self) -> Cow<str> {
         match self.raw.as_ref() {
             Some(raw) => Cow::Borrowed(raw),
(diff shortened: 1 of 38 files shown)


crates/oxc_prettier/src/format/mod.rs
```diff
@@ -2125,7 +2125,11 @@ impl<'a> Format<'a> for PropertyKey<'a> {
                     if need_quote {
                         dynamic_text!(
                             p,
-                            string::print_string(p, literal.raw, p.options.single_quote)
+                            string::print_string(
+                                p,
+                                literal.raw.as_ref().unwrap(),
```
overlookmotel, line 2130: It's possible that prettier will be run on an AST which has been through transformer, so may contain generated nodes, which may have `raw: None`. So need to fall back to converting `literal.value` to a string if `raw.is_none()`.

crates/oxc_semantic/src/checker/javascript.rs
```diff
@@ -293,10 +293,12 @@ pub fn check_number_literal(lit: &NumericLiteral, ctx: &SemanticBuilder<'_>) {
...
     if ctx.strict_mode() {
         match lit.base {
-            NumberBase::Octal if leading_zero(lit.raw) => {
+            NumberBase::Octal if leading_zero(lit.raw.as_ref().unwrap().as_str()) => {
                 ctx.error(legacy_octal(lit.span));
             }
-            NumberBase::Decimal | NumberBase::Float if leading_zero(lit.raw) => {
+            NumberBase::Decimal | NumberBase::Float
+                if leading_zero(lit.raw.as_ref().unwrap().as_str()) =>
+            {
```
overlookmotel, line 301: Semantic checker can be run on an AST which has been through transformer or minifier, or otherwise programatically generated. So it's possible `raw` is `None`, and shouldn't unwrap.

Here I think solution is just to skip the checks if `raw` is `None`. e.g.:

```rs
fn leading_zero(s: &Option<Atom>) -> bool {
    if let Some(s) = s.as_ref() {
        let mut chars = s.bytes();
        if let Some(first) = chars.next() {
            if let Some(second) = chars.next() {
                return first == b'0' && second.is_ascii_digit();
            }
        }
    }
    false
}

if ctx.strict_mode() {
    match lit.base {
        NumberBase::Octal if leading_zero(&lit.raw) => {
            ctx.error(legacy_octal(lit.span));
        }
        NumberBase::Decimal | NumberBase::Float if leading_zero(&lit.raw) => {
            ctx.error(leading_zero_decimal(lit.span));
        }
        _ => {}
    }
}
```

crates/oxc_traverse/src/ast_operations/gather_node_parts.rs
```diff
@@ -466,7 +466,7 @@ impl<'a> GatherNodeParts<'a> for StringLiteral<'a> {
 
 impl<'a> GatherNodeParts<'a> for NumericLiteral<'a> {
     fn gather<F: FnMut(&str)>(&self, f: &mut F) {
-        f(self.raw);
+        f(self.raw.as_ref().unwrap().as_str());
```
overlookmotel, line 469: `raw` could be `None`. Need a fallback to use `value`, instead of unwrapping.

overlookmotel (changes requested):
Thanks loads for tackling this.

Turns out this is more complicated than I thought! Sorry for the litany of comments below.

First off, could you please rebase on main and fix the merge conflicts so CI can run?

Most of my comments are the same - they relate to not unwrapping when can't be sure that `raw` is `Some`.

In lots of places we need a fallback for converting `NumericLiteral`'s `value` to a string if `raw` is `None`. Probably it'd be good to add a method `NumericLiteral::raw_str` which can be used in all those places (which I guess will need to return a `Cow<str>`).

crates/oxc_ast/src/ast_impl/literal.rs
```diff
@@ -82,20 +82,20 @@ impl<'a> ContentHash for NumericLiteral<'a> {
 
 impl<'a> fmt::Display for NumericLiteral<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.raw.fmt(f)
+        self.raw.as_ref().unwrap().fmt(f)
```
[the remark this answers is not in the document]
ShuiRuTian: NumericLiteral::to_string is only used in linter now, so I choose the first approach for now.

ShuiRuTian:
Thanks a lot for the detailed comments!

Sorry, I was indeed abusing unwrap...

Add a new method called `raw_str` as suggested.

Most problems should already be fixed.

overlookmotel:
> Thanks a lot for the detailed comments!

Glad you found them useful and not annoying!

Nice speed bump on transformer benchmarks.

Could you please try and fix the failing test, and fix clippy, and then I'll review again?

A couple of quick things in meantime:

1. When formatting `NumericLiteral`'s `value`, I don't think there's any need to look at what `base` is. Just `Cow::Owned(format!("{}", self.value))` should do it.
2. Please make sure all changes related to `BigIntLiteral` are removed. I can see a few remaining in the diff.

Side note: I think we can also remove `NumericLiteral::base` field later on, or make it an `Option`. Like `raw`, it only has meaning where the `NumericLiteral` came from source code. In a `NumericLiteral` which is generated, it's just a value. It has no "base".

overlookmotel:
I believe the CI failure on "Test NAPI" CI job is unrelated. I've also seen it on one of my PRs #7618.

Please let me know when ready for me to review again.

ShuiRuTian:
> I believe the CI failure on "Test NAPI" CI job is unrelated to content of this PR. I've also seen it on one of my recent PRs https://github.com/oxc-project/oxc/pull/7618.

Cool, if so, the PR should be ready to be reviewed now.

There might also be some comment issues, like whether I should add "Safety: xxx" above unwrap or some comments is not clearly enough.  but the logic part should be done.

overlookmotel (approved):
Branch got out of date with main. I've rebased on main and fixed conflicts. I've also taken the liberty of pushing a few small commits to fix a few things, rather than going through another round of review - I hope you don't mind.

Am going to merge now. @ShuiRuTian please take a look at the extra commits I pushed if you're interested.

Thanks loads for your work on this!

ShuiRuTian:
Thanks much for the detailed guide! It really helps a lot

overlookmotel:
To be honest, I don't think this one is correct: https://github.com/oxc-project/oxc/pull/7547/commits/4874660b93646d1cc624f2b95e8818111239c72e

If `raw` is `None` then it returns `None`, which it shouldn't. But how `RegExpLiteral` stores `raw` needs changes anyway. At present it stores it in 2 places.

PR #7553, opened 2024-11-30
fix(oxc_transformer): correct generate `ThisExpr` and `import.meta` in jsx pragma

IWANABETHATGUY:


Boshen:
@overlookmotel need your review.

overlookmotel (changes requested):
Sorry to say I see problems.

Also, could you please add a test for `import.meta`?

We can make this more performant by changing `Pragma` into an enum, to avoid string comparisons on every call to `create_expression`. Something like:

```rs
enum Pragma<'a> {
  /// `createElement`
  Single(Atom<'a>),
  /// `React.createElement`
  Double(Atom<'a>, Atom<'a>),
  /// `foo.bar.qux`
  Many(Vec<Atom<'a>>),
  /// `this`, `this.foo`, `this.foo.bar.qux`
  This(Vec<Atom<'a>>),
  /// `import.meta`, `import.meta.foo`, `import.meta.foo.bar.qux`
  ImportMeta(Vec<Atom<'a>>),
}
```

In practice, it's going to always be `Single` or `Double`. All the others are edge cases.

But we can do that optimization in a follow-up.

crates/oxc_transformer/src/jsx/jsx_impl.rs
```diff
@@ -1067,3 +1069,60 @@ fn create_static_member_expression<'a>(
...
+        };
+        let pragma =
+            Pragma::parse(Some(&"import.meta.prop".to_string()), "createElement", ast, &ctx);
+
+        let expr = pragma.create_expression(&mut traverse_ctx);
+        let member_expr = expr.as_member_expression().unwrap();
+        assert_eq!(member_expr.static_property_name(), Some("prop"));
+        match member_expr.object() {
+            Expression::MetaProperty(_) => {}
+            _ => panic!("Expected ThisExpression"),
```
overlookmotel, line 1125:
```suggestion
            _ => panic!("Expected MetaProperty"),
```

overlookmotel: Or how about:

```rs
assert!(matches!(member_expr.object(), Expression::MetaProperty(_)));
```

overlookmotel (approved):
Thanks for adding the tests.

I've pushed a couple more commits:

1. Get rid of the `Box`es in `create_arbitrary_length_member_expr_or_ident`. They didn't appear to add anything.
2. Tweak the tests to make them more explicit.

overlookmotel:
### Merge activity

* **Dec 3, 8:53 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7554, opened 2024-11-30
feat(parser,linter)!: use a different `ModuleRecord` for linter

Boshen:
The parser returns a simple `ModuleRecord` that is allocated in the arena for performance reasons.

The linter uses a more complicated, `Send` + `Sync` `ModuleRecord` that will hold more cross-module information.

The next step is to return more esm information from the parser to eliminated the need of the `oxc_module_lexer` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1986,7 +1986,6 @@ version = "0.38.0"
 dependencies = [
  "assert-unchecked",
  "bitflags 2.6.0",
- "dashmap 6.1.0",
  "nonmax",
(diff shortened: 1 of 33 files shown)


Boshen:
### Merge activity

* **Nov 30, 10:56 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7566, opened 2024-12-01
fix(apps/oxlint): incorrect matching in `.oxlintignore`

shulaoda:
The issue was discovered while updating test cases in `rolldown`.

In the `.oxlintignore` file, we have:
```ignore
tests/**
```

When running the command by `lint-staged`:

```bash
oxlint -c .oxlintrc.json --ignore-path=.oxlintignore --deny-warnings "tests/function/main.js"
```

The file `main.js` gets linted despite being ignored. This happens because, before using `Gitignore::new`, we converted paths to absolute paths, causing the pattern `tests/**` to not match `D:/rolldown/tests/function/main.js` correctly.

https://github.com/oxc-project/oxc/blob/c61a383e8cc453fb92f85e6047426a86e50daba7/apps/oxlint/src/lint.rs#L60-L78


Boshen (commented):
Can we add a test for catching regressions?

pumano:
Today I test it too, and got same problem.

PR #7590, opened 2024-12-03
fix(napi/transform): respect `options.sourcemap` for id

sxzz:
Don't pass `source_map_path` if `options.sourcemap` is undefined or false; then, `IsolatedDeclarationsResult.map` should be undefined.

Downstream issue https://github.com/unplugin/unplugin-isolated-decl/issues/42


Dunqing (approved):

Boshen:
Let me add a test.

PR #7593, opened 2024-12-03
test(oxc_transformer): define works differently with esbuild

sapphi-red:
See https://github.com/oxc-project/oxc/issues/7594


crates/oxc_transformer/tests/integrations/plugins/replace_global_defines.rs
```diff
@@ -59,6 +59,14 @@ fn dot() {
     test("process['env'].NODE_ENV", "production", config.clone());
 }
 
+#[ignore]
```
sapphi-red, line 62: Added for now as this test is failing now

IWANABETHATGUY (approved):

PR #7600, opened 2024-12-03
refactor(prettier): Unified method of `Doc` generation

leaysgur:
Before:

- `Doc::Xxx()` + `XXX.with_xxx()`
- `macro!()`
- `p.xxx()`

After:

- `p.xxx()` with `p_vec!` macro


Boshen (approved):
I thought the macros make the code easier to read ;-)

But you are the boss now.

leaysgur:
My original intention was just to unify the way of creation, and in fact I'm still undecided about whether to use a macro or a function. 😅 

Please let me think a bit~!

PR #7343, opened 2024-11-18
refactor(ast)!: remove `impl GetAddress for Function`

overlookmotel:
`impl GetAddress for Function` was added as a hack, in the absence of another way to get the `Address` of a `&Function` (https://github.com/oxc-project/backlog/issues/140).

Remove it, and use `Address:from_ptr` instead in JSX Refresh transform, which is only place using it.

PR #7621, opened 2024-12-03
fix(oxc_transform): overlap replacement

### LINKED ISSUE #7594: `define` works differently from esbuild if it includes `import.meta.something` and `import.meta.something.anotherthing`
I made a test case as that was easier than explaining in text: https://github.com/oxc-project/oxc/pull/7593
[esbuild try](https://esbuild.github.io/try/#dAAwLjI0LjAALS1kZWZpbmU6aW1wb3J0Lm1ldGEuZW52LkZPTz1pbXBvcnQubWV0YS5lbnYuRk9PIC0tZGVmaW5lOmltcG9ydC5tZXRhLmVudj1fX2Zvb19fAGNvbnNvbGUubG9nKGltcG9ydC5tZXRhLmVudiwgaW1wb3J0Lm1ldGEuZW52LkZPTyk)

IWANABETHATGUY:
closed  https://github.com/oxc-project/oxc/issues/7594

### Reference
https://github.com/oxc-project/oxc/pull/7343/files

### DIFF
--- a/crates/oxc_transformer/src/plugins/replace_global_defines.rs
+++ b/crates/oxc_transformer/src/plugins/replace_global_defines.rs
@@ -1,7 +1,7 @@
 use std::{cmp::Ordering, sync::Arc};
 
 use lazy_static::lazy_static;
-use oxc_allocator::Allocator;
+use oxc_allocator::{Address, Allocator, GetAddress};
 use oxc_ast::ast::*;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_parser::Parser;
@@ -220,12 +220,31 @@ pub struct ReplaceGlobalDefinesReturn {
 pub struct ReplaceGlobalDefines<'a> {
     allocator: &'a Allocator,
     config: ReplaceGlobalDefinesConfig,
+    /// Since `Traverse` did not provide a way to skipping visiting sub tree of the AstNode,
+    /// Use `Option<Address>` to lock the current node when it is `Some`.
+    /// during visiting sub tree, the `Lock` will always be `Some`, and we can early return, this
+    /// could acheieve same effect as skipping visiting sub tree.
+    /// When `exit` the node, reset the `Lock` to `None` to make sure not affect other
+    /// transformation.
+    ast_node_lock: Option<Address>,
 }
 
 impl<'a> Traverse<'a> for ReplaceGlobalDefines<'a> {
     fn enter_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
-        self.replace_identifier_defines(expr, ctx);
-        self.replace_dot_defines(expr, ctx);
+        if self.ast_node_lock.is_some() {
+            return;
+        }
+        let is_replaced =
+            self.replace_identifier_defines(expr, ctx) || self.replace_dot_defines(expr, ctx);
+        if is_replaced {
+            self.ast_node_lock = Some(expr.address());
+        }
+    }
+
+    fn exit_expression(&mut self, node: &mut Expression<'a>, _ctx: &mut TraverseCtx<'a>) {
+        if self.ast_node_lock == Some(node.address()) {
+            self.ast_node_lock = None;
+        }
     }
 
     fn enter_assignment_expression(
@@ -233,13 +252,30 @@ impl<'a> Traverse<'a> for ReplaceGlobalDefines<'a> {
         node: &mut AssignmentExpression<'a>,
         ctx: &mut TraverseCtx<'a>,
     ) {
-        self.replace_define_with_assignment_expr(node, ctx);
+        if self.ast_node_lock.is_some() {
+            return;
+        }
+        if self.replace_define_with_assignment_expr(node, ctx) {
+            // `AssignmentExpression` is stored in a `Box`, so we can use `from_ptr` to get
+            // the stable address
+            self.ast_node_lock = Some(Address::from_ptr(node));
+        }
+    }
+
+    fn exit_assignment_expression(
+        &mut self,
+        node: &mut AssignmentExpression<'a>,
+        _: &mut TraverseCtx<'a>,
+    ) {
+        if self.ast_node_lock == Some(Address::from_ptr(node)) {
+            self.ast_node_lock = None;
+        }
     }
 }
 
 impl<'a> ReplaceGlobalDefines<'a> {
     pub fn new(allocator: &'a Allocator, config: ReplaceGlobalDefinesConfig) -> Self {
-        Self { allocator, config }
+        Self { allocator, config, ast_node_lock: None }
     }
 
     pub fn build(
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
* **#7621** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7621?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

IWANABETHATGUY:
Since `TraverseCtxt` has no functionality to skip visit sub tree(Fix me if I am wrong, I don't find the usage in codebase).
Need to lock the sub tree if one ast node is replaced by some value. 
TODO: 
if overlap appears in `AssignmentExpr` the output should still be wrong, since we can't get address of `AssignmentExpression` for now. 

@overlookmotel any chance to implement `ToAddress` for `AssignmentExpression` ?

overlookmotel:
> Since `TraverseCtxt` has no functionality to skip visit sub tree(Fix me if I am wrong, I don't find the usage in codebase). 

Correct. `Traverse` has no facility to skip visiting sub tree at present. We could look at adding that, but it might cause quite a performance regression, as it'd introduce an extra check and branch after every single AST node.

> TODO: if overlap appears in `AssignmentExpr` the output should still be wrong, since we can't get address of `AssignmentExpression` for now.
> 
> @overlookmotel any chance to implement `ToAddress` for `AssignmentExpression` ?

I'm not familiar with the define plugin, so not sure what you mean by "overlap". Can you give an example showing the problem with `AssignmentExpression`?

IWANABETHATGUY:
> > Since `TraverseCtxt` has no functionality to skip visit sub tree(Fix me if I am wrong, I don't find the usage in codebase).
> 
> Correct. `Traverse` has no facility to skip visiting sub tree at present. We could look at adding that, but it might cause quite a performance regression, as it'd introduce an extra check and branch after every single AST node.
> 
> The way you are handling it here seems like a good option in meantime, especially as the subtree is likely small, so not a great cost to visiting it.
> 
> > TODO: if overlap appears in `AssignmentExpr` the output should still be wrong, since we can't get address of `AssignmentExpression` for now.
> > @overlookmotel any chance to implement `ToAddress` for `AssignmentExpression` ?
> 
> I'm not familiar with the define plugin, so not sure what you mean by "overlap". Can you give an example showing the problem with `AssignmentExpression`?

Here is one, https://esbuild.github.io/try/#dAAwLjI0LjAAewogIGRlZmluZTogewogICAgJ2ltcG9ydC5tZXRhLmVudi5GT08nOiAnaW1wb3J0Lm1ldGEuZW52LkZPTycsCiAgICAnaW1wb3J0Lm1ldGEuZW52JzogJ19fZm9vX18nLAogIH0sCn0ACmNvbnNvbGUubG9nKAoJW2ltcG9ydC5tZXRhLmVudi5GT08gPSAwXQopCg, you could see 
both 
`import.meta.env.FOO` and `import.meta.env` matched part of  `import.meta.env.FOO`
for esbuild it will replace `import.meta.env.FOO` to `import.meta.env.FOO` and skip visit sub tree. 

but for us, 
we first replace `import.meta.env.FOO` to `import.meta.env.FOO`, then visit the child `import.meta.env`, and replace it to 
`__foo__`, so the final output should be `__foo__.FOO`

crates/oxc_transformer/src/plugins/replace_global_defines.rs
```diff
@@ -220,26 +220,55 @@ pub struct ReplaceGlobalDefinesReturn {
...
         &mut self,
         node: &mut AssignmentExpression<'a>,
         ctx: &mut TraverseCtx<'a>,
     ) {
-        self.replace_define_with_assignment_expr(node, ctx);
+        if self.ast_node_lock.is_some() {
+            return;
+        }
+        if self.replace_define_with_assignment_expr(node, ctx) {
+            // `AssignmentExpression` is always stored in a `Box`, since it came from Traverse Visitor
```
overlookmotel, line 253: It is true that `AssignmentExpression` is always stored in a `Box`, but the reason for that isn't that it came from the `Traverse` visitor. The reason is that the only place `AssignmentExpression` appears in AST types is in:

```rs
enum Expression<'a> {
  // ...
  AssignmentExpression(Box<'a, AssignmentExpression<'a>>),
  // ...
}
```

This doesn't generalize to all types that come from the `Traverse` visitor. e.g. `IdentifierName` isn't always stored in a `Box` (e.g. in `StaticMemberExpression`).

Not trying to lecture you! Just explaining, as I know you're interested in using `GetAddress` in Rolldown, and it's useful to know its limitations/gotchas.

crates/oxc_transformer/tests/integrations/plugins/replace_global_defines.rs
```diff
@@ -68,7 +67,22 @@ fn dot_with_overlap() {
     ])
     .unwrap();
     test("import.meta.env", "__foo__", config.clone());
-    test("import.meta.env.NODE_ENV", "import.meta.env.NODE_ENV", config.clone());
```
overlookmotel, line 71: Shouldn't we keep this test, but change it's output?

```rs
test("import.meta.env.NODE_ENV", "__foo__.NODE_ENV", config.clone());
```

crates/oxc_transformer/tests/integrations/plugins/replace_global_defines.rs
```diff
@@ -68,7 +67,22 @@ fn dot_with_overlap() {
...
+)
+        ",
+        "
+
+console.log(
+	[import.meta.env.FOO = 0]
+)
+",
+        config.clone(),
+    );
```
overlookmotel, line 85: This test seems overly complicated, and doesn't cover all cases.

```suggestion
    test("import.meta.env = 0", "__foo__ = 0");
    test("import.meta.env.NODE_ENV = 0", "__foo__.NODE_ENV = 0");
    test("import.meta.env.FOO = 0", "import.meta.env.FOO = 0");
```

crates/oxc_transformer/src/plugins/replace_global_defines.rs
```diff
@@ -220,26 +220,55 @@ pub struct ReplaceGlobalDefinesReturn {
...
 impl<'a> Traverse<'a> for ReplaceGlobalDefines<'a> {
     fn enter_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
-        self.replace_identifier_defines(expr, ctx);
-        self.replace_dot_defines(expr, ctx);
+        if self.ast_node_lock.is_some() {
+            return;
+        }
+        let mut flag = self.replace_identifier_defines(expr, ctx);
+        flag |= self.replace_dot_defines(expr, ctx);
+        if flag {
```
overlookmotel, line 233:
```suggestion
        let is_replaced = self.replace_identifier_defines(expr, ctx)
            || self.replace_dot_defines(expr, ctx);
        if is_replaced {
```

overlookmotel (changes requested):
Apart from the comments I've made, looks good.

But could you please add some comments about what `ast_node_lock` is and how it works?

crates/oxc_transformer/src/plugins/replace_global_defines.rs
IWANABETHATGUY: thanks

crates/oxc_transformer/tests/integrations/plugins/replace_global_defines.rs
IWANABETHATGUY: Yes

crates/oxc_transformer/tests/integrations/plugins/replace_global_defines.rs
IWANABETHATGUY: Done

overlookmotel (approved):

overlookmotel:
### Merge activity

* **Dec 6, 10:26 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7575, opened 2024-12-02
feat(transformer/class-properties): transform `ChainExpression`

Dunqing:
Fully support transforming ChainExpression in class properties. Our implementation has some differences in execution order compared to Babel, but we have passed all execution-related tests.

PR #7652, opened 2024-12-04
perf(transformer/class-properties): replace recursion with loop

overlookmotel:
Follow-on after #7575. Small optimization. Replace recursive function calls with a loop.

### DIFF
--- a/crates/oxc_transformer/src/es2022/class_properties/private.rs
+++ b/crates/oxc_transformer/src/es2022/class_properties/private.rs
@@ -1216,20 +1216,35 @@ impl<'a, 'ctx> ClassProperties<'a, 'ctx> {
     /// Recursively check if the expression has optional expression.
     #[inline]
     fn has_optional_expression(expr: &Expression<'a>) -> bool {
-        match expr {
-            Expression::CallExpression(call) => {
-                call.optional || Self::has_optional_expression(call.callee.get_inner_expression())
+        let mut expr = expr;
+        loop {
+            match expr {
+                Expression::CallExpression(call) => {
+                    if call.optional {
+                        return true;
+                    }
+                    expr = call.callee.get_inner_expression();
+                }
+                Expression::StaticMemberExpression(member) => {
+                    if member.optional {
+                        return true;
+                    }
+                    expr = &member.object;
+                }
+                Expression::ComputedMemberExpression(member) => {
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#7652** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

camc314 (approved):
For my own self developement, are there any good resources on why this is faster? 🙂

overlookmotel:
> For my own self developement, are there any good resources on why this is faster? 🙂

I don't have a go-to article or anything on this. But the general principle is that a function recursively calling itself increases the call stack. That is:

1. More costly than a simple loop in terms of CPU ops, register use, and cache utilization.
2. Somewhat dangerous, because you can exhaust the stack space and crash if recursion goes very deep (unlikely in practice, it'd have to be *really* weird code to trigger that).

Clever compilers will convert recursive functions to loops anyway ("tail call optimization"), but often the function has to be quite simple, or written in a certain way, for the compiler to be able to see that optimization is possible. So *maybe* compiler would have done this itself anyway here, maybe not.

I believe that in principle *any* recursive algorithm can be rewritten as a loop, but with complex algorithms often the way you have to write the code to achieve that makes it pretty much unreadable. But on a fairly simple functions like this where it doesn't make the code completely incomprehensible, personally I think it's worthwhile doing.

Here's a rather silly example showing the perf impact, from a conversation I was having with someone the other week about exactly this. Flattening an array of arrays in JS: https://x.com/onlyspaceghost/status/1859398231915925988 The loopy version is 1310x faster!

If you're interested in learning more, search for "loop vs recursion" or "tail call optimization".

camc314:
🙏 thank you
that example is crazy aha

Dunqing:
### Merge activity

* **Dec 4, 7:58 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7664, opened 2024-12-05
refactor(transformer/class-properties): use `duplicate_object` in `transform_expression_to_wrap_nullish_check`

Dunqing:
The `duplicate_object` is great, it has handled unbound identifiers and `this` expressions, so we can remove a lot of code. But it is still a little bit annoying because we need two `object` here,  although we can use `clone_in` it needs a special logic for `Expression::Identifier`.

### DIFF
--- a/crates/oxc_transformer/src/es2022/class_properties/private.rs
+++ b/crates/oxc_transformer/src/es2022/class_properties/private.rs
@@ -3,7 +3,7 @@
 
 use std::mem;
 
-use oxc_allocator::Box as ArenaBox;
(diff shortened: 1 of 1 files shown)


overlookmotel (approved):
Rebased on main and made a couple of small tweaks.

overlookmotel:
### Merge activity

* **Dec 5, 12:41 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7697, opened 2024-12-06
feat(transformer/class-properties): support for transforming `AssignmentTarget`

Dunqing:
Instance prop:
 * `[object.#prop] = arr` -> `[_toSetter(_classPrivateFieldSet2, [_prop, object])._] = arr`
 * `({x: object.#prop} = obj)` -> `({ x: _toSetter(_classPrivateFieldSet2, [_prop, object])._ } = obj)`

 Static prop:
 * `[object.#prop] = arr` -> `[_assertClassBrand(Class, object, _prop)._] = arr`
 * `({x: object.#prop} = obj)` -> `({ x: _assertClassBrand(Class, object, _prop)._ } = obj)`

PR #7700, opened 2024-12-06
refactor(transformer/class-properties): shorten code

overlookmotel:
Follow-on after #7697. No need to cast twice. You can go from `MemberExpression` direct to `AssignmentTarget`.

PR #7701, opened 2024-12-06
refactor(transformer/class-properties): move code out of `transform_assignment_target`

overlookmotel:
Follow-on after #7697.

`transform_assignment_target` is inlined into `enter_assignment_target` visitor, so we want it to be as small as possible. Move assigning to `target` into `transform_assignment_target_impl`, which is the cold path.

The principle is that in the transformer the most important thing for performance is to optimize for the path which is "nothing to do here, exit".

Here we are only transforming `object.#prop`, but very few `AssignmentTarget`s are `object.#prop`, so 99% of the time there is nothing to do. So we want to keep that "do nothing and exit" path as fast and small as possible.

In practice, the `*target =` assignment is only a single assembly operation, so this PR is a micro-optimization. But why not? Every little helps.

### DIFF
--- a/crates/oxc_transformer/src/es2022/class_properties/private.rs
+++ b/crates/oxc_transformer/src/es2022/class_properties/private.rs
@@ -1432,7 +1432,7 @@ impl<'a, 'ctx> ClassProperties<'a, 'ctx> {
         // `object.#prop` in assignment pattern.
         // Must be in assignment pattern, as `enter_expression` already transformed `AssignmentExpression`s.
         if matches!(target, AssignmentTarget::PrivateFieldExpression(_)) {
-            *target = self.transform_assignment_target_impl(target, ctx);
+            self.transform_assignment_target_impl(target, ctx);
         }
     }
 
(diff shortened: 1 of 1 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7701?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7701** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7701?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7701?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7700** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

Dunqing (approved):
Fair enough!

Issue #7696, opened 2024-12-06
linter: the rule `import/namespace` doesn't handle external dependencies correctly


shulaoda:
### What version of Oxlint are you using?

0.14.1

### What does your `.oxlint.json` config file look like?

```json
{
  "$schema": "./node_modules/oxlint/configuration_schema.json",
  "plugins": ["import"],
  "rules": {
    "import/namespace": [
      "error",
      {
        "allowComputed": true
      }
    ]
  }
}
```

### What happened?

Before this, we need to install the dependency `acorn`.
```js
import * as acorn from 'acorn'; acorn.parse()
```

```bash
  ⚠ eslint-plugin-import(namespace): "parse" not found in imported namespace "acorn".
   ╭─[index.js:1:39]
 1 │ import * as acorn from 'acorn'; acorn.parse()
   ·                                       ─────
   ╰────
```

PR #7723, opened 2024-12-08
test(linter): add regression tests for `import/namespace`

shulaoda:
Related to #7696


Boshen:
I didn't add tests because this is just testing the behavior of the resolver, which is already covered by the resolver itself.

PR #7737, opened 2024-12-09
feat(napi/transform): return helpers information

### LINKED ISSUE #7599: [transformer] bundle runtime helpers / remove runtime helper import code from output
I'm trying to use the transform api as a replacement of `esbuild.transform`. When esbuild needs to use a helper function for lowering, it injects them at the top of the output ([esbuild try](https://esbuild.github.io/try/#dAAwLjI0LjAALS10YXJnZXQ9ZXMyMDE3AGNvbnN0IGJhciA9IHsuLi4iYmIifQ)).
On the other hand, oxc injects a `import` or `require` for the helper function.

This makes it difficult to use it as an replacement for my case. I'm running esbuild in `renderChunk` hook of rollup and at that time the chunking has already done. I have to analyze the output and run a different build to get the helper functions and connect the chunks.

If oxc has a flag that injects the helper function into the output, it will be really easy to swap esbuild with oxc.

Or if oxc has a flag that makes the output to not include the `import`/`require`, I won't need to analyze the output.
An example API looks like this:
```js
const result = transformWithOxc(name, 'const bar = {..."bb"}', {
  target: 'es2017',
  sourceType: 'module',
  helpers: 'external' // proposed flag
})

console.log(result.helpers) // { '_objectSpread': '@babel/runtime/helpers/objectSpread2' }
```
Then, I can know which helpers are needed without analyzing the output.

Boshen:
closes #7599

### DIFF
--- a/crates/oxc_transformer/src/common/helper_loader.rs
+++ b/crates/oxc_transformer/src/common/helper_loader.rs
@@ -156,7 +156,7 @@ pub enum Helper {
 }
 
 impl Helper {
-    const fn name(self) -> &'static str {
+    pub const fn name(self) -> &'static str {
         match self {
             Self::AwaitAsyncGenerator => "awaitAsyncGenerator",
             Self::AsyncGeneratorDelegate => "asyncGeneratorDelegate",
@@ -184,6 +184,7 @@ pub struct HelperLoaderStore<'a> {
     mode: HelperLoaderMode,
     /// Loaded helpers, determined what helpers are loaded and what imports should be added.
     loaded_helpers: RefCell<FxHashMap<Helper, BoundIdentifier<'a>>>,
+    pub(crate) used_helpers: RefCell<FxHashMap<Helper, String>>,
 }
 
 impl<'a> HelperLoaderStore<'a> {
@@ -192,6 +193,7 @@ impl<'a> HelperLoaderStore<'a> {
             module_name: options.module_name.clone(),
             mode: options.mode,
             loaded_helpers: RefCell::new(FxHashMap::default()),
+            used_helpers: RefCell::new(FxHashMap::default()),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Boshen:
* **#7737** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7737?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

Boshen (commented):
need review

Boshen:
### Merge activity

* **Dec 9, 4:35 AM EST**: [Graphite](https://app.graphite.dev/github/pr/oxc-project/oxc/7737) disabled "merge when ready" on this PR due to: a merge conflict with the target branch; resolve the conflict and try again..

overlookmotel:
I think this needs a separate option `HelperLoaderMode::Global`. As I understand it from #7599, the requirement is:

1. Do NOT generate `import` / `require`.
2. Create a unique symbol name for the helper function.
3. Helper usage includes that symbol name e.g. `_defineProperty(...)`.
4. Return object mapping symbol names to import paths.

This is different from the existing modes:

* `Runtime` mode inserts `import`/`require`.
* `External` mode doesn't create a symbol name for each helper - it outputs `babelHelpers.defineProperty(...)` not `_defineProperty(...)`.

`HelperLoaderMode::Global` should use `TraverseCtx::generate_uid_name` to generate symbol name, instead of `generate_uid_in_root_scope`. This will create a unique name and add it to set of used UIDs so it doesn't get used again, but will NOT create a binding. The `IdentifierReference` returned should be unbound (i.e. global reference).

PR #7742, opened 2024-12-09
feat(linter): support `expectTypeOf`, `assert` and `assertType` in `vitest/expect-expect`

baseballyama:
[Vitest](https://vitest.dev/api/expect.html) has 4 assertions. [expect](https://vitest.dev/api/expect), [expectTypeOf](https://vitest.dev/api/expect-typeof), [assert](https://vitest.dev/api/assert) and [assertType](https://vitest.dev/api/assert-type).

But now only `expect`. Therefore this PR supports there rest of assertions.
And jest doesn't have such assertions, so I added branching based on the test framework.


camc314 (approved):
thank you! 🙏

PR #7745, opened 2024-12-09
fix(transformer/logic-assignment-operator): always create `IdentifierReference`s with `ReferenceId`

overlookmotel:
Use `TransformCtx::duplicate_expression` (introduced in #7754) to decide when to create temp vars for member expression object and computed property.

This fixes a bug where `IdentifierReference`s created when transforming `key` in `object[key] &&= value` were created without a `ReferenceId` (due to `clone_in`).

We didn't catch this before because Babel's test fixtures only cover `object[key++] &&= value` not the simpler `object[key] &&= value`. Add tests for this.


overlookmotel:
Ah ha no! I made a mistake in the test. The problem *is* in this transform. I'll try to fix.

overlookmotel:
@Dunqing Could you please take a look and make sure I'm not missing anything? I *think* this simplification of the code retains its original behavior, but I don't know this transform, so I might have made a mistake.

overlookmotel:
Ah shit. I did miss something, and Babel doesn't have enough tests to catch me! Will fix...

Dunqing (approved):

PR #7746, opened 2024-12-09
perf(traverse): faster `TraverseScoping::is_static`

overlookmotel:
Speed up `TraverseScoping::is_static` by checking if symbol is `const`. That check is faster than iterating through all references to see if they write to the binding.

This also subtly changes behavior. `is_static(x)` will now return `true` for `x` in `const x = 1; x = 2;`. But that's valid, because the purpose of `is_static` is to say whether the value of the var can be mutated, which it can't if it's `const`.

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -378,13 +378,14 @@ impl TraverseScoping {
     pub fn is_static(&self, expr: &Expression) -> bool {
         match expr {
             Expression::ThisExpression(_) | Expression::Super(_) => true,
-            Expression::Identifier(ident) => {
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#7746** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7746?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

overlookmotel:
0.1% perf boost. Woo hoo!

Dunqing (approved):

overlookmotel:
Actually, I'm refactoring things. This method can be deleted entirely. I'll do that in another PR.

PR #7771, opened 2024-12-10
chore(tasks/transform-conformance): support `override` to replace `takeover` mode

Dunqing:
This PR does the following things.

1. Move the override output of the `snapshots` folder to the `overrides` folder.
2. Support `override` mode to replace `takeover` mode
3. The `update_fixtures.js` no longer uses `overrides`'s `output.js` to replace Babel's `output.js`.

### How does `override` mode work?

￼When running each test, it checks whether an output file for that test exists in the ⁠`overrides` directory. If it does, the output file will be used to compare with the transformed code.


Dunqing:
* **#7771** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

overlookmotel:
I see one problem. There are 3 overrides for class properties transform where the `options.json` file is overridden too. To replace `update_fixture.js`, the new mechanism would need to look for `options.json` files in the `overrides` folder too, and use that as the source of truth if there is one (note: required for `exec` tests too).

This explains the difference in snapshots for class properties transform in this PR.

2 of the overrides for class properties transform have no `output.js` files in them, but I've changed that in #7777. If you want to rebase on top of that, it'll solve that particular problem.

What explains the difference in snapshots for the other transforms, I cannot say. Looks like something's not working!

overlookmotel:
PS I can clean up `update_fixtures.js` to remove dead code after this PR is merged. Probably faster for me to do it, as I'm familiar with it.

Dunqing:
> What explains the difference in snapshots for the other transforms, I cannot say. Looks like something's not working!


Looks like the `takeover` mode hasn't stored any information in snapshots, we even don't know if the semantic check is passed or not, now we does.

overlookmotel:
An unwelcome discovery! 😭

But yes better to know than not to know.

But every single test for object-rest-spread is now "Output mismatch"!

Dunqing:
> But every single test for object-rest-spread is now "Output mismatch"! I guess that's where `--override` comes in. Going through them all and checking our output is actually correct is going to be fun...

Fixed! By using `--override` 😂.

Dunqing:
The `private-loose/foobar/input.js` test still mismatch, can you take a look more?

overlookmotel:
> The `private-loose/foobar/input.js` test still mismatch, can you take a look more?

Looks like the problem is that it's not looking for `options.json` in `overrides` folder and using that over the original Babel one if it's there. i.e. Currently you can override `output.js`, but not `options.json`.

Sorry if I was unclear above. #7777 solved the problem of when `overrides` folder contains *only* an `options.json` file and no `output.js` file (by just adding `output.js` files to those overrides). But still `options.json` in `overrides` needs to be obeyed!

Dunqing:
> > The `private-loose/foobar/input.js` test still mismatch, can you take a look more?
> 
> Looks like the problem is that it's not looking for `options.json` in `overrides` folder and using that over the original Babel one if it's there. i.e. Currently you can override `output.js`, but not `options.json`.
> 
> Sorry if I was unclear above. #7777 solved the problem of when `overrides` folder contains _only_ an `options.json` file and no `output.js` file (solved by just adding `output.js` files to those override folders). But still `options.json` in `overrides` needs to be obeyed!

Solved! Thank you for finding the root cause!

overlookmotel (approved):
I rebased this on top of #7807, so it's easier to see that all the old takeover snapshots have just moved to overrides, and this change hasn't actually altered the `output.js` files.

PR #7796, opened 2024-12-11
fix(linter): Fix unicorn/prefer-query-selector to use the correct replacement for getElementsByClassName

### LINKED ISSUE #7794: prefer-query-selector fixer is incorrect and breaks code
the fixer here is not correct, the arguments are different for `getElementsByClassName` and `querySelectorAll`.  One is a list of class names and the other is a query selector.

_Originally posted by @ezpuzz in https://github.com/oxc-project/oxc/issues/1068#issuecomment-2533449188_

nrayburn-tech:
Note: This uses a regex to replace multiple instances of whitespace with ` .`.  May not be the most performant, so if there's a simple alternative I can change to that instead.

cc @camc314, I know this was assigned to you but I just wanted to throw something quick together while I had a minute.  Feel free to use this, or decline it and write your own.

Fixes #7794.

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/prefer_query_selector.rs
+++ b/crates/oxc_linter/src/rules/unicorn/prefer_query_selector.rs
@@ -3,7 +3,6 @@ use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::{GetSpan, Span};
 use phf::phf_map;
-use regex::Regex;
 
 use crate::{context::LintContext, rule::Rule, utils::is_node_value_not_dom_node, AstNode};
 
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/unicorn/prefer_query_selector.rs
```diff
@@ -127,9 +128,19 @@ impl Rule for PreferQuerySelector {
                     // argument_expr.span().source_text(ctx.source_text());
                     let source_text = fixer.source_range(argument_expr.span());
                     let quotes_symbol = source_text.chars().next().unwrap();
-                    let sharp = if cur_property_name.eq(&"getElementById") { "#" } else { "" };
+                    let argument = match *cur_property_name {
+                        "getElementById" => format!("#{literal_value}"),
+                        "getElementsByClassName" => format!(
+                            ".{}",
+                            Regex::new(r"\s+").unwrap().replace_all(literal_value, " .")
```
Boshen, line 135: We should avoid `Regex` at all cost, a quick split and then filter out whitespace should also work.

nrayburn-tech: I’ll change this when I get a chance, if somebody else gets to it sooner feel free to close this out.
Marking as a draft for now.

camc314 (approved):
thank you!

perf regression is unrelated

PR #7799, opened 2024-12-11
refactor(transformer): handle `<CWD>` in test runner

Dunqing:
We should move the handling of  `<CWD>` to the test runner because this is just only used in testing, and it causes us always get a path by `self.ctx.source_path` like `<CWD>/xxx/xxx.js`, we should get a real path for this.

### DIFF
--- a/crates/oxc_transformer/src/context.rs
+++ b/crates/oxc_transformer/src/context.rs
@@ -52,14 +52,10 @@ impl<'a> TransformCtx<'a> {
             .file_stem() // omit file extension
             .map_or_else(|| String::from("unknown"), |name| name.to_string_lossy().to_string());
 
-        let source_path = source_path
-            .strip_prefix(&options.cwd)
-            .map_or_else(|_| source_path.to_path_buf(), |p| Path::new("<CWD>").join(p));
-
         Self {
             errors: RefCell::new(vec![]),
             filename,
-            source_path,
+            source_path: source_path.to_path_buf(),
             source_type: SourceType::default(),
             source_text: "",
             module: options.env.module,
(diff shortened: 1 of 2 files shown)


Dunqing:
* **#7799** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7799?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7799?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

overlookmotel:
I would be very happy to see this change happen. The path manipulation in `TransformCtx::new` is one of the largest sources of variance in our transformer benchmarks, so removing it will help a lot with that.

But... just to check, does not stripping `options.cwd` from start of source path affect the output of JSX Source transform, which uses `source_path`?

overlookmotel:
> Yes, it will.

Is that not a problem?

Dunqing:
> > Yes, it will.
> 
> Is that not a problem?

It is not a problem, we should always get a real path except for testing

overlookmotel:
Agreed we can remove the special case for `<CWD>`, but doesn't Babel remove `cwd` from start of the path?

i.e.:

```js
var _jsxFileName = "/src/foo/bar.jsx";
```

not full path:

```js
var _jsxFileName = "/Users/bozo/projects/clownhouse-app/src/foo/bar.jsx";
```

overlookmotel:
OK great. Thanks for bearing with all my questions!

overlookmotel (approved):

overlookmotel:
### Merge activity

* **Dec 12, 12:21 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 12, 12:21 PM EST**: A user added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #7812, opened 2024-12-12
fix(semantic): missing references when `export {}` references a type-only binding and a normal

hyfdev:
https://playground.oxc.rs/#eNpdTzEOgzAM/AryzFAqdUHq2rlD1YklUIMihTiyTQGh/L0BBEMn31l35/MCDZTQkBfNXmRE3xbHQKzZPSsqX3mcNrZkOgfcJU+mIPmfPEIOBOUCPPh1yOzVTFAqD5iDs14PLA0FPMnc1+QOpmy8tMT9vog5BMOCnAITXkMOfCq3BajhDtMFQLleihskRUMf7HDtkkhvvW3tYU7vKpN7OBpX8xe5JkmVWuMEY4w/P71iMw==

Current:

image

Expected:

https://babeljs.io/repl#?browsers=&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=MYewdgzgLgBAKiAhtAagSwKYHcAOIBOsAvDAIwBQ5UAnjhvEtAAr4g4QwkXkYAeehGAG8YNOg2RQWbCABoJqTLgKwAvkA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ctypescript&prettier=false&targets=&version=7.26.4&externalPlugins=&assumptions=%7B%7D

image


crates/oxc_semantic/src/builder.rs
```diff
@@ -2016,9 +2013,7 @@ impl<'a> SemanticBuilder<'a> {
                 self.class_table_builder.pop_class();
             }
             AstKind::ExportSpecifier(_) => {
-                if !self.current_reference_flags.is_type_only() {
-                    self.current_reference_flags = ReferenceFlags::empty();
-                }
+                self.current_reference_flags = ReferenceFlags::empty();
```
hyf0, line 2016: Should I use

```
s.export_kind.is_type()
                    || matches!(self.nodes.parent_kind(self.current_node_id), Some(AstKind::ExportNamedDeclaration(decl)) if decl.export_kind.is_type())
```

and only `self.current_reference_flag-=ReferenceFlags::Type`.

crates/oxc_semantic/src/builder.rs
```diff
@@ -1845,13 +1845,10 @@ impl<'a> SemanticBuilder<'a> {
         /* cfg */
 
         match kind {
-            AstKind::ExportNamedDeclaration(decl) => {
```
hyf0, line 1848: Should I also remove `AstKind::ExportNamedDeclaration` in leave kind?

crates/oxc_semantic/src/builder.rs
Dunqing: We may need to handle this in `visit_export_specifier`; I think checking the parental `ExportNamedDeclaration` every time here is inefficient. 

You can refer https://github.com/oxc-project/oxc/blob/877a83a81bdf93dcb890df174132cfff381a58d1/crates/oxc_semantic/src/builder.rs#L1777-L1812 to do that

hyf0: Yeah. I thought of this too. But it seems can't be done to check `if previous item is a type` to determine if this item is a type. You have to check it's parent.

Maybe we could store the check result of ` parental ExportNamedDeclaration`? But that would lead to add a extra field. To reduce changes, I didn't do it in the first place.

hyf0:
All tests pass now. I handle this PR over to you, feel free to add necessary changes.

crates/oxc_semantic/src/builder.rs
Dunqing: It seems to work as well if you revert this and remove `|| matches!(self.nodes.parent_kind(self.current_node_id), Some(AstKind::ExportNamedDeclaration(decl)) if decl.export_kind.is_type())` in `ExportSpecifier`

hyf0: Feel free to modify the PR directly if you think it's more efficient.

Dunqing:
Similar to #7183, waiting for a solution from it

Dunqing:
Just formatted comments and added two tests. Merging it, I will make a small refactor for it later.

Dunqing:
### Merge activity

* **Dec 12, 9:25 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Dunqing (approved):

PR #7815, opened 2024-12-12
feat(transformer/class-properties): transform super member expressions that are inside static prop initializer

Dunqing:
This PR support for transforming `super.prop` to  `babelHelpers.superPropGet(_B, "prop", _B)`

Input:
```js
class A {
  static prop = 1;
}

class B extends A {
  static prop = 2;
  static propA = super.prop;
  static getPropA = () => super.prop;
}
```

Output:
```js
var _B;
class A {}
babelHelpers.defineProperty(A, "prop", 1);
class B extends A {}
_B = B;
babelHelpers.defineProperty(B, "prop", 2);
babelHelpers.defineProperty(B, "propA", babelHelpers.superPropGet(_B, "prop", _B));
babelHelpers.defineProperty(B, "getPropA", () => babelHelpers.superPropGet(_B, "prop", _B));
```

### DIFF
--- a/crates/oxc_transformer/src/es2022/class_properties/supers.rs
+++ b/crates/oxc_transformer/src/es2022/class_properties/supers.rs
@@ -1,0 +1,86 @@
+//! ES2022: Class Properties
+//! Transform of `super` expressions.
+
+use oxc_ast::ast::*;
+use oxc_traverse::TraverseCtx;
+
+use crate::Helper;
+
+use super::ClassProperties;
+
+impl<'a, 'ctx> ClassProperties<'a, 'ctx> {
+    /// Transform static member expression where object is `super`.
+    ///
+    /// `super.prop` -> `_superPropGet(_Class, "prop", _Class)`
+    //
+    // `#[inline]` so that compiler sees that `expr` is an `Expression::StaticMemberExpression`.
+    #[inline]
+    pub(super) fn transform_static_member_expression(
+        &mut self,
+        expr: &mut Expression<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) {
+        let Expression::StaticMemberExpression(member) = expr else { unreachable!() };
+        if matches!(member.object, Expression::Super(_)) {
+            *expr = self.transform_static_member_expression_impl(member, ctx);
+        }
+    }
+
+    fn transform_static_member_expression_impl(
+        &mut self,
+        member: &mut StaticMemberExpression<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) -> Expression<'a> {
+        let property = &member.property;
+        let property = ctx.ast.expression_string_literal(
+            property.span,
+            property.name.clone(),
+            Some(property.name.clone()),
+        );
+        self.create_super_prop_get(member.span, property, ctx)
+    }
+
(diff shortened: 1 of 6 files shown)


crates/oxc_transformer/src/es2022/class_properties/supers.rs
```diff
@@ -0,0 +1,76 @@
...
+                ctx.ast.move_expression(&mut member.expression)
+            }
+            Expression::PrivateFieldExpression(_) => {
+                unreachable!("`super` cannot access private fields")
+            }
+            _ => return,
+        };
+
+        *expr = self.create_super_prop_get(expr.span(), property, ctx);
+    }
```
overlookmotel, line 58: It would be better to have 2 different visitors here for `StaticMemberExpression` and `ComputedMemberExpression`. Currently you funnel both through `transform_member_expression` and then you branch on what kind of `Expression` it is again in `transform_member_expression_impl`. This is unnecessary work.

crates/oxc_transformer/src/es2022/class_properties/supers.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    // `_superPropGet(_classBinding, "prop", _classBinding)`
+    fn create_super_prop_get(
+        &mut self,
+        span: Span,
+        property: Expression<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) -> Expression<'a> {
+        let class_binding = self.class_bindings.get_or_init_temp_binding(ctx);
```
overlookmotel, line 67: This needs to use `ClassProperties::get_temp_binding` instead (defined in `static_prop_init.rs`, you'll need to make that method `pub(super)`).

```suggestion
        let class_binding = self.get_temp_binding(ctx);
```

It's really annoying we have to do this, and it's not at all obvious. The problem is that we store `class_bindings` in 2 places and which is the "source of truth" depends on whether the class has private properties or not. This is very bad design (my fault), but I couldn't find a better way at present.

overlookmotel (changes requested):

crates/oxc_transformer/src/es2022/class_properties/supers.rs
```diff
@@ -0,0 +1,76 @@
...
+use oxc_ast::ast::*;
+use oxc_span::GetSpan;
+use oxc_traverse::TraverseCtx;
+
+use crate::Helper;
+
+use super::ClassProperties;
+
+impl<'a, 'ctx> ClassProperties<'a, 'ctx> {
+    /// Transform member expression where object is a super.
```
overlookmotel, line 13: Nit:

```suggestion
    /// Transform member expression where object `super`.
```

overlookmotel (changes requested):

crates/oxc_transformer/src/es2022/class_properties/supers.rs
Dunqing: Thanks for pointing this out, we should do that!

Dunqing:
Thank @overlookmotel for reviewing, now it more clearer!

overlookmotel (approved):
Pushed 4 nitty commits. @Dunqing If you're OK with them, please merge.

PR #7822, opened 2024-12-12
ci: add zizmor for github actions security

Sysix:
> zizmor is a static analysis tool for GitHub Actions. It can find many common security issues in typical GitHub Actions CI/CD setups.

https://woodruffw.github.io/zizmor/

EDIT: what is the right PR-syntax for this?

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,7 +1,7 @@
 name: Check PR
 
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - edited
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,7 +1,7 @@
 name: Check PR
 
 on:
-  pull_request:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     types:
       - opened
       - edited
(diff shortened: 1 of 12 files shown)


.github/workflows/prepare_release_oxlint.yml
```diff
@@ -10,11 +10,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
```
Sysix, line 13: I hope I did split them right and did not missed one on a job

.github/workflows/pr.yml
```diff
@@ -1,7 +1,7 @@
 name: Check PR
 
 on:
-  pull_request_target:
```
Sysix, line 4: https://woodruffw.github.io/zizmor/audits/#dangerous-triggers

Boshen: https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#event-triggers

Boshen: I'm in favor of removing of this check with something else, but not right now.

Boshen (approved):
Thank you for this, I've always wanted to harden the security, but the only thing I did was changing tokens to another account with less privileges.

PR #7860, opened 2024-12-13
feat(ast): add `AstBuilder::vec_convert`

Dunqing:
Give a try for https://github.com/oxc-project/backlog/issues/153

I am not sure whether it a 100% safe, we only can guarantee that we use it safely

### DIFF
--- a/crates/oxc_ast/src/ast_builder_impl.rs
+++ b/crates/oxc_ast/src/ast_builder_impl.rs
@@ -75,6 +75,26 @@ impl<'a> AstBuilder<'a> {
         Vec::from_array_in(array, self.allocator)
     }
 
+    /// Convert a [`Vec`] of one type to a [`Vec`] of another type.
+    ///
+    /// This method is useful where the `Src` is inherited from `Dst` or vice versa, so that you can it
+    /// without any runtime overhead. For example, you can convert a `Vec<'a, Expression>` to a
+    /// `Vec<'a, Argument>`
+    #[inline]
+    pub fn vec_convert<Src, Dst>(&self, vec: Vec<'a, Src>) -> Vec<'a, Dst>
+    where
+        Src: 'a,
+        Dst: 'a,
+    {
+        // Verify types are layout-compatible
+        debug_assert!(std::mem::size_of::<Src>() == std::mem::size_of::<Dst>());
+        debug_assert!(std::mem::align_of::<Src>() == std::mem::align_of::<Dst>());
+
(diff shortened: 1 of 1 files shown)


overlookmotel:
Sorry to say this is wildly unsound!

2 types being same size and alignment is not enough. They also need *to be* the same type.

The only reason why we can convert from `Expression` to `Argument` safely is because `Argument` is defined with the `inherit_variants!` macro. This makes `Argument` a *superset* of `Expression`, so any `Expression` is also a valid `Argument`.

* `Expression::BooleanLiteral` -> `Argument::BooleanLiteral`
* `Expression::ArrayExpression` -> `Argument::ArrayExpression`
* ...same for all `Expression`'s variants

The `inherit_variants!` macro generates code that checks at compile time that this is true, and that both `enum`s have same discriminants i.e. discriminant for `Expression::BooleanLiteral` is same as discriminant for `Argument::BooleanLiteral`, etc for all the variants.

https://github.com/oxc-project/oxc/blob/0804916b283e0dd2b26d5a84ce0142f4216ab180/crates/oxc_ast/src/ast/macros.rs#L858-L870

So there's a lot of checks that make sure this conversion is sound.

But it's *not* sound, for example, to go in the other direction - from `Argument` back to `Expression` - because `Argument::SpreadElement` has no equivalent in `Expression`.

And it's even more unsound to perform conversion between *any* types which have same size and alignment, which is what `vec_convert` allows.

`vec_convert` could convert from `Vec<u8>` to `Vec<bool>`, for example. But that's not valid. `0u8` -> `false` and `1u8` -> `true`, but what is the `bool` value that corresponds to `2`? Or `255`? There isn't one, so it's undefined behavior.

I think we *probably* can make a sound implementation of `Vec<Expression>` to `Vec<Argument>` conversion, but we should lock it down to only allow conversions which are valid, and codegen the code with either `inherit_variants!` macro or `ast_tools` to make sure we don't make mistakes.

Beyond soundness issues, there's also a question of whether allowing this conversion loses us anything else useful. At present, we have an invariant that any pointer to an element in a `Vec<Expression>` remains a valid `Expression` forever. but if you can convert it to a `Vec<Argument>`, then that invariant breaks. We don't *currently* make use of that invariant, but we may want to in future (e.g. `Traverse` v2). That's the part I'm not sure about.

Anyway... TLDR... it's complicated!

Dunqing:
> Anyway... TLDR... it's complicated!

Ah, closing it as you said!

PR #7881, opened 2024-12-14
fix(linter): false positive in `jsx-a11y/label-has-associated-control`

### LINKED ISSUE #7849: linter: false positives for label-has-associated-control

0.15.0 / main

### What command did you run?

oxlint --import-plugin --jsx-a11y-plugin -D label-has-associated-control

_No response_

### What happened?

In our projects we currently run eslint with this rule enabled but it doesn't warn us about these issues while with oxlint it does. 

```
  × eslint-plugin-jsx-a11y(label-has-associated-control): A form label must be associated with a control.
     ╭─[src/app.tsx:173:13]
 172 │         <Page>
 173 │             <FilesContext.Provider value={{ addAlert, cwdInfo }}>
     ·             ─────────────────────────────────────────────────────
 174 │                 <WithDialogs>
     ╰────
  help: Either give the label a `htmlFor` attribute with the id of the associated control, or wrap the label around the control.
```
Code: https://github.com/cockpit-project/cockpit-files/blob/13/src/app.tsx#L173

```
  ⚠ eslint-plugin-jsx-a11y(label-has-associated-control): A form label must have accessible text.
    ╭─[src/ContainerDetails.jsx:66:53]
 65 │                         <DescriptionListTerm>{_("Created")}</DescriptionListTerm>
 66 │                         <DescriptionListDescription><utils.RelativeTime time={container.Created} /></DescriptionListDescription>
    ·                                                     ───────────────────────────────────────────────
 67 │                     </DescriptionListGroup>
    ╰────
  help: Ensure the label either has text inside it or is accessibly labelled using an attribute such as `aria-label`, or `aria-labelledby`. You can mark more attributes as
accessible labels by configuring the `labelAttributes` option.
```

shulaoda:
closes #7849

This is essentially because `get_element_type` does not support recognizing tag names like `<a.b />`. I'm unsure whether we should support it.

### DIFF
--- a/crates/oxc_linter/src/rules/jsx_a11y/label_has_associated_control.rs
+++ b/crates/oxc_linter/src/rules/jsx_a11y/label_has_associated_control.rs
@@ -206,10 +206,12 @@ impl Rule for LabelHasAssociatedControl {
             return;
         };
 
-        if let Some(element_type) = get_element_type(ctx, &element.opening_element) {
-            if self.label_components.binary_search(&element_type.into()).is_err() {
-                return;
-            }
+        let Some(element_type) = get_element_type(ctx, &element.opening_element) else {
+            return;
+        };
+
+        if self.label_components.binary_search(&element_type.into()).is_err() {
+            return;
         }
 
         let has_html_for = has_jsx_prop(&element.opening_element, "htmlFor").is_some();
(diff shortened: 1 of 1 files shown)


Boshen:
I think it should if makes sense.

PR #7894, opened 2024-12-14
fix(linter): add option "allowTypeImports" for rule "no-restricted-imports"

### LINKED ISSUE #7810: linter: `no-restricted-imports` missing option support for `allowTypeImports`, `allowImportNames`, `group`, ..

0.15.0

### What command did you run?

--

---

### What happened?

Because this is originally only a eslint rule, it does not support the option of the alias rule [@typescript-eslint/no-restricted-imports](https://typescript-eslint.io/rules/no-restricted-imports/).

See the implementation here: https://github.com/oxc-project/oxc/blob/74bf141b1a0b27123479156264eed0924fe8c36d/crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs#L35-L40

I expect to use `allowTypeImports`  option

Sysix:
closes #7810


crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs
```diff
@@ -1832,6 +2600,270 @@ fn test() {
         // ),
     ];
 
+    let fail_typescript = vec![
+        ("import foo from 'import1';", Some(serde_json::json!(["import1", "import2"]))),
+        // ("import foo = require('import1');", Some(serde_json::json!(["import1", "import2"]))),
```
Sysix, line 2605: @camc314 should we support this?
Notice the `import` at the start. There is no entry in `ModuleRecord` for it.
I searched a bit on the internet and it looks like this is a old syntax. 

There is 2 options how to solve this:
- fix `ModuleRecord` to implement this use case 
- loops to the AST nodes and find the require

I am not that good in typescript to understand what the best solution for the absurd syntax

Sysix: the team worked on some semantic about this. So it is still needed. Need to rethink the implementation

crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs
```diff
@@ -2268,6 +2714,270 @@ fn test() {
...
+        ("export { foo } from 'import1';", Some(serde_json::json!(["import1", "import2"]))),
+        (
+            "import foo from 'import1';",
+            Some(serde_json::json!([{ "paths": ["import1", "import2"] }])),
+        ),
+        (
+            "export { foo } from 'import1';",
+            Some(serde_json::json!([{ "paths": ["import1", "import2"] }])),
+        ),
+        // (
```
Sysix, line 2729: `paths` and `patterns` combination is currently not supported. Looks like a typescript-eslint only feature.

crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs
```diff
@@ -503,7 +548,28 @@ impl RestrictedPattern {
...
+        &self,
+        literal: &StringLiteral,
+        is_type: bool,
+    ) -> ImportNameResult {
+        if is_type && self.allow_type_imports.is_some_and(|x| x) {
+            return ImportNameResult::Allowed;
+        }
+
+        let name = literal.value.into_compact_str();
+        let unused_name = &CompactStr::from("__<>import_name_that_cant_be_used<>__");
```
Sysix, line 561: reimplementing `is_name_span_allowed_in_pattern` sounds silly for me. This works too.

shulaoda: We can further improve it in the future.

crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs
```diff
@@ -493,7 +534,11 @@ impl RestrictedPattern {
...
 
-    fn get_export_name_result(&self, name: &ExportImportName) -> ImportNameResult {
+    fn get_export_name_result(&self, name: &ExportImportName, is_type: bool) -> ImportNameResult {
+        if is_type && self.allow_type_imports.is_some_and(|x| x) {
+            return ImportNameResult::Allowed;
+        }
+
         match &name {
             ExportImportName::Name(import) => {
                 match is_name_span_allowed_in_pattern(&import.name, self) {
```
shulaoda, line 544: Could we implement `is_name_span_allowed_in_pattern` and `is_name_span_allowed_in_path` within `RestrictedPattern` and `RestrictedPath`?

shulaoda (approved):
LGTM

PR #7919, opened 2024-12-15
feat(transformer/class-properties): do not create temp var for template literal computed key

overlookmotel:
Do not create temp var for computed key which is `TemplateLiteral` with no expressions. Evaluating such a template literal cannot have side effects.

```js
class C {
  [`foo`] = 1;
}
```

But this *can* have side effects:

```js
class C {
  [`foo${bar}`] = 1;
}
```


Dunqing (approved):

Dunqing:
### Merge activity

* **Dec 15, 11:54 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #7943, opened 2024-12-16
fix(linter): rule `no-restricted-imports`: fix option "importNames"

Sysix:
all test cases with `importNames` are now passing.

Now working on the other options


Sysix:
/ping @camc314 

I got already another PR up.
This rule will go into multiple iterations  :)

camc314 (approved):

PR #7959, opened 2024-12-17
feat(transformer/class-properties): transform super assignment expressions that are inside static prop initializer

Dunqing:


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7959?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7959** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7959?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7959?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7956** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7955** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=7959).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/es2022/class_properties/supers.rs
```diff
@@ -34,9 +34,7 @@ impl<'a, 'ctx> ClassProperties<'a, 'ctx> {
         is_callee: bool,
         ctx: &mut TraverseCtx<'a>,
     ) -> Expression<'a> {
-        let property = &member.property;
-        let property =
-            ctx.ast.expression_string_literal(property.span, property.name.clone(), None);
+        let property = Expression::StringLiteral(ctx.ast.alloc(member.property.clone().into()));
```
overlookmotel, line 37: This is shorter, but I'm not really a fan of `impl From<IdentifierName> for StringLiteral`.

Reason is: You're dealing with an owned type on the stack. Usually, it's better to allocate it into the arena as quickly as possible, and then only hold `Box<T>` (8 bytes), references `&T` or `&mut T` (8 bytes), enum types e.g. `Expression`, `Statement` etc (16 bytes), or `Atom`s (16 bytes).

Generally, if you find yourself having to use `ctx.ast.alloc`, you're not in an optimal situation, because it means you have a large type on the stack.

In the original version, you only had to clone the `Atom` (which is essentially a `Copy`), whereas in this new version you have to clone the whole `IdentifierName` to be able to pass it to `from`.

In this case, it's not a big deal, because `IdentifierName` is only 24 bytes, but in general in my opinion it's not the best pattern.

If you wanted to, you could implement this conversion instead as a method on `AstBuilder` - `AstBuilder:: expression_string_literal_from_identifier_name`. But personally I think that'd be overkill unless we need to do that conversion in many places.

crates/oxc_transformer/src/es2022/class_properties/static_prop_init.rs
```diff
@@ -221,6 +221,7 @@ impl<'a, 'ctx, 'v> VisitMut<'a> for StaticInitializerVisitor<'a, 'ctx, 'v> {
             // `object.#prop = value`, `object.#prop += value`, `object.#prop ??= value` etc
             Expression::AssignmentExpression(_) => {
                 self.class_properties.transform_assignment_expression(expr, self.ctx);
+                self.class_properties.transform_super_assignment_expression(expr, self.ctx);
```
overlookmotel, line 224: I think this is the cause of the panic in conformance. Problem is that `transform_assignment_expression` sometimes mutates `expr` so it's not an `AssignmentExpression` any more. So then `transform_super_assignment_expression` enters `unreachable!()`.

So we need to check that it's still an `AssignmentExpression`:

```suggestion
                if matches!(expr, Expression::AssignmentExpression(_)) {
                    self.class_properties.transform_super_assignment_expression(expr, self.ctx);
                }
```

Having to do this check again is inefficient, but don't worry - this visitor will be removed anyway in my forthcoming refactoring, so we can change it then.

PR #7964, opened 2024-12-17
test(transformer/class-properties): add static super tagged template test

Dunqing:
I just found that we don't need to transform `TaggedTemplateExpression` because its `tag` can be transformed by `transform_static_member_expression` and `transform_computed_member_expression`

### DIFF
--- a/tasks/transform_conformance/snapshots/oxc.snap.md
+++ b/tasks/transform_conformance/snapshots/oxc.snap.md
@@ -1,6 +1,6 @@
 commit: 54a8389f
 
-Passed: 112/126
+Passed: 112/127
 
 # All Passed:
 * babel-plugin-transform-class-static-block
@@ -16,7 +16,10 @@ Passed: 112/126
 * regexp
 
 
-# babel-plugin-transform-class-properties (13/15)
+# babel-plugin-transform-class-properties (13/16)
+* static-super-tagged-template/input.js
+x Output mismatch
+
 * typescript/optional-call/input.ts
 Symbol reference IDs mismatch for "X":
 after transform: SymbolId(0): [ReferenceId(0), ReferenceId(2), ReferenceId(6), ReferenceId(11), ReferenceId(16)]
--- a/tasks/transform_conformance/snapshots/oxc_exec.snap.md
+++ b/tasks/transform_conformance/snapshots/oxc_exec.snap.md
@@ -2,4 +2,10 @@ commit: 54a8389f
 
 node: v22.12.0
 
-Passed: 3 of 3 (100.00%)
+Passed: 3 of 4 (75.00%)
+
+Failures:
+
+./fixtures/oxc/babel-plugin-transform-class-properties-test-fixtures-static-super-tagged-template-exec.test.js
+AssertionError: expected undefined to be [Function C] // Object.is equality
+    at ./tasks/transform_conformance/fixtures/oxc/babel-plugin-transform-class-properties-test-fixtures-static-super-tagged-template-exec.test.js:15:17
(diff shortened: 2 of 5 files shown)


Dunqing:
* **#7964** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7964?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7964?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=7964).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Unfortunately this is a bug in Babel (another one!)

```js
import assert from 'assert';

let receiver;
class S {
  static method() {
    receiver = this;
  }
}

class C extends S {
  static prop = super.method`xyz`;
}

assert(receiver === C);
```

The assertion passes.

After Babel transform:

```js
import assert from "assert";

let receiver;
class S {
  static method() {
    receiver = this;
  }
}

class C extends S {}
_C = C;
_defineProperty(C, "prop", _superPropGet(_C, "method", _C)`xyz`);

assert(receiver === C);
```

Assertion fails!

Same as for private field as template tag, it needs extra `.bind(_C)`:

```js
_defineProperty(C, "prop", _superPropGet(_C, "method", _C).bind(_C)`xyz`);
```

[Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAQwM5IKazgMyhEcDkyas-A3AFDkA2q8UqAxqsAG7oUNVFwDKcA3uThwkMBDGAM4IWgAsIAEwAUASgFDhcek1bo4AXjgxZwJBWEBfclfKduAYTioAHjFQA7BUl7rho8ZJwYDhgBiIArmDoAHQyxooABs4AngBeCRQ2ROgwStrMbFAG-ob2KhRAA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.26.4&externalPlugins=%40babel%2Fplugin-transform-class-properties%407.25.9&assumptions=%7B%7D)

I will add this to my list of bugs to report to Babel!

We don't need to fix this right now. It's not important.

But I don't think we should add a test with wrong output either. Maybe better to adapt this test to be a failing test, so we remember to come back to it later.

Dunqing:
Thank you for finding the bug, I changed the output to the correct output and also added an execution test from your example

overlookmotel (approved):
Great. Your version of the exec test is better than mine.

I pushed a commit to remove the 2nd test case, which had the wrong output. Is there a reason that was there that I don't understand?

Also:

* Removed `import { expect } from "vitest";` - test runner already injects `expect` into environment.
* Made `input.js` a copy of `exec.js`. This follows what Babel does. I guess the reason is that when debugging the exec test, it's helpful to be able to see the transformed output.

@Dunqing If there wasn't a good reason why the 2nd test was there, please merge. If there was a good reason, and I shouldn't have removed it, please say!

Dunqing:
### Merge activity

* **Dec 18, 5:24 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 18, 5:24 AM EST**: A user added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #7970, opened 2024-12-17
refactor(transformer/react-refresh): calculate signature key once

Dunqing:
I can't remember why I wrote such code. This change reduced big allocation.

### DIFF
--- a/crates/oxc_transformer/src/jsx/refresh.rs
+++ b/crates/oxc_transformer/src/jsx/refresh.rs
@@ -1,4 +1,5 @@
 use base64::prelude::{Engine, BASE64_STANDARD};
+use compact_str::{CompactString, CompactStringExt};
 use rustc_hash::FxHashMap;
 use sha1::{Digest, Sha1};
 
@@ -111,7 +112,7 @@ pub struct ReactRefresh<'a, 'ctx> {
     /// (eg: hoc(() => {}) -> _s1(hoc(_s1(() => {}))))
     last_signature: Option<(BindingIdentifier<'a>, ArenaVec<'a, Argument<'a>>)>,
     // (function_scope_id, (hook_name, hook_key, custom_hook_callee)
-    hook_calls: FxHashMap<ScopeId, Vec<(Atom<'a>, Atom<'a>)>>,
+    hook_calls: FxHashMap<ScopeId, Vec<CompactString>>,
     non_builtin_hooks_callee: FxHashMap<ScopeId, Vec<Option<Expression<'a>>>>,
 }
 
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#7970** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7970?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7970?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=7970).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):
Nice. Avoids allocating lots of `Atom`s into arena which never end up in the AST.

overlookmotel:
### Merge activity

* **Dec 17, 12:11 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 17, 12:11 PM EST**: A user added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

overlookmotel:
I think there's one further thing you could do if you wanted.

If I understand right (which maybe I don't), the values of `self.hook_calls` always end up being inserted into AST.

So instead of building a `Vec<CompactString>`, you could build a single `oxc_allocator::String` and push new keys onto the end of it. Then when you come to use it, convert to an `Atom` with `Atom::from(the_string.into_bump_str())`.

That remove *all* heap allocations, as the string is built entirely in the arena.

overlookmotel:
Or, after #7973, can just do `Atom::from(arena_string)`.

Dunqing:
> I think there's one further thing you could do if you wanted.
> 
> If I understand right (which maybe I don't), the values of `self.hook_calls` always end up being inserted into AST.
> 
> So instead of building a `Vec<CompactString>`, you could build a single `oxc_allocator::String` and push new keys onto the end of it. Then when you come to use it, convert to an `Atom` with `Atom::from(arena_string.into_bump_str())`.
> 
> That remove _all_ heap allocations, as the string is built entirely in the arena.

By default, we also need to hash the key by `sha1`, so we can't call `arena_string.into_bump_str()` directly
https://github.com/oxc-project/oxc/blob/67d255c15fcc0a2786497866e114131b71201e6b/crates/oxc_transformer/src/jsx/refresh.rs#L521-L530

overlookmotel:
Ah I missed that. Doh!

Still, I think you could probably make a gain by replacing the `Vec<CompactString>` with a single `String` (`std`'s `String`, not `ArenaString`) which you push onto the end of each time a hook is found. That'd be 1 allocation rather than many, and no need for the `join` call at the end.

PR #7982, opened 2024-12-18
fix(transformer/class-properties): run other transforms on static properties, static blocks, and computed keys

overlookmotel:
Large re-architecting of class properties transform. Split transform into 3 phases:

1. Transform instance properties when entering class.
2. Transform private fields during traversal of class body.
3. Transform static properties and static blocks when exiting class.

This ensures that code which has to be moved outside of the class (static property initializers, static blocks, computed keys) can get transformed by other transforms before they're moved out.

Also fixes a problem where we previously registered private properties too early - on entering the class, rather than the class *body*, so private fields in `extends` clause of a nested class were misinterpretted.

PR #8007, opened 2024-12-19
feat(traverse): record current block scope

overlookmotel:
Record "block" scope ID along with "hoist" scope ID in `Traverse`.

"Block" scope is the scope where a `let` statement would be inserted above current position in AST.

Block scope and current scope differ from each other inside classes. For example, if want to create a `let` temp var for `foo()` or `bar()` in example below, should use block scope not current scope. Current scope is the class itself, but the `let` statement will be inserted *outside* the class.

```js
class C {
  [foo()]: bar();
}
```

All transforms which create `let` bindings should use block scope not current scope. We should add `VarDeclarationsStore::insert_let` method which uses block scope, to accompany `insert_var` (which uses hoist scope).

This is implemented in a rather hacky way, and we should improve it later. Notably, we're not considering `for` statements as block scopes because we currently have no way to insert `let` statements into them if they don't have a body block.

### DIFF
--- a/crates/oxc_traverse/scripts/lib/walk.mjs
+++ b/crates/oxc_traverse/scripts/lib/walk.mjs
@@ -129,6 +129,23 @@ function generateWalkForStruct(type, types) {
       `;
       exitScopeCode += 'ctx.set_current_hoist_scope_id(previous_hoist_scope_id);';
     }
+
+    // TODO: Type names shouldn't be hard-coded here. Block scopes should be signalled by attrs in AST.
+    let isBlockScope = [
+      'Program',
+      'BlockStatement',
+      'Function',
+      'ArrowFunctionExpression',
+      'StaticBlock',
+      'TSModuleDeclaration',
+    ].includes(type.name);
+    if (isBlockScope) {
+      enterScopeCode += `
+        let previous_block_scope_id = ctx.current_block_scope_id();
+        ctx.set_current_block_scope_id(current_scope_id);
+      `;
+      exitScopeCode += 'ctx.set_current_block_scope_id(previous_block_scope_id);';
+    }
   }
 
   const fieldsCodes = visitedFields.map((field, index) => {
--- a/crates/oxc_traverse/src/context/mod.rs
+++ b/crates/oxc_traverse/src/context/mod.rs
@@ -188,6 +188,14 @@ impl<'a> TraverseCtx<'a> {
         self.scoping.current_hoist_scope_id()
     }
 
+    /// Get current block scope ID.
+    ///
+    /// Shortcut for `ctx.scoping.current_block_scope_id`.
+    #[inline]
+    pub fn current_block_scope_id(&self) -> ScopeId {
+        self.scoping.current_block_scope_id()
+    }
+
     /// Get current scope flags.
     ///
     /// Shortcut for `ctx.scoping.current_scope_flags`.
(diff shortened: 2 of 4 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8007?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#7982** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8007** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8007?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8007?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8001** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8001?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7995** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7995?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7998** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/7998?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8007).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
I have taken a look at the whole stack. From the usages, it seems we can get `current_block_scope` by looking up the parent scope of `current_scope_id` because when you want to use `current_block_scope`, you are always at the `class` scope, so you only need to look up once and store it somewhere. That I understand correctly?

To clarify, I am not rejecting this API at all; on the contrary, I am happy to add various APIs in `Traverse` and `TransformerCtx` to solve annoying duplicating code. So If you find that can be fixed or refactored in other places by it, tell me!

overlookmotel:
> it seems we can get `current_block_scope` by looking up the parent scope of `current_scope_id` because when you want to use `current_block_scope`, you are always at the `class` scope, so you only need to look up once and store it somewhere. That I understand correctly?

That's what I thought but actually no!

This gem was in one of Babel's tests:

```js
class Outer extends class Inner {
  [foo()] = 123
} {}
```

So when you're in the inner class, is *not* where a `let` binding gets inserted - the parent scope is the scope of the outer class!

Dunqing:
> > it seems we can get `current_block_scope` by looking up the parent scope of `current_scope_id` because when you want to use `current_block_scope`, you are always at the `class` scope, so you only need to look up once and store it somewhere. That I understand correctly?
> 
> That's what I thought but actually no!
> 
> This gem was in one of Babel's tests:
> 
> ```js
> class Outer extends class Inner {
>   [foo()] = 123;
> } {}
> ```
> 
> So when you're in the inner class, the parent scope is _not_ where a `let` binding gets inserted - the parent scope is the scope of the _outer_ class!

Aha, I see the problem, fu** JavaScript!

Issue #7792, opened 2024-12-11
Invalid output from async-to-generator transform when async arrow function in class constructor


overlookmotel:
Input:

```js
class C extends S {
  constructor() {
    super();
    this.f = async () => this;
  }
}
```

Transformed by async-to-generator:

```js
class C extends S {
  constructor() {
    var _this = this;
    super();
    this.f = _asyncToGenerator(function*() {
      return _this;
    });
  }
}
```

[Oxc Playground](https://playground.oxc.rs/#eNo9UEFqw0AM/IrQqQUT2kIvLumlkA/0mstmI6db1isjyW2C8d8jr3FgYTWjmZHQhBFbjDmowhfQ1aicFb5hOhaAyEVNxmgsT88rBaDjQA4/VmQ/SXcd7CHorURw2f6zkrU/H4s/bJCxnVDGsnwutHDF1pOpwZyKYduFrA408kBbR2/9ifOGTELRjqVfibnBIYiS1EQeJdIhZSqhdz8aqe1+FV215FeV14+M1WZBLuTDkfTt5fV9UUc+04WWNR30qaQubWY/hgnnQ+b/xfxHcmL1YXX1eZ7vxWVwYg==)

Problem is that `this` is now accessed before `super()`, which is an error.

Babel puts `_this = this` after `super()`:

```js
class C extends S {
  constructor() {
    var _this;
    super();
    _this = this;
    this.f = _asyncToGenerator(function*() {
      return _this;
    });
  }
}
```

[...]

PR #8024, opened 2024-12-19
fix(transformer/arrow-functions): `_this = this` should be inserted after super call expression

Dunqing:
related: #7792 

This PR doesn't contain fixing the async arrow function in `super()`, which is difficult for our architecture. I just found that `esbuild`'s implementation is quite simpler! https://esbuild.github.io/try/#dAAwLjI0LjAALS10YXJnZXQ9ZXMyMDE2AGNsYXNzIEMgZXh0ZW5kcyBTIHsKICBjb25zdHJ1Y3RvcigpIHsKICAgIHN1cGVyKGFzeW5jICgpID0+IHRoaXMpOwogICAgYXN5bmMoKSA9PiB7fQogIH0KfQ


Dunqing:
* **#8024** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8024?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8024?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8024).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/common/arrow_function_converter.rs
```diff
@@ -703,6 +728,25 @@ impl<'a> ArrowFunctionConverter<'a> {
...
+                            matches!(
+                                ancestor,
+                                // const A = super():
+                                Ancestor::VariableDeclarationDeclarations(_)
+                                // super();
+                                | Ancestor::ExpressionStatementExpression(_)
+                            )
+                        })
+                        .unwrap()
+                        .address()
```
Dunqing, line 747: Unwrap failed here, we need to find all possible statements, and this may not be a good implementation

Dunqing:
@overlookmotel Can you take a look at this PR, I am not sure this is a good. I really want to change everything according to esbuild's implementation

overlookmotel:
Concerning ESBuild's version:

Yes it's simpler to implement, but I think less performant at runtime than Babel's version. In Babel's version, there's only 1 call to `asyncToGenerator` when the async function is defined. But in ESBuild's version, `__async` gets called and constructs a new generator function *every time you call the async function*.

I don't know for sure, but that sounds more expensive (more garbage collection).

I think we could do our own version which is better than both! Maybe something like this:

Input:

```js
class C extends S {
  constructor() {
    super();
    let f = async () => this;
  }
}
```

Output:

```js
class C extends S {
  constructor() {
    var _getThis = () => this;
    super();
    let f = oxcHelpers.asyncToGenerator(function*() { return this; }, _getThis, "f");
  }
}
```

(the last param takes care of naming the function correctly, instead of Babel's extra function wrappers)

But... I suggest we follow Babel for now, make our implementation correct, and look at optimizing output later. There's many places we can improve on Babel in *lots* of transforms.

overlookmotel:
PS: I've made a small improvement to the `super()`-replacer visitor in #8028.

crates/oxc_transformer/src/common/arrow_function_converter.rs
overlookmotel: You can use `Ancestor::is_via_statement` for this.

Note it doesn't tell you if that ancestor *is* a statement, but whether its *child* is a statement.

So if you are visiting a `Function`, it could be either an expression or a statement. `ctx.parent().is_via_statement()` will tell you.

To find the closest statement, loop through ancestors, and when `ancestor.is_via_statement()` returns `true`, then the *previous* ancestor you visited is the statement.

The naming `is_via_*` is weird, and the fact that you have to look at the ancestor above is confusing as hell. So it's a terrible API! But it does do what you need.

overlookmotel: #8031 renames this method to `Ancestor::is_parent_of_statement`. I think that's clearer.

Dunqing: I just found that we have these methods!

Dunqing:
That's a trade-off for us considering the different implementations. When I am a developer for the transformer, I prefer a 
way easy to implement save my hair!

Dunqing:
To clarify, so in this case, you prefer to double-visited? I have no strong opinion about this. Just think of this way is easier to implement, but cannot solve `super(async () => {})`.

overlookmotel:
I think double-visit is the best way, yes.

I *think* it'll be able to handle the `super(async () => this)` case too.

We could later try and optimize it for common case where `super()` is a top level statement, by storing the `Address` of `super()` during traversal, and then you know exactly where to insert `_this = this` without another visit.

But:

1. I'm not actually sure if that'll be a perf gain or not, because you have to do a lot more checking for "is this node `super()`?", when most of the time the answer is "no". And when the class constructor *doesn't* contain an async arrow function (most won't), then that's wasted work.
2. That won't handle all the edge cases anyway.
3. Make it correct first, then optimize.

Dunqing:
I have reimplemented the fixer in https://github.com/oxc-project/oxc/pull/8024/commits/89eb36f541d6ca4381e827580834eceabd01e1f1. The new implementation covers `super(async () => { this })` case

overlookmotel (approved):
Very nice!

I pushed a couple of commits to extend the tests, and will follow-up with some refactoring PRs. But it's all small stuff. Overall, looks great.

overlookmotel:
### Merge activity

* **Jan 7, 8:45 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #8038, opened 2024-12-20
fix(binder): bind declare class

### LINKED ISSUE #7278: No `SymbolId` for `declare class C {}`
Semantic does not produce a `SymbolId` for `C` in `declare class C {}`.

I believe it should - all `BindingIdentifier`s should have a `SymbolId`, and we do produce `SymbolId`s for other TS-only identifiers e.g. `type N = number;`.

In this example, only `A` gets assigned a `SymbolId`:

```ts
class A {}
declare class B {}
export declare class C {}
```

[Playground](https://playground.oxc.rs/#eNpVTs0OwiAMfpWl5z0BNzXxCTx6YawzJECXFnTLwrsLzGk8td9vu4EBBcZpke7UbfkeRiyIsdu5c+NwmYlj9y9dmgQ9EKgNOIU6ZA1RL6AiJ+zB2RBBTdpJAWJoxkOR1Q/kDhRZB5mI/cece5g1C3KrpMQGb2sNg6cxOYRiqN3NUPZv/iAMjfjA+lEB3gY72Z8UIpO7OnrV9ifyQFKq98M5vwGfPGFl)

Relevant code:

https://github.com/oxc-project/oxc/blob/ff2a1d4937fa252c0ccf2983bf47d78e0d4ab67a/crates/oxc_semantic/src/binder.rs#L103-L118

ShuiRuTian:
Fix #7278

### DIFF
--- a/crates/oxc_semantic/src/binder.rs
+++ b/crates/oxc_semantic/src/binder.rs
@@ -100,20 +100,14 @@ impl<'a> Binder<'a> for VariableDeclarator<'a> {
 
 impl<'a> Binder<'a> for Class<'a> {
     fn bind(&self, builder: &mut SemanticBuilder) {
-        if !self.declare {
-            let Some(ident) = &self.id else { return };
-            let symbol_id = builder.declare_symbol(
-                ident.span,
-                &ident.name,
-                SymbolFlags::Class,
-                if self.is_declaration() {
-                    SymbolFlags::ClassExcludes
-                } else {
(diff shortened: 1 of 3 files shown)


ShuiRuTian:
Not that easy, convert to draft for first.

Need to consider about `redeclaration`(the term used in oxc now)

This should be valid:
```
declare class CCC { }

function CCC() { }
```

PR #8063, opened 2024-12-22
feat(minifier): fold `foo === undefined || foo === null`

sapphi-red:
This PR implements folding `foo === undefined || foo === null` into `foo == null`.

I checked the minified output diff this time, so hoping that there isn't a bug.


crates/oxc_minifier/src/ast_passes/peephole_substitute_alternate_syntax.rs
```diff
@@ -265,6 +276,123 @@ impl<'a, 'b> PeepholeSubstituteAlternateSyntax {
...
 
+    /// Compress `foo === null || foo === undefined` into `foo == null`.
+    ///
+    /// `foo === null || foo === undefined` => `foo == null`
+    /// `foo !== null && foo !== undefined` => `foo != null`
+    ///
+    /// This compression assumes that `document.all` is a normal object.
+    /// If that assumption does not hold, this compression is not allowed.
+    /// - `document.all === null || document.all === undefined` is `false`
+    /// - `document.all == null` is `true`
```
sapphi-red, line 287: terser makes this assumption.
https://github.com/terser/terser#:~:text=document.all%20is%20not%20%3D%3D%20null

PR #8067, opened 2024-12-23
feat(napi/parser): add `crate-type: "lib"`

underfin:
The rolldown need export it to instead of `rollup/parseAst`, ref https://rollupjs.org/javascript-api/#accessing-the-parser


underfin:
@Dunqing Could you publish it to `crate.io`?

underfin:
> @Dunqing Could you publish it to `crate.io`?

cc @Boshen

PR #8123, opened 2024-12-26
feat(linter): add `curly` rule

baseballyama:
implement https://eslint.org/docs/latest/rules/curly

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/curly.rs
+++ b/crates/oxc_linter/src/rules/eslint/curly.rs
@@ -109,73 +109,88 @@ impl Rule for Curly {
 
         match node.kind() {
             AstKind::IfStatement(stmt) => {
-                let mut statements_are_else = vec![false];
-                let mut statements = vec![&stmt.consequent];
-                let mut should_have_braces_list =
-                    vec![should_have_braces(root_node, &self.0.options, &stmt.consequent, ctx)];
-                let mut has_braces_list = vec![has_braces(&stmt.consequent)];
+                struct StatementInfo<'a> {
+                    statement: &'a Statement<'a>,
+                    is_else: bool,
+                    should_have_braces: Option<bool>,
+                    has_braces: bool,
+                }
+
+                let mut infos = vec![StatementInfo {
+                    statement: &stmt.consequent,
+                    is_else: false,
+                    should_have_braces: should_have_braces(
+                        root_node,
+                        &self.0.options,
+                        &stmt.consequent,
+                        ctx,
+                    ),
+                    has_braces: has_braces(&stmt.consequent),
+                }];
+
                 let mut current_statement = &stmt.alternate;
 
                 while let Some(statement) = current_statement {
-                    let (consequent, alternate) = if let Statement::IfStatement(node) = statement {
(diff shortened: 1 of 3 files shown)


camchenry:
Thank you for the contribution! I will review this when I have some time.

crates/oxc_linter/src/rules/eslint/curly.rs
```diff
@@ -0,0 +1,1904 @@
...
+            matches!(decl.kind, VariableDeclarationKind::Const | VariableDeclarationKind::Let)
+        }
+        Statement::FunctionDeclaration(_) | Statement::ClassDeclaration(_) => true,
+        _ => false,
+    }
+}
+
+fn is_followed_by_else_keyword(root_node: &AstNode, node: &Statement, ctx: &LintContext) -> bool {
+    let start = node.span().end + 1;
+    let end = root_node.span().end;
```
camchenry, line 420: I don't think the root node is necessary here? You can get the last byte offset with `ctx.source_text().len()`

crates/oxc_linter/src/rules/eslint/curly.rs
```diff
@@ -0,0 +1,1904 @@
...
+                    Span::new(parent.span().start, parent.span().start)
+                })
+        },
+        oxc_span::GetSpan::span,
+    );
+
+    let text = ctx.source_range(Span::new(
+        before_node_span.end + 1,
+        span.end
+            - (u32::try_from(node_string.len()).expect("Failed to convert to u32") - trimmed_len),
```
camchenry, line 373: No need to use `try_from` and expect here, it's OK to assume files are not larger than 4GB, it's a general assumption we make. You can use `as u32` and add the annotation `#[allow(clippy::cast_possible_truncation)]`

crates/oxc_linter/src/rules/eslint/curly.rs
```diff
@@ -0,0 +1,1904 @@
...
+        }
+    }
+}
+
+fn get_root_node<'a, 'b>(mut node: &'a AstNode<'b>, ctx: &'a LintContext<'b>) -> &'a AstNode<'b> {
+    while let Some(parent) = ctx.nodes().parent_node(node.id()) {
+        node = parent;
+    }
+    node
+}
```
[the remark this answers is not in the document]
baseballyama: Finally I don't need to use root_node.

fixed: https://github.com/oxc-project/oxc/pull/8123/commits/d760bc434f99e25970e59bce3e0c7e933f327f12

crates/oxc_linter/src/rules/eslint/curly.rs
baseballyama: fixed: https://github.com/oxc-project/oxc/commit/d760bc434f99e25970e59bce3e0c7e933f327f12

crates/oxc_linter/src/rules/eslint/curly.rs
baseballyama: OK! updated: https://github.com/oxc-project/oxc/pull/8123/commits/ea76f7d3b52ff30ac56d8a9186a6b00e2374de04

PR #8126, opened 2024-12-26
chore(ast): use `assert_eq!` instead of `assert!` (`crates/oxc_ast/src/generated/assert_layouts.rs`)

baseballyama:
related to https://github.com/oxc-project/oxc/pull/8124


baseballyama:
This file can not convert to `assert_eq!`.

PR #8138, opened 2024-12-27
feat(semantic): add `get_node_by_span` function

baseballyama:
Many lint rules have already been implemented, so this function may not be necessary. However, It's useful for me, so I submitted this PR. I would appreciate the opportunity to discuss its necessity.

For example, when a function that takes a Statement (e.g. `Statement::BlockStatement`) as an argument needs to retrieve its parent Node, it requires a node_id. Since node_id cannot be obtained directly from a Statement, the only way is to fetch the Node from the span. Here’s an example:

```rust
fn get_node_by_statement<'a>(statement: &'a Statement, ctx: &'a LintContext) -> &'a AstNode<'a> {
    let span = statement.span();
    ctx.nodes().iter().find(|n| n.span() == span).expect("Failed to get node by statement")
}
```

However, this process is costly, and it would be better to make it more efficient. To address this, this PR adds a function that creates a Map in advance, allowing subsequent lookups to retrieve the Node directly from the Map.

### DIFF
--- a/crates/oxc_semantic/src/node.rs
+++ b/crates/oxc_semantic/src/node.rs
@@ -1,7 +1,8 @@
+use hashbrown::HashMap;
 use oxc_ast::AstKind;
 use oxc_cfg::BlockNodeId;
 use oxc_index::IndexVec;
-use oxc_span::GetSpan;
+use oxc_span::{GetSpan, Span};
 use oxc_syntax::{
     node::{NodeFlags, NodeId},
     scope::ScopeId,
@@ -108,6 +109,7 @@ pub struct AstNodes<'a> {
     nodes: IndexVec<NodeId, AstNode<'a>>,
     /// `node` -> `parent`
     parent_ids: IndexVec<NodeId, Option<NodeId>>,
+    node_by_span: HashMap<Span, AstNode<'a>>,
 }
 
 impl<'a> AstNodes<'a> {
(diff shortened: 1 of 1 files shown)


baseballyama:
As expected, this seems to impact performance. I’m not sure how to achieve this without affecting performance—or perhaps it’s better not to implement it at all?

Boshen:
This is not ideal. All linter rules should be implemented via going up or down the AST.

baseballyama:
@Boshen 
I understand it that way, but how can I retrieve stmt.body as an AstNode in cases like the one below?


```rust
fn foo(node: &AstNode) {
    if let AstKind::BlockStatement(stmt) = node.kind() {
        // How to get body as AstNode?
        stmt.body
    }
}
```

Boshen:
> @Boshen I understand it that way, but how can I retrieve stmt.body as an AstNode in cases like the one below?
> 
> ```rust
> fn foo(node: &AstNode) {
>     if let AstKind::BlockStatement(stmt) = node.kind() {
>         // How to get body as AstNode?
>         stmt.body
>     }
> }
> ```

I don't follow, can you elaborate?

As for this PR - close as not planned.

baseballyama:
If we need to get an `AstNode` for `stmt.body` (for instance, when a `node_id` is needed), currently, there are no straightforward methods available. It is necessary to get an AstNode in some way. I think I experienced confusion when I started implementing oxlint rules because, unlike ESLint where we could simply walk the AST, with oxlint we need to handle both AstNode and specific statements/declarations, etc.
In any case, as I gradually become more accustomed to implementing the rules, I might be able to share some different ideas as I learn from others.

PR #8144, opened 2024-12-27
fix(ecmascript): incorrect `to_int_32` value for Infinity

sapphi-red:
https://tc39.es/ecma262/#sec-toint32
Infinity should return `0` when passed to `ToInt32` abstract operation.

### DIFF
--- a/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
+++ b/crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
@@ -1505,9 +1505,9 @@ mod test {
 
     #[test]
     fn test_fold_left_child_op() {
-        test_same("x & infinity & 2"); // FIXME: want x & 0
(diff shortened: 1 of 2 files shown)


crates/oxc_minifier/src/ast_passes/peephole_fold_constants.rs
```diff
@@ -1505,9 +1505,9 @@ mod test {
 
     #[test]
     fn test_fold_left_child_op() {
-        test_same("x & infinity & 2"); // FIXME: want x & 0
-        test_same("x - infinity - 2"); // FIXME: want "x-infinity"
-        test_same("x - 1 + infinity");
+        test("x & Infinity & 2", "x & 0");
+        test_same("x - Infinity - 2"); // FIXME: want "x-Infinity"
+        test_same("x - 1 + Infinity");
```
sapphi-red, line 1510: Also fixed the incorrect casing of `Infinity` I wrote in past 😅

PR #8162, opened 2024-12-28
feat(minifier): minimize computed property access

sapphi-red:
Ported `ConvertToDottedProperties` partially. The rest seems to be mostly handled by #8147.

### DIFF
--- a/crates/oxc_minifier/src/ast_passes/convert_to_dotted_properties.rs
+++ b/crates/oxc_minifier/src/ast_passes/convert_to_dotted_properties.rs
@@ -1,0 +1,277 @@
+use oxc_ast::ast::*;
+use oxc_span::GetSpan;
+use oxc_syntax::identifier::is_identifier_name;
+use oxc_traverse::{traverse_mut_with_ctx, ReusableTraverseCtx, Traverse, TraverseCtx};
+
+use crate::CompressorPass;
+
+/// Converts property accesses from quoted string or bracket access syntax to dot or unquoted string
+/// syntax, where possible. Dot syntax is more compact.
+/// <https://github.com/google/closure-compiler/blob/v20240609/src/com/google/javascript/jscomp/ConvertToDottedProperties.java>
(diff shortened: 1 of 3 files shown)


crates/oxc_minifier/src/ast_passes/convert_to_dotted_properties.rs
```diff
@@ -0,0 +1,277 @@
...
+        test_same("a[':']");
+        test_same("a['.']");
+        test_same("a['0']");
+        test_same("a['p ']");
+        test_same("a['p' + '']");
+        test_same("a[p]");
+        test_same("a[P]");
+        test_same("a[$]");
+        test_same("a[p()]");
+        // test_same("a['default']");
```
sapphi-red, line 106: Keywords seems to be not allowed for ES3, but I guess this is fine for us.

crates/oxc_minifier/src/ast_passes/convert_to_dotted_properties.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Ignorable control characters are ok in Java identifiers, but not in JS.
+        test_same("a['A\\u0004']");
+        // upper case lower half of o from phonetic extensions set.
+        // valid in Safari, not in Firefox, IE.
+        // test_same("a['\\u1d17A']");
+        // Latin capital N with tilde - nice if we handled it, but for now let's
+        // only allow simple Latin (aka ASCII) to be converted.
+        // test_same("a['\\u00d1StuffAfter']");
```
sapphi-red, line 114: `\u1d17A` is invalid for ES5 (ES5 uses Unicode 3.0 and that character was introduced in Unicode 3.1) and valid for ES2015 (ES2015 uses Unicode 5.1).
`\u00d1` is valid for ES3+.

Boshen:
This falls under substitute alternative syntax. I'll move the code around when I get the time.

sapphi-red:
=> #8175, #8176

PR #8192, opened 2024-12-30
fix(transformer): ensure last expression statement in arrow function expression is wrapped in return

Dunqing:
The class properties plugin will insert new statements after the expression statement, which causes the expression statement to no longer be at the end. This PR fixed it by reverse looping the statements to find the expression statement and transforming it to wrapped in `return`. 

This is not a good way to do it, a potential problem is if there is a plugin that inserts a new expression statement after the original expression statement, which will cause a wrong expression wrapped in a `return`


Dunqing:
### Merge activity

* **Dec 30, 8:29 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Dunqing:
To unblock the class-properties plugin development, so merge it without review

PR #8193, opened 2024-12-30
fix(parser): allow line breaks between `const` and `enum`

wan9chi:
https://www.typescriptlang.org/play/?ts=5.7.2#code/MYewdgzgLgUApmArgWwAQEFUG8C+Mg

### DIFF
--- a/crates/oxc_parser/src/modifiers.rs
+++ b/crates/oxc_parser/src/modifiers.rs
@@ -318,7 +318,7 @@ impl<'a> ParserImpl<'a> {
         }
 
         match self.cur_kind() {
-            Kind::Const => !self.peek_token().is_on_new_line && self.peek_kind() == Kind::Enum,
+            Kind::Const => self.peek_kind() == Kind::Enum,
             Kind::Export => {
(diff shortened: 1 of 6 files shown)


Boshen:
.d.ts panel prints 

```
declare const enum A {
}
```

Double confirmed with babel.

branchseer:
> Your proposal isn't correct.
> 
> .d.ts panel prints
> 
> ```
> declare const enum A {
> }
> ```
> 
> Double confirmed with babel in ast explorer.

Sorry I don't follow. Doesn't the .d.ts panel show that `const (line break) enum` is valid?

Boshen:
> > Your proposal isn't correct.
> > .d.ts panel prints
> > ```
> > declare const enum A {
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Double confirmed with babel in ast explorer.
> 
> Sorry I don't follow. Doesn't the .d.ts panel show that `const (line break) enum` is valid?

Oh sorry I confused myself.

PR #8337, opened 2025-01-08
refactor(prettier): Update tasks/prettier to correctly handle snapshots

leaysgur:
I refactored the code in `tasks/prettier_conformance` primarily to make the output more readable when using `--filter`.

But I also discovered that our previous implementation did not correctly handle Prettier's behavior of adding a blank line at the EOF.

In addition, I resolved a problem where test specs that used patterns like `runFormatTest(_, parsers)` were unable to locate the correct snapshot output.

As a result, compatibility has also improved slightly. 😉

### DIFF
--- a/crates/oxc_prettier/src/format/js.rs
+++ b/crates/oxc_prettier/src/format/js.rs
@@ -33,8 +33,7 @@ impl<'a> Format<'a> for Program<'a> {
 
             if let Some(body_doc) = block::print_block_body(p, &self.body, Some(&self.directives)) {
                 parts.push(body_doc);
-                // XXX: Prettier seems to add this, but test results don't match
-                // parts.extend(hardline!());
+                parts.extend(hardline!());
             }
 
             array!(p, parts)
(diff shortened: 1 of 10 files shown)


crates/oxc_prettier/src/format/js.rs
```diff
@@ -33,8 +33,7 @@ impl<'a> Format<'a> for Program<'a> {
 
             if let Some(body_doc) = block::print_block_body(p, &self.body, Some(&self.directives)) {
                 parts.push(body_doc);
-                // XXX: Prettier seems to add this, but test results don't match
-                // parts.extend(hardline!());
+                parts.extend(hardline!());
```
leaysgur, line 36: Previously, this was hard-coded into the conformance test process.

tasks/prettier_conformance/src/spec.rs
```diff
@@ -33,39 +43,69 @@ impl SpecParser {
...
-                    })
-                    .collect::<Vec<String>>();
+                for el in &arr_expr.elements {
+                    if let ArrayExpressionElement::StringLiteral(literal) = el {
+                        parsers.push(literal.value.to_string());
+                    }
+                }
+            }
+            // If variable
+            if let Expression::Identifier(_) = argument_expr {
```
leaysgur, line 97: In some cases it was written as `runFormatTest(_, parsers)` instead of `runFormatTest(_, ["babel"])`.

Boshen:
Nice!

PR #8387, opened 2025-01-09
test(transformer/async-to-generator): failing test for async arrow function in class static block

overlookmotel:
It's arguable whether we should support this, because async functions is ES2018 and class static blocks is ES2022. So there should not be any browser which needs async-to-generator transform, and not the static block transform too. And when class-static-block transform is enabled, this works fine.

But personally, I think we should support it, because:

1. We allow user to enable/disable arbitrary transforms via `TransformOptions`.
2. We support `this` in async arrow functions in class static blocks, so it makes sense to complete that support.
3. I don't think it should be too hard, as the framework for it is already there in the async-to-generator transform.

[Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYGwhgzhAEDC0FMAeAXBA7AJjAytA3gFDTQQpgoCWwBxJ0KAFpRAHQBm60AvNJAJ7oaACgCUPAHykArgAcEAJw4B7ZQG46AX0LagA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.26.4&externalPlugins=%40babel%2Fplugin-transform-async-to-generator%407.25.9%2C%40babel%2Fplugin-external-helpers%407.25.9&assumptions=%7B%7D)

### DIFF
--- a/tasks/transform_conformance/snapshots/oxc.snap.md
+++ b/tasks/transform_conformance/snapshots/oxc.snap.md
@@ -1,6 +1,6 @@
 commit: 54a8389f
 
-Passed: 126/147
+Passed: 126/148
 
 # All Passed:
 * babel-plugin-transform-class-static-block
@@ -10,7 +10,6 @@ Passed: 126/147
 * babel-plugin-transform-optional-catch-binding
 * babel-plugin-transform-async-generator-functions
 * babel-plugin-transform-object-rest-spread
-* babel-plugin-transform-async-to-generator
 * babel-plugin-transform-exponentiation-operator
 * babel-plugin-transform-arrow-functions
 * babel-preset-typescript
@@ -45,6 +44,11 @@ rebuilt        : SymbolId(0): [ReferenceId(0), ReferenceId(2), ReferenceId(8), R
 Symbol reference IDs mismatch for "X":
 after transform: SymbolId(0): [ReferenceId(0), ReferenceId(2), ReferenceId(6), ReferenceId(9), ReferenceId(12)]
 rebuilt        : SymbolId(0): [ReferenceId(0), ReferenceId(2), ReferenceId(6), ReferenceId(10)]
+
+
+# babel-plugin-transform-async-to-generator (20/21)
+* class/static-block/input.js
+x Output mismatch
 
 
 # babel-plugin-transform-typescript (2/13)
--- a/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/class/static-block/input.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/class/static-block/input.js
@@ -1,0 +1,5 @@
+class C extends S {
+  static {
+    this.fn = async () => super.foo;
+  }
+}
--- a/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/class/static-block/output.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-async-to-generator/test/fixtures/class/static-block/output.js
@@ -1,0 +1,8 @@
+class C extends S {
+  static {
+    var _superprop_getFoo = () => super.foo;
+    this.fn = babelHelpers.asyncToGenerator(function* () {
+      return _superprop_getFoo();
+    });
+  }
+}


Dunqing:
Even if the async arrow function is not in the class static block, it also needs to transform `super`, because `async-to-generator` will transform the async arrow function to a function generator, thus `super` will be in the wrong place.

But there is a troublesome example

Input:
```js
class C { static f = async () => super.prop;  }
```

Babel's Output:
```js
var _superprop_getProp = () => super.prop;
class C {
  static f = babelHelpers.asyncToGenerator(function* () {
    return _superprop_getProp();
  });
}
```

Babel's implementation is wrong because it inserts `_superprop_getProp` outside the class. But how to fix it? Where it is the correct place to insert `_superprop_getProp`?

overlookmotel:
> Just found that Babel's output is incorrect, `super` can't be outside a class. It wasted a few hours for me.

Oh bollocks sorry my fault. I posted the Babel REPL link meaning to say "look Babel is wrong again!" but I forgot to actually write that. The test in this PR shows what I think the correct output is for static blocks.

For static blocks, I don't expect it's too hard because there's already a block for the `var` statement to go in, and the transform already correctly handles `this` correctly in static blocks: `class C { static { async () => this; } }`

overlookmotel:
For properties (instance prop *or* static prop), it's much more annoying. Babel's solution for instance props looks correct, but for some reason it gets it wrong for static props: [Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYGwhgzhAEDC0FMAeAXBA7AJjAytA3gFDTQBm60AvNJAJ7rDQAUAlFQHzQDaKAFgJYQANNAgBXAA4IATgDpSAewUBdANzFRKMCn6MIWncABiFanQbM2lTjwHDRkmfKVrCAX0JA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.26.4&externalPlugins=%40babel%2Fplugin-transform-async-to-generator%407.25.9%2C%40babel%2Fplugin-external-helpers%407.25.9&assumptions=%7B%7D)

But I'm not sure we should attempt to cover async arrow functions in class properties. It's maybe not worth the work. I only suggested that we might want to deal with static *blocks* because I thought it'd be quite easy.

We also have this problem with other transforms. e.g. `class C { prop = x ?? y; }` should probably create the temp var inside the property initializer, not outside the class. Or maybe in this cases it's fine ([Babel thinks so](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYGwhgzhAEDC0G8BQ1oAcBOB7N0C80WARgFYB0AHtAPzXQCeA3EgL5JA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.26.4&externalPlugins=%40babel%2Fplugin-external-helpers%407.25.9%2C%40babel%2Fplugin-transform-nullish-coalescing-operator%407.25.9&assumptions=%7B%7D)), but in other transforms it's probably not fine.

We have other weird cases where we need to create temp vars as part of a transform inside:

* function f(x = somethingThatNeedsALocalTempVarWhenTransformed) {}
* `for (let [x = somethingThatNeedsALocalTempVarWhenTransformed] of arr) {}`
* Probably more

I put class property initializers in the same category.

Rather than tackling this in individual transforms, I think we should probably try to figure out a way to handle these problems universally with helpers that create temp vars, and handle these edge cases for us.

overlookmotel:
It may also be easier to deal with when we have our own helper library. e.g.:

```js
class C {
  static f = async () => super.prop;
}
```

->

```js
class C {
  static f = oxcHelpers.asyncToGenerator(
    function*(_superprop_getProp) {
      return _superprop_getProp();
    },
    () => super.prop
  );
}
```

Dunqing:
> It may also be easier to deal with when we have our own helper library. e.g. something like:
> 
> ```js
> class C {
>   static f = async (x, y) => super.prop;
> }
> ```
> 
> ->
> 
> ```js
> class C {
>   static f = oxcHelpers.asyncToGenerator(
>     function*(x, y, _superprop_getProp) {
>       return _superprop_getProp();
>     },
>     () => super.prop
>   );
> }
> ```

I just came up with a way by `SequenceExpression``
```js
class C {
  static f = (var _superprop_getProp = () => super.prop, babelHelpers.asyncToGenerator(function* () {
    return _superprop_getProp();
  }));
}
```

overlookmotel:
Nice! That does work for static props, but it won't work for instance props because you need a *different* `_superprop_getProp` for each class *instance*.

Developing your idea further, I think this works for both static and instance:

Input:

```js
class C {
  f = async (x, y) => [this, super.prop];
  static f = async (x, y) => [this, super.prop];
}
```

Output:

```js
class C {
  f = (function(_this, _superprop_getProp) {
    var _ref = babelHelpers.asyncToGenerator(function*(x, y) {
      return [_this, _superprop_getProp()];
    });
    return function (_x, _x2) {
      return _ref.apply(this, arguments);
    };
  })(this, () => super.prop);

  static f = (function(_this2, _superprop_getProp2) {
    var _ref = babelHelpers.asyncToGenerator(function*(x, y) {
      return [_this2, _superprop_getProp2()];
    });
    return function (_x, _x2) {
      return _ref.apply(this, arguments);
    };
  })(this, () => super.prop);
}
```

BUT... Do we want to do this now? I think in the end we can do it better once we have our own helpers lib, so we will end up rewriting it anyway.

And supporting async-to-generator transform *without* class properties transform is not required for our core purpose which is to support `target: 'es2015'` or `target: 'es2020'`.

Dunqing:
I have tried to solve this problem at https://github.com/oxc-project/oxc/pull/8435, luckily it didn't take me a long time.

PR #8435, opened 2025-01-11
fix(transformer/arrow-functions): transform `this` and `super` incorrectly in async arrow function

### LINKED ISSUE #8385: Suggested changes to arrow functions transform
Continuing from conversation on https://github.com/oxc-project/oxc/pull/8335#discussion_r1907885466:

I said:

> The principle is this: You _must_ make sure that every `pop` follows a `push` so the stack and traversal stay in sync. So if you push/pop conditionally e.g.:
> 
> ```rust
> fn enter_some_node(...) {
>   if node.whatever {
>     stack.push(data);
>   }
> }
> 
> fn exit_some_node(...) {
>   if node.whatever {
>     let data = stack.pop();
>   }
> }
> ```
> 
> Then you need to be absolutely sure that the value of `node.whatever` is the same in both `enter_*` and `exit_*`. That includes what any _other_ transform may do to the node in between.
> 
> You can only rely on 2 things:
> 
> 1. Every `enter_*` visitor call has a corresponding `exit_*` call.
> 2. The ancestors of the node are unchanged between `enter_*` and `exit_*` (because you cannot mutate upwards).
> 
> But all the properties of the node itself are fluid - any other transform can change them in any way.
> 
> As much as possible, I think it's ideal if individual transforms are not tightly coupled. When working on one transform, you shouldn't need to understand what _other_ transforms may do to AST. None of us knows all the transforms inside out, so that'd be impossible! So I have tended to code defensively, and assume other transforms could do _anything_ in between `enter_*` and `exit_*`.
> 
> I try to _always_ push to a stack when entering a node, and _always_ pop when exiting it, so it's impossible for the stack and the traversal to get out of sync. Then use the saved value to decide what to do.
> 
> `SparseStack` is ideal for this because pushing `None` to it is very cheap indeed (internally it just writes a single `bool`). Then in `exit_*` you pop from the stack unconditionally, and decide whether you need to make a modification to the AST or not depending on if the value is `Some` or `None`.

@Dunqing replied:

> Another thing I want to avoid is passing too many arguments to `insert_variable_statement_at_the_top_of_statements`, we have to pass all stack values in call site, because only program we need to use `last_mut`, otherwise we use `pop`.

2 suggestions to improve on #8335 based on that conversation:

### Defensive coding

As I said above, you cannot rely on the node not being changed by other transforms between `enter_*` and `exit_*` visitors.

In this case, `self.is_async_only() && Self::is_class_method_like_ancestor(ctx.parent())` is guaranteed to be the same in `enter_function` and `exit_function`. BUT it's possible that `func.r#async` is `false` in `enter_function` but has been changed to `true` by another transform before `exit_function`. That would cause a panic because stack goes out of sync with traversal.

So I suggest instead of:

https://github.com/oxc-project/oxc/blob/335065d8c8f5d2995bc3c4e66cec1d4baf482f0b/crates/oxc_transformer/src/common/arrow_function_converter.rs#L216-L242

```rs
if self.is_async_only() && Self::is_class_method_like_ancestor(ctx.parent()) {
    let super_methods = if func.r#async { Some(FxIndexMap::default()) } else { None };
    self.super_methods_stack.push(super_methods);
}
```

Then in `exit_function`, do the same.

i.e. Don't rely on `func.r#async` being the same in `enter_function` and `exit_function` to ensure stack stays in sync with traversal.

This does have a small perf cost, because you push to the stack for *every* function now. But that cost is very small because pushing `None` to a `SparseStack` is only a 1-byte write, and `SparseStack::push` is very optimized. So personally I think the small perf cost is worth it for the greater safety.

However... of course it's a problem if some other transform *has* converted a non-async class method to an async one between `enter_function` and `exit_function` because our transform won't work correctly. But we can catch that in `exit_function` with a debug assertion:

```rs
// Check that another transform hasn't altered value of `func.r#async` since our `enter_function` visitor
debug_assert!(func.r#async == super_methods.is_some());
```

### Don't pass `super_methods` to `insert_variable_statement_at_the_top_of_statements`

I think you can achieve your goal of not having to pass `super_methods` to `insert_variable_statement_at_the_top_of_statements` by changing:

https://github.com/oxc-project/oxc/blob/335065d8c8f5d2995bc3c4e66cec1d4baf482f0b/crates/oxc_transformer/src/common/arrow_function_converter.rs#L1110-L1114

to:

```rs
if let Some(super_methods) = self.super_methods_stack.take_last() {
  // same as before
}
```

Then call `self.super_methods_stack.pop()` in `exit_function` *after* calling `insert_variable_statement_at_the_top_of_statements`. At that point the popped value will always be `None`, and you can discard it.

You can also do the same with `this_var` and `arguments_var`.

Dunqing:
close: #8385

This PR is to solve the missing `super` transform in the async arrow function, and `_this = this` inserts to an incorrect place. These problems are all about the async arrow function, which is a part of the init of class property. Learn more at #8387. 

The output matches Babel's output except for static prop as Babel transforms incorrectly.

### DIFF
--- a/crates/oxc_transformer/src/common/arrow_function_converter.rs
+++ b/crates/oxc_transformer/src/common/arrow_function_converter.rs
@@ -144,7 +144,8 @@ pub struct ArrowFunctionConverter<'a> {
     renamed_arguments_symbol_ids: FxHashSet<SymbolId>,
     // TODO(improve-on-babel): `FxHashMap` would suffice here. Iteration order is not important.
     // Only using `FxIndexMap` for predictable iteration order to match Babel's output.
-    super_methods_stack: SparseStack<FxIndexMap<SuperMethodKey<'a>, SuperMethodInfo<'a>>>,
+    super_methods_stack: NonEmptyStack<FxIndexMap<SuperMethodKey<'a>, SuperMethodInfo<'a>>>,
+    super_needs_transform_stack: NonEmptyStack<bool>,
 }
 
 impl ArrowFunctionConverter<'_> {
@@ -164,7 +165,8 @@ impl ArrowFunctionConverter<'_> {
             constructor_super_stack: NonEmptyStack::new(false),
             arguments_needs_transform_stack: NonEmptyStack::new(false),
             renamed_arguments_symbol_ids: FxHashSet::default(),
-            super_methods_stack: SparseStack::new(),
+            super_methods_stack: NonEmptyStack::new(FxIndexMap::default()),
+            super_needs_transform_stack: NonEmptyStack::new(false),
         }
     }
 }
@@ -198,10 +200,12 @@ impl<'a> Traverse<'a> for ArrowFunctionConverter<'a> {
         debug_assert!(self.constructor_super_stack.len() == 1);
         // TODO: This assertion currently failing because we don't handle `super` in arrow functions
         // in class static properties correctly.
-        // e.g. `class C { static f = async () => super.prop; }`
+        // e.g. `class C { static f = () => super.prop; }`
         // debug_assert!(self.constructor_super_stack.last() == &false);
         debug_assert!(self.super_methods_stack.len() == 1);
-        debug_assert!(self.super_methods_stack.last().is_none());
+        debug_assert!(self.super_methods_stack.last().is_empty());
+        debug_assert!(self.super_needs_transform_stack.len() == 1);
+        debug_assert!(self.super_needs_transform_stack.last() == &false);
     }
 
     fn enter_function(&mut self, func: &mut Function<'a>, ctx: &mut TraverseCtx<'a>) {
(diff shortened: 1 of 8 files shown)


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8435?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#8435** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8387** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8387?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8435).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):

overlookmotel:
### Merge activity

* **Jan 13, 6:30 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
@Dunqing I've fixed merge conflicts on #8387 below locally, but cannot push because Github is currently having an outage. I'll try again in the morning.

PR #8436, opened 2025-01-11
refactor(linter): add output formatter

Sysix:
I want to start grouping all the different Formats for action X, Y and Z into own place.
This is the first step and probably not the best one to be honest :)

~~I prefer that the `OutputFormatter` is a part of `oxlint` and not `oxc_linter`~~
~~but all `use create::rules::RULES` is not public from outside.~~
EDIT: I pushed a commit with this changes can easily be reverted and move back to `oxc_linter`

Also their is a crate `oxc_diagnostics` which has already the concept too but only for the lint part of `oxlint` and not for the other parts. The next goal would be splitting the `DiagnosticService` with its reporters to `oxlint`.


crates/oxc_linter/src/lib.rs
```diff
@@ -15,17 +15,16 @@ mod module_graph_visitor;
 mod module_record;
 mod options;
 mod rule;
-mod rules;
 mod service;
 mod utils;
 
 pub mod loader;
+pub mod rules;
```
Sysix, line 22: ⚠️ I need `apps/oxlint` to get `RULES`

PR #8449, opened 2025-01-12
perf(codegen): always inline `wrap`

Boshen:
Function call overhead seems to be expensive for this hot path.


Boshen:
no difference.

PR #8454, opened 2025-01-12
refactor(linter): move DiagnosticsReporters to oxlint

Sysix:
it was never the plan that oxc_diagnostics should be the `std::io::stdout` writer:
https://github.com/oxc-project/oxc/blob/8fc238ac34462945cac66dfa3c5c5f7993cc7df4/crates/oxc_diagnostics/src/service.rs#L84-L85

This PR does refactor all reporters to `oxlint` crate and make the current implementation of `oxc_diagnostics` public for others to consume.

I added some tests to reflect to expected output (and found some bugs^^)

For the future I think the BufWriter for `std::io::stdout` should come from outside.   
Or maybe `std::fmt::stdout`? I do not know^^"

I could not test 100% of the code, I hope I can fix this with the next PR which will include a own Tester for oxlint (like `oxc_linter`). Please be extra careful when reviewing it.


apps/oxlint/src/output_formatter/github.rs
```diff
@@ -79,3 +81,33 @@ fn escape_property(value: &str) -> String {
...
+    fn reporter_error() {
+        let mut reporter = GithubReporter;
+        let error = OxcDiagnostic::warn("error message")
+            .with_label(Span::new(0, 8))
+            .with_source_code(NamedSource::new("file://test.ts", "debugger;"));
+
+        let result = reporter.render_error(error);
+
+        assert!(result.is_some());
+        assert_eq!(result.unwrap(), "::warning file=file%3A//test.ts,line=1,endLine=1,col=1,endColumn=1,title=oxlint::error message\n");
```
Sysix, line 111: `endColumn=1` is wrong, `oxc_diagnostics::reporter:Info` does not contain and end column

PR #8455, opened 2025-01-12
fix(semantic)!: ensure program outlives semantic

### LINKED ISSUE #8437: ast: Out of bounds memory access crash in AST Vititor
## Summary
After adding some unrelated code to a function in my project `oxc::ast::Visit::visit_program` started crashing with out-of-bounds memory error:
```
error: process didn't exit successfully: `target\debug\test-runner.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

### Stack Trace
Attaching a debugger to the program I got the following stack trace:
```rust
test-runner.exe!oxc_span::atom::Atom::as_str() Line 34 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_span-0.44.0\src\atom.rs:34)
test-runner.exe!oxc_span::atom::impl$17::eq<ref$<str$>>(oxc_span::atom::Atom * self, ref$<str$> * other) Line 159 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_span-0.44.0\src\atom.rs:159)
test-runner.exe!oxc_ast::ast::js::Directive::is_use_strict() Line 787 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_ast-0.44.0\src\ast_impl\js.rs:787)
test-runner.exe!core::ops::function::FnMut::call_mut<bool (*)(ref$<oxc_ast::ast::js::Directive>),tuple$<ref$<oxc_ast::ast::js::Directive>>>(bool(*)(oxc_ast::ast::js::Directive *) *, oxc_ast::ast::js::Directive *) Line 166 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:166)
test-runner.exe!core::slice::iter::impl$182::any<oxc_ast::ast::js::Directive,bool (*)(ref$<oxc_ast::ast::js::Directive>)>(core::slice::iter::Iter<oxc_ast::ast::js::Directive> * self, bool(*)(oxc_ast::ast::js::Directive *) f) Line 285 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\slice\iter\macros.rs:285)
test-runner.exe!oxc_ast::ast::js::Program::has_use_strict_directive() Line 20 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_ast-0.44.0\src\ast_impl\js.rs:20)
test-runner.exe!oxc_ast::generated::visit::walk::walk_program<test_runner::TypeVisitorImpl<core::iter::adapters::flatten::FlatMap<core::slice::iter::Iter<alloc::vec::Vec<test_runner::Assertion,alloc::alloc::Global>>,core::slice::iter::Iter<test_runner::Assertion>,test_runner::impl$0::run::closure_env$1>>>(test_runner::TypeVisitorImpl<core::iter::adapters::flatten::FlatMap<core::slice::iter::Iter<alloc::vec::Vec<test_runner::Assertion,alloc::alloc::Global>>,core::slice::iter::Iter<test_runner::Assertion>,test_runner::impl$0::run::closure_env$1>> * visitor, oxc_ast::ast::js::Program * it) Line 1347 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_ast-0.44.0\src\generated\visit.rs:1347)
test-runner.exe!oxc_ast::generated::visit::Visit::visit_program<test_runner::TypeVisitorImpl<core::iter::adapters::flatten::FlatMap<core::slice::iter::Iter<alloc::vec::Vec<test_runner::Assertion,alloc::alloc::Global>>,core::slice::iter::Iter<test_runner::Assertion>,test_runner::impl$0::run::closure_env$1>>>(test_runner::TypeVisitorImpl<core::iter::adapters::flatten::FlatMap<core::slice::iter::Iter<alloc::vec::Vec<test_runner::Assertion,alloc::alloc::Global>>,core::slice::iter::Iter<test_runner::Assertion>,test_runner::impl$0::run::closure_env$1>> * self, oxc_ast::ast::js::Program * it) Line 52 (c:\Users\valja\.cargo\registry\src\index.crates.io-6f17d22bba15001f\oxc_ast-0.44.0\src\generated\visit.rs:52)
test-runner.exe!test_runner::TypeVisitor::run() Line 59 (c:\Users\valja\source\rust\rustify\crates\test_runner\src\main.rs:59)
test-runner.exe!test_runner::main() Line 15 (c:\Users\valja\source\rust\rustify\crates\test_runner\src\main.rs:15)
test-runner.exe!core::ops::function::FnOnce::call_once<void (*)(),tuple$<>>(void(*)()) Line 250 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:250)
test-runner.exe!std::sys::backtrace::__rust_begin_short_backtrace<void (*)(),tuple$<>>(void(*)() f) Line 160 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\sys\backtrace.rs:160)
test-runner.exe!std::rt::lang_start::closure$0<tuple$<>>(std::rt::lang_start::closure_env$0<tuple$<>> *) Line 195 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\rt.rs:195)
[Inline Frame] test-runner.exe!std::rt::lang_start_internal::closure$1() Line 174 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\rt.rs:174)
[Inline Frame] test-runner.exe!std::panicking::try::do_call() Line 557 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\panicking.rs:557)
[Inline Frame] test-runner.exe!std::panicking::try() Line 520 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\panicking.rs:520)
[Inline Frame] test-runner.exe!std::panic::catch_unwind() Line 358 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\panic.rs:358)
test-runner.exe!std::rt::lang_start_internal() Line 174 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\rt.rs:174)
test-runner.exe!std::rt::lang_start<tuple$<>>(void(*)() main, __int64 argc, unsigned char * * argv, unsigned char sigpipe) Line 194 (c:\Users\valja\.rustup\toolchains\1.84.0-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\std\src\rt.rs:194)
test-runner.exe!main() (Unknown Source:0)
```

It seems that `self.directives` is NULL here:
https://github.com/oxc-project/oxc/blob/01722f34ff5d9d118f9940574c4d9f40eacf826e/crates/oxc_ast/src/ast_impl/js.rs#L17-L20
```rust
{__0={buf={ptr=NonNull(0x0000000000000008: 0x0000000000000008 {span={start=??? end=??? } expression={span={start=??? end=??? } value={__0=??? } ...} ...}) ...} ...} }
```

<details>
<summary>

## Code to reproduce (click to expand)

</summary>

```rust
use oxc::{
  allocator::{self, Allocator, IntoIn},
  ast::{AstKind, Visit},
  semantic::Semantic,
};

/// # Panics
pub fn main() {
  let alloc = oxc::allocator::Allocator::default();
  let semantic = parse_file("test.ts", &alloc);
  let visitor = TypeVisitor {
    semantic: &semantic,
    assertions: vec![vec![Assertion { expr: "Foo", expected_type: "Foo" }]],
  };
  visitor.run();
}

fn parse_file<'a>(path: &'_ str, alloc: &'a Allocator) -> Semantic<'a> {
  let source_text: allocator::String = "class Foo {}".to_string().into_in(alloc);
  let source_text = source_text.into_bump_str();
  let source_type = oxc::span::SourceType::from_path(path).unwrap();
  let parser = oxc::parser::Parser::new(alloc, source_text, source_type);

  let parse_result = parser.parse();

  let program = parse_result.program;
  let builder = oxc::semantic::SemanticBuilder::new().with_check_syntax_error(true);
  let semantic_result = builder.build(&program);

  semantic_result.semantic
}

pub struct Assertion<'a> {
  pub expr: &'a str,
  pub expected_type: &'a str,
}

pub struct TypeVisitor<'a> {
  pub semantic: &'a Semantic<'a>,
  pub assertions: Vec<Vec<Assertion<'a>>>,
}

impl TypeVisitor<'_> {
  /// # Panics
  pub fn run(&self) {
    // !!! Removing this statement fixes the crash
    let _ = self.assertions.iter().flat_map(|x| x.iter().map(|x| x.expr)).collect::<String>();

    let source_text = self.semantic.source_text();
    let assertions = self.assertions.iter().flat_map(|x| x.iter());
    let mut visitor = TypeVisitorImpl { _source_text: source_text, _assertions: assertions };
    let AstKind::Program(program) =
      self.semantic.nodes().root_node().expect("root node to exist").kind()
    else {
      panic!("Expected root AST node to be Program");
    };
    visitor.visit_program(program);
  }
}

struct TypeVisitorImpl<'a, T: Iterator<Item = &'a Assertion<'a>>> {
  _source_text: &'a str,
  _assertions: T,
}

impl<'a, T: Iterator<Item = &'a Assertion<'a>>> Visit<'a> for TypeVisitorImpl<'a, T> {}
```

</details>

## System Details

- `oxc: 0.44.0`
- ```
  rustc 1.84.0 (9fc6b4312 2025-01-07)
  binary: rustc
  commit-hash: 9fc6b43126469e3858e2fe86cafb4f0fd5068869
  commit-date: 2025-01-07
  host: x86_64-pc-windows-msvc
  release: 1.84.0
  LLVM version: 19.1.5
  ```
- `os: Windows 11`

valeneiko:
fixes: oxc-project/oxc#8437 

In semantic builder make sure `Program` reference has a lifetime of the Arena.

### DIFF
--- a/crates/oxc_semantic/src/builder.rs
+++ b/crates/oxc_semantic/src/builder.rs
@@ -223,7 +223,7 @@ impl<'a> SemanticBuilder<'a> {
     /// Finalize the builder.
     ///
     /// # Panics
-    pub fn build(mut self, program: &Program<'a>) -> SemanticBuilderReturn<'a> {
+    pub fn build(mut self, program: &'a Program<'a>) -> SemanticBuilderReturn<'a> {
         self.source_text = program.source_text;
         self.source_type = program.source_type;
         if self.build_jsdoc {
--- a/tasks/coverage/src/driver.rs
+++ b/tasks/coverage/src/driver.rs
@@ -1,10 +1,10 @@
-use std::{ops::ControlFlow, path::PathBuf};
+use std::{hint::unreachable_unchecked, ops::ControlFlow, path::PathBuf};
 
 use rustc_hash::FxHashSet;
 
 use oxc::{
     allocator::Allocator,
-    ast::{ast::Program, Comment},
+    ast::{ast::Program, AstKind, Comment},
     codegen::{CodegenOptions, CodegenReturn},
     diagnostics::OxcDiagnostic,
     minifier::CompressOptions,
@@ -80,12 +80,12 @@ impl CompilerInterface for Driver {
         ControlFlow::Continue(())
     }
 
-    fn after_semantic(
-        &mut self,
-        program: &mut Program<'_>,
-        ret: &mut SemanticBuilderReturn,
-    ) -> ControlFlow<()> {
+    fn after_semantic(&mut self, ret: &mut SemanticBuilderReturn) -> ControlFlow<()> {
         if self.check_semantic {
+            let AstKind::Program(program) = ret.semantic.nodes().root_node().unwrap().kind() else {
+                // SAFETY: root_node is guaranteed to be Program
+                unsafe { unreachable_unchecked() };
+            };
             if let Some(errors) = check_semantic_ids(program) {
                 self.errors.extend(errors);
                 return ControlFlow::Break(());
--- a/tasks/coverage/src/driver.rs
+++ b/tasks/coverage/src/driver.rs
@@ -1,4 +1,4 @@
-use std::{hint::unreachable_unchecked, ops::ControlFlow, path::PathBuf};
+use std::{ops::ControlFlow, path::PathBuf};
 
 use rustc_hash::FxHashSet;
 
@@ -86,8 +86,7 @@ impl CompilerInterface for Driver {
                 return ControlFlow::Break(());
             };
             let AstKind::Program(program) = root_node.kind() else {
-                // SAFETY: root_node is guaranteed to be Program
-                unsafe { unreachable_unchecked() };
+                return ControlFlow::Break(());
             };
             if let Some(errors) = check_semantic_ids(program) {
                 self.errors.extend(errors);
--- a/tasks/coverage/src/driver.rs
+++ b/tasks/coverage/src/driver.rs
@@ -1,10 +1,10 @@
-use std::{ops::ControlFlow, path::PathBuf};
+use std::{hint::unreachable_unchecked, ops::ControlFlow, path::PathBuf};
 
 use rustc_hash::FxHashSet;
 
 use oxc::{
     allocator::Allocator,
-    ast::{ast::Program, Comment},
+    ast::{ast::Program, AstKind, Comment},
     codegen::{CodegenOptions, CodegenReturn},
     diagnostics::OxcDiagnostic,
     minifier::CompressOptions,
@@ -80,12 +80,15 @@ impl CompilerInterface for Driver {
         ControlFlow::Continue(())
     }
 
-    fn after_semantic(
-        &mut self,
-        program: &mut Program<'_>,
-        ret: &mut SemanticBuilderReturn,
-    ) -> ControlFlow<()> {
+    fn after_semantic(&mut self, ret: &mut SemanticBuilderReturn) -> ControlFlow<()> {
         if self.check_semantic {
+            let Some(root_node) = ret.semantic.nodes().root_node() else {
+                return ControlFlow::Break(());
+            };
+            let AstKind::Program(program) = root_node.kind() else {
+                // SAFETY: root_node is guaranteed to be Program
+                unsafe { unreachable_unchecked() };
+            };
             if let Some(errors) = check_semantic_ids(program) {
                 self.errors.extend(errors);
                 return ControlFlow::Break(());
--- a/tasks/coverage/src/driver.rs
+++ b/tasks/coverage/src/driver.rs
@@ -1,10 +1,10 @@
-use std::{hint::unreachable_unchecked, ops::ControlFlow, path::PathBuf};
+use std::{ops::ControlFlow, path::PathBuf};
 
 use rustc_hash::FxHashSet;
 
 use oxc::{
     allocator::Allocator,
-    ast::{ast::Program, AstKind, Comment},
+    ast::{ast::Program, Comment},
     codegen::{CodegenOptions, CodegenReturn},
     diagnostics::OxcDiagnostic,
     minifier::CompressOptions,
@@ -80,12 +80,12 @@ impl CompilerInterface for Driver {
         ControlFlow::Continue(())
     }
 
-    fn after_semantic(&mut self, ret: &mut SemanticBuilderReturn) -> ControlFlow<()> {
+    fn after_semantic(
+        &mut self,
+        program: &mut Program<'_>,
+        ret: &mut SemanticBuilderReturn,
+    ) -> ControlFlow<()> {
         if self.check_semantic {
-            let AstKind::Program(program) = ret.semantic.nodes().root_node().unwrap().kind() else {
-                // SAFETY: root_node is guaranteed to be Program
-                unsafe { unreachable_unchecked() };
-            };
             if let Some(errors) = check_semantic_ids(program) {
                 self.errors.extend(errors);
                 return ControlFlow::Break(());
--- a/crates/oxc/README.md
+++ b/crates/oxc/README.md
@@ -103,7 +103,7 @@ let SemanticBuilderReturn {
     .with_check_syntax_error(true) // Enable extra syntax error checking
     .with_build_jsdoc(true)        // Enable JSDoc parsing
     .with_cfg(true)                // Build a Control Flow Graph
-    .build(program);              // Produce the `Semantic`
+    .build(&program);              // Produce the `Semantic`
 
 errors.extend(semantic_errors);
 if errors.is_empty() {
--- a/crates/oxc/README.md
+++ b/crates/oxc/README.md
@@ -103,7 +103,7 @@ let SemanticBuilderReturn {
     .with_check_syntax_error(true) // Enable extra syntax error checking
     .with_build_jsdoc(true)        // Enable JSDoc parsing
     .with_cfg(true)                // Build a Control Flow Graph
-    .build(&program);              // Produce the `Semantic`
+    .build(program);              // Produce the `Semantic`
 
 errors.extend(semantic_errors);
 if errors.is_empty() {
--- a/crates/oxc/src/compiler.rs
+++ b/crates/oxc/src/compiler.rs
@@ -101,11 +101,7 @@ pub trait CompilerInterface {
         ControlFlow::Continue(())
     }
 
-    fn after_semantic(
-        &mut self,
-        _program: &mut Program<'_>,
-        _semantic_return: &mut SemanticBuilderReturn,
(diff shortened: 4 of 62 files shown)


tasks/coverage/src/driver.rs
```diff
@@ -80,12 +80,12 @@ impl CompilerInterface for Driver {
...
-        &mut self,
-        program: &mut Program<'_>,
-        ret: &mut SemanticBuilderReturn,
-    ) -> ControlFlow<()> {
+    fn after_semantic(&mut self, ret: &mut SemanticBuilderReturn) -> ControlFlow<()> {
         if self.check_semantic {
+            let AstKind::Program(program) = ret.semantic.nodes().root_node().unwrap().kind() else {
+                // SAFETY: root_node is guaranteed to be Program
+                unsafe { unreachable_unchecked() };
+            };
```
valeneiko, line 88: `root_node().unwrap()` fails here: https://github.com/oxc-project/oxc/actions/runs/12736474264/job/35496404900?pr=8455#step:7:73

But I am not sure what the correct action should be. Afterall, before my changes program should have existed, so I just assumed it would still exist.

Boshen:
`pub program: &'a mut Program<'a>` from `ParserReturn` doesn't feel right, let's see if @overlookmotel knows what the underly problem is.

overlookmotel:
I've written up my thoughts on this problem in https://github.com/oxc-project/oxc/issues/8437#issuecomment-2586884874 and #8461.

Personally, I quite like the idea of parser returning `&mut Program`. BUT that's a large breaking change, and I don't think it's necessary to fix this issue. I think just changing `SemanticBuilder::build` to take a `&'a Program<'a>` should be sufficient to squash the bug.

So I suggest we go one step at a time - fix #8437 first, then discuss API changes after. I have some other thoughts about API, and if we're going to make breaking changes, it'd be preferable to do that only once.

Or have I misunderstood? Is parser returning `&mut Program` *necessary* to fix the bug?

valeneiko:
Changing parser return type is not necessary. But we would still need to update all usages of semantic to first copy the program into the allocator and then use the new reference.
```rust
let program = alloc.alloc(parserReturn.program);
let semanticReturn = SemanticBuilder::new().build(program);
```

valeneiko:
@overlookmotel I've updated the PR to just fix semantic.

tasks/coverage/src/driver.rs
overlookmotel: Just out of interest, did you figure out what the problem was here?

valeneiko: No. Just replaced unwrap with a return statement and all tests passed. Didn't look into the reason why it was needed.

crates/oxc_semantic/src/builder.rs
```diff
@@ -223,7 +223,7 @@ impl<'a> SemanticBuilder<'a> {
     /// Finalize the builder.
     ///
     /// # Panics
-    pub fn build(mut self, program: &Program<'a>) -> SemanticBuilderReturn<'a> {
```
Boshen, line 226: Beginners will have trouble getting the lifetime `&'a`, I think we need to document how to get it on this API.

overlookmotel (commented):
Thanks very much for doing this, and thanks for reverting the changes to `ParserReturn`.

Unfortunately this does make the API harder to use, but that's necessary to make it correct. `SemanticBuilderReturn` *does* borrow the `Program` for the duration of its existence, so we need to make sure it does actually work that way.

I've pushed 1 extra commit to remove `unreachable_unchecked`. The purpose of that part of the coverage runner is to make sure that `Semantic` contains what it should do, so it's best not to make unsafe assertions about the very thing we're trying to check.

Apart from that, great!

This is a breaking change, so before we merge this we should make sure it doesn't break downstream consumers in a terrible way. I'll try it out with Rolldown now, and will report back.

tasks/coverage/src/driver.rs
overlookmotel: I took a look. There is something wrong - that method should never be called if semantic failed, so the unwrap should never fail. But that's a separate issue, and no need to resolve it right now.

overlookmotel (approved):
Not causing major problems in Rolldown. Given that's a large codebase which makes heavy use of Oxc, likely it's not going to cause too many problems in other downstream projects either.

crates/oxc_semantic/src/builder.rs
overlookmotel: To be honest, I wouldn't be sure how to document this. I'm not sure I entirely get the lifetimes myself.

But I *think* it's not as simple as "`SemanticBuilderReturn<'a>` borrows the `Program` for the lifetime of the allocator" because `build` doesn't know what the lifetime `'a` means. It just knows that `SemanticBuilderReturn<'a>` borrows `&Program` for `'a`, but `'a` can be anything. `Program` just has to live at least as long as `SemanticBuilderReturn` does.

So often you don't have to call `allocator.alloc(program)`. Often the compiler will figure out how to stretch the borrow accordingly to make your code compile.

None of the examples for parser, semantic, transformer etc need any changes. They all still compile fine after this PR.

What you *can't* do any more is pass an owned `Program` and an owned `SemanticBuilderReturn` to another function, because `SemanticBuilderReturn` borrows `Program` (so the 2 together is like a self-referential struct).

But you *can* pass `&Program` and owned `SemanticBuilderReturn` to another function, or passing `&Program` and `&SemanticBuilderReturn` also works.

This is definitely an unwelcome complication to the API, but I'm not sure how to explain it simply. I think we might be better off merging this as is and then seeing what problems people come to us with on Discord. It might be easier to figure out how to explain it when we have a concrete example of a problem.

david1234121 (commented):
DavidMashby.com

PR #8479, opened 2025-01-14
feat(span): add `Atom::r#static`

kermanx:
This allows

```rs
const TYPEOF_UNDEFINED: Atom<'static> = Atom::r#static("undefined");

#[derive(Clone, Copy)]
enum Value<'a> {
  /// Not `Atom<'a>`, which takes one more byte
  StringLiteral(&'a Atom<'a>),
  // ...
}

fn get_typeof<'a>(value: Value<'a>) -> Value<'a> {
  if ... {
    return Value::StringLiteral(&TYPEOF_UNDEFINED);
  }
}
```


overlookmotel:
@KermanX Please see #8480.

PR #8486, opened 2025-01-14
refactor(semantic): remove Ambient symbol flag

ShuiRuTian:
#7278

The preparation for 
- #8038 
- #8039

This approach is agreed by the origin committer https://github.com/oxc-project/oxc/issues/7278#issuecomment-2556698726

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_unused_vars/mod.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_unused_vars/mod.rs
@@ -337,8 +337,7 @@ impl NoUnusedVars {
     }
 
     fn should_skip_symbol(symbol: &Symbol<'_, '_>) -> bool {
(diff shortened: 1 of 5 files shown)


Boshen:
This change breaks a few namespace tests:

```ts
+Expect to Parse: tasks/coverage/typescript/tests/cases/compiler/mixingFunctionAndAmbientModule1.ts
+
+  × Identifier `My` has already been declared
+   ╭─[typescript/tests/cases/compiler/mixingFunctionAndAmbientModule1.ts:2:20]
+ 1 │ module A {
+ 2 │     declare module My {
+   ·                    ─┬
+   ·                     ╰── `My` has already been declared here
+ 3 │         export var x: number;
+ 4 │     }
+ 5 │     function My(s: string) { }
+   ·              ─┬
+   ·               ╰── It can not be redeclared here
+ 6 │ }
+   ╰────
```

PR #8500, opened 2025-01-15
fix(transformer/async-to-generator): move parameters to the inner generator function when they could throw errors

Dunqing:
The new implementation port from [esbuild](https://github.com/evanw/esbuild/blob/df815ac27b84f8b34374c9182a93c94718f8a630/internal/js_parser/js_parser_lower.go#L355-L467), before from `Babel`. 

Babel's transform implementation for the async method is incorrect because the async method should return a rejecting promise when it throws an error. Everything is good if the errors are thrown in the async method body, but the following case will throw an error in the parameters which causes the whole program crushed not a rejecting promise. So we should move the parameters to the inner generator function when the parameters could throw an error.

Input: 
```js
class Cls {
  // ReferenceError: Cannot access 'b' before initialization
  async method(a = b, b = 0) {}
}
```

Before output
```js
class Cls {
  method(a = b, b = 0) {
    return babelHelpers.asyncToGenerator(function* () {})();
  }
}
```

After output:
```js
class Cls {
  method() {
     // ReferenceError: Cannot access 'b' before initialization
    return babelHelpers.asyncToGenerator(function* (a = b, b = 0) {}).apply(this, arguments);
  }
}
```

No override tests because Babel doesn't cover this case.

### DIFF
--- a/crates/oxc_transformer/src/es2017/async_to_generator.rs
+++ b/crates/oxc_transformer/src/es2017/async_to_generator.rs
@@ -220,23 +220,70 @@ impl<'a, 'ctx> AsyncGeneratorExecutor<'a, 'ctx> {
             return;
         };
 
+        // If parameters could throw errors, we need to move them to the inner function,
+        // because it is an async function, which should return a rejecting promise if
+        // there is an error.
+        let needs_move_parameters_to_inner_function =
+            Self::could_throw_errors_parameters(&func.params);
+
         let (generator_scope_id, wrapper_scope_id) = {
             let new_scope_id = ctx.create_child_scope(ctx.current_scope_id(), ScopeFlags::Function);
             let scope_id = func.scope_id.replace(Some(new_scope_id)).unwrap();
             // We need to change the parent id to new scope id because we need to this function's body inside the wrapper function,
             // and then the new scope id will be wrapper function's scope id.
             ctx.scopes_mut().change_parent_id(scope_id, Some(new_scope_id));
-            // We need to transform formal parameters change back to the original scope,
-            // because we only move out the function body.
-            Self::move_formal_parameters_to_target_scope(new_scope_id, &func.params, ctx);
+            if !needs_move_parameters_to_inner_function {
+                // We need to change formal parameters's scope back to the original scope,
+                // because we only move out the function body.
+                Self::move_formal_parameters_to_target_scope(new_scope_id, &func.params, ctx);
+            }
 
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#8500** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8500).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
I noticed a potential improvement, we don't need to change `this` to `_this` in the async method and async arrow function because both `asyncToGenerator` and `wrapAsyncGenerator` are binding the `this` for the inner generator function. However this will cause a lot of tests to fail, thus we need to override them, this is not a good moment to change this.

overlookmotel (approved):
Nice!

overlookmotel:
### Merge activity

* **Jan 15, 7:01 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
One things for follow-up: There are more things beyond `AssignmentPattern` which can throw. e.g.:

```js
const throwIt = () => { throw new Error('gotcha!'); };

const constant = 1;

Object.defineProperty(global, 'foo', { set(v) {} }); // No getter

class C {
    async method1( [a = b], b = 0 ) {}
    async method2( {x: a = b}, b = 0 ) {}
    async method3( { [throwIt()]: a } ) {}
    async method4( { [constant = 2]: a } ) {}
    async method5( { [foo]: a } ) {} // Just trying to read `global.foo` throws an error!
    async method6( a = foo ) {} // Ditto
}
```

It might be best to conservatively assume that any computed keys in object destructuring can throw - they'll be rare anyway.

Any accesses of unbound variables can also throw, and no way to know statically if they will or not.

overlookmotel:
Force pushed to resolve merge conflicts in snapshots.

PR #8519, opened 2025-01-15
refactor(ast, regular_expression): shorten `ContentEq` implementations

overlookmotel:
Shorten codegen-ed implementations of `ContentEq` for enums, matching what Rust produces for `#[derive(PartialEq)]` (inspired by #8517).

### DIFF
--- a/crates/oxc_ast/src/generated/derive_content_eq.rs
+++ b/crates/oxc_ast/src/generated/derive_content_eq.rs
@@ -47,175 +47,57 @@ impl ContentEq for Program<'_> {
 
 impl ContentEq for Expression<'_> {
     fn content_eq(&self, other: &Self) -> bool {
-        match self {
-            Self::BooleanLiteral(it) => match other {
-                Self::BooleanLiteral(other) if ContentEq::content_eq(it, other) => true,
-                _ => false,
-            },
-            Self::NullLiteral(it) => match other {
-                Self::NullLiteral(other) if ContentEq::content_eq(it, other) => true,
-                _ => false,
-            },
-            Self::NumericLiteral(it) => match other {
-                Self::NumericLiteral(other) if ContentEq::content_eq(it, other) => true,
-                _ => false,
-            },
-            Self::BigIntLiteral(it) => match other {
-                Self::BigIntLiteral(other) if ContentEq::content_eq(it, other) => true,
-                _ => false,
-            },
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#8519** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8519?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8519).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
@camc314 I half-hoped this would produce a speed-up too, but no. I think `ContentEq` is just inherently expensive, especially when the 2 AST fragments being compared are large and *mostly* the same, so you have to crawl a long way through the structures to find out if they're the same or not.

Boshen:
No change in benchmark, so merging.

Boshen:
### Merge activity

* **Jan 17, 12:58 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jan 17, 12:58 AM EST**: A user added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

overlookmotel:
> it might be worth double checking this. but i think this approach has worse performance than before. If we consider "left" and "other", if "left" matches the first enum variant on the first thing of the match statement, and right does not, it will try ever match pattern rather than bailing out early. << this might not be correct, but it's what i understood from godbolt and some rust forums

I would assume compiler is smart enough to see that each match arm is mutually exclusive, and so will turn it into "exit if discriminants are not the same, then jump table" rather than looping through each condition. I would have thought we can safely assume that what Rust produces for `PartialEq` is optimal. If it wasn't, it would have been corrected long ago, as it's such a common construct and critical for performance.

That's all assumption though. I'd be interested to see what you saw in Godbolt which suggested otherwise.

Issue #8350, opened 2025-01-08
Incorrect data in `SymbolTable` for TS enums


overlookmotel:
Possibly relevant to #8342. The way we record references between TS enum fields in `SymbolTable` is wonky.

### The problem

This is legal:

```ts
enum Foo {
  A = 123,
  B = A, // `A` evaluates to 123
}

enum Foo {
  C = A, // `A` evaluates to 123
}
```

[Playground](https://playground.oxc.rs/#eNpVTs1OwzAMfpXI5wixIS6VOAyknnmAXrLgTpESu7LTwVT13XFWiuCU74u/vwUidIA0F9czu2Ug507uxR2OT77hV8Mn7wZaBxrov+7tzw08MHQLyEzt0RvV8AVdlRk95EQVujFkNaKRJ9wveitnzjurEkhHlrJ9rB6mIIpyT+RZIvYpI4Vifqio9aGqFU+CJrriexAk/Skyc6u9mw3/Rm9pNcgFbROgHh8Pz2CKyB94wbbeSEmUxrSbI1MVzn3mz2a+opxZbcNWtK7fLi1t9g==)

i.e. it's essentially the same as:

```ts
enum Foo {
  A = 123,
  B = A, // `A` evaluates to 123
  C = A, // `A` evaluates to 123
}
```

Our output from transformer for this case is correct, but the data in `SymbolTable` is not:

* `A` here gets a `Symbol`. But it only has 1 reference, not 2.
* `A` in 2nd enum block has a `ReferenceId`, but that reference has no `SymbolId` (i.e. `SymbolTable` says it resolves to `global.A`.

If you add `let A = 'hello';` at top level, then `Reference` for `A` in 2nd enum block has `SymbolId` pointing to `let A` (similar to the problem in #8342).

### Why this matters

1. If we can correctly resolve identifiers in `Semantic` it'd make solving #8342 trivial.
2. Correct symbol resolution matters in the linter for e.g. `no-unused-vars` rule.
3. We want our semantic data to be correct!

### Why it's hard to solve

The binding of enum fields does not follow our model for symbols/reference. It's unlike anything else.

Currently the `Symbol` for `A` is bound in the scope for 1st enum's body block. When there's only 1 enum block for an enum, this makes sense. That binding is not visible outside the block. But when there are 2 or more `enum {}` blocks comprising the enum, it breaks down.

So we need to hoist the binding up to enclosing scope (top level scope in this case) so it is in scope of both `enum` blocks.

[...]

PR #8543, opened 2025-01-16
fix(transformer/typescript): correctly resolve references to non-constant enum members

### LINKED ISSUE #8342: transformer: tsc vs babel on enum transformation
### Discussed in https://github.com/oxc-project/oxc/discussions/8341

<div type='discussions-op-text'>

<sup>Originally posted by **branchseer** January  8, 2025</sup>
oxc-transform currently follows babel's rules on enum transformation. IMHO it's a little bit weird about whether an identifier should be resolved to the enum member:

https://playground.oxc.rs/#eNqNkcFqwzAMhl9F+NIEwroOdunooD0Udhh7gVycREkNjlRsp11X8u6Tk6UbXRk7+Zcs/f+HfValWqqckiSF1TOccwI4aAdrWMFsh9by7Cn2kLoWtszjBAz3i2zUG9HrDOZz6RoPDj3bA1YQGMIOx9UW2wIdmBpMmHkomXzQFKJBn1OfJqnE/BNjIxc/MF512N05TRW3SfoL6U0Q3NF4zGRYiK7puAvC5Uveo3AVhipDzTVWTl/BH2Pw7djv0Jc6ervoSHw7IhsfQjfa0F9PJhQ5qUyxWp6V6yge/kRBv6tlcB1myhoKkx5CLsWpLdhOVRBWX7Nrx0afqb12Hp0Yio4mk75MDg0VtGtQEhT6h/vFo5KJkitsMLJI0RoytZmW5WeDY7u1fIzLB3QFe0GqtfXY9/0nDIDKdw==

In contrast, tsc has a simpler rule:

https://www.typescriptlang.org/play/?#code/PTAEEFQSwZ1BDANgd3gTzgJwKYwPaIBu2AJqAC54UAW2o2AdgK4C2oL2LARtpgHQAoIQAphASlABeAHygA3gNChC8TBCmgA5AHNEeLkk0BuRfWZsAYnioKlSyJNABGADSmlAIQ3gXoUwF8BfzFxEwFRCRl5UxU1By1dfUMTJUZWUA9VaLt1RwBZeHJqPkx4BhI8FnE3HK9HHwCgkLEwgTS2TIAvbNzQAqKSsoqqsRqM7zdAoA

Would it be better to switch to tsc's rule?

</div>

wan9chi:
fixes #8342


crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -199,44 +205,34 @@ impl<'a> TypeScriptEnum<'a> {
...
-                        if !has_binding {
-                            IdentifierReferenceRename::new(
-                                param_binding.name.clone(),
-                                previous_enum_members.clone(),
-                                ctx,
-                            )
-                            .visit_expression(&mut new_initializer);
-                        }
+                        IdentifierReferenceRename::new(
+                            param_binding.name.clone(),
```
branchseer, line 217: Question: any reason not making `Atom` implements `Copy`? `.clone()` always makes me feel the operation is somewhat expensive.

overlookmotel: I agree. https://github.com/oxc-project/backlog/issues/155

Originally we were considering some form of interning and reference-counting, so we didn't make it `Copy` to leave the door open for that. But now all strings are stored in the arena anyway, so even if we did decide to intern strings, reference-counting would be irrelevant.

overlookmotel:
@Dunqing You wrote this transform and are more familiar with it. Would you be able to review please?

crates/oxc_transformer/src/typescript/enum.rs
overlookmotel: A PR would be very welcome. But it'd be boring as hell to do! *So many* `.clone()`s to remove!

branchseer: Thanks for both the answer and the backlog link! There are lots of interesting readings over there:)

overlookmotel: Yeah, it's kind of our "brain dump". We try to keep main issue tracker just for stuff which is being worked on currently, to keep things manageable. But if there's anything in backlog that interests you, please ping me on our [Discord](https://discord.gg/9uXCAwqQZW) and I'll transfer the issue over to main repo so we can discuss.

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -527,56 +525,68 @@ impl<'a> TypeScriptEnum<'a> {
...
+            enum Foo {
+                A,
+                B = (() => {
+                    const A = 1;
+                    return A;
+                })())
+            } */
+            return false;
+        }
+        true
```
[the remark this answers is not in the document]
Dunqing: It looks like as long as the identifier reference refers to an enum member, then we need to transform it, do I understand correctly? If so we only need to check the symbol flags if it is an `EnumMember`.  Please see [OXC Playground]( https://playground.oxc.rs/#eNo9TzFuwzAM/ArByQY8JAW6uEgBZ8grvCguHRiQyYCS2waG/h7KiaxBujvxjuSKA7ZIvMxwEYG1Z4CuyfcZTlBVNZy+X2o+g3CI0NnP8atoSnFRhm4TUl3VPSfouect9Oz0HWqmwx6cWyRsULBdURfOT3hwdP/YRl2oQT9xLDgMcqedPOar+MKiOg6j6PwSUoN3p4HUAg3nkIL3yk3A6PRG1gEpfByOn2gVg/zQjfIsRuaJp3EqZls8qviLl79s/iW9SrCRRucDpZSe7xBo8w==), and switch to the `Symbol` panel, you will see `Foo`'s flag is `EnumMember`.

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -527,56 +525,68 @@ impl<'a> TypeScriptEnum<'a> {
...
+            return true;
+        }
+        if ancestors.any(|scope_id| scope_id == self.enum_scope_id) {
+            /* The symbol is declared inside enum:
+            enum Foo {
+                A,
+                B = (() => {
+                    const A = 1;
+                    return A;
+                })())
```
Dunqing, line 571:
```suggestion
                })()
```

Dunqing (changes requested):
It looks great overall, we can make further improvements after https://github.com/oxc-project/oxc/issues/8350 is done!

crates/oxc_transformer/src/typescript/enum.rs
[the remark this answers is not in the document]
branchseer: > It looks like as long as the identifier reference refers to an enum member, then we need to transform it, do I understand correctly?

Yes that's the rule I want to implement. 

> If so we only need to check the symbol flags if it is an EnumMember.

The `EnumMember` flag seems not sufficient if there's a nested enum ([playground](https://playground.oxc.rs/#eNo9jtFqwzAMRX9F6GUJhHUt7CXQQfrQryijbqq0AVsqsrOtBP/77ATXL7pXPlfSjD22SDw5OIrAfGKADvawbbI6JFVVNey/1h+AhTwYLX6ldwudX050DRS72QB8g7/LZK/8FkCJjSMI99FDEDinne/deaVjLrGu6hMniQ0KtjPqxLn4Jwfzh23QiRq0I4eifS8Pepmnu4gtLqhhP4i6tREbfBj1pGlg0nlI0S9yaWAweqO0AcnvPrafmIhernSjfEsybuRxGEu4Fw4q9mjlN4d/SC/i00mDsZ5ijP8qY3AG)).

But you remind me that it's the symbol's scope we're checking with, not the reference's, so no need to go through the ancestors (I was confused with the [fix for babel](https://github.com/babel/babel/blob/4b2608b4fc240038a731428b249d3ab53acda735/packages/babel-plugin-transform-typescript/src/enum.ts#L178-L181)). A 
 simple comparison with the enum's scope id should suffice.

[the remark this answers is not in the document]
Dunqing: > The `EnumMember` flag seems not sufficient if there's a nested enum ([playground](https://playground.oxc.rs/#eNo9jtFqwzAMRX9F6GUJhHUt7CXQQfrQryijbqq0AVsqsrOtBP/77ATXL7pXPlfSjD22SDw5OIrAfGKADvawbbI6JFVVNey/1h+AhTwYLX6ldwudX050DRS72QB8g7/LZK/8FkCJjSMI99FDEDinne/deaVjLrGu6hMniQ0KtjPqxLn4Jwfzh23QiRq0I4eifS8Pepmnu4gtLqhhP4i6tREbfBj1pGlg0nlI0S9yaWAweqO0AcnvPrafmIhernSjfEsybuRxGEu4Fw4q9mjlN4d/SC/i00mDsZ5ijP8qY3AG)).
>

It still will work, you can see 
https://github.com/oxc-project/oxc/blob/8dd0013e7220ac9460cb99f1d6ce864da77502a4/crates/oxc_transformer/src/typescript/enum.rs#L547-L556

Here can cover `A` to  `Foo.A`. So I prefer to check symbol flags over scope id because you don't need to pass `enum_scope_id` to here, and using `SymbolFlags` will be clear than scope id.


> But you remind me that it's the symbol's scope we're checking with, not the reference's, so no need to go through the ancestors (I was confused with the [fix for babel](https://github.com/babel/babel/blob/4b2608b4fc240038a731428b249d3ab53acda735/packages/babel-plugin-transform-typescript/src/enum.ts#L178-L181)). A simple comparison with the enum's scope id should suffice.
> 
> Updated in [aec2efc](https://github.com/oxc-project/oxc/commit/aec2efc1c1eb77e1c1dfa0ffdceb6444867ae437) (I assume all symbols declared in a enum's scope are `EnumMember`. Is that correct?)

That's correct!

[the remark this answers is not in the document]
branchseer: Just to be clear about that nested enum [example](https://playground.oxc.rs/#eNo9jtFqwzAMRX9F6GUJhHUt7CXQQfrQryijbqq0AVsqsrOtBP/77ATXL7pXPlfSjD22SDw5OIrAfGKADvawbbI6JFVVNey/1h+AhTwYLX6ldwudX050DRS72QB8g7/LZK/8FkCJjSMI99FDEDinne/deaVjLrGu6hMniQ0KtjPqxLn4Jwfzh23QiRq0I4eifS8Pepmnu4gtLqhhP4i6tREbfBj1pGlg0nlI0S9yaWAweqO0AcnvPrafmIhernSjfEsybuRxGEu4Fw4q9mjlN4d/SC/i00mDsZ5ijP8qY3AG), `A` at line 6 should refer to `Bar.A`. If I only check its `EnumMember` flag while transforming enum `Foo`, I would incorrectly resolve it to `Foo.A` because I can't tell which enum it belongs to without the enum's scope id. The `match_member_expression` arm can't solve this, can it?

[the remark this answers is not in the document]
Dunqing: Oh, I got it, you are right!

Dunqing (approved):
Thanks for working on this!

PR #8552, opened 2025-01-16
feat(tasks/e2e): transformer + minifier runtime tests for popular npm packages

Boshen:
Code extracted from https://github.com/privatenumber/minification-benchmarks


Boshen:
This shouldn't be for minifier only ... 

- [ ] change to testing transformer + minifier + combinatorial of options.

PR #8562, opened 2025-01-17
feat(mangler): reuse variable names

sapphi-red:
Changed the mangler to reuse variable names where possible.

This will reduce the code size as shorter variable names can be used in more places. But requires global information and limits parallelism in a single file and requires more memory.

### DIFF
--- a/crates/oxc_mangler/src/lib.rs
+++ b/crates/oxc_mangler/src/lib.rs
@@ -1,5 +1,7 @@
+use std::iter;
 use std::ops::Deref;
 
+use fixedbitset::FixedBitSet;
 use itertools::Itertools;
 use rustc_hash::FxHashSet;
 
@@ -88,11 +90,15 @@ impl Mangler {
 
     #[must_use]
     pub fn build(self, program: &Program<'_>) -> Mangler {
-        let semantic = SemanticBuilder::new().build(program).semantic;
+        let semantic =
+            SemanticBuilder::new().with_scope_tree_child_ids(true).build(program).semantic;
         let (symbol_table, scope_tree) = semantic.into_symbol_table_and_scope_tree();
         self.build_with_symbols_and_scopes(symbol_table, &scope_tree, program)
     }
 
+    /// # Panics
+    ///
+    /// Panics if the child_ids does not exist in scope_tree.
     #[must_use]
     pub fn build_with_symbols_and_scopes(
         self,
@@ -117,6 +123,8 @@ impl Mangler {
         program: &Program<'_>,
         generate_name: G,
     ) -> Mangler {
+        assert!(scope_tree.has_child_ids(), "child_id needs to be generated");
+
         let (exported_names, exported_symbols) = if self.options.top_level {
             Mangler::collect_exported_symbols(program)
         } else {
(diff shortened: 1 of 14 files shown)


sapphi-red:
* **#8562** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8562).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red:
The perf regression looks really bad, but if you take a step back and see the overall minifier timings, it's not that big regression.

antd: 196.8ms (6.4ms is 3.3%)
react: 1.8ms (125.1µs is 7.0%)
typescript: 344.6ms (24.9ms is 7.2%)

overlookmotel:
Just to raise one other concern:

The addition of `scope_id` field to `Reference` will cause difficulties in transformer. In transformer, we update semantic data as we mutate the AST, to keep AST + semantic in sync.

This is quite difficult already, and if `Reference` has a `scope_id` field, we'd need to update that field too whenever we move an `IdentifierReference` into a different scope, or e.g. wrap some code in an IIFE (which introduces a new scope). Both of these are fairly common operations in the transformer.

The changes to `TraverseScoping::create_bound_reference` and `create_unbound_reference` in this PR aren't correct - often we are creating a `Reference` in a *different* scope from the current one.

FYI: Currently we run semantic analysis multiple times during the pipeline. The hope (well certainly *my* hope) is that we can eventually make all stages of the pipeline keep semantic data in sync with AST changes. Then we'll be able to remove these extra semantic passes, which will be a large perf improvement.

---

The performance issue needs to be looked at first. But if we decide that it's worth it (very possibly it is), I would suggest the following:

* Split the introduction of `Reference::scope_id` field into a separate PR.
* Add checks for correctness of that field to the [transformer conformance checker](https://github.com/oxc-project/oxc/blob/007e8c017c25ffc81c92ae141e02f06743c7dd14/tasks/transform_checker/src/lib.rs) (which checks semantic data is correct after transform), and see how many errors we have due to incorrect `Reference::scope_id`.
* Assess how much work required to handle updating this extra state in transformer.

It will of course be achievable with some effort, but it'd be useful to understand *how much* effort before we go ahead.

cc @Dunqing

---

Last thing: I'm not very familiar with the mangler. Could you possibly give a brief explanation of how this algorithm works? I'm wondering if any other way to achieve the effect without the `Reference::scope_id` field.

Dunqing:
This is a significant minification improvement. I've thought about this before, but I have the same concern as @overlookmotel. 

I am guessing the smaller output is more important than performance here? So I think we can add an extra AST pass to collect identifier references that belong to which scope, this way we can only change in the mangler crate.

crates/oxc_syntax/src/reference.rs
```diff
@@ -234,6 +234,8 @@ impl<'alloc> CloneIn<'alloc> for ReferenceFlags {
 pub struct Reference {
     /// The AST node making the reference.
     node_id: NodeId,
+    /// The scope of the AST node making the reference.
+    scope_id: ScopeId,
```
Boshen, line 238: This is not desirable, I'll study your algorithm later.

sapphi-red:
> The addition of `scope_id` field to `Reference` will cause difficulties in transformer. In transformer, we update semantic data as we mutate the AST, to keep AST + semantic in sync.

Yeah, I had that concern while writing this code. I went this way at that time to see how much this change would improve the minification rate.
@Dunqing 's idea of collecting the scope_ids of the references before running the mangler sounds worth exploring.

> Could you possibly give a brief explanation of how this algorithm works?

This PR introduces a concept of "liveness". This is the set of scopes that a given variable / slot is used in.
For example, if we have the following code:
```js
var top_level_a = 0;
var top_level_b = 1;

function foo() {
  var foo_a = 1;
  console.log(top_level_b, foo_a);
}

function bar() {
  var bar_a = 1;
  console.log(top_level_b, bar_a);
}

console.log(top_level_a, foo(), bar())
```
Each liveness of the variables will be:
- `top_level_a`: {root}
- `top_level_b`: {root, foo function, bar function}
- `foo_a`: {foo function}
- `bar_a`: {bar function}
- `foo`: {root}
- `bar`: {root}

Since `top_level_a` is only used in the root scope and `foo_a` is only used in the `foo` function and `bar_a` is only used in the `bar` function, `foo_a` and `bar_a` can shadow `top_level_a`.
In other words, since the liveness of `top_level_a` and `foo_a` and `bar_a` does not overlap, they can be assigned to the same slot.

For this example, the slots assigned to the variables will be:
- slot 0: `top_level_a`
- slot 1: `top_level_b`, `foo_a`, `bar_a`
- slot 2: `foo`
- slot 3: `bar`

Boshen:
@sapphi-red To be honest I spent the last hour and still don't comprehend any of this ... probably due to my lack knowledge of JavaScript.

May I ask you a favor. Can you update the top level documentation, reference all reading materials, and write a line by line explanation for your newly added code? Just like how I did it with the next one ... so the next person can understand what's going in.

crates/oxc_minifier/tests/mangler/snapshots/mangler.snap
Boshen, line 1: I think we need more tests to cover the new implementation

crates/oxc_mangler/src/lib.rs
```diff
@@ -125,59 +128,80 @@ impl Mangler {
...
+            tmp_bindings.clear();
+            tmp_bindings.extend(bindings.values().copied());
+            tmp_bindings.sort_unstable();
+            for (symbol_id, assigned_slot) in tmp_bindings.iter().zip(assignable_slots) {
+                slots[symbol_id.index()] = assigned_slot;
+
+                let lived_scope_ids = symbol_table
+                    .get_resolved_references(*symbol_id)
+                    .flat_map(|reference| {
+                        let scope_id = ast_nodes.get_node(reference.node_id()).scope_id();
```
Boshen, line 184: I fixed the previous attempt by going through the nodes table ... @overlookmotel please don't hate me :-)

crates/oxc_mangler/src/lib.rs
```diff
@@ -125,59 +128,80 @@ impl Mangler {
...
         for _ in 0..symbol_table.len() {
             slots.push(0);
         }
 
-        // Keep track of the maximum slot number for each scope
-        let mut max_slot_for_scope = Vec::with_capacity_in(scope_tree.len(), &allocator);
-        for _ in 0..scope_tree.len() {
-            max_slot_for_scope.push(0);
-        }
+        let mut slot_liveness: std::vec::Vec<FixedBitSet> = vec![];
```
Boshen, line 137: I don't even understand what `FixedBitSet` is and used for 😅

Boshen:
> ## [CodSpeed Performance Report](https://codspeed.io/oxc-project/oxc/branches/01-17-feat_mangler_reuse_variable_names)
> ### Merging #8562 will **degrade performances by 75.24%**
> Comparing `01-17-feat_mangler_reuse_variable_names` ([faaa1ee](https://github.com/oxc-project/oxc/commit/faaa1ee1603d9053cb39064b3147fb8e6af65069)) with `main` ([b977678](https://github.com/oxc-project/oxc/commit/b97767874f744f9264454a104532ad8425b8ef96))
> 
> ### Summary
> `❌ 3` regressions `✅ 29` untouched benchmarks
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/oxc-project/oxc/branches/01-17-feat_mangler_reuse_variable_names)._
> 
> ### Benchmarks breakdown
> Benchmark	`BASE`	`HEAD`	Change
> ❌	`mangler[antd.js]`	6 ms	14.6 ms	-58.67%
> ❌	`mangler[react.development.js]`	139.2 µs	295.6 µs	-52.91%
> ❌	`mangler[typescript.js]`	9.6 ms	38.7 ms	-75.24%

```
oxc  main ❯ hyperfine './target/release/examples/mangler --nospace --mangle ./target/typescript.js' './target/release/examples/minifier --nospace --mangle ./target/typescript.js'
Benchmark 1: ./target/release/examples/mangler --nospace --mangle ./target/typescript.js
  Time (mean ± σ):     142.2 ms ±   1.1 ms    [User: 129.1 ms, System: 11.4 ms]
  Range (min … max):   140.9 ms … 145.8 ms    20 runs

Benchmark 2: ./target/release/examples/minifier --nospace --mangle ./target/typescript.js
  Time (mean ± σ):     137.5 ms ±   0.5 ms    [User: 125.2 ms, System: 10.6 ms]
  Range (min … max):   136.4 ms … 138.3 ms    21 runs

Summary
  ./target/release/examples/minifier --nospace --mangle ./target/typescript.js ran
    1.03 ± 0.01 times faster than ./target/release/examples/mangler --nospace --mangle ./target/typescript.js
```

crates/oxc_mangler/src/lib.rs
```diff
@@ -129,51 +137,73 @@ impl Mangler {
...
         // Walk the scope tree and compute the slot number for each scope
         let mut tmp_bindings = std::vec::Vec::with_capacity(100);
-        for scope_id in scope_tree.descendants_from_root() {
+        for scope_id in iter::once(scope_tree.root_scope_id())
+            .chain(scope_tree.iter_all_child_ids(scope_tree.root_scope_id()))
+        {
+            let nearest_var_scope_id = scope_tree
+                .ancestors(scope_id)
+                .find(|s_id| scope_tree.get_flags(*s_id).is_var())
+                .unwrap_or(scope_tree.root_scope_id());
```
sapphi-red: I had this code and "treat all symbols as var for now" comment before. But it turns out this was already working so I removed these.

sapphi-red:
I added a comment that describes the algorithm and added some test cases. I hope that my explanation makes sense, but let me know if there's any places that are unclear.

PR #8588, opened 2025-01-18
refactor(linter): improve `eslint/no-lone-blocks`

shulaoda:
The code `let mut lone_blocks = Vec::new();` is unnecessary because it will contain at most one element throughout its usage.

Additionally, special test cases has been added for when `parent_node.kind()` is `AstKind::FunctionBody(_)`.


camc314 (approved):
thanks!

PR #8590, opened 2025-01-18
fix(linter): `ignorePatterns` does not work when files are provided as command arguments

### LINKED ISSUE #8505: linter: ignorePatterns does not work when files are provided as command arguments

0.15.6

### What command did you run?

oxlint  -c .oxlintrc.json --quiet --fix "./tools/config/src/app.config.d.ts"

```
{
  "ignorePatterns": ["**/app.config.d.ts", "global.d.ts"],
...
}
```

### What happened?

It seems that the `ignorePatterns` attribut within the config is ignored when an file path is provided as argument.
So my file `./tools/config/src/app.config.d.ts` still get lint in this case. If I don't put any argument in the commande,`ignorePatterns` works fine.

Referring to this [issue](https://github.com/oxc-project/oxc/pull/7566) it seems to be a bug?

Thanks !

shulaoda:
closes #8505 

Here is why this issue occurs:
https://github.com/oxc-project/oxc/blob/c15af02e52e124f3968b098d487e28b49d937b56/apps/oxlint/src/lint.rs#L66-L70


apps/oxlint/src/lint.rs
```diff
@@ -63,18 +63,65 @@ impl Runner for LintRunner {
...
+        let config_search_result =
+            Self::find_oxlint_config(&self.cwd, basic_options.config.as_ref());
+
+        if let Err(err) = config_search_result {
+            return err;
+        }
+
+        let mut oxlintrc = config_search_result.unwrap();
+
+        let mut override_builder = OverrideBuilder::new(&self.cwd);
```
shulaoda, line 86: The only destructive change here is that I replaced `oxlint_wd` with `self.cwd`.

This is because `ignore_options.ignore_pattern` and `oxlintrc.ignore_patterns` conflict with each other, and we cannot resolve this conflict within a single `OverrideBuilder`.

In general, the `oxlintrc` is expected to be in the `self.cwd` directory. If the user specifies `oxlint -c special/path/oxlintrc.json`, we should apply the `ignorePatterns` from that configuration relative to the current working directory (`self.cwd`).

@Sysix cc

Sysix: When I started to fix `ignorePatterns` for windows I asked myself how to solve exactly this problem.  
I would expect that the paths inside the config file is relative to the config file directory and not the cwd.

I searched a bit in eslint repo, and it looks like eslint will append the cfd-cwd path
to each ignorePattnern inside the config file:
https://github.com/eslint/eslint/blob/17f2aaec16d5afbb0d219bce6ae01d7b15d74828/lib/config/config-loader.js#L515-L538

Sysix: @shulaoda did you have time to change this behavior?
I think patterns from the CLI (`oxlint --ignore-patterns=...`) should start from the CWD
Patterns from the Config should start from the `.oxlintrc.json` file directory.

shulaoda: Due to some recent personal matters, it was delayed. Today, I picked it up again and proceeded according to your suggestions. @Sysix

apps/oxlint/src/lint.rs
```diff
@@ -326,6 +358,26 @@ impl LintRunner {
             Err(error)
         }
     }
+
+    fn adjust_ignore_patterns(base: &str, path: &str, ignore_patterns: Vec<String>) -> Vec<String> {
+        if base == path {
+            ignore_patterns
+        } else {
+            let relative_ignore_path =
+                pathdiff::diff_paths(path, base).unwrap_or_else(|| PathBuf::from("."));
```
Sysix, line 367: side note: we can remove the dependency when we expect that the config file is always inside the cwd.
(`path.strip_prefix(base)`)
But then this would not work correctly: `oxlint -c ../.oxlintrc.json` :/

apps/oxlint/src/lint.rs
```diff
@@ -107,9 +84,64 @@ impl Runner for LintRunner {
         let mut oxlintrc = config_search_result.unwrap();
         let oxlint_wd = oxlintrc.path.parent().unwrap_or(&self.cwd).to_path_buf();
 
-        let paths = Walk::new(&oxlint_wd, &paths, &ignore_options, &oxlintrc.ignore_patterns)
-            .with_extensions(Extensions(extensions))
-            .paths();
+        let mut override_builder = OverrideBuilder::new(&self.cwd);
```
Sysix, line 87: now we are building `OverrideBuilder` even when `ignore_options.no_ignore` is active.
Can you build it as an Option?

apps/oxlint/src/lint.rs
```diff
@@ -107,9 +84,64 @@ impl Runner for LintRunner {
...
+        // To accommodate this, unless `--no-ignore` is passed,
+        // pre-filter the paths.
+        if !paths.is_empty() && !ignore_options.no_ignore {
+            let (ignore, _err) = Gitignore::new(&ignore_options.ignore_path);
+
+            paths.retain_mut(|p| {
+                // Append cwd to all paths
+                let mut path = self.cwd.join(&p);
+
+                std::mem::swap(p, &mut path);
```
Sysix, line 121: Do not enough about Rust to tell if this is safe

shulaoda: Don't worry, this is a safe operation.

Sysix (approved):

PR #8595, opened 2025-01-19
feat(npm/runtime): add `@oxc-project/runtime` package

### LINKED ISSUE #7422: `@oxc-project/runtime` NPM package
We currently use [@babel/runtime](https://babeljs.io/docs/babel-runtime) for helper functions used in transformer.

Should we publish our own `@oxc-project/runtime` NPM package and use that instead?

Initially, it could just re-export from `@babel/runtime`, but I can see various advantages of having our own package:

1. `@babel/runtime` may be updated and break our transforms at any time. We can lock the version of `@babel/runtime` used in `@oxc-project/runtime` to prevent that.
2. We have decided to diverge from Babel where we can produce better output, or need to do things differently because Rust. We may want to alter some of the helpers too.
3. We only use a fraction of the helpers because we're only supporting transform down to ES6. Our package can be much smaller than `@babel/runtime`.
4. It's just weird to tell users to install `@babel/runtime` as a dependency in their project to use Oxc.

Boshen:
closes #7422

All code are copied manually, nothing is automated yet. (cjs files and package.json#exports should be generated from the esm directory.)

The following code are not copied over yet, as we may not need them.
 
@babel/runtime/regenerator/index.js`

Package dependency:

```
  "dependencies": {
    "regenerator-runtime": "^0.14.0"
  },
```

### DIFF
--- a/.github/workflows/release_runtime.yml
+++ b/.github/workflows/release_runtime.yml
@@ -1,0 +1,59 @@
+name: Release @oxc-project/runtime
+
+permissions: {}
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - npm/runtime/package.json
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: Check version
+    runs-on: ubuntu-latest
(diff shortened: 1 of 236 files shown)


Harsh97711:
PULL THE REQUEST

PR #8596, opened 2025-01-19
refactor(span): derive `Copy` on `Atom`

wan9chi:
Follow up from https://github.com/oxc-project/oxc/pull/8543#discussion_r1918592423

> I agree. https://github.com/oxc-project/oxc/issues/8606

> Originally we were considering some form of interning and reference-counting, so we didn't make it Copy to leave the door open for that. But now all strings are stored in the arena anyway, so even if we did decide to intern strings, reference-counting would be irrelevant - our bump allocator doesn't allow freeing individual allocations anyway.

Most of the changes are done automatically by `just fix` (`cargo clippy --fix` && `cargo fmt --all`). See the commit list for the manual edits.


overlookmotel:
Thanks loads for doing this. To my shame, I didn't know `cargo clippy --fix` existed!

I've made some follow-up changes in #8641.

PR #8600, opened 2025-01-19
fix(linter): can't disable `no-nested-ternary` rule anymore

### LINKED ISSUE #8485: linter: cannot disable no-nested-ternary rule anymore

0.15.6

### What command did you run?

oxlint -c .oxlintrc.json src

{
  "env": {
    "browser": true,
    "es2024": true
  },
  "settings": {},
  "plugins": ["oxc", "typescript", "unicorn"],
  "categories": {
    "correctness": "error",
    "perf": "error",
    "suspicious": "error",
    "restriction": "error",
    "style": "warn",
    "pedantic": "warn",
    "nursery": "off"
  },
  "rules": {
    "no-nested-ternary": "off",
    "eslint/no-nested-ternary": "off",
    "unicorn/no-nested-ternary": "off"
  }
}

### What happened?

Since the new 'eslint/no-nested-ternary' rule appeared in 0.15.x (duplicating/replacing the previous unicorn/no-nested-ternary rule?), it seems it cannot be deactivated anymore. With the above setup the rule is still active and generates warning.
It was not the case in 0.14.x

shulaoda:
closes #8485 

Since we currently support two rules with the same `rule_name` but different `plugin_names`, some of the original logic is no longer applicable. As a result, I have made some adjustments.

### DIFF
--- a/crates/oxc_linter/src/config/rules.rs
+++ b/crates/oxc_linter/src/config/rules.rs
@@ -504,18 +504,4 @@ mod test {
             assert_eq!(rule.severity, AllowWarnDeny::Warn, "{config:?}");
         }
     }
-
-    #[test]
-    fn test_two_rules_with_same_name_and_different_plugin_names() {
-        // Issue: <https://github.com/oxc-project/oxc/issues/8485>
-        let configs =
-            [json!({ "eslint/no-nested-ternary": "off", "unicorn/no-nested-ternary": "off" })];
-
-        for config in &configs {
-            let mut rules = RuleSet::default();
-            r#override(&mut rules, config);
-
-            assert_eq!(rules.len(), 0, "{config:?}");
-        }
-    }
 }
--- a/crates/oxc_linter/src/config/rules.rs
+++ b/crates/oxc_linter/src/config/rules.rs
@@ -110,28 +110,43 @@ impl OxlintRules {
                     }
                 }
                 _ => {
-                    // For overlapping rule names, use the "error" one
-                    // "no-loss-of-precision": "off",
-                    // "@typescript-eslint/no-loss-of-precision": "error"
-                    if let Some(rule_config) =
-                        rule_configs.iter().find(|r| r.severity.is_warn_deny())
-                    {
-                        let config = rule_config.config.clone().unwrap_or_default();
+                    let rules = rules_for_override
+                        .iter()
+                        .filter_map(|r| {
+                            if r.name() == *name {
+                                Some((r.plugin_name(), r))
+                            } else {
+                                None
+                            }
+                        })
+                        .collect::<FxHashMap<_, _>>();
 
-                        if let Some(rule) = rules_for_override.iter().find(|r| r.name() == *name) {
-                            rules_to_replace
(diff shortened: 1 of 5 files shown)


crates/oxc_linter/src/config/rules.rs
```diff
@@ -131,8 +131,10 @@ impl OxlintRules {
                             ));
                         }
                     } else if rule_configs.iter().all(|r| r.severity.is_allow()) {
-                        if let Some(rule) = rules_for_override.iter().find(|r| r.name() == *name) {
```
shulaoda, line 134: Currently, we support multiple identical rules with different plugins, which may introduce potential issues. However, I am only addressing the related issues at the moment, without making further changes to avoid introducing other problems.

https://github.com/oxc-project/oxc/blob/c390fdf332e7e0cbb519b3cb352ccee2fc6a687a/crates/oxc_linter/src/config/rules.rs#L121-L123

crates/oxc_linter/src/config/rules.rs
```diff
@@ -465,9 +469,8 @@ mod test {
     fn test_override_plugin_prefix_duplicates() {
         let configs = [
             // FIXME: this should be valid
-            // json!({ "@typescript-eslint/no-unused-vars": "error" }),
-            json!({ "no-unused-vars": "off", "typescript/no-unused-vars": "error" }),
-            json!({ "no-unused-vars": "off", "@typescript-eslint/no-unused-vars": "error" }),
```
shulaoda, line 470: `@typescript-eslint/no-unused-vars` does not exist.

crates/oxc_linter/src/config/rules.rs
```diff
@@ -465,9 +469,8 @@ mod test {
...
+            // json!({ "@unicorn-eslint/no-nested-ternary": "error" }),
+            json!({ "no-nested-ternary": "off", "unicorn/no-nested-ternary": "error" }),
```
shulaoda, line 473: `"no-nested-ternary"` => `"eslint/no-nested-ternary"`

Currently, we do not support rules without an explicit plugin_name, which would automatically match all rules:
`"no-nested-ternary"` => `"eslint/no-nested-ternary"` and  `"unicorn/no-nested-ternary"`

crates/oxc_linter/src/config/rules.rs
Sysix: here is the rule: https://typescript-eslint.io/rules/no-unused-vars/

Internally we are remapping some eslint rules to typescript here:
https://github.com/oxc-project/oxc/blob/d966e0addafc4e0160ea27bf2d65e8b97ef299cb/crates/oxc_linter/src/utils/mod.rs#L55-L74

shulaoda: > Internally we are remapping some eslint rules to typescript here:

Got it.

Sysix (commented):
Thanks you for your work 🫶 
Could you please create a CLI snapshot test for it too? 
Then we can be sure that the implementation from start to end is correct.

Sysix (approved):

PR #8607, opened 2025-01-19
feat(oxlint): add stylish formatter

### LINKED ISSUE #8422: lint: stylish reporter to match ESLint stylish formatter
I've been experimenting with moving a rather large monorepo over to oxlint and I must say, very impressive thus far. I'd like to request a reporter to match the [ESLint stylish reporter](https://eslint.org/docs/latest/use/formatters/#stylish).

shellscape:
👋  This implements a reporter for `--format` on `oxlint` which aims to be visually similar to https://eslint.org/docs/latest/use/formatters/#stylish

Please note that this is my first time working with Rust and my knowledge is very limited. I'm unlikely to understand best-practice or best-pattern references outside of what clippy/cargo lint has already had me change. If this needs modification, please help me out by making code suggestions that can be merged to this PR. 

Resolves #8422

### DIFF
--- a/apps/oxlint/src/output_formatter/mod.rs
+++ b/apps/oxlint/src/output_formatter/mod.rs
@@ -20,6 +20,8 @@ use crate::output_formatter::{default::DefaultOutputFormatter, json::JsonOutputF
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum OutputFormat {
     Default,
+    /// GitHub Check Annotation
+    /// <https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message>
     Github,
     Json,
     Unix,
--- a/apps/oxlint/src/output_formatter/mod.rs
+++ b/apps/oxlint/src/output_formatter/mod.rs
@@ -2,6 +2,7 @@ mod checkstyle;
 mod default;
 mod github;
 mod json;
+mod stylish;
 mod unix;
 
 use std::io::{BufWriter, Stdout, Write};
@@ -9,6 +10,7 @@ use std::str::FromStr;
 
 use checkstyle::CheckStyleOutputFormatter;
 use github::GithubOutputFormatter;
+use stylish::StylishOutputFormatter;
 use unix::UnixOutputFormatter;
 
 use oxc_diagnostics::reporter::DiagnosticReporter;
@@ -18,12 +20,11 @@ use crate::output_formatter::{default::DefaultOutputFormatter, json::JsonOutputF
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum OutputFormat {
     Default,
-    /// GitHub Check Annotation
-    /// <https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message>
     Github,
     Json,
     Unix,
     Checkstyle,
+    Stylish,
 }
 
 impl FromStr for OutputFormat {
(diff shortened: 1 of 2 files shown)


apps/oxlint/src/output_formatter/mod.rs
```diff
@@ -18,12 +20,11 @@ use crate::output_formatter::{default::DefaultOutputFormatter, json::JsonOutputF
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum OutputFormat {
     Default,
-    /// GitHub Check Annotation
```
Boshen, line 21: These comments shouldn't be removed.

shellscape: fixed

apps/oxlint/src/output_formatter/stylish.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn get_diagnostic_reporter(&self) -> Box<dyn DiagnosticReporter> {
+        Box::new(StylishReporter::default())
+    }
+}
+
+#[derive(Default)]
+struct StylishReporter {
+    diagnostics: Vec<Error>,
```
Sysix, line 26: can this be `FxHashMap<String, Vec<&Error>>` and the grouping will be done inside `render_error` :)?

apps/oxlint/src/output_formatter/stylish.rs
```diff
@@ -0,0 +1,146 @@
...
+        let info = Info::new(diagnostic);
+        grouped.entry(info.filename).or_default().push(diagnostic);
+    }
+
+    for diagnostics in grouped.values() {
+        let diagnostic = diagnostics[0];
+        let info = Info::new(diagnostic);
+        let filename = info.filename;
+        let filename = if let Some(path) =
+            std::env::current_dir().ok().and_then(|d| d.join(&filename).canonicalize().ok())
```
Sysix, line 64: `canonicalize` adds under widows a  `\\?\` prefix to it.
Is there a problem using `info.filename` without it? 
Would like to avoid `current_dir` and `canonicalize` :)

shellscape: @Sysix a note from my PR description:

> Please note that this is my first time working with Rust and my knowledge is very limited. I'm unlikely to understand best-practice or best-pattern references outside of what clippy/cargo lint has already had me change. If this needs modification, please help me out by making code suggestions that can be merged to this PR.

I'm also not a Windows user, so I'd have no valuable insight into that question. Please do make a code suggestion

Sysix: I would like to see this code block removed:

```rust
        let filename = if let Some(path) =
            std::env::current_dir().ok().and_then(|d| d.join(&filename).canonicalize().ok())
        {
            path.display().to_string()
        } else {
            filename
        };
```

`let filename = info.filename;` should be just fine

shellscape: > let filename = info.filename; should be just fine.

When I was testing the output using `just watch` the filename was relative to the current working directory, and would print just `test.js` using `let filename = info.filename;`. ESLint v9 always prints the full file path, and so I was attempting to get to 1:1 with that. Personally, I'd very much prefer the full file path. Maybe I did something else incorrectly in conjunction, but the solution as it is was the only way I could get the reporter to get to the full path. 

> PS: thanks you for the work 🫶

very welcome. this is an impressive project, happy to try and make a modest contribution.

Sysix: Got it! No work on your side. Great job 🥳 
You can resolve the discussion, I do not have the rights for this :)

Sysix (approved):

shellscape:
Thanks for the quick merge!

PR #8608, opened 2025-01-20
feat(oxc_allocator):  Add new method `clone_in_with_semantic_ids` for `CloneIn` trait

IWANABETHATGUY:
1. Unlike `clone_in` the new method also clone `semantic` fields, this is used for scenario like caching ast and semantic data at the same time.

### DIFF
--- a/crates/oxc_allocator/src/clone_in.rs
+++ b/crates/oxc_allocator/src/clone_in.rs
@@ -31,6 +31,13 @@ pub trait CloneIn<'new_alloc>: Sized {
     /// Clone `self` into the given `allocator`. `allocator` may be the same one
     /// that `self` is already in.
     fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned;
+
+    /// Almost same as `clone_in`, but for some special type, it will also clone the semantic ids.
+    /// Please use this method only if you make sure the semantic info is synced with the ast.
+    #[inline]
+    fn clone_in_with_semantic_ids(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
+        self.clone_in(allocator)
+    }
 }
 
 impl<'alloc, T, C> CloneIn<'alloc> for Option<T>
@@ -41,6 +48,10 @@ where
 
     fn clone_in(&self, allocator: &'alloc Allocator) -> Self::Cloned {
         self.as_ref().map(|it| it.clone_in(allocator))
+    }
+
+    fn clone_in_with_semantic_ids(&self, allocator: &'alloc Allocator) -> Self::Cloned {
+        self.as_ref().map(|it| it.clone_in_with_semantic_ids(allocator))
     }
 }
 
@@ -53,6 +64,10 @@ where
     fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
         Box::new_in(self.as_ref().clone_in(allocator), allocator)
     }
+
+    fn clone_in_with_semantic_ids(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
+        Box::new_in(self.as_ref().clone_in_with_semantic_ids(allocator), allocator)
+    }
 }
 
 impl<'new_alloc, T, C> CloneIn<'new_alloc> for Vec<'_, T>
@@ -63,6 +78,10 @@ where
 
     fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
         Vec::from_iter_in(self.iter().map(|it| it.clone_in(allocator)), allocator)
+    }
+
+    fn clone_in_with_semantic_ids(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
+        Vec::from_iter_in(self.iter().map(|it| it.clone_in_with_semantic_ids(allocator)), allocator)
     }
 }
 
--- a/crates/oxc_syntax/src/reference.rs
+++ b/crates/oxc_syntax/src/reference.rs
@@ -1,7 +1,7 @@
 #![allow(missing_docs)] // fixme
 use bitflags::bitflags;
 use nonmax::NonMaxU32;
-use oxc_allocator::CloneIn;
+use oxc_allocator::{Allocator, CloneIn};
 use oxc_index::Idx;
 #[cfg(feature = "serialize")]
 use serde::{Serialize, Serializer};
@@ -11,6 +11,18 @@ use crate::{node::NodeId, symbol::SymbolId};
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct ReferenceId(NonMaxU32);
 
+impl<'alloc> CloneIn<'alloc> for ReferenceId {
+    type Cloned = Self;
+
+    fn clone_in(&self, _: &'alloc Allocator) -> Self {
+        // `clone_in` should never reach this, because `CloneIn` skips semantic ID fields
+        unreachable!();
+    }
+
+    fn clone_in_with_semantic_ids(&self, _: &'alloc Allocator) -> Self {
+        *self
+    }
+}
 impl Idx for ReferenceId {
     #[allow(clippy::cast_possible_truncation)]
     fn from_usize(idx: usize) -> Self {
--- a/tasks/ast_tools/src/derives/clone_in.rs
+++ b/tasks/ast_tools/src/derives/clone_in.rs
@@ -54,49 +54,118 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
         .collect_vec();
 
     let alloc_ident = if used_alloc { format_ident!("allocator") } else { format_ident!("_") };
-    let body = quote! {
+    let clone_in_body = quote! {
         match self {
             #(#matches),*
         }
     };
 
-    impl_clone_in(&ty_ident, def.has_lifetime, &alloc_ident, &body)
+    let clone_in_with_semantic_ids_token_stream = if used_alloc {
+        let matches = def
+        .all_variants()
+        .map(|var| {
+            let ident = var.ident();
+            if var.is_unit() {
+                quote!(Self :: #ident => #ty_ident :: #ident)
+            } else {
+                quote!(Self :: #ident(it) => #ty_ident :: #ident(CloneIn::clone_in_with_semantic_ids(it, allocator)))
+            }
+        })
+        .collect_vec();
+
+        let alloc_ident_param = if def.has_lifetime {
+            quote!(#alloc_ident: &'new_alloc Allocator)
+        } else {
+            quote!(#alloc_ident: &'alloc Allocator)
+        };
+        quote!(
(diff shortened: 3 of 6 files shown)


crates/oxc_ast/src/ast/js.rs
```diff
@@ -41,7 +41,6 @@ pub struct Program<'a> {
     pub directives: Vec<'a, Directive<'a>>,
     pub body: Vec<'a, Statement<'a>>,
     #[estree(skip)]
-    #[clone_in(default)]
```
IWANABETHATGUY, line 44: This is not right, should use `cfg(attribete)`

Boshen:
Can you provide some context and link your Rolldown prototype for other reviewers?

IWANABETHATGUY:
> Can you provide some context and link your Rolldown prototype for other reviewers?

OK, I will link it once the new poc is finished.

overlookmotel:
I'm not quite sure what you're trying to achieve here, but this may be relevant: https://github.com/oxc-project/backlog/issues/127

`CloneIn` is a bit of a nightmare. I tend to think we should not have introduced it in the first place, because it's a footgun.

IWANABETHATGUY:
> I'm not quite sure what you're trying to achieve here, but this may be relevant: [oxc-project/backlog#127](https://github.com/oxc-project/backlog/issues/127)
> 
> `CloneIn` is a bit of a nightmare. I tend to think we should not have introduced it in the first place, because it's a footgun.
> 
> At present, it unsets all `scope_id` etc fields to `None`, which is not good because a post-semantic AST should always have these fields set with an ID.
> 
> But if we go the other way (as in your PR), then it _will_ have ID fields set, but they're likely to be the _wrong_ IDs.
> 
> So either way is wrong! `CloneIn` cannot set the IDs correctly without context about _where_ the cloned node is going to be inserted.
> 
> See also: [oxc-project/backlog#96](https://github.com/oxc-project/backlog/issues/96)

incremental build, since we need to modify the ast, we should clone `Ast` to avoid parse it next round, 
I assume the source code, should have same AST, and same semantic data. In this scenario, the semantic data(like ScopeId, SymbolId)  could be cloned safely

overlookmotel:
Ah I see.

If you clone both the AST and also the matching `ScopeTree` and `SymbolTable`, then yes the IDs will all remain valid. Ditto if you don't clone `ScopeTree` or `SymbolTable`, but you don't modify them in any way either.

But I suggest we introduce a new trait for this use case `CloneIntoWithIds`, and document that it's only valid to use in these specific circumstances. Other code currently relies on the current implementation of `CloneIn`, so I don't think we should change it.

`CloneIn` is codegen-ed, so we can easily codegen `CloneIntoWithIds` too.

Note on naming: I suggest `CloneIntoWithIds` not `CloneInWithIds` to indicate that you're cloning *into* a different allocator, instead of *in* the same allocator (see https://github.com/oxc-project/backlog/issues/96).

We need `Allocator::with_capacity` method for other purposes, so I've broken that out into a separate PR #8620.

I assume you'll also need a way to get the capacity required for the allocator you're cloning into, so I've also added that in #8621. Please take a look and confirm that covers what you need.

overlookmotel:
Actually maybe we don't need a new trait. We could just add a 2nd method to `CloneIn`: `clone_in_with_semantic_ids`.

...and document the pitfalls of using this method!

IWANABETHATGUY:
> Actually maybe we don't need a new trait. We could just add a 2nd method to `CloneIn`: `clone_in_with_semantic_ids`.
> 
> ...and document the pitfalls of using this method!
> 
> @IWANABETHATGUY Are you comfortable enough with `oxc_ast_tools` to make that change yourself?

Thanks for your guiding, I will try it tomorrow. I am working on other thing today.

IWANABETHATGUY:
* **#8608** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8608?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8608?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8608).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

tasks/ast_tools/src/derives/clone_in.rs
```diff
@@ -60,52 +62,92 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
...
+                    }
+                    CloneInAttribute::None => {
+                        quote!(#ident: CloneIn::clone_in(&self.#ident, allocator))
+                    }
                 }
-            }
-        });
-        (format_ident!("allocator"), quote!(#ty_ident { #(#fields),* }))
+            })
+            .collect::<Vec<_>>()
```
IWANABETHATGUY, line 91: The `collect` to make sure the fn in `iter` is executed and updates the `needs_override_clone_in_with_semantic_ids` variable

crates/oxc_syntax/src/reference.rs
```diff
@@ -11,6 +11,14 @@ use crate::{node::NodeId, symbol::SymbolId};
...
 pub struct ReferenceId(NonMaxU32);
 
+impl<'alloc> CloneIn<'alloc> for ReferenceId {
+    type Cloned = Self;
+
+    fn clone_in(&self, _: &'alloc oxc_allocator::Allocator) -> Self::Cloned {
+        *self
+    }
+}
+
```
overlookmotel, line 21:
```suggestion
impl<'alloc> CloneIn<'alloc> for ReferenceId {
    type Cloned = Self;

    fn clone_in(&self, _: &'alloc Allocator) -> Self {
        // `clone_in` should never reach this, because `CloneIn` skips semantic ID fields
        unreachable!();
    }

    fn clone_in_with_semantic_ids(&self, _: &'alloc Allocator) -> Self {
        *self
    }
}
```

(Please add `use oxc_allocator::Allocator;` with other imports. Our code style generally avoids local imports)

crates/oxc_allocator/src/clone_in.rs
```diff
@@ -31,6 +31,12 @@ pub trait CloneIn<'new_alloc>: Sized {
     /// Clone `self` into the given `allocator`. `allocator` may be the same one
     /// that `self` is already in.
     fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned;
+
+    /// Almost same as `clone_in`, but for some special type, it will also clone the semantic ids.
+    #[inline]
+    fn clone_in_with_semantic_ids(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
+        self.clone_in(allocator)
+    }
```
overlookmotel, line 39: Please expand this comment to explain why it's generally a bad idea to use this method.

tasks/ast_tools/src/derives/clone_in.rs
```diff
@@ -60,52 +62,92 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
...
-    impl_clone_in(&ty_ident, def.has_lifetime, &alloc_ident, &body)
+    impl_clone_in(&ty_ident, def.has_lifetime, &alloc_ident, &body, &quote!())
 }
 
 fn derive_struct(def: &StructDef) -> TokenStream {
     let ty_ident = def.ident();
-
-    let (alloc_ident, body) = if def.fields.is_empty() {
-        (format_ident!("_"), quote!(#ty_ident))
+    let needs_override_clone_in_with_semantic_ids = Cell::new(false);
```
overlookmotel, line 70: Why do you need `Cell` here? Just a mutable var should do.

tasks/ast_tools/src/derives/clone_in.rs
```diff
@@ -60,52 +62,92 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
...
+            .into_iter();
+        let clone_in_with_semantic_ids_token_stream = if needs_override_clone_in_with_semantic_ids
+            .get()
+        {
+            let fields = def.fields.iter().map(|field| {
+                let ident = field.ident();
+                quote!(#ident: CloneIn::clone_in(&self.#ident, allocator))
+            });
+            quote!(
+                fn clone_in_with_semantic_ids(&self, #alloc_ident: &'new_alloc Allocator) -> Self::Cloned {
```
overlookmotel, line 101: Add `///@@line_break` before this line to make the generated code nicer.

tasks/ast_tools/src/derives/clone_in.rs
```diff
@@ -60,52 +62,92 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
...
-        (format_ident!("allocator"), quote!(#ty_ident { #(#fields),* }))
+            })
+            .collect::<Vec<_>>()
+            .into_iter();
+        let clone_in_with_semantic_ids_token_stream = if needs_override_clone_in_with_semantic_ids
+            .get()
+        {
+            let fields = def.fields.iter().map(|field| {
+                let ident = field.ident();
+                quote!(#ident: CloneIn::clone_in(&self.#ident, allocator))
```
overlookmotel, line 98: This isn't correct. It should be calling `clone_in_with_semantic_ids`, not `clone_in`.

tasks/ast_tools/src/derives/clone_in.rs
```diff
@@ -60,52 +62,92 @@ fn derive_enum(def: &EnumDef) -> TokenStream {
         }
     };
 
-    impl_clone_in(&ty_ident, def.has_lifetime, &alloc_ident, &body)
+    impl_clone_in(&ty_ident, def.has_lifetime, &alloc_ident, &body, &quote!())
```
overlookmotel, line 65: You will need a `clone_in_with_semantic_ids` impl for most enums too. If you call `clone_in_with_semantic_ids` on a `&Program` you want it to be using `clone_in_with_semantic_ids` all the way through the AST, not diverting to `clone_in` as soon as it hits a `Statement`.

Otherwise `Program::clone_in_with_semantic_ids` will do exactly the same as `Program::clone_in` - cloned AST won't have any IDs.

overlookmotel (changes requested):

tasks/ast_tools/src/derives/clone_in.rs
overlookmotel: Makes sense!

tasks/ast_tools/src/derives/clone_in.rs
IWANABETHATGUY: ah, you are right, I use variable directly before, it violates borrow rules, it worked after I use `collect::<Vec<_>>`

Boshen:
To make sure this PR work as intended, let's review and merge once there's a corresponding working PR in Rolldown.

IWANABETHATGUY:
> To make sure this PR work as intended, let's review and merge once there's a corresponding working PR in Rolldown.

that make sense

Issue #4047, opened 2024-07-04
transformer: legacy decorators


Boshen:
# legacy (typescript `--experimentalDecorators`)

esbuild: `--loader=ts --tsconfig-raw='{"compilerOptions":{"experimentalDecorators":true}}'`
tsc: https://www.typescriptlang.org/docs/handbook/decorators.html

Boshen:
Will reopen when planned.

Boshen:
> experimental decorators and useDefineForClassFields: false is not support by oxc

From rolldown-vite integration.

OskarLebuda:
@Boshen do we now any EOT?

yyx990803:
For reference: https://github.com/evanw/decorator-tests

OskarLebuda:
@Boshen  @Dunqing do you have guys any kind of ETA?

Boshen:
> @Boshen @Dunqing do you have guys any kind of ETA?

What are you using oxc for that requires decorators, and which version of decorators do you need?

OskarLebuda:
> What are you using oxc for that requires decorators, and which version of decorators do you need?

I need this to build project using rolldown. See https://github.com/rolldown/rolldown/issues/2296

This project uses e.g. decorators for dependencie-injection. 
```typescript
export class StorageService {
  
  @Inject($MyStorage)
  private myStorage: MyStorage;

  public get(key: string): string {
    return myStorage.get(key);
  }
  
}
```

This are [TypeScripts decorators](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#decorators)

OskarLebuda:
@Boshen do you have any updates for this issue? 😞

Boshen:
The team has agreed to implement typescript `--experimentalDecorators` first, which is simpler and what most people use today. Research this week, implementation will probably start next weerk.

Boshen:
Legacy decorator has been implemented and is being integrated into Rolldown. 

A `decorator?: DecoratorOptions` option has been added to the `oxc-transform` package.

Stage 3 ecma decorators is tracked in

* https://github.com/oxc-project/oxc/issues/9170

alfaproject:
[...]

PR #8614, opened 2025-01-20
feat(transformer): support for transforming legacy decorator

Dunqing:
related: #4047
related: https://github.com/rolldown/rolldown/issues/2296

This is also known as "Experimental Decorator" in `TypeScript` by [experimentalDecorators](https://www.typescriptlang.org/tsconfig/#experimentalDecorators) enabling.

### Testing

- Six tests fail due to [emitDecoratorMetadata](https://www.typescriptlang.org/tsconfig/#emitDecoratorMetadata), as we haven't supported it yet. I found `esbuild` doesn't support it as well.
- A few tests fail due to different unique binding generator
- ...

### DIFF
--- a/crates/oxc_transformer/src/common/helper_loader.rs
+++ b/crates/oxc_transformer/src/common/helper_loader.rs
@@ -164,6 +164,8 @@ pub enum Helper {
     ReadOnlyError,
     WriteOnlyError,
     CheckInRHS,
+    Decorate,
+    DecorateParam,
 }
 
 impl Helper {
@@ -193,6 +195,8 @@ impl Helper {
             Self::ReadOnlyError => "readOnlyError",
             Self::WriteOnlyError => "writeOnlyError",
             Self::CheckInRHS => "checkInRHS",
+            Self::Decorate => "decorate",
+            Self::DecorateParam => "decorateParam",
         }
     }
 }
--- a/tasks/transform_conformance/snapshots/oxc.snap.md
+++ b/tasks/transform_conformance/snapshots/oxc.snap.md
@@ -321,10 +321,20 @@ x Output mismatch
 
 # legacy-decorators (0/62)
 * typescript/accessor/decoratorOnClassAccessor1/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Unresolved references mismatch:
+after transform: ["TypedPropertyDescriptor", "babelHelpers", "dec"]
+rebuilt        : ["babelHelpers", "dec"]
 
 * typescript/accessor/decoratorOnClassAccessor2/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Unresolved references mismatch:
+after transform: ["TypedPropertyDescriptor", "babelHelpers", "dec"]
+rebuilt        : ["babelHelpers", "dec"]
 
 * typescript/accessor/decoratorOnClassAccessor3/input.ts
 
@@ -340,10 +350,20 @@ x Output mismatch
 
 
 * typescript/accessor/decoratorOnClassAccessor4/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Unresolved references mismatch:
+after transform: ["TypedPropertyDescriptor", "babelHelpers", "dec"]
+rebuilt        : ["babelHelpers", "dec"]
 
 * typescript/accessor/decoratorOnClassAccessor5/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Unresolved references mismatch:
+after transform: ["TypedPropertyDescriptor", "babelHelpers", "dec"]
+rebuilt        : ["babelHelpers", "dec"]
 
 * typescript/accessor/decoratorOnClassAccessor6/input.ts
 
@@ -365,7 +385,12 @@ x Output mismatch
 x Output mismatch
 
 * typescript/constructableDecoratorOnClass01/input.ts
-x Output mismatch
+Symbol span mismatch for "C":
+after transform: SymbolId(1): Span { start: 74, end: 75 }
+rebuilt        : SymbolId(1): Span { start: 0, end: 0 }
+Symbol span mismatch for "C":
+after transform: SymbolId(2): Span { start: 0, end: 0 }
+rebuilt        : SymbolId(2): Span { start: 74, end: 75 }
 
 * typescript/constructor/decoratorOnClassConstructor1/input.ts
 x Output mismatch
@@ -374,7 +399,18 @@ x Output mismatch
 x Output mismatch
 
 * typescript/constructor/parameter/decoratorOnClassConstructorParameter1/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Symbol span mismatch for "C":
+after transform: SymbolId(3): Span { start: 155, end: 156 }
+rebuilt        : SymbolId(0): Span { start: 0, end: 0 }
+Symbol span mismatch for "C":
+after transform: SymbolId(5): Span { start: 0, end: 0 }
+rebuilt        : SymbolId(1): Span { start: 155, end: 156 }
+Unresolved references mismatch:
+after transform: ["Function", "babelHelpers", "dec"]
+rebuilt        : ["babelHelpers", "dec"]
 
 * typescript/constructor/parameter/decoratorOnClassConstructorParameter4/input.ts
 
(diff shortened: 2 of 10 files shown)


Boshen:
Can you make space for ecma decorators? i.e. name current decorator as `LegacyDecorator`.

crates/oxc_transformer/src/decorator/legacy.rs
```diff
@@ -0,0 +1,284 @@
+//! Typescript Experimental Decorator
```
Boshen, line 1: `Experimental` has become confusing, let's not mention it except use the term legacy decorator (`--experimentalDecorators`) explicitly.

crates/oxc_transformer/src/common/helper_loader.rs
```diff
@@ -193,6 +195,8 @@ impl Helper {
             Self::ReadOnlyError => "readOnlyError",
             Self::WriteOnlyError => "writeOnlyError",
             Self::CheckInRHS => "checkInRHS",
+            Self::Decorate => "decorate",
+            Self::DecorateParam => "decorateParam",
```
Dunqing, line 199: Open to suggestions for the naming, in TypeScript, it calls `decorate` and `param`, but I think `param` is not intuitive, so I renamed it to `decorateParam`

tasks/transform_conformance/snapshots/oxc.snap.md
```diff
@@ -410,49 +456,174 @@ x Output mismatch
...
+rebuilt        : SymbolId(1): Span { start: 107, end: 108 }
 
 * typescript/decoratorOnClass8/input.ts
-x Output mismatch
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1)]
+Symbol span mismatch for "C":
+after transform: SymbolId(0): Span { start: 134, end: 135 }
+rebuilt        : SymbolId(0): Span { start: 0, end: 0 }
```
Dunqing, line 601: I am curious, does the span match matter for semantic correctness?

Dunqing:
@overlookmotel This PR is ready to review, don't try to understand what the logic does for, hopefully, it will not take you too much time.

crates/oxc_transformer/src/common/helper_loader.rs
overlookmotel: That naming sounds good to me!

tasks/transform_conformance/snapshots/oxc.snap.md
overlookmotel: I don't know what we use `SymbolTable::spans` for. Error reporting, I guess?

But this looks a bit fishy. The fact that 2 symbols with the same name are being swapped around suggests something else may be wrong. It may be related to the Scope children mismatch.

Since all these `C` symbols originate from `class C` in source, shouldn't they all have the same span anyway?

overlookmotel (approved):
Hooray!

I have "browsed" the implementation, rather than going through it in detail, as you suggested. But the lengthy comments are really helpful.

One question: What about class expressions? Can't they be decorated too?

Side note: As far as I can see, TS's implementation has some bugs. e.g.: [TS Playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&target=99#code/MYewdgzgLgBMA2BDCECmEYF4YG0C6A3AFABmIATjABTyqwCWWMADATIwDzYBMb9A1PwCUMAN5EYkmAAEAJqmASpCZBgDCYpVKnREUesBgAPKiPHaLU8nQCu5MDDUA6AJ7FLkgL5aLu-YZcmencpb20VFHQnAAcbCAALKjUhYm8iAHp0mABJGCh4+jAAazyCjFVUcn1wGBJEengMACMFRDjUGAADNQB9AEZOmAB3ZBg0WCgQLuwItAgcbjxOjKzCmCQoSry7BxASUo74EBBopxWu3oGx+JAbeFkYFphELtooQcKIenlHkFlAvYHdbHaIAGhgYBAsBenQAbohyINblAvj96FAzhUqlRZugcMw8E4TCJMNhmCkiEA)

I also I'm not sure that it's handling correctly that `C` in `export default class C {}` is a "live binding". I *think* maybe it should be transformed to `export { C as default }` rather than `export default C`. Maybe... or maybe I remember my JS semantics wrongly.

Dunqing:
> One question: What about class expressions? Can't they be decorated too?
> 

No, class expressions can't be decorated. https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=2#code/MYewdgzgLgBAgjAvDAAgEwKbBsANgQwghgGEYBvAXyA 

> Side note: As far as I can see, TS's implementation has some bugs. e.g.: [TS Playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&target=99#code/MYewdgzgLgBMA2BDCECmEYF4YG0C6A3AFABmIATjABTyqwCWWMADATIwDzYBMb9A1PwCUMAN5EYkmAAEAJqmASpCZBgDCYpVKnREUesBgAPKiPHaLU8nQCu5MDDUA6AJ7FLkgL5aLu-YZcmencpb20VFHQnAAcbCAALKjUhYm8iAHp0mABJGCh4+jAAazyCjFVUcn1wGBJEengMACMFRDjUGAADNQB9AEZOmAB3ZBg0WCgQLuwItAgcbjxOjKzCmCQoSry7BxASUo74EBBopxWu3oGx+JAbeFkYFphELtooQcKIenlHkFlAvYHdbHaIAGhgYBAsBenQAbohyINblAvj96FAzhUqlRZugcMw8E4TCJMNhmCkiEA)
> 

Luckily, this bug doesn't exist in our implementation.

OXC Output:
```js
const classes = [];
for (let i = 0; i <= 2; i++) {
        var _C;
        let C = _C = class C {
                static x() {
                        return _C.y;
                }
                static y = i;
        };
        C = _C = babelHelpers.decorate([dec], C);
        classes.push(C);
}
assert(classes[0].x() == 0);
```

> I also I'm not sure that it's handling correctly that `C` in `export default class C {}` is a "live binding". I _think_ maybe it should be transformed to `export { C as default }` rather than `export default C`. Maybe... or maybe I remember my JS semantics wrongly.


As far as I know, there seems to be no difference here, and we got output the same as [TypeScript](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=2#code/KYDwDg9gTgLgBAE2AMwIYFcA28DGnUDOBcAwnAN4C+AUNThAHYESbAB0mEA5gBQkCUQA).

tasks/transform_conformance/snapshots/oxc.snap.md
Dunqing: > I don't know what we use `SymbolTable::spans` for. Error reporting, I guess?
Yes, only the Linter will use it.

> But this looks a bit fishy. The fact that 2 symbols with the same name are being swapped around suggests something else may be wrong. It may be related to the Scope children mismatch.
> 

The scope is correct here, handling logic is 

https://github.com/oxc-project/oxc/blob/9281314ff1ed671ff90833ec46fc89cabeb526ed/crates/oxc_transformer/src/decorator/legacy_decorator.rs#L344-L356

the following example is what the logic for

Input:
```ts
export default @dec class C {}
```

Output:
```js
let C = class C {
//            ^ This is a new symbol for the class 
//  ^ This is an original class symbol that was taken from class
};
C = __decorate([
    dec
], C);
export default C;
```

The class declaration transforms into a class expression, which is assigned to a variable with the same name as that class, we have to move out the class symbol, and then as a symbol of that variable(C) because after transformed as a class expression, the class only can be self-referenced, this way make sure that all class references are still correct. 

> Since all these `C` symbols originate from `class C` in source, shouldn't they all have the same span anyway?

No, the symbol of class expression is newly generated, thus its span is 0,0, and no way to modify it now.

> (I don't think it's a blocker to merging, anyway)

Yes!

overlookmotel:
> No, class expressions can't be decorated.

OK great! That simplifies things.

> Luckily, this bug doesn't exist in our implementation.

I'm afraid it does. `var _C;` needs to be `let _C;` so that each class gets *its own* `_C`.

But we have the same problem in lots of transforms using `var` for temp vars when it should be `let` when inside a loop. We can address that problem later on as a transformer-wide thing. I'll add a failing test once this PR is merged, so we don't forget to come back to it.

We should probably raise this as a bug report on TypeScript though.

> As far as I know, there seems to be no difference here, and we got output the same as [TypeScript](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=2#code/KYDwDg9gTgLgBAE2AMwIYFcA28DGnUDOBcAwnAN4C+AUNThAHYESbAB0mEA5gBQkCUQA).

Yes, I suppose since there is no spec for legacy decorators, the only spec is "match TypeScript"!

Here's the problem I saw with `export default`: [TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=11#code/AIEwpgxgUGAeAOB7ATgFwATgGYEMCuANhhATgM5noDC6A3gL5RQ0C86AjAEwDMA3FEA)

If this was plain JS, it's legal to assign to `C`, in which case the default export needs to be a live binding (`export {C as default}`). But in TS, it looks like it's illegal to assign to a class name, so this problem doesn't arise. I wonder if our syntax checker catches such illegal assignments? Or does that fall under "type checking" and therefore out of scope for Oxc?

Dunqing:
> I'm afraid it does. `var _C;` needs to be `let _C;` so that each class gets _its own_ `_C`.
> 
> But we have the same problem in lots of transforms using `var` for temp vars when it should be `let` when inside a loop. We can address that problem later on as a transformer-wide thing. I'll add a failing test once this PR is merged, so we don't forget to come back to it.
> 

I got it!

> Yes, I suppose since there is no spec for legacy decorators, the only spec is "match TypeScript"!
> 
> Here's the problem I saw with `export default`: [TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=11#code/AIEwpgxgUGAeAOB7ATgFwATgGYEMCuANhhATgM5noDC6A3gL5RQ0C86AjAEwDMA3FEA)
> 
> If this was plain JS, it's legal to assign to `C`, in which case the default export needs to be a live binding (`export {C as default}`). But in TS, it looks like it's illegal to assign to a class name, so this problem doesn't arise. I wonder if our syntax checker catches such illegal assignments? Or does that fall under "type checking" and therefore out of scope for Oxc?
> 

This is just a "type checking" thing, we don't need to care about.

Dunqing:
### Merge activity

* **Feb 9, 9:51 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #8618, opened 2025-01-20
refactor(minifier): remove all the unnecessary fake ast passes

Boshen:
This also removes handling of making cjs-module-lexer to work.


Boshen:
Will fix the performance regression.

PR #8622, opened 2025-01-20
feat(semantic): memoize `Bindings` hashmap's key hash (POC)

Boshen:
Attempt to reduce the total number of hashes here

https://github.com/oxc-project/oxc/blob/6f95cd599a8870da410c03a07e9941b0d0ed3dfe/crates/oxc_semantic/src/builder.rs#L490

This is a dump of code with no intention of being merged.

### DIFF
--- a/crates/oxc_semantic/src/scope.rs
+++ b/crates/oxc_semantic/src/scope.rs
@@ -1,6 +1,12 @@
-use std::{fmt, mem};
+use std::{
+    borrow::Borrow,
+    fmt,
+    hash::{BuildHasherDefault, Hash, Hasher},
+    mem,
+    ops::Deref,
+};
 
-use oxc_allocator::{Allocator, HashMap as ArenaHashMap, Vec as ArenaVec};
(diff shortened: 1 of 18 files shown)


crates/oxc_semantic/src/scope.rs
```diff
@@ -398,23 +480,26 @@ impl ScopeTree {
...
+            inner.bindings[scope_id].insert(Key::new(name), symbol_id);
         });
     }
 
     /// Remove an existing binding from a scope.
     pub fn remove_binding(&mut self, scope_id: ScopeId, name: &str) {
-        self.cell.with_dependent_mut(|_allocator, inner| {
-            inner.bindings[scope_id].remove(name);
+        self.cell.with_dependent_mut(|allocator, inner| {
+            let key = Key::new(allocator.alloc_str(name));
```
Boshen, line 490: This is wasteful, need to use the entry API instead.

Boshen:
Only 2%?

PR #8634, opened 2025-01-21
feat(tasks/prettier): Visualize test details

leaysgur:
For example, if there are 3 tests in 1 spec and all of them fail:

- 💥💥💥

After some implementation, if 1 test passes:

- 💥💥✨

However, in this case, the coverage as number does not change.

This PR visualizes these details for better mental well-being. 😃
(But in practice, specs are so detailed, there is not much opportunity for this...)


Boshen:
neat!

PR #8638, opened 2025-01-21
feat(minifier): compress `typeof foo === 'object' && foo !== null` to `typeof foo == 'object' && !!foo`

sapphi-red:
If `typeof foo == 'object'`, then `foo` is guaranteed to be an object or null. In that case, `foo !== null` can be replaced with `!!foo` because objects return `true` for `!!foo` and null returns `false` for it.

**References**
- [Spec of `typeof`](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-typeof-operator)
- [Spec of `!`](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-logical-not-operator)
- [Spec of `ToBoolean`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toboolean)

### DIFF
--- a/crates/oxc_minifier/src/peephole/substitute_alternate_syntax.rs
+++ b/crates/oxc_minifier/src/peephole/substitute_alternate_syntax.rs
@@ -7,6 +7,7 @@ use oxc_ecmascript::{
 };
 use oxc_span::cmp::ContentEq;
 use oxc_span::GetSpan;
+use oxc_span::SPAN;
 use oxc_syntax::{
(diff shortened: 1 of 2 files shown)


Boshen:
### Merge activity

* **Jan 21, 9:58 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #8647, opened 2025-01-21
fix(linter): fix github `endColumn` output

Sysix:
added `start` and `end` for `Info`, so every reporter can use both if they want.
Then end calculation is a bit hacky, but i looks like it works.


Boshen:
We need a bit of snapshot testing for these reporters at some point 😅

PR #8653, opened 2025-01-22
fix(semantic): fix const assertions in `UnresolvedReferencesStack`

overlookmotel:
The `_SIZE_CHECK` assertion was ignored, because `const`s are only evaluated if they're referenced in a function which is called. Fix that by referencing the const in `UnresolvedReferencesStack::new`.

Also add more assertions to cover all the invariants.


Boshen:
### Merge activity

* **Jan 22, 8:12 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #8666, opened 2025-01-22
feat(linter): add DiagnosticResult to the Reporters for receiving a sub part result

Sysix:
We are currently outputting only for the default-outputter some extra information:

https://github.com/oxc-project/oxc/blob/3be03926e8a5097e5c1fe249b8bff0009ec4468d/apps/oxlint/src/result.rs#L61-L87

My goal is that all information will be passed to our new DiagnosticReporter / OutputFormatter.
This will break the output format in the next PR. **Merging this PR is the OK for me to make this change** ⚠️ 
The only breaking point:
`"Found {number_of_warnings} warning{} and {number_of_errors} error{}."` will still be outputted when `max_warnings_exceeded` is true.

Because this is something the `DiagnosticReporter` should do and not the `OutputFormatter`.

The end goal is:
- no `println!`, our `OutputFormatter` and his `DiagnosticReporter` will return `Option<String>` and we output it the our `stdout`
- `LintResult` will only handle `ExitCode` result and nothing more
- `stdout` can be changed from outside (for the next part)
- add snapshots with a custom `stdout`

I do not know if all goals can be done easily. Last two parts should be a bit tricky for me, never did test setups in rust. 
But we do never stop to learn ;)

### DIFF
--- a/crates/oxc_diagnostics/src/service.rs
+++ b/crates/oxc_diagnostics/src/service.rs
@@ -1,11 +1,13 @@
 use std::{
-    cell::Cell,
     io::{ErrorKind, Write},
     path::{Path, PathBuf},
     sync::{mpsc, Arc},
 };
 
-use crate::{reporter::DiagnosticReporter, Error, NamedSource, OxcDiagnostic, Severity};
+use crate::{
+    reporter::{DiagnosticReporter, DiagnosticResult},
+    Error, NamedSource, OxcDiagnostic, Severity,
+};
 
 pub type DiagnosticTuple = (PathBuf, Vec<Error>);
 pub type DiagnosticSender = mpsc::Sender<Option<DiagnosticTuple>>;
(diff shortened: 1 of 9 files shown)


crates/oxc_diagnostics/src/service.rs
```diff
@@ -165,20 +142,27 @@ impl DiagnosticService {
     /// # Panics
     ///
     /// * When the writer fails to write
-    pub fn run(&mut self, writer: &mut dyn Write) {
+    ///
+    /// ToDo:
+    /// We are passing [`DiagnosticResult`] to the [`DiagnosticReporter`] already
+    /// currently for the GraphicalReporter there is another extra output,
+    /// which does some more things. This is the reason why we are returning it.
+    /// Let's check at first it we can easily change for the default output before removing this return.
```
Sysix, line 150: fyi

PR #8674, opened 2025-01-23
feat(linter): Add "strict" option to `promise/prefer-await-to-then` rule

NeilTheFisher:
Implementation of the strict option [specified in the eslint docs for `promise/prefer-await-to-then`](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md)

```js
// Examples of **incorrect** code with `{ strict: true }`:
async function hi() {
  await thing().then(x => {})
}
```


camc314 (approved):
thanks!

PR #8691, opened 2025-01-24
fix(transformer/typescript): enum merging when same name declared in outer scope

wan9chi:
```typescript
var x = 10;
enum Merge { x = Math.random() }
enum Merge {
    y = x // <-- refers to Merge.x
}
```

This case wasn't covered in #8543 and by the [babel test case](https://github.com/babel/babel/blob/e568916ef3e5fd288039a65c904b12a7eba2a73a/packages/babel-plugin-transform-typescript/test/fixtures/enum/non-constant-member-reference/input.ts).  To handle it we still have to go through the scope ancestors.


Dunqing:
I spend a bit of time to figure out what happens here, I think storing scopes of expression and using it would be easier and more efficient because we don't need to go through the whole ancestor scope, and usually it only happens in the enum scope.

Dunqing:
@branchseer Please review https://github.com/oxc-project/oxc/pull/8691/commits/9729ad238dd7a5ce5c2c4a1102fe81e6cd11cd5d, if you don't think so I am happy you to revert or modify it.

branchseer:
@Dunqing LGTM. I agree your approach is more optimal.

In 154d4c53c367ca02873969d3f89d836d12fc0b4f I took the liberty to add a non-`Option` version of `first()` for `NonEmptyStack`. cc @overlookmotel

crates/oxc_transformer/src/typescript/enum.rs
```diff
@@ -533,23 +535,68 @@ impl<'a, 'ctx> IdentifierReferenceRename<'a, 'ctx> {
...
+            // The resolved symbol is declared outside the enum,
+            // and we have checked that the name exists in previous_enum_members:
+            //
+            // ```ts
+            // const A = 0;
+            // enum Foo { A }
+            // enum Foo { B = A }
+            //                ^ This should be renamed to Foo.A
+            // ```
+        || !self.scope_stack.contains(&symbol_scope_id)
```
branchseer, line 587: @Dunqing  Maybe a wild idea: is it ok to assume that a inner scope id is always greater than a outer scope id (because the inner scope is created later)? If so we can just do `enum_scope_id >= symbol_scope_id` here.

Dunqing: Actually, I think it will work, but it's not a good idea, because we can't guarantee that we won't break that assumption in the future.

Dunqing (approved):
Thanks！

overlookmotel:
> In [154d4c5](https://github.com/oxc-project/oxc/commit/154d4c53c367ca02873969d3f89d836d12fc0b4f) I took the liberty to add a non-`Option` version of `first()` for `NonEmptyStack`. cc @overlookmotel

Sorry I was slow to come back on this. Good idea! I've implemented in #8908.

PR #8703, opened 2025-01-25
chore(oxc_release.toml): add `crates/oxc_language_server/Cargo.toml`

### LINKED ISSUE #8603: oxc_language server still on version 0.0.1
Would like to see a new version when releasing it with oxlint.

Maybe add it to https://github.com/oxc-project/oxc/blob/main/oxc_release.toml?
Probably the CI needs to be updated too.

This is confusing for debugging reason:
```
2025-01-19 14:27:52.016 [info] [2025-01-19T13:27:52Z INFO  oxc_language_server] initialize: Options { run: OnType, enable: true, config_path: ".oxlintrc.json" }
[2025-01-19T13:27:52Z INFO  oxc_language_server] language server version: "0.0.1"
```

re-taro:
close #8603 

Add `crates/oxc_language_server/Cargo.toml` to release LS with linter's release

### DIFF
--- a/oxc_release.toml
+++ b/oxc_release.toml
@@ -20,4 +20,5 @@ versioned_files = [
   "crates/oxc_linter/Cargo.toml",
   "editors/vscode/package.json",
   "npm/oxlint/package.json",
+  "crates/oxc_language_server/Cargo.toml",
 ]


Boshen:
This is not the correct fix. We need to find where the `language server version: "0.0.1"` value is set, and set it correctly when we build the binary in CI.

re-taro:
> This is not the correct fix. This file is used for publishing crates.
> 
> We need to find where the `language server version: "0.0.1"` value is set, and set it correctly when we build the binary in CI.

Oops, my bad. I'll fix it.

re-taro:
Hmmm, there appears to be no problem with the implementation.

https://github.com/oxc-project/oxc/blob/06f3287685344184d86b61659037b831c2ba8854/crates/oxc_language_server/src/main.rs#L102

It is thought that this will change if the version of oxc_language_server is also upgraded and built according to the release of oxlint...

PR #8756, opened 2025-01-27
feat(linter): Junit reporter

### LINKED ISSUE #7960: linter: support JUnit output format
Many CI tools can parse JUnit output format and display the results in a more user-friendly UI. Adding support for JUnit output would greatly improve the ability to understand and analyze issues when running oxlint in CI environments.

tapanprakasht:
closes https://github.com/oxc-project/oxc/issues/7960


Cargo.toml
```diff
@@ -178,6 +178,7 @@ phf = "0.11.2"
 pico-args = "0.5.0"
 prettyplease = "0.2.25"
 project-root = "0.2.2"
+quick-junit = "0.5.1"
```
Sysix, line 181: I do not like that we need a dependency just for a output format. 
You could probably self build the XML, like we did with the checkstyle format:
https://github.com/oxc-project/oxc/blob/e929f268260b10e6ec0076bef2b150b0cb1469f3/apps/oxlint/src/output_formatter/checkstyle.rs#L40-L67

apps/oxlint/src/output_formatter/mod.rs
```diff
@@ -40,6 +43,7 @@ impl FromStr for OutputFormat {
             "checkstyle" => Ok(Self::Checkstyle),
             "github" => Ok(Self::Github),
             "stylish" => Ok(Self::Stylish),
+            "junit" => Ok(Self::JUnit),
```
Sysix, line 46: in this file there is a snapshot test for the different formats. Could you please add one test for it too?
The test should also work in windows.

Cargo.toml
tapanprakasht: Done. Removed quick-junit dependency.

apps/oxlint/src/output_formatter/mod.rs
tapanprakasht: Done

Sysix (approved):

PR #8770, opened 2025-01-28
refactor(minifier): port esbuild's `mangleStmts`

Boshen:


Boshen:
@sapphi-red Please take a look, The previous google closure compiler architecture is not holding up well :-/

The current direction for performance is 1 AST pass with lots of small "minimize expression" calls when new expressions are created ... we'll get there!

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -0,0 +1,380 @@
...
+                            };
+                            let b = last_return.argument.as_mut().unwrap();
+                            let argument = Self::join_sequence(&mut expr_stmt.expression, b, ctx);
+                            let right_span = last_return.span;
+                            let last_return_stmt =
+                                ctx.ast.statement_return(right_span, Some(argument));
+                            result.push(last_return_stmt);
+                        }
+                        // Merge the last two statements
+                        Statement::IfStatement(if_stmt) => {
```
sapphi-red, line 97: Similar to the previous question, I wonder if this can be done inside `self.minimize_statement` by running `self.minimize_statement` when the last item of `result` changed.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -0,0 +1,380 @@
...
+                                ctx.ast.alloc(var_decl),
+                            ));
+                            result.pop();
+                            self.mark_current_function_as_changed();
+                        }
+                    }
+                    _ => {}
+                }
+                result.push(Statement::ForStatement(for_stmt));
+            }
```
sapphi-red, line 346: (I guess you're aware of it, but in case) I think `ForInStatement`, `LabeledStatement`, `BlockStatement` is missing here.
https://github.com/oxc-project/oxc/blob/29417ddc03085a49caaf31b2a6477a24cf4544f5/crates/oxc_minifier/src/peephole/statement_fusion.rs#L70-L80
https://github.com/oxc-project/oxc/blob/29417ddc03085a49caaf31b2a6477a24cf4544f5/crates/oxc_minifier/src/peephole/statement_fusion.rs#L134-L150
Maybe `BlockStatement` should be handled in a different place though.

sapphi-red, line 346: (I guess you're aware of it, but in case) I think merging `VariableDeclaration` + `ForInStatement`/`ForOfStatement` is missing too.
https://github.com/oxc-project/oxc/blob/0fcff207c7e538bc2becf2197e00efeefd6d0ffa/crates/oxc_minifier/src/peephole/collapse_variable_declarations.rs#L177-L215

sapphi-red (commented):
Great! I left some comments

PR #8788, opened 2025-01-30
feat(tasks/prettier): Show total match ratio with updating snapshot format

leaysgur:
Previously, #8634 visualized each spec result, but this was not enough.

If the implementation has been updated but the spec still fails, there is no way to know if there is an improvement or not.

This time, finally, all progress is visualized! 🤩 

image

e.g. here `arrow_function_expression` is mostly passing, except for comment handling.


Boshen:
NICE!

Boshen:
NICE!

PR #8817, opened 2025-01-31
refactor(transformer): don't need to transform `JSXExpression::EmptyExpression` in JSX attributes

Dunqing:


crates/oxc_transformer/src/jsx/jsx_impl.rs
```diff
@@ -872,8 +872,10 @@ impl<'a> JsxImpl<'a, '_> {
                     // SAFETY: `ast.copy` is unsound! We need to fix.
                     unsafe { ctx.ast.copy(e.to_expression()) }
                 }
-                JSXExpression::EmptyExpression(e) => {
-                    ctx.ast.expression_boolean_literal(e.span, true)
+                JSXExpression::EmptyExpression(_) => {
+                    // `<div attr={}></div>`
+                    //            ^^ Invalid empty expression here
+                    unreachable!()
```
Boshen, line 878: We should not make this `unreachable` in situations where the semantic checker is not run or missed the check, do not emit anything instead.

Dunqing: That makes sense. To achieve "do not emit anything" here we must return an `Expression, which will cause more changes, I will do this later.

Dunqing: This is part of `typescript-eslint`, See https://ast.sxzz.dev/#eNpNzMEKwjAMgOFXGTlPvI+5o08gnnoJNUhH15SmFaX03Y1WZLfk+0kqeJhgxQeKTS5mGCEq5FekDgcS78LHrfp8Zh7wVNtiwlCbCfNRZdHKWqviYEC4JEsXfWFg0n3jW/E6jz2T3fBKSRyH3j1mkvzvCcP9e5lToZ+t8txJg/YGroc72A==.

I agree to throw an error in the parser as this is a pointless AST node.

PR #8821, opened 2025-02-01
test(oxlint): test two real rules with same name but from different plugins

shulaoda:
Related to #8814 

This test case is designed to verify that two different rules, having the same name but coming from different plugins, are treated as distinct rules rather than simple aliases.

In `unicorn`, there is only one rule, `no-negated-condition`, which is aliased to `eslint`.

https://github.com/oxc-project/oxc/blob/655b2126889a3307929a1a20047b317c1eefb140/crates/oxc_linter/src/config/rules.rs#L166-L178


Sysix (approved):
You are right! Sorry for the trouble

PR #8443, opened 2025-01-12
feat(minifier): minify `String::concat` into template literal

sapphi-red:
Compress `"".concat(a, "b", c)` into `` `${a}b${c}` ``.

~~I'm not sure if this should be merged. It works for antd but it doesn't for typescript.~~ Now with #8839, it works well for typescript as well.

**References**
- [Spec of `String::concat`](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.concat)
- [Spec of template literal](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-template-literals-runtime-semantics-evaluation)

PR #8839, opened 2025-02-02
feat(codegen): prefer backquotes over double / single quotes

sapphi-red:
When thinking about why #8443 made the gzip size increased, I thought it maybe because the backquotes appear more than before the PR. So I changed the codegen to prefer backquotes than double/single quotes. It seems this separate change have a positive effect with some projects.
My guess is that in most projects, back quotes does not appear in string literals than double quotes and single quotes.

The gzipped size slightly increases for `bundle.min.js`, `victory.js`, `antd.js`, but those will be decreased by #8443 except for `bundle.min.js`.

### DIFF
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -952,7 +952,7 @@ impl Gen for ImportDeclaration<'_> {
             p.print_str("from");
         }
         p.print_soft_space();
-        self.source.print(p, ctx);
+        p.print_quoted_utf16(&self.source.value, false);
         if let Some(with_clause) = &self.with_clause {
             p.print_soft_space();
             with_clause.print(p, ctx);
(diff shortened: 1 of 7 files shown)


sapphi-red:
* **#8443** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8443?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8839** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8839?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/8839?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=8839).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sapphi-red:
It seems I need to add `allow_backtick: false` to more places for TypeScript codegen.

longlho:
@sapphi-red I'm getting an issue/bug with this. The gist of it is we have something like:

```
const pathRegex = /(['"])((\w+:(\/?\/?))?[^\s,]+)\1/;

  function getImportPath(importFn) {
    return importFn.toString().match(pathRegex)?.[2] ?? null;
  }

  async function withRetry(importFn) {
    try {
      return await importFn();
    } catch {
      const path = getImportPath(importFn);
      if (path) {
        // Retry with cache-busting to bypass stale CDN cache
        return await import(`${path}?r=${Date.now()}`);
      }
      throw new Error('Dynamic import failed');
    }
  }

  // Usage:
  withRetry(() => import("./MyComponent-abc123.js"));
```

Before minification, `importFn.toString()` returns `() => import("./MyComponent-abc123.js")` — the path is extractable. After Rolldown minification, it returns 
```
() => import(`./MyComponent-abc123.js`)
```
 — the regex no longer matches because backticks aren't quotes.

Terser and esbuild both preserve string literal quote types. The minifier should not convert string literals to template literals, as it changes observable behavior via .toString().

Is there a way to opt out of this behavior?

sapphi-red:
> The minifier should not convert string literals to template literals, as it changes observable behavior via .toString().

No minifier preserves the output of `Function.prototype.toString` and Oxc won't as well.

I guess you can modify the code to handle backquotes as well.

longlho:
OK that's fair we'll handle this on our side

PR #8840, opened 2025-02-02
fix(linter): correctly replace rule severity with duplicate rule name configurations

shulaoda:
See the snapshot changes of `apps/oxlint/fixtures/typescript_eslint/eslintrc.json`.


Sysix (approved):
Thanks for fixing this bug.

PR #8854, opened 2025-02-03
perf(linter): reduce unnecessary rule method executions to improve performance

shulaoda:
We can make full use of `should_run` to avoid unnecessary iterations over nodes and rules.

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -25,6 +25,7 @@ pub mod table;
 use std::{path::Path, rc::Rc, sync::Arc};
 
 use oxc_semantic::{AstNode, Semantic};
+use rule::ShouldRunMeta;
 
 pub use crate::{
     config::{
@@ -105,11 +106,11 @@ impl Linter {
             Rc::new(ContextHost::new(path, semantic, module_record, self.options, config));
 
         let rules = rules.iter().filter_map(|rule| {
-            let state = rule.should_run(&ctx_host);
-            if state.enable {
-                Some(Rc::new((rule, Rc::clone(&ctx_host).spawn(rule), state)))
-            } else {
+            let meta = rule.should_run(&ctx_host);
+            if meta.is_empty() {
                 None
+            } else {
+                Some((Rc::new((rule, Rc::clone(&ctx_host).spawn(rule))), meta))
             }
         });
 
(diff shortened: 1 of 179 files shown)


shulaoda:
> ### Benchmarks breakdown
> Benchmark	`BASE`	`HEAD`	Change
> ⚡	`linter[RadixUIAdoptionSection.jsx]`	2.6 ms	2.3 ms	+9.05%
> ⚡	`linter[cal.com.tsx]`	1.2 s	1.1 s	+8.09%
> ❌	`linter[checker.ts]`	2.8 s	3.5 s	-20.85%

When running on my windows machine (`13th Gen Intel(R) Core(TM) i5-13600KF 3.50 GHz`), the performance of `checker.ts` significantly improved by `50%`. I believe this is influenced by the configuration of each person's computer. At this point, I have not enabled the optimizations from #6600.

shulaoda:
> ### Benchmarks breakdown
> Benchmark	`BASE`	`HEAD`	Change
> ⚡	`linter[RadixUIAdoptionSection.jsx]`	2.6 ms	2.3 ms	+9.05%
> ⚡	`linter[cal.com.tsx]`	1.2 s	1.1 s	+8.09%
> ❌	`linter[checker.ts]`	2.8 s	3.5 s	-20.85%

When running on my windows machine (`13th Gen Intel(R) Core(TM) i5-13600KF 3.50 GHz`), the performance of `checker.ts` significantly improved by `50%`. I believe this is influenced by the configuration of each person's computer. At this point, I have not enabled the optimizations from #6600.

shulaoda:
Introducing `bitflags` did not result in a performance improvement, and in fact, it caused a slight decrease of around `1%` to `2%`. However, it benefits memory management and provides better maintainability, so I have decided not to roll it back.

Boshen (commented):
I like the overall direction, but dislike requiring 

```
    fn should_run(&self, _: &crate::ContextHost) -> crate::rule::ShouldRunMeta {
        crate::rule::ShouldRunMeta::IS_RUN_ONCE
    }
```

everywhere.

shulaoda:
> I like the overall direction, but dislike requiring
> 
> ```
>     fn should_run(&self, _: &crate::ContextHost) -> crate::rule::ShouldRunMeta {
>         crate::rule::ShouldRunMeta::IS_RUN_ONCE
>     }
> ```
> 
> everywhere.

By default, If use `run`, It shouldn’t require explicitly implementing `should_run`. Perhaps we can use a procedural `macro` to automate this? Do you have any good ideas?

camchenry:
Agreed with @Boshen: something like this is definitely worth doing, but maybe we can polish the  experience a little bit more. My suggestion would be that maybe we can make it as part of the `declare_oxc_lint!` macro: https://github.com/oxc-project/oxc/blob/bec8feecf20070d4f2a00aa5397c5b535fde7d4f/crates/oxc_macros/src/declare_oxc_lint.rs#L62

So maybe it would look more like this (without doc comments):

```rs
declare_oxc_lint!(
    SomeLintRule(IsRunOnSymbol),
    oxc,
    nursery
);
```

Not 100% sure how to parse it like that though, not too familiar with macros

shulaoda:
> So maybe it would look more like this (without doc comments):
> 
> ```rust
> declare_oxc_lint!(
>     SomeLintRule(IsRunOnSymbol),
>     oxc,
>     nursery
> );
> ```
> 
> Not 100% sure how to parse it like that though, not too familiar with macros

This doesn't seem feasible (because `(` and `)` cannot be parsed as a token), we can:

```rust
declare_oxc_lint!(
    SomeLintRule,
    oxc,
    nursery,
    IsRunOnSymbol,
);
```

or
```rust
#[should_run(is_run_once)]
pub struct PreferToBeObject;
```

Boshen:
Ok my decision is that the total amount of changes is not worth it for `+8.67%`, unless we find a better approach to get the same performance boost.

PR #8858, opened 2025-02-03
feat(oxc_language_server): implement `oxc.fixAll` workspace command

marekvospel:
This pull request focuses on optimizing the implementation of the `oxlint.applyAllFixesFile` vscode command by adding the command interface to the LSP instead of requesting all code actions and executing the preferred ones. This PR contains an abstraction above the LSP workspace commands, the `oxc.fixAll` command itself and minor changes to the vscode extension.

Since the `workspace/executeCommand` handler is new, I've created an abstraction to register the commands and parse their arguments. While it isn't necessary, I feel like it makes future additions to the LS easier.

I tested the command in both vscode and neovim, doubt it will be hard to use this in the zed / intellij projects.

### DIFF
--- a/crates/oxc_language_server/Cargo.toml
+++ b/crates/oxc_language_server/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 
 [[bin]]
 name = "oxc_language_server"
-test = false
+test = true
 doctest = false
 
 [dependencies]
--- a/crates/oxc_language_server/src/capabilities.rs
+++ b/crates/oxc_language_server/src/capabilities.rs
@@ -20,17 +20,17 @@ pub struct Capabilities {
 impl From<ClientCapabilities> for Capabilities {
     fn from(value: ClientCapabilities) -> Self {
         // check if the client support some code action literal support
-        let code_action_provider = value.text_document.is_some_and(|capability| {
-            capability.code_action.is_some_and(|code_action| {
-                code_action.code_action_literal_support.is_some_and(|literal_support| {
+        let code_action_provider = value.text_document.as_ref().is_some_and(|capability| {
+            capability.code_action.as_ref().is_some_and(|code_action| {
+                code_action.code_action_literal_support.as_ref().is_some_and(|literal_support| {
                     !literal_support.code_action_kind.value_set.is_empty()
                 })
             })
         });
         let workspace_apply_edit =
-            value.workspace.clone().is_some_and(|workspace| workspace.apply_edit.is_some());
+            value.workspace.as_ref().is_some_and(|workspace| workspace.apply_edit.is_some());
         let workspace_execute_command =
-            value.workspace.is_some_and(|workspace| workspace.execute_command.is_some());
+            value.workspace.as_ref().is_some_and(|workspace| workspace.execute_command.is_some());
 
         Self { code_action_provider, workspace_apply_edit, workspace_execute_command }
     }
@@ -40,10 +40,7 @@ impl From<Capabilities> for ServerCapabilities {
     fn from(value: Capabilities) -> Self {
         let commands = LSP_COMMANDS
             .iter()
-            .filter_map(|c| match c.available(value.clone()) {
-                true => Some(c.command_id()),
-                false => None,
-            })
+            .filter_map(|c| if c.available(value.clone()) { Some(c.command_id()) } else { None })
             .collect();
 
         Self {
@@ -69,9 +66,10 @@ impl From<Capabilities> for ServerCapabilities {
             } else {
                 None
             },
-            execute_command_provider: match value.workspace_execute_command {
-                true => Some(ExecuteCommandOptions { commands, ..Default::default() }),
-                false => None,
+            execute_command_provider: if value.workspace_execute_command {
+                Some(ExecuteCommandOptions { commands, ..Default::default() })
+            } else {
+                None
             },
             ..ServerCapabilities::default()
         }
@@ -162,6 +160,7 @@ mod test {
                         code_action_kind: CodeActionKindLiteralSupport {
                             // nvim 0.10.3
                             value_set: vec![
+                                #[allow(clippy::manual_string_new)]
                                 "".into(),
                                 "quickfix".into(),
                                 "refactor".into(),
--- a/crates/oxc_language_server/src/capabilities.rs
+++ b/crates/oxc_language_server/src/capabilities.rs
@@ -1,14 +1,19 @@
 use tower_lsp::lsp_types::{
-    ClientCapabilities, CodeActionKind, CodeActionOptions, CodeActionProviderCapability, OneOf,
-    ServerCapabilities, TextDocumentSyncCapability, TextDocumentSyncKind, WorkDoneProgressOptions,
-    WorkspaceFoldersServerCapabilities, WorkspaceServerCapabilities,
+    ClientCapabilities, CodeActionKind, CodeActionOptions, CodeActionProviderCapability,
+    ExecuteCommandOptions, OneOf, ServerCapabilities, TextDocumentSyncCapability,
+    TextDocumentSyncKind, WorkDoneProgressOptions, WorkspaceFoldersServerCapabilities,
+    WorkspaceServerCapabilities,
 };
+
+use crate::commands::LSP_COMMANDS;
 
 pub const CODE_ACTION_KIND_SOURCE_FIX_ALL_OXC: CodeActionKind =
     CodeActionKind::new("source.fixAll.oxc");
 
+#[derive(Clone)]
 pub struct Capabilities {
     pub code_action_provider: bool,
+    pub workspace_edit: bool,
 }
 
 impl From<ClientCapabilities> for Capabilities {
@@ -21,13 +26,23 @@ impl From<ClientCapabilities> for Capabilities {
                 })
             })
         });
+        let workspace_edit =
+            value.workspace.is_some_and(|capability| capability.workspace_edit.is_some());
 
-        Self { code_action_provider }
+        Self { code_action_provider, workspace_edit }
     }
 }
 
 impl From<Capabilities> for ServerCapabilities {
     fn from(value: Capabilities) -> Self {
+        let commands = LSP_COMMANDS
+            .iter()
+            .filter_map(|c| match c.available(value.clone()) {
+                true => Some(c.command_id()),
+                false => None,
+            })
+            .collect();
+
         Self {
             text_document_sync: Some(TextDocumentSyncCapability::Kind(TextDocumentSyncKind::FULL)),
             workspace: Some(WorkspaceServerCapabilities {
@@ -51,6 +66,10 @@ impl From<Capabilities> for ServerCapabilities {
             } else {
                 None
             },
+            execute_command_provider: Some(ExecuteCommandOptions {
+                commands,
+                ..Default::default()
+            }),
             ..ServerCapabilities::default()
         }
     }
--- a/crates/oxc_language_server/src/capabilities.rs
+++ b/crates/oxc_language_server/src/capabilities.rs
@@ -79,7 +79,8 @@ impl From<Capabilities> for ServerCapabilities {
 mod test {
     use tower_lsp::lsp_types::{
         ClientCapabilities, CodeActionClientCapabilities, CodeActionKindLiteralSupport,
-        CodeActionLiteralSupport, TextDocumentClientCapabilities,
+        CodeActionLiteralSupport, TextDocumentClientCapabilities, WorkspaceClientCapabilities,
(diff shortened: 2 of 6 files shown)


crates/oxc_language_server/src/capabilities.rs
```diff
@@ -60,7 +79,8 @@ impl From<Capabilities> for ServerCapabilities {
 mod test {
```
marekvospel, line 79: I'm not sure the tests are being run. I tried adding `assert!(false)` and didn't get any test to fail... maybe it needs to be enabled in the LS Cargo.toml?

crates/oxc_language_server/src/commands.rs
```diff
@@ -0,0 +1,115 @@
...
+        if let Some(value) = backend.diagnostics_report_map.get(&url.to_string()) {
+            for report in value.iter() {
+                if let Some(fixed) = &report.fixed_content {
+                    edits.push(TextEdit { range: fixed.range, new_text: fixed.code.clone() })
+                }
+            }
+            let _ = backend
+                .client
+                .send_request::<ApplyWorkspaceEdit>(ApplyWorkspaceEditParams {
+                    label: Some("Oxlint fix all".into()),
```
marekvospel, line 101: This message is shown to the user (nvim) and in the undo history (vscode). Maybe better label is applicable (such as Oxlint: {n} fixes applied)

crates/oxc_language_server/src/commands.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+}
+
+pub struct FixAllCommand;
+
+#[derive(Deserialize)]
+pub struct FixAllCommandArg {
+    uri: String,
+}
```
marekvospel, line 68: As talked about in discord, we could add `Option<FixType>` and fall back to the `SafeFix` or `SafeFixOrSuggestion` in the future (and either do another ServerLinter run or filter current diagnostic)

(note for my future PR, you can close this once you read it)

marekvospel:
I've added some notes that I'd like to have your input on 😀

crates/oxc_language_server/src/capabilities.rs
```diff
@@ -21,13 +26,23 @@ impl From<ClientCapabilities> for Capabilities {
                 })
             })
         });
+        let workspace_edit =
+            value.workspace.is_some_and(|capability| capability.workspace_edit.is_some());
```
Sysix, line 30:
```
let workspace_execute_command = value.workspace.is_some_and(|workspace| workspace.execute_command.is_some());
```

See ClientCapabilities here:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand

crates/oxc_language_server/src/capabilities.rs
Sysix: Good catch, I didn't noticed it. Feel free to create a bug issue

crates/oxc_language_server/src/capabilities.rs
```diff
@@ -51,6 +66,10 @@ impl From<Capabilities> for ServerCapabilities {
             } else {
                 None
             },
+            execute_command_provider: Some(ExecuteCommandOptions {
```
Sysix, line 69: Only provide the Commands when the server supports them (reflecting the client support)

crates/oxc_language_server/src/capabilities.rs
```diff
@@ -92,12 +112,20 @@ mod test {
...
+                }),
+                ..WorkspaceClientCapabilities::default()
+            }),
             ..ClientCapabilities::default()
         };
 
         let capabilities = Capabilities::from(client_capabilities);
 
         assert!(capabilities.code_action_provider);
+        assert!(capabilities.workspace_edit);
```
Sysix, line 128: Could you please create own tests for `workspace_execute_command`.
Feel free to create your own tests for nvim too.

Please use right config from the program.

VsCode:
```
                    execute_command: Some(
                        DynamicRegistrationClientCapabilities {
                            dynamic_registration: Some(
                                true,
                            ),
                        },
                    ),
```

IntelliJ:
```
    executeCommand = ExecuteCommandCapabilities [
      dynamicRegistration = false
    ]
```

crates/oxc_language_server/src/commands.rs
Sysix: I could not tests this with vscode. Could to provide me the shortcude `CTRL + Shift + P` did not work.
I guess maybe we need to extend the `package.json` information for vscode like eslint did:

https://github.com/microsoft/vscode-eslint/blob/main/package.json#L564-L595

My test setup:
```
cd editors/vscode
pnpm build
pnpm install-extension
// restart vscode, but wait a bit to clear the cache
```

Sysix (commented):
Nice work. I could not test it (see one discussion) but from the outside this looks great 👍

crates/oxc_language_server/src/capabilities.rs
Sysix: Good catch! I did not notice it. Feel free to create a bug issue

marekvospel:
@Sysix 
I tested the code on the default vscodium installation on nixos and the command worked.

Do you perhaps have `oxlint` installed in the project you are testing in? The vscode extension prefers the binary in node modules if available. (Resulting in `Method not found` error. - This is not optimal, but I'm not sure how to handle this other than creating an issue that users can find telling them to update oxlint to latest)

The commands in package.json are only the `CTRL + Shift + P` commands, not the LSP ones (at least from what I know)

Sysix:
> The commands in package.json are only the CTRL + Shift + P commands, not the LSP ones (at least from what I know)

How to I tell vscode to execute the LSP Command?

> Do you perhaps have oxlint installed in the project you are testing in?

Yes. But it still works (somehow). I printed the vscode ClientCapability with this method for you.

Sysix:
Thanks for the explanation. I tested this command. worked!

crates/oxc_language_server/src/capabilities.rs
marekvospel: I enabled it in this PR since this doesn't seem significant enough to create a separate one.

crates/oxc_language_server/src/capabilities.rs
marekvospel: My bad, I didn't read the test name :sweat_smile: I thought it was a general editor capability test.

crates/oxc_language_server/src/capabilities.rs
```diff
@@ -20,17 +20,17 @@ pub struct Capabilities {
...
                     !literal_support.code_action_kind.value_set.is_empty()
                 })
             })
         });
         let workspace_apply_edit =
-            value.workspace.clone().is_some_and(|workspace| workspace.apply_edit.is_some());
+            value.workspace.as_ref().is_some_and(|workspace| workspace.apply_edit.is_some());
         let workspace_execute_command =
-            value.workspace.is_some_and(|workspace| workspace.execute_command.is_some());
+            value.workspace.as_ref().is_some_and(|workspace| workspace.execute_command.is_some());
```
marekvospel, line 33: No need for clone! :tada:

crates/oxc_language_server/src/commands.rs
```diff
@@ -0,0 +1,119 @@
...
+                .send_request::<ApplyWorkspaceEdit>(ApplyWorkspaceEditParams {
+                    label: Some(match edits.len() {
+                        1 => "Oxlint: 1 fix applied".into(),
+                        n => format!("Oxlint: {n} fixes applied"),
+                    }),
+                    edit: WorkspaceEdit {
+                        #[expect(clippy::disallowed_types)]
+                        changes: Some(std::collections::HashMap::from([(
+                            url.clone(),
+                            edits.clone(),
```
Sysix, line 109: I think both clones are not needed

crates/oxc_language_server/src/commands.rs
```diff
@@ -0,0 +1,119 @@
...
+        let url = Url::parse(&args.0.uri);
+        if let Err(e) = url {
+            error!("Invalid uri passed to {:?}: {e}", self.command_id());
+            return Err(Error::invalid_request());
+        }
+        let url = url.unwrap();
+
+        let mut edits = vec![];
+        if let Some(value) = backend.diagnostics_report_map.get(&url.to_string()) {
+            for report in value.iter() {
```
Sysix, line 93: thanks for looping before. we did missed that with the code action. 
No need for this PR: We can use the code action the execute this command or vice versa.

crates/oxc_language_server/Cargo.toml
```diff
@@ -18,7 +18,7 @@ workspace = true
 
 [[bin]]
 name = "oxc_language_server"
-test = false
+test = true
```
Sysix, line 21: 😍

Sysix (commented):
thanks you. there can  probably be more optimization done in this package.

crates/oxc_language_server/src/commands.rs
marekvospel: I'm not sure I understand.

Doing `fixAll` through code actions feels like using the wrong tool for the thing. - Instead of letting the server do the edits directly, there has to be multiple back and forth (client has to request all code actions, loop through all of them and make a separate request for each fix).

Moving this into a command reduces the back and forth between lsp and an editor and I think it's the right choice, as this makes the editor extensions simpler. (Why should each and every editor implement the fix all logic? In case we want this to be updated we'd have to go around 4 editors implementing a change in multiple repositories)

Sysix: > Doing fixAll through code actions feels like using the wrong tool for the thing

oxlint --fix is only modifying the code. And so does codeActions do for LSP:

> The code action request is sent from the client to the server to compute commands for a given text document and range. These commands are typically code fixes to either fix problems or to beautify/refactor code. The result of a [textDocument/codeAction](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) request is an array of [Command](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command) literals which are typically presented in the user interface. To ensure that a server is useful in many clients the commands specified in a code actions should be handled by the server and not by the client (see [workspace/executeCommand](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand) and ServerCapabilities.executeCommandProvider). If the client supports providing edits with a code action then that mode should be used.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction

The reason behind this is vscode supports `editor.codeActionsOnSave`. Maybe this has nothing to do with LSP Code Actions. But the author of it implemented it here because I requested it: https://github.com/oxc-project/oxc/pull/7672#issuecomment-2521221424 

> My Idea: also support editor.codeActionsOnSave in settings.json in the future.
There is source.fixAll.eslint, source.fixAll.ts and source.fixAll.biome already :)

Tell me what do you think about it. Maybe you have better knowledge about this feature then me.

marekvospel: Ah, I think I understand now. It's probably the difference between editors that's causing the misunderstanding.

Vscode queries for all code actions all the time (for the whole file), while neovim does a query after an action button is clicked.

Since vscode has all code actions for the current buffer pre-fetched, running each code action is a viable solution - espetially for workspace edits, which (at least I think) can be done in bulk and only then synchronizing the document? (I don't know much about how vscode works, but that would be my guess to make stuff effective) Thus `source.fixAll` makes sense. (Or it could be because it was a way for vscode to get this to work for a lot of language servers)

For neovim and vscode commands code actions don't make sense, as you have to make a new code action request and run apply for each of them. That's why I want to implement the LSP command - to get rid of this initial query AND to apply the fixes in one single bigger request - and thus remove the overhead of parsing the LSP request and task switching (async await)

I'm fairly sure using workspace commands is the most effective way to do this (only two requests are made + document synchronization) and should be preffered for non-editor-specific features.

Sysix: > Since vscode has all code actions for the current buffer pre-fetched, running each code action is a viable solution - espetially for workspace edits, 

I think vscode will request code actions on hover for the specific diagnostic. 

> I'm fairly sure using workspace commands is the most effective way to do this (only two requests are made + document synchronization) and should be preffered for non-editor-specific features.

I agree. Sorry when it was not clear from the beginning. 
I only wanted to point out that we have almost the same code already to support code actions (for vscode settings)

👍

marekvospel: Ah, my bad 😅 I maybe got a bit too defensive about my code. 😆 I was thinking about putting this into it's own utility function, but it seemed like way too short snippet for the hussle

Sysix (approved):

Issue #5068, opened 2024-08-22
Rework `oxc_prettier`


Boshen:
[`crates/oxc_prettier`](https://github.com/oxc-project/oxc/tree/main/crates/oxc_prettier) was my attempt at the [prettier bounty](https://console.algora.io/challenges/prettier). 

I thought I could finish it in time, except the fact that I rushed too quickly without looking at all the requirements ... It was too late when I got blocked by printing comments.

In order to rework `oxc_prettier`, we need to understand at least:

* How to debug prettier
* The [`Doc`](https://github.com/prettier/prettier/tree/main/src/document) IR https://github.com/prettier/prettier/blob/main/commands.md
* How to handle comments https://github.com/prettier/prettier/blob/main/src/language-js/comments/handle-comments.js
* How to handle parentheses https://github.com/prettier/prettier/blob/main/src/language-js/needs-parens.js

As for the infrastructure, we already have most of the code:

* https://github.com/oxc-project/oxc/tree/main/tasks/prettier_conformance shows our js compat is at js compatibility: 264/586 (45.05%). But we need to update prettier to latest make sure this conformance runner still works.
* All the code in https://github.com/oxc-project/oxc/tree/main/crates/oxc_prettier mimics and ports prettier https://github.com/prettier/prettier/tree/main/src/language-js

Feel free to remove everything and start from scratch, and copy over the format code https://github.com/oxc-project/oxc/tree/main/crates/oxc_prettier/src/format

parkin-lin:
I'm sorry to take up some of your time, but I feel confused about the following issues, so please allow me to ask you a few questions...
As we konw, [the next minor release of ESLint will deprecate core formatting rules. ](https://eslint.org/blog/2023/10/deprecating-formatting-rules) But this also means that the linter itself originally had code formatting capabilities, so why isn't Oxlint **directly** compatible with plugins like [ESLint Stylistic](https://github.com/eslint-stylistic)? Are there some corner cases where Oxlint doesn't work as well as Prettier, so Oxlint had to be compatible with ESLint and Prettier **separately**, or it's just a divide-and-conquer engineering trade-off？
I'm guessing one of the reasons for this is that Prettier can support different file formats (not only .js, but also .css and more...), but ESLint or Oxlint can do the same job via plugins.
Thanks.

DonIsaac:
We made an intentional decision to only support code correctness rules, not stylistic rules. @Boshen can explain reasoning more.

Boshen:
[...]

PR #8886, opened 2025-02-04
refactor(prettier): Verify AST nodes round 2

leaysgur:
Part of #5068 

- VariableDeclaration
- AwaitExpression
- SequenceExpression
- (ParenthesizedExpression)


crates/oxc_prettier/src/lib.rs
```diff
@@ -105,6 +105,10 @@ impl<'a> Prettier<'a> {
         (len > n).then(|| self.stack[len - n - 1])
     }
 
+    fn find_ancestor(&self, predicate: impl Fn(AstKind<'a>) -> bool) -> Option<AstKind<'a>> {
+        self.stack.iter().rev().skip(2).find(|&&kind| predicate(kind)).copied()
```
Boshen, line 109: This `skip(2)` looks out of place for `find_ancestor `, probably need a comment?

crates/oxc_prettier/src/format/js.rs
```diff
@@ -1299,10 +1322,42 @@ impl<'a> Format<'a> for Super {
...
+            {
+                // avoid printing `await (await` on one line
+                if let Some(ancestor) = p.find_ancestor(|kind| {
+                    matches!(kind, AstKind::BlockStatement(_) | AstKind::AwaitExpression(_))
+                }) {
+                    if (match ancestor {
+                        AstKind::BlockStatement(_) => true,
+                        AstKind::AwaitExpression(await_expr) => {
+                            // TODO:
+                            // && !startsWithNoLookaheadToken(
```
Boshen, line 1341: We don't have any tokens, although we have plenty of similar cases in the linter where we scan through the source text.

crates/oxc_prettier/src/lib.rs
leaysgur: Thanks! This should be `skip(1)` for the current node. I also added a comment for this.

crates/oxc_prettier/src/format/js.rs
leaysgur: This is helpful. 😍 
I have left a comment to address this in future iterations.

PR #8902, opened 2025-02-05
fix(oxc_laguage_server): code actions if report range overlap request

marekvospel:
Requesting code actions in neovim's normal mode doesn't return any code action, as the report range isn't fully contained in the requested range - which is only the current cursor position. Filtering based on overlap seems to be done by most relevant language servers (based on my tesing) - such as `rust-analyzer` or `ts_ls`. `eslint` filters code actions based on intercepted lines.

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -259,13 +259,10 @@ impl LanguageServer for Backend {
 
         let mut code_actions_vec: Vec<CodeActionOrCommand> = vec![];
         if let Some(value) = self.diagnostics_report_map.get(&uri.to_string()) {
-            let reports = value
-                .iter()
-                .filter(|r| {
-                    r.diagnostic.range == params.range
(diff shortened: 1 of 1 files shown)


crates/oxc_language_server/src/main.rs
```diff
@@ -259,13 +259,10 @@ impl LanguageServer for Backend {
 
         let mut code_actions_vec: Vec<CodeActionOrCommand> = vec![];
         if let Some(value) = self.diagnostics_report_map.get(&uri.to_string()) {
-            let reports = value
-                .iter()
-                .filter(|r| {
-                    r.diagnostic.range == params.range
-                        || range_includes(params.range, r.diagnostic.range)
-                })
-                .collect::<Vec<_>>();
```
marekvospel, line 268: Bonus performance improvement! :tada: No need for collect, we need `Iterator`

Sysix (approved):

PR #8915, opened 2025-02-05
feat(linter): support `env` and `globals` in `overrides` configuration

Sysix:
This PR supports `env` and `globals` keys in `overrides[x]`.

I added tests where I could. Snapshot CLI tests are working 🥳

PR #8920, opened 2025-02-06
fix(linter): correct the `is_reference_to_global_variable`

shulaoda:
Can you do a fix PR for the global reference problem (semantic + `globals`) to reduce the changes of this PR?

_Originally posted by @Boshen in https://github.com/oxc-project/oxc/pull/8845#pullrequestreview-2597750513_

### DIFF
--- a/crates/oxc_linter/src/config/globals.rs
+++ b/crates/oxc_linter/src/config/globals.rs
@@ -1,4 +1,4 @@
-use std::{borrow, fmt, hash};
+use std::{borrow, fmt, hash, ops::Deref};
 
 use rustc_hash::FxHashMap;
 use schemars::JsonSchema;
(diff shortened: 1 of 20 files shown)


crates/oxc_linter/src/config/globals.rs
```diff
@@ -40,6 +40,14 @@ impl OxlintGlobals {
     {
         self.0.get(name).is_some_and(|value| *value != GlobalValue::Off)
     }
+
+    pub fn is_disabled<Q>(&self, name: &Q) -> bool
```
shulaoda, line 44: Next, I will confirm if `is_enabled` is correct. @Sysix cc

Sysix: I am currently trying to implement the overrides for env and globals.
I think my expection was right:
https://github.com/oxc-project/oxc/pull/8915/commits/53aa4f766252f6582a5a9f10bbda90c74cd56cf1#diff-4803bbea196e86eff70b44ffa5fef5e38573bfeeaf90cb578cba00b20e2694acR351-R405

When no entry is found we should return true for `is_disabled`.

Sysix: After creating CLI Snapshot tests I can confirm there is a bug somewhere: https://github.com/oxc-project/oxc/pull/8915
Still I think te implemenation is wrong. `is_enabled` and `is_disabled` are sounding the opposite of the other. 
With your fix we are just creating another bug. I think we have somewhere else the bug and not here.

Sysix: Feel free to copy the changed snapshot tests and look if the expected result is coming out (currently broken because of this bug):
https://github.com/oxc-project/oxc/pull/8915/files
Maybe we have a bug in another place

shulaoda: I've decided not to add the potentially ambiguous `is_disabled` method for now and will use it through `Deref` instead.

Sysix (approved):

PR #9053, opened 2025-02-12
fix(linter): false positive in `jest/no-conditional-expect`

### LINKED ISSUE #9026: linter: jest/no-conditional-expect false positive for async expects

0.15.10

### What command did you run?

oxlint -c ../../packages/linting/oxlintrc.json

```jsonc
{
  "$schema": "../../node_modules/oxlint/configuration_schema.json",
  "plugins": ["import", "jest", "react-perf", "react", "unicorn", "typescript", "oxc"],
  "rules": {
    "jest/valid-title": ["error", { "ignoreTypeOfDescribeName": true, "ignoreTypeOfTestName": true }]
  }
}
```

### What happened?

The test looks as follows
```
    it("throws error", async () => {
     const fn = async () => throw new Error();

      const promise = fn();

      await expect(promise).rejects.toThrow();
    });
```
and the linter reports
```
  ⚠ eslint-plugin-jest(no-conditional-expect): Unexpected conditional expect
    ╭─[test.spec.ts:29:13]
 28 │ 
 29 │       await expect(promise).rejects.toThrow(MissingIterationsError);
    ·             ──────
 30 │     });
    ╰────
```
But this totally fine jest / vitest code to test promises.

The documentation states this is incorrect use
> it("throws an error", async () => {
  await foo().catch((error) => expect(error).toBeInstanceOf(error));
});

But it is correct use and is not included in the original docs of the rule: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md

shulaoda:
closes #9026 

- Improve the document
- Inline the function `run`

### DIFF
--- a/crates/oxc_linter/src/rules/jest/no_conditional_expect.rs
+++ b/crates/oxc_linter/src/rules/jest/no_conditional_expect.rs
@@ -31,11 +31,16 @@ declare_oxc_lint!(
     ///
     /// ### Why is this bad?
     ///
-    /// Jest only considers a test to have failed if it throws an error, meaning if calls to assertion functions like expect occur in conditional code such as a catch statement, tests can end up passing but not actually test anything.
-    /// Additionally, conditionals tend to make tests more brittle and complex, as they increase the amount of mental thinking needed to understand what is actually being tested.
+    /// Jest only considers a test to have failed if it throws an error, meaning if calls to
+    /// assertion functions like expect occur in conditional code such as a catch statement,
+    /// tests can end up passing but not actually test anything. Additionally, conditionals
+    /// tend to make tests more brittle and complex, as they increase the amount of mental
+    /// thinking needed to understand what is actually being tested.
     ///
-    /// ### Example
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/jest/no_conditional_expect.rs
```diff
@@ -137,7 +166,6 @@ fn check_parents<'a>(
         | AstKind::SwitchStatement(_)
         | AstKind::IfStatement(_)
         | AstKind::ConditionalExpression(_)
-        | AstKind::AwaitExpression(_)
```
shulaoda, line 140: The main change is here, `AwaitExpression` is not a conditional situation.

Sysix: Checked with the original jest rule. they do not have `AwaitExpression` too. `CatchClause` should be enough

Sysix (approved):

PR #9061, opened 2025-02-12
feat(transformer/statement-injector): add a series of `insert_*_before/after_current_statement` methods

Dunqing:
These methods allow inserting statements before or after the current statement without needing the statement address. And can reduce the address looking up logic like

https://github.com/oxc-project/oxc/blob/c35f52bac58ad618df5bcfce1301ded3c6da26ce/crates/oxc_transformer/src/es2022/class_properties/class.rs#L434-L440 

 This is also an alternative to https://github.com/oxc-project/backlog/issues/140.

### DIFF
--- a/crates/oxc_transformer/src/common/mod.rs
+++ b/crates/oxc_transformer/src/common/mod.rs
@@ -65,6 +65,16 @@ impl<'a> Traverse<'a> for Common<'a, '_> {
         self.statement_injector.exit_statements(stmts, ctx);
     }
 
+    #[inline]
+    fn enter_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.statement_injector.enter_statement(stmt, ctx);
+    }
+
+    #[inline]
+    fn exit_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.statement_injector.exit_statement(stmt, ctx);
+    }
+
     fn enter_function(&mut self, func: &mut Function<'a>, ctx: &mut TraverseCtx<'a>) {
         self.arrow_function_converter.enter_function(func, ctx);
     }
--- a/crates/oxc_transformer/src/common/statement_injector.rs
+++ b/crates/oxc_transformer/src/common/statement_injector.rs
@@ -12,7 +12,7 @@
 //! self.ctx.statement_injector.insert_many_after(address, statements);
 //! ```
 
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 
 use rustc_hash::FxHashMap;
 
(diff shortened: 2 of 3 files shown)


Dunqing:
* **#9061** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9061?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9061?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9061).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/common/mod.rs
```diff
@@ -65,6 +65,16 @@ impl<'a> Traverse<'a> for Common<'a, '_> {
...
 
+    #[inline]
+    fn enter_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.statement_injector.enter_statement(stmt, ctx);
+    }
+
+    #[inline]
+    fn exit_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.statement_injector.exit_statement(stmt, ctx);
+    }
```
overlookmotel, line 76: I don't think this can be correct. When exiting a statement, don't you need to restore the address of the *parent* statement? e.g.:

```js
const x = () => {
  let y;
};
```

After exiting the statement `let y`, the current statement is `const x = ...` again.

I guess would need a stack to achieve this.

overlookmotel (commented):
I'm afraid I'm not convinced this is a good idea overall.

On the positive side, as you say, it removes common repeated code - great.

But on the negative side, it's doing extra work on *every* statement (very hot path) to update `current_statement_address`. For the majority of statements, no transform ever needs to know their address, so that work is wasted.

This PR is currently showing -1% on some of the transformer benchmarks. I believe that to make it correct, it'll require a stack (see comment below) which is more expensive, so the perf impact will increase further.

I wonder if there's another way to get the nice API without the perf impact?

How about `insert_before_current_statement` etc taking `&TraverseCtx` and they search up the ancestry chain to find the current statement? Then the work would only be done when it needs to be done, rather than eagerly? Do you think something like that could work?

Dunqing:
> I wonder if there's another way to get the nice API without the perf impact?
> 
> How about `insert_before_current_statement` etc taking `&TraverseCtx` and they search up the ancestry chain to find the current statement? Then the work would only be done when it needs to be done, rather than eagerly? Do you think something like that could work?

This is feasible in most scenarios.  

I added these APIs to solve a problem when we want to do some transformation in `enter_statement` or `exit_statement` and we want to insert a statement in a deeper function. In this case, we have to pass the address all the way to where it is truly needed. Taking `&TraverseCtx` cannot solve the problem because we are already at the statement.

crates/oxc_transformer/src/common/mod.rs
Dunqing: Ah, This big problem I've overlooked. Thanks for pointing it out

Dunqing:
Converted to draft as it is not ideal, and I changed another way to solve #9062, please review that PR, which blocks Rolldown from advancing the legacy decorator feature.

overlookmotel:
@Dunqing #9062 is now merged, and from our discussion today, I think we agreed that these APIs don't solve our problems. So I'm closing. But feel free to re-open if I misunderstood.

PR #9074, opened 2025-02-12
feat(linter): support `import-x` plugin name

### LINKED ISSUE #8779: Oxlint: Support eslint-plugin-import-x rules
We have eslint-plugin-import-x which touts itself to have the similar set of rules but more performant than eslint-plugin-import

> https://github.com/un-ts/eslint-plugin-import-x

Would it make sense for this plugin to also support or map over those eslint rules to the corresponding rules in oxlint?

Sysix:
closes #8779

The `import-x` rules are the reimplementation of the `import` rules. No rules added or removed.
Some eslint-plugins/configs are handling the plugin the same way.

Here a snippet from a config which `next-auth` is using:
https://github.com/antfu/eslint-config/blob/998887cedb0b09c5b95d08602689a2c86e6662c7/src/factory.ts#L56-L67

Could not find more places where some mapping is happening. Please have a look because I did not test it.


camchenry (approved):

Sysix:
### Merge activity

* **Feb 18, 2:39 PM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Sysix:
added test, it is working with CLI 🥳 
Merging because vuejs/core is using it too:
https://github.com/oxc-project/oxlint-migrate/blob/a63d8159386e9da1706a885c6f25fe7cc47b14c8/integration_test/__snapshots__/vuejs-core.spec.ts.snap#L179-L187

PR #9088, opened 2025-02-13
refactor(minifier): inline more minification methods

Boshen:


Boshen:
Need to fix:

```
var flags = symbol.getFlags();
if ((flags & 2885600 ) === 0) {
    return undefined;
}
else if (B) {
    return 11;
}
else if (C) {
    return 12;
}
else if (D) {
    return 16;
}
else {
    return undefined;
}
```

PR #9089, opened 2025-02-13
refactor(linter): improve `jsx-a11y/heading-has-content`

shulaoda:
- Enable a repaired test case
- Simplify logical code

### DIFF
--- a/crates/oxc_linter/src/rules/jsx_a11y/heading_has_content.rs
+++ b/crates/oxc_linter/src/rules/jsx_a11y/heading_has_content.rs
@@ -48,7 +48,7 @@ declare_oxc_lint!(
     /// this could either confuse users or even prevent them
     /// from accessing information on the page's structure.
     ///
-    /// ### Example
+    /// ### Examples
     ///
     /// Examples of **incorrect** code for this rule:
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/jsx_a11y/heading_has_content.rs
```diff
@@ -19,21 +19,10 @@ fn heading_has_content_diagnostic(span: Span) -> OxcDiagnostic {
 }
 
 #[derive(Debug, Default, Clone)]
-pub struct HeadingHasContent(Box<HeadingHasContentConfig>);
```
Boshen, line 22: Boxing this struct ensures a small enum

```
thread 'size_asserts' panicked at crates/oxc_linter/src/lib.rs:56:5:
  left: 24
 right: 16
```

shulaoda: Got it. Thanks for pointing this out. I now understand why the configuration for some rules uses `Box` while others do not.

Sysix (approved):

camchenry (approved):

PR #9154, opened 2025-02-16
refactor(minifier): `Math.pow(a,b)` -> `a ** (+b)` instead of `(+a) ** (+b)`

Boshen:
It's currently minifying twice through:

`Math.pow(a, b)` -> `(+a) ** (+b)` -> `a ** (+b)`

Correct me if this is wrong, i.e. we need to keep `(+a)` in the fold pass.

### DIFF
--- a/crates/oxc_minifier/src/peephole/replace_known_methods.rs
+++ b/crates/oxc_minifier/src/peephole/replace_known_methods.rs
@@ -392,6 +392,7 @@ impl<'a> PeepholeOptimizations {
 
         Some(ctx.ast.expression_binary(
             span,
+            // see [`PeepholeOptimizations::is_binary_operator_that_does_number_conversion`] why it does not require `wrap_with_unary_plus_if_needed` here
             ctx.ast.move_expression(first_arg),
             BinaryOperator::Exponential,
             wrap_with_unary_plus_if_needed(second_arg),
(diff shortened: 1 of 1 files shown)


crates/oxc_minifier/src/peephole/replace_known_methods.rs
```diff
@@ -392,7 +392,7 @@ impl<'a> PeepholeOptimizations {
 
         Some(ctx.ast.expression_binary(
             span,
-            wrap_with_unary_plus_if_needed(first_arg),
+            ctx.ast.move_expression(first_arg),
```
sapphi-red, line 395:
```suggestion
            // see [`PeepholeOptimizations::is_binary_operator_that_does_number_conversion`] why it does not require `wrap_with_unary_plus_if_needed` here
            ctx.ast.move_expression(first_arg),
```

sapphi-red (approved):
Yeah, this is safe. I didn't do it this way initially because it was a bit difficult to verify if it's safe.

sapphi-red:
@magic-akari Like other minifiers OXC minifier assumes `.valueOf` to be sideeffect free, so that order change is fine.
https://github.com/oxc-project/oxc/tree/main/crates/oxc_minifier#:~:text=.toString()%2C,console.log(%27sideeffect%27)%20%7D%20%7D

PR #9160, opened 2025-02-16
feat(linter): add eslint/no-unneeded-ternary rule

cdrikd:
Implement the [eslint/no-unneeded-ternary](https://eslint.org/docs/latest/rules/no-unneeded-ternary) rule.

I have a question, where to put the options of the rules ? in the examples or in a specific chapter "Rule options" before examples ?

It's my first rust contribution, I just start to learn the language so don't hesitate to tell me enhancement.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_unneeded_ternary.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_unneeded_ternary.rs
@@ -1,0 +1,280 @@
+use crate::{context::LintContext, rule::Rule, AstNode};
+use oxc_allocator::Box;
+use oxc_ast::{ast::Expression, ast::IdentifierReference, AstKind};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+
+fn no_unneeded_ternary_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Unnecessary use of boolean literals in conditional expression")
+        .with_help("Remove this ternary operator")
+        .with_label(span)
+}
+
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/no_unneeded_ternary.rs
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        if let (Some(ze_test), Some(ze_cons)) =
+            (supply_identifier(&expr.test), supply_identifier(&expr.consequent))
+        {
+            if !self.default_assignment && ze_test.name == ze_cons.name {
+                ctx.diagnostic(no_unneeded_ternary_conditional_expression_diagnostic(expr.span));
+            }
+        }
+    }
```
camchenry, line 90: a few suggestions:

1. since these two cases are mutually exclusive, this second `if` block can be moved into an `else` case so that we don't do unnecessary work
2. the `supply_identifier` can be implemented using built-in methods: `get_inner_expression()` will remove parentheses and such, among other "transparent" expressions; `get_identifier_reference()` will return the ident, if one exists.

```suggestion
        } else if let (Some(test), Some(cons)) = (
            (&expr.test.get_inner_expression().get_identifier_reference()),
            (&expr.consequent.get_inner_expression().get_identifier_reference()),
        ) {
            if !self.default_assignment && test.name == cons.name {
                ctx.diagnostic(no_unneeded_ternary_conditional_expression_diagnostic(expr.span));
            }
        }
```

cdrikd: thank you for your comment

camchenry (approved):

PR #9162, opened 2025-02-16
refactor(traverse): simplify pointer manipulation and improve readability

### LINKED ISSUE #17: Use `ptr::addr_of` for offset calculations in `Traverse`'s `walk_*` functions
`Traverse`'s `walk_*` functions are currently full of code like this:

```rs
walk_expression(
    traverser,
    (node as *mut u8).add(ancestor::OFFSET_BINARY_EXPRESSION_LEFT) as *mut Expression,
    ctx,
);
```

`Ancestor` methods contain similar code:

```rs
pub fn left(&self) -> &Expression<'a> {
    unsafe {
        &*((self.0 as *const u8).add(OFFSET_BINARY_EXPRESSION_LEFT) as *const Expression<'a>)
    }
}
```

Could probably use `std::ptr::addr_of!` and `std::ptr::addr_of_mut!` for these instead:

```rs
use std::ptr::addr_of_mut;

walk_expression(
    traverser,
    addr_of_mut!((*node).left),
    ctx,
);
```

```rs
use std::ptr::addr_of;

pub fn left(&self) -> &Expression<'a> {
    let node = self.0;
    unsafe { &*addr_of!((*node).left) }
}
```

Much easier to read, and can get rid of all the `OFFSET_*` constants.

I hadn't clocked that you can use `addr_of!` starting with another pointer, and deref that pointer without actually deref-ing it.

https://doc.rust-lang.org/std/ptr/macro.addr_of.html

However, I'd like to implement more transformers using upwards seeking first and run it through Miri before making any changes. Also `addr_of!` may affect performance (either positively or negatively).

Dunqing:
close: https://github.com/oxc-project/backlog/issues/17

I don't know too much about raw pointers, the benchmark shows no difference both ways. I don't use [addr_of](https://doc.rust-lang.org/std/ptr/macro.addr_of.html) because the documentation has noted this API has been soft-deprecated


Dunqing:
* **#9162** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9162?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9162).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (commented):
I am pretty sure this change will cause undefined behavior.

There's a subtle difference between these 3:

(assume `ptr` is a `*const Program`)

1. `&*( (ptr as *const u8).add( offset_of!(Program, span) ) as *const Span )`
2. `&*addr_of!( (*ptr).span )`
3. `&(&*ptr).span`

(1) is what we have now, and (2) should be equivalent (if my understanding is correct).

They both convert the pointer to `Program` into a pointer to the `Span`, *without* creating a `&Program` reference in between.

But this PR uses something similar to (3), which is different from the first 2 because it creates an intermediate `&Program` reference (the `&*ptr` part deferences pointer `*const Program` to `&Program`).

And that's exactly what we need to avoid, because there already exists a live `&mut Program` - and holding a `&` reference and a `&mut` reference to the same thing is a violation of Rust's aliasing rules.

At least this is my understanding. And this is why I'm so cautious about touching `Traverse` - because it's really at the edges of my understanding. I know enough to be aware that there are subtleties, but I'm not sure if I understand them all. For example, the docs talk about a "value" and a "place" being different, but what *exactly* is that difference?

The good news is... now that you've removed `AstBuilder::copy`, we are finally in a position to deploy Miri and find out!

I propose to put this PR on hold for now. Next week I'd like to run the entire transformer conformance suite under Miri (which is slooooow - may take something like 12 - 48 hours) and we can find out whether it's sound in its current state. If it is, *then* we can look at modifying it, and then test again.

I'm going to mark this PR as "draft" in the meantime.

(sorry this is an incomplete explanation - I don't mean to invoke mysterious "dark magic" and avoid explaining myself - it's just really hard to put into words when I don't have a completely clear mental model myself)

PR #9176, opened 2025-02-17
chore(linter): improve the documentation of eslint/no-ternary

therewillbecode:
Relates to  [#6050](https://github.com/oxc-project/oxc/issues/6050)

Adds correctness and incorrectness examples following rule documentation template.


therewillbecode:
The examples I have added from the eslint documentation aren't actually included in the test cases for this rule. I can add them either in this PR or a separate one if that is an issue?

therewillbecode:
> > The example for correct code I have added from the eslint documentation isn't currently included in the test cases. I can add this test case in this PR or a separate one if that is an issue?
> 
> In this PR is fine.

Okay, all done :)

Sysix (approved):

PR #9183, opened 2025-02-17
chore(napi/transform): validate napi dts files

ArnaudBarre:
I was updating my vite plugin with the latest version of oxc-transform and I now have type issues with TransformResult.sourceMap being any because of theses line being removed: https://github.com/oxc-project/oxc/pull/8478/files?w=1#diff-02e5402041f98a1b1ae2abc7b3bafedb773bbb88578ab48e7faabb2b13d511d4L208-L218

I'm not sure how to fix the Rust code to make the source map type being generated again, but at least with the PR this will avoid regression like this


Dunqing:
It seems NAPI doesn't generate a correct DTS file in CI. Can @Brooooooklyn take a look at this?

PR #9191, opened 2025-02-18
feat(mangler): opt-out of direct eval

Boshen:
Context: 

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval
* https://esbuild.github.io/content-types/#direct-eval

### DIFF
--- a/crates/oxc_mangler/src/lib.rs
+++ b/crates/oxc_mangler/src/lib.rs
@@ -192,6 +192,11 @@ impl Mangler {
 
         assert!(scope_tree.has_child_ids(), "child_id needs to be generated");
 
+        // TODO: implement opt-out of direct-eval in a branch of scopes.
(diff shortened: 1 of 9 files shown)


crates/oxc_mangler/src/lib.rs
```diff
@@ -192,6 +192,11 @@ impl Mangler {
 
         assert!(scope_tree.has_child_ids(), "child_id needs to be generated");
 
+        // TODO: implement opt-out of direct-eval in a branch of scopes.
```
Boshen, line 195: cc @sapphi-red

tasks/coverage/snapshots/runtime.snap
Boshen, line 1: This contains previous changes which weren't update before.

I double checked test cases regarding eval, all seemed correct.

PR #9212, opened 2025-02-18
feat(linter): add unicorn/no-invalid-fetch-options rule

baevm:
This PR implements [unicorn/no-invalid-fetch-options](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v56.0.1/docs/rules/no-invalid-fetch-options.md) rule

issue: https://github.com/oxc-project/oxc/issues/684

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_invalid_fetch_options.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_invalid_fetch_options.rs
@@ -1,3 +1,6 @@
+use std::borrow::Cow;
+
+use cow_utils::CowUtils;
 use oxc_allocator::Box;
 use oxc_ast::{
     ast::{Argument, Expression, ObjectExpression, ObjectPropertyKind, PropertyKey},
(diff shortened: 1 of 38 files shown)


shulaoda (approved):

shulaoda:
LGTM! Thanks!

crates/oxc_linter/src/rules/unicorn/no_invalid_fetch_options.rs
```diff
@@ -0,0 +1,209 @@
...
+    ///
+    /// Examples of **correct** code for this rule:
+    /// ```javascript
+    /// const response = await fetch('/', {method: 'POST', body: 'foo=bar'});
+    ///
+    /// const request = new Request('/', {method: 'POST', body: 'foo=bar'});
+    /// ```
+    NoInvalidFetchOptions,
+    unicorn,
+    correctness,
```
shulaoda, line 54: Need to review if it should be set to `correctness`.

Sysix: because it will throw a `TypeError` I will say yes this is correct(ness).

Sysix (approved):

PR #9223, opened 2025-02-19
feat(linter): support `--report-unused-disable-directive`

### LINKED ISSUE #7544: linter: `--report-unused-disable-directives`
Hope to support [--report-unused-disable-directives](https://eslint.org/docs/latest/use/command-line-interface#--report-unused-disable-directives) to perform a strong check in CI

1zumii:
and `--report-unused-disable-directives-severity`

closed #7544

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -17,7 +17,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use serde_json::Value;
 
 use crate::{
-    cli::{CliRunResult, LintCommand, MiscOptions, Runner, WarningOptions},
+    cli::{CliRunResult, LintCommand, MiscOptions, ReportUnusedDirectives, Runner, WarningOptions},
     output_formatter::{LintCommandInfo, OutputFormatter},
     walk::{Extensions, Walk},
 };
@@ -57,6 +57,7 @@ impl Runner for LintRunner {
             enable_plugins,
             misc_options,
             disable_nested_config,
+            inline_config_options,
             ..
         } = self.options;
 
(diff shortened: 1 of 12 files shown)


Boshen (dismissed):
There's a test failure.

crates/oxc_linter/src/disable_directives.rs
```diff
@@ -604,3 +693,169 @@ semi*/
...
+
+        assert_eq!(unused.len(), 2);
+        assert_eq!(
+            unused.first().map(|(_, span)| *span),
+            comments.first().map(Comment::content_span)
+        );
+        assert_eq!(
+            unused.last().map(|(_, span)| *span),
+            comments.first().map(Comment::content_span)
+        );
```
camchenry, line 760: this goes for all of the other tests in this file where there is a rule name, we should check the rule name too, not just the comment span:
```suggestion
        assert_eq!(unused.len(), 2);
        let comment_span = comments.first().unwrap().content_span();
        let disable_debugger = unused.first().unwrap();
        let disable_console = unused.last().unwrap();
        assert_eq!(disable_debugger.span, comment_span);
        assert_eq!(disable_debugger.rule_name, "no-debugger");
        assert_eq!(disable_console.span, comment_span);
        assert_eq!(disable_console.rule_name, "no-console"
```

apps/oxlint/src/lint.rs
camchenry, line 1: There should be an end-to-end test at the bottom of this file which showcases the report unused directives feature. Add a fixture in `apps/oxlint/fixtures/report_unused_directives` and add a file which has a number of different kinds of lint disables, then reference that file in a test which adds a snapshot.

crates/oxc_linter/src/disable_directives.rs
```diff
@@ -604,3 +693,169 @@ semi*/
...
+    }
+
+    #[test]
+    fn unused_enable_all() {
+        let allocator = Allocator::default();
+        let semantic = process_source(
+            &allocator,
+            r"
+            console.log();
+            /* eslint-enable */
```
camchenry, line 721: Can you modify these tests to also use `oxlint-enable-*` and `oxlint-disable-*` as well, to ensure we have coverage there too?

camchenry (commented):
Thanks for the contribution! This looks good overall, I have a few suggestions on some ways we can improve the performance a little bit. My main suggestion is for some improvements to the testing, such as adding an integration test with an example file so we can easily see what the output of this is.

1zumii:
Thank you so much for your review. I've learned a lot about optimizing memory allocation in Rust from your feedback.

Sorry for the delay in getting back to this PR. Work has been keeping me busy lately.

I've made the changes you suggested. Could you please review it again? Thanks a lot!

camchenry (approved):
There is some future room for improvement in our diagnostic messages and probably performance, but this is a great contribution, thank you!

donaldpipowitch:
Is it by design or a bug that unknown ESLint rules (e.g. custom rules) are reported as unused directives instead of being ignored?

PR #9231, opened 2025-02-19
feat(linter): promote `eslint/no-eval` to `correctness`

### LINKED ISSUE #9188: linte: promote `no-eval` to `correctness`
Direct eval should not be used according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) and [esbuild documentation](https://esbuild.github.io/content-types/#direct-eval).

This rule is currently hidden under `restriction`. I suggest we move it to `correctness` (turned on by default).

https://github.com/oxc-project/oxc/blob/e8fa09174fdcd04d59b6831197a9d9c1d2d7e1b9/crates/oxc_linter/src/rules/eslint/no_eval.rs#L48

---

This rule is incomplete with lots of commented out test cases, so PR is welcomed to fix these tests.

https://github.com/oxc-project/oxc/blob/e8fa09174fdcd04d59b6831197a9d9c1d2d7e1b9/crates/oxc_linter/src/rules/eslint/no_eval.rs#L129-L238

shulaoda:
closes #9188

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_eval.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_eval.rs
@@ -1,51 +1,100 @@
-// Ported from https://github.com/eslint/eslint/tree/v9.9.1/lib/rules/no-eval.js
-use oxc_ast::{ast::Expression, AstKind};
+use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_semantic::AstNode;
-use oxc_span::Span;
+use oxc_span::{GetSpan, Span};
 
-use crate::{context::LintContext, rule::Rule};
+use crate::{
+    ast_util::{self},
+    config::GlobalValue,
+    context::LintContext,
+    rule::Rule,
+    AstNode,
+};
 
 fn no_eval_diagnostic(span: Span) -> OxcDiagnostic {
-    OxcDiagnostic::warn("eval can be harmful.").with_label(span)
+    OxcDiagnostic::warn("The `eval` can be harmful.").with_label(span)
 }
 
 #[derive(Debug, Default, Clone)]
 pub struct NoEval {
-    /// Whether to allow references to the `eval` function as long as they are
-    /// not called. For example, the following code is valid if this property is
-    /// true:
-    ///
-    /// ```javascript
-    /// const foo = eval;
-    /// foo();
-    ///
-    /// (function(exec) {
-    ///     exec();
-    /// })(eval);
-    /// ```
-    ///
-    /// The default value is `false`.
-    pub allow_indirect: bool,
+    allow_indirect: bool,
 }
 
 declare_oxc_lint!(
     /// ### What it does
-    /// Disallows referencing the 'eval' function.
+    ///
+    /// Disallows referencing the `eval` function. This rule is aimed at preventing
+    /// potentially dangerous, unnecessary, and slow code by disallowing the use of
+    /// the `eval()` function.
     ///
     /// ### Why is this bad?
-    /// Calling 'eval' is not supported in some secure contexts and can lead to
-    /// vulnerabilities.
     ///
-    /// ### Example
-    /// ```javascript
-    /// const someString = "console.log('pwned')"
-    /// eval(someString);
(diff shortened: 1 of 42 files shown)


crates/oxc_linter/src/rules/eslint/no_eval.rs
```diff
@@ -58,184 +107,284 @@ impl Rule for NoEval {
...
+        ("window.setTimeout('foo')", None, env_with_browser(), None),
+        ("window.setInterval('foo')", None, env_with_browser(), None),
+        ("window.eval('foo')", None, None, None),
+        ("window.eval('foo')", None, None, Some(PathBuf::from("foo.cjs"))),
+        ("window.noeval('foo')", None, env_with_browser(), None),
+        ("function foo() { var eval = 'foo'; window[eval]('foo') }", None, env_with_browser(), None),
+        ("global.eval('foo')", None, None, None),
+        ("global.noeval('foo')", None, None, Some(PathBuf::from("foo.cjs"))),
+        ("function foo() { var eval = 'foo'; global[eval]('foo') }", None, env_with_node(), Some(PathBuf::from("foo.cjs"))),
+        // ("globalThis.eval('foo')", None, None, None), // { "ecmaVersion": 2017 }, globalThis is not supported
```
shulaoda, line 257: Since we do not support the `ecmaVersion` option, it will be ignored.

crates/oxc_linter/src/rules/eslint/no_eval.rs
```diff
@@ -58,184 +107,284 @@ impl Rule for NoEval {
...
-        // ("(window?.window).eval('foo')", None),
-        // ("class C { [this.eval('foo')] }", None),
-        // ("'use strict'; class C { [this.eval('foo')] }", None),
-        // ("class A { static {} [this.eval()]; }", None),
-        // ("function foo() { 'use strict'; this.eval(); }", None),
+        //     "function foo() { 'use strict'; this.eval(); }",
+        //     None,
+        //     None,
+        //     Some(PathBuf::from("foo.cjs")),
+        // ), // { "ecmaVersion": 3 }, in es3, "use strict" directives do not apply
```
shulaoda, line 368: Since we do not support the `ecmaVersion` option, it will be ignored.

crates/oxc_linter/fixtures/import/foo.cjs
Boshen, line 1: Unintentional file?

crates/oxc_linter/src/rules/eslint/no_eval.rs
```diff
@@ -1,51 +1,100 @@
...
+    ast_util::{self},
+    config::GlobalValue,
+    context::LintContext,
+    rule::Rule,
+    AstNode,
+};
 
 fn no_eval_diagnostic(span: Span) -> OxcDiagnostic {
-    OxcDiagnostic::warn("eval can be harmful.").with_label(span)
+    OxcDiagnostic::warn("The `eval` can be harmful.").with_label(span)
```
Boshen, line 15: https://github.com/eslint/eslint/blob/80b048535e1d951692e838fe502fb0edb72c837f/lib/rules/no-eval.js#L66

`eval can be harmful.` was correct.

Boshen (changes requested):

crates/oxc_linter/fixtures/import/foo.cjs
shulaoda: This is necessary, we need files with a `.cjs` extension to set the `SourceType` to `script`.

https://github.com/oxc-project/oxc/pull/9231/files#diff-852124295944eec39f4b94fbef7050e6f77e09a73a9d52dc423d616a30ceb1faR314

shulaoda: I didn't find any other files that met the criteria, so I added it.

crates/oxc_linter/src/rules/eslint/no_eval.rs
shulaoda: Okay, revert it.

shulaoda: BTW, many of our messages are not exactly the same as those in `eslint` or other plugins. I’ve modified it solely to improve the `diagnostic`.

Boshen (approved):

PR #9242, opened 2025-02-19
chore(linter): Improve the documentation of eslint/default-case

therewillbecode:
Add more documentation aligned with doc template for eslint rule default-case.

Relates to [#6050](https://github.com/oxc-project/oxc/issues/6050).

Inline with @shulaoda comment on this PR I have made the documented option type consistent in the other two rules I recently updated.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/default_case.rs
+++ b/crates/oxc_linter/src/rules/eslint/default_case.rs
@@ -48,41 +48,41 @@ declare_oxc_lint!(
     /// Examples of **incorrect** code for this rule:
     /// ```javascript
     /// switch (foo) {
-    ///     case 1:
-    ///         break;
+    ///   case 1:
+    ///     break;
     /// }
     /// ```
     ///
     /// Examples of **correct** code for this rule:
     /// ```javascript
     /// switch (a) {
-    ///     case 1:
-    ///         /* code */
-    ///         break;
+    ///   case 1:
+    ///     /* code */
+    ///     break;
     ///
-    ///     default:
-    ///         /* code */
-    ///         break;
+    ///   default:
+    ///     /* code */
+    ///     break;
     /// }
     /// ```
     ///
     /// ```javascript
     /// switch (a) {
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/default_case.rs
```diff
@@ -35,14 +35,76 @@ declare_oxc_lint!(
...
+    /// ```
+    ///
+    ///  ### Options
+    ///
+    ///  ### commentPattern
+    ///
+    ///  This option is for specifying an alternative regular expression which
+    ///  will override the default `/^no default$/i` comment test pattern.
+    ///
+    ///  `{ "commentPattern": string }`
```
shulaoda, line 96: I suggest placing the option type first, followed by the description of the option.

therewillbecode: Good idea, thanks!

crates/oxc_linter/src/rules/eslint/default_case.rs
```diff
@@ -35,14 +35,76 @@ declare_oxc_lint!(
...
+    /// ### Examples
+    ///
+    /// Examples of **incorrect** code for this rule:
     /// ```javascript
     /// switch (foo) {
-    ///   case 1:
-    ///     break;
+    ///     case 1:
+    ///         break;
+    /// }
```
shulaoda, line 53: I recommend using two spaces for indentation. Since each rule plugin has different indentation conventions for examples, I’d like to standardize them.

therewillbecode: Also should we add the indentation standardisation to one of the goals in this [issue](https://github.com/oxc-project/oxc/issues/6050)?

therewillbecode: Thanks for updating the [issue](https://github.com/oxc-project/oxc/issues/6050) guidelines btw!

PR #9244, opened 2025-02-19
feat(linter): add eslint/id-denylist

therewillbecode:
wip. Commented out these kinds of tests
```
// ("/* global myGlobal: readonly */ myGlobal = 5;", Some(serde_json::json!(["myGlobal"]))),
``` 
As not sure what do do with them. Seems as though all  the other rules comment out tests related to inline global conf comments.


camchenry:
You can leave the inline configuration tests commented out. We don't support inline configuration, so we don't need to test it.

PR #9247, opened 2025-02-19
perf(transformer/jsx): add `#[inline]` to `exit_expression`

overlookmotel:


overlookmotel:
Well that didn't work!

PR #9318, opened 2025-02-24
refactor(data_structures): stack types const assert `T` is not zero-size type

overlookmotel:
`Stack`, `NonEmptyStack` and `SparseStack` do not support zero-sized types, for simplicity of implementation (a `Stack<()>` would be pretty pointless anyway).

Promote panics if attempt to construct a `Stack<()>` from run time to build time, by putting the size assertions in `const {}` blocks. These assertions should be const-folded out by compiler anyway, but the `const {}` blocks make sure.


Boshen:
### Merge activity

* **Feb 24, 4:49 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #9329, opened 2025-02-24
feat(allocator): add `String::from_strs_array_in`

Dunqing:
This method is similar to in [Vec::from_array_in](https://github.com/oxc-project/oxc/blob/5acc6ec3e9b51b3c6649409759e5039b6bdce8eb/crates/oxc_allocator/src/vec.rs#L140-L167), this aims to solve the problem that we want to construct an `ArenaString` with the given `&str`s where from different variables. 

For example:

https://github.com/oxc-project/oxc/blob/36a90a61e85bd132040dc9a562efc12e5ae59673/crates/oxc_transformer/src/common/helper_loader.rs#L309-L318

This can refactored to

```diff
- let mut source = ArenaString::with_capacity_in(len, ctx.ast.allocator); 
- source.push_str(&self.module_name); 
- source.push_str("/helpers/"); 
- source.push_str(helper_name);
+ ArenaString::from_array_in([&self.module_name, "/helpers/", helper_name], ctx.ast.allocator);
```

### DIFF
--- a/crates/oxc_allocator/src/string.rs
+++ b/crates/oxc_allocator/src/string.rs
@@ -10,6 +10,7 @@ use std::{
     hash::{Hash, Hasher},
     mem::ManuallyDrop,
     ops::{Deref, DerefMut},
+    ptr,
 };
 
 use bumpalo::collections::String as BumpaloString;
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#9329** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9329?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9329?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9329).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (dismissed):
The idea for this API is good. And we can also make it more performant with judicious use of some unsafe code! (to remove bounds checks)

However, the implementation I'm pretty sure is not correct. Add a test, and you'll see what I mean...

Dunqing:
> The idea for this API is good. And we can also make it more performant with judicious use of some unsafe code! (to remove bounds checks)
> 
> However, the implementation I'm pretty sure is not correct. Add a test, and you'll see what I mean...

Ah, I see the problem, I reimplemented it. But I am not sure this if it has become more performant.

overlookmotel (commented):
Nice! This is, in my opinion, exactly how unsafe code should be used - enclosed in a small function where you can understand the logic and check it's sound, and presenting a safe interface to user, so they don't have to worry about checking that logic again every time they use it.

In addition to my suggested optimization below, I'd suggest:

Add more `// SAFETY` comments. Usually I go through the docs for the unsafe method that I'm calling and make a comment that "ticks off" each safety requirement one by one - and explains for each one how the code guarantees that requirement is satisfied.

This is boring and lengthy, but has 2 advantages: (a) it's a rigorous approach which makes it harder to miss anything and (b) it makes it easier for someone else to review the correctness of the logic.

For example, if one of the string is `""`, I'm wondering if it is legal to call `ptr::copy_nonoverlapping` with `len = 0`? Your comments don't reveal whether (a) you checked the docs and it *is* legal or (b) we don't know if it is or not. That kind of ambiguity isn't ideal for unsafe code.

Secondly, unsafe code should ideally have unit tests, and they should include edge cases e.g. empty array `[&str; 0]` or empty strings `["", "x", ""]`.

overlookmotel:
On other thing... What if the total length of the strings together exceeds `usize::MAX`? Does `sum()` panic in case of arithmetic overflow in release mode? If it doesn't, then we have opportunity for UB.

That's not really possible on 64-bit machines, but it could be on WASM (32-bit).

```rs
let (s1: &str, s2: &str) = get_strings_somehow();
// Lets say these 2 strings are both huge
assert!(s1.len() == 2 * 1024 * 1024 * 1024); // 2 GiB
assert!(s2.len() == 2 * 1024 * 1024 * 1024); // 2 GiB
let len = s1.len() + s2.len(); // Panics in debug mode, but wraps around to 0 in release mode on 32-bit machine
assert!(len == 0);
// `len` is 0 so this does not allocate
let vec = Vec::with_capacity_in(len, allocator);
// Later... boom! Overwrite 2 GiB of arbitrary memory
unsafe { ptr::copy_nonoverlapping(s1.as_ptr(), dst, s1.len());
```

overlookmotel:
One other thing: Could we call this method `String::from_strs_array_in`? Otherwise the question is "array of *what*?". We might in future want to add e.g. `String::from_bytes_array_in`.

Dunqing:
> One other thing: Could we call this method `String::from_strs_array_in`? Otherwise the question is "array of _what_?". We might in future want to add e.g. `String::from_bytes_array_in`.

I like the new method name!

Dunqing:
Thank you for helping me make this API work and safe! I really appreciate it. @overlookmotel

PR #9344, opened 2025-02-24
feat(linter): add `eslint/dot-notation`

ammubhave:
Relates to #479

Rule details: https://eslint.org/docs/latest/rules/dot-notation

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/dot_notation.rs
+++ b/crates/oxc_linter/src/rules/eslint/dot_notation.rs
@@ -7,14 +7,14 @@ use oxc_span::Span;
 use regex::{Regex, RegexBuilder};
 
 fn dot_notation_use_dot_diagnostic(span: Span, incorrect: &str, correct: &str) -> OxcDiagnostic {
-    OxcDiagnostic::warn(format!("[{}] is better written in dot notation", incorrect))
-        .with_help(format!("Write it as .{}", correct))
+    OxcDiagnostic::warn(format!("[{incorrect}] is better written in dot notation"))
+        .with_help(format!("Write it as .{correct}"))
         .with_label(span)
 }
 
 fn dot_notation_use_brackets_diagnostic(span: Span, key: &str) -> OxcDiagnostic {
-    OxcDiagnostic::warn(format!(".{} is a syntax error", key))
-        .with_help(format!("Write it using bracket notation [\"{}\"]", key))
+    OxcDiagnostic::warn(format!(".{key} is a syntax error"))
+        .with_help(format!("Write it using bracket notation [\"{key}\"]"))
         .with_label(span)
 }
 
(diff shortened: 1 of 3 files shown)


Sysix:
I think there are some possible improvements. One is: instead of `VALID_IDENTIFER: Regex` there can be a simple char parser.
@camchenry @shulaoda  you guys have more knowledge about performance in rust

shulaoda:
> I think there are some possible improvements. One is: instead of `VALID_IDENTIFER: Regex` there can be a simple char parser. 

Perhaps we can take inspiration from https://github.com/rolldown/rolldown/pull/3677. I’ll review it later.

crates/oxc_linter/src/rules/eslint/dot_notation.rs
```diff
@@ -0,0 +1,343 @@
...
+    "typeof",
+    "var",
+    "void",
+    "volatile",
+    "while",
+    "with",
+];
+lazy_static! {
+    static ref VALID_IDENTIFER: Regex = Regex::new(r"^[a-zA-Z_$][a-zA-Z0-9_$]*$").unwrap();
+}
```
camchenry, line 150: We have a strong preference to avoid using `Regex` unless it's necessary (for example, the `allowPattern` needs to support that syntax), as it is relatively slow (https://oxc.rs/docs/contribute/rules.html#performance). We can rewrite this using a mini-parser instead, here is my suggestion:

```suggestion
fn is_valid_identifier(s: &str) -> bool {
    s.as_bytes().first().map_or(false, |&c| (c.is_ascii_alphabetic() || c == b'_' || c == b'$'))
        && s.bytes().skip(1).all(|c| c.is_ascii_alphanumeric() || c == b'_' || c == b'$')
}
```

crates/oxc_linter/src/rules/eslint/dot_notation.rs
```diff
@@ -0,0 +1,343 @@
...
+            if let Some(val) = config.get("allowKeywords").and_then(serde_json::Value::as_bool) {
+                cfg.allow_keywords = val;
+            }
+        }
+
+        Self(Box::new(cfg))
+    }
+
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::MemberExpression(node) = &node.kind() {
```
camchenry, line 169: Use an early return here instead, to remove a level of indentation:

```suggestion
        let AstKind::MemberExpression(node) = &node.kind() else { return };
```

crates/oxc_linter/src/rules/eslint/dot_notation.rs
```diff
@@ -0,0 +1,343 @@
...
+                        Expression::TemplateLiteral(x) => {
+                            if x.expressions.len() == 0 && x.quasis.len() == 1 {
+                                let s = x.quasis[0].value.cooked.unwrap().into_string();
+                                Some((s.clone(), s))
+                            } else {
+                                None
+                            }
+                        }
+                        _ => None,
+                    } {
```
camchenry, line 192: 1. we have a method for getting the expression inside called `get_inner_expression`
2. there is no need to clone the strings inside, I don't think we have a helper that currently does what this is doing, but it can at least be simplified:

```suggestion
                    let value = match expr.expression.get_inner_expression() {
                    Expression::NullLiteral(x) => Some(x.to_string()),
                    Expression::BooleanLiteral(x) => Some(x.to_string()),
                    Expression::StringLiteral(x) => Some(x.to_string()),
                    Expression::TemplateLiteral(x) => {
                        if x.expressions.len() == 0 && x.quasis.len() == 1 {
                            Some(x.quasis[0].value.cooked.unwrap().to_string())
                        } else {
                            None
                        }
                    }
                    _ => None,
                };
                let Some(value) = value else {
                    return;
                };
```

crates/oxc_linter/src/rules/eslint/dot_notation.rs
```diff
@@ -0,0 +1,343 @@
...
+                        if VALID_IDENTIFER.is_match(&after)
+                            && (self.0.allow_keywords || !KEYWORDS.contains(&after.as_str()))
+                            && !(self.0.allow_pattern.is_some()
+                                && self.0.allow_pattern.as_ref().unwrap().is_match(&after))
+                        {
+                            ctx.diagnostic(dot_notation_use_dot_diagnostic(
+                                expr.span, &before, &after,
+                            ))
+                        }
+                    }
```
camchenry, line 202: no need to unwrap the regex pattern, you can use `if-let` blocks. I would also recommend flattening this out a bit with early returns to make it a little easier to follow. taking other suggestions into account:

```suggestion
                if !is_valid_identifier(&value) {
                    return;
                }
                if !self.0.allow_keywords && KEYWORDS.contains(&value.as_str()) {
                    return;
                }

                if let Some(pattern) = &self.0.allow_pattern {
                    if pattern.is_match(&value) {
                        return;
                    }
                }

                ctx.diagnostic(dot_notation_use_dot_diagnostic(expr.span, &value, &value))
```

crates/oxc_linter/src/rules/eslint/dot_notation.rs
```diff
@@ -0,0 +1,343 @@
...
+    DotNotation,
+    eslint,
+    style,
+    // Fixes are possible
+    // See <https://github.com/eslint/eslint/blob/main/lib/rules/dot-notation.js>
+    pending
+);
+
+/// List of ES3 keywords.
+const KEYWORDS: &[&str] = &[
```
camchenry, line 87: given the size of this, I'd recommend using hash set instead. here's an example: https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/valid_typeof.rs#L141

camchenry (commented):
Thanks for the contribution! I've made a few suggestions to help this rule better fit the style of our codebase as well as perform better.

crates/oxc_linter/src/rules/eslint/dot_notation.rs
ammubhave: good point, I was curious so I ran benchmarks for linear search vs using `match` and using `phf_set`.
```
test tests::bench_linear            ... bench:          18.53 ns/iter (+/- 2.30)
test tests::bench_match             ... bench:           2.01 ns/iter (+/- 0.13)
test tests::bench_set               ... bench:          10.67 ns/iter (+/- 0.89)
```

I changed the implementation to a `is_keyword` function that uses `match`.


btw, for the valid_typeof link that you mentioned above, I ran the same benchmark with that list of valid types. looks like the set lookup takes significantly longer in that case presumably because the number of elements is much smaller.
```
test tests::bench_valid_type_linear ... bench:           0.29 ns/iter (+/- 0.03)
test tests::bench_valid_type_match  ... bench:           0.28 ns/iter (+/- 0.03)
test tests::bench_valid_type_set    ... bench:          10.81 ns/iter (+/- 1.01)
```

ref: https://gist.github.com/ammubhave/ca9f531f0e3760616f664f4230da7983

Boshen (changes requested):
We have similar code in the minifier where you can copy from https://github.com/oxc-project/oxc/blob/main/crates/oxc_minifier/src/peephole/convert_to_dotted_properties.rs

Boshen:
@camc314 would be great if you can finish this, by copying over what the minifier does.

connorshea:
Thank you for the contribution! We've implemented this rule as a type-aware rule through tsgolint, based on the typescript-eslint version. As such, we're opting not to implement the ESLint version of this rule. Apologies for the trouble. 🙇  https://github.com/oxc-project/tsgolint/pull/658

PR #9349, opened 2025-02-25
perf(ast/estree): faster UTF-8 to UTF-16 span conversion

overlookmotel:
Speed up UTF-8 to UTF-16 `Span` conversion by processing span offsets in ascending order.

This involves, for each node, processing `span.start`, then visiting all children and processing their `Span`s, and then finally processing `span.end`.

When the AST has come direct from the parser, this means all offsets are processed 100% in ascending order (and the visitor used here has manually-written visitation methods for the few types where that's not always the case otherwise e.g. `export {x}`).

If the AST *has* been modified, code may have moved around, so ascending order won't necessarily be preserved, but still it *mostly* will.

Optimize for this visitation order. Instead of doing a binary search through the whole UTF8-UTF16 translation table on every single span offset, assume the current offset is in same region as the last. Only resort to binary search as a de-opt when that's not the case. Most JS/TS files don't contain many non-ASCII characters, so these cases will be rare.

### DIFF
--- a/.github/.generated_ast_watch_list.yml
+++ b/.github/.generated_ast_watch_list.yml
@@ -18,6 +18,7 @@ src:
   - 'crates/oxc_ast/src/generated/derive_get_span.rs'
   - 'crates/oxc_ast/src/generated/derive_get_span_mut.rs'
   - 'crates/oxc_ast/src/generated/get_id.rs'
+  - 'crates/oxc_ast/src/generated/utf8_to_utf16_converter.rs'
   - 'crates/oxc_ast/src/generated/visit.rs'
   - 'crates/oxc_ast/src/generated/visit_mut.rs'
   - 'crates/oxc_ast/src/serialize.rs'
--- a/crates/oxc_ast/src/generated/utf8_to_utf16_converter.rs
+++ b/crates/oxc_ast/src/generated/utf8_to_utf16_converter.rs
@@ -1,0 +1,1198 @@
+// Auto-generated code, DO NOT EDIT DIRECTLY!
+// To edit this generated file you have to edit `tasks/ast_tools/src/generators/utf8_to_utf16.rs`
+
+use oxc_span::GetSpan;
+use oxc_syntax::scope::ScopeFlags;
+
+use crate::{
+    ast::*,
+    utf8_to_utf16::Utf8ToUtf16Converter,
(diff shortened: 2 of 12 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9349?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9349** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9349?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9348** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9348?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9342** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9342?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9349).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Boshen:
The latest https://crates.io/crates/oxc_ast/0.52.0 crate is 184kb, I think it's time to split all the visitors to speed up compilation?

overlookmotel:
> The latest https://crates.io/crates/oxc_ast/0.52.0 crate is 184kb, I think it's time to split all the visitors to speed up compilation?

Yes. The visitor in this PR is not compiled unless `serialize` feature is enabled, though.

I also wonder if we should move `Visit` and `VisitMut` traits into a new `oxc_visit` crate. Some use cases only need the parser and AST itself, so having to compile these huge files is unhelpful in those cases.

Unfortunately `AstBuilder` I think needs to stay in `oxc_ast` crate, because we want to eventually force all node creation to go through `AstBuilder`. But we could put it behind a feature. I don't think linter uses it, for example.

Issue #8070, opened 2024-12-23
How to get child nodes 


ityuany:
Suppose I want to get all the child nodes of `JSXClosingElement` , based on `oxc_semantic` what should I do? 

Thank you for your help

<img width="861" alt="image" src="https://github.com/user-attachments/assets/321c6a45-3134-4772-a02d-fc9884b05181" />

Boshen:
You can't. What's your usecase?

ityuany:
When I iterate through all nodes, I want to determine whether a node is a sub-node belonging to JSXClosingElement.
The actual scenario is that I need to determine the number of times a member imported from'react 'is used in the following code.

Since'React 'is the'default import', what I need to determine is the number of times all properties of'React 'are used, which in the following code is the number of times'Fragment' is used.
I followed the steps below.
- Get the ImportDefaultSpecifier of'react ', get'React'
- Get all the references of'React ', and the resulting array contains the'React' in '</React. Fragment >'.
At this point I need to determine whether the'React 'in' </React. Fragment > 'belongs to the sub-node of JSXClosingElement. Otherwise it will be counted repeatedly with' < React. Fragment > '

```jsx
import React from 'react';
function Component() {
    return <React.Fragment>Hello</React.Fragment>;
}
```

ityuany:
我现在的做法是写了一个很糟糕的，不优雅的工具方法向上寻找

```rust
pub fn is_in<F>(
    &self,
    node: &AstNode,
    max_depth: usize,
    predicate: F,
  ) -> Option<&AstNode>
  where
    F: Fn(&AstKind) -> bool,
  {
    let mut depth = 0;
    let mut current_node_id = node.id();
    while let Some(pn) = self.semantic.nodes().parent_node(current_node_id) {
      if depth >= max_depth {
        return None;
      }

      if predicate(&pn.kind()) {
        return Some(pn);
      }

      current_node_id = pn.id();
      depth += 1;
    }
    None
  }
```

我是向下面这样使用的

```rust
 let is_in_closing = self
    .semantic_handler
    .is_in(reference_node, 6, |kind| {
      matches!(kind, AstKind::JSXClosingElement(_))
    })
    .is_some();

  if is_in_closing {
    continue;
  }
```

我认为这个做法并不好， 出于性能考虑， 我不得不加了一个 max_depth 的判定

ityuany:
I think you've come up with a great idea, but the problem is I also need to differentiate

[...]

Issue #5493, opened 2024-09-05
Wish to introduce  ast node location


ityuany:
I really like the range of crates provided by oxc and have been using them a lot in my projects, but I'm experiencing some problems with them.

I can't use ast node's span information to accurately restore the source location via oxc_sourcemap

So I hope

As below, there are LOC fields, which are more friendly to using sourcemap to directly reflect the source code

```json
{
          "type": "ImportDefaultSpecifier",
          "start": 7,
          "end": 12,
          "loc": {
            "start": {
              "line": 1,
              "column": 7
            },
            "end": {
              "line": 1,
              "column": 12
            }
          },
          "range": [
            7,
            12
          ],
          "local": {
            "type": "Identifier",
            "start": 7,
            "end": 12,
            "loc": {
              "start": {
                "line": 1,
                "column": 7
              },
              "end": {
                "line": 1,
                "column": 12
              }
            },
            "range": [
              7,
              12
            ],
            "name": "React"
          }
        }
```

Looking forward to your reply

IWANABETHATGUY:
I assume you mean adding `location` information in the playground?

IWANABETHATGUY:
Each `Location` will increase `span` size by 16 bytes, which could greatly impact performance.

ityuany:
Great, thank you for providing the code snippet. I will come to learn your code later.

IWANABETHATGUY:
Closed since oxc is not gonna adding this info in the core

overlookmotel:
Actually we *may* add other forms of span (e.g. UTF-16 spans, line/column spans) in future, but it's likely a way off. #959

IWANABETHATGUY:
> Actually we _may_ add other forms of span (e.g. UTF-16 spans, line/column spans) in future. But it's likely a way off. #959

Yeah, but the issue owner want `Location`, so he needs to convert anyway.

PR #9357, opened 2025-02-25
refactor(linter): Change private modules to public modules to improve…

ityuany:
… accessibility

- Change the visibility of modules such as `import`, `eslint`, `typescript`, `jest`, `react`, `react_perf`, `unicorn`, `jsx_a11y`, `oxc`, `nextjs`, `jsdoc`, `promise`, `vitest`, `node`, etc. from `mod` to `pub mod`

- Change `use` to `pub use` in `lib.rs` to export module interfaces

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -45,7 +45,7 @@ pub use crate::{
     rule::{RuleCategory, RuleFixMeta, RuleMeta, RuleWithSeverity},
     service::{LintService, LintServiceOptions},
 };
-use crate::{
+pub use crate::{
     config::{LintConfig, OxlintEnv, OxlintGlobals, OxlintSettings, ResolvedLinterState},
     context::ContextHost,
     fixer::{Fixer, Message},
--- a/crates/oxc_linter/src/rules.rs
+++ b/crates/oxc_linter/src/rules.rs
@@ -6,7 +6,7 @@
 //! are inside a proc macro.
 
 /// <https://github.com/import-js/eslint-plugin-import>
-mod import {
+pub mod import {
     // pub mod no_deprecated;
     // pub mod no_unused_modules;
     pub mod default;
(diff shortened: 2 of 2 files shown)


camchenry:
@ityuany what is your purpose with this PR? we don't publish `oxc_linter` to `crates.io` and we are not planning on building the linter as a standalone library as far as I know. I don't think we should do this, so that we can modify the internal APIs however we need to improve performance and make breaking changes and such.

cc @Boshen for your thoughts

ityuany:
> @ityuany what is your purpose with this PR? we don't publish `oxc_linter` to `crates.io` and we are not planning on building the linter as a standalone library as far as I know. I don't think we should do this, so that we can modify the internal APIs however we need to improve performance and make breaking changes and such.
> 
> 
> 
> cc @Boshen for your thoughts

I like oxc very much. In fact, I have used a lot of oxc crate in my daily work. Now, because of the excellent performance of oxc_linter, I hope to use oxc_linter as a crate. I can Install and use through the github link, and it is not necessary to publish it to crate.io. At the same time, I can also accept the continuous changes of the internal api, and I can modify my code to adapt it.

Boshen:
I'm conflicted. Let me think about this.

ityuany:
There is another issue ,  [because macro is used here](https://github.com/oxc-project/oxc/blob/bff83c96a21d582775b65d97123d3a2ffabf30be/crates/oxc_linter/src/rules.rs#L544)

Boshen:
The primary focus of our linter is to deliver features to end users via the binary application `oxlint`.

At present, we are unable to accommodate changes that fall outside of this scope, such as modifying our linter infrastructure for consumption as a crate.

Your available options are:

* Fork the repository and implement your desired changes.
* Contact us for commercial, customized linter support.

---

You have already made similar requests:

* https://github.com/oxc-project/oxc/issues/8070
* https://github.com/oxc-project/oxc/issues/7509
* https://github.com/oxc-project/oxc/issues/5493
* https://github.com/oxc-project/oxc/pull/9357

Please do not create similar issues that will increase our maintenance burden, thank you for your understanding.

PR #9364, opened 2025-02-25
fix(linter): support more cases for no_redundant_constructor_init

BenShelton:
Follow-up from https://github.com/oxc-project/oxc/pull/9299

As pointed out in discussions there, the initial implementation doesn't handle assignments nested in blocks as only the body's statements were being checked.

This instead checks for any references to each public param, checks it is a read (which it would be in the right-side of an assignment) and then if it's part of an assignment to `this`.

Adds tests for these new cases, previous tests still pass.

### DIFF
--- a/crates/oxc_linter/src/rules/oxc/no_redundant_constructor_init.rs
+++ b/crates/oxc_linter/src/rules/oxc/no_redundant_constructor_init.rs
@@ -1,6 +1,6 @@
 use oxc_ast::{
     AstKind,
-    ast::{AssignmentTarget, Expression, MethodDefinitionKind, Statement},
+    ast::{AssignmentTarget, Expression, MethodDefinitionKind},
 };
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/oxc/no_redundant_constructor_init.rs
```diff
@@ -58,53 +58,48 @@ impl Rule for NoRedundantConstructorInit {
...
-            }
+                // check for assigning to this: this.x = ?
+                let AssignmentTarget::StaticMemberExpression(static_member_expr) =
+                    &assignment_expr.left
+                else {
+                    continue;
+                };
+                let Expression::ThisExpression(_this_expr) = &static_member_expr.object else {
+                    continue;
+                };
```
camchenry, line 89: if you don't need `_this_expr`:

```suggestion
                if !matches!(&static_member_expr.object, Expression::ThisExpression(_)) {
                    continue;
                }
                let Expression::ThisExpression(_this_expr) = &static_member_expr.object else {
                    continue;
                };
```

crates/oxc_linter/src/rules/oxc/no_redundant_constructor_init.rs
```diff
@@ -58,53 +58,48 @@ impl Rule for NoRedundantConstructorInit {
...
-                &assignment_expr.left
-            else {
-                return;
-            };
-            let Expression::ThisExpression(_this_expr) = &static_member_expr.object else {
-                return;
-            };
-            let assignment_name = static_member_expr.property.name;
+                let Some(parent_node) = ctx.nodes().parent_node(reference.node_id()) else {
+                    continue;
```
Dunqing, line 75:
```suggestion
                let Some(parent_kind) = ctx.nodes().parent_kind(reference.node_id()) else {
                    continue;
```

Dunqing (approved):

crates/oxc_linter/src/rules/oxc/no_redundant_constructor_init.rs
BenShelton: That's a handy shortcut, thanks for pointing it out, will update 👍🏻

PR #9371, opened 2025-02-26
feat(transformer): support `ModuleRunnerTransform`

### LINKED ISSUE #9169: transformer: vite ssrTransform
Port from 
https://github.com/vitejs/vite/blob/00deea4ff88e30e299cb40a801b5dc0205ac913d/packages/vite/src/node/ssr/ssrTransform.ts

Tests at https://github.com/vitejs/vite/blob/00deea4ff88e30e299cb40a801b5dc0205ac913d/packages/vite/src/node/ssr/__tests__/ssrTransform.spec.ts

Dunqing:
close: #9169 

NOTE: I didn't copy any `sourcemap` tests to here, because these are not directly related to this plugin.

### DIFF
--- a/crates/oxc_transformer/src/plugins/module_runner_transform.rs
+++ b/crates/oxc_transformer/src/plugins/module_runner_transform.rs
@@ -124,7 +124,9 @@ impl<'a> ModuleRunnerTransform<'a> {
         // Reserve enough space for new statements
         let mut new_stmts: ArenaVec<'a, Statement<'a>> =
             ctx.ast.vec_with_capacity(program.body.len() * 2);
+
         let mut hoist_index = None;
+        let mut hoist_imports = Vec::with_capacity(program.body.len());
 
         for stmt in program.body.drain(..) {
             match stmt {
@@ -132,9 +134,8 @@ impl<'a> ModuleRunnerTransform<'a> {
                     let ImportDeclaration { span, source, specifiers, .. } = import.unbox();
                     let import_statement = self.transform_import(span, source, specifiers, ctx);
                     // Needs to hoist import statements to the above of the other statements
-                    if let Some(index) = hoist_index {
-                        new_stmts.insert(index, import_statement);
-                        hoist_index.replace(index + 1);
+                    if hoist_index.is_some() {
+                        hoist_imports.push(import_statement);
                     } else {
                         new_stmts.push(import_statement);
                     }
@@ -158,6 +159,12 @@ impl<'a> ModuleRunnerTransform<'a> {
             // Found that non-import statements, we can start hoisting import statements after this
             if hoist_index.is_none() {
                 hoist_index.replace(new_stmts.len() - 1);
+            }
+        }
+
+        if let Some(index) = hoist_index {
+            if !hoist_imports.is_empty() {
+                new_stmts.splice(index..index, hoist_imports);
             }
         }
 
--- a/crates/oxc_transformer/src/plugins/module_runner_transform.rs
+++ b/crates/oxc_transformer/src/plugins/module_runner_transform.rs
@@ -536,11 +536,12 @@ impl<'a> ModuleRunnerTransform<'a> {
                         Statement::FunctionDeclaration(func),
                         Self::create_export(span, ident, DEFAULT, ctx),
                     ]);
-                    return;
+                } else {
+                    func.r#type = FunctionType::FunctionExpression;
+                    let right = Expression::FunctionExpression(func);
+                    new_stmts.push(Self::create_export_default_assignment(span, right, ctx));
                 }
-                // Without name, treat it as an expression
-                func.r#type = FunctionType::FunctionExpression;
-                Expression::FunctionExpression(func)
+                return;
             }
             ExportDefaultDeclarationKind::ClassDeclaration(mut class) => {
                 if let Some(id) = &class.id {
@@ -549,12 +550,12 @@ impl<'a> ModuleRunnerTransform<'a> {
                         Statement::ClassDeclaration(class),
                         Self::create_export(span, ident, DEFAULT, ctx),
                     ]);
-                    return;
+                } else {
+                    class.r#type = ClassType::ClassExpression;
+                    let right = Expression::ClassExpression(class);
+                    new_stmts.push(Self::create_export_default_assignment(span, right, ctx));
                 }
-
-                // Without name, treat it as an expression
-                class.r#type = ClassType::ClassExpression;
-                Expression::ClassExpression(class)
+                return;
             }
             ExportDefaultDeclarationKind::TSInterfaceDeclaration(_) => {
                 // Do nothing for `export default interface Foo {}`
@@ -767,6 +768,22 @@ impl<'a> ModuleRunnerTransform<'a> {
         ]);
 
         ctx.ast.statement_expression(span, Self::create_define_property(arguments, ctx))
+    }
+
+    // __vite_ssr_exports__.default = right;
+    fn create_export_default_assignment(
+        span: Span,
+        right: Expression<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) -> Statement<'a> {
+        let object =
+            ctx.create_unbound_ident_expr(SPAN, SSR_MODULE_EXPORTS_KEY, ReferenceFlags::Read);
+        let property = ctx.ast.identifier_name(SPAN, DEFAULT);
+        let member = ctx.ast.member_expression_static(SPAN, object, property, false);
+        let target = AssignmentTarget::from(member);
+        let operator = AssignmentOperator::Assign;
+        let assignment = ctx.ast.expression_assignment(SPAN, operator, target, right);
+        ctx.ast.statement_expression(span, assignment)
     }
 }
 
@@ -1233,28 +1250,10 @@ Object.defineProperty(__vite_ssr_exports__, 'B', {
     #[test]
     fn should_handle_default_export_variants() {
         // default anonymous functions
-        test_same(
-            "export default function() {}",
-            "Object.defineProperty(__vite_ssr_exports__, 'default', {
-  enumerable: true,
-  configurable: true,
-  get() {
-    return function() {};
-  }
-});",
-        );
+        test_same("export default function() {}", "__vite_ssr_exports__.default = function() {};");
 
         // default anonymous class
-        test_same(
-            "export default class {}",
-            "Object.defineProperty(__vite_ssr_exports__, 'default', {
-  enumerable: true,
-  configurable: true,
-  get() {
-    return class {};
-  }
-});",
-        );
+        test_same("export default class {}", "__vite_ssr_exports__.default = class {};");
 
         // default named functions
         test_same(
--- a/crates/oxc_transformer/src/plugins/module_runner_transform.rs
+++ b/crates/oxc_transformer/src/plugins/module_runner_transform.rs
@@ -60,6 +60,7 @@ use crate::utils::ast_builder::{
     create_compute_property_access, create_member_callee, create_property_access,
 };
 
+#[derive(Debug, Default)]
 pub struct ModuleRunnerTransform<'a> {
     /// Uid for generating import binding names.
     import_uid: u32,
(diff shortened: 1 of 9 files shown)


Dunqing:
* **#9371** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9371?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9371?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9371).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/vite/module_runner_transform.rs
```diff
@@ -0,0 +1,2006 @@
...
+            "function foo() {}
+Object.defineProperty(__vite_ssr_exports__, 'default', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    return foo;
+  }
+});
+foo.prototype = Object.prototype;",
+        );
```
Dunqing: A little difference in this case, the `Object.defineProperty(__vite_ssr_exports__` is inserted following original export declaration but in Vite, it is inserted following `foo.prototype = Object.prototype;`

See https://github.com/vitejs/vite/blob/00deea4ff88e30e299cb40a801b5dc0205ac913d/packages/vite/src/node/ssr/__tests__/ssrTransform.spec.ts#L504-L514

I think it doesn't matter because there is no runtime behavior change if I understand correctly.

crates/oxc_transformer/src/vite/module_runner_transform.rs
```diff
@@ -0,0 +1,2006 @@
...
+});
+class B extends A {}
+Object.defineProperty(__vite_ssr_exports__, 'B', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    return B;
+  }
+});",
+        );
```
Dunqing: The output difference is the same as above case

Dunqing:
Hello @sapphi-red, the PR is done. It would be great if you could review this PR!

Dunqing:
The next thing I going to do is expose a function that run this plugin in NAPI, so that `Vite` can integrate it.

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+        );
+    }
+
+    #[test]
+    fn export_then_import_minified() {
+        test_same(
+            "export * from 'vue';import {createApp} from 'vue'",
+            "const __vite_ssr_import_0__ = await __vite_ssr_import__('vue');
+__vite_ssr_exportAll__(__vite_ssr_import_0__);
+const __vite_ssr_import_1__ = await __vite_ssr_import__('vue', { importedNames: ['createApp'] });",
```
sapphi-red, line 1098: The order is different from Vite.
https://github.com/vitejs/vite/blob/2c51565ec044904a080ef5649034c37f02212c7b/packages/vite/src/node/ssr/__tests__/ssrTransform.spec.ts#L211
But it should be fine. (just pointing out for future reference)

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+        let result = transform(source_text, true).unwrap();
+        if result != expected {
+            let diff = TextDiff::from_lines(&expected, &result);
+            print_diff_in_terminal(&diff);
+            panic!("Expected code does not match the result");
+        }
+    }
+
+    #[test]
+    fn default_import() {
```
sapphi-red, line 847: If we are going to support `.deps` / `.dynamicDeps`, I think we need to add tests for it (`ssrTransform.spec.ts` has them).

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+            "import { Foo } from './dependency'; class A extends Foo {}",
+            "const __vite_ssr_import_0__ = await __vite_ssr_import__('./dependency', { importedNames: ['Foo'] });
+class A extends __vite_ssr_import_0__.Foo {}",
+        );
+    }
+
+    #[test]
+    fn should_declare_exported_super_class() {
+        test_same(
+            "import { Foo } from './dependency'; export default class A extends Foo {} export class B extends Foo {}",
```
sapphi-red, line 1204:
```suggestion
            "import { Foo } from './dependency'; export default class A extends Foo {}; export class B extends Foo {}",
```
Isn't `;` or `\n` needed here? My brain interpreter fails to parse this code 🤔

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+        test_same(
+            "export default function() {}",
+            "Object.defineProperty(__vite_ssr_exports__, 'default', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    return function() {};
+  }
+});",
+        );
```
sapphi-red, line 1238: I think this should be
```js
const temp = function () {};
Object.defineProperty(__vite_ssr_exports__, 'default', {
  enumerable: true,
  configurable: true,
  get() {
    return temp;
  }
});
// or
Object.defineProperty(__vite_ssr_exports__, 'default', {
  enumerable: true,
  configurable: true,
  value: function () {}
});
// or
__vite_ssr_exports__.default = function () {};
```
otherwise
```js
// foo.js
import bar from './bar.js'
import baz from './baz.js'
if (bar !== baz) {
  throw new Error("should be same reference")
}
// bar.js
import baz from './baz.js'
export default baz
// baz.js
export default function () {}
```
would probably fail.
(same for anonymous classes)

crates/oxc_transformer/src/vite/module_runner_transform.rs
```diff
@@ -0,0 +1,2006 @@
...
+                add = null
+            }",
+            "
+            const __vite_ssr_import_0__ = await __vite_ssr_import__('vue', { importedNames: ['remove', 'add'] });
+            class A {
+              remove = 1;
+              add = null;
+            }
+            ",
+        );
```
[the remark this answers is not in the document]
sapphi-red: > I think the Vite's output is incorrect

These variables are injected but not referenced. So it is correct; just that it doesn't have to.

> Further improvement here, I think we can remove the import if no identifiers refer to this import's bindings.

We can strip `const __vite_ssr_import_0__ = ` and `, { importedNames: ['remove', 'add'] }`. But we still need to keep `await __vite_ssr_import__('vue')` as it may have a side effect.

sapphi-red (commented):
Awesome!

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+/// Based on [Vite](https://github.com/vitejs/vite/blob/00deea4ff88e30e299cb40a801b5dc0205ac913d/packages/vite/src/node/ssr/ssrTransform.ts)'s ssrTransform.
+///
+/// All tests are copy-pasted from (ssrTransform.spec.ts)[https://github.com/vitejs/vite/blob/00deea4ff88e30e299cb40a801b5dc0205ac913d/packages/vite/src/node/ssr/__tests__/ssrTransform.spec.ts]
+///
+/// ## Integrate into main `Transformer` in future
+///
+/// There are few problems to integrate this transform into the main transformer:
+///
+/// 1. In Vite, it will collect import deps and dynamic import deps during the transform process, and return them
+/// at the end of function. We can do this, but how to pass them into the js side?
```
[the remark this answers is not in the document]
Dunqing: Could you take a look in Vite it's possible to not collect `deps` and `importDeps`. In my opinion, I don't think this is what the transform plugin should do. Also will unblock integrating into the main transformer in the future. Thanks!

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+        ctx.generate_binding_in_current_scope(name, SymbolFlags::BlockScopedVariable)
+    }
+
+    // { importedNames: ['foo', 'bar'] }
+    fn create_imported_names_object(
+        elements: ArenaVec<'a, ArrayExpressionElement<'a>>,
+        ctx: &TraverseCtx<'a>,
+    ) -> Argument<'a> {
+        let value = ctx.ast.expression_array(SPAN, elements, None);
+        let key = ctx.ast.property_key_static_identifier(SPAN, Atom::from("importedNames"));
```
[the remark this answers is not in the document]
Dunqing: There is no performance difference, it's just a personal preference. `Atom::new_const` is used to solve defining constant variable. `importedNames` only used here, so we don't need to hoist it to a constant variable

crates/oxc_transformer/src/plugins/module_runner_transform.rs
Dunqing: Fixed in https://github.com/oxc-project/oxc/pull/9371/commits/328009e5415b84c2fd8d5624da8d4905a718feb5

crates/oxc_transformer/src/plugins/module_runner_transform.rs
Dunqing: Fixed in https://github.com/oxc-project/oxc/pull/9371/commits/67d2504c1579aa70bf29f5c085a778bddfced2a7

crates/oxc_transformer/src/plugins/module_runner_transform.rs
Dunqing: Thank you for pointing it out.

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+        let mut hoist_index = None;
+
+        for stmt in program.body.drain(..) {
+            match stmt {
+                Statement::ImportDeclaration(import) => {
+                    let ImportDeclaration { span, source, specifiers, .. } = import.unbox();
+                    let import_statement = self.transform_import(span, source, specifiers, ctx);
+                    // Needs to hoist import statements to the above of the other statements
+                    if let Some(index) = hoist_index {
+                        new_stmts.insert(index, import_statement);
```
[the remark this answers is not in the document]
Dunqing: This is also what I concern problem here, thank you for inspiring me to optimize it. See https://github.com/oxc-project/oxc/pull/9371/commits/328009e5415b84c2fd8d5624da8d4905a718feb5

crates/oxc_transformer/src/plugins/module_runner_transform.rs
```diff
@@ -0,0 +1,2055 @@
...
+});
+Object.defineProperty(__vite_ssr_exports__, 'c', {
+  enumerable: true,
+  configurable: true,
+  get() {
+    return __vite_ssr_import_0__.computed;
+  }
+});",
+        );
+    }
```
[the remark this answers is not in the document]
Dunqing: Added in https://github.com/oxc-project/oxc/pull/9371/commits/1dd1c0ce3b9060e93bd2085bd5c5d5d0676833ef

crates/oxc_transformer/src/vite/module_runner_transform.rs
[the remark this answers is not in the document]
Dunqing: Thank you confirm this.

> We can strip const __vite_ssr_import_0__ =  and , { importedNames: ['remove', 'add'] }. But we still need to keep await __vite_ssr_import__('vue') as it may have a side effect.

I will try to improve this after this transform plugin adapted by `Vite`

crates/oxc_transformer/src/plugins/module_runner_transform.rs
Dunqing: I am prefer not to support them. See https://github.com/oxc-project/oxc/pull/9371#discussion_r1974791415

crates/oxc_transformer/src/plugins/module_runner_transform.rs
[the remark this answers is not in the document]
sapphi-red: I'll take a look 👍

sapphi-red (approved):

PR #9377, opened 2025-02-26
feat(parser): support V8 intrinsics

srijan-paul:


Boshen:
Your fork main is 1102 commits behind 😅

srijan-paul:
my bad. will close and re-open another pr

PR #9379, opened 2025-02-26
feat(parser): support V8 intrinsics

srijan-paul:
V8 has custom syntax for runtime calls, commonly called "[intrinsics](https://source.chromium.org/chromium/v8/v8.git/+/c843aa75886abbd9aa4ea65818ece3d661ab2ed3:src/runtime/runtime.h)".
This PR implements support for this across the toolchain.


crates/oxc_ast/src/ast/js.rs
```diff
@@ -2562,3 +2565,14 @@ pub enum ModuleExportName<'a> {
...
 }
+
+/// Intrinsics in Google's V8 engine `%GetOptimizationStatus`.
+/// See: https://github.com/v8/v8/blob/5fe0aa3bc79c0a9d3ad546b79211f07105f09585/src/runtime/runtime.h#L43
+#[ast(visit)]
+#[derive(Debug)]
+#[generate_derive(CloneIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
+pub struct V8IntrinsicExpression<'a> {
+    pub span: Span,
+    pub name: IdentifierReference<'a>,
```
overlookmotel: I think this should be `IdentifierName` not `IdentifierReference`.

```js
const Foo = 123;
%Foo(123);
```

`Foo` in `%Foo` does not act as a reference to `const Foo`.

V8 intrinsic names should probably be considered a separate namespace from symbols, in the same way as:

* `label` in `label: while (whatever) { break label; }`
* `div` in JSX `<div>stuff</div>`

srijan-paul: Yep. Already fixed in the followup.
This is in a WIP state, I can drop a ping in the discord server when it's ready, if that helps :)

overlookmotel: Note: `Foo` in `<Foo>` is an `IdentifierReference`, but `div` in `<div>` is an `IdentifierName`, because it doesn't refer to variable called `div`.

overlookmotel: Sorry, not trying to step on your toes before you're ready. Just I was interested, so I took a look.

leaysgur:
Just FYI: Babel AST only has a `V8IntrinsicIdentifier`, not a whole expression. 👀 

> https://github.com/babel/babel/blob/main/packages/babel-parser/src/plugins/v8intrinsic.ts#L18

srijan-paul:
@leaysgur That's right! I was going for something closer to what [V8 itself does](https://github.com/v8/v8/blob/3c8ed98716870fdeed8e5acea2b99ace054af464/src/ast/ast.h#L1871).

Does OXC want to run Babel's [V8 intrinsic test fixtures](https://github.com/babel/babel/tree/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-parser/test/fixtures/v8intrinsic)? If so, then I can add a `V8IntrinsicIdentifier` node instead.

However I feel we can do slightly better than Babel. Intrinsics are *always* runtime calls (intrinsic identifiers on their own don't occur anywhere else), and have stricter syntax rules than other `CallExpression`s – for instance, an argument can't be a `SpreadElement`. We can enforce both invariants in the type system if we keep `V8IntrinsicExpression` a different node type. What do you think?

cc: @overlookmotel

Boshen:
Closer to v8 so that Nova can implement its own debugging APIs, which was our original intention of adding this.

Boshen (commented):
LGTM. A parser option and a few codegen tests should be added.

leaysgur:
As for the `oxc_prettier` related code, there are a few things that want to confirm & fix, but OK, I'll deal with them later. 🙆🏻 

You may need to run `cargo run -p oxc_prettier_conformance` to update the snapshot.

PR #9382, opened 2025-02-26
fix(oxlint): current dir as arg

### LINKED ISSUE #9023: Linter: `"ignorePatterns"` not causing files to be ignored when `.` is passed in to CLI
Thanks for making Oxc! It's fast and very helpful. 🙂

I've noticed that, in certain cases, `"ignorePatterns"` in `.oxlintrc.json` does not ignore certain files when running `oxlint --fix`. This only seems to occur with deep directories, and only when adding `.` to the end of the CLI command.

[Here is a minimal repro of the issue](https://github.com/nolanlawson/repro-oxlint-ignore-issue):

```bash
git clone git@github.com:nolanlawson/repro-oxlint-ignore-issue.git repro
cd repro
npm i
npm run lint
```

Here is what's happening. `"ignorePatterns"`:

```json
  "ignorePatterns": [
    "foo.js",
    "a/bar.js"
  ],
```

This works, and ignores both files:

```bash
oxlint -c .oxlintrc.json --fix
```

However, adding `.` to the end of the command causes `bar.js` to no longer be ignored:

```bash
oxlint -c .oxlintrc.json --fix .
```

If you run this second command, you'll see that `foo.js` is correctly ignored, but `a/bar.js` is not:

```diff
--- a/a/bar.js
+++ b/a/bar.js
@@ -1,2 +1,2 @@
-debugger
+
 console.log("Don't lint me!!")
```

BenShelton:
closes #9023

When passing `.` as an arg it would result in a path like `{cwd}/.`, which when being used to resolve other paths obviously caused issues. This canonicalizes any path args before doing anything else.

Adds test to verify this based on original issue. This also caused another snapshot to change for an invalid file extension, changing from `LintSucceeded` to `LintNoFilesFound`, which I believe is actually correct? If you only pass files that do not exist it would make more sense to return an error to me so I'm viewing this as a side effect fix.

### DIFF
--- a/apps/oxlint/fixtures/ignore_file_current_dir/.oxlintrc.json
+++ b/apps/oxlint/fixtures/ignore_file_current_dir/.oxlintrc.json
@@ -1,0 +1,10 @@
+{
+  "ignorePatterns": [
+    "foo.js",
+    "a/bar.js"
+  ],
+  "rules": {
+    "no-debugger": "deny"
+  }
+}
+
--- a/apps/oxlint/fixtures/ignore_file_current_dir/a/bar.js
+++ b/apps/oxlint/fixtures/ignore_file_current_dir/a/bar.js
@@ -1,0 +1,2 @@
+debugger
+console.log("Don't lint me!!")
--- a/apps/oxlint/fixtures/ignore_file_current_dir/foo.js
+++ b/apps/oxlint/fixtures/ignore_file_current_dir/foo.js
@@ -1,0 +1,2 @@
+debugger
+console.log("Don't lint me!!")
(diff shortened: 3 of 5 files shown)


Sysix:
my windows setup is broken. 
I expect that some other bugs will be introduces with `canonicalize` under windows.
`canonicalize` will generate a `\\?\` prefix, which caused other ignore bugs. See the PR for more info:

https://github.com/oxc-project/oxc/pull/8214

Boshen:
`canonicalize` is also a very slow sys call, which should avoid calling it on all paths.

shulaoda (approved):

Sysix:
@Boshen what was the reason of marking it as a draft?

Issue #8631, opened 2025-01-21
linter: import-plugin lints unexpected files


spmiller:
### What version of Oxlint are you using?

latest

### What command did you run?

npx oxlint@latest --import-plugin -A all

### What does your `.oxlint.json` config file look like?

No config file

### What happened?

Enabling the `import-plugin` starts linting files in the `node_modules` directory which cannot be disabled. Errors are not detected when the linter is run without the `import-plugin` enabled.

Steps to reproduce:

1. Add a library with code that triggers a linting error (we used [temp@0.8.3](https://www.npmjs.com/package/temp))
2. Include at least one TS file that imports the library (the bug is not triggered without an import; both .ts and .mjs files trigger the issue)
3. Run `oxlint` enabling the `import-plugin`

Expected output:
```
Finished in 12ms on 1 file with 0 rules using 16 threads.
Found 0 warnings and 0 errors.
```

Actual output:
```
  × '0'-prefixed octal literals and octal escape sequences are deprecated
     ╭─[node_modules/temp/lib/temp.js:226:21]
 225 │   var dirPath = generateName(affixes, 'd-');
 226 │   fs.mkdir(dirPath, 0700, function(err) {
     ·                     ────
 227 │     if (!err) {
     ╰────
  help: for octal literals use the '0o' prefix instead

....

Finished in 11ms on 1 file with 0 rules using 16 threads.
Found 0 warnings and 5 errors.
```

Using `--ignore-pattern` does not seem to have any effect, e.g. the bug is still reproduced when using `--ignore-pattern "**/*.js"` or `--ignore-pattern "node_modules`.

`package.json`:
```
{
  "name": "repo",
  "version": "1.0.0",
  "dependencies": {
    "temp": "^0.8.3"
  }
}
```

`test.ts`:
```
import * as temp from 'temp';
```

shulaoda:
I'm unable to reproduce your issue. It seems to have been fixed in the latest version. Could you try again, or provide more detailed information?

Boshen:
There's a race condition in the multi-file runtime.

camchenry:
Closing as completed in https://github.com/oxc-project/oxc/pull/9383

spmiller:
Many thanks @Boshen for your hard work. I really appreciate your efforts.

Issue #7996, opened 2024-12-18
linter: `import/named` is unstable


shulaoda:
Related to https://github.com/rolldown/rolldown/pull/3169

Repo: https://github.com/shulaoda/repo-cases/tree/oxc/linter-import-named

The import/named rule seems to be unstable.

```bash
  ! eslint-plugin-import(named): named import "createApp" not found
   ,-[examples/basic-vue/index.js:1:10]
 1 | import { createApp } from 'vue'
   :          ^^^^^^^^^
 2 | 
   `----
  help: does "vue" have the export "createApp"?
```

This change seems to have been introduced in [oxlint v0.15.0](https://github.com/oxc-project/oxc/releases/tag/oxlint_v0.15.0). I suspect it might be a regression caused by the fix for `import/namespace`.

shulaoda:
Actually, if you clone `rolldown` and enable `import/named` in `.oxlintrc.json`, then run `pnpm lint-code`, this issue may occur intermittently. Sometimes it throws an error, and other times it doesn’t — it's quite strange!

camchenry:
Closing as completed in https://github.com/oxc-project/oxc/pull/9383

PR #9383, opened 2025-02-26
feat(linter): a new multi-file analysis runtime

### LINKED ISSUE #9077: linter: no-cycle rule is unstable
There are few cycle imports in https://github.com/toeverything/AFFiNE latest canary branch, but oxlint can detect it only very occasionally.

I can't reproduce it on my computer, and CI is also fine, but sometimes it can be found during precommit:

Image

wan9chi:
fixes #9077, #7996, #8631.

## Results of running `oxlint -c oxlint.json` on AFFiNE@[97cc814a](https://github.com/toeverything/AFFiNE/tree/97cc814a2211b07160e3d0f02f3e2d967e8e970f)

**1.3x** slower, takes **0.86x** memory.

### Before

```
$ /usr/bin/time -al ../oxc/target/release/oxlint_main -c oxlint.json
[lint error messages omitted]
Found 0 warnings and 3 errors.
Finished in 588ms on 4990 files with 135 rules using 10 threads.
        0.61 real         3.69 user         0.94 sys
           402784256  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               31315  page reclaims
                   0  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
                   2  voluntary context switches
                6127  involuntary context switches
         42060793125  instructions retired
         13307821784  cycles elapsed
           398427968  peak memory footprint
```

### After 

```
$ /usr/bin/time -al ../oxc/target/release/oxlint -c oxlint.json
[lint error messages omitted]
Found 0 warnings and 6 errors.
Finished in 764ms on 4990 files with 135 rules using 10 threads.
        0.77 real         4.07 user         1.49 sys
           348061696  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               31412  page reclaims
                   0  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
                   5  voluntary context switches
                6625  involuntary context switches
         44886241869  instructions retired
         15977031204  cycles elapsed
           343606976  peak memory footprint
```


crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+        // all these paths must stay in memory (because they are both entries and part of `src/index.js` dependencies)
+        // until the last dependency is processed.
+        // The more efficient way is to start from "leaf" modules: their dependencies are ready earlier, thus we
+        // can run lint on them and then released their content earlier.
+        //
+        // But it's impossible to know which ones are "leaf" modules before parsing even starts. Here we assume
+        // deeper paths are more likely to be leaf modules  (src/very/deep/path/baz.js is likely to have
+        // fewer dependencies than src/index.js).
+        // This heuristic is not always true, but it works well enough for real world codebases.
+        self.paths.par_sort_unstable_by(|a, b| Path::new(b).cmp(Path::new(a)));
```
Boshen, line 241: Sort by `Path` is going to be slow. Maybe sort by number of `/` or `\\` with `sort_by_cached_key`?

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -2,43 +2,121 @@ use std::{
...
+    source: JavaScriptSource<'a>,
+    /// None if section parsing failed. The corresponding item with the same index in
+    /// `ProcessedModule.section_module_records` would be `Err(Vec<OxcDiagnostic>)`.
+    semantic: Option<Semantic<'a>>,
+}
+
+/// A module ready for linting. A `EntryModule` is generated for each path in `runtime.paths`
+///
+/// It's basically the same as `ProcessedModule`, except `content` is non-Option.
+struct EntryModule {
```
Boshen, line 102: `Entry` is a really vague term ... We need a longer and more descriptive name.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+        // that is small enough to hold in memory but big enough to make use of the rayon thread pool. We build the
+        // module graph from this group of entries, run lint on them, drop their content but keep the module
+        // graph, and then move on to the next group.
+        // This size is empirical based on AFFiNE@97cc814a.
+        let entry_group_size = rayon::current_num_threads() * 4;
+
+        // only stores entry modules in current group
+        let mut entry_modules: Vec<EntryModule> = Vec::with_capacity(entry_group_size);
+
+        // downgrade self to immutable reference so it can be shared among spawned tasks.
```
Boshen, line 253: The wording "downgrade" feels weird, may "set to ..."?

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -2,43 +2,121 @@ use std::{
...
+}
+
+/// A module processed from a path
+#[derive(Default)]
+struct ProcessedModule {
+    /// Module records of source sections, or diagnostics if parsing failed on that section.
+    ///
+    /// Modules with special extensions such as .vue could contain multiple source sections (see `PartialLoader::PartialLoader`).
+    /// Plain ts/js modules have one section. Using `SmallVec` to avoid allocations for plain modules.
+    section_module_records: SmallVec<[Result<ResolvedModuleRecord, Vec<OxcDiagnostic>>; 1]>,
```
Boshen, line 59: The term `section` is vague, what is `source sections`?

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+        // the module graph
+        let mut modules_by_path =
+            FxHashMap::<Arc<OsStr>, SmallVec<[Arc<ModuleRecord>; 1]>>::default();
+
+        // `encountered_paths` prevents duplicated processing.
+        // It is a superset of keys of `modules_by_path` as it also contains paths that are queued to process.
+        let mut encountered_paths =
+            FxHashSet::<Arc<OsStr>>::with_capacity_and_hasher(me.paths.len(), FxBuildHasher);
+
+        let mut module_relationships =
```
Boshen, line 265: What is `module_relationships`? Probably need a better naming and a comment.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+                    return;
+                }
+                let records = &modules_by_path[&path];
+                assert_eq!(records.len(), requested_module_paths.len());
+                for (record, requested_module_paths) in
+                    records.iter().zip(requested_module_paths.into_iter())
+                {
+                    let mut loaded_modules = record.loaded_modules.write().unwrap();
+                    for (specifier, dep_path) in requested_module_paths {
+                        // TODO: revise how to store multiple sections in loaded_modules
```
Boshen, line 365: Let's figure out out the TODO, it's unlikely we'll revisit this part of code and fix it later.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -228,69 +629,29 @@ impl Runtime {
...
             // Retrieve all dependent modules from this module.
             let dir = path.parent().unwrap();
-            module_record
+            requested_module_paths = module_record
                 .requested_modules
-                .keys()
-                .par_bridge()
-                .map_with(self.resolver.as_ref().unwrap(), |resolver, specifier| {
-                    resolver.resolve(dir, specifier).ok().map(|r| (specifier, r))
+                .par_iter()
```
Boshen, line 648: `par_iter` is probably not required anymore, since there's no further processing other than `resolver.resolve`. Spawning a new thread here may be more expensive than `resolver.resolve` it self. We need the threads for heavy duties such as parse + lint.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+
+        // `encountered_paths` prevents duplicated processing.
+        // It is a superset of keys of `modules_by_path` as it also contains paths that are queued to process.
+        let mut encountered_paths =
+            FxHashSet::<Arc<OsStr>>::with_capacity_and_hasher(me.paths.len(), FxBuildHasher);
+
+        let mut module_relationships =
+            Vec::<(Arc<OsStr>, SmallVec<[Vec<(/*specifier*/ CompactStr, Arc<OsStr>)>; 1]>)>::new();
+
+        let (tx_resolve_output, rx_resolve_output) = mpsc::channel::<ModuleProcessOutput>();
```
Boshen, line 268: I think it's best to write an algorithm overview here, because there are so many loops and multi-thread logic here.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+
+                if let Some(entry_module) =
+                    EntryModule::from_processed_module(path, processed_module)
+                {
+                    entry_modules.push(entry_module);
+                }
+            }
+
+            // Writing to `loaded_modules` based on `module_relationships`
+            module_relationships.par_drain(..).for_each(|(path, requested_module_paths)| {
```
Boshen, line 354: Why `par_drain`? I'm not seeing any heavy processing, spawning threads may be slower than the acutaly work done here.

crates/oxc_linter/src/service/runtime.rs
```diff
@@ -102,105 +191,424 @@ impl Runtime {
...
+                if requested_module_paths.is_empty() {
+                    return;
+                }
+                let records = &modules_by_path[&path];
+                assert_eq!(records.len(), requested_module_paths.len());
```
Boshen, line 359: using `assert_eq` will crash in release build, we probably need a panic message here.

crates/oxc_linter/src/service/runtime.rs
branchseer: I tried to sort by counting `/` or `\\` and found the memory usage regressed a lot. I'm 100% sure why but my guess is it's important that files under same folders are near each other, because they are more likely to be in a same module graph. Processing files under same folders in a group reduces the possibility of prematurely discovering other modules to lint.

Profiling shows that the time of this sort is just a tiny fraction of the whole workflow. I think we can keep it for the better grouping strategy.

crates/oxc_linter/src/service/runtime.rs
branchseer: Renamed to `ModuleToLint`, as opposed to modules required by the import plugin, not for linting. Let's me know if you have better ideas.

crates/oxc_linter/src/service/runtime.rs
branchseer: Updated in c074e3b

crates/oxc_linter/src/service/runtime.rs
branchseer: The term `section` is taken from https://github.com/oxc-project/oxc/blob/cfdcfdb17cdded5cd0b5757ae2dec6aaf750c3f0/crates/oxc_linter/src/loader/partial_loader/mod.rs#L16. `PartialLoader` is also mentioned in the docs of `section_module_records`. 

Personally I find this term precise and couldn't come up with a better name.

crates/oxc_linter/src/service/runtime.rs
branchseer: resolved in c074e3b

crates/oxc_linter/src/service/runtime.rs
branchseer: The `for_each` body is indeed lightweight, but the size of `module_relationships` (now renamed to `module_paths_and_resolved_requests`) can be huge if the current group of modules have lots of dependencies. At worse case it could be the relationships of the whole module graph (unlikely with our new strategy but not impossible).

crates/oxc_linter/src/service/runtime.rs
branchseer: Added in `b0aa5f6`. I added a generic message to not confuse users with implementation details.

PR #9415, opened 2025-02-27
feat(linter): add import/no-absolute-path rule

huangtiandi1999:
Relates to #1117 
Rule detail：https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/no-absolute-path.md


camc314 (approved):
thanks!

PR #9423, opened 2025-02-27
refactor(linter): improve `eslint/no-new`

therewillbecode:
- Improve the documentation
- Add a missing case for parenthesised new expression which is covered by the equivalent eslint rule. (See eslint playground for rule [here](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tbmV3OiBcImVycm9yXCIqL1xuXG5uZXcgVGhpbmcoKTsifQ==)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_new.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_new.rs
@@ -1,7 +1,4 @@
-use oxc_ast::{
-    AstKind,
-    ast::{Expression, NewExpression},
-};
+use oxc_ast::AstKind;
 use oxc_diagnostics::OxcDiagnostic;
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/eslint/no_new.rs
```diff
@@ -22,10 +23,25 @@ declare_oxc_lint!(
...
+    /// ```
+    ///
+    /// Examples of **correct** code for this rule:
+    /// ```javascript
+    /// var a = new Date()
+    /// ```
+    ///
+    /// ```javascript
+    /// (() => new Date())
+    /// ```
```
shulaoda, line 44: I suggest placing these simple examples in the same code block and separating them with a single line.

PR #9469, opened 2025-03-01
docs(linter): add example of how configure rule

cdrikd:
I don't find an example of how a rule can be configured. It's like Eslint but I think it's better to have example in the website doc for newcomers.


Sysix:
To fix the snapshot, run:
`cargo test -p oxc_linter` and
`cargo insta review` and accept the changes.

cdrikd:
> To fix the snapshot, run: `cargo test -p oxc_linter` and `cargo insta review` and accept the changes.

done, i fix also snapshot for website.

PR #9474, opened 2025-03-01
feat(linter): add react/jsx-filename-extension rule

cdrikd:
Implement the `react/jsx-filename-extension` rule

Eslint reference : https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
@@ -96,7 +96,7 @@ declare_oxc_lint!(
     /// ```
     JsxFilenameExtension,
     react,
-    restriction, 
+    restriction,
     pending
 );
 
--- a/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
@@ -228,7 +228,7 @@ fn test() {
             Some(serde_json::json!([{ "allow": "as-needed" }])),
             None,
             Some(PathBuf::from("foo.jsx")),
-        ),        
+        ),
         (
             "module.exports = function MyComponent() { return <><Comp /><Comp /></>; }",
             Some(serde_json::json!([{ "allow": "as-needed" }])),
@@ -267,7 +267,7 @@ fn test() {
             Some(serde_json::json!([{ "extensions": [".js", ".jsx"] }])),
             None,
             Some(PathBuf::from("foo.js")),
-        ),        
+        ),
         (
             "//test\n\n//comment",
             Some(serde_json::json!([{ "allow": "as-needed" }])),
@@ -331,7 +331,6 @@ fn test() {
             None,
             Some(PathBuf::from("foo.js")),
         ),
-
         (
             "module.exports = function MyComponent() { return <div>\n<div />\n</div>; }",
             Some(serde_json::json!([{ "extensions": [".js"] }])),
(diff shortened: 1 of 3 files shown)


Sysix (approved):

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
```diff
@@ -0,0 +1,288 @@
...
+        if jsx_elt.is_some() {
+            if !has_ext_allowed {
+                let span_elt = jsx_elt.map(GetSpan::span);
+                ctx.diagnostic(no_jsx_with_filename_extension_diagnostic(
+                    file_extension,
+                    span_elt.unwrap(),
+                ));
+            }
+            return;
+        }
```
camchenry, line 162: you can use `if-let` here:

```suggestion
        if let Some(jsx_elt) = jsx_elt {
            if !has_ext_allowed {
                ctx.diagnostic(no_jsx_with_filename_extension_diagnostic(
                    file_extension,
                    jsx_elt.span(),
                ));
            }
            return;
        }
```

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        if matches!(self.allow, AllowType::AsNeeded) && has_ext_allowed {
+            let Some(root) = ctx.nodes().root_node() else {
+                return;
+            };
+            let AstKind::Program(program) = root.kind() else { unreachable!() };
+            if self.ignore_files_without_code && program.body.is_empty() {
+                return;
+            }
```
camchenry, line 171: We want to avoid using `unreachable!()`, because it could cause a panic at runtime. instead of checking the program node directly we could check `ctx.nodes().len()` and see if it is greater than 1, indicating it is not empty.

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+#[test]
+fn test() {
+    use crate::tester::Tester;
+    use std::path::PathBuf;
+
+    let pass = vec![
+        (
+            "module.exports = function MyComponent() { return <div>jsx\n<div />\n</div>; }",
```
camchenry, line 184: could we include some tests for typical export syntax like `export function MyComponent()`?

camchenry (changes requested):

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
cdrikd: thank you, i fix it.

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
cdrikd: yes you're right, I had more test with new syntax.

crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
```diff
@@ -0,0 +1,368 @@
...
+            .unwrap_or(vec![CompactStr::from("jsx")]);
+
+        Self(Box::new(JsxFilenameExtensionConfig { allow, extensions, ignore_files_without_code }))
+    }
+
+    fn run_once(&self, ctx: &LintContext) {
+        let jsx_elt = ctx
+            .nodes()
+            .iter()
+            .find(|&&x| matches!(x.kind(), AstKind::JSXElement(_) | AstKind::JSXFragment(_)));
```
Boshen, line 149: This check is slow, probably should be moved to after checking the extension?

cdrikd: I push a new commit, what do you think about this refactor c0d3550 ? 
Iteration is not always done now, is it what you think about ?

camchenry (approved):

PR #9519, opened 2025-03-03
ci(benchmarks): benchmark NAPI parser

overlookmotel:
Add benchmarks for NAPI parser. Benchmarks cover both "standard" mode and "raw transfer" mode.

Codspeed say they've sorted out the problems with wide variance in NodeJS benchmarks, so hopefully they won't wave all over the place, the way they did last year when we first tried to add these benchmarks.

What we also found last year is that Codspeed's measures didn't reflect real-world wallclock measurements at all when comparing code running on Rust side vs JS side. Raw transfer moves a lot of work from Rust side to JS side, and so likely the bench results will be misleading if comparing "standard" vs "raw".

I've added benchmarks for both, not to compare the 2, but because I think there's room to improve the performance of both "standard" and "raw" independently.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -159,3 +159,65 @@ jobs:
         with:
           token: ${{ secrets.CODSPEED_TOKEN }}
           run: cargo codspeed run
+
+  # Build NAPI parser
+  build-parser-napi:
+    name: Build NAPI parser
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Branch
+        uses: taiki-e/checkout-action@b13d20b7cda4e2f325ef19895128f7ff735c0b3d # v1.3.1
+
+      - uses: oxc-project/setup-rust@cd82e1efec7fef815e2c23d296756f31c7cdc03d # v1.0.0
+        with:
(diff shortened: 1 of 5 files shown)


overlookmotel:
These benchmarks are really slow. I've skipped the worst ones, and sharded all the rest. But still should use a filter to skip running them unless relevant files have been touched. I'll look into that.

overlookmotel:
Argh. We *can't* skip these benchmarks conditionally as it'll foul up CodSpeed. Every PR that doesn't touch NAPI parser would get a "dropped benchmarks" warning.

I've [reached out to Codspeed](https://discord.com/channels/1065233827569598464/1065233828098097285) to ask if possible to allow benchmarks to be intentionally skipped, and they've put it on their issue tracker. But who knows how long it'll take.

We may need to take matters into our own hands and build something to skip benchmarks when their dependences haven't changed, and submit the previous benchmark run's results in its place to keep Codspeed happy (https://github.com/oxc-project/backlog/issues/3).

But in the meantime, I think we do need benchmarks for NAPI parser, as I'm going to be working on optimizing it.

camchenry (approved):

PR #9537, opened 2025-03-04
feat(linter): Add `promise/no-return-wrap`

therewillbecode:
Relates to #4655 .

Rule details: https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-return-wrap.md

### DIFF
--- a/crates/oxc_linter/src/rules/promise/no_return_wrap.rs
+++ b/crates/oxc_linter/src/rules/promise/no_return_wrap.rs
@@ -275,6 +275,11 @@ fn test() {
         ("doThing().then(() => { return Promise.a() })", None),
         ("doThing().then(() => a())", None),
         ("doThing().then(() => Promise.a())", None),
+        (
+            "class Promise { constructor(){} resolve(){} };
+             doThing().then(function() { return Promise.resolve(4) })",
+            None,
+        ),
     ];
 
     let fail = vec![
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/promise/no_return_wrap.rs
```diff
@@ -0,0 +1,541 @@
...
+    ast::{ArrowFunctionExpression, CallExpression, Expression, FunctionBody, Statement},
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{CompactStr, Span};
+
+use crate::{AstNode, context::LintContext, rule::Rule, utils::is_promise};
+
+fn no_return_wrap_diagnostic(span: Span, issue: &ReturnWrapper) -> OxcDiagnostic {
+    let warn_msg: CompactStr = match issue {
```
Boshen, line 13: Using `CompactStr` is redundant, all these messages exceed the `CompactStr` inline size limit. using `&'static str` should be fine here?

crates/oxc_linter/src/rules/promise/no_return_wrap.rs
```diff
@@ -0,0 +1,541 @@
...
+    /// })
+    /// ```
+    ///
+    /// ```js
+    /// myPromise().then().catch(() => Promise.reject("err"))
+    /// ```
+    ///
+    NoReturnWrap,
+    promise,
+    nursery,
```
Boshen, line 138: Need a explanation if you want this in nursery.

therewillbecode: Should be fine to go into `style`. What do you think?

PR #9545, opened 2025-03-04
docs(linter): improve the documentation of `typescript-no-namespace`

therewillbecode:
- Document the options for the rule as well as adding correctness examples according to #6050 .
- Format the strings to make them more readable in the test cases.

### DIFF
--- a/crates/oxc_linter/src/snapshots/typescript_no_namespace.snap
+++ b/crates/oxc_linter/src/snapshots/typescript_no_namespace.snap
@@ -79,217 +79,200 @@ source: crates/oxc_linter/src/tester.rs
   help: Replace the namespace with an ES2015 module or use `declare module`
 
   ⚠ typescript-eslint(no-namespace): ES2015 module syntax is preferred over namespaces.
-   ╭─[no_namespace.tsx:2:7]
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/snapshots/typescript_no_namespace.snap
```diff
@@ -79,217 +79,200 @@ source: crates/oxc_linter/src/tester.rs
...
-   ╭─[no_namespace.tsx:3:9]
- 2 │             namespace A {
- 3 │               namespace B {
-   ·               ─────────
- 4 │                 declare namespace C {}
+   ╭─[no_namespace.tsx:2:10]
+ 1 │ namespace A {
+ 2 │                namespace B {
+   ·                ─────────
+ 3 │                  declare namespace C {}
```
therewillbecode, line 111: Making the test case strings prettier makes the snapshots less pretty. Should I revert the test case formatting changes?

Boshen: The snapshots are mainly used for checking whether the spans are pointing to the correct thing. So no need.

PR #9565, opened 2025-03-05
refactor(ast_tools): add `Output::yaml_watch_list` method

overlookmotel:
Pure refactor. Move the logic for creating a YAML watch list into it's own function.

This was originally part of a change I was working on to make NAPI parser benchmarks only run when relevant files have changed. I've now realized that is not workable because of how Codspeed works, but I think this refactor is good anyway.


Boshen:
### Merge activity

* **Mar 5, 8:15 AM EST**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #9612, opened 2025-03-08
docs(linter): Improve the documentation of `eslint-no-console`

therewillbecode:
- Add `Why is this bad section`
- Add correctness examples
- Correct the documented option name which is `allow` and not `allowed`
- Add options documentation

Above relates to #6050 

- Also adds the following missing parenthesized test case for `(console.log())`.


Boshen:
Thank you for all the hard work.

PR #9650, opened 2025-03-10
perf(allocator): remove overflow checks from `String::from_strs_array_in`

overlookmotel:
Optimize `String::from_strs_array_in` to remove overflow checks in common cases.

`&str` has a maximum length of `isize::MAX`, so in many cases, if we communicate this constraint to the compiler, it can remove the overflow checks, as it can see that overflowing `usize` is impossible.

I was unsure if max length of `&str` was guaranteed, but wiser minds confirm it is: https://users.rust-lang.org/t/does-str-reliably-have-length-isize-max/126777

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1589,6 +1589,7 @@ name = "oxc_allocator"
 version = "0.57.0"
 dependencies = [
  "allocator-api2",
+ "assert-unchecked",
  "bumpalo",
  "hashbrown 0.15.2",
  "oxc_estree",
--- a/crates/oxc_allocator/Cargo.toml
+++ b/crates/oxc_allocator/Cargo.toml
@@ -22,6 +22,7 @@ doctest = false
 oxc_estree = { workspace = true, optional = true }
 
 allocator-api2 = { workspace = true }
+assert-unchecked = { workspace = true }
 bumpalo = { workspace = true, features = ["allocator-api2", "collections"] }
 hashbrown = { workspace = true, default-features = false, features = ["inline-more", "allocator-api2"] }
 rustc-hash = { workspace = true }
--- a/crates/oxc_allocator/src/string.rs
+++ b/crates/oxc_allocator/src/string.rs
@@ -13,6 +13,7 @@ use std::{
     ptr,
 };
 
+use assert_unchecked::assert_unchecked;
 use bumpalo::collections::String as BumpaloString;
 pub use simdutf8::basic::Utf8Error;
 use simdutf8::basic::from_utf8;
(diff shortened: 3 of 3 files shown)


overlookmotel:
* **#9650** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9650?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9650?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9650).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
It looks like we don't even need to add any checks?

Dunqing (approved):

Dunqing:
### Merge activity

* **Mar 11, 2:03 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
> It looks like we don't even need to add any checks?

Yes I think we still do. We have to deal correctly with *any* possible input to the function.

Without overflow checks, this example would cause incorrect `total_len` calculation, and therefore writing out of bounds.

```rs
let s1 = "$_";
let s2 = get_long_string();
let s3 = get_long_string();
assert_eq!(s1.len(), 2);
assert_eq!(s2.len(), isize::MAX as usize);
assert_eq!(s3.len(), isize::MAX as usize);

// This does not panic in release mode without `checked_add`
let total_len = [s1, s2, s3].iter().fold(0usize, |total_len, s| total_len + s.len());
// Oh no!
assert_eq!(total_len, 0);
```

But... at least this change means that in our actual use cases, the compiler has enough information to prove that the checks aren't needed, and to remove them.

So we get complete safety no matter what the input is, but we don't have to pay for it - as long as only 1 of the input strings is dynamic, and the rest are static (so compiler knows their length).

overlookmotel:
> The example case seems like it would not happen, as https://users.rust-lang.org/t/does-str-reliably-have-length-isize-max/126777/2 explained, the maximum allocatable allocation object cannot exceed `isize::MAX`; therefore, its impossible for total allocated string length to surpass `isize::MAX` either. Am I correct in understanding this?

Yes, it's true that attempting to allocate more than `isize::MAX` bytes will fail. `Vec::with_capacity_in` will panic in this case.

But the potential problem here is *before* we allocate the `Vec`.

If we calculate `total_len` without checks and the calculation "overflows", then `total_len` can end up "wrapping around" back to 0.

So, if `total_len` is incorrectly calculated as 0, then we call `Vec::with_capacity_in(0, allocator)`. That doesn't panic, and it doesn't allocate any memory either.

But we are assuming that we *have* allocated enough memory for all the strings, and we get the pointer to the `Vec`, and write the first string at that memory address. But we didn't allocate any memory! So we just wrote that string over unallocated memory. This may result in a segfault, or it may overwrite any other data in memory. Bad news!

In safe code, Rust will prevent you from doing this - it checks these things and panics if you try to do anything illegal, and it's extremely good at covering every conceivable edge case. But in "unsafe land", we we have told Rust "don't check anything, we already checked". So the responsibility for covering every possible case is on us.

I'm not sure if this makes any more sense?

Dunqing:
> Yes, it's true that attempting to allocate more than `isize::MAX` bytes will fail. `Vec::with_capacity_in` will panic in this case.
> 
> But the potential problem here is _before_ we allocate the `Vec`.
> 
> If we calculate `total_len` without checks and the calculation "overflows", then `total_len` can end up "wrapping around" back to 0.
> 
> So, if `total_len` is incorrectly calculated as 0, then we call `Vec::with_capacity_in(0, allocator)`. That doesn't panic, and it doesn't allocate any memory either.
> 
> But we are assuming that we _have_ allocated enough memory for all the strings, and we get the pointer to the `Vec`, and write the first string at that memory address. But we didn't allocate any memory! So we just wrote that string over unallocated memory. This may result in a segfault, or it may overwrite any other data in memory. Bad news!
> 
> In safe code, Rust will prevent you from doing this - it checks these things and panics if you try to do anything illegal, and it's extremely good at covering every conceivable edge case. But in "unsafe land", we we have told Rust "don't check anything, we already checked". So the responsibility for covering every possible case is on us.
> 
> I'm not sure if this makes any more sense?


Thank you for explaining this again, that makes sense, I misunderstood that the allocated object can't be more than one https://doc.rust-lang.org/std/ptr/index.html#allocated-object

PR #9659, opened 2025-03-10
fix(linter): Add missing fail cases in `eslint-no-array-constructor`

therewillbecode:
Adds the following missing fail cases for the rule where the last argument to the array constructor is a spread element:

```js
new Array(...args);
Array(...args);
Array(1, 2, ...args);
```

I setup an eslint playground to show the cases failing [here](https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8tYXJyYXktY29uc3RydWN0b3I6IFwiZXJyb3JcIiovXG5cblxubmV3IEFycmF5KC4uLmFyZ3MpO1xuXG5BcnJheSguLi5hcmdzKTtcblxuQXJyYXkoMSwgMiwgLi4uYXJncyk7Iiwib3B0aW9ucyI6eyJydWxlcyI6e30sImxhbmd1YWdlT3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFGZWF0dXJlcyI6e319fX19)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
@@ -10,7 +10,7 @@ use oxc_span::Span;
 use crate::{AstNode, context::LintContext, rule::Rule};
 
 fn no_array_constructor_diagnostic(span: Span) -> OxcDiagnostic {
-    OxcDiagnostic::warn("Do not use `new` to create arrays")
+    OxcDiagnostic::warn("The array literal notation [] is preferable.")
         .with_help("Use an array literal instead")
         .with_label(span)
 }
@@ -67,19 +67,6 @@ impl Rule for NoArrayConstructor {
                 &new_expr.type_parameters,
                 false,
             ),
-            AstKind::ExpressionStatement(expr_stmt) => {
-                if let Expression::CallExpression(call_expr) = &expr_stmt.expression {
-                    (
-                        call_expr.span,
-                        &call_expr.callee,
-                        &call_expr.arguments,
-                        &call_expr.type_parameters,
-                        call_expr.optional,
-                    )
-                } else {
-                    return;
-                }
-            }
             _ => {
                 return;
             }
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/snapshots/eslint_no_array_constructor.snap
```diff
@@ -36,6 +43,20 @@ source: crates/oxc_linter/src/tester.rs
...
+  ⚠ eslint(no-array-constructor): Do not use `new` to create arrays
+   ╭─[no_array_constructor.tsx:1:1]
+ 1 │ new Array(...args);
+   · ──────────────────
+   ╰────
+  help: Use an array literal instead
+
+  ⚠ eslint(no-array-constructor): Do not use `new` to create arrays
+   ╭─[no_array_constructor.tsx:1:1]
+ 1 │ Array(x, y)
```
therewillbecode, line 55: Not sure why I am getting duplicate snapshots here. Must be the diagnostic being reported multiple times when it shouldn't be.

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
```diff
@@ -64,6 +67,19 @@ impl Rule for NoArrayConstructor {
                 &new_expr.type_parameters,
                 false,
             ),
+            AstKind::ExpressionStatement(expr_stmt) => {
```
Boshen, line 70: This duplicates the `CallExpression` logic above, hence the duplicated error message.

crates/oxc_linter/src/snapshots/eslint_no_array_constructor.snap
Boshen: The error message is off btw, there is no `new` in the code.

crates/oxc_linter/src/rules/eslint/no_array_constructor.rs
therewillbecode: Oh of course! Thanks.

crates/oxc_linter/src/snapshots/eslint_no_array_constructor.snap
therewillbecode: Good catch! Thanks.

camchenry (approved):

PR #9660, opened 2025-03-10
feat(linter): Add `eslint/no-lonely-if`

therewillbecode:
Relates to #479 .

Rule details: https://eslint.org/docs/latest/rules/no-lonely-if

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
@@ -98,16 +98,15 @@ impl Rule for NoLonelyIf {
             return;
         };
 
-        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_node(node.id()).map(AstNode::kind)
-        {
+        let Some(only_stmt) = b.body.first() else {
+            return;
+        };
+
+        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_kind(node.id()) {
             return;
         };
 
         if b.body.len() == 1 {
-            let Some(only_stmt) = b.body.first() else {
-                return;
-            };
-
             if let Statement::BlockStatement(inner_block) = only_stmt {
                 if inner_block.body.len() == 1 {
                     if let Some(Statement::IfStatement(lonely_if)) = inner_block.body.first() {
@@ -138,8 +137,8 @@ fn test() {
         "if (a) {
            if (b) {} else { }
          } else {
-           if (c) {  } 
-           if (d) {  } 
+           if (c) {  }
+           if (d) {  }
           }",
     ];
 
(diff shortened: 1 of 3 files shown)


crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,168 @@
...
+        };
+
+        let BlockStatement(b) = alternate else {
+            return;
+        };
+
+        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_node(node.id()).map(AstNode::kind)
+        {
+            return;
+        };
```
camchenry, line 104: use `parent_kind` instead:
```suggestion
        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_kind(node.id())
        {
            return;
        };
```

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        if b.body.len() == 1 {
```
camchenry, line 106: early return here?
```suggestion
        if b.body.len() != 1 {
            return;
        }
```

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,168 @@
...
+        "if (a) if (a) {} else { if (b) {} } else {}",
+        "if (a) {
+           if (b) {} else { }
+         } else {
+           if (c) {  } 
+           if (d) {  } 
+          }",
+    ];
+
+    let fail = vec![
```
camchenry, line 146: It looks like we are missing some test cases from eslint. could you port those as well? e.g.:

```
if (foo) {
} else {
  if (bar) baz()
}
[1, 2, 3].forEach(foo);
```

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
camchenry, line 106: can we move this check above `parent_kind` check? this is a cheaper check to do first

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
therewillbecode: Sure thing, added them!

therewillbecode:
Thanks so much for the review @camchenry.

I have addressed all the feedback so far.

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,296 @@
...
+            return;
+        };
+
+        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_kind(node.id()) {
+            return;
+        };
+
+        if b.body.len() != 1 {
+            return;
+        };
```
camchenry, line 111: these are still mis-ordered I think, more expensive checks should happen as far down as possible:
```suggestion
        if b.body.len() != 1 {
            return;
        };
        
        let Some(only_stmt) = b.body.first() else {
            return;
        };

        if let Some(AstKind::IfStatement(_)) = ctx.nodes().parent_kind(node.id()) {
            return;
        };        
```

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        if let Statement::BlockStatement(inner_block) = only_stmt {
+            if inner_block.body.len() == 1 {
+                if let Some(Statement::IfStatement(lonely_if)) = inner_block.body.first() {
+                    ctx.diagnostic(no_lonely_if_diagnostic(Span::new(
+                        lonely_if.span.start,
+                        lonely_if.span.start + 2,
+                    )));
```
camchenry, line 119: can use `Span::sized` to make this shorter:
```suggestion
                    ctx.diagnostic(no_lonely_if_diagnostic(Span::sized(
                        lonely_if.span.start,
                        2,
                    )));
```

crates/oxc_linter/src/rules/eslint/no_lonely_if.rs
```diff
@@ -0,0 +1,296 @@
...
+                };
+            }
+        } else if let Statement::IfStatement(lonely_if) = only_stmt {
+            ctx.diagnostic(no_lonely_if_diagnostic(Span::new(
+                lonely_if.span.start,
+                lonely_if.span.start + 2,
+            )));
```
camchenry, line 126:
```suggestion
            ctx.diagnostic(no_lonely_if_diagnostic(Span::sized(
                lonely_if.span.start,
                2,
            )));
```

camchenry (approved):
looks good! I had a small recommendation for the span creation as well and the check order, but otherwise looks good to merge to me

camchenry (approved):

camchenry:
Thanks for the contribution!

therewillbecode:
> Thanks for the contribution!

My pleasure. Thanks again for the helpful review.

PR #9662, opened 2025-03-11
fix(parser)!: parse TS import attributes options as object expression

ArnaudBarre:
TS-ESLint spec: https://github.com/typescript-eslint/typescript-eslint/blob/9aa0b6b49807599b1c2b704b778df74c91084a7c/packages/ast-spec/src/type/TSImportType/spec.ts#L8

ast-explorer: https://ast-explorer.dev/#eNoty8EKQiEQheFXGc6q4PYCQoteotVsRIy8eB1xhiLEd0+p7XfO35HhsPuX19BSNWyoE+xT4w8uUXMqy8Pf6UZXSkeVZifGQ4SxUX8nezrq6+CIsasUxhjnGcoMOxeaHMPh77FpWuv6ZW9RjcFlYHwB6tsulA==

I'm block by the fact that `parse_object_expression` return an expression and not an objectExpression

Feel free to close another PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ target/
 /editors/vscode/out/
 /editors/vscode/*.vsix
 
+# JetBrains
+/.idea/
+
 # Cloned conformance repos
 tasks/coverage/babel/
 tasks/coverage/test262/
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -1280,38 +1280,10 @@ pub struct TSImportType<'a> {
     pub span: Span,
     /// `true` for `typeof import("foo")`
     pub is_type_of: bool,
-    pub parameter: TSType<'a>,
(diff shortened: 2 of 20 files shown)


.gitignore
```diff
@@ -22,6 +22,9 @@ target/
 /editors/vscode/out/
 /editors/vscode/*.vsix
 
+# JetBrains
```
Boshen, line 25: https://github.com/oxc-project/oxc/blob/29edb51c09721088a9878d6c4b2ffd0c2909b0cd/.gitignore#L42-L46

Boshen:
Our AST is easier to work with, let's serialize this node as `ObjectExpression` instead.

.gitignore
ArnaudBarre: The issue is that I don't want to ignore it on my repos
But I will use a local gitignore

Boshen:
I'll fix this.

PR #9646, opened 2025-03-10
feat(allocator): add `Vec2`

Dunqing:
related: https://github.com/oxc-project/oxc/issues/9706

These two files are copied from Bumplo's `vec` and `raw_vec`, they will be used to replace allocator_api2's `Vec` later PRs.

PR #9704, opened 2025-03-12
feat(allocator): add `no_std` support

CrazyboyQCD:
Add `no_std` support for `oxc_allocator` for downstream crates that need `no_std`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
           cache-key: warm
       - run: cargo ck
       - run: cargo test --all-features
+      - run: cargo test -p oxc_allocator --no-default-features --features serialize
       - run: git diff --exit-code # Must commit everything
 
   test-windows:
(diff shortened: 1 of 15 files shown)


overlookmotel:
`oxc_allocator` crate is currently undergoing a lot of changes. We are in the process of replacing `Vec` with our own custom implementation (#9646 and later PRs). We also intend to replace `bumpalo` with our own allocator soon. So it's not an ideal time to be making this kind of change.

Out of interest, can you please outline what your use case is that requires `no_std`?

CrazyboyQCD:
This pr could be postponed when refactor is done:)

> Out of interest, can you please outline what your use case is that requires `no_std`?

Shader parser, which is needed to implement in some targets like `wasm32v1-none`

Boshen:
I'm hesitant to maintain `no_std`.

And I also couldn’t find any public repositories from you that rely on Oxc.

Please reach out / provide more context if this is for a company project.

CrazyboyQCD:
> I'm hesitant to maintain `no_std`.

I think `oxc_allocator` itself doesn't take much effort to maintain as `no_std` (no multithread code now), and making it `no_std` will allow more users to use it.

> And I also couldn’t find any public repositories from you that rely on Oxc.
> 
> Please reach out / provide more context if this is for a company project.

It is personal project currently.

Boshen:
> oxc_allocator

It would be easiest if you vendor `oxc_allocator` in your private project, it's a simple wrapper around `bumpalo` anyway.

PR #9618, opened 2025-03-08
fix(linter): turn oxc/no-redundant-constructor-init into typescript/no-unnecessary-parameter-property-assignment

ulrichstark:
Closes #9599

I switched the category to nursery because in a followup PR I will import all tests from typescript-eslint and attempt to fix all of them. This separation into two PRs makes the changes easier to review and preserves the history and original author of the lint rule.

PR #9720, opened 2025-03-12
fix(linter): import and fix tests for typescript::no_unnecessary_parameter_property_assignment

ulrichstark:
Followup to #9618.
PRs for oxc/no-redundant-constructor-init: #9299 #9364
typescript-eslint rule: [Docs](https://typescript-eslint.io/rules/no-unnecessary-parameter-property-assignment) [Source](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-unnecessary-parameter-property-assignment.ts) [Tests](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/no-unnecessary-parameter-property-assignment.test.ts)

Changes of this PR:

- Imported tests from typescript-eslint
- Added one more passing test to the end
- Fixed all new test cases
- Changed rule category back to correctness

This is my first major change to OXC.
If there's anything wrong or not the OXC-way, I will happily improve on that.


Boshen:
Nice work!

PR #9724, opened 2025-03-12
feat(oxc_language_server): Support nested configs

nrayburn-tech:
Just putting this up to have the idea validated before I clean it up and separate it out into different PRs.

Left comments inline to sort of communicate what I'm doing or why. I don't do much in Rust, so I think it might be helpful to communicate what I'm trying to do in case there are better alternatives.

- There are multiple places doing `.unwrap` that I want to review and fix.
- I want to split this into at least a separate PR for the language server changes and the VS Code changes. Maybe even additional PRs that aren't explicitly required for nested config support, such as the change on what files are watched from VS Code.


crates/oxc_language_server/src/main.rs
```diff
@@ -33,13 +32,16 @@ mod linter;
...
+const OXC_CONFIG_FILE: &str = ".oxlintrc.json";
+
 struct Backend {
     client: Client,
     root_uri: OnceCell<Option<Url>>,
     server_linter: RwLock<ServerLinter>,
     diagnostics_report_map: ConcurrentHashMap<String, Vec<DiagnosticReport>>,
     options: Mutex<Options>,
     gitignore_glob: Mutex<Vec<Gitignore>>,
+    nested_configs: RwLock<FxHashMap<PathBuf, ConfigStore>>,
```
nrayburn-tech, line 44: Store a map of directory to config for all nested configs in the workspace.

crates/oxc_language_server/src/main.rs
```diff
@@ -54,11 +56,17 @@ struct Options {
     run: Run,
     enable: bool,
     config_path: String,
+    use_nested_configs: bool,
```
nrayburn-tech, line 59: Allow opting into nested configs. If false, there shouldn't be any behavior change.

crates/oxc_language_server/src/main.rs
```diff
@@ -180,8 +188,38 @@ impl LanguageServer for Backend {
         }
     }
 
-    async fn did_change_watched_files(&self, _params: DidChangeWatchedFilesParams) {
+    async fn did_change_watched_files(&self, params: DidChangeWatchedFilesParams) {
         debug!("watched file did change");
+        if self.options.lock().await.use_nested_configs {
```
nrayburn-tech, line 193: If using nested configs, check the watched files to determine if they are `.oxlintrc.json` files (which are the only nested config files that are supported to my knowledge).

This just keeps the nested config map up to date with any workspace changes.

crates/oxc_language_server/src/main.rs
```diff
@@ -492,21 +530,46 @@ impl Backend {
...
+            );
+            config_store = Some(
+                ConfigStoreBuilder::from_oxlintrc(true, oxlintrc.clone().unwrap())
+                    .expect("failed to build config")
+                    .build()
+                    .expect("failed to build config"),
+            );
+        }
+
+        if self.options.lock().await.use_nested_configs {
```
nrayburn-tech, line 550: If using nested configs, use `Linter::new_with_nested_configs` and pass in the nested config files. Otherwise, should be no behavior change. However, I did do some refactoring here to make this easier for me to read through.

editors/vscode/client/extension.ts
```diff
@@ -171,8 +191,10 @@ export async function activate(context: ExtensionContext) {
     synchronize: {
       // Notify the server about file config changes in the workspace
       fileEvents: [
-        workspace.createFileSystemWatcher('**/.oxlint{.json,rc.json}'),
-        workspace.createFileSystemWatcher('**/oxlint{.json,rc.json}'),
+        workspace.createFileSystemWatcher(oxlintConfigFileGlob),
+        ...workspaceConfigPatterns.map((workspaceConfigPattern) => {
+          return workspace.createFileSystemWatcher(workspaceConfigPattern);
+        }),
```
nrayburn-tech, line 197: Watch any .oxlintrc.json file in the workspace.
Watch the file that the user has configured within their configuration.

This stops watching all variations of oxlint config files that were previously watched. Which, I think is safe.

editors/vscode/client/extension.ts
```diff
@@ -242,7 +264,17 @@ export async function activate(context: ExtensionContext) {
...
+
+  const initialConfigFiles =
+    (await Promise.all([oxlintConfigFileGlob, ...workspaceConfigPatterns].map(async globPattern => {
+      return workspace.findFiles(globPattern);
+    }))).flat();
+  await client.sendNotification(DidChangeWatchedFilesNotification.type, {
+    changes: initialConfigFiles.map(file => {
+      return { uri: file.toString(), type: FileChangeType.Created };
+    }),
+  });
```
nrayburn-tech, line 277: On startup, send all found config files to the language server as "created", so that they get added to the nested config map.

crates/oxc_language_server/src/main.rs
nrayburn-tech, line 44: I don't know if this is the correct type. `RwLock<FxHashMap<PathBuf, ConfigStore>>`. Maybe something else should be used, but this is the only one I could figure out how to make work.

crates/oxc_language_server/src/main.rs
```diff
@@ -180,8 +188,38 @@ impl LanguageServer for Backend {
...
 
-    async fn did_change_watched_files(&self, _params: DidChangeWatchedFilesParams) {
+    async fn did_change_watched_files(&self, params: DidChangeWatchedFilesParams) {
         debug!("watched file did change");
+        if self.options.lock().await.use_nested_configs {
+            let mut config_files = FxHashMap::default();
+            let existing_nested_configs = (*self.nested_configs.read().await).clone();
+            for (key, value) in existing_nested_configs {
+                config_files.insert(key.clone(), value.clone());
+            }
```
nrayburn-tech, line 198: I am copying the map here, so I can mutate it later, before setting the value back into `self.nested_configs`. There might be a better way to do this that I don't know about.

crates/oxc_language_server/src/main.rs
```diff
@@ -492,21 +530,46 @@ impl Backend {
...
+                    .expect("failed to build config"),
+            );
+        }
+
+        if self.options.lock().await.use_nested_configs {
+            let mut config_files: FxHashMap<PathBuf, ConfigStore> = FxHashMap::default();
+            let existing_nested_configs = (*self.nested_configs.read().await).clone();
+            for (key, value) in existing_nested_configs {
+                config_files.insert(key.clone(), value.clone());
+            }
```
nrayburn-tech, line 555: Copying the map so that I can pass it into `Linter::new_with_nested_configs`. Might be a better way to do the copy that I'm not aware of.

crates/oxc_linter/src/config/config_store.rs
```diff
@@ -29,7 +29,7 @@ struct Config {
 }
 
 /// Resolves a lint configuration for a given file, by applying overrides based on the file's path.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
nrayburn-tech, line 32: Needed so that I can clone my map that contains this struct.

nrayburn-tech:
@camchenry can you review this to make sure the logic lines up with how the CLI works?

nrayburn-tech:
@Sysix can you make sure this functionality makes sense to you from the language server perspective? (And any feedback with how the code is actually written would be appreciated too.)

crates/oxc_language_server/src/main.rs
camchenry: Seems ok to me, but I'm not an expert in the async data structures area 😄 Might be worth looking at https://docs.rs/papaya/latest/papaya/, as I know we have been using that recently?

crates/oxc_language_server/src/main.rs
```diff
@@ -180,8 +188,38 @@ impl LanguageServer for Backend {
...
     }
 
-    async fn did_change_watched_files(&self, _params: DidChangeWatchedFilesParams) {
+    async fn did_change_watched_files(&self, params: DidChangeWatchedFilesParams) {
         debug!("watched file did change");
+        if self.options.lock().await.use_nested_configs {
+            let mut config_files = FxHashMap::default();
+            let existing_nested_configs = (*self.nested_configs.read().await).clone();
+            for (key, value) in existing_nested_configs {
+                config_files.insert(key.clone(), value.clone());
```
camchenry, line 197: instead of inserting one-by-one, maybe we can try extending the hash map all at once? this should be more efficient because it allocates the needed memory upfront: https://doc.rust-lang.org/nightly/std/collections/struct.HashMap.html#impl-Extend%3C(K,+V)%3E-for-HashMap%3CK,+V,+S%3E

crates/oxc_language_server/src/main.rs
camchenry: Might be worth looking at https://doc.rust-lang.org/std/mem/fn.take.html. It still requires allocating an empty hash map, but that is much cheaper compared to cloning a full hash map. Then after you've taken the hash map, you have full ownership over it and can insert into as needed.

crates/oxc_language_server/src/main.rs
```diff
@@ -180,8 +188,38 @@ impl LanguageServer for Backend {
...
+            }
+
+            params.changes.iter().for_each(|x| {
+                if x.uri.to_file_path().unwrap().file_name().unwrap() != OXC_CONFIG_FILE {
+                    return;
+                }
+                // spellchecker:off -- "typ" is accurate
+                if x.typ == FileChangeType::CREATED || x.typ == FileChangeType::CHANGED {
+                    // spellchecker:on
+                    let file_path = x.uri.to_file_path().unwrap();
```
camchenry, line 207: we could avoid a duplicate unwrap here and some work by saving `to_file_path()`:

```suggestion
                let Some(file_path) = x.uri.to_file_path() else {
                    return;
                };
                if file_path.file_name().unwrap() != OXC_CONFIG_FILE {
                    return;
                }
                // spellchecker:off -- "typ" is accurate
                if x.typ == FileChangeType::CREATED || x.typ == FileChangeType::CHANGED {
                    // spellchecker:on
```

crates/oxc_language_server/src/main.rs
camchenry: maybe std::mem::take? https://doc.rust-lang.org/std/mem/fn.take.html but not sure if that works in async context like this

camchenry (commented):
I think this all looks right so far. I guess one advantage of having the actual config resolution in just oxc_linter is that it should "just work". As long as the config files are found correctly and passed in, the config files should just get resolved as expected, same as in the CLI (hopefully).

nrayburn-tech:
Closing this as it has been replaced with smaller PRs along with some of the feedback from here addressed.

https://github.com/oxc-project/oxc/pull/9731 - Only watch relevant config files
https://github.com/oxc-project/oxc/pull/9739 - Language server nested config support.
TBD - VS Code send all config files to the Language server on extension startup.

PR #9743, opened 2025-03-13
feat(editor): Support nested configs

nrayburn-tech:
Based off of https://github.com/oxc-project/oxc/pull/9739.

### DIFF
--- a/crates/oxc_language_server/src/main.rs
+++ b/crates/oxc_language_server/src/main.rs
@@ -543,15 +543,15 @@ impl Backend {
         let Ok(root_path) = uri.to_file_path() else {
             return None;
         };
-        let mut config_path = None;
-        let config = root_path.join(self.options.lock().await.get_config_path().unwrap());
-        if config.exists() {
-            config_path = Some(config);
-        }
+        let relative_config_path = self.options.lock().await.get_config_path();
+        let oxlintrc = if relative_config_path.is_some() {
+            let config = root_path.join(relative_config_path.unwrap());
+            config.try_exists().expect("Invalid config file path");
+            Oxlintrc::from_file(&config).expect("Failed to initialize oxlintrc config")
+        } else {
+            Oxlintrc::default()
+        };
 
-        let config_path = config_path?;
-        let oxlintrc = Oxlintrc::from_file(&config_path)
-            .expect("should have initialized linter with new options");
         let config_store = ConfigStoreBuilder::from_oxlintrc(true, oxlintrc.clone())
             .expect("failed to build config")
             .build()
--- a/editors/vscode/client/extension.ts
+++ b/editors/vscode/client/extension.ts
@@ -7,11 +7,18 @@ import {
   StatusBarAlignment,
   StatusBarItem,
   ThemeColor,
+  Uri,
   window,
   workspace,
 } from 'vscode';
 
-import { ExecuteCommandRequest, MessageType, ShowMessageNotification } from 'vscode-languageclient';
+import {
+  DidChangeWatchedFilesNotification,
+  ExecuteCommandRequest,
+  FileChangeType,
+  MessageType,
+  ShowMessageNotification,
+} from 'vscode-languageclient';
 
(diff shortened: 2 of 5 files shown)


editors/vscode/client/extension.ts
```diff
@@ -251,3 +277,34 @@ export function deactivate(): Thenable<void> | undefined {
...
+
+  return [
+    workspace.createFileSystemWatcher(`**/${oxlintConfigFileName}`),
+    ...workspaceConfigPatterns.map((pattern) => {
+      return workspace.createFileSystemWatcher(pattern);
+    }),
+  ];
+}
+
+async function findAllConfigFiles(configRelativePath: string) {
```
Sysix, line 294: is this really needed? 
I expect `workspace.createFileSystemWatcher("**/${oxlintConfigFileName}")` to find all configs and send the `workspace/didChangeWatchedFiles` to the server

nrayburn-tech: It's only needed with a user supplied path, such as `oxlint.json` or whatever it may be that isn't using the standard `.oxlintrc.json`.

Sysix: `oxlint` does not support custom filenames for nested configuration, see PR https://github.com/oxc-project/oxc/pull/9153

nrayburn-tech: With how this is used, I don't think we need to find the user supplied config files.

We do still need to be able to find the `.oxlintrc.json` files and send those over on language server startup. `workspace.createFileSystemWatcher` only watches for events, it doesn't trigger a create event on startup for files that were already there before the watcher was registered.

nrayburn-tech: Updated to only find the one config file name. Resolving this, but please re-open if it needs more discussion.

crates/oxc_language_server/src/main.rs
```diff
@@ -487,26 +552,40 @@ impl Backend {
...
-                .expect("failed to build config");
-            *linter = ServerLinter::new_with_linter(
-                Linter::new(LintOptions::default(), config_store).with_fix(FixKind::SafeFix),
-            );
-            return Some(config);
+        let oxlintrc;
+        let relative_config_path = self.options.lock().await.get_config_path();
+        if let Some(relative_config_path) = relative_config_path {
+            let config = root_path.join(relative_config_path);
+            config.try_exists().expect("Invalid config file path");
```
nrayburn-tech, line 559: I think this is a change behavior, not sure if it needs to be called out or not.

Previous behavior would just return `None` if the config file didn't exist. However, now this expects that if you supply a config file name then it should exist.

If you're using nested configs, you should supply `''`.

nrayburn-tech:
Marking as draft until I can accommodate changes from https://github.com/oxc-project/oxc/pull/9760.

Sysix:
I am sorry, I had not time this weekend to check if this works as expected :) 
Will try to look into it the next days

nrayburn-tech:
No rush. I was looking at implementing this for IntelliJ and it has me second guessing my approach.

Should the client be sending the initial configs that are found OR should the language server search the root_uri for this instead? I'm not sure how much filesystem access should be done from the language server, which is why I did it this way originally. Maybe it's a non-issue though.

Sysix (approved):
> Should the client be sending the initial configs that are found OR should the language server search the root_uri for this instead? 

My first idea was to do all the work on the server side, but after reading about the LSP Specification:

> Servers are allowed to run their own file system watching mechanism and not rely on clients to provide file system events. However this is not recommended due to the following reasons:
> - to our experience getting file system watching on disk right is challenging, especially if it needs to be supported across multiple OSes.
> - file system watching is not for free especially if the implementation uses some sort of polling and keeps a file system tree in memory to compare time stamps (as for example some node modules do)
> - a client usually starts more than one server. If every server runs its own file system watching it can become a CPU or memory problem.
> - in general there are more server than client implementations. So this problem is better solved on the client side.

_https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles_

Tested it today and it worked  🥳

PR #9656, opened 2025-03-10
feat(allocator): replace allocator_ap2's Vec with Vec2

Dunqing:
Just replace allocator_ap2's Vec with Vec2, and make some changes to make it compile successfully.

PR #9792, opened 2025-03-15
perf(allocator/vec2): optimize reserving memory

Dunqing:
resolve: https://github.com/oxc-project/oxc/pull/9656#issuecomment-2723678880

#9656 brought a small performance improvement for the transformer but also led the parser to 1% performance hits. This PR returns performance by splitting `reserve_internal` to `reserve_exact_internal` and `reserve_amortized_internal`  respectively the internal implementation of `reserve_exact` and `reserve`.

Why the change can improve performance?

The original `reserve_internal` implementation has a check for reserve strategy, https://github.com/oxc-project/oxc/blob/fef680a4775559805e99622fb5aa6155cdf47034/crates/oxc_allocator/src/vec2/raw_vec.rs#L664-L668 which is can be avoided because the caller of `reserve_internal` already knows the reserve strategy. After the change, the `reserve_exact` and `reserve` can call the corresponding internal implementation directly, which can avoid unnecessary checks.

Likewise, the `Fallibility` check can also be avoided, https://github.com/oxc-project/oxc/blob/fef680a4775559805e99622fb5aa6155cdf47034/crates/oxc_allocator/src/vec2/raw_vec.rs#L681-L683

because we know where the errors should be handled. 

~~Due to this change, I also replaced Bumpalo's `CollecitonAllocErr` with allocator-api2's `TryReserveError` because `CollecitonAllocErr::AllocErr` cannot pass in a `Layout`.~~ I ended up reverting https://github.com/oxc-project/oxc/pull/9792/commits/937c61a65815f84d6998764e77b1aabac4a8dc71 as it caused transformer performance 1%-2% regression (See [codspeed](https://codspeed.io/oxc-project/oxc/branches/03-15-pref_allocator_vec2_optimize_reserving_memory) and switch to "replace CollectionAllocErr with TryReserveError" commit), and replaced by https://github.com/oxc-project/oxc/pull/9792/commits/84edacdcc131e1891a02e9f5862cbf52da71a355 I've tried various way to save the performance but it not work. I suspect the cause is that `TryReserveError` is 16 bytes whereas `CollecitonAllocErr` is only 1 byte.

So, after both checks are removed, the performance returns to the original. The whole change is according to standard `RawVec`'s implementation. See https://doc.rust-lang.org/src/alloc/raw_vec.rs.html

image

### DIFF
--- a/crates/oxc_allocator/src/vec2/raw_vec.rs
+++ b/crates/oxc_allocator/src/vec2/raw_vec.rs
@@ -318,7 +318,11 @@ impl<'a, T> RawVec<'a, T> {
         used_cap: usize,
         needed_extra_cap: usize,
     ) -> Result<(), CollectionAllocErr> {
-        self.fallible_reserve_internal(used_cap, needed_extra_cap, Exact)
+        if self.needs_to_grow(used_cap, needed_extra_cap) {
+            self.reserve_exact_internal(used_cap, needed_extra_cap)?
+        }
+
+        Ok(())
     }
 
     /// Ensures that the buffer contains at least enough space to hold
@@ -342,7 +346,9 @@ impl<'a, T> RawVec<'a, T> {
     ///
     /// Aborts on OOM
     pub fn reserve_exact(&mut self, used_cap: usize, needed_extra_cap: usize) {
-        self.infallible_reserve_internal(used_cap, needed_extra_cap, Exact)
+        if let Err(err) = self.try_reserve_exact(used_cap, needed_extra_cap) {
+            handle_error(err)
+        }
     }
 
     /// Calculates the buffer's new size given that it'll hold `used_cap +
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#9792** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9792?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9792).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):
Excellent!

overlookmotel:
### Merge activity

* **Mar 17, 8:25 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #9856, opened 2025-03-18
perf(allocator/vec2): replace `self.reserve(1)` calls with `self.grow_one()` for better efficiency

Dunqing:
The places calling `self.reserve(1)` already checked `len == self.buf.cap()`, in order to avoid checking again in `reserve`, we should call `grow_one` instead which is a shortcut of `grow_amortized(len, 1)`.

### DIFF
--- a/crates/oxc_allocator/src/vec2/mod.rs
+++ b/crates/oxc_allocator/src/vec2/mod.rs
@@ -1237,7 +1237,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
 
         // space for the new element
         if len == self.buf.cap() {
-            self.reserve(1);
+            self.buf.grow_one();
         }
 
         unsafe {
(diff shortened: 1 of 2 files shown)


Dunqing:
Unbelievable 😢
image

Dunqing:
I've noticed the current `extend` implementation will call the `push` method, `extend` is quite a common method for our scenario, so I think `#[inline(never)]` is a deoptimization. I decided to get #9881 done first and then continue this.

overlookmotel (approved):

overlookmotel:
### Merge activity

* **May 3, 10:33 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #9870, opened 2025-03-18
perf(isolated_declarations): replace `HashMap` of internal annotation spans with `Vec`

overlookmotel:
Statements, class elements etc which have an `@internal` comment need to be excluded from output if `strip_internal` option is enabled.

Previously we checked whether AST nodes have a comment by storing start offsets of AST nodes which have an `@internal` comment in a hash map.

But we traverse the AST in order that AST nodes appear in source, and therefore we'll always be querying spans against the lookup table in ascending order. So we can replace the hash map with a `Vec` and move through it in order. Those lookups are much faster.

Additionally, `program.comments` is also in source text order, so we can rely on that property when constructing the lookup table.

### DIFF
--- a/crates/oxc_isolated_declarations/src/internal_annotations.rs
+++ b/crates/oxc_isolated_declarations/src/internal_annotations.rs
@@ -53,20 +53,15 @@ impl InternalAnnotations {
     pub fn build(&mut self, program: &Program) {
         // `u32::MAX` cannot be the start of any node, because source text is limited to `u32::MAX` bytes
         let mut last_span_start = u32::MAX;
-        self.span_starts = program
-            .comments
-            .iter()
-            .filter_map(|comment| {
-                let has_internal =
-                    comment.content_span().source_text(program.source_text).contains("@internal");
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#9870** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9870?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9870).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
I'm seeing +3% speed-up locally. But CodSpeed only says +1%.

PR #9879, opened 2025-03-19
feat(parser): Add regex cargo feature to oxc_parser

### LINKED ISSUE #9839: parser: feature gate `oxc_regular_expression`
To help with binary size in places where it's not needed.

r58Playz:
Fixes #9839.

### DIFF
--- a/crates/oxc_parser/Cargo.toml
+++ b/crates/oxc_parser/Cargo.toml
@@ -23,7 +23,7 @@ oxc_allocator = { workspace = true }
 oxc_ast = { workspace = true }
 oxc_diagnostics = { workspace = true }
 oxc_ecmascript = { workspace = true }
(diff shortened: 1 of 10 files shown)


crates/oxc_parser/Cargo.toml
```diff
@@ -45,3 +45,5 @@ pico-args = { workspace = true }
 [features]
```
Boshen, line 45: I guess we need to add a `default` feature with `regex` turned on by default.

r58Playz: Doing this makes it hard to disable the `oxc_parser/regex` feature in the `oxc` umbrella crate, so that `oxc/regex` can control it instead. Using `{ workspace = true, default-features = false }` in `oxc` errors.

PR #9752, opened 2025-03-13
feat(allocator/vec2): align the `retain` method with the standard implementation

Dunqing:
The documentation and the implementation are copied from https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2095-209.

~~After this change, `drain_filter` is no longer needed, so remove it.~~

PR #9742, opened 2025-03-13
feat(allocator/vec): remove `ManuallyDrop` wrapper

Dunqing:
Partially revert https://github.com/oxc-project/oxc/pull/6623, It still keeps the check for non-`drop` element of `Vec`

PR #9679, opened 2025-03-11
feat(allocator): remove drop operations from Vec2

Dunqing:
Just like #6623. Since we have control over the `Vec` implementation, we can remove the `drop` operations and eventually revert the changes made in #6623 directly.

PR #9655, opened 2025-03-10
feat(allocator): add `Vec2::retain_mut` method

Dunqing:
OXC has a few places using this API, so we need to add this method before replacing allocator-api2's `Vec`. The implementation is copied from the https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2121-2123, and modified a little to make it fit the `Vec2`.

PR #9881, opened 2025-03-19
feat(allocator/vec2): implement `extend_desugared` method as `extend` internal implemention

Dunqing:


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9881?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#9881** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9881?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9752** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9752?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#9858](https://github.com/oxc-project/oxc/pull/9858) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9858?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#9742** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9679** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9679?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9656** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9656?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9655** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9655?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9647** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9730** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9729** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9728** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9728?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9881).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -2271,11 +2286,81 @@ impl<'a, 'bump, T> IntoIterator for &'a mut Vec<'bump, T> {
...
+        // 2. `T` is always non-`Copy` and non-`Clone` in our use case, so we don't need a special
+        //   implementation for `T: Clone` or `T: Copy`.
         let iter = iter.into_iter();
-        self.reserve(iter.size_hint().0);
+        let (_, high) = iter.size_hint();
+        if high.is_some() {
+            self.extend_trusted(iter);
+        } else {
+            self.extend_desugared(iter);
+        }
```
overlookmotel: Calling `extend_trusted` here is unsound, because the iterator *isn't* trusted.

Here's an example which writes out of bounds of the `Vec` (and will continue overwriting all memory with `123` over and over until the process crashes):

```rs
struct MyIter;

impl Iterator for MyIter {
    type Item = u64; // Or whatever

        // Yield `123` endlessly forever
        Some(123)
    }

    fn size_hint(&self) -> (usize, Option<usize>) {
        // Report that no more values to yield (which is a lie!)
        (0, Some(0))
    }
}

vec.extend(MyIter);
```

The point of `TrustedLen` trait is that it indicates that the values returned by `size_hint` are definitely accurate. The only iterators that implement `TrustedLen` are those in std lib where the implementation is guaranteed to uphold that promise.

But it's legal for any other `Iterator` to provide inaccurate `size_hint`. That might be considered a bug, but it's not a violation of safety guarantees. So my terrible iterator above is stupid, but not illegal. `Vec::extend` must handle it without causing UB.

Dunqing: Yes, we may need to do a runtime check for this but I don't know if it is worth it. I am trying to make all benchmarks are positive first and then come back to handle this

overlookmotel: Unfortunately there is no runtime check available. How do you determine whether an `Iterator` is e.g. a [slice iterator](https://doc.rust-lang.org/std/slice/struct.Iter.html) which does have a trusted length, or another kind of iterator which doesn't?

We can get the optimization for the most common cases by using `Vec::extend_from_slice`, `Vec::extend_from_array` or `Vec::append` in calling code, instead of `Vec::extend`. Slices, arrays, and `Vec`s all have a fixed length which you can trust to be reported correctly, so they can safely use `extend_trusted`.

But due to the lack of [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html) in stable Rust, what we *can't* do is what std lib does - choose whether to use `extend_trusted` or not in `extend`, depending on the type of the iterator.

---

By the way, I found the docs that specify that `size_hint` cannot be trusted: https://doc.rust-lang.org/std/iter/trait.Iterator.html#implementation-notes

---

If we want to get fancy, we could introduce our own `TrustedLen` trait and implement it on iterators which are known to have a trusted length. Then we can expose a method `Vec::extend_trusted`:

```rs
unsafe trait TrustedLen {}

impl Vec<T> {
    pub fn extend_trusted<I: IntoIterator<Item = T> + TrustedLen>(&mut self, iter: I) { /* ... */ }
}
```

That could be fun to do! But probably `extend_from_slice`, `extend_from_array` and `append` will get us most of the perf benefit, without the complication.

Dunqing: Looks good! 

> That could be fun to do! But probably extend_from_slice, extend_from_array and append will get us most of the perf benefit, without the complication.

Yes, I agree!

I added a `vec_from_iter_with_capacity` API in this PR to optimize the usage of `from_iter_in` with `iter.filter_map` because filter_map's `size_hint` doesn't have a relatively exact size, to avoid calling extra `reserve` in-middle of iter, we can pass a capacity in to allocate enough space Vec for the iter. Just an experiment, I am glad to hear your opinion on this case.

PR #9895, opened 2025-03-19
perf(ast/estree): speed up raw deser for `JSXElement`

overlookmotel:
Follow-on after #9863. Speed up the raw deserializer code for `JSXElement` that PR introduced. Using `delete` in JS is usually a deoptimization. Better to create a new object.


overlookmotel:
Makes no difference to output, and should be a perf improvement (though we don't have benchmarks to answer that one), so merging without review.

overlookmotel:
### Merge activity

* **Mar 19, 1:04 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #9904, opened 2025-03-20
chore(syntax,napi/playground): enable to list all supported es targets

wangziling:
Maybe the playground [https://playground.oxc.rs/](https://playground.oxc.rs/) should list out all supported es targets.

image

Just as the ts playground did.

image

### DIFF
--- a/crates/oxc_syntax/src/es_target.rs
+++ b/crates/oxc_syntax/src/es_target.rs
@@ -66,3 +66,22 @@ impl fmt::Display for ESTarget {
         write!(f, "{s}",)
     }
 }
+
+/// List all supported ECMAScript targets.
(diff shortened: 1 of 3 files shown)


Boshen:
I think you can hard code this in the playground. https://github.com/oxc-project/playground

FYI our target also support esbuild targets, https://esbuild.github.io/api/#target

hence the textfield.

wangziling:
> I think you can hard code this in the playground. https://github.com/oxc-project/playground
> 
> FYI our target also support esbuild targets, https://esbuild.github.io/api/#target
> 
> hence the textfield.

Okay, I got it~

Let me close this PR, thanks for your interpretation~ 😉

PR #9920, opened 2025-03-20
fix(napi/playground): fix the issue that missing `reset` the data before re-running the compilation process

wangziling:
This PR is aimed to fix a slight issue, it didn't reset all the related data before re-running the wasm compilation process.

The repro steps:
1. Visit https://playground.oxc.rs
2. Select `Symol` or `Scope` tab on the right side of the page.
3. Check bettwen the `D.TS` and `TSX` boxes.
4. The contents under the `Symbol` and `Scope` tabs will not change as the checkboxes' changes.

gif

And the 3rd step can trigger as this way:
Edit the json, make the `run.symbol` or `run.scope` as `false`.

gif1

### DIFF
--- a/napi/playground/src/lib.rs
+++ b/napi/playground/src/lib.rs
@@ -75,7 +75,7 @@ impl Oxc {
     #[napi]
     #[allow(clippy::allow_attributes, clippy::needless_pass_by_value)]
     pub fn run(&mut self, source_text: String, options: OxcOptions) -> napi::Result<()> {
-        self.diagnostics = vec![];
+        self.reset();
 
         let OxcOptions {
             run: run_options,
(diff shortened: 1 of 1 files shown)


napi/playground/src/lib.rs
```diff
@@ -299,6 +301,10 @@ impl Oxc {
         }
     }
 
+    fn reset(&self) {
```
Boshen, line 304: remove?

wangziling:
Firstly I reset all the properties of the `Oxc` instance.

But I found that we need to keep the previous `ast` and `ast_json` values, otherwise the `EsTree` tab will show nothing in the `Program`, but the previous result should be accepted.

PR #9926, opened 2025-03-20
perf(lexer): pad `Token` to 16 bytes

overlookmotel:
#9918 caused a small regression on lexer and parser benchmarks because it introduced an uninitialized padding byte into `Token`.

Add another dummy `u8` to fill the uninitialized hole.

### DIFF
--- a/crates/oxc_parser/src/lexer/token.rs
+++ b/crates/oxc_parser/src/lexer/token.rs
@@ -39,6 +39,7 @@ pub struct Token {
     // Padding to fill to 16 bytes.
     // This makes copying a `Token` 1 x xmmword load & store, rather than 1 x dword + 1 x qword
     // and `Token::default()` is 1 x xmmword store, rather than 1 x dword + 1 x qword.
+    _padding: u8,
     _padding2: u16,
 }
 


overlookmotel:
* **#9926** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9926?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9926).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Boshen:
I'm not seeing any difference, both are `size = 16 (0x10), align = 0x4`. What are property do I need to inspect? Can this be a compile time check?

We already have this

```
#[cfg(test)]
mod size_asserts {
    use super::Token;
    const _: () = assert!(std::mem::size_of::<Token>() == 16);
}
```

why is this not enough?

overlookmotel:
Unfortunately, Rust offers no tools to detect whether a type contains padding bytes. So I don't think it's possible to implement a test.

This PR does not affect type size or alignment, only whether any of the bytes representing the type are uninitialized. When they are, creating and copying a `Token` is more instructions, as compiler avoids ever reading or writing the uninitialized byte. So e.g. copying a `Token` is 2 x 8-byte reads + 2 x 8-byte writes (overlapping). Whereas with the `_padding: u8, _padding2: u16` fields added, there are no unitialized bytes, do copying `Token` is just 1 x 16-byte read, and 1 x 16-byte write.

This is a micro-micro-optimization, but creating and copying `Token`s is such an extremely hot path that it makes a measureable difference. #3283 was the one that revealed this - 5% speed-up just due to the fields of `Token` getting re-arranged.

But this PR isn't succeeding in fixing the perf loss. Looks like changing the `u32` field to a `u16` has caused the other fields to move around, which is also a regression. I'll try to fix it.

PR #9954, opened 2025-03-21
fix(ast/estree): Fix `FormalParameter`

therewillbecode:
Align our AST's ESTree output with that of TS-ESLint's for the `FormalParameter` node.  Relates to #9705 

Skips the estree serialization of `accessibility` `readonly` and `overrides` fields.


overlookmotel (approved):
Bravo!

PR #9959, opened 2025-03-21
fix(ast/estree): Fix `TSFunctionType` and `TSCallSignatureDeclaration` 

therewillbecode:
Skips estree serialisation for the field named `this_param` for both `TSFunctionType` and `TSCallSignatureDeclaration` nodes.

PR is part of ongoing work to align our AST's ESTree output with that of TS-ESLint's. Relates to #9705.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -1319,6 +1319,7 @@ pub struct TSFunctionType<'a> {
     /// type T = (this: string, a: number) => void;
     /// //        ^^^^^^^^^^^^
     /// ```
+    #[estree(skip)]
     pub this_param: Option<Box<'a, TSThisParameter<'a>>>,
     /// Function parameters. Akin to [`Function::params`].
     pub params: Box<'a, FormalParameters<'a>>,
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -975,6 +975,7 @@ pub struct TSIndexSignature<'a> {
 pub struct TSCallSignatureDeclaration<'a> {
     pub span: Span,
     pub type_parameters: Option<Box<'a, TSTypeParameterDeclaration<'a>>>,
+    #[estree(skip)]
     pub this_param: Option<TSThisParameter<'a>>,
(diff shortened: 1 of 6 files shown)


overlookmotel (approved):
👍

I'd guess (but don't know) that the `this_param` field isn't ignored entirely. It probably gets added to the start of `params`, same as in `Function`:

https://github.com/oxc-project/oxc/blob/0cdeedd19dab4b7c29793efd5b5468742a878384/crates/oxc_ast/src/ast/js.rs#L1701-L1705

`#[estree(via = FunctionFormalParameters)]` refers to this custom serializer (a bit tricky):

https://github.com/oxc-project/oxc/blob/0cdeedd19dab4b7c29793efd5b5468742a878384/crates/oxc_ast/src/serialize.rs#L384-L421

But maybe better to catch more of the "low hanging fruit" first, and come back to these more complex ones.

overlookmotel:
Oh damn it! Merge conflicts (probably conflicts with your other PR which I just merged).

If you can rebase on main and run `just test-estree` again, that should sort it.

overlookmotel:
I fixed the conflicts. Will merge as soon as CI goes green.

therewillbecode:
> 👍
> 
> I'd guess (but don't know) that the `this_param` field isn't ignored entirely. It probably gets added to the start of `params`, same as in `Function`:
> 
> https://github.com/oxc-project/oxc/blob/0cdeedd19dab4b7c29793efd5b5468742a878384/crates/oxc_ast/src/ast/js.rs#L1701-L1705
> 
> `#[estree(via = FunctionFormalParameters)]` refers to this custom serializer (a bit tricky):
> 
> https://github.com/oxc-project/oxc/blob/0cdeedd19dab4b7c29793efd5b5468742a878384/crates/oxc_ast/src/serialize.rs#L384-L421
> 
> But maybe better to catch more of the "low hanging fruit" first, and come back to these more complex ones.

Ah so that is what `via` does it references the newtype wrapper which implements the custom serialization logic. So I guess that is how we take case of dynamically serialising a field on a node or not depending on what its parent looks like.

therewillbecode:
Thanks for the review and fixing the conflicts @overlookmotel !

PR #9967, opened 2025-03-22
fix(ast/estree): Add decorators field to `AssignmentPattern`

therewillbecode:
Adds a `decorators` field with an empty array to align our AST's ESTree output with that of TS-ESLint's for the `AssignmentPattern` node. Relates to #9705


hi-ogawa (approved):
Thanks!

PR #9968, opened 2025-03-22
fix(ast/estree): Add missing estree fields to `TSIndexSignature` and `TSIndexSignatureName`

therewillbecode:
Add missing estree fields to TSIndexSignature and TSIndexSignatureName

PR is part of ongoing work to align our AST's ESTree output with that of TS-ESLint's. Relates to #9705.


hi-ogawa (approved):
I wonder when `TSIndexSignature.accessibility` becomes non null. Maybe something legacy  and tseslint doens't use it? :thinking: https://github.com/typescript-eslint/typescript-eslint/blob/84744742c79cf3c1a81cbb08fd2cbaf75b45afc7/packages/ast-spec/src/element/TSIndexSignature/spec.ts#L9

Either way, I think we can proceed and revisit if we spot something different in the future. Thanks!

PR #9971, opened 2025-03-22
feat(linter): add unicorn/no-accessor-recursion rule

huangtiandi1999:
Relates to #684 
Rule detail：https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v57.0.0/docs/rules/no-accessor-recursion.md

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
@@ -50,7 +50,7 @@ declare_oxc_lint!(
     /// ```
     NoAccessorRecursion,
     unicorn,
-    correctness,
+    suspicious,
 );
 
 impl Rule for NoAccessorRecursion {
--- a/crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
+++ b/crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
@@ -1,15 +1,18 @@
 use oxc_ast::{
     AstKind,
-    ast::{Expression, MemberExpression, MethodDefinitionKind, PropertyKind},
+    ast::{
+        BindingPatternKind, Expression, MemberExpression, MethodDefinition, MethodDefinitionKind,
+        ObjectProperty, PropertyKey, PropertyKind,
+    },
 };
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
-use oxc_span::Span;
+use oxc_span::{GetSpan, Span};
 
 use crate::{AstNode, context::LintContext, rule::Rule};
 
-fn no_accessor_recursion_diagnostic(span: Span) -> OxcDiagnostic {
-    OxcDiagnostic::warn("Disallow recursive access to `this` within getters and setters.")
+fn no_accessor_recursion_diagnostic(span: Span, kind: &str) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!("Disallow recursive access to `this` within {kind}."))
         .with_label(span)
 }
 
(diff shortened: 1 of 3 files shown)


huangtiandi1999:
hi @Sysix, this job is finish, please help to review

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,581 @@
...
+    /// ```
+    NoAccessorRecursion,
+    unicorn,
+    suspicious,
+);
+
+impl Rule for NoAccessorRecursion {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::VariableDeclarator(decl) => {
```
Sysix, line 59: VariableDeclarator and MemberExpression are the core structure of any JS program.
Maybe we can get more performance when we check for `ThisExpression` and its possible valid parent.
What do you think @camchenry

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,581 @@
...
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{GetSpan, Span};
+
+use crate::{AstNode, ast_util::nth_outermost_paren_parent, context::LintContext, rule::Rule};
+
+fn no_accessor_recursion_diagnostic(span: Span, kind: &str) -> OxcDiagnostic {
+    OxcDiagnostic::warn(format!("Disallow recursive access to `this` within {kind}."))
+        .with_label(span)
```
camchenry, line 16: Every rule should have at least a short help message, to help users understand how to fix it the issue. How about something like this:
```suggestion
    let method_kind = match kind {
      "setters" => "set",
      _ => "get",
    };
    OxcDiagnostic::warn(format!("Disallow recursive access to `this` within {kind}."))
        .with_help("Remove this property access, or remove `{method_kind}` from the method")
        .with_label(span)
```

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
camchenry: I think it's a good idea. @huangtiandi1999 to put it more concretely: try restructuring the rule so that it only searches for `AstKind::ThisExpression`. Then, look up/down the tree as needed to match the cases. It might require a bit of restructuring, but it should be a lot faster, as we'll be able to skip evaluating lots of nodes that are common like @Sysix pointed out.

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,581 @@
...
+        }
+    } else {
+        false
+    }
+}
+
+fn get_closest_function<'a>(node: &AstNode, ctx: &'a LintContext) -> Option<&'a AstNode<'a>> {
+    let mut parent = ctx.nodes().parent_node(node.id())?;
+
+    loop {
```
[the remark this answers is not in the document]
camchenry: if it's possible, I'd even recommend using `ctx.nodes().ancestor_kinds()`. here is an example elsewhere, where we do something similar: https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/no_new_func.rs#L66-L72

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,581 @@
...
+                    }
+                    let Some(key_name) = get_property_or_method_def_name(func, ctx) else {
+                        return;
+                    };
+                    if let BindingPatternKind::ObjectPattern(obj_pattern) = &decl.id.kind {
+                        let exist = obj_pattern
+                            .properties
+                            .iter()
+                            .find(|ident| ident.key.name().is_some_and(|name| name == key_name));
+                        if exist.is_some() {
```
camchenry, line 80:
```suggestion
                        let exists = obj_pattern
                            .properties
                            .iter()
                            .any(|ident| ident.key.name().is_some_and(|name| name == key_name));
                            
                        if exists {
```

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,581 @@
...
+    suspicious,
+);
+
+impl Rule for NoAccessorRecursion {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::VariableDeclarator(decl) => {
+                // Here we deal with deconstructive access to this inside getter
+                // e.g. "const { baz } = this"
+                if let Some(init) = &decl.init {
```
camchenry, line 62: I'd suggest doing an early return here, to remove a level of indentation:
```suggestion
                let Some(init) = &decl.init else {
                    return;
                }
```

huangtiandi1999:
Hi @camchenry @Sysix, all comments are solved, and now that it only searches for `AstKind::ThisExpression`, please help to review.

crates/oxc_linter/src/rules/unicorn/no_accessor_recursion.rs
```diff
@@ -0,0 +1,574 @@
...
+    /// };
+    /// ```
+    NoAccessorRecursion,
+    unicorn,
+    suspicious,
+);
+
+impl Rule for NoAccessorRecursion {
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        if let AstKind::ThisExpression(this_expr) = node.kind() {
```
Sysix, line 65: NIT: you can fast return with `else` and avoid one nesting

Sysix (approved):

Sysix:
Thanks for the hard work 🫶

camchenry (approved):

PR #9973, opened 2025-03-22
feat(linter): Add suggested fix to `typescript/no_unnecessary_parameter_property_assignment` and fix false positive

ulrichstark:
Hey, I just added a suggested fix to the lint rule as continuation of my work in #9720 and #9618.
The original typescript-eslint rule isn't capable of fixing its reported issue, but the fix should be pretty safe.
I decided to configure it as `suggestion` and not as 100% safe `fix`.
Also found one false positive while implementing the fix suggestion, fixed it and added a corresponding passing test.

### DIFF
--- a/crates/oxc_linter/src/rules/typescript/no_unnecessary_parameter_property_assignment.rs
+++ b/crates/oxc_linter/src/rules/typescript/no_unnecessary_parameter_property_assignment.rs
@@ -2,12 +2,13 @@ use oxc_ast::{
     AstKind,
     ast::{
         AssignmentExpression, AssignmentOperator, AssignmentTarget, ClassElement, Expression,
-        FormalParameter, MethodDefinitionKind, Statement,
+        FormalParameter, Function, MethodDefinitionKind, Statement,
     },
 };
 use oxc_ast_visit::Visit;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
+use oxc_semantic::ScopeFlags;
 use oxc_span::{Atom, Span};
 use rustc_hash::FxHashSet;
 
@@ -105,13 +106,17 @@ impl Rule for NoUnnecessaryParameterPropertyAssignment {
             }
         }
 
+        let Some(function_body) = &method.value.body else {
+            return;
+        };
+
         let mut visitor = AssignmentVisitor {
             ctx,
             parameter_properties,
             assigned_before_unnecessary: FxHashSet::default(),
             assigned_before_constructor,
         };
-        visitor.visit_method_definition(method);
+        visitor.visit_function_body(function_body);
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/typescript/no_unnecessary_parameter_property_assignment.rs
```diff
@@ -176,9 +182,15 @@ impl<'a> Visit<'a> for AssignmentVisitor<'a, '_> {
...
-            self.ctx.diagnostic(no_unnecessary_parameter_property_assignment_diagnostic(
-                assignment_expr.span,
-            ));
+            self.ctx.diagnostic_with_suggestion(
+                no_unnecessary_parameter_property_assignment_diagnostic(assignment_expr.span),
+                |fixer| {
+                    fixer.delete_range(Span::new(
+                        assignment_expr.span.start,
+                        assignment_expr.span.end + 1,
+                    ))
```
ulrichstark, line 191: I'm unsure about this.
- Should `.dangerously()` be added to the `RuleFix`?
- Is adding 1 a stable enough solution for a suggested fix to achieve the expected result in most cases?
- How can I get the parent of the assignment expression to maybe improve this fix?

shulaoda: > * Should `.dangerously()` be added to the `RuleFix`?

What is this? I don't quite understand.

shulaoda: > * Is adding 1 a stable enough solution for a suggested fix to achieve the expected result in most cases?

I think that's sufficient. But I'm curious why you chose not to remove the `public` modifier.

shulaoda: > * How can I get the parent of the assignment expression to maybe improve this fix?

It seems difficult to search using the `Visit` approach. Instead, you can obtain the corresponding `NodeId` through the `symbol_id` of `BindingIdentifier`, then traverse upwards to find the `AssignmentExpression` with a matching `span`. Finally, you can retrieve its parent node.

shulaoda (approved):

crates/oxc_linter/src/rules/typescript/no_unnecessary_parameter_property_assignment.rs
ulrichstark: > > * Should `.dangerously()` be added to the `RuleFix`?
> 
> What is this? I don't quite understand.

I mean calling the following method `dangerously` on my `RuleFix` to signal that the rulefix might be unsafe. Docs for the method state "This is especially useful for fixer functions that are safe in some cases but not in others.". Because I only have one possible rulefix and the whole fix is marked as "suggestion", it's probably fine without calling `dangerously` or?
https://github.com/oxc-project/oxc/blob/c3af9a45100ec0a30695b20163af24237661d7c1/crates/oxc_linter/src/fixer/fix.rs#L219

ulrichstark: > > * Is adding 1 a stable enough solution for a suggested fix to achieve the expected result in most cases?
> 
> I think that's sufficient. But I'm curious why you chose not to remove the `public` modifier?

Okay great! The `public` or other modifiers aren't the issue. These modifiers turn parameters into TypeScript parameter properties (https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties). And because they are parameter properties with implicit assignment capability, the explicit assignments in the constructor body are unnecessary and can be removed. The goal of this rule is to report and remove these unnecessary explicit assignments of parameter properties.

PR #9977, opened 2025-03-22
refactor(linter): use delete instead of delete_range to shorten calls

ulrichstark:
Some of these shorter calls are only possible because of the also added `impl GetSpan for oxc_allocator::Box`. I decided to put this impl block into `oxc_span/src/span.rs`, because it feels a bit wrong to add it to the similar impl blocks in `oxc_allocator/src/boxed.rs`, because that would make oxc_allocator depend on oxc_span.


Sysix:
I do not like this change. You are calling `span()` instead of `.span`. 
Second the implementation is only used in oxc_linter, your `delete` code should be there, but I could not find a right place for it.

Boshen:
I agree, `delete_range(x.span)` is explicit and more readable.

PR #9990, opened 2025-03-23
refactor(language_server)!: remove `enable` configuration, the client should shutdown the server instead

Sysix:


Sysix:
* **#9990** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9990?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/9990?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=9990).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

editors/vscode/client/extension.ts
```diff
@@ -265,8 +279,17 @@ export async function activate(context: ExtensionContext) {
...
 
+  if (configService.config.enable) {
+    await startClient();
+  }
+}
+
+// Starts the client, it does not check if it is already started
+async function startClient() {
+  await client.start();
+  // ToDo: refactor it on the server side.
```
Sysix, line 291: @nrayburn-tech I think we should do the init for all `.oxlintrc.json` files on the server side,  
like you suggested in your nested config PR.

The client is able to start/stop the language server on any time 
and remembering to send the watchedFiles-request is not ideal.

nrayburn-tech: I haven’t been able to do much investigation, but I think there’s a lot of common code that should be between the CLI and the language server that could be shared if there was a common crate made for it. From memory, below are some areas that I think make sense to share. (Note, I still think we need to retain watched files in the language server.)

- Loading the nested configs.
  - Same for CLI and language server for initial load.
  - CLI doesn’t need subsequent loads as it only runs once.
  - Language server needs to retain all configs in memory and listen for changes, applying and modifications on top of the existing config.
- Default ignore patterns and where they are loaded from.
- Reporting diagnostics.
  - CLI has different reporters.
  - Language server just sends the diagnostics to the client.
- An “init” command for creating a config file.
- A way to list rules.

I think if the CLI does it then the language server probably needs to be able to do it too. Obviously some functionality is more important in a language server than others, but I think feature parity would be ideal.

Sysix: Yes writing 2 code bases for CLI and the langauge server feels wierd.  
I hope some rust experts can refactor it in the future ;) 

> Language server needs to retain all configs in memory and listen for changes, applying and modifications on top of the existing config.

I do not agree with this one. FileWatchers should be done on the client side. See the LSP docs for more details:  
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles

> Servers are allowed to run their own file system watching mechanism and not rely on clients to provide file system events. However this is not recommended due to the following reasons:

---

> Default ignore patterns and where they are loaded from.

Should the client really send open/change/save events to the server when it should be ignored?

> A way to list rules.

https://github.com/oxc-project/oxc/blob/c3af9a45100ec0a30695b20163af24237661d7c1/apps/oxlint/src/output_formatter/json.rs#L23-L27

nrayburn-tech: > > Language server needs to retain all configs in memory and listen for changes, applying and modifications on top of the existing config.

> I do not agree with this one. FileWatchers should be done on the client side. See the LSP docs for more details:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles

I didn't mean to imply that the server would do the watching. The server just needs to keep the values in memory and be aware of any changes, so that it can layer the new configs without the client needing to restart the language server. (Config file changes requiring a language server restart isn't ideal.)

> > Default ignore patterns and where they are loaded from.

> Should the client really send open/change/save events to the server when it should be ignored?

Could we have the defaults in the language server and publish them to clients and then the clients watch/update watchers accordingly? In any case, I would consider this a very low priority to resolve (considering it might be difficult).

> > A way to list rules.

I was referring to the ability to do this from an editor. Again, low priority.

The reason for some of the duplicate functionality within the editor and the CLI is to stay within a single workflow and not need to switch in and out of a terminal for something that is only available in the terminal. However, if something is available within the terminal then it's probably low priority to get that same functionality in the editor.
Diagnostics and code actions working without any bugs should definitely be the highest priority in my opinion.

nrayburn-tech (approved):
Went ahead and looked at this even though it's draft, I'm not sure when I would have a chance to get back to it.

Left a few comments on things you might want to review, but nothing big.

editors/vscode/client/extension.ts
```diff
@@ -78,8 +82,18 @@ export async function activate(context: ExtensionContext) {
...
   const toggleEnable = commands.registerCommand(
     OxcCommands.ToggleEnable,
-    () => {
-      configService.config.updateEnable(!configService.config.enable);
+    async () => {
+      await configService.config.updateEnable(!configService.config.enable);
+
+      if (client.isRunning()) {
+        if (!configService.config.enable) {
+          await client.stop();
```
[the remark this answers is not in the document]
Sysix: Oh of course :) Forget that the Diagnostics are owned by the Server

editors/vscode/client/extension.ts
```diff
@@ -232,7 +246,7 @@ export async function activate(context: ExtensionContext) {
 
   configService.onConfigChange = function onConfigChange(event) {
     let settings = this.config.toLanguageServerConfig();
-    updateStatsBar(settings.enable);
+    updateStatsBar(this.config.enable);
```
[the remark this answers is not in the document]
Sysix: Agree, but this is the current implementation. I did not change it. 
Feel free to create another PR to fix this

camchenry (approved):
Thanks for working on this and updating the docs a bit. I'm still learning the language server protocol, so I appreciate any extra comments.

PR #9994, opened 2025-03-23
feat(linter): add suggested fix to `unicorn/prefer-structured-clone`

ulrichstark:
I added the suggested fix to `unicorn/prefer-structured-clone`.
Also fixed the incorrect reported spans, cleaned up the code and added a new fail & fix test with a more complex first argument.


camchenry (approved):
thanks!

PR #2509, opened 2024-02-26
refactor(ast): remove `TSEnumBody`

Boshen:

PR #10017, opened 2025-03-24
fix(ast/estree): Add `TSEnumBody` to `TSEnumDeclaration.body`

leaysgur:
Part of #9705 

This PR fixes AST structure of `TSEnumDeclaration` from:

```js
{
  "type": "TSEnumDeclaration",
  "id": { /* ... */ },
  "members": [
    {
      "type": "TSEnumMember",
      "id": { /* ... */ },
      "initializer": null
    }
  ],
  "const": false,
  "declare": false
}
```

to:

```js
{
  "type": "TSEnumDeclaration",
  "id": { /* ... */ },
  "body": {
    "type": "TSEnumBody",
    "members": [
	    {
	      "type": "TSEnumMember",
          "id": { /* ... */ },
	      "initializer": null
	    }
	  ],
  },
  "const": false,
  "declare": false
}
```

> https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/declaration/TSEnumDeclaration/spec.ts
> https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/special/TSEnumBody/spec.ts

Historically, we had this node in the past, but it seems to have been deleted in #2509 .


leaysgur:
This change causes many tests to pass, but I've just realized that this implementation may not be sufficient...

I believe it might not work for code like this.

```ts
enum Foo        { A }
//       ^^^^^^^
```

I'm not sure how to handle these gaps in serialize step. Do we need to revive the `TSEnumBody` node in the Rust AST?

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -68,9 +68,19 @@ pub struct TSThisParameter<'a> {
...
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
+#[estree(
+    add_ts_def = "
+        interface TSEnumBody extends Span {
+            type: 'TSEnumBody';
+            members: TSEnumMember[];
+        }
+    ",
+    add_fields(body = TSEnumDeclarationBody)
```
overlookmotel, line 78: Thanks for getting involved!

I think you can achieve the same thing in a slightly cleaner way by:

* Removing `add_fields(body = TSEnumDeclarationBody)`.
* Removing `#[estree(skip)]` on `members` field.
* Instead, add this before `members` field:

```rs
#[scope(enter_before)]
#[estree(rename = "body", via = TSEnumDeclarationBody)]
pub members: Vec<'a, TSEnumMember<'a>>,
```

You shouldn't need to change the implementation of `TSEnumDeclarationBody` converter - it should work exactly as before.

crates/oxc_ast/src/serialize.rs
```diff
@@ -728,6 +728,40 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
...
+///          |     | TSEnumBody.span
+///        ^^ id_end + 1 for space
+/// ```
+#[ast_meta]
+#[estree(
+    ts_type = "TSEnumBody",
+    raw_deser = "
+        const tsEnumDeclId = DESER[BindingIdentifier](POS_OFFSET.id);
+        const tsEnumDeclMembers = DESER[Vec<TSEnumMember>](POS_OFFSET.members);
+        const bodyStart = tsEnumDeclId.end + 1;
```
overlookmotel, line 747: I think this will work, and is more efficient as it only deserializes `id` field once instead of twice:

```suggestion
        const tsEnumDeclMembers = DESER[Vec<TSEnumMember>](POS_OFFSET.members);
        const bodyStart = THIS.id.end + 1;
```

overlookmotel:
> This change causes many tests to pass, but I've just realized that this implementation may not be sufficient...
> 
> I believe it might not work for code like this.
> 
> ```ts
> enum Foo        { A }
> //       ^^^^^^^
> ```
> 
> I'm not sure how to handle these gaps in serialize step. Do we need to revive the `TSEnumBody` node in the Rust AST?

Yes, we'll probably need to alter the Rust AST so it includes the `Span` we need.

But this PR already makes more tests pass, so let's get it merged, and we can then improve on it.

crates/oxc_ast/src/serialize.rs
overlookmotel: Sorry the `raw_deser` syntax is so dreadful! I will try to improve it at some point...

leaysgur:
@overlookmotel Thanks for the review~! I think I've fixed your points.

> Yes, we'll probably need to alter the Rust AST so it includes the Span we need.
> But this PR already makes more tests pass, so let's get it merged, and we can then improve on it.

🫡 I will create another issue for this later.

overlookmotel (approved):

overlookmotel:
Thank you!

PR #10018, opened 2025-03-24
fix(ast/estree): Fix `ObjectProperty` node

therewillbecode:
Align our AST's ESTree output with that of TS-ESLint's for the`ObjectProperty` node. Relates to #9705 

Add `optional` field and set to `false` fo the `ObjectProperty` when serializing to TS estree.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -379,7 +379,7 @@ pub enum ObjectPropertyKind<'a> {
 #[ast(visit)]
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
-#[estree(rename = "Property", field_order(span, method, shorthand, computed, key, value, kind))]
+#[estree(rename = "Property", add_fields(optional = TsFalse), field_order(span, optional, method, shorthand, computed, key, value, kind))]
 pub struct ObjectProperty<'a> {
     pub span: Span,
     pub kind: PropertyKind,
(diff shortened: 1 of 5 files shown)


crates/oxc_ast/src/ast/js.rs
```diff
@@ -379,7 +379,7 @@ pub enum ObjectPropertyKind<'a> {
 #[ast(visit)]
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
-#[estree(rename = "Property", field_order(span, method, shorthand, computed, key, value, kind))]
+#[estree(rename = "Property", add_fields(optional = TsFalse), field_order(span, optional, method, shorthand, computed, key, value, kind))]
```
overlookmotel, line 382: Could you move the `optional` field to last in the field order? Our convention is to put TS fields last.

This line might also be more readable split into multiple lines:

```suggestion
#[estree(
    rename = "Property",
    add_fields(optional = TsFalse),
    field_order(span, method, shorthand, computed, key, value, kind, optional),
)]
```

overlookmotel: Also please rebase on latest main, or it'll conflict with your other 2 PRs which I've just merged.

therewillbecode: Sure thing. Done :)

therewillbecode: I am surprised that moving the `optional` field to the end in `field_order` doesn't change the conformance snapshot as based on my understanding of the macro I would expect this change to alter the order of the serialized node fields.

overlookmotel (approved):

PR #10041, opened 2025-03-25
fix(parser): store lone surrogates as escape sequence

### LINKED ISSUE #3526: codegen: string surrogates printed incorrectly

overlookmotel:
Closes #3526.

Fully parse strings containing lone surrogates and encode the string in `value`.

Encoding schema is to encode a lone surrogate as the lossy replacement character, followed by the code point in hex. i.e. `"\uD800"` in source code would be encoded as `\u{FFFD}d800` in `value`. The lossy replacement character itself is encoded as `\u{FFFD}fffd`.

There's nothing special about the lossy replacement character. Just had to choose *some* valid Unicode character to be the escape marker, and that seemed like a reasonable choice of a character which is likely to be rare in real-world code.

All the ESTree-Test262 test cases related to lone surrogates now pass.

WIP. A bit of tidying up to do yet.

### DIFF
--- a/crates/oxc_ast/src/ast/literal.rs
+++ b/crates/oxc_ast/src/ast/literal.rs
@@ -85,10 +85,14 @@ pub struct StringLiteral<'a> {
     #[content_eq(skip)]
     pub raw: Option<Atom<'a>>,
 
-    /// The string value contains replacement character (U+FFFD).
+    /// The string value contains lone surrogates.
+    ///
+    /// `value` is encoded using `\u{FFFD}` (the lossy replacement character) as an escape character.
+    /// Lone surrogates are encoded as `\u{FFFD}XXXX`, where `XXXX` is the code unit in hex.
+    /// The lossy escape character itself is encoded as `\u{FFFD}fffd`.
     #[builder(default)]
     #[estree(skip)]
-    pub lossy: bool,
+    pub lone_surrogates: bool,
 }
 
 /// BigInt literal
--- a/crates/oxc_ast/src/generated/assert_layouts.rs
+++ b/crates/oxc_ast/src/generated/assert_layouts.rs
@@ -776,7 +776,7 @@ const _: () = {
     assert!(offset_of!(StringLiteral, span) == 0);
     assert!(offset_of!(StringLiteral, value) == 8);
     assert!(offset_of!(StringLiteral, raw) == 24);
-    assert!(offset_of!(StringLiteral, lossy) == 40);
+    assert!(offset_of!(StringLiteral, lone_surrogates) == 40);
 
     assert!(size_of::<BigIntLiteral>() == 32);
     assert!(align_of::<BigIntLiteral>() == 8);
@@ -2167,7 +2167,7 @@ const _: () = {
     assert!(offset_of!(StringLiteral, span) == 0);
     assert!(offset_of!(StringLiteral, value) == 8);
     assert!(offset_of!(StringLiteral, raw) == 16);
-    assert!(offset_of!(StringLiteral, lossy) == 24);
+    assert!(offset_of!(StringLiteral, lone_surrogates) == 24);
 
     assert!(size_of::<BigIntLiteral>() == 20);
     assert!(align_of::<BigIntLiteral>() == 4);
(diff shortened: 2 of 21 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10041?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10041** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10041?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10041?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10040** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10040?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10039** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10039?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10036** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10036?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10041).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Boshen:
cc @andreubotella for a quick review.

andreubotella:
> cc @andreubotella for a quick review.

IIUC, if `lone_surrogates` is false, then any replacement characters in the value are true replacement characters; and if true, then the value string would need decoding, right? I think we can work with that in Nova. (cc @aapoalas)

That said, I worry that other users of oxc_ast might not be aware of this change, and might use `value` without checking for `lone_surrogate`. You could say that no `AsRef<str>` would accurately represent the JS string, but I'd say that decoding any lone surrogate lossily to the replacement character (as `String::from_utf16_lossy` does) would be a lot more useful for most use cases. If the extra memory use is not an issue, I'd suggest having `value` be this lossy decode, and having `lone_surrogate_string` be an `Option<Atom<'a>>` that is only `Some` if there are lone surrogates.

overlookmotel:
@andreubotella Thanks for your review.

Yes, this isn't entirely ideal for the 2 reasons you've pointed out:

1. The extra `lone_surrogates` flag adds 8 bytes to `StringLiteral`.
2. It's confusing that `str_lit.value.as_str()` sometimes returns an inaccurate string - you have to check the `lone_surrogates` flag to see if it needs to be decoded.

I considered the main priority to encode the string correctly *somehow* and to be able to pass the Test262 tests.

But I see this as a stop-gap. In the end, I think we need our own string types, to enable some other optimizations:

* String length is always `<= u32::MAX`.
* Separate `Ident` type for identifiers, which stores a 32-bit hash inline, to make using them as hash table keys faster.
* Faster string comparison and hashing.

https://github.com/oxc-project/backlog/issues/46#issuecomment-2603013242
https://github.com/oxc-project/backlog/issues/46#issuecomment-2603153807

So, when we do that, I'd imagine changing the implementation to more like:

```rs
struct StringLiteralValue {
    ptr: NonNull<u8>,
    len: u32,
    lone_surrogates: bool,
    _padding: [u8; 3],
}
```

`StringLiteralValue` would not dereference to `&str`, and when you print it, it would automatically decode the escapes. I think this would address both shortcomings.

But in meantime, I propose that we go with the approach in this PR, as it's at least is correct, even if it's messy. Does that sound OK to you?

overlookmotel:
Something is wrong. This PR should make no difference to result of `cargo minsize`, because codegen is only using the value of `raw` when `lone_surrogates` flag is set (prior to #10044). I'll investigate.

overlookmotel:
> Something is wrong. This PR should make no difference to result of `cargo minsize`, because codegen is only using the value of `raw` when `lone_surrogates` flag is set (prior to #10044). I'll investigate.

Nope. Snapshot was just out of date.

overlookmotel:
@overlookmotel Note to self: Add some tests. Especially for strings containing both lone surrogates and lossy replacement characters.

overlookmotel:
Tests added in #10175.

PR #10081, opened 2025-03-27
fix(isolated_declarations,linter,semantic,transformer): use `FormalParameter::has_modifier` to detect parameter property

ulrichstark:
Using `FormalParameter::has_modifier` makes code shorter, more explicit and also fixes code which hasn't checked for the `override` and `readonly` modifier before.


Dunqing:
Thank you! Could you split this PR into twos?

The first one is to shorten the code by using `FormalParameter::has_modifier`, and the second one is to fix the isolated declaration issue. So that what you want to do will be clearer, and easier for maintainers to review.

PR #10092, opened 2025-03-28
feat(ast)!: Add `computed` property to `TSEnumMember` and `TSEnumMemberName::TemplateString`

### LINKED ISSUE #10088: ast: Missing `computed` in `TSEnumMember` and `TemplateElement` pattern
Part of #9705 

For TS-ESLint AST, we need to add `computed` property for `TSEnumMember`.

https://github.com/typescript-eslint/typescript-eslint/blob/5b187f4332a6b30ba465e9321639e25857a7788f/packages/ast-spec/src/element/TSEnumMember/spec.ts#L11

```ts
enum X {
  ["d1"], // computed: true
  "d2",   // computed: false
}
```

But our Rust AST does not have a `computed` property.

https://github.com/oxc-project/oxc/blob/8e55395efb03505c2e3bf6ff496248a21290d07a/crates/oxc_ast/src/ast/ts.rs#L102-L106

At the same time, I found that `TemplateLiteral` seemed to be treated as `TSEnumMemberName::String` = `StringLiteral`.

https://github.com/oxc-project/oxc/blob/8e55395efb03505c2e3bf6ff496248a21290d07a/crates/oxc_parser/src/ts/statement.rs#L70-L71

```ts
enum Y {
  [`t`],
  // `t2`, // This is syntax error, so it's OK 
}
```

But in this case, `id` should be `TemplateLiteral` instead of `(String)Literal` for now.

leaysgur:
Fixes #10088

### DIFF
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -3757,13 +3757,30 @@ impl Gen for TSEnumMember<'_> {
     fn r#gen(&self, p: &mut Codegen, ctx: Context) {
         match &self.id {
             TSEnumMemberName::Identifier(decl) => decl.print(p, ctx),
-            TSEnumMemberName::String(decl) => p.print_string_literal(decl, false),
+            TSEnumMemberName::String(decl) => {
+                if self.computed {
+                    p.print_ascii_byte(b'[');
+                }
+                p.print_string_literal(decl, false);
+                if self.computed {
+                    p.print_ascii_byte(b']');
+                }
+            }
             TSEnumMemberName::TemplateString(decl) => {
+                if self.computed {
+                    p.print_ascii_byte(b'[');
+                }
                 let quasi = decl.quasis.first().unwrap();
                 p.add_source_mapping(quasi.span);
+                p.print_ascii_byte(b'`');
                 p.print_str(quasi.value.raw.as_str());
+                p.print_ascii_byte(b'`');
+                if self.computed {
+                    p.print_ascii_byte(b']');
+                }
             }
         }
+
         if let Some(init) = &self.initializer {
             p.print_soft_space();
             p.print_equal();
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -3755,16 +3755,14 @@ impl Gen for TSEnumBody<'_> {
 
 impl Gen for TSEnumMember<'_> {
     fn r#gen(&self, p: &mut Codegen, ctx: Context) {
-        if self.computed {
-            p.print_ascii_byte(b'[');
-        }
         match &self.id {
             TSEnumMemberName::Identifier(decl) => decl.print(p, ctx),
             TSEnumMemberName::String(decl) => p.print_string_literal(decl, false),
-            TSEnumMemberName::TemplateString(decl) => decl.print(p, ctx),
-        }
-        if self.computed {
-            p.print_ascii_byte(b']');
+            TSEnumMemberName::TemplateString(decl) => {
+                let quasi = decl.quasis.first().unwrap();
+                p.add_source_mapping(quasi.span);
+                p.print_str(quasi.value.raw.as_str());
+            }
         }
         if let Some(init) = &self.initializer {
             p.print_soft_space();
--- a/crates/oxc_codegen/src/gen.rs
+++ b/crates/oxc_codegen/src/gen.rs
@@ -3755,9 +3755,16 @@ impl Gen for TSEnumBody<'_> {
 
 impl Gen for TSEnumMember<'_> {
     fn r#gen(&self, p: &mut Codegen, ctx: Context) {
+        if self.computed {
+            p.print_ascii_byte(b'[');
+        }
         match &self.id {
             TSEnumMemberName::Identifier(decl) => decl.print(p, ctx),
             TSEnumMemberName::String(decl) => p.print_string_literal(decl, false),
+            TSEnumMemberName::TemplateString(decl) => decl.print(p, ctx),
+        }
+        if self.computed {
+            p.print_ascii_byte(b']');
         }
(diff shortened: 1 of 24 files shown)


leaysgur:
I am having trouble updating the `minsize.snap` properly... 🤔

When I run `cargo run -p oxc_minsize` on my local, it creates these diffs.

```diff
diff --git a/tasks/minsize/minsize.snap b/tasks/minsize/minsize.snap
index 6be7c45c6..e9e3ed57f 100644
--- a/tasks/minsize/minsize.snap
+++ b/tasks/minsize/minsize.snap
@@ -21,7 +21,7 @@ Original   | minified   | minified   | gzip       | gzip       | Fixture

 3.20 MB    | 1.01 MB    | 1.01 MB    | 324.14 kB  | 331.56 kB  | echarts.js

-6.69 MB    | 2.28 MB    | 2.31 MB    | 466.12 kB  | 488.28 kB  | antd.js
+4.12 MB    | 1.54 MB    | 2.31 MB    | 447.69 kB  | 488.28 kB  | antd.js

 10.95 MB   | 3.35 MB    | 3.49 MB    | 860.95 kB  | 915.50 kB  | typescript.js
```

This also appears to be very different from the results in CI.

overlookmotel:
Thanks for this.

Since we know for sure that the template string must have no expressions and only 1 quasi in this case, I wonder if we should add a specific type for it?

```rs
pub enum TSEnumMemberName<'a> {
    Identifier(Box<'a, IdentifierName<'a>>) = 0,
    String(Box<'a, StringLiteral<'a>>) = 1,
    Template(Box<'a, SimpleTemplateLiteral<'a>>) = 2,
}

/// A template literal with no expressions and only 1 quasi.
/// Basically a copy of `StringLiteral`, but with backtick quotes.
pub struct SimpleTemplateLiteral<'a> {
    span: Span,
    #[estree(via = StringLiteralValue)]
    pub value: Atom<'a>,
    #[content_eq(skip)]
    pub raw: Option<Atom<'a>>,
    #[builder(default)]
    #[estree(skip)]
    pub lone_surrogates: bool,
}
```

Personally I think that'd be preferable as, generally speaking, I think it's ideal for the AST to statically be unable to represent syntax which is illegal.

But @Boshen what do you think?

---

Additionally, instead of a `computed` property, we could choose to represent whether the property is computed or not with extra enum variants:

```rs
pub enum TSEnumMemberName<'a> {
    Identifier(Box<'a, IdentifierName<'a>>) = 0,
    String(Box<'a, StringLiteral<'a>>) = 1,
    Template(Box<'a, SimpleTemplateLiteral<'a>>) = 2,
    ComputedString(Box<'a, StringLiteral<'a>>) = 3,
    ComputedTemplate(Box<'a, SimpleTemplateLiteral<'a>>) = 4,
}
```

Again, the rationale is that this makes it impossible to represent an illegal syntax in AST:

```ts
enum Foo {
    // Illegal syntax - identifiers cannot be computed
    [x] = 1,
}
```

This also makes `TSEnumMember` 8 bytes smaller than adding a `computed: bool` property.

leaysgur:
Thanks for the review~!
I too prefer that idea because it is explicit.

Let's wait for the Boshen's decision. 🙏🏻

overlookmotel:
One other thing: Doesn't [codegen](https://github.com/oxc-project/oxc/pull/10092/files#diff-cc3a14aa7f110fa2c7adf486701a5ae105f6a31d5ba260cae72b9a3a048b3777) need to wrap in `[...]` when `computed: true` to preserve the original syntax?

Otherwise this:

```ts
export enum X {
  [`A`] = 123,
}
```

is printed as:

```ts
export enum X {
  `A` = 123,
}
```

...which is invalid syntax.

overlookmotel:
@Boshen Could you please give your opinion on https://github.com/oxc-project/oxc/pull/10092#issuecomment-2774875698? Personally, I like the tighter constraint that a `SimpleTemplateLiteral` would give. But I know you're not keen to add more AST types.

crates/oxc_codegen/src/gen.rs
```diff
@@ -3770,8 +3770,30 @@ impl Gen for TSEnumMember<'_> {
...
+            }
+            TSEnumMemberName::TemplateString(decl) => {
+                if self.computed {
+                    p.print_ascii_byte(b'[');
+                }
+                let quasi = decl.quasis.first().unwrap();
+                p.add_source_mapping(quasi.span);
+                p.print_ascii_byte(b'`');
+                p.print_str(quasi.value.raw.as_str());
+                p.print_ascii_byte(b'`');
```
Dunqing, line 3790:
```suggestion
                decl.print(p)
```
Does it work here?

Dunqing:
I think adding a `computed` field is good, which is also compatible with `typescript-eslint`.

crates/oxc_codegen/src/gen.rs
overlookmotel: It would. But we know for sure that there is only 1 `quasi` and 0 `expression`s, so this should be faster.

overlookmotel:
Rebased on main to fix merge conflicts. I'm going to merge this when CI passes.

---

I realized my idea of a `SimpleTemplateLiteral` is not good. The problem is that visitors for `TemplateLiteral` need to visit this node, and they won't if it's a different type.

I do think my other suggestion of removing the `computed` field and replacing with additional enum variants on `TSEnumMemberName` is good though:

```rs
pub enum TSEnumMemberName<'a> {
    Identifier(Box<'a, IdentifierName<'a>>) = 0,
    String(Box<'a, StringLiteral<'a>>) = 1,
    Template(Box<'a, SimpleTemplateLiteral<'a>>) = 2,
    ComputedString(Box<'a, StringLiteral<'a>>) = 3,
    ComputedTemplate(Box<'a, SimpleTemplateLiteral<'a>>) = 4,
}
```

I'll open an issue for making that change. But let's get this merged in meantime!

PR #10096, opened 2025-03-28
fix(semantic): use `FormalParameter::has_modifier` to detect parameter properties

ulrichstark:
Checks for the `override` and `readonly` modifier were missing before.


Dunqing (approved):
Good catch! I am going to merge it later. Because there is a huge [stack](#10095) that is going to merge, and it probably will conflict with this PR.

ulrichstark:
I rebased onto main to fix the merge conflicts.

PR #10097, opened 2025-03-28
refactor(linter): use `FormalParameter::has_modifier` to detect parameter properties

ulrichstark:
By using this function, the code is shorter and the missing check for the `override` modifier is added.


Dunqing (approved):
Looks good!

PR #10099, opened 2025-03-28
refactor(transformer): use `FormalParameter::has_modifier` to detect parameter properties

ulrichstark:
By using this function, the code is shorter and aligns with the changes of the original PR.


Dunqing (approved):

Dunqing:
BTW, I think this PR and #10097, and #10098 should merge into one because their purposes are the same, just for "shorten code by using FormalParameter::has_modifier". But anyway, maybe I am okay merging these PRs. Thank you for your work!

PR #10100, opened 2025-03-28
refactor(linter): use items of `oxc_ast::ast` module directly

ulrichstark:
All other lint rules do it this way and it leads to less noisy code.


Boshen:
Thank you.

PR #10116, opened 2025-03-29
feat(quote): implement `jsquote!` and `jsquote_expr!`

Qix-:
Hi! This is just a draft. Wanted to see what you might think about this implementation for a `quote!`-like macro for OXC, inspired heavily by the infamous `quote` crate. Please just consider it a proposal; if it isn't the best way forward for this, that's alright too.

ref oxc-project/backlog#151

This commit introduces the `jsquote!` and `jsquote_expr!` procedural macros, which utilize the auto-derivation system to build a proto-AST when using them in order to produce in-situ Rust invocations.

- `jsquote_expr!` returns an `Expression<'_>` and uses `.parse_expression()`
- `jsquote!` returns a `Vec<'_, Statement<'_>>` and uses `.parse()`

~~It only has half of the placeholder implementation right now but anything without placeholders _should_ work as you'd expect.~~ They are implemented in the latest commit but have a few caveats (see comments below).

---

I want to say up-front that there are some really unfortunate hacks in the proc macro to make this work on stable. Nothing "unsafe", just fragile. Until some of the long-time unstable features are no longer so, they can be improved.

I've also tried to enable the unstable features automatically depending on which compiler is used (I did it the way I did, instead of using features, since all the CI runs with `--all-features`, but I didn't want to muck with the `rust-toolchain.toml` and cause a bigger stir than this PR probably already does).

```rust
let a = Allocator::default();
let span = Span::empty(0);

let expr: Expression = jsquote_expr!(&a, span, { 1234.5 + 2 });

assert!(expr.content_eq(&Expression::BinaryExpression(Box::new_in(
    BinaryExpression {
        span,
        left: Expression::NumericLiteral(Box::new_in(
            NumericLiteral {
                span,
                value: 1234.5,
                raw: Some(Atom::new_const("1234.5"),),
                base: NumberBase::Float,
            },
            &a
        )),
        operator: BinaryOperator::Addition,
        right: Expression::NumericLiteral(Box::new_in(
            NumericLiteral {
                span,
                value: 2.0,
                raw: Some(Atom::new_const("2"),),
                base: NumberBase::Decimal,
            },
            &a
        )),
    },
    &a
))));
```


Qix-:
I'm grappling a bit with substitutions. All of the differentiating types (`Expression` vs `Argument`, as well as `BindingIdentifier`) makes it tricky to substitute in items that aren't exactly the AST type.

Throwing an `.into()` onto everything causes issues with how vectors are built up, too, given that Rust isn't (yet) able to solve for these types until Chalk comes around, I'd think. So that negates things like `identifier_expr.into::<BindingIdentifier>()` (pseudocode) being the 'magic' sauce here.

Further, doing what `quote!` actually does, which is generating tokens, and then re-parsing is out of the question here since the parser infrastructure is rigid and doesn't allow for a generic `.parse()?` to work either - especially given the disparity in `parse()` and `parse_expression()`. Otherwise, for the dubious cases, a `#(#ident?)` sort of syntax could be crafted such that `ident.codegen().parse()?` (pseudocode) could be done at a small performance overhead.

So for now, I'll just add as many exceptions as I can with the hopes that maybe someday this can be formalized in a bit more of a generic way.

Qix-:
Another way to go about this is to have a dedicated set of utility types that can be converted into various like-types. For example:

```rust
let a = oxc_allocator::Allocator::default();
let s = oxc_span::Span::empty(0);
let ident = oxc_quote::Ident(&a, s, "foobar");

let _ = jsquote!(&a, s, {
    var #ident = 10;
    console.log(#ident);
});
```

Which would expand to something like...

```rust
// ...
VariableDeclarator {
    // ...
    id: BindingPattern {
        kind: ident.into(), // Ident -> BindingPatternKind::BindingIdentifier(...)
        // ...
    },
    // ...
}
// ...
{
    arguments: [
        ident.into() // Ident -> Argument::IdentifierReference(...)
    ]
}
```

Qix-:
For some reason I'm getting this on CI, but I can't reproduce locally by running the same command.

```rust
error[E0609]: no field `lossy` on type `&ast::literal::StringLiteral<'_>`
    --> crates/oxc_ast/src/generated/derive_to_rust.rs:5726:32
     |
57[26](https://github.com/oxc-project/oxc/actions/runs/14159257451/job/39662477714?pr=10116#step:4:27) |                 ("lossy", self.lossy.to_rust())
     |                                ^^^^^ unknown field
     |
     = note: available fields are: `span`, `value`, `raw`, `lone_surrogates`
```

Also, grepping for `lone_surrogates` across the entire project gives me nothing, so I'm not sure what I'm missing here.

Dunqing:
Thank you for working on this great staff! We will take a look at this later.

overlookmotel:
Unfortunately, no-one has had time to go through this PR in several months (the problem with large PRs).

But just to say, obviously this was a significant amount of work, and we do appreciate your efforts. I've added a comment to https://github.com/oxc-project/backlog/issues/151#issuecomment-3337651422 to go through this PR and see if we can finish it off and get it merged once the core team have more bandwidth.

So, sorry we've not done anything with it. But hopefully your work won't go to waste in the end.

PR #10131, opened 2025-03-30
chore(deps): Update pnpm-lock to fix dependabot alerts#18

leaysgur:
Fixes https://github.com/oxc-project/oxc/security/dependabot/18

(I don't think we've ever used that feature, but warning shows up every push. 🥲 )


Boshen:
FYI lockfile gets regenerated every monday by renovate file.

PR #10141, opened 2025-03-30
fix(linter): span disable directive correctly on next line

### LINKED ISSUE #10115: lint: disable-next-line on max-params doesn't work when placed before function declaration
The disable-next-line directive doesn't seem to span large enough to cover the full function declaration. The following code snippets should not trigger lint errors:

```tsx
// eslint-disable-next-line max-params
function someMaxParamsFunction(
  param1: number,
  param2: number,
  param3: number,
  param4: number,
  param5: number,
  param6: number
) {

}
```

A workaround is to put the disable-next-line directive before the list of params instead but this doesn't read as nicely and is at odds with how eslint handles this exact scenario.

```ts
function someMaxParamsFunction(
  // eslint-disable-next-line max-params
  param1: number,
  param2: number,
  param3: number,
  param4: number,
  param5: number,
  param6: number
) {

}
```

ulrichstark:
Previous code didn't account for the newline characters (0..N carriage return characters and one line feed character) between the line of the disable directive comment and the next line.

Fixes #10115

### DIFF
--- a/crates/oxc_linter/src/disable_directives.rs
+++ b/crates/oxc_linter/src/disable_directives.rs
@@ -173,9 +173,8 @@ impl<'a> DisableDirectivesBuilder<'a> {
                     }
 
                     if let Some(next_line) = lines_after_comment_end.next() {
-                        let next_line = next_line.strip_suffix('\n').unwrap_or(next_line);
-                        let next_line = next_line.strip_suffix('\r').unwrap_or(next_line);
-                        stop += next_line.len() as u32;
+                        let next_line_trimmed = next_line.trim_end_matches(['\n', '\r']);
+                        stop += next_line_trimmed.len() as u32;
                     }
 
                     if text.trim().is_empty() {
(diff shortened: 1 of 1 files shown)


ulrichstark:
Just converted this PR to draft because I think some tests won't hurt

crates/oxc_linter/src/disable_directives.rs
```diff
@@ -164,10 +164,20 @@ impl<'a> DisableDirectivesBuilder<'a> {
...
+                    let mut stop = comment_span.end;
+                    let mut lines_after_comment_end =
+                        source_text[comment_span.end as usize..].split_inclusive('\n');
+
+                    if let Some(rest_of_line) = lines_after_comment_end.next() {
+                        stop += rest_of_line.len() as u32;
+                    }
+
+                    if let Some(next_line) = lines_after_comment_end.next() {
+                        let next_line = next_line.strip_suffix('\n').unwrap_or(next_line);
```
Boshen, line 176: These APIs accept a pattern https://doc.rust-lang.org/std/str/pattern/index.html

So a ['\n', '\r'] should be possible.

And it seems like you can use trim_end_matches? https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches

PR #10157, opened 2025-04-01
fix(linter): nested configuration directory resolution

### LINKED ISSUE #10156: linter: Nested configuration rules are not detected in subdirectories
While using nested configurations, there is an issue which is causing Oxlint CLI to bypass rules in nested directories.

```
├── .oxlintrc.json
├── packageA
│   ├── .oxlintrc.json
│   └── src
│       └── helpers.js
├── packageB
│   ├── .oxlintrc.json
│   └── helpers.js
```

In the example above, we have two nested configs.
- `packageA` has a `src` directory with a file nested inside it. Lint configuration is at top-level.
- `packageB` has a file which is adjacent to the lint configuration.

If the root config contains a rule `no-console: warn`, and a nested configuration has `no-console: off`, the rule is still applied in the nested directory when running the Oxlint CLI command. **This only happens when files are nested within a subdirectory relative to the configuration file. The functionality works as expected if the subject file is located adjacently to .oxlintrc.json**

The expected behaviour is that the rule should be turned off in the nested directory since there is a nested `.oxlintrc.json` file present.

This only applies to the Oxlint CLI. In the language server, it appears that the nested configuration is behaving correctly as the warning does disappear when the rule is turned off.  

I've produced a minimal reproducible example of the issue in this project - https://github.com/zubhav/oxlint-nested-config-issue-example

1) Clone project and run `yarn`
2) Run `yarn lint`

Output is:
```
  ⚠ eslint(no-console): eslint(no-console): Unexpected console statement.
   ╭─[packageA/src/helpers.js:1:1]
 1 │ console.log("Hello from packageA");
   · ───────────
   ╰────
  help: Delete this console statement.

Found 1 warning and 0 errors.
```

3) Based on the output above, the nested configuration in `packageA` is not being used to lint files within subdirectories of `packageA`. The expected behaviour is that running `yarn lint` will output no warnings at all, since the rule is turned off in all nested configurations.

zubhav:
Fixes https://github.com/oxc-project/oxc/issues/10156

I'm not very familiar with the inner workings of this project - this was a fix based on what I could understand after exploring the codebase for the first time. @camchenry It would be great to get your eyes on this since you are familiar with the nested config implementation.

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -185,6 +185,8 @@ impl Runner for LintRunner {
                 // Start from the file's parent directory and walk up the tree
                 let mut current = path.parent();
                 while let Some(dir) = current {
+                    // NOTE: Initial benchmarking showed that it was faster to iterate over the directories twice
+                    // rather than constructing the configs in one iteration. It's worth re-benchmarking that though.
                     directories.insert(dir);
                     current = dir.parent();
                 }
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -176,15 +176,17 @@ impl Runner for LintRunner {
 
         if use_nested_config {
             // get all of the unique directories among the paths to use for search for
-            // oxlint config files in those directories
-            // e.g. `/some/file.js` and `/some/other/file.js` would both result in `/some`
+            // oxlint config files in those directories and their ancestors
+            // e.g. `/some/file.js` will check `/some` and `/`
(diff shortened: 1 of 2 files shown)


zubhav:
@camchenry Could you also take a look at https://github.com/oxc-project/oxc/pull/10157 please? Its another fix for nested config

camchenry (approved):

camchenry:
I think this makes sense, thanks for the contribution!

PR #10174, opened 2025-04-02
fix(linter): Make extended configs properly inherit plugins

### LINKED ISSUE #10173: linter: extended configs do not inherit plugins from parent config

0.16.3

### What command did you run?

`npx oxlint . --print-config`:
`npx oxlint ./packageB --print-config --config packageB/.oxlintrc.json`

Base config:

```jsonc
{
  "plugins": ["import"],
  "rules": {
    "import/export": "error"
  }
}
```

Extended config:

```jsonc
{
  "extends": ["../.oxlintrc.json"],
  "rules": {
    "import/export": "warn"
  }
}
```

### What happened?
See here for reproducible example: https://github.com/zubhav/oxlint-nested-config-issue-example/tree/oxlint-nested-config-vitest-issue-example

`npx oxlint . --print-config`

```
{
  "plugins": [
    "import"
  ],
  ...
}
```

`npx oxlint ./packageB --print-config --config packageB/.oxlintrc.json`

```
{
  "plugins": [
    "react",
    "unicorn",
    "typescript",
    "oxc"
  ],
  ...
}
```

I initially thought this issue was specifically related to Vitest plugin, but it seems to be happening for all plugins.

**Current behaviour**
- Extended configurations do not inherit plugins from parent configuration. 
- If no plugins are specified in the extended config, the plugins default to react, unicorn, typescript, oxc
- If a plugins array is specified explicitly in extended config, the final config will include it

**Expected behaviour**
- If extended configuration does not specify `plugins` attribute, it should be inherited from the parent config
- If the extended configuration does specify `plugins` attribute, it should be appended after plugins of parent config
- This is the current behaviour that ESLint uses

zubhav:
Fixes https://github.com/oxc-project/oxc/issues/10173

When extending configuration files in Oxlint, plugins from the extended config were not being properly inherited, which caused rules requiring those plugins to be filtered out. This was encountered in issue #10173 where rules like vitest/no-focused-tests were being ignored even when the vitest plugin was defined in a base configuration.

The issue occurred because in the implementation of from_oxlintrc. The following changes were made:

Modified plugin inheritance logic in from_oxlintrc:
- Now checks if the child configuration has explicitly specified plugins
- If the child has default plugins (not explicitly specified), it inherits from the parent
- If the child has explicitly specified plugins, it combines them with the parent's plugins (like ESLint)
- Empty plugin arrays are treated as explicit specifications

Added tests to verify the behavior:
- Tests for empty configurations
- Tests for configurations with empty plugin arrays
- Tests for configurations with explicitly specified plugins
- Tests for configurations that extend other configurations

This ensures that when rules are added from an extended config, the necessary plugins are already enabled, so the rules won't be filtered out in the build step.

cc @camchenry


crates/oxc_linter/src/config/config_builder.rs
```diff
@@ -149,7 +149,19 @@ impl ConfigStoreBuilder {
                                 ConfigStoreBuilder::from_oxlintrc(true, extended_config)?;
                             let rules = std::mem::take(&mut extended_config_store.rules);
                             builder = builder.with_rules(rules);
-                            builder = builder.and_plugins(extended_config_store.plugins(), true);
+
+                            // Handle plugin inheritance
+                            let parent_plugins = extended_config_store.plugins();
+                            let child_plugins = builder.plugins();
```
zubhav, line 155: We seem to add standard plugins by default when building this. 

So, below we need to check if child_plugins points to the same reference as `LintPlugins::default()` to infer whether the child config was initialized without `plugins` attribute defined.

I'm open to other ideas here!

camchenry:
Haven't had a chance to review this yet, but I will take a look next time I am able.

crates/oxc_linter/src/config/config_builder.rs
```diff
@@ -908,49 +920,149 @@ mod test {
 
     #[test]
     fn test_extends_plugins() {
+        use std::fs;
+
+        // Create a temporary directory for test files
+        let temp_dir = tempfile::tempdir().unwrap();
+        let parent_path = temp_dir.path().join("parent.json");
+        let child_no_plugins_path = temp_dir.path().join("child_no_plugins.json");
+        let child_with_plugins_path = temp_dir.path().join("child_with_plugins.json");
```
camchenry, line 929: Let's try to avoid adding a dependency unless we really need it, so I would recommend creating some additional files in the `fixtures` directory that have the content specified here.

camchenry (commented):
These tests look good, but let's refactor it to not need an extra dependency, to help keep things simpler

crates/oxc_linter/src/config/config_builder.rs
zubhav: Thanks @camchenry I've just addressed this

zubhav:
@camchenry Could you also take a look at https://github.com/oxc-project/oxc/pull/10157 please? Its another fix for nested config

camchenry (approved):
Thank you for the updates!

zubhav:
@camchenry Could you help with the graphite-app comment above? I've configured the account so I'm guessing we'll need to re-add the label. Thanks!

PR #10201, opened 2025-04-03
fix(ast/estree): Fix `TSEnumMember`

therewillbecode:
Adds the `computed` field to the Estree serialiser for  the `TSEnumMember` node.


overlookmotel:
Ah sorry, this one is already in progress on #10092.

overlookmotel:
If these discrepancies are blocking progress, we could always merge this PR first, and continue refining it in #10092. Please let me know if you think that's advisable.

therewillbecode:
> Ah sorry, this one is already in progress on #10092.

No problem :) 

Can close as not a blocker on progress for me.

PR #10206, opened 2025-04-03
fix(ast/estree): Fix `TSInstantiationExpression`

therewillbecode:
Renames field from `typeParameters` to `typeArguments` field for the Estree serialiser on  the `TSInstantiationExpression` node.


overlookmotel:
On this one, I think it'd be better to rename the field in the Rust type, to match what #9815 did for various other types.

```diff
pub struct TSInstantiationExpression<'a> {
    pub span: Span,
    pub expression: Expression<'a>,
-   pub type_parameters: Box<'a, TSTypeParameterInstantiation<'a>>,
+   pub type_arguments: Box<'a, TSTypeParameterInstantiation<'a>>,
}
```

overlookmotel:
Implemented in #10327.

Issue #10002, opened 2025-03-24
ast: `#[estree(add_fields(...))]` for `JSXOpeningFragment` should not be applied to TS-ESLint AST


leaysgur:
Part of #9703 

In PR #9747, `attributes` and `selfClosing` props were added to `JSXOpeningFragment`.

https://github.com/oxc-project/oxc/blob/094dec1e92e01ce1965f132887124fd458ede359/crates/oxc_ast/src/ast/jsx.rs#L133

![Image](https://github.com/user-attachments/assets/41071682-547f-480c-8e6d-c5583bc49f5f)

As a result, it matched correctly with ESTree, but it appears these were unnecessary for TS-ESLint.

|ESTree(acorn-jsx)|TS-ESLint|
|-|-|
|![Image](https://github.com/user-attachments/assets/a3287103-c475-41a7-9599-48321ad1bba7)|![Image](https://github.com/user-attachments/assets/aebbc42f-7823-4e69-b16e-f2c5d878723e)|

Coverage diffs can be found by `--filter jsxJsxsCjsTransformCustomImport`.

overlookmotel:
Do you think we should raise this as an issue on TS-ESLint parser? My understanding was that they intend their AST to be a superset of ESTree. Yes, JSX is not part of ESTree, but `acorn-jsx` might be considered as close to a reference implementation as you can get?

Or we can abandon trying to align with `acorn-jsx` and just align with TS-ESLint for JSX syntax?

overlookmotel:
We could alternatively add a `#[js]` attribute which behaves like `#[ts]` but in reverse.

But personally I don't think that's a great option. I think we want the TS AST to only extend the JS AST, not remove things from it.

leaysgur:
Ah~ sorry, I was short-sighted. 🙈 

> just align with TS-ESLint for JSX syntax?
> add a #[js] attribute which behaves like #[ts] but in reverse

I don't think either is necessary in this case.

Let's leave our implementation as-is and just ignore the coverage diff is enough.

For `JSXOpeningFragment`, since `selfClosing` is always `false` and `attributes` is always empty.
IMO, it seems unnecessary for Acorn JSX as well. 😅

overlookmotel:
Fixed in #10208.

We've used a hack to include these fields only in JS AST. I've made a TODO comment to come back to this later and find a better way.

PR #10208, opened 2025-04-03
fix(ast/estree): Fix `JSXOpeningFragment`

therewillbecode:
Fix `JSXOpeningFragment` node estree TS serialization by skipping `selfClosing` and `attributes` fields when AST is TS.

### DIFF
--- a/crates/oxc_ast/src/ast/jsx.rs
+++ b/crates/oxc_ast/src/ast/jsx.rs
@@ -128,7 +128,7 @@ pub struct JSXFragment<'a> {
 #[ast(visit)]
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
-#[estree(via = JSXOpeningFragmentConverter)]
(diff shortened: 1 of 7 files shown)


overlookmotel:
Related: #10002

overlookmotel:
We decided not to care about the raw transfer deserializer yet, but on this type we do need to, because it exists in the JS AST, and we don't want to break JSX support.

#10238 adds the necessary feature to implement `raw_deser` correctly for this type. Once it's merged, could you please rebase on top of it? I'll add a commit to this PR to fill in `raw_deser`.

---

We also have a problem with the TS type defs. This is annoying, but I think we can mis-use the `#[ts]` attr on meta types to get what we need.

I think you need to:

1. Add `#[estree(add_fields(attributes = JSXOpeningFragmentAttributes, selfClosing = TsFalse))]` to the type.
2. Bring back the `JSXOpeningFragmentAttributes` meta type. But add a `#[ts]` attr to it.

That's wrong - these fields aren't TS AST-only, they're JS AST-only. But still I think it'll have the desired effect of adding those 2 fields to the TS type def, and making them optional.

overlookmotel:
I've rebased on latest main and pushed a commit to fix the `raw_deser` impl. If you could fix the TS type def, I think we'll be good to go.

overlookmotel (approved):
Thanks!

PR #10228, opened 2025-04-04
fix(various): unwrap `Result` of `write!` macro

overlookmotel:
#10196 replaced the anti-pattern of `str.push_str(&format!("..."));` with `let _ = write!(str, "...");`.

However, this ignores the `Result` returned by `write!` macro. It's unspecified what happens when the `String` reaches capacity - it may return `Result::Err`, or may panic. Ensure we get a panic, rather than silently generating incorrect output by calling `unwrap()` on the `Result`.

This problem could never arise in practice on 64-bit systems because max capacity of `String` is enormous, but it's not impossible on 32-bit systems (e.g. WASM).

### DIFF
--- a/apps/oxlint/src/output_formatter/stylish.rs
+++ b/apps/oxlint/src/output_formatter/stylish.rs
@@ -72,7 +72,7 @@ fn format_stylish(diagnostics: &[Error]) -> String {
             .max()
             .unwrap_or(0);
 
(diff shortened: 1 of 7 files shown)


Boshen:
### Merge activity

* **Apr 4, 10:37 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10236, opened 2025-04-04
perf(cfg, diagnostics, lexer, syntax, tasks): remove `write!` macro where unnecessary

overlookmotel:
Same as #10230, for small changes to various crates.

Replace usages of `write!` macro with either `fmt.write_str(...)` for static strings or `Display::fmt` / `Debug::fmt` for other values, where it's not being used to concatenate multiple values.

`Formatter::write_str` is more performant, as it avoids various checks. `Display::fmt` / `Debug::fmt` may also perform a little better in some cases, and will be equivalent in others. But in all cases it should be better for compile times, due to avoiding macro expansion and trait resolution.

### DIFF
--- a/crates/oxc_cfg/src/dot.rs
+++ b/crates/oxc_cfg/src/dot.rs
@@ -1,4 +1,7 @@
-use std::{borrow::Cow, fmt};
+use std::{
+    borrow::Cow,
+    fmt::{self, Debug, Display},
+};
 
 use itertools::Itertools as _;
 use petgraph::{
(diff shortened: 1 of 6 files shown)


Boshen:
### Merge activity

* **Apr 4, 11:00 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10237, opened 2025-04-04
feat(linter): add eslint/block-scoped-var

huangtiandi1999:
Relates to #479 
Rule detail：https://eslint.org/docs/latest/rules/block-scoped-var


camc314 (approved):
thanks!

PR #10258, opened 2025-04-05
docs(linter): rule `eslint/eqeqeq` add "null" & "smart" options

### LINKED ISSUE #10255: `eslint/eqeqeq`: option to allow `==` for nullish checks
It's quite common and perfectly fine to check nullish instead of explicitly `undefined` or `null`.

```js
if (something == null) return;
```

vs

```js
if (something === undefined || something === null) return;
```

JakobJingleheimer:
Sorry if I didn't format it correctly. I followed [max nested callbacks](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/max_nested_callbacks.rs).

Closes https://github.com/oxc-project/oxc/issues/10255

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/eqeqeq.rs
+++ b/crates/oxc_linter/src/rules/eslint/eqeqeq.rs
@@ -23,42 +23,67 @@ pub struct Eqeqeq {
 
 declare_oxc_lint!(
     /// ### What it does
-    /// Requires the use of the === and !== operators
+    ///
+    /// Requires the use of the `===` and `!==` operators.
     ///
     /// ### Why is this bad?
+    ///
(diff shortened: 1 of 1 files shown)


Sysix (changes requested):
Thanks for updating the docs by yourself 🫶 
The doc structure is old. Can you update the complete doc to follow the structure like: https://github.com/therewillbecode/oxc/blob/main/crates/oxc_linter/src/rules/eslint/valid_typeof.rs

Note: Add examples for `correct` and `incorrect` code. 
Some smalls things like `Example` to `Examples` are included in my request :')

Sysix (approved):

PR #10259, opened 2025-04-05
refactor(linter): update comments, improve tests, add variant All to LintFilterKind

ulrichstark:
I split this PR into easily reviewable commits. While working on this PR, I noticed two possible follow up changes:

`LintFilterKind::Rule` and `LintFilterKind::Generic` behave exactly the same in `ConfigStoreBuilder::with_filter` because `LintFilterKind::Rule` never actually uses or checks for the plugin it stores. Should I include the check for plugin?

The proposed fifth `LintFilterKind` isn't implemented yet (see: `// TODO: plugin + category? e.g -A react:correctness)`). Should I add it?


camchenry (approved):
Nice! I like this change.

ulrichstark:
@camchenry Thanks for reviewing and merging! Are you interested in the possible follow up changes I mentioned in the PR description?

PR #10268, opened 2025-04-06
feat(language_server): use linter runtime

### LINKED ISSUE #7118: linter: oxlint VSCode extension not show warn or error for rules like `import/no-cycle` and `import/no-duplicates`

0.11.0

### What command did you run?

oxlint -c oxlintrc.json

```json
{
  "plugins": ["import", "unicorn"],
  "rules": {
    "import/no-commonjs": "error",
    "import/no-cycle": "error",
    "import/no-duplicates": "error",
    "import/no-self-import": "error"
  }
}
```

### What happened?

Rule `import/no-commonjs` works perfectly, but `import/no-cycle` or `import/no-duplicates` not show any sign of error or warn. Maybe problem in multiple spans in rule diagnostic or maybe in incorrect in terms of `vscode` large help info that not correctly handled by VSCode extension?

P. S. all rules works well in oxlint console output

Sysix:
closes #7118

Added tests for:
- (language_server & VSCode) nested configuration
- (language_server & VSCode) nested configuration extends
- (language_server) without nested configuration

The implementation requires at the moment a feature flag for `oxc_linter`. I added documentation in `runtime.rs` to let others know why this is needed. Maybe someone can refactor it to avoid the flag. My first idea is a Trait for file system, but you guys know what to do <3 

grafik

### DIFF
--- a/crates/oxc_language_server/src/linter/tester.rs
+++ b/crates/oxc_language_server/src/linter/tester.rs
@@ -6,7 +6,10 @@ use tower_lsp_server::{
     lsp_types::{CodeDescription, NumberOrString, Uri},
 };
 
-use super::{error_with_position::DiagnosticReport, server_linter::ServerLinter};
+use super::{
+    error_with_position::DiagnosticReport, isolated_lint_handler::IsolatedLintHandlerOptions,
+    server_linter::ServerLinter,
+};
 
 /// Given a file path relative to the crate root directory, return the URI of the file.
 pub fn get_file_uri(relative_file_path: &str) -> Uri {
@@ -89,11 +92,27 @@ pub struct Tester<'t> {
 
 impl Tester<'_> {
     pub fn new() -> Self {
-        Self { snapshot_suffix: None, server_linter: ServerLinter::new() }
+        Self {
+            snapshot_suffix: None,
+            server_linter: ServerLinter::new(IsolatedLintHandlerOptions {
+                use_cross_module: false,
+                root_path: std::env::current_dir().expect("could not get current dir"),
+            }),
+        }
     }
 
     pub fn new_with_linter(linter: Linter) -> Self {
-        Self { snapshot_suffix: None, server_linter: ServerLinter::new_with_linter(linter) }
+        Self::new_with_server_linter(ServerLinter::new_with_linter(
+            linter,
+            IsolatedLintHandlerOptions {
+                use_cross_module: false,
+                root_path: std::env::current_dir().expect("could not get current dir"),
+            },
+        ))
+    }
+
+    pub fn new_with_server_linter(server_linter: ServerLinter) -> Self {
+        Self { snapshot_suffix: None, server_linter }
     }
 
     pub fn with_snapshot_suffix(mut self, suffix: &'static str) -> Self {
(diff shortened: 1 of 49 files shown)


Sysix:
* **#10268** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10268).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camchenry:
Just commenting here to say I think this is generally the right direction: we should use the same runtime as `oxlint`, which should also enable us to use the cross-module analysis for import plugin and such. We'll probably need to implement a walker for getting all lintable files though. Then, the other hard part is determining which files need to be re-linted based on which file was saved.

Sysix:
> We'll probably need to implement a walker for getting all lintable files though. Then, the other hard part is determining which files need to be re-linted based on which file was saved.

I thought about that. My preferred way is to handle it inside the runtime. 
When oxlint uses a command like `npx oxlint /some/deep/folder/` and the files inside using imports like `../<filename>` we should still analyze them.
We can give the runtime with the paths a good hint to pre-work the files, but I do not think we can solve that completely without reading files when needed.
I found already some code in the runtime to diff between only semantics vs lint files. Maybe the goal to it is not so far away.

Sysix:
@camchenry I can see the finish line for the first time. '
Can you look please at the architecture if this is OK how I want to implement it :) 
- an oxc_linter Feature Flag
- Runtime linear and not using a Sender/receiver
- Clones

crates/oxc_language_server/src/linter/tester.rs
```diff
@@ -89,11 +92,39 @@ pub struct Tester<'t> {
...
+                linter,
+                IsolatedLintHandlerOptions {
+                    use_cross_module: false,
+                    root_path: std::env::current_dir().expect("could not get current dir"),
+                },
+            ),
+        }
+    }
+
+    pub fn new_with_linter_cross_module(linter: Linter) -> Self {
```
camchenry, line 117: rather than adding lots of `new_*` functions, should we replace this with a `new_with_server_linter` method, and let the user pass in whatever options and linter they want?

crates/oxc_linter/src/service/offset_to_position.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl SpanPosition {
+    pub fn new(line: u32, column: u32) -> Self {
+        Self { line, character: column }
+    }
+
+    pub fn line(&self) -> u32 {
+        self.line
+    }
```
camchenry, line 50: do we need these as methods, or could we just make these fields `pub`?

Boshen (commented):
I'll test and release the lsp changes before merging this.

crates/oxc_language_server/src/linter/tester.rs
Sysix: To be honest, I do not like the current implementation of the Server Linter and co. The test setup reflects it^^.
Added your change and will look into refactoring.

PR #10283, opened 2025-04-07
feat(minifier): remove useless unary operator in mathematical operation

Brooooooklyn:


crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -910,12 +959,14 @@ mod test {
...
-        test("var a, b; 'a' + b", "var a, b; '' + b");
+        test("var a, b; 'a' + b", "var a, b;");
         test_same("var a, b; a + '' + b");
         test("var a, b, c; 'a' + (b === c)", "var a, b, c;");
-        test("var a, b; 'a' + +b", "var a, b; '' + +b"); // can be improved to "var a, b; +b"
+        test("var a, b; 'a' + +b", "var a, b; +b");
         test_same("var a, b; a + ('' + b)");
         test("var a, b, c; a + ('' + (b === c))", "var a, b, c; a + ''");
+        test("Math.floor(+d / 1000)", "Math.floor(d / 1000)");
+        test("2 - +this._x.call(null, node.data)", "2 - this._x.call(null, node.data)");
```
sapphi-red, line 969: This is not safe unless it is known that `this._x.call(null, node.data)` does not return a BigInt.
```js
2 - +0n // TypeError: Cannot convert a BigInt value to a number
2 - 0n // no error
```

sapphi-red (dismissed):

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
sapphi-red: Ah, no, forget about this 😅

Brooooooklyn: I thinks it's the same in this case

CleanShot 2025-04-07 at 14 43 03@2x

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -626,6 +650,31 @@ impl<'a> PeepholeOptimizations {
...
+        state: &mut State,
+        ctx: Ctx<'a, '_>,
+    ) -> bool {
+        match e {
+            Expression::UnaryExpression(unary_expr) => {
+                if unary_expr.operator.is_arithmetic() {
+                    *e = unary_expr.argument.take_in(ctx.ast.allocator);
+                    state.changed = true;
+                }
+            }
```
sapphi-red, line 665: I think the unary `+` cannot be removed for comparisons.
```js
+s > 0 // error when s is bigint
s > 0 // no error
```

sapphi-red: There's a related implementation here: 
https://github.com/oxc-project/oxc/blob/691c2ab0740e2b232dbaced47feb66b2dbb73be8/crates/oxc_minifier/src/peephole/substitute_alternate_syntax.rs#L304-L333
Maybe it can be improved.

crates/oxc_minifier/src/peephole/minimize_expression_in_boolean_context.rs
```diff
@@ -137,8 +137,8 @@ mod test {
         test("!!a ? b : c", "a ? b : c");
         test("if (!!!a);", "a");
         test("Boolean(!!a)", "a");
-        test("if ((a | +b) !== 0);", "a | +b");
-        test("if ((a | +b) === 0);", "a | +b");
+        test("if ((a | +b) !== 0);", "a | b");
+        test("if ((a | +b) === 0);", "a | b");
```
sapphi-red, line 141: This is not safe for `a: 0n, b: 0n` (error -> no error).

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -626,6 +650,31 @@ impl<'a> PeepholeOptimizations {
...
+            Expression::StringLiteral(literal) => {
+                if literal.value.is_empty() {
+                    *e = ctx.ast.number_0();
+                    state.changed = true;
+                    return true;
+                }
+            }
```
sapphi-red, line 672:
```suggestion
            Expression::StringLiteral(literal) => {
                if literal.value.is_empty() {
                    // NOTE: this changes the result value, but it's fine because the return value is not used
                    // (example case: `"" > "-1"`)
                    *e = ctx.ast.number_0();
                    state.changed = true;
                    return true;
                }
            }
```
Let's add a comment to make it clear that this changes the result value.

sapphi-red (changes requested):
see the comments above

PR #10284, opened 2025-04-07
feat(ast)!: Re-introduce `TSEnumBody` AST node

### LINKED ISSUE #10087: ast: `TSEnumBody` should be revived?
Part of #9705 

In #10017, I added `TSEnumBody` as `TSEnumDeclaration.body` for TS-ESLint AST by generating it in the serialization phase.

But in detail it only supports cases like:

```ts
enum Foo {} // Exact one space between id and {
```

It does not support such cases:

```ts
enum Foo{} // No space between id and {
enum Foo     {} // Extra space between id and {
```

To properly support these, it seems we need to add `TSEnumBody` back into the Rust AST.

I'm not sure why `TSEnumBody` was removed in #2509, but it's time to bring it back?

leaysgur:
Fixes #10087 

---

Follow up for #10017 .

This PR eliminates the part where nodes were manually generated for JS-land AST.

NOTE: This node was originally present, but was removed at #2509 and is now being reintroduced to get the correct `Span` position for #9705 .


crates/oxc_codegen/src/gen.rs
```diff
@@ -3754,12 +3754,12 @@ impl Gen for TSEnumDeclaration<'_> {
         p.print_str("enum ");
         self.id.print(p, ctx);
         p.print_space_before_identifier();
-        p.print_curly_braces(self.span, self.members.is_empty(), |p| {
-            for (index, member) in self.members.iter().enumerate() {
+        p.print_curly_braces(self.span, self.body.members.is_empty(), |p| {
+            for (index, member) in self.body.members.iter().enumerate() {
```
overlookmotel, line 3758: Could simplify this a bit by implementing `Gen` for `TSEnumBody`.

Also, is the span wrong?

```diff
- p.print_curly_braces(self.span, self.members.is_empty(), |p| {
+ p.print_curly_braces(self.body.span, self.members.is_empty(), |p| {
```

(that's a question, not a statement - I'm not very familiar with codegen)

overlookmotel (commented):
It would make sense to move the scope from `TSEnumDeclaration` to `TSEnumBody`. But let's leave that for a follow-up PR.

Apart from my notes on codegen, all looks good. Thank you!

crates/oxc_codegen/src/gen.rs
leaysgur: Thanks, I think both are true. Fixed in 31d09f1

leaysgur:
> It would make sense to move the scope from TSEnumDeclaration to TSEnumBody. But let's leave that for a follow-up PR.

🙆🏻

Thanks as always for your thoughtful review~!

overlookmotel (approved):
Looks good to me!

overlookmotel:
@Boshen do you want to take a look? Or merge it?

I've been through it fairly carefully, but I'm not familiar with codegen. It might need a `p.add_source_mapping(self.span);` somewhere?

Dunqing:
> @Boshen or @Dunqing do you want to take a look? Or merge it?
> 
> I've been through it fairly carefully, but I'm not familiar with codegen. It might need a `p.add_source_mapping(self.span);` somewhere?

Me either!

Dunqing (approved):

overlookmotel:
Boshen says no need for `p.add_source_mapping(self.span);` on TS nodes. We have no use for sourcemaps when printing TS at present.

Have rebased on latest main and fixed conflicts. Merging now.

PR #10292, opened 2025-04-07
fix(traverse): `ChildScopeCollector` visit all scopes

### LINKED ISSUE #9702: Traverse: Fix and improve design of `ChildScopeCollector`
`ChildScopeCollector` in `oxc_traverse` has a bug and is poorly designed (both my fault).

`ChildScopeCollector` traverses the AST until it hits an AST node which has a scope. It records the `ScopeId` of that scope in a `Vec<ScopeId>`, and then does not traverse any further from that node.

Result is that it only collects the IDs of *direct* child scopes of the node you start from (and not more deeply nested scopes - grandchild scopes etc).

### The bug

Some nodes enter the scope late or exit it early e.g. `Class`.

https://github.com/oxc-project/oxc/blob/f437a95fbd5f3f6fbde474af54413fe250f346e8/crates/oxc_ast/src/ast/js.rs#L1855-L1923

The current implementation in `ChildScopeCollector` for `Class` is:

https://github.com/oxc-project/oxc/blob/f437a95fbd5f3f6fbde474af54413fe250f346e8/crates/oxc_traverse/src/generated/scopes_collector.rs#L75-L77

The problem is that it's assuming that all child nodes within `Class` are within `Class`'s scope. This is wrong - `decorators` field is outside of the `Class`'s scope.

So in this example the scope of the arrow function within the decorator is a direct child scope, and should be collected, but currently it's not:

```js
@((...args) => decorateIt(args))
class C {}
```

It should be:

```rs
    fn visit_class(&mut self, it: &Class<'a>) {
        self.visit_decorators(&it.decorators);
        self.add_scope(&it.scope_id);
    } 
```

There are probably other similar bugs for other node types where the node's scope does not enclose all its fields.

### The bad design

Currently `ChildScopeCollector` collects child scope IDs into a `Vec`.

This temporary allocation is unnecessary. It would be better if it took a callback function which is called on each scope, to avoid constructing a `Vec`.

Or it could be a trait:

```rs
trait ChildScopesVisitor {
    fn visit_child_scope(&mut self, scope_id: ScopeId);
}

impl<'a, V: ChildScopesVisitor> Visit<'a> for V {
    fn visit_block_statement(&mut self, block: &BlockStatement<'a>) {
        self.visit_child_scope(block.scope_id());
    }

    // ... visitors for all other types which have scopes ...
}
```

Then user would implement as:

```rs
struct MyVisitor;

impl ChildScopesVisitor for MyVisitor {
    fn visit_child_scope(&mut self, scope_id: ScopeId) {
        // Do whatever with `scope_id`
    }
}

let mut visitor = MyVisitor;
visitor::visit_statement(&stmt);
```

### Re-implement in `oxc_ast_tools`

The code which generates `ChildScopeCollector` is currently implemented in JS in `oxc_traverse`. I did this for speed of implementation, but it's not ideal. We should re-implement the generator in Rust in `oxc_ast_tools`.

Probably we should do that first, before making the other changes described above.

overlookmotel:
Fixes #9702.

Fix `ChildScopeCollector` so it doesn't miss out scopes in some cases.

Previously when a struct has its own scope, the visitor does not visit any of the struct's fields. But this is incorrect where the type's scope does not cover all of its fields (the scope starts late / ends early). If the fields outside of the scope may themselves have scopes, then those scopes are direct children of the starting point.

These types are:

* `Class` (`decorators` is outside `Class`'s scope).
* `SwitchStatement` (`discriminant` is outside `SwitchStatement`'s scope).
* `TSConditionalType` (`check_type` and `false_type` are outside `TSConditionalType`'s scope).

Additionally, make the visitor more efficient by not bothering to visit any struct fields / enum variants which don't contain a scope anywhere in their dependencies.


Dunqing (approved):

Dunqing:
### Merge activity

* **Apr 9, 5:53 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10327, opened 2025-04-09
fix(estree): Rename `TSInstantiationExpression`.`type_parameters` to `type_arguments`

leaysgur:
Part of #9705 

~~There seemed to be other nodes that originally has the `type_arguments` property.~~
~~Is it OK to just apply `estree(rename)` here...?~~

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -1595,8 +1595,7 @@ pub struct TSNamespaceExportDeclaration<'a> {
 pub struct TSInstantiationExpression<'a> {
     pub span: Span,
     pub expression: Expression<'a>,
-    #[estree(rename = "typeArguments")]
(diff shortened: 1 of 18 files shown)


crates/oxc_ast/src/ast/ts.rs
```diff
@@ -1595,6 +1595,7 @@ pub struct TSNamespaceExportDeclaration<'a> {
 pub struct TSInstantiationExpression<'a> {
     pub span: Span,
     pub expression: Expression<'a>,
+    #[estree(rename = "typeArguments")]
     pub type_parameters: Box<'a, TSTypeParameterInstantiation<'a>>,
```
Boshen, line 1599: Let's rename this to `type_parameters` directly. The original implementation mixed these up.

leaysgur: 🙆🏻 I will update.

PR #10330, opened 2025-04-09
feat(linter): add import/group-exports rule

huangtiandi1999:
Relates to #1117 
Rule detail：https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/group-exports.md

### DIFF
--- a/crates/oxc_linter/src/rules/import/group_exports.rs
+++ b/crates/oxc_linter/src/rules/import/group_exports.rs
@@ -1,0 +1,370 @@
+use oxc_ast::{
+    AstKind,
+    ast::{ImportOrExportKind, MemberExpression},
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::Span;
+use rustc_hash::FxHashMap;
+
+use crate::{context::LintContext, rule::Rule};
+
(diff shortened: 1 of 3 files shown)


shulaoda (approved):

crates/oxc_linter/src/rules/import/group_exports.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// ```
+    GroupExports,
+    import,
+    style
+);
+
+impl Rule for GroupExports {
+    fn run_once(&self, ctx: &LintContext<'_>) {
+        let semantic = ctx.semantic();
+        let mut modules_source_record: FxHashMap<String, Vec<Span>> = FxHashMap::default();
```
Sysix, line 68: We have `ctx.module_record()` which includes all possible exports. It should be faster to iterate them instead of all nodes.
@camchenry can probably tell a bit more about how to get the references/symbols

huangtiandi1999: Yes i have tried using `ctx.module_record()`, but it does not contain the type export
https://github.com/oxc-project/oxc/blob/426f8cba64e34bffc303bc53c995cb2ff30772c0/crates/oxc_parser/src/module_record.rs#L319-L326

Sysix: my bad!

Sysix (approved):

PR #10346, opened 2025-04-10
refactor(ast): Add `TSEnumMemberName` variant to replace `computed` field

### LINKED ISSUE #10332: Replace `computed` field on `TSEnumMember` with more enum variants on `TSEnumMemberName`
#10092 introduced a `computed` field on `TSEnumMember`, in order to align with TS-ESLint.

https://github.com/oxc-project/oxc/blob/fa97b6cea1720a292d7d568b929eeea47f6759e8/crates/oxc_ast/src/ast/ts.rs#L122-L127

I suggest that we remove this field and instead have more variants on `TSEnumMemberName`:

```diff
pub enum TSEnumMemberName<'a> {
    Identifier(Box<'a, IdentifierName<'a>>),
+   ComputedIdentifier(Box<'a, IdentifierName<'a>>),
    String(Box<'a, StringLiteral<'a>>),
+   ComputedString(Box<'a, StringLiteral<'a>>),
-   TemplateString(Box<'a, TemplateLiteral<'a>>),
+   ComputedTemplateString(Box<'a, TemplateLiteral<'a>>),
}
```

There are 2 reasons I feel this is preferable:

1. Shrink `TSEnumMember` by 8 bytes.
2. Make it impossible to represent an illegal syntax in AST:

```rs
enum Foo {
    // Illegal syntax - template strings must be computed
    `X` = 1,
}
```

`computed: false` is not valid when member name is a `TemplateLiteral`.

---

Note: I had thought that computed `Identifier`s are also illegal (`[A] = 1`), but it appears that's a typechecking error, not a parse error.

Both TS and `@typescript-eslint/parser` accept this syntax when parsing only (type-checking disabled):

[TS Playground](https://www.typescriptlang.org/play/?noCheck=true#code/MYewdgzgLgBAgjAvDARHFBuAUFgpmAVwFsYAxEEGAbyxjpgG04BdJGARgBpb6AhNgEzd6jFAGEUrZAGZh9FABEUbACxy6DAAYBRTVJgBWdTE2lNbAGzcAvliA)
[TS-ESLint AST explorer](https://ast-explorer.dev/#eNo1js0KgzAQhF8lzNki/T1YerA/PkIvbiEh5GDRjZi0UMR371rTPc18M7AzokWBp3mbYIemj8jQC4if3i1g5ULb8MytcOs5RFWqkyKUhCMxseNXpyrv1Uis5OryIfk6W9xZ9CbpmnAhzOk2EcKVIH73b+ibnvN98nmudKUFHARMxDLDy4zfJ4Kznbm7ITSeCYWA1kQXIkG6mL4Yrj30)

leaysgur:
Fixes #10332


crates/oxc_ast/src/serialize.rs
```diff
@@ -951,6 +951,21 @@ impl ESTree for JSXOpeningFragmentConverter<'_> {
 // TS
 // --------------------
 
+/// Serializer for `computed` field of `TSEnumMember`.
+#[ast_meta]
+#[estree(ts_type = "boolean", raw_deser = "false")]
```
leaysgur, line 956: @overlookmotel Would you please help me to implement `raw_deser` here? I'm not sure how to touch `enum` variants in JS.

overlookmotel: Sure. I'll push a commit to this PR.

overlookmotel: Done. It's a dreadful hack!

crates/oxc_ast/src/ast/ts.rs
```diff
@@ -119,21 +119,38 @@ pub struct TSEnumBody<'a> {
...
+/// ## Example
+/// ```ts
+/// enum ValidEnum {
+///   identifier,
+///   'string',
+///   ['computed-string'],
+///   [`computed-template`],
+///   // These are invalid syntax
+///   // `template`,
+///   // [computedIdentifier],
```
overlookmotel, line 143: Turns out I was wrong about this. `[computedIdentifier]` causes a *type check* error, but not a *parser* error, so I think we do need to support it.

Please see bottom of my comment on #10332.

overlookmotel: If you put that as 3rd variant (after `String`), the `raw_deser` implementation I've added will remain valid. If you want to put the variants in a different order, you'd need to change the `raw_deser` implementation.

crates/oxc_ast/src/serialize.rs
leaysgur: Thank you so much! 🙌🏻

crates/oxc_ast/src/ast/ts.rs
leaysgur: > [computedIdentifier] causes a type check error, but not a parser error

Ah, I finally understand!

It turns out that we should fix the current parser implementation, doesn't it?

https://github.com/oxc-project/oxc/blob/63d4aa6ac5cc552712071349059194e6e07f960b/crates/oxc_parser/src/ts/statement.rs#L82

> I'm not sure what static_name should return for ComputedIdentifier. None?

Indeed... 🤔 First, let me check all usages of `static_name'.

overlookmotel: > It turns out that we should fix the current parser implementation, doesn't it?

I think so, yes.

leaysgur: After some research, I found these are also the same category. 😓 

```ts
 const x = 'a';
 enum Foo { [x] }

// Expression
 enum Bar { ['a' + 'b'] }
```

overlookmotel: Good find! Yes, TS-ESLint also parses that without complaint: [AST explorer](https://ast-explorer.dev/#eNolyjEOwjAMBdCrWH/JAFygI4dgwQwm8hCUulESkFCUu9dV16c3kLHgIz9psabScUVx6P+iJ9y05WSHR3e170p3qTToGSTQhcI7vGiyedg8DDYihsZVHlpb2oyxOGTp2jqDbWLul5olMg==)

OK, this is getting too complicated. Let's ignore the whole of this comment chain, and leave it as you have it in this PR. If we find that this fails some of TS-ESLint conformance tests, then we can revisit the decision.

Sorry I have wasted time with this red herring!

leaysgur: > Sorry I have wasted time with this red herring! But thank you for delving into it deeply.

Not at all! Thank you for bringing it up. I appreciate it~😃

leaysgur: NOTE: For the future...

- `TSEnumMember`.`id`: `PropertyNameComputed | PropertyNameNonComputed`
  - https://github.com/typescript-eslint/typescript-eslint/blob/7f23300d0d23e5abbb7b86f8c40bafd7730901e1/packages/ast-spec/src/element/TSEnumMember/spec.ts#L12-L14
- Computed = `Expression`, NonComputed = `Identifier` + `StringLiteral` + `NumberLiteral`
  - https://github.com/typescript-eslint/typescript-eslint/blob/7f23300d0d23e5abbb7b86f8c40bafd7730901e1/packages/ast-spec/src/unions/PropertyName.ts#L11-L15

overlookmotel (approved):

PR #10363, opened 2025-04-11
fix(parser): Handle `JSDocUnknownType` correctly

leaysgur:
Current behavior:

```ts
type G<T> = T;
type C1 = G<?>; // Parse failed 👀
```

TS parses this as `TypeReference > JSDocUnknownType`.

> REPL: https://www.typescriptlang.org/play/?noCheck=true#code/C4TwDgpgBA4gPAFQHxQLxQQbgFCklAYQEY1Y4B+JTIA
> Parser: https://github.com/microsoft/TypeScript/blob/83dc0bb2ed91fe0815ab28dc3ff95fae7425e413/src/compiler/parser.ts#L3847

And also this PR fixes these:

```ts
type X = [
  1?,  // TSOptionalType
  ?1,  // JSDocUnknownType 👀
  ?1?, // JSDocUnknownType 👀
  ?1!, // JSDocUnknownType 👀
];
```

to

```ts
type X = [
  1?,  // TSOptionalType
  ?1,  // JSDocNullableType
  ?1?, // JSDocNullableType > JSDocNullableType
  ?1!, // JSDocNullableType > JSDocNonNullableType
];
```

This will result in the same behavior as TS and ESLint-TS.

### DIFF
--- a/crates/oxc_ast/src/ast/ts.rs
+++ b/crates/oxc_ast/src/ast/ts.rs
@@ -119,21 +119,38 @@ pub struct TSEnumBody<'a> {
 #[ast(visit)]
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
+#[estree(
+    add_fields(computed = TSEnumMemberComputed),
+    field_order(span, id, computed, initializer),
+)]
 pub struct TSEnumMember<'a> {
     pub span: Span,
     pub id: TSEnumMemberName<'a>,
-    pub computed: bool,
     pub initializer: Option<Expression<'a>>,
 }
 
 /// TS Enum Member Name
+///
+/// ## Example
+/// ```ts
+/// enum ValidEnum {
+///   identifier,
+///   'string',
+///   ['computed-string'],
+///   [`computed-template`],
+///   // These are invalid syntax
+///   // `template`,
+///   // [computedIdentifier],
+/// }
+/// ```
 #[ast(visit)]
 #[derive(Debug)]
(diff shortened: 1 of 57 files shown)


leaysgur:
How should I handle diffs in parser coverage...?

This is the same as https://github.com/oxc-project/oxc/pull/10346#discussion_r2037104675 , another one that "doesn't result in a parse error, but does result in a type check error".

Boshen:
Thank you. I think I got lazy with these AST nodes at the time of their writing :-)

--
> How should I handle diffs in parser coverage...?

Add logic to https://github.com/oxc-project/oxc/blob/main/crates/oxc_semantic/src/checker/typescript.rs if you want to pass some typescript tests.

In terms of the correct jargon to use, there is actually no parser error or type check error, they are all syntax errors.

Boshen (approved):

overlookmotel:
> In terms of the correct jargon to use, there is actually no parser error or type check error, they are all syntax errors.

Maybe I've been using the wrong terms, but there seems to be a difference between the two. I don't understand this `JSDocUnknownType` business, but using the example from https://github.com/oxc-project/oxc/issues/10332#issuecomment-2796525747:

```ts
enum Foo {
    [bar()],
}
```

TS parser (and TS-ESLint parser) will both parse this without error - therefore I don't think it's a *syntax error*.

But still it's not valid - you can't transform this to JS. That's what I was calling a "typecheck error" because it only gets flagged by TS when you enable type checking.

Please correct me if I've got this wrong. We're probably going to encounter more of these cases as we continue with TS-ESTree compat work, so would be good to understand how we should be handling them.

Taking off the merge label until we resolve this ambiguity.

Boshen:
They parse (produce an AST) because tsc is a recoverable parser, it's a syntax error regardless where it's reported from.

Boshen:
`noCheck` turns off type checking (stop running the checker), but it's still syntax error regardless of where it's reported from.

PR #10367, opened 2025-04-11
fix(linter): config path resolution when path contains '..' syntax

flowreaction:
Improved the `find_oxlint_config` function to handle paths containing parent references ( ../.. syntax )
Added unit tests to verify behavior for valid and invalid config paths, including cases with parent directory references.


Sysix (changes requested):
canonicalize will generate a `\\?\` prefix, which caused others bugs. And it is a slow syscall:

Try the approach from https://github.com/oxc-project/oxc/pull/9382 with absolute.

flowreaction:
> canonicalize will generate a `\\?\` prefix, which caused others bugs. And it is a slow syscall:
> 
> Try the approach from #9382 with absolute.

oh ok, thanks for the link, i changed it and tests still pass so looks to be the better option 🙂

Sysix (approved):
Thanks!

flowreaction:
@Sysix what is the process from here on? I noticed that this Project uses graphite merge queues. So no need to merge from here right?

PR #10387, opened 2025-04-14
fix(estree): `ExportNamedDeclaration.exportKind` should be `type` for `declare` export

leaysgur:
Part of #9705

### DIFF
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -820,6 +820,27 @@ impl ESTree for ExportAllDeclarationWithClause<'_, '_> {
     }
 }
 
+#[ast_meta]
+#[estree(
+    ts_type = "ImportOrExportKind",
+    raw_deser = "
+        const exportKind = DESER[ImportOrExportKind](POS_OFFSET.export_kind);
+        THIS.declaration?.declare ? 'type' : exportKind
+    "
+)]
+pub struct ExportNamedDeclarationExportKind<'a, 'b>(pub &'b ExportNamedDeclaration<'a>);
+
+impl ESTree for ExportNamedDeclarationExportKind<'_, '_> {
+    fn serialize<S: Serializer>(&self, serializer: S) {
+        if let Some(decl) = &self.0.declaration {
+            if decl.declare() {
+                return ImportOrExportKind::Type.serialize(serializer);
+            }
+        }
+        self.0.export_kind.serialize(serializer);
+    }
+}
+
 // --------------------
 // JSX
 // --------------------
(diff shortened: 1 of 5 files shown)


crates/oxc_ast/src/serialize.rs
```diff
@@ -820,6 +820,28 @@ impl ESTree for ExportAllDeclarationWithClause<'_, '_> {
     }
 }
 
+#[ast_meta]
+#[estree(
+    ts_type = "ImportOrExportKind",
+    raw_deser = "
+        const exportKind = DESER[ImportOrExportKind](POS_OFFSET.export_kind);
+        THIS.declaration?.declare ? 'type' : exportKind
```
leaysgur, line 828:
```suggestion
        THIS.declaration?.declare ? 'type' : THIS.exportKind
```

First I tried this. But this seems to generate invalid JS code as a result. 🤔

crates/oxc_ast/src/serialize.rs
```diff
@@ -820,6 +820,28 @@ impl ESTree for ExportAllDeclarationWithClause<'_, '_> {
...
+    "
+)]
+pub struct ExportNamedDeclarationExportKind<'a, 'b>(pub &'b ExportNamedDeclaration<'a>);
+
+impl ESTree for ExportNamedDeclarationExportKind<'_, '_> {
+    fn serialize<S: Serializer>(&self, serializer: S) {
+        if let Some(decl) = &self.0.declaration {
+            if decl.declare() {
+                ImportOrExportKind::Type.serialize(serializer);
```
Boshen, line 837: I wonder whether this is a parser bug, where `declare`d exports should be `type`.

leaysgur: Thanks for the review.

I tried to fix parser at first, but other tests like codegen, isolated_decls failed, so I switched to this one. 😓 
I will take another look at the parser side.

Boshen: I think we need to fix the parser so that consumers of this AST node knows it's a type export.

To test the theory, we can find a linter rule and double check, e.g.

https://github.com/typescript-eslint/typescript-eslint/blob/fa265a970535e9b8d3ac8bfde9f3a24f1a1ec94a/packages/eslint-plugin/src/rules/consistent-type-exports.ts#L205

I haven't checked, but we may be linting this node incorrectly.

leaysgur: OK, I think I got it this time.  => #10389 

Since it seems that `consistent-type-exports` is a type-checking rule, so we don't have it yet. 👀 
And other rules are not affected by `exportKind`.

PR #10389, opened 2025-04-14
fix(parser): `ExportNamedDeclaration.exportKind` should be `type` for `declare` declaration

leaysgur:
Retry for #10387


Boshen (commented):
Nice! Much cleaner!

PR #10392, opened 2025-04-14
fix(estree): Align `TSMappedType` fields

leaysgur:
Part of #9705 

In this PR, I've tried to align `TSMappedType` with TS-ESLint.

> https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/ast-spec/src/type/TSMappedType/spec.ts

```js
// BEFORE
{
  type: "TSMappedType",
  typeParameter: {
    name: Identifier,
    constraint: TypeNode,
  },
  // ...
}

// AFTER
{
  type: "TSMappedType",  
  key: Identifier,
  constraint: TypeNode,
  // ...
}
```

And also fixed serializer for `optional` and `readonly` types.

They are defined as `boolean | '+' | '-' | undefined`, but we use `true | '+' | '-' | null`.

### DIFF
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -1040,10 +1040,7 @@ impl ESTree for TSMappedTypeKey<'_, '_> {
 // NOTE: Variable `type_parameter` in 'raw_deser' is declared in `TSMappedTypeKey`'s `raw_deser`.
 // They will be concatenated in the generated code.
 #[ast_meta]
-#[estree(
-    ts_type = "TSTypeParameter['constraint']",
-    raw_deser = "type_parameter.constraint"
-)]
+#[estree(ts_type = "TSTypeParameter['constraint']", raw_deser = "type_parameter.constraint")]
 pub struct TSMappedTypeConstraint<'a, 'b>(pub &'b TSMappedType<'a>);
 
 impl ESTree for TSMappedTypeConstraint<'_, '_> {
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -1037,6 +1037,7 @@ impl ESTree for TSMappedTypeKey<'_, '_> {
         self.0.type_parameter.name.serialize(serializer);
     }
 }
+
 // NOTE: Variable `typeParameter` in `raw_deser` is shared between `key` and `constraint` serializers.
 // They will be concatenated in the generated code.
 #[ast_meta]
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -1066,7 +1066,7 @@ impl ESTree for TSMappedTypeModifierOperatorConverter<'_> {
             TSMappedTypeModifierOperator::True => true.serialize(serializer),
             TSMappedTypeModifierOperator::Plus => JsonSafeString("+").serialize(serializer),
             TSMappedTypeModifierOperator::Minus => JsonSafeString("-").serialize(serializer),
-            // This is typed as `undefined`(= key is not present) in TS-ESTree.
+            // This is typed as `undefined` (= key is not present) in TS-ESTree.
             // But we serialize it as `null` to align result in snapshot tests.
             TSMappedTypeModifierOperator::None => Null(()).serialize(serializer),
         }
--- a/crates/oxc_ast/src/serialize.rs
+++ b/crates/oxc_ast/src/serialize.rs
@@ -1026,8 +1026,8 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
 #[estree(
     ts_type = "TSTypeParameter['name']",
     raw_deser = "
-        const type_parameter = DESER[Box<TSTypeParameter>](POS_OFFSET.type_parameter);
-        type_parameter.name
+        const typeParameter = DESER[Box<TSTypeParameter>](POS_OFFSET.type_parameter);
+        typeParameter.name
     "
 )]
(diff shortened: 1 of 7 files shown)


overlookmotel:
This PR is marked as draft, so I've not reviewed yet. Please let me know when it's ready for review.

leaysgur:
Thank you for your concern! 🙏🏻 
There are a couple of points I would like to confirm..., but I had no time to put them together. I will let you know soon!

crates/oxc_ast/src/serialize.rs
```diff
@@ -1021,6 +1021,57 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
...
+        self.0.type_parameter.constraint.serialize(serializer);
+    }
+}
+
+#[ast_meta]
+#[estree(
+    ts_type = "true | '+' | '-' | null",
+    raw_deser = "
+        const operator = uint8[pos];
+        [true, '+', '-', null][operator]
```
leaysgur, line 1057: And I don't even think this is the right way to do it. Sorry, could you please give me some guidance? 🙇🏻

leaysgur:
@overlookmotel Thank you for your patience each time~, please take a look?

crates/oxc_ast/src/serialize.rs
```diff
@@ -1021,6 +1021,57 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
...
+    "
+)]
+pub struct TSMappedTypeKey<'a, 'b>(pub &'b TSMappedType<'a>);
+
+impl ESTree for TSMappedTypeKey<'_, '_> {
+    fn serialize<S: Serializer>(&self, serializer: S) {
+        self.0.type_parameter.name.serialize(serializer);
+    }
+}
+// NOTE: Variable `typeParameter` in `raw_deser` is shared between `key` and `constraint` serializers.
```
overlookmotel, line 1040: Nit: Could you add a line break before this line please?

crates/oxc_ast/src/serialize.rs
```diff
@@ -1021,6 +1021,57 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
...
+        const operator = uint8[pos];
+        [true, '+', '-', null][operator]
+    "
+)]
+pub struct TSMappedTypeModifierOperatorConverter<'a>(pub &'a TSMappedTypeModifierOperator);
+
+impl ESTree for TSMappedTypeModifierOperatorConverter<'_> {
+    fn serialize<S: Serializer>(&self, serializer: S) {
+        match self.0 {
+            TSMappedTypeModifierOperator::True => True(()).serialize(serializer),
```
overlookmotel, line 1065: This could be slightly shortened:

```suggestion
            TSMappedTypeModifierOperator::True => true.serialize(serializer),
```

But it's fine to leave it as it is if you think that's clearer - up to you.

crates/oxc_ast/src/serialize.rs
```diff
@@ -1021,6 +1021,57 @@ impl ESTree for TSModuleDeclarationGlobal<'_, '_> {
...
+            TSMappedTypeModifierOperator::Plus => JsonSafeString("+").serialize(serializer),
+            TSMappedTypeModifierOperator::Minus => JsonSafeString("-").serialize(serializer),
+            // This is typed as `undefined`(= key is not present) in TS-ESTree.
+            // But we serialize it as `null` to align result in snapshot tests.
+            TSMappedTypeModifierOperator::None => TsNull(()).serialize(serializer),
```
overlookmotel, line 1070: Slightly shorter:

```suggestion
            TSMappedTypeModifierOperator::None => Null(()).serialize(serializer),
```

`().serialize(serializer)` also does the same, but I think the more explicit `Null(())` is preferable.

overlookmotel (commented):
Looks good! All my comments are just nits.

overlookmotel:
> @overlookmotel Thank you for your patience each time~, please take a look?

No no no! Thank *you* for doggedly pursuing this effort. Really appreciate it.

crates/oxc_ast/src/serialize.rs
overlookmotel: Applied in 51b7e2b0e6bc54d533c288377b96cfc365d2ae58.

crates/oxc_ast/src/serialize.rs
overlookmotel: Applied in 51b7e2b0e6bc54d533c288377b96cfc365d2ae58.

crates/oxc_ast/src/serialize.rs
overlookmotel: Applied in ef688ae85ef8a341c9057dd1d5f1316a4ad38ac4.

crates/oxc_ast/src/serialize.rs
overlookmotel: Applied in 089df24b877a684db25abd107b283fc18b3c6e62.

overlookmotel:
I realized we're in completely different time zones, and I didn't want for this PR to end up with merge conflicts by the time you'd applied changes, and then waited for me to wake up in the morning and merge it. As my change suggestions were all minor nits, I've taken the liberty of making the changes myself, and will merge this as soon as CI passes.

I hope you don't mind. If anything you don't like in the changes I've made, please feel free to say so and revert them!

overlookmotel:
Now over 80% passing 🎉

leaysgur:
> I realized we're in completely different time zones,

Ah~, indeed! I live in JST(GMT+9) and it is time to start working now. 🫖 

> I hope you don't mind. If anything you don't like in the changes I've made, please feel free to say so and revert them!

I don't mind at all and would rather thank you for doing it instead.

> Now over 80% passing 🎉

🥳

PR #10413, opened 2025-04-15
fix(tasks/lint_rules): use default import for `@typescript-eslint/eslint-plugin`

shulaoda:
Fixed https://github.com/oxc-project/oxc/actions/runs/14454100852/job/40545889650


shulaoda:
Fixed by #10407

PR #10417, opened 2025-04-15
fix(parser): Report errors for duplicate extends/implements clauses (TS1172/TS1175)

leaysgur:
```ts
interface I extends A extends B {}
//                    ^^^^^^^
```

At first I thought this should be done by semantic checker.
However, only the last `extends` or `implements` information was kept in the AST, so I report this during parsing.


leaysgur:
Closing this once, I have a few to think about...

PR #10478, opened 2025-04-17
feat(semantic)!: remove `SymbolFlags::ConstVariable`

Dunqing:


Dunqing:
* **#10478** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10478).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Sorry to be a pain, but can I ask: What's the advantage of removing `SymbolFlags::ConstVariable`? Checking that flag is faster than `self.get_resolved_references(symbol_id).any(Reference::is_write)`.

It's also more accurate. In this case, `x` cannot be mutated, but `any(Reference::is_write)` will say it is:

```js
const x = 1;
const throwError = () => { x = 2; };
```

Dunqing:
> Sorry to be a pain, but can I ask: What's the advantage of removing `SymbolFlags::ConstVariable`? Checking that flag is faster than `self.get_resolved_references(symbol_id).any(Reference::is_write)`.
> 
> It's also more accurate. In these cases, `x` cannot be mutated, but `.any(Reference::is_write)` will say it is:
> 
> ```js
> const x = 1;
> const throwError = () => { x = 2; };
> ```
> 
> ```js
> let f = function x() {
>   x = 123;
> };
> ```

I think we can get the following advantages:

1. Remove unnecessary `SymbolFlags` so that we can shrink the size from `u32` to `u16`
2. As much as possible for make each Symbol have only one flag to reduce mismatched semantic flags reports in the Transformer.
3. Align TypeScript better because the flag does not come from TypeScript's [SymbolFlags](https://github.com/microsoft/TypeScript/blob/15392346d05045742e653eab5c87538ff2a3c863/src/compiler/types.ts#L5886-L5976)

I have checked all usages of `SymbolFlags::ConstVariable`, and I didn't see an unavoidable reason to use this flag. So I tried to remove it and see if performance improves or regresses.

From the benchmark result, we can see no performance hits due to this flag being removed, but it also no performance improvement, which is not what I expected. But memory certainly reduced, I think.

As far as I don't think removing this is worth it, and I am going to leave it for a while, and look back when I have more thought on this.

overlookmotel:
We should be able to reduce `SymbolFlags` to a `u8` if we move away from using `bitflags!` for it.

There are lots of bit patterns which make no sense e.g. `RegularEnum + ConstEnum + EnumMember + Class`. A symbol can only be *one* of these things at a time, so it doesn't require a bit for each - only 2 bits to represent those 4 possibilities (more like an `enum`).

If `SymbolFlags` was limited to combinations which are actually possible, it'd be less error-prone (statically impossible to create a symbol which is both a `RegularEnum` and a `Class`) and would allow squeezing down all the possible states into a single byte.

PR #10492, opened 2025-04-18
refactor(transformer/class-properties): streamline handling scope of instance property initializer

Dunqing:
Move the `instance_inits_scope_id` and `instance_inits_constructor_scope_id` fields from the `ClassProperties` struct to local variables.

Using struct-level fields as temporary variables is unnecessary within three function depths, and you also need to consider when to reset the states, otherwise, you may cause a bug due to using previous temporary information. Moreover, I think it also improves a bit readability.

NOTE: This is not only a refactor, but also along with #10495 and #10493 to prepare #10491.

### DIFF
--- a/crates/oxc_transformer/src/es2022/class_properties/class.rs
+++ b/crates/oxc_transformer/src/es2022/class_properties/class.rs
@@ -239,6 +239,16 @@ impl<'a> ClassProperties<'a, '_> {
             return;
         }
 
+        // Scope that instance property initializers will be inserted into.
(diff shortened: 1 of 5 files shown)


overlookmotel (approved):

overlookmotel:
### Merge activity

* **Apr 22, 7:20 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10494, opened 2025-04-18
fix(linter): fix auto-fix issue for eslint/no-else-return

huangtiandi1999:
Part of fix issue： #10439 

The auto-fix feature removed the comments in this case.
```js
if (a) { return 1 } else { /** comments */ return 2 }
```


Dunqing (changes requested):
This does not look efficient. Can you try to use the `span` of `BlockStatement`?

huangtiandi1999:
> This does not look efficient. Can you try to use the `span` of `BlockStatement`?

do you mean modified like this? 
```rust
- ctx.comments().iter().find()
+ ctx.comments_range(block.span.start..block.span.end).find()
```

Dunqing:
> > This does not look efficient. Can you try to use the `span` of `BlockStatement`?
> 
> do you mean modified like this?
> 
> ```rust
> - ctx.comments().iter().find()
> + ctx.comments_range(block.span.start..block.span.end).find()
> ```

Sorry that I did not explain clearly. I mean that you don't even need to access `comments`, return `span` of BlockStatement is enough because that `BlockStatement` always contains comments. Note that  BlockStatement's span contains `{` and `}`, you need to exclude them.

huangtiandi1999:
Hi @Dunqing
I have updated the `replacement_span` to use the `span` of `BlockStatement` and adjusted the related auto-fix test cases. Could you please review the changes when you have a moment?

crates/oxc_linter/src/rules/eslint/no_else_return.rs
```diff
@@ -224,25 +225,21 @@ fn no_else_return_diagnostic_fix(
         return;
     }
     ctx.diagnostic_with_fix(diagnostic, |fixer| {
-        let target_span = Span::new(prev_span.end, else_content_span.end);
+        // find the "else" word start
+        let else_word_start = ctx
+            .source_range(else_keyword_span)
+            .find("else")
+            .map_or(else_keyword_span.start, |pos| else_keyword_span.start + pos as u32);
+        let target_span = Span::new(else_word_start, else_content_span.end);
```
Dunqing, line 233:
```suggestion
        let target_span = Span::new(else_keyword_span.start, else_content_span.end);
```
Can we simplify the code to the above? I don't understand why we need to `+ pos`. And see the test changes, this will produce many unnecessary spaces, although it doesn't matter.

Dunqing (changes requested):
Great! Now the code looks clearer and simpler. One thing I guess we can make it better, can you take a look?

crates/oxc_linter/src/rules/eslint/no_else_return.rs
huangtiandi1999: Thank you for your review, i will simplify the code.
By the way, the unnecessary spaces are actually caused by use the `span` of `BlockStatement`, which aligns with ESLint's behavior.

crates/oxc_linter/src/rules/eslint/no_else_return.rs
```diff
@@ -660,176 +633,224 @@ fn test() {
...
+            ",
             None,
         ),
-        ("if (foo) { return true; } else { let a; }", "if (foo) { return true; } let a;", None),
     ];
+
+    // let pass = vec![];
+    // let fail = vec![
+    //     ("("if(0)return;else r", "if(0)return; r", None),")
+    // ];
```
Dunqing, line 852:
```suggestion

```

Dunqing (approved):
Looks good to me, we can polish the fixer code later if needed

PR #10516, opened 2025-04-20
fix(parser): fix missing type export in module information

### LINKED ISSUE #10505: parser: module information does not include type exports
I'm using the `oxc-parser` v0.64.0 npm module that's called from TypeScript code running in Node.js 23.10.0 to do some custom import graph analysis. First off, I want to say the performance of oxc is incredible, well done!

I've encountered what I suspect is a bug, but might also be user error on my part: module information for a test file doesn't include type re-exports.

Repro code in TypeScript:

```ts
import oxc from 'oxc-parser';

const code = `export type { Foo } from './a';`;

const result = oxc.parseSync('/project/b.ts', code, {
  sourceType: 'module',
});

console.log(result.module.staticExports);
```

Actual output:

```
[] // empty array
```

Expected output:

An array with an entry describing the type export.

ulrichstark:
Closes #10505

### DIFF
--- a/crates/oxc_parser/src/module_record.rs
+++ b/crates/oxc_parser/src/module_record.rs
@@ -317,11 +317,8 @@ impl<'a> ModuleRecordBuilder<'a> {
     }
 
     fn visit_export_named_declaration(&mut self, decl: &ExportNamedDeclaration<'a>) {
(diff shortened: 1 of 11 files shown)


Boshen:
I'll add some tests.

PR #10532, opened 2025-04-21
fix(parser,linter): consider typescript declarations for named exports

### LINKED ISSUE #10318: linter: `eslint-plugin-import(namespace)` gives false warning for TypeScript enum

0.16.5

### What command did you run?

`oxlint`

```jsonc
{
  "plugins": ["import"],
  "rules": {
    "import/namespace": "error"
  }
}
```

### What happened?

The rule is triggered:

```
  × eslint-plugin-import(namespace): "Foo" not found in imported namespace "./lib".
   ╭─[test.ts:3:24]
 2 │ 
 3 │ export const bar = lib.Foo.BAR;
```

However, `Foo` is defined in `lib.ts`:

```
export enum Foo {
  BAR = 0
}
```

Here's a full repro: https://github.com/timostamm/oxlint-import-namespace-enum

### LINKED ISSUE #10556: parser: module information does not include type, interface, or namespace exports
This one is something of a follow-up to https://github.com/oxc-project/oxc/issues/10505#issuecomment-2818961656 for some other instances where types are not listed in module outputs (sorry for not thinking to check this edge case when I filed the first bug).

```js
import oxc from 'oxc-parser';

const code = `export type Foo = {}

export interface Bar {}

export namespace Baz {}`;

const result = oxc.parseSync('/project/a.ts', code, {
  sourceType: 'module',
});

console.log(result.module.staticExports);
```

Actual output:

```
[] // empty array
```

Expected output:

An array with three entries describing the three type exports.

ulrichstark:
Only VariableDeclaration, FunctionDeclaration and ClassDeclaration were considered when collecting named export declarations for the module record. Now the parser also considers TSTypeAliasDeclaration, TSInterfaceDeclaration, TSEnumDeclaration, TSImportEqualsDeclaration and TSModuleDeclaration named export declarations of the module.

Maybe the other `decl.declaration.is_typescript_syntax()` check in `module_record.rs` also needs to be removed?

Closes #10318
Closes #10556

### DIFF
--- a/tasks/coverage/snapshots/parser_typescript.snap
+++ b/tasks/coverage/snapshots/parser_typescript.snap
@@ -3,7 +3,7 @@ commit: 15392346
 parser_typescript Summary:
 AST Parsed     : 6523/6531 (99.88%)
 Positive Passed: 6511/6531 (99.69%)
-Negative Passed: 1308/5754 (22.73%)
+Negative Passed: 1309/5754 (22.75%)
 Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/ClassDeclaration24.ts
 
 Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/ExportAssignment7.ts
@@ -2539,8 +2539,6 @@ Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/jsdocTypeCas
 Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/jsdocTypeNongenericInstantiationAttempt.ts
 
 Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/jsdocTypedefMissingType.ts
-
-Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/jsdocTypedefNoCrash2.ts
 
 Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/jsxCallElaborationCheckNoCrash1.tsx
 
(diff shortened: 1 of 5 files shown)


Boshen:
I'll need to recall the reason for skipping typescript declarations (including merge declarations).

tasks/coverage/snapshots/parser_typescript.snap
```diff
@@ -13415,6 +13413,20 @@ Expect to Parse: tasks/coverage/typescript/tests/cases/conformance/salsa/private
...
+  × Duplicated export 'foo'
+   ╭─[typescript/tests/cases/compiler/jsdocTypedefNoCrash2.ts:1:13]
+ 1 │ export type foo = 5;
+   ·             ─┬─
+   ·              ╰── Export has already been declared here
+ 2 │ /**
+   ╰────
+   ╭─[typescript/tests/cases/compiler/jsdocTypedefNoCrash2.ts:6:14]
+ 5 │  */
+ 6 │ export const foo = 5;
```
Boshen, line 13425: This duplicated error is wrong.

Boshen:
> Maybe the other decl.declaration.is_typescript_syntax() check in module_record.rs also needs to be removed?

Probably yes, although it complicates semantic checks :-/

tasks/coverage/snapshots/parser_typescript.snap
ulrichstark: Do we need to split `exported_bindings` of `ModuleRecord` into `exported_value_bindings` and `exported_type_bindings` to mirror TypeScript's behavior?

If you want me to make this and/or the other change of removing the other `is_typescript_syntax` check, I can do that. If you instead want to take over my PR, that's also no problem.

Boshen: https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/jsdocTypedefNoCrash2.errors.txt

It's reporting the same error 😦

PR #10548, opened 2025-04-22
fix(ast/estree): fix raw deser for `FormalParameter`

overlookmotel:
Follow-on after #10534. Fix the raw deser implementation for `FormalParameter` to handle `TSParameterProperty`.

This involves moving some of the logic into a converter for `FormalParameter`, instead of doing it all in `FormalParameters` converter.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1793,8 +1793,15 @@ pub struct FormalParameters<'a> {
 #[plural(FormalParameterList)]
 #[estree(
     no_type,
+    via = FormalParameterConverter,
     add_ts_def = "
-        interface TSParameterProperty extends Span {
+        type FormalParameter =
+            & ({
+                decorators?: Array<Decorator>;
+            })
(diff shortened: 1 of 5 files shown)


overlookmotel:
@leaysgur Can you please check I have this correct? Or does this foul things up for `TSCallSignatureDeclarationFormalParameters`, `TSMethodSignatureFormalParameters`, and `TSFunctionTypeFormalParameters`?

leaysgur:
Thanks for always cleaning up after me!
(EXCUSE: I also tried to implement it directly for `FormalParameter` at first, but for some reason I couldn't... 🙉)

From my quick check, there did not seem to be any problems.
I did find differences due to other areas than this one, but I will address them again later.

leaysgur (approved):

overlookmotel:
No excuses required! The syntax is so horrible it makes it very difficult. Anyway, we agreed not to worry about `raw_deser` until later anyway. It's nice if we can keep it in sync as we go, but we don't have to.

PR #10557, opened 2025-04-23
fix(estree): Add missing fields for `RestElement` raw deser

leaysgur:
I happened to find this when checking #10548 👀 

(It might not have been necessary to fix it right away, but since I found it 😅)


overlookmotel (approved):
Thank you!

PR #10558, opened 2025-04-23
fix(ast): Box `this_param` in `TSCallSignatureDeclaration`

leaysgur:
All other types of `this_param` were alreadey `Box`ed, but only here it was not.

Because of this, `raw_deser` was behaving wrongly.

https://github.com/oxc-project/oxc/blob/2eb1f88a7d8f382a9a0d87f177e1a043eaf1e32b/crates/oxc_ast/src/serialize.rs#L1197


overlookmotel (approved):
Great!

This also fixes a bug in raw deser. Previously this was wrong for `TSCallSignatureDeclarationFormalParameters`:

```js
const thisParam = DESER[Option<Box<TSThisParameter>>](POS_OFFSET.this_param);
```

But now it's right!

overlookmotel:
I just noticed you already pointed this out in PR description. I read the code, not the description!

PR #10562, opened 2025-04-23
perf(traverse): use `ArenaString` instead `CompactString` to store UID name

### LINKED ISSUE #10545: `TraverseScoping::generate_uid_name` return `Atom` not `CompactStr`
It used to be the case that `Semantic` contained `CompactStr`s, so it made sense to generate UIDs as `CompactStr`s. But now `Semantic` contains only `Atom`s.

We can avoid allocations by constructing UIDs directly in the arena as `ArenaString`s, and returning an `Atom` from `TraverseScoping::generate_uid_name` and friends.

Dunqing:
close: #10545

### DIFF
--- a/crates/oxc_traverse/src/context/scoping.rs
+++ b/crates/oxc_traverse/src/context/scoping.rs
@@ -1,10 +1,9 @@
 use std::str;
 
-use compact_str::CompactString;
 use itoa::Buffer as ItoaBuffer;
 use rustc_hash::FxHashSet;
 
-use oxc_allocator::Vec as ArenaVec;
+use oxc_allocator::{Allocator, String as ArenaString, Vec as ArenaVec};
 use oxc_ast::ast::*;
 use oxc_ast_visit::Visit;
 use oxc_semantic::{NodeId, Reference, Scoping};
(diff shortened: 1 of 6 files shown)


overlookmotel:
bench

Holy smokes!

crates/oxc_traverse/src/context/scoping.rs
```diff
@@ -492,14 +491,14 @@ impl TraverseScoping {
...
-    fn get_uid_names(&self) -> FxHashSet<CompactStr> {
+    fn get_uid_names(&self, allocator: &'a Allocator) -> FxHashSet<&'a str> {
         self.scoping
             .root_unresolved_references()
             .keys()
             .copied()
             .chain(self.scoping.symbol_names())
             .filter(|name| name.as_bytes().first() == Some(&b'_'))
-            .map(CompactStr::from)
+            .map(|str| allocator.alloc_str(str))
```
overlookmotel, line 501: It's really unfortunate we have to do this. But right now with `Scoping` having it's own allocator, I don't see any good alternative.

Maybe it doesn't matter much as most identifiers don't start with `_`, so there'll be fairly few entries in `uid_names`?

Or maybe you were right that it's better to create `Atom`s in `Scoping`'s allocator first? Maybe `uid_names` should live in `Scoping`? (`SemanticBuilder` could populate it)

Ergh... eventually I hope we can combine the 2 arenas into `Allocator`, and all `Atom`s live in the AST arena. That'll avoid copying `Atom`s from one arena to the other in lots of places.

Anyway... this PR already is achieving a big perf gain. So probably best we go with this initial approach for now, and we can experiment with moving `uid_names` into `Scoping` in a follow-up if we want to.

Dunqing: This is unavoidable now. Let's go with this and revisit when `Scoping` reuses the same AST allocator

Dunqing:
I found another possible improvement. Since generating UID name has used `ArenaString`, we can add a `generate_uid_name_arena_string` method to take `ArenaString` rather than `&str` because we still have a few places that use `CompactString` or `format!()` to construct a temporary name to pass in the `generate_uid_name` method. If we have this method, then we can use `ArenaString` to construct and pass in to get a UID name without double allocation.

Boshen:
### Merge activity

* **Apr 26, 10:38 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10588, opened 2025-04-24
perf(parser): remove `-> Result<T>` from all parsing methods

Boshen:
All credits to @overlookmotel

In parser, optimize for common case that file parses successfully.

Instead of parse_* methods returning Result<T>s which are checked with `?` in the caller, always return a valid node T.

In case of a fatal parsing error:

* Record the error in fatal_error property of ParserImpl.
* Also record current length of errors.
* Fast forward the lexer to EOF, so parsing will come to an end swiftly.
* Create a dummy node of the required type and return it.
* At end of parsing, if `fatal_error.is_some()`, then there was a fatal error. In that case:

Truncate errors to the length it was at time of the fatal error (because more errors may occur as the parser finds it's unexpectedly at EOF).
Add fatal_error to errors.
i.e. all parse methods never fail, they just return nonsense if there's a fatal error. This removes all the branches implicit in `?`.

---

Wall clock benchmark:

```
oxc  main ❯ hyperfine './target/release/examples/parser-new ./target/typescript.js' './target/release/examples/parser ./target/typescript.js'
Benchmark 1: ./target/release/examples/parser-new ./target/typescript.js
  Time (mean ± σ):      42.7 ms ±   2.1 ms    [User: 36.7 ms, System: 5.3 ms]
  Range (min … max):    41.6 ms …  59.6 ms    68 runs

Benchmark 2: ./target/release/examples/parser ./target/typescript.js
  Time (mean ± σ):      44.2 ms ±   0.7 ms    [User: 38.3 ms, System: 5.4 ms]
  Range (min … max):    43.7 ms …  48.6 ms    64 runs

Summary
  ./target/release/examples/parser-new ./target/typescript.js ran
    1.04 ± 0.05 times faster than ./target/release/examples/parser ./target/typescript.js
```

/usr/bin/time -al:

```
old: 517424069  instructions retired
new: 492334561  instructions retired
```

uses 5% less CPU instructions.

---

Testing:
* passed https://github.com/oxc-project/monitor-oxc
* passed Rolldown
* Did some fuzzing with https://github.com/oxc-project/fuzz-oxc, found no crashes.

### DIFF
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -28,9 +28,7 @@ impl<'a> ParserImpl<'a> {
 
     #[inline]
     pub(crate) fn end_span(&self, start: u32) -> Span {
-        let end = self.prev_token_end;
-        debug_assert!(end >= start);
-        Span::new(start, end)
+        Span::new(start, self.prev_token_end)
     }
 
     /// Get current token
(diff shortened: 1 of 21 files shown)


crates/oxc_parser/src/cursor.rs
```diff
@@ -28,9 +28,7 @@ impl<'a> ParserImpl<'a> {
 
     #[inline]
     pub(crate) fn end_span(&self, start: u32) -> Span {
-        let end = self.prev_token_end;
-        debug_assert!(end >= start);
```
Boshen, line 32: TODO: add a check in conformance suite.

PR #10597, opened 2025-04-24
fix(linter): cross_module of LintService not being enabled despite enabled import plugin

ulrichstark:
Running `oxlint --import-plugin` with no `oxlintrc.json` didn't enable `cross_module` of `LintService`, because `use_nested_config` was `true` and `nested_configs` empty in that case. This leads to rules like `import/no-cycle` not working.


Sysix (approved):
Nice catch! Thank you ❤️

PR #10614, opened 2025-04-25
perf(parser): mark error paths as cold

DonIsaac:
## What This PR Does
Adds `#[cold]` to parser functions that record errors.

I'm making a draft for now to run benchmarks and see if my hypothesis is correct. Locally I saw some improvement, but my machine gets finicky when it heats up, so it could be noise.

## Hypothesis
LLVM will [consider call sites of `cold` functions to also be `cold`](https://arc.net/l/quote/byvkkmns) (assuming that the function is always called). By marking error recording functions as `#[cold]`, `if` statements that check for diagnostics _should_ be considered cold themselves, thus hinting to branch predictors that the happy path is more likely. This should give branch prediction a higher success rate and lower the odds the CPU's pipeline gets thrown out.

e.g.
```rs
fn parse_foo(&self) -> FooNode {
  if some_bad_thing {
    // this block post-dominates call to `error`, suggesting that it won't be called
    // frequently
    self.error(diagnostics::bad_thing_happened());
  }
  // happy path continues
}
```

### DIFF
--- a/crates/oxc_parser/src/error_handler.rs
+++ b/crates/oxc_parser/src/error_handler.rs
@@ -40,6 +40,7 @@ impl<'a> ParserImpl<'a> {
     }
 
     /// Push a Syntax Error
+    #[cold]
     pub(crate) fn error(&mut self, error: OxcDiagnostic) {
         self.errors.push(error);
     }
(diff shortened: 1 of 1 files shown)


Boshen:
I tried in https://github.com/oxc-project/oxc/pull/10599, but I think it makes sense to merge this change.

I just realized codspeed doesn't weigh in branch predictions, so we need to rely on wall clock benchmark + cpu instrustions.

PR #10654, opened 2025-04-28
perf(transformer): optimize inserting var/let statements

Dunqing:
Found while working on #9881, this PR aims to optimize inserting statements by reducing an unnecessary `Vec` allocation.

Previously, `get_var_statement` would collect `var` statement and `let` statement into a `std::vec::Vec` and return it, then immediately `extend`/`insert` to another `ArenaVec`. Considering here only have two elements, we don't need to store them in a `Vec`, just return them, and use them depending on how they are used.

Seems to have very minor improvement, and I guess `transformer/checker.ts` has 0.1 regression is noisy.

image

### DIFF
--- a/crates/oxc_transformer/src/common/var_declarations.rs
+++ b/crates/oxc_transformer/src/common/var_declarations.rs
@@ -214,15 +214,30 @@ impl<'a> VarDeclarationsStore<'a> {
             return;
         }
 
-        if let Some(insert_stmts) = self.get_var_statement(ctx) {
-            stmts.splice(0..0, insert_stmts);
+        if let Some((var_statement, let_statement)) = self.get_var_statement(ctx) {
+            let mut new_stmts = ctx.ast.vec_with_capacity(stmts.len() + 2);
+            match (var_statement, let_statement) {
+                (Some(var_statement), Some(let_statement)) => {
+                    // Insert `var` and `let` statements
+                    new_stmts.extend([var_statement, let_statement]);
+                }
+                (Some(statement), None) | (None, Some(statement)) => {
+                    // Insert `var` or `let` statement
+                    new_stmts.push(statement);
+                }
+                (None, None) => return,
+            }
+            new_stmts.append(stmts);
+            *stmts = new_stmts;
         }
     }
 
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#10654** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10654).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):
It's definitely an improvement not to allocate a temp `Vec`.

What is less good is that we always copy the whole of `stmts` into a new `ArenaVec`. If `stmts` has spare capacity, it'd be preferable to shift up the existing `Statement`s to make room for the new ones at the start, and then move the new statements into the newly-created space. That doesn't reduce the amount of memory copying that needs to happen, but at least it does it without consuming more memory in arena for a brand new `Vec`.

However, I was surprised to see that `Vec::splice` was doing the same unoptimal thing - allocate a new `Vec` and copy everything over. `Vec::insert` is also surprisingly bad! If the `Vec` is already full to capacity, it copies all the existing elements of the `Vec` *twice*, which is completely unnecessary.

To get this (and the many similar operations in our code) optimal, we'd need new APIs like:

```rs
impl<'a, T> Vec<'a, T> {
    /// Same as `vec.insert(0, value)`
    fn push_start(&mut self, value: T);
    /// Same as `vec.splice(0..0, iter)`
    fn extend_start<I: IntoIterator<Item = T>>(&mut self, iter: I);
}
```

We could optimize these to use the "shift up" method.

overlookmotel:
### Merge activity

* **Apr 29, 12:00 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10676, opened 2025-04-28
fix(allocator): Add `allocator_api` feature to allow usage of `core::alloc::`

TimDiekmann:
Hi,

The crate `allocator-api2` has a feature flag "nightly" to either use the `#![feature(allocator_api)]`-based types from `core::alloc` or use their own. If `allocator-api2` is used by another crate at the same time which enables the `allocator-api2/nightly` feature, a compile error will happen due to feature flag unification of Cargo as `oxc_-_allocator` will then use the re-exported types from `core::alloc` through `allocator_api2` which will result in the error that the `allocator_api` feature is not enabled.

There is a quite easy fix: Simply add `#![feature(allocator_api)]` to the top of `oxc-allocator`'s `lib.rs`. However, as this requires a nightly compiler, the possibly easiest way to stay compatible is to add a `nightly` feature.

I don't know the exact procedure to contribute to this repository, but as it's effectively a 4-line change I thought I'd just open a PR.

Relevant issue:
- https://github.com/zakarumych/allocator-api2/issues/19


Boshen:
Can you provide some context on how you are using this crate which requires nightly?

TimDiekmann:
It's not specifically that this crate requires a nightly compiler, but as `allocator_api2` does not follow Cargo guidelines about feature unification, it has to be handled by downstream crates. I'll try to summarize it in a quick way:
- `allocator_api2` effectively has this
```rust
// When the `nightly` feature is enabled, use `core::alloc` ...
#[cfg(feature = "nightly")]
pub use core::alloc::*

// ... otherwise use own definitions to allow usage on stable toolchain
#[cfg(not(feature = "nightly"))]
pub use self::alloc::*
```
- As soon one crate is using `allocator_api2` with the `nightly` feature enabled (e.g. `hashbrown` has the option), every other crate depending on `allocator_api2` will have the same set of features enabled
- This means, that `oxc_allocator` effectively will see `core::alloc::Allocator` instead of `allocator_api2::Allocator`
- `core::alloc::Allocator`, however, requires the nightly feature `allocator_api`

Generally, this means that any crate depending on `allocator_api2`, directly or indirectly, needs to handle this case. To handle it it's as easy as simply enabling the feature flag depending on a feature flag.

TimDiekmann:
And a bit more concrete: We utilize the allocator `A` in `Box<T, A>` and use bumpalo as an allocator. This means, we require the feature flag `allocator_api` on `bumpalo`. Similarly, we require the `Allocator` trait from the std-library in `hashbrown`. This implies, that `hashbrown` and `bumpalo` both use `allocator-api2/nightly`, which then implies that `oxc_allocator` fails to compile because that crate is not allowed to use `core::alloc::Allocator`.

Dunqing:
I see the reason. We only have a few places that rely on `allocator-api2,` and those can be replaced later, but not quickly. Is this your blocker? If so, I think we can merge first. We may remove it if we get rid of `allocator-api2` later.

TimDiekmann:
I currently use my branch directly, so I would not see updates from your side. I would prefer to not use a fork, so if we could merge it and if not required anymore removing it would be great!

overlookmotel:
We'll need to keep `allocator_api2` for as long as we use `hashbrown` for `HashMap`s which store their data in an arena (which we do in `Semantic`). Getting rid of that would require replacing `hashbrown` with our own `HashMap` impl - which could have some advantages, but is obviously no small undertaking!

In short: I expect we'll be using `allocator_api2` for a long time.

This PR is a good way to make that workable for downstream users who are on nightly. My only comments are:

1. Can we call the feature `allocator_api` instead of `nightly`? "nightly" is a bit broad.

2. We should document what this feature does, and in what circumstances you'd want to use it. Those docs would explain the problem it's designed to solve, much as @TimDiekmann has in this issue. Those docs should be in `crates/oxc_allocator/lib.rs`. I think it'd be better for those docs to err on side of being lengthy. The difference between `allocator-api` and `allocator-api2` may be a bit confusing for those not familiar with this area.

TimDiekmann:
> The CI failures are because we run `cargo check`, `cargo test`, and `cargo clippy` with `--all-features`. We'd need to change that.
> 
> Maybe add a feature `tests` to each crate which enables all features _except_ ones which rely on nightly? (we'd do that in a separate PR which we'd merge before this one)

I was about to ask about what we do about the CI, it just took me a bit to figure out where `cargo lint` is defined 😄 
Do you want me to open the PR or would you do that?

TimDiekmann:
It was already asked if a new release could be done in https://github.com/zakarumych/allocator-api2/pull/30#issuecomment-2668503530. But even if the crate would be released, `hashbrown` and `bumpalo` also would need an update, which implies that they need to add logic to determine the nightly toolchain and conditionally enable the `allocator_api`. For the same reason, a `[patch.crates-io]` would not work because this also would apply to `hashbrown` and `bumpalo` and it would result in an error that the `allocator-api2/nightly` feature does not exist. 
I think the feature flag would be the easiest way forward actually.

TimDiekmann:
If `allocator_api2`, `hashbrown`, and `bumpalo` got an update, I'm happy to contribute to `oxc` to update the usage as well. The least I could say is that I have quite some experience with the (std) allocator API 😉

overlookmotel:
> I think the feature flag would be the easiest way forward actually.

OK. I see your point.

Thanks for making the changes I requested. Some of the docs for other features wasn't quite correct (our fault - they should have been documented already). I've added docs for them in #10695. Please could you rebase on top of it?

Concerning the CI failures: I'd like to consult with @Boshen on this, as it's not my area. My suggestion is:

> Add a feature to each crate which enables all features except ones which rely on nightly? (we'd do that in a separate PR which we'd merge before this one)

Maybe the best name for that feature is `all_stable`. Then we'd run tests with `cargo test --features all_stable` instead of `cargo test --all-features`.

Unfortunately, Boshen is largely away from desk for next few days, so it may take a little time for him to come back on this.

TimDiekmann:
Ah, great, thank you! Do you squash-merge PRs or do should I squash all commits from my PR?

On the feature name: It seems that `full` is quite established. Derived from that, we could go with `full_stable`. I don't know how you want to expand on the CI and how much you care about nightly toolchains, but it might be worse considering also have a matrix element in the CI to test the nightly channel, though, this is probably a high priority for you.

It's not crucial to have this merged now. I can just use my own branch using (and for simplicity `[patch.crates-io]`), but mid-term it would just be good to have this fix being merged so we can easily update the dependency on `oxc`.

TimDiekmann:
Any updates on the CI issue?

NeuralModder:
@overlookmotel @TimDiekmann A few updates:
- The PR for allocator-api2 (https://github.com/zakarumych/allocator-api2/pull/30) has been merged and a new release has been made
- Hashbrown didn't actually need the nightly feature for allocator-api2 so it was taken care of on their end a while back (https://github.com/rust-lang/hashbrown/pull/606), though they should probably still update the allocator-api2 version
- For bumpalo, there's a PR open to update allocator-api2, in case that's needed

overlookmotel:
Thanks very much @NeuralModder. We're in the process of replacing `bumpalo` with our own allocator - that should be done within a few days.

So, if I'm understanding right, if we bump `allocator-api2` to latest, and we get rid of `bumpalo`, the need for a `nightly` feature goes away. @TimDiekmann Please can you confirm if I have this right?

TimDiekmann:
Yes, as far as I can see, we would not need this anymore then. Instead, we'd need a small build-script to detect the nightly toolchain and conditionally enable the allocator-api feature. I'm happy to adjust this PR to account for this. I'd do something like I did for `error-stack`:
https://github.com/hashintel/hash/blob/9cdaaeca48e921a0e03b87c54c6afdc70b07db4d/libs/error-stack/build.rs#L6-L9

TimDiekmann:
Sure, that works as well, that prevents an additional dependency as well 🙂

TimDiekmann:
Actually, it seems that `rustversion` does not support enabling feature-flags such as `#![rustversion::attr(nightly, feature(allocator_api))]`
```
warning: crate-level attribute should be an inner attribute: add an exclamation mark: `#![foo]`
  --> crates/oxc_allocator/src/lib.rs:23:31
   |
23 | #![rustversion::attr(nightly, feature(allocator_api))]
   |                               ^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_attributes)]` on by default
```

A workaround would be to at least use `rustversion` over `rustc_version` (to avoid additional dependencies), and add a super simple `build.rs` (that's the full file):
```rust
fn main() {
    println!("cargo:rustc-check-cfg=cfg(nightly)");
    if rustversion::cfg!(nightly) {
        println!("cargo:rustc-cfg=nightly");
    }
}
```

which would allow checking `nightly` in `cfg_attr` directly:
```rust
#![cfg_attr(nightly, feature(allocator_api))]
```

But let's wait for `bumpalo` to be removed (or updated).

Cargo.toml
```diff
@@ -211,6 +211,7 @@ project-root = "0.2.2"
 rayon = "1.10.0"
 ropey = "1.6.1"
 rust-lapper = "1.1.0"
+rustversion = "1.0.20"
```
TimDiekmann, line 214: I chose the version which was listed in `Cargo.lock`, so it does not change any other behavior.

crates/oxc_allocator/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(nightly)");
```
TimDiekmann, line 2: This is required to prevent a warning on `#[cfg(nightly)]`. This effectively adds `nightly` to the list of allowed cfg-attributes for the crate.

crates/oxc_allocator/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 //!   Usage of this feature is not advisable, and it will be removed as soon as we're able to.
 
 #![warn(missing_docs)]
+#![cfg_attr(nightly, feature(allocator_api))]
```
TimDiekmann, line 23: Now, we simply use `allocator_api` when we're on nightly. This is the correct usage of `allocator_api2` since version `0.3.0`.

overlookmotel:
Just to update you on where we're at...

We have a prototype of our new allocator which is going to replace `bumpalo`. However, we've hit a snag. We also use `bumpalo::collections::String` in various places, and it doesn't have a generic `A: Allocator` bound - it depends directly on `bumpalo::Bump`.

So before we can replace `bumpalo::Bump` with our own allocator, we need our own `String` type which is a wrapper around `oxc_allocator::Vec` instead of `bumpalo::collections::Vec` (or perhaps replace `String` with a more efficient `StringBuilder` type, since we never actually store `String`s anywhere, they're just temporaries which get converted to a `&'a str` once they've been constructed).

So, in short, we are progressing towards replacing `bumpalo`, but it's taking longer than expected.

Cargo.toml
TimDiekmann:
```suggestion
```

TimDiekmann:
@overlookmotel could you take another look at this PR? I updated the changes to use `std::alloc` over `allocator_api2` on nightly with is now properly working on the nightly channel. The stable channel is unaffected and CI should just pass. I tested that locally with my code to make sure everything is fine.

TimDiekmann:
Actually, it seems that `bumpalo` still causing issues for us, so nevermind probably 😅

overlookmotel:
Just to let you know what the hold-up here is:

We bumped `bumpalo` to latest, but found it caused stack overflow on 1 test case in ecosystem CI. It's not clear at this point whether this is a bug in latest `bumpalo`, or just an incidental side-effect e.g. it doesn't inline a function somewhere, which is enough to exhaust the stack in this specific case (the test case is a *very* deeply nested AST).

Meanwhile, I made progress with replacing `bumpalo` with our own allocator, but not finished yet, and have been tied up with other tasks. I'll come back to it as soon as I can.

overlookmotel:
Update: It seems I was out of date. We've now upgraded to latest `bumpalo` v3.19.0 and the mysterious test fail is gone - possibly related to bumping Rust version too.

However, it seems bumpalo reverted the update to `allocator_api2` 0.3.0, because it was a breaking change. So the problem persists.

We could downgrade to bumpalo v3.18.0 (which *does* use `allocator_api2` 0.3.0), as a stopgap until we replace bumpalo entirely. If we did, would that help you?

TimDiekmann:
Thanks for the update!

V3.18.0 of bumpalo was yanked, so it probably introduce more issues than it solves. We're still on the version of `oxc` from when I opened this PR, I guess waiting a bit more doesn’t hurt too much. 👍
I guess it makes sense to wait until bumpalo either releases v4 or you finished the implementation.

overlookmotel:
Just to say, still not finished ripping out bumpalo and replacing it with our own allocator. I wrote the new `Allocator` implementation, but it's got a weird bug in it somewhere, and not yet got a chance to track it down. It *will* happen at some point in not-too-distant future.

If something else we can do to help with this problem in short term, please let us know.

TimDiekmann:
Thanks for the update @overlookmotel!
We're currently on the old version from when this PR was opened and for now this works for us. As this PR was closed now it would be great if I could be notified when you managed to finish your allocator.

PR #10699, opened 2025-04-29
feat(linter): add react/jsx-equals-spacing rule

huangtiandi1999:
Relates to #1022 
Rule detail：https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-equals-spacing.md

Note: This implementation ignores cases where comments are placed between the attribute name and its value.
```jsx
<App foo /** comments */={2} />
```


camc314 (commented):
@Boshen given the work on the formatter, should we just implement it there? i think the two are going to conflict

Boshen:
yeah this is a stylistic rule that does not belong to the linter.

PR #10707, opened 2025-04-30
fix(estree): Serialize adjusted span for `TSTypePredicate`.`typeAnnotation`

leaysgur:
Part of #9705 

- [x] Rust
- [ ] Raw deser
- [ ] Comments

---

NOTE: I tried to deal with this during serialization, but this may be a parser issue? I'll do some research... #10711


leaysgur:
OK, IMO, it is more appropriate to handle this in the parser, so close this.

PR #10719, opened 2025-04-30
perf(transformer/refresh): `RefreshIdentifierResolver::parse` search string for `.` only once

overlookmotel:
Small optimization. `input.split('.')` searches the string for `.`, so use the result of `split` instead of also calling `input.contains('.')`.

### DIFF
--- a/crates/oxc_transformer/src/jsx/refresh.rs
+++ b/crates/oxc_transformer/src/jsx/refresh.rs
@@ -32,20 +32,20 @@ enum RefreshIdentifierResolver<'a> {
 impl<'a> RefreshIdentifierResolver<'a> {
     /// Parses a string into a RefreshIdentifierResolver
     pub fn parse(input: &str, ast: AstBuilder<'a>) -> Self {
-        if !input.contains('.') {
+        let mut parts = input.split('.');
+
+        let first_part = parts.next().unwrap();
+        let Some(second_part) = parts.next() else {
             // Handle simple identifier reference
             return Self::Identifier(ast.identifier_reference(SPAN, input));
-        }
-
-        let mut parts = input.split('.');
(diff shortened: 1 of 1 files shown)


crates/oxc_transformer/src/jsx/refresh.rs
```diff
@@ -32,20 +32,20 @@ enum RefreshIdentifierResolver<'a> {
...
-            return Self::Identifier(ast.identifier_reference(SPAN, input));
-        }
-
         let mut parts = input.split('.');
+
         let first_part = parts.next().unwrap();
+        let Some(second_part) = parts.next() else {
+            // Handle simple identifier reference
+            return Self::Identifier(ast.identifier_reference(SPAN, input));
+        };
```
Dunqing, line 41: Just curious, is there a way to get the second part first?

Dunqing (approved):

Dunqing:
### Merge activity

* **Apr 30, 8:09 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

crates/oxc_transformer/src/jsx/refresh.rs
overlookmotel: I don't think so. `Split` is an iterator, so you have to consume it in order. `first_part` is just a slice of `input` so it's not costly to generate.

I would also *hope* that compiler knows that `Split` always yields at least 1 item (even `"".split('.')` yields a single item `""`), so it should be able to remove the `.unwrap()` in `let first_part = parts.next().unwrap();`. I don't know that for sure, though.

PR #10791, opened 2025-05-04
perf(parser): use visitor instead of JSON.parse reviver

### LINKED ISSUE #10783: perf(napi/parser): json reviver is slow
From discord:

One year ago this was not the case, so I bisect the perf regression to v0.50.
Parsing @typescript-eslint/types/dist/generated/ast-spec.d.ts takes ~5ms for <= 0.49 but ~20ms for >=0.50.

ArnaudBarre:
Fixes #10783

This requires having a list of ~135 nodes with their non primitive keys. 
Technically this list could be generated alongside the raw deserializer but I don't know this code enough to do it.
I've chosen to inline visitor keys to avoid adding [3 dependencies](https://pkg-size.dev/@typescript-eslint%2Fvisitor-keys) for that.
I didn't find how the main LICENCE was injected during the publish so I couldn't append the bundled LICENSES and choose to use `@licence` comments with links


overlookmotel:
I've squashed to 1 commit and rebased on main, just to fix merge conflicts and make it easier to review.

I'll try and figure out cause of the test failures.

overlookmotel:
I found the problem. Our AST has 2 node types which TS-ESLint doesn't:

* `ParenthesizedExpression`
* `TSParenthesizedType`

Have pushed a commit to fix, and also done a little refactoring to (in my opinion) make the code a bit clearer.

ArnaudBarre:
Thanks for finding that! Changes looks good to me!

PR #10813, opened 2025-05-05
perf(napi/parser): simplify recursion and avoid function calls in fixup visitor

overlookmotel:
Follow-on after #10791, part of #10783.

In the fixup visitor, replace `visitNode` with a more specialized `transformNode` function. The difference is that `transformNode` doesn't need to be passed `fn` as 2nd param.

Also, move the check for whether a node is a `Literal` into `transformNode` to avoid the cost of a function call for every node (most of which are not `Literal`s).

Running benchmarks locally (`pnpm run bench` in `napi/parser`), I'm seeing a 2%-3% speed-up.

### DIFF
--- a/napi/parser/wrap.cjs
+++ b/napi/parser/wrap.cjs
@@ -27,41 +27,41 @@ module.exports.wrap = function wrap(result) {
 
 function jsonParseAst(programJson) {
   const program = JSON.parse(programJson);
-  visitNode(program, transform);
+  transformNode(program);
   return program;
 }
 
-function transform(node) {
-  // Set `value` field of `Literal`s for `BigInt`s and `RegExp`s.
-  // This is not possible to do on Rust side, as neither can be represented correctly in JSON.
-  if (node.type === 'Literal') {
-    if (node.bigint) {
-      node.value = BigInt(node.bigint);
-    }
-    if (node.regex) {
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#10813** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10813?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10813?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10813).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
@ArnaudBarre Since we don't have benchmarks for NAPI in CI (we had to remove them as they took way to long to run), I wonder if you could also run this change on the test cases you're using and confirm if you also see an improvement?

ArnaudBarre:
I didn't saw changes on my tests, only +-1% changes depending on the run, but my test suite is very biased (only TSX)

PR #10816, opened 2025-05-05
feat(formatter): fill in implementations

Boshen:


crates/oxc_formatter/src/write/mod.rs
```diff
@@ -2762,14 +2762,36 @@ impl<'a> FormatWrite<'a> for TSIndexSignature<'a> {
         write!(
             f,
             ["[", self.parameters.first().unwrap(), "]", self.type_annotation, OptionalSemicolon]
-        )?;
+        )
+    }
+}
+
+impl<'a, T, const N: usize> Format<'a> for [&T; N]
```
Boshen, line 2769: Is this necessary? I don't see where it's used.

Dunqing: Removed! This isn't needed now. I tried to use this to solve print `ThisParam` and `FormalParameters` in the same `group`, but it doesn't work.

Dunqing (approved):
I am going to support printing binary-like expressions in the next PR.

Dunqing:
### Merge activity

* **May 17, 12:07 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10821, opened 2025-05-05
fix(ast_tools): generate valid output for `#[estree(skip)]` on enum variants with fields

overlookmotel:
We currently only use `#[estree(skip)]` on enum variants which are fieldless. But fix the generator for `ESTree` so it can also handle enum variants which have fields.


overlookmotel:
Does not affect output, only the codegen, so merging without review.

overlookmotel:
### Merge activity

* **May 5, 1:50 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10823, opened 2025-05-06
fix(transformer/typescript): class fields without initializers and with legacy decorators were not removed when `removeClassFieldsWithoutInitializer=true`

### LINKED ISSUE #10738: transformer: `removeClassFieldsWithoutInitializer=true` should remove fields with legacy decorators
For this input
```ts
class Foo {
  @Decorator
  foo: string | undefined;
}
```
TypeScript outputs ([TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&useDefineForClassFields=true#code/MYGwhgzhAEBiD29oG8BQ1oAEAiBTY8ATmAC5HrQBmiAXNBCYQJYB2A5tAD7QCuLAJrkqtc-ANyoAvkA))
```js
"use strict";
var __decorate = /* omit */;
class Foo {
}
__decorate([
    Decorator
], Foo.prototype, "foo", void 0);
```
(with `useDefineForClassFields=false, experimentalDecorators=true`). The field is removed.
esbuild outputs the same ([esbuild try](https://esbuild.github.io/try/#dAAwLjI1LjMAewogIGxvYWRlcjogJ3RzJywKICB0YXJnZXQ6ICdlczIwMjQnLAogIHRzY29uZmlnUmF3OiB7CiAgICBjb21waWxlck9wdGlvbnM6IHsKICAgICAgdXNlRGVmaW5lRm9yQ2xhc3NGaWVsZHM6IGZhbHNlLAogICAgICBleHBlcmltZW50YWxEZWNvcmF0b3JzOiB0cnVlCiAgICB9CiAgfQp9AGNsYXNzIEZvbyB7CiAgQERlY29yYXRvcgogIGZvbzogc3RyaW5nIHwgdW5kZWZpbmVkOwp9)).

But Oxc outputs
```js
import _decorate from "@oxc-project/runtime/helpers/decorate";
class Foo {
        constructor() {
                this.foo = void 0;
        }
}
_decorate([Decorator], Foo.prototype, "foo", void 0);
```
The field is **not** removed.
[reproduction using transform function exposed from rolldown](https://stackblitz.com/edit/github-xv4yss-zchmbhwp?file=build.mjs)

Note that TypeScript does not remove the field when standard decorators is used (`experimentalDecorators=false`).

Probably this `&& prop.decorators.is_empty()` should only be checked for standard decorators.
https://github.com/oxc-project/oxc/blob/0105253cadca872b38b9827c4d9de452d6293680/crates/oxc_transformer/src/typescript/annotations.rs#L235-L237
babel has [the same condition](https://github.com/babel/babel/blob/952f873a84d8a582a07ddc5b32ab134b264b714c/packages/babel-plugin-transform-typescript/src/index.ts#L172-L179) though.

refs #10576

cc @Dunqing sorry I missed this one

Dunqing:
* close #10738

Needs to handle `remove_class_fields_without_initializer` in `class-properties` as well, because the class fields transformation runs after removing class fields without initializers. We have no way to tell `class-properties` SKIP class fields include decorators, except for adding a `remove_class_fields_without_initializer` state in `class-properties` as well.

I've tried moving instance property transformation from `enter_class` to `exit_class`, but it will break the `async-to-generator` plugin, because the transformation of `async` function/arrow function in between class property value or call expression is different. Although it doesn't affect runtime, it will produce duplicate code like `_this = this`.


Dunqing:
* **#10823** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10823).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
Closing as all changes are included in https://github.com/oxc-project/oxc/pull/10854. It was caused by pushing some commits in #10854 by @overlookmotel, and `Graphite` merged them all in that PR. 🥲

PR #10841, opened 2025-05-07
feat(linter): vitest/valid-title

mhheise:
This rule implementation is almost entirely based on the Jest rule of the same name. I'm happy to extract the common code from each if there's a good place to house it.

It's my first contribution to this repository, so let me know if I'm missing any steps!


Sysix (changes requested):
Thanks for your contribution. ❤️ 

We are not separating the jest/vitest rules which are identical,  
instead, we are implementing them in jest and remapping them for vitest:
https://github.com/oxc-project/oxc/blob/ad4fbf44dbce764c9e71d1c8c3dae440c69f33af/crates/oxc_linter/src/utils/mod.rs#L22-L58

Could you please update the jest rule to include all vitest rules test  + options?
And add the rule in the linked mod.rs file :)

PR #10849, opened 2025-05-07
feat(ast_tools): support `#[estree(prepend_to)]`

overlookmotel:
Add an `#[estree(prepend_to = ...)]` attribute to accompany `#[estree(append_to = ...)]`.

* `#[estree(prepend_to = target)]` prepends the field to *start* of the target field.
* `#[estree(append_to = target)]` appends the field to *end* of the target field.

This also allow both prepending *and* appending to the same field. e.g. combining `TSThisParameter`, `FormalParameter`s, and `BindingRestElement` into a single array.


overlookmotel:
Only alters the ESTree generator, no changes to how AST is serialized. So merging without review.

PR #10854, opened 2025-05-07
feat(transformer/typescript): support transforming class fields into this assignment

### LINKED ISSUE #10649: transformer: a way to transform public class fields with `set` semantics without setting lower targets
Currently `assumption.setPublicClassFields: true` only works when target is set lower than es2022, because the option only has effect when class lowering is done.

But in typescript, setting `useDefineForClassFields: false` would enable that transformation even if the target is higher or equal to es2022. To solve this, rolldown-vite always sets a target lower than es2022, when `useDefineForClassFields: false` is set, and this is not ideal.

[reproduction using transform function exposed from rolldown](https://stackblitz.com/edit/github-xv4yss-ahcp6ug8?file=package.json,build.mjs)

refs https://github.com/oxc-project/oxc/issues/6735#issuecomment-2513866362

Dunqing:
Close https://github.com/oxc-project/oxc/issues/10649


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10854?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10854** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10854?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10854?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10853** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10853?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10852** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10852?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10851** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10851?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10825** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10825?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10823** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10823?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10854).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/typescript/class.rs
```diff
@@ -22,29 +33,90 @@ impl<'a> TypeScript<'a, '_> {
...
+    ///
+    // The computed key transformation behavior is the same as `TypeScript`, computed key assignments are
+    // inserted into a static block rather than Babel that inserts them before class. We follow `TypeScript` just for
+    // simplicity because Babel handles class expression and class declaration differently, which quite troublesome to
+    // implement. Anyway, `TypeScript` is the source of truth for the typescript transformation.
+    //
+    // TODO: Not handling static property yet. Might no need to handle it because static property wouldn't affect the
+    //       constructor body, which means never breaks https://github.com/oxc-project/oxc/issues/9192. And the
+    //       `class-properties` plugin has covered it.
+    pub(super) fn transform_class(&self, class: &mut Class<'a>, ctx: &mut TraverseCtx<'a>) {
```
Dunqing, line 72: Could you take a look at this? Should we support transforming static property key for `useDefineForClassFields: false`?

sapphi-red: I guess static properties also needs to be transformed.
For this input,
```ts
class Base {
    static set foo(v: unknown) { console.log('set!') }
}
class Foo extends Base {
    static foo = 0;
}
```
TypeScript outputs
```js
"use strict";
class Base {
    static set foo(v) { console.log('set!'); }
}
class Foo extends Base {
    static { this.foo = 0; }
}
```
and this outputs `set!`.

If this is not transformed, `set!` won't be output.

crates/oxc_transformer/src/typescript/class.rs
```diff
@@ -22,29 +33,90 @@ impl<'a> TypeScript<'a, '_> {
...
+    ///   constructor() {
+    ///     this.x = 1;
+    ///     this[_y] = 2;
+    ///   }
+    /// }
+    /// ```
+    ///
+    // The computed key transformation behavior is the same as `TypeScript`, computed key assignments are
+    // inserted into a static block rather than Babel that inserts them before class. We follow `TypeScript` just for
+    // simplicity because Babel handles class expression and class declaration differently, which quite troublesome to
```
sapphi-red, line 66:
```suggestion
    // simplicity because Babel handles class expression and class declaration differently, which is quite troublesome to
```

crates/oxc_transformer/src/typescript/class.rs
sapphi-red: Another case that is also good to test is:
```ts
const fields = {
  get name() {
    console.log("field")
    return "field"
  }
}

class Foo {
    static {
      console.log('before')
    }

    static [fields.name] = 0;

    static {
      console.log('after')
    }
}
```
this should output `field`, `before`, `after`.

Dunqing: Thank you for reviewing and bringing this case up! My brain is small, I hadn't thought of this case.😢. Yes, you are right, we should also transform static properties.

overlookmotel (commented):
I'll taken the liberty of adding a couple of commits for small things.

1. Avoid copying existing `Atom`s.
2. Comment nits.

I don't think the implementation for computed keys is quite right. For example, this:

```ts
class C {
  [x()]() {}
  [y()] = 123;
  [z()]() {}
}
```

Will be transformed to something like this:

```js
var _y;
class C {
    constructor() {
        this[_y] = 123;
    }
    static {
        _y = y();
    }
    [x()]() {}
    [z()]() {}
}
```

In original, `x()`, `y()` and `z()` are evaluated in that order. After transform, `z()` will be called before `y()`, because static blocks are evaluated after computed keys.

Typescript handles this by moving the temp var assignment into the next computed key:

```js
var _y;
class C {
    constructor() {
        this[_y] = 123;
    }
    [x()]() {}
    [(_y = y(), z())]() {}
}
```

[TS Playground](https://www.typescriptlang.org/play/?target=99&ssl=5&ssc=2&pln=1&pc=1#code/MYGwhgzhAEDC0G8BQ1oG0AeAKAlAXV0QF8V0BPXPaAXmgEYAmAZgG5S0AvSwhEkoA)

I guess this should be addressed. However, probably methods with computed keys are rare in practice, so if you want to, you could merge this as is, and fix this oddity in a follow-up.

Dunqing:
> I'll taken the liberty of adding a couple of commits for small things.
> 
> 1. Avoid copying existing `Atom`s.
> 2. Comment nits.
> 
> I don't think the implementation for computed keys is quite right. For example, this:
> 
> ```ts
> class C {
>   [x()]() {}
>   [y()] = 123;
>   [z()]() {}
> }
> ```
> 
> Will be transformed to something like this:
> 
> ```js
> var _y;
> class C {
>     constructor() {
>         this[_y] = 123;
>     }
>     static {
>         _y = y();
>     }
>     [x()]() {}
>     [z()]() {}
> }
> ```
> 
> In original, `x()`, `y()` and `z()` are evaluated in that order. After transform, `z()` will be called before `y()`, because static blocks are evaluated after computed keys.
> 
> Typescript handles this by moving the temp var assignment into the next computed key:
> 
> ```js
> var _y;
> class C {
>     constructor() {
>         this[_y] = 123;
>     }
>     [x()]() {}
>     [(_y = y(), z())]() {}
> }
> ```
> 
> [TS Playground](https://www.typescriptlang.org/play/?target=99&ssl=5&ssc=2&pln=1&pc=1#code/MYGwhgzhAEDC0G8BQ1oG0AeAKAlAXV0QF8V0BPXPaAXmgEYAmAZgG5S0AvSwhEkoA)
> 
> I guess this should be addressed. However, probably methods with computed keys are rare in practice, so if you want to, you could merge this as is, and fix this oddity in a follow-up.

Good catch. The thing is going to be crazy. It looks like handling only the property definition is not enough. When the computed key comes up, we need to make sure the evaluation order is the same as before.

Dunqing:
I've converted this PR to a draft, I want to solve that problem in this PR. If you, @overlookmotel, think the others look good, please merge them first, Thanks!

Dunqing:
Thank God! The TypeScript implementation is quite simple to follow.

tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/use-define-for-class-fields-without-class-properties/input.ts
```diff
@@ -0,0 +1,60 @@
...
+	[y] = 1;
+	@dce
+	[z]: number;
+}
+
+class ClsWithConstructor extends Cls {
+	constructor() {
+		console.log("constructor");
+		super();
+	}
```
sapphi-red, line 17:
```suggestion
	constructor() {
		console.log("constructor1");
		super();
		console.log("constructor2");
	}
```
I think it'd be nice to change the test case like this to ensure the assignments are injected right after the `super` call.
[TS playground](https://www.typescriptlang.org/play/?target=9&useDefineForClassFields=false#code/MYGwhgzhAEBCkFNoG8C+AodpIwGIHt9oEAPAFwQDsATGeCJZdaaMaAXmgEZnph9KEMgCcArsDL5hACgCUKXi36D8IBADoQ+AObSA5MqFiJXAIR7ZAbkXQIogA4IZVm4dUatugwKPiyAJnMXFgwWACMObnQMIA)

tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/use-define-for-class-fields-without-class-properties/output.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+  constructor() {
+    console.log("constructor");
+    super();
+    this.y = 1;
+    this[_y2] = 1;
+  }
+}
+
+class StaticCls {
```
sapphi-red, line 30:
```suggestion
class StaticCls {
  static {
    _y3 = y;
  }
```
I guess this is missing here.

tasks/transform_conformance/tests/legacy-decorators/test/fixtures/oxc/use-define-for-class-fields/output.js
```diff
@@ -0,0 +1,10 @@
+class Cls {
+  constructor() {
+    this.y = 1;
+  }
+}
+babelHelpers.decorate([dce], Cls.prototype, "z", void 0);
+
+class StaticCls { }
+babelHelpers.decorate([dce], StaticCls, "z", void 0);
+StaticCls.y = 1;
```
sapphi-red, line 10:
```suggestion
StaticCls.y = 1;
babelHelpers.decorate([dce], StaticCls, "z", void 0);
```
I guess this has to be in the reverse order.
[TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&useDefineForClassFields=false#code/MYGwhgzhAEDKAuZ4EtgGEQwN4CgCQEiKw0AHgFzQB2ArgLYBGApgE4Dc+hSq0AntAF5oARg54AAgBNgTTkR4AvDgF8gA)
(probably this can be tackled separately)

tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/use-define-for-class-fields-without-class-properties/input.ts
Dunqing: Good suggestion!

tasks/transform_conformance/tests/babel-plugin-transform-typescript/test/fixtures/use-define-for-class-fields-without-class-properties/output.js
Dunqing: Good catch! 

Oxc output is:

```js
class StaticCls {
  static {
    x;
    _y3 = y;
    z;
  }
  static {
    this.y = 1;
  }
  static {
    this[_y3] = 1;
  }
}
```

We treat the computed key needs to be kept or not differently from `TypeScript`, I've noted in https://github.com/oxc-project/oxc/blob/993f6518a174bed3ced3cccc4fd93bdb857fce28/crates/oxc_transformer/src/typescript/class.rs#L112-L116. Keep it as what we do in other places.

tasks/transform_conformance/tests/legacy-decorators/test/fixtures/oxc/use-define-for-class-fields/output.js
Dunqing: I am not sure whether it implies runtime, but what is certain is that adjusting the order is not easy and has some cost. I prefer to keep it now.

sapphi-red: Here's a example that the behavior changes.
```ts
function decorator(target: any, propertyKey: string) {
  console.log('y', target.y)
}

class StaticCls {
	static y = 1;
	@decorator
	static z: string;
}
```
[TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&useDefineForClassFields=false#code/GYVwdgxgLglg9mABAEwKYTgJwIZSwCim0wHNUoAuRbMATwBpEAHTOJ1TKWgaVVqoDOUTDDAkAlIgDeAKESIMYAXAA2qAHQq4JfAHJauxkVLl1tcTIC+MmRBXYBAxAGUisCAGEVT2QEghuDAQiLSIALyIAIwA3DK+AAJoGDh4mHEB7ogAXoLCoiSx1kA)

I don't think people would rely on this behavior so I think it's fine to keep it, too.

I noticed that TypeScript doesn't keep the execution order in some cases as well ([example](https://www.typescriptlang.org/play/?useDefineForClassFields=true&experimentalDecorators=true&target=99#code/GYVwdgxgLglg9mABAEwKYTgJwIZSwCim0wHNUoAuRbMATwBpEAHTOJ1TKWgaVVqoDOUTDDAkAlIgDeAKESIMYAXAA2qAHQq4JfAHJauxkVLl1tcTIC+MmRBXYBAxAGUisCAGEVT2QEghuDAQiLSIALyIAIwA3DK+AAJoGDh4mHEB7ogAXoLCoiSx-m5B2QBM4VGx1kA)).

sapphi-red (approved):
I only checked the test cases, but at least that part looked good to me

tasks/transform_conformance/tests/legacy-decorators/test/fixtures/oxc/use-define-for-class-fields/output.js
Dunqing: I see. I think it should handle decorator initialization like handling computed keys does, but still very complicated because it separated between two plugins. Let's keep it now.

Dunqing:
@overlookmotel I am going to merge the stack, feel free to review when you are free.

Dunqing:
### Merge activity

* **May 9, 4:42 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #10834, opened 2025-05-06
refactor(ast)!: simplify `RegExpPattern`

overlookmotel:
`RegExpPattern` was an enum with `Raw`, `Invalid`, and `Pattern` variants. i.e. only the raw string *or* the `Pattern` could be stored, but not both.

This required complications in various places, because if the regexp has been parsed, in order to examine or print it, you have to either convert the `Pattern` back to a `String`, or rely on the `Span` and slice the source text.

The `Invalid` variant is not very useful. It's only created if `parse_regular_expression` option is enabled, in which case an error is reported from the parser anyway.

Instead, always store the pattern string as an `Atom`, and store the parsed pattern as `Option<Box<Pattern>>`.

This simplifies code in various places, and makes printing `RegExpLiteral`s faster in `oxc_codegen` when `parse_regular_expression` is enabled, because it can just print the pattern string, rather than converting a `Pattern` back to a `String`.

PR #10855, opened 2025-05-07
refactor(estree): remove regular expression types from ESTree AST

### LINKED ISSUE #10780: regular_expression AST: `Modifier` type is included in `Node` union but doesn't extend `Span`
This forbid to do something like `(node: Node) => node.start`

I'm not sure if this is either a missing `extends Span` or that it should be dropped from the Node union

overlookmotel:
`Pattern` and other types in `oxc_regular_expression` are not part of ESTree AST. Don't implement `ESTree`, and don't generate raw transfer deserializers for these types. None of that code is actually used.

#10834 moved `Pattern` to an optional field of `RegExpPattern`, which is skipped in ESTree serialization, which enables now removing it from ESTree AST entirely.

This reduces the scope of ESTree serialization.


Boshen (approved):
I'll let @leaysgur take a final look.

leaysgur (approved):
I don't think there's any particular issue as a decision.
(BTW, I'm probably the only person in the world who uses [this](https://github.com/leaysgur/js-regexp-ast-viewer/blob/main/src/oxc_regular_expression_wasm/src/lib.rs#L19), so I might need to think about how to handle. 😅)

I'll also update #10846 accordingly.

Boshen:
### Merge activity

* **May 7, 7:50 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Boshen:
oops the other branch got merged first.

ArnaudBarre:
Thanks for cleaning generated types 🎉

PR #10858, opened 2025-05-07
feat(linter): add import/consistent-type-specifier-style rule

huangtiandi1999:
Relates to https://github.com/oxc-project/oxc/issues/1117
Rule detail：https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/consistent-type-specifier-style.md


camc314 (approved):
thank you!

PR #10879, opened 2025-05-08
feat(allocator/vec2): introduce `Cap(NonMaxU32)` for cap field

Dunqing:


Dunqing:
* **#10879** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10879?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10879?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10879).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (commented):
From what I can see, std lib uses `core::num::niche_types::UsizeNoHighBit`, which is a bit different from `NonMaxUsize`.

https://github.com/rust-lang/rust/blob/e964ccafedcf7a505f90f31370d568e649286176/library/alloc/src/raw_vec/mod.rs#L40

Valid ranges:

* `usize`: `0..=usize::MAX` (`1 << 64` possible values)
* `NonZeroUsize`: `1..=usize::MAX` (`(1 << 64) - 1` possible values)
* `NonMaxUsize`: `0..=(usize::MAX - 1)` (`(1 << 64) - 1` possible values)
* `UsizeNoHighBit`: `0..=(isize::MAX as usize)` (`1 << 63` possible values)

i.e. `U64NoHighBitSet` could have been called `U63`.

Anyway, if we're switching to 4-byte `len` and `capacity`, I imagine we won't gain anything from using `NonMaxU32`, and it has a (small) runtime cost. I suggest just using `u32`.

`u32` already has a more restricted range than `usize`, so we may be able to get some optimizations out of that.

---

By the way, I know I brought up this restricted range `Cap` thing in a conversation we had before. What I was trying to say was it might have explained a small perf loss on a previous `Vec` PR, not that we should try to replicate it. Sorry I was unclear.

Dunqing:
> From what I can see, std lib uses `core::num::niche_types::UsizeNoHighBit`, which is a bit different from `NonMaxUsize`.
> 
> https://github.com/rust-lang/rust/blob/e964ccafedcf7a505f90f31370d568e649286176/library/alloc/src/raw_vec/mod.rs#L40
> 
> Valid ranges:
> 
> * `usize`: `0..=usize::MAX` (`1 << 64` possible values)
> * `NonZeroUsize`: `1..=usize::MAX` (`(1 << 64) - 1` possible values)
> * `NonMaxUsize`: `0..=(usize::MAX - 1)` (`(1 << 64) - 1` possible values)
> * `UsizeNoHighBit`: `0..=(isize::MAX as usize)` (`1 << 63` possible values)
> 
> i.e. `U64NoHighBitSet` could have been called `U63`.
> 
> Anyway, if we're switching to 4-byte `len` and `capacity`, I imagine we won't gain anything from using `NonMaxU32`, and it has a (small) runtime cost. I suggest just using `u32`.
> 
> `u32` already has a more restricted range than `usize`, so we may be able to get some optimizations out of that.
> 
> By the way, I know I brought up this restricted range `Cap` thing in a conversation we had before. What I was trying to say was it might have explained a small perf loss on a previous `Vec` PR, not that we should try to replicate it. Sorry I was unclear.

I probably messed up my local Rust source code, which is why I see std lib uses `core::num::niche_types::U32NotAllOnes`. You are right, there is no meaning if we want to change uszie to u32, so close it.

Issue #9706, opened 2024-05-18
Reduce size of `Vec` to 24


overlookmotel:
`Vec` is currently 32 bytes. It consists of 4 x 8-bytes fields:

* Pointer to `Vec`'s contents (`NonNull<u8>`)
* Length (`usize`)
* Capacity (`usize`)
* Pointer to allocator (`*const Allocator`)

### Reduce to 24 bytes

Make length and capacity fields `u32`.

This is pretty easy.

### Reduce to 16 bytes

If arena allocator always created chunks aligned on 4 GiB boundaries, and stored chunk metadata in a header at start of chunks, the pointer field could do double-duty.

Pointer field would be a pointer to the `Vec`'s contents. But also if you zero out the bottom 32 bits it would give you a pointer to the chunk metadata. The chunk metadata would contain a pointer back to the `Allocator`, with which you can grow/reallocate the `Vec` etc.

This would require replacing bumpalo with a custom allocator.

overlookmotel:
#10884 has shrunk `Vec` to 24 bytes. We have more work to do to shrink it further to 16.

Boshen:
Consider this issue done. Move 16 to backlog.

* https://github.com/oxc-project/oxc/issues/9706

PR #10884, opened 2025-05-08
feat(allocator/vec2): change `len` and `cap` fields from `usize` to `u32`

Dunqing:
Related: 
* #6488 
* #9706

Finished first step(32 -> 24) of #9706.

Change `len` and `cap` fields from `usize` to `u32`, so that the `Vec` size can reduced size to `24`. Also, to avoid some unnecessary unit conversion. I changed many `RawVec` methods to take `u32` rather than `usize`.

The performance looks great now! And we reduced 1%-3% memory usage in the parser.

### DIFF
--- a/crates/oxc_allocator/src/vec2/mod.rs
+++ b/crates/oxc_allocator/src/vec2/mod.rs
@@ -724,7 +724,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// ```
     #[inline]
     pub fn capacity(&self) -> usize {
-        self.buf.cap()
+        self.buf.cap_usize()
     }
 
     /// Reserves capacity for at least `additional` more elements to be inserted
@@ -735,7 +735,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the new capacity overflows `usize`.
+    /// Panics if the new capacity overflows `u32`.
     ///
     /// # Examples
     ///
@@ -762,7 +762,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the new capacity overflows `usize`.
+    /// Panics if the new capacity overflows `u32`.
     ///
     /// # Examples
     ///
@@ -786,7 +786,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the new capacity overflows `usize`.
+    /// Panics if the new capacity overflows `u32`.
     ///
     /// # Examples
     ///
@@ -813,7 +813,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the new capacity overflows `usize`.
+    /// Panics if the new capacity overflows `u32`.
     ///
     /// # Examples
     ///
@@ -848,7 +848,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// assert!(vec.capacity() >= 3);
     /// ```
     pub fn shrink_to_fit(&mut self) {
-        if self.capacity() != self.buf.len {
+        if self.buf.cap() != self.buf.len {
             self.buf.shrink_to_fit(self.buf.len);
         }
     }
@@ -955,7 +955,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// [`clear`]: #method.clear
     /// [`drain`]: #method.drain
     pub fn truncate(&mut self, len: usize) {
-        let current_len = self.buf.len;
+        let current_len = self.buf.len_usize();
         unsafe {
             let mut ptr = self.as_mut_ptr().add(current_len);
             // Set the final length at the end, keeping in mind that
@@ -1105,6 +1105,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Safety
     ///
+    /// - `new_len` must be less than or equal to `u32::MAX`.
     /// - `new_len` must be less than or equal to [`capacity()`].
     /// - The elements at `old_len..new_len` must be initialized.
     ///
@@ -1169,8 +1170,12 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// assert_eq!(b"aaaa", &*vec);
     /// ```
     #[inline]
+    #[expect(clippy::cast_possible_truncation)]
     pub unsafe fn set_len(&mut self, new_len: usize) {
-        self.buf.len = new_len;
+        // `new_len as u32` is safe because caller must ensure that
+        // `new_len` is less than or equal to `u32::MAX`, otherwise
+        // it is UB.
+        self.buf.len = new_len as u32;
     }
 
     /// Removes an element from the vector and returns it.
@@ -1205,7 +1210,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
             // bounds check on hole succeeds there must be a last element (which
             // can be self[index] itself).
             let hole: *mut T = &mut self[index];
-            let last = ptr::read(self.get_unchecked(self.buf.len - 1));
+            let last = ptr::read(self.get_unchecked(self.buf.len_usize() - 1));
             self.buf.len -= 1;
             ptr::replace(hole, last)
         }
@@ -1232,11 +1237,11 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// assert_eq!(vec, [1, 4, 2, 3, 5]);
     /// ```
     pub fn insert(&mut self, index: usize, element: T) {
-        let len = self.len();
+        let len = self.buf.len_usize();
         assert!(index <= len);
 
         // space for the new element
-        if len == self.buf.cap() {
+        if self.buf.len == self.buf.cap() {
             self.buf.grow_one();
         }
 
@@ -1252,7 +1257,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
                 // element.
                 ptr::write(p, element);
             }
-            self.set_len(len + 1);
+            self.buf.len += 1;
         }
     }
 
@@ -1547,7 +1552,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the number of elements in the vector overflows a `usize`.
+    /// Panics if the number of elements in the vector overflows a `u32`.
     ///
     /// # Examples
     ///
@@ -1568,7 +1573,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
             self.buf.grow_one();
         }
         unsafe {
-            let end = self.buf.ptr().add(self.buf.len);
+            let end = self.buf.ptr().add(self.buf.len_usize());
             ptr::write(end, value);
             self.buf.len += 1;
         }
@@ -1606,7 +1611,8 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     ///
     /// # Panics
     ///
-    /// Panics if the number of elements in the vector overflows a `usize`.
+    /// Panics if the number of elements in the vector overflows a `u32`.
+    /// Panics if the length of `Self` and `other` add up to more than `u32::MAX`.
     ///
     /// # Examples
     ///
@@ -1630,13 +1636,24 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     }
 
     /// Appends elements to `Self` from other buffer.
+    ///
+    /// # SAFETY
+    ///
+    /// The caller must ensure that the length of buffer passed in is less than or
+    /// equal to `u32::MAX`.
+    /// The length of the `Self` and `other` buffer add up to must be less than or
+    /// equal to `u32::MAX`.
     #[inline]
+    #[expect(clippy::cast_possible_truncation)]
     unsafe fn append_elements(&mut self, other: *const [T]) {
         let count = (*other).len();
         self.reserve(count);
         let len = self.len();
         ptr::copy_nonoverlapping(other as *const T, self.as_mut_ptr().add(len), count);
-        self.buf.len += count;
+        // `unwrap_unchecked()` is okay because that caller needs to ensure that
+        // the length of the buffer passed in is less than or equal to `u32::MAX`,
+        // otherwise it is UB.
+        self.buf.len = self.buf.len.checked_add(count as u32).unwrap_unchecked()
     }
 
     /// Creates a draining iterator that removes the specified range in the vector
@@ -1753,7 +1770,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// ```
     #[inline]
     pub fn len(&self) -> usize {
-        self.buf.len
+        self.buf.len_usize()
     }
 
     /// Returns `true` if the vector contains no elements.
@@ -1802,7 +1819,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     pub fn split_off(&mut self, at: usize) -> Self {
         assert!(at <= self.len(), "`at` out of bounds");
 
-        let other_len = self.buf.len - at;
+        let other_len = self.buf.len_usize() - at;
         let mut other = Vec::with_capacity_in(other_len, self.buf.bump());
 
         // Unsafely `set_len` and copy items to `other`.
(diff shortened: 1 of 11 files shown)


Dunqing:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10884?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#10884** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10883** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10883?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10879** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/10879?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=10884).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Boshen:
Can you report back memory change https://github.com/oxc-project/bench-javascript-parser-written-in-rust/blob/main/memory.sh

Dunqing:
> Can you report memory change https://github.com/oxc-project/bench-javascript-parser-written-in-rust/blob/main/memory.sh

Not too much

Before:
```shell
./files/cal.com.tsx
oxc 11.8 mb
swc 15.0 mb
biome 0.9 MB

./files/typescript.js
oxc 59.1 mb
swc 84.0 mb
biome 0.9 MB
```

After:
```shell
./files/cal.com.tsx
oxc 11.6 mb
swc 15.0 mb
biome 0.9 mb

./files/typescript.js
oxc 57.3 mb
swc 84.0 mb
biome 0.9 MB
```

overlookmotel:
So 2% memory reduction in `cal.com.tsx`, 3% in `typescript.js`.

In terms of speed, I think it's hard to judge the implications at present.

1. Implementation at present is probably not as optimal as it could be.
2. If we go forwards with this, next step is to reduce the size further to 16 (#9706). Presumably we'll get another gain from that.
3. If we go forwards with this, it lays the groundwork for other optimizations. The gain will grow further.

I've made a suggestion on #10883 for a faster way to get a more optimal implementation without tons of code changes. Maybe we should consider trying that and then assess perf again?

Dunqing:
> So 2% memory reduction in `cal.com.tsx`, 3% in `typescript.js`.
> 
> In terms of speed, I think it's hard to judge the implications at present.
> 
> 1. Current implementation is probably not as optimal as it could be.
> 2. If we go forwards with this, next step is to reduce the size further to 16 ([Reduce size of `Vec` #9706](https://github.com/oxc-project/oxc/issues/9706)). Presumably we'd get another perf gain from that.
> 3. If we go forwards with this, it lays the groundwork for other optimizations. The gain will likely grow further.
> 
> I've made a suggestion on #10883 for a faster way to get a more optimal implementation without tons of code changes. Maybe we should consider trying that and then assess perf again?

Thank you for reviewing this! I've taken your suggestion to re-implement this. Unfortunately, the current implementation has a 2% ~ 3% performance regression in the transformer. Probably caused by a subtle change somewhere, I will investigate a little bit on this. Anyway, your way is good! It is simpler than I thought.

Dunqing:
I may have found a little clue about the performance regression. When only changing `cap` to u32 hurts, exceeding the 4% performance regression in the transformer.

overlookmotel:
A couple of thoughts:

1. I think some of the u32-to-usize / usize-to-u32 conversions are the wrong way around or unnecessary.

e.g. in `push` you have this:

```rs
if self.buf.len == usize_to_u32(self.buf.cap()) {
```

The `cap` field is already `u32`, so converting it from `u32` -> `usize` -> `u32` is unnecessary. This may make no difference, but I think it's worth trying. You could add a `RawVec::cap_u32` method and use that to avoid conversions.

`push`, `extend` and `Index` access are probably the most heavily used APIs, so they're probably the ones to investigate first.

2. Try putting `len` field before `cap`.

Again, this may make no difference, but I think worth a giving a go. `len` is the more often accessed field, so it may be preferable to have it aligned on 8. Getting a 32-bit uint from lower bits of a 64-bit uint is very cheap. Getting the higher bits costs an extra shift operation. Given that these methods are on such a hot path, even a tiny change like that might make a difference.

---

By the way, don't worry about the raw transfer test fails. It's just because the layout of `Vec` is changing and the codegen doesn't know that. That'll be easy to fix once we have something we're happy with.

Dunqing:
> A couple of thoughts:
> 
> 1. I think some of the u32-to-usize / usize-to-u32 conversions are the wrong way around or unnecessary.
> 
> e.g. in `push` you have this:
> 
> ```rust
> if self.buf.len == usize_to_u32(self.buf.cap()) {
> ```
> 
> The `cap` field is already `u32`, so converting it from `u32` -> `usize` -> `u32` is unnecessary. This may make no difference, but I think it's worth trying. You could add a `RawVec::cap_u32` method and use that to avoid conversions.
> 
> `push`, `extend` and `Index` access are probably the most heavily used APIs, so they're probably the ones to investigate first.
> 
> 2. Try reversing order of `len` and `cap` fields.
> 
> Again, this may make no difference, but I think worth a giving a go. `len` is the more often accessed field, so it may be preferable to have it aligned on 8. Getting a 32-bit uint from lower bits of a 64-bit uint is very cheap. Getting the higher bits costs an extra shift operation. Given that these methods are on such a hot path, even a tiny change like that might make a difference.
> 
> By the way, don't worry about the raw transfer test fails. It's just because the layout of `Vec` is changing and the codegen doesn't know that. That'll be easy to fix once we have something we're happy with.

Thank you! Inspired by you, I've re-implemented some code, now the performance has turned positive.

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -748,7 +748,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// assert!(vec.capacity() >= 11);
     /// ```
     pub fn reserve(&mut self, additional: usize) {
-        self.buf.reserve(self.buf.len, additional);
+        self.buf.reserve(self.buf.len, usize_to_u32(additional));
```
overlookmotel, line 751: `usize_to_u32` just truncates `additional` if it's greater than `u32:MAX`:

* `usize_to_u32(1 << 32) == 0`
* `usize_to_u32((1 << 32) + 123) == 123`

This is dangerous because code may rely on that if you call `vec.reserve(n)`, there is space available for at least `n` further elements. But if `n > u32::MAX`, this isn't true.

This problem is present in other code in `Vec` itself. e.g. in `extend_from_slice_copy`:

https://github.com/oxc-project/oxc/blob/27a65f7a8e9de914279e43d2723b19fe0cf7ede3/crates/oxc_allocator/src/vec2/mod.rs#L1982-L1995

Here's how you can exploit this safety hole to write out of bounds:

```rs
const TOO_BIG: usize = 1 << 32;
let origin = std::vec::Vec::from_iter(0u8.repeat_n(TOO_BIG));
let slice = origin.as_slice();
assert!(slice.len() == TOO_BIG);

let mut vec = oxc_allocator::Vec::new_in(allocator);
vec.extend_from_slice_copy(slice);
// Boom! Just wrote 4 GiB of zeros across a random unallocated section of memory.
```

* `extend_from_slice_copy` calls `reserve` with `TOO_BIG` (usize).
* `reserve` uses `usize_to_u32` to convert to `u32`, but `usize_to_u32(TOO_BIG) == 0`, so it doesn't reserve any extra capacity.
* `extend_from_slice_copy` expects that space for `TOO_BIG` elements is now available (but it's not).
* `extend_from_slice_copy` writes `TOO_BIG` elements into memory which isn't reserved. Likely segfault, but may just corrupt memory.

What we want to happen here is that `Vec::reserve` panics, because it's illegal for capacity to exceed `u32::MAX`. Then we never get to the last step where bad things happen.

---

Assuming that we want external APIs to continue using `usize`, and only use `u32` internally, we need to be really careful about converting between the two.

In this case, I think we need to pass `additional` as `usize` to `RawVec::reserve`. `RawVec::needs_to_grow` also needs to take `additional: usize`, and then it can do:

```rs
fn needs_to_grow(&self, len: u32, additional: usize) -> bool {
    additional > (self.cap().wrapping_sub(len) as usize)
}
```

`grow_amortized` also need to take `additional: usize` and panic if new required capacity exceeds `u32::MAX as usize`.

i.e. we stay in `usize`-land until we've done our bounds checks.

---

This same problem is also evident in various other places `usize_as_u32` is used.

Personally, I think it might be best to remove the `usize_as_u32` function and just do `blah as u32` wherever there's a conversion. Clippy's "possible truncation" warnings will be annoying, but they're also helpful because often there *is* a genuine problem where the warning appears.

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -775,7 +775,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// assert!(vec.capacity() >= 11);
     /// ```
     pub fn reserve_exact(&mut self, additional: usize) {
-        self.buf.reserve_exact(self.buf.len, additional);
+        self.buf.reserve_exact(self.buf.len, usize_to_u32(additional));
```
overlookmotel, line 778: Same problem as with `reserve`.

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -1170,7 +1170,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// ```
     #[inline]
     pub unsafe fn set_len(&mut self, new_len: usize) {
-        self.buf.len = new_len;
+        self.buf.len = usize_to_u32(new_len);
```
overlookmotel, line 1173: Add to safety comment that `new_len` must be `<= u32::MAX`. `capacity` is always `<= u32::MAX`, and it's in the safety docs that `new_len` must be `<= self.capacity()`, so that's kind of already covered, but I still think it's worth repeating.

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -1636,7 +1636,7 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
         self.reserve(count);
         let len = self.len();
         ptr::copy_nonoverlapping(other as *const T, self.as_mut_ptr().add(len), count);
-        self.buf.len += count;
+        self.buf.len += usize_to_u32(count);
```
overlookmotel, line 1639: This is sound because `append_elements` is only called by `append`, and append receives a `&mut Vec` (which can't have `len` over `u32::MAX`).

But still it's a bit weird. `append_elements` should have safety docs stating that `other` must not have length over `u32::MAX`. Or we could merge `append_elements` into `append`.

There's also an optimization to be made here. `self` can't have `len > u32::MAX`, and `other` can't either, so `self.len() + other.len()` cannot overflow `usize`. So there's no need for checked addition here, which I think `reserve` does - one less branch.

I suggest we don't try to make these kind of optimizations in this PR, though - there's probably other places similar optimizations can be made, and we can do that in a follow-up.

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -2822,7 +2822,7 @@ impl<'a, 'bump, T> Drain<'a, 'bump, T> {
     unsafe fn move_tail(&mut self, extra_capacity: usize) {
         let vec = self.vec.as_mut();
         let used_capacity = self.tail_start + self.tail_len;
-        vec.buf.reserve(used_capacity, extra_capacity);
+        vec.buf.reserve(usize_to_u32(used_capacity), usize_to_u32(extra_capacity));
```
overlookmotel, line 2825: I'm not sure if this is sound or not. Are `used_capacity ` and `extra_capacity` definitely `<= u32::MAX`? `Drain` is hard to wrap your head around!

crates/oxc_allocator/src/vec2/raw_vec.rs
```diff
@@ -113,7 +122,7 @@ impl<'a, T> RawVec<'a, T> {
                 }
             };
 
-            RawVec { ptr, a, cap, len: 0 }
+            RawVec { ptr, a, cap: usize_to_u32(cap), len: 0 }
```
overlookmotel, line 125: Similar problem to `Vec::reserve`. You call `Vec::with_capacity_in` with size `n` and you get back a `Vec` with a smaller capacity than you asked for. It should panic instead.

Probably `alloc_guard` is the place for this check.

Note: On 32 bit systems (e.g. WASM), max capacity is `i32::MAX` bytes (half of `u32::MAX`) because Rust requires that no allocation can exceed `isize::MAX` bytes (and on 32-bit, `isize::MAX == i32::MAX`).

overlookmotel (approved):
Sorry to say I've found problems!

My comments below aren't an exhaustive list. Basically everywhere there's `usize_to_u32`, there's likely to be a problem.

I suspect the solution will result in diff for `Vec`'s methods being much reduced, because it'll need to pass `usize` into `RawVec`'s methods in most places. And then `RawVec` will need to do most maths in `usize` and only convert to `u32` once it's done bounds checks to make sure that `usize` really can't be `> u32::MAX`.

Hope my string of comments doesn't come across as criticism. I'm really excited to see all this work on `Vec` coming to fruition at last, and a really good payoff in terms of perf.

Dunqing:
I am looking at your review and checking the code, I will make a change in this PR.

crates/oxc_allocator/src/vec2/raw_vec.rs
Dunqing: If I understand correctly, we should change `alloc_guard` to check `alloc_size > ::core::u32::MAX` rather than `alloc_size > ::core::isize::MAX`, right?

https://github.com/oxc-project/oxc/blob/27a65f7a8e9de914279e43d2723b19fe0cf7ede3/crates/oxc_allocator/src/vec2/raw_vec.rs#L752-L759

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: I found a unsound case when `Self` and `other` both have length (u32::MAX - 1), then `self.buf.len += count` overflows. I will add a safety comment here. We could merge `append_elements` into `append` later.

For the optimization, I will add a comment to help us revisit it

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: It is sound!
https://github.com/oxc-project/oxc/blob/7c5d8afddafdebdca2591547290959a5ee25941c/crates/oxc_allocator/src/vec2/mod.rs#L2843-L2845

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: Yes, I added a safety comment here.

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: Done!

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: Done!

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: You are right; it would be unsound if we converted usize to u32 early. I removed `usize_as_u32` and `u32_as_usize` functions, and changed to `as usize` and `as u32` in place when needed. That's a good way to avoid an unsound conversion, because we can check one by one if it is sound.

Dunqing:
I've refactored the code to ensure that all conversions are sound and added necessary bound checks. Hopefully, it wouldn't hurt performance.

Dunqing:
image

The benchmarks have a 1% performance regression on the transformer, the parser has a 1%-2% performance improvement, considering the parser execution time is higher than the transformer's, so the whole pipeline is faster than before.

overlookmotel (approved):
Merge merge merge merge!

I removed the benchmarks image from PR description, because it's outdated.

overlookmotel:
### Merge activity

* **May 15, 5:50 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

crates/oxc_allocator/src/vec2/mod.rs
overlookmotel: Great. I've checked the conversions, and they all look good now.

overlookmotel:
@Dunqing Just to make sure you saw my comments above. I had to mark them as "resolved" just to be able to merge the PR. But they're not resolved...

Dunqing:
> @Dunqing Just to make sure you saw my comments above. I had to mark them as "resolved" just to be able to merge the PR. But they're not resolved...

Thanks! I saw! I am going to solve the https://github.com/oxc-project/oxc/pull/10884#discussion_r2091967631 problem.

PR #10907, opened 2025-05-08
refactor(language_server): avoid cloning linter by taking reference in LintService

ulrichstark:


Sysix (approved):
Thank you ❤️ 
Forgot the To-Do because I thought this could never be resolved.
Glad to be proven wrong.

PR #8153, opened 2024-12-28
perf(parser): reduce `Token` size to 8 bytes from 16

wan9chi:
- Replace `end: u32` with `len: u16`. Ends of long tokens (which are rare) are stored in `lexer.long_token_ends`;
- Pack bools into bitflags;
- Now that `end` is calculated from `start + len`, `start` must be properly set. In some places they were not. This PR fixes them and adds a `debug-assertion` check in the lexer.

PR #10933, opened 2025-05-10
perf(parser): Explore packed tokens

tomgasson:
Pack the parser tokens into a single u128.

I was musing with o3 about compiler performance, and it told me that Hermes used packed tokens to improve parser performance so I thought I'd try and see if that would also work here. In local benchmarks it appears to have a decent improvement.

closes #9926

### DIFF
--- a/crates/oxc_parser/src/lexer/token.rs
+++ b/crates/oxc_parser/src/lexer/token.rs
@@ -4,146 +4,142 @@ use oxc_span::Span;
 
 use super::kind::Kind;
 
-#[derive(Debug, Clone, Copy, Default)]
-pub struct Token {
-    /// Token Kind
-    kind: Kind,
+// Bit layout for u128:
+// - Bits 0-31 (32 bits): `start`
+// - Bits 32-63 (32 bits): `end`
+// - Bits 64-71 (8 bits): `kind` (as u8)
+// - Bit 72 (1 bit): `is_on_new_line`
+// - Bit 73 (1 bit): `escaped`
+// - Bit 74 (1 bit): `lone_surrogates`
+// - Bit 75 (1 bit): `has_separator`
 
-    /// Start offset in source
-    start: u32,
+const START_SHIFT: u32 = 0;
+const END_SHIFT: u32 = 32;
+const KIND_SHIFT: u32 = 64;
+const IS_ON_NEW_LINE_SHIFT: u32 = 72;
+const ESCAPED_SHIFT: u32 = 73;
+const LONE_SURROGATES_SHIFT: u32 = 74;
+const HAS_SEPARATOR_SHIFT: u32 = 75;
 
-    /// End offset in source
-    end: u32,
+const START_MASK: u128 = 0xFFFF_FFFF; // 32 bits
+const KIND_MASK: u128 = 0xFF; // 8 bits
+const END_MASK: u128 = 0xFFFF_FFFF; // 32 bits
 
-    /// Indicates the token is on a newline
-    is_on_new_line: bool,
+const IS_ON_NEW_LINE_FLAG: u128 = 1 << IS_ON_NEW_LINE_SHIFT;
+const ESCAPED_FLAG: u128 = 1 << ESCAPED_SHIFT;
+const LONE_SURROGATES_FLAG: u128 = 1 << LONE_SURROGATES_SHIFT;
+const HAS_SEPARATOR_FLAG: u128 = 1 << HAS_SEPARATOR_SHIFT;
 
-    /// True if the identifier / string / template kinds has escaped strings.
-    /// The escaped strings are saved in [Lexer::escaped_strings] and [Lexer::escaped_templates] by
-    /// [Token::start].
-    ///
-    /// [Lexer::escaped_strings]: [super::Lexer::escaped_strings]
-    /// [Lexer::escaped_templates]: [super::Lexer::escaped_templates]
-    escaped: bool,
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct Token(u128);
 
-    /// True if a string contains lone surrogates.
-    lone_surrogates: bool,
-
-    /// True if for numeric literal tokens that contain separator characters (`_`).
-    ///
-    /// Numeric literals are defined in Section 12.9.3 of the ECMAScript
-    /// standard and include [`Kind::Decimal`], [`Kind::Binary`],
-    /// [`Kind::Octal`], [`Kind::Hex`], etc.
-    has_separator: bool,
-
-    // Padding to fill to 16 bytes.
-    // This makes copying a `Token` 1 x xmmword load & store, rather than 1 x dword + 1 x qword
-    // and `Token::default()` is 1 x xmmword store, rather than 1 x dword + 1 x qword.
-    _padding2: u16,
+impl Default for Token {
+    fn default() -> Self {
+        let mut token = Self(0);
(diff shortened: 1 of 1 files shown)


Boshen:
Hi Tom! Long time no see!

Very promising performance improvements:

image

I'll get some of the API changes into main first to make this PR easier to review.

Previous attempt was https://github.com/oxc-project/oxc/pull/8153 and it made similar API changes I forgot to do that first ...

Boshen:
> I was musing with o3 about compiler performance

Curious how you chatted with o3 ;-)

eyestorefivemstore:
Great implementation of this feature. I particularly like the approach to error handling.

eyestoreoye:
Impressive project structure! The architecture really demonstrates solid design principles.

magic-akari:
After this modification, can it parse this library properly?  
Just to highlight, there's a Base64-encoded WASM file embedded directly in the source code as a very long string.

https://www.npmjs.com/package/@shikijs/engine-oniguruma?activeTab=code
https://cdn.jsdelivr.net/npm/@shikijs/engine-oniguruma@3.4.0/dist/wasm-inlined.mjs

tomgasson:
@magic-akari thanks for that example. I was aware of this new constraint - but struggled to identify where someone would actually have such a large single individual token.

From some iteration locally, it seems that moving to u128 performs similarly.

I also couldn't get stable results between using `start & len` vs `start & end` so I've reverted that to `end` to reduce what's changing.


Screenshot 2025-05-12 at 10 30 08 AM
_baseline: main_

tomgasson:
> Curious how you chatted with o3 ;-)

Just conversationally exploring, rough overall structure
 - I want to build a language from scratch (in rust), ask me what features it should have
 - [answer questions like JS]
 - Read about the following [oxc, hermes, v8, javascriptcore, typescript, flow, hhvm, JVM, c#, go, mold, wild, 1brc, ripgrep, ninja, n2, buck2, bazel, simdjson, simdutf, ada, sep] (filling context with high perf insights)
 - Let's build my language. Starting at the [lexer|parser|printer]
 - [simple example]
 - What techniques is X using which I could use?
 
 
 
_Relevant snippet of convo_

----
> Consider Hermes (Facebook/meta). Are there any tricks we can take from there to apply to our lexer/parser?

Yes — Hermes, Facebook/Meta’s JS engine, is famous for its focus on startup time, memory compactness, and ahead-of-time execution, especially on low-power devices. There are several powerful tricks in its lexer/parser/compiler that you could absolutely adapt to your high-performance parser.

Let’s break them down:

### 1. Compact Token & AST Representation
#### Hermes:
- Uses bit-packed token headers (u32 or u64):
  - kind: 8 bits
  - start_offset: 24 bits
  - flags/metadata: remaining bits

AST nodes are tight structs, often with union-like variants.

Apply to your language:
 - You can bit-pack token metadata:
   - Use u32 with {kind: u8, start: u24}
 - Replace heap-allocated structs with:
   - flat arenas
   - small, Copy AST nodes
   - Possibly #[repr(C)] + enum-discriminant layouts for traversals

Benefit: Faster parsing, smaller memory footprint, better locality

### 2. Pre-parsed Bytecode: Skipping Parse at Runtime
...

> Tell me more about bit-packing token

### 4. Bit-Packed Tokens: What, Why, and How
#### Why: 
 - Your lexer is blazing fast — but each token is 12–24 bytes
 - That adds up fast in large files (>10K tokens)
 - Packing reduces memory, improves cache locality

#### Idea:
Represent a Token as a single u32 or u64:


#### Layout:
Field | Bits | Example
-- | -- | --
Kind | 8 | 0–255 enum
Start offset | 24 | Byte offset
End offset | 24 | Byte offset
Flags | 8 | custom

#### Accessors:
```rust
impl PackedToken {
    fn kind(self) -> TokenKind {
        TokenKind::from_u8((self.0 & 0xFF) as u8)
    }
    fn start(self) -> u32 {
        ((self.0 >> 8) & 0xFFFFFF) as u32
    }
    fn end(self) -> u32 {
        ((self.0 >> 32) & 0xFFFFFF) as u32
    }
}
```
Optionally make TokenKind: u8 with #[repr(u8)] to simplify.

tomgasson:
From CodSpeed it look like this has now shifted much of the work over to the lexer (more so with u128 than u64). This might need a bit more work on the lexing side to improve

overlookmotel (approved):
Great!

On Mac Mini M4, I'm getting this result from benchmarks:

```
lexer/checker.ts        time:   [3.5806 ms 3.5878 ms 3.5953 ms]
                        change: [+2.8823% +3.1912% +3.4929%] (p = 0.00 < 0.05)
                        Performance has regressed.
lexer/cal.com.tsx       time:   [1.4846 ms 1.4899 ms 1.4961 ms]
                        change: [+3.5426% +4.5244% +5.7089%] (p = 0.00 < 0.05)
                        Performance has regressed.
lexer/RadixUI.jsx       time:   [1.5299 µs 1.5328 µs 1.5358 µs]
                        change: [−4.0315% −3.7330% −3.4303%] (p = 0.00 < 0.05)
                        Performance has improved.
lexer/pdf.mjs           time:   [918.41 µs 920.58 µs 923.02 µs]
                        change: [+1.2432% +1.7523% +2.2282%] (p = 0.00 < 0.05)
                        Performance has regressed.
lexer/antd.js           time:   [6.8587 ms 6.8660 ms 6.8756 ms]
                        change: [+3.8498% +3.9941% +4.1547%] (p = 0.00 < 0.05)
                        Performance has regressed.

parser/checker.ts       time:   [7.1211 ms 7.1489 ms 7.1802 ms]
                        change: [−3.8394% −3.3014% −2.7636%] (p = 0.00 < 0.05)
                        Performance has improved.
parser/cal.com.tsx      time:   [3.5647 ms 3.5728 ms 3.5822 ms]
                        change: [−6.5004% −6.2345% −5.9451%] (p = 0.00 < 0.05)
                        Performance has improved.
parser/RadixUI.jsx      time:   [4.3214 µs 4.3289 µs 4.3367 µs]
                        change: [−12.434% −12.188% −11.924%] (p = 0.00 < 0.05)
                        Performance has improved.
parser/pdf.mjs          time:   [2.0171 ms 2.0206 ms 2.0243 ms]
                        change: [−6.4245% −5.8924% −5.4479%] (p = 0.00 < 0.05)
                        Performance has improved.
parser/antd.js          time:   [13.776 ms 13.836 ms 13.905 ms]
                        change: [−4.3647% −3.9299% −3.5010%] (p = 0.00 < 0.05)
                        Performance has improved.
```

+12% perf improvement on the RadixUI benchmark - which I think is the one that's most representative of real-world code that gets thrown at Oxc, as all the other benchmark files are huge.

I'm really surprised that this change nets such a gain. I've previously tried making `Token` `#[repr(C, align(16))]` with padding `u8` fields, so all 16 bytes are initialized. i.e. so the memory layout of `Token` is same as a `u128` and with all the fields in same location as here. I'd have expected the compiler to boil that down to exactly the same assembly as this. But it seems not! 🤷

Boshen:
Thank you @tomgasson! Can't wait to see what you find next.

Conaclos:
> I'm really surprised that this change nets such a gain. I've previously tried making Token #[repr(C, align(16))] with padding u8 fields, so all 16 bytes are initialized. i.e. so the memory layout of Token is same as a u128 and with all the fields in same location as here. I'd have expected the compiler to boil that down to exactly the same assembly as this. But it seems not!

This is the kind of thing you could raise as an issue on the Rust repository.

PR #10934, opened 2025-05-10
feat(napi/transform): expose `env.module`

sxzz:


napi/transform/test/transform.test.ts
```diff
@@ -112,6 +112,30 @@ describe('target', () => {
...
+      "
+    `);
+  });
+
+  it('should configure cjs', () => {
+    const ret = transform('test.js', code, { modules: 'cjs' });
+    expect(ret.errors.length).toBe(0);
+    expect(ret.code).toBeDefined();
+    expect(ret.code).toMatchInlineSnapshot(`
+      "import * as foo from "mod";
```
Boshen, line 133: This is not working as intended because we don't have esm to cjs tranformation yet @Dunqing ?

Boshen:
@sxzz we don't have esm to cjs tranform, why do you need this option exposed?

sxzz:
It's not related to tsdown, but to my side project. esbuild has the ability to transform ESM to CJS, but Oxc can't do this.

Boshen:
Unfortunately we don't have a esm to cjs transformer yet.

PR #10953, opened 2025-05-12
fix(estree): Ensure the same key order for `AssignmentPattern`

leaysgur:
Part of #9705 

With this PR, all AST nodes now have a consistent key order for each node type.
(Checked by https://github.com/leaysgur/oxc_estree_ts-ast-diff-viewer/blob/main/scripts/verify-key-order-oxc.js)

Except for:

- `Literal`: `regex` and `bigint` keys
- `TSModuleDeclaration`: `body` key is missing when `declare module "jq";`


overlookmotel (approved):
Thank you!

PR #10966, opened 2025-05-12
fix(transformer): Always update jsx options from comments

### LINKED ISSUE #10956: transformer: `@jsxRuntime` pragma comment is ignored when `typescript.onlyRemoveTypeImports: true` is set
For the following input, Oxc ignores `@jsxRuntime classic` pragma, when `typescript.onlyRemoveTypeImports: true` is set.
```jsx
/** @jsx h */
/** @jsxRuntime classic */
export const foo = <div></div>
```
[reproduction using transform function exposed from rolldown](https://stackblitz.com/edit/github-xv4yss-ervsp63v?file=build.mjs)

If `typescript.onlyRemoveTypeImports: true` is not set, the output uses classic runtime.

magic-akari:
> [!NOTE]  
> The explanation in the comment is incorrect because the `h` function might be locally defined rather than imported. Removing the conditional check will fix the issue.

- Closes: #10956


Dunqing (approved):
Thanks!

PR #10997, opened 2025-05-13
feat(codegen): rework printing normal / legal / annotation comments

### LINKED ISSUE #10948: `CodegenOptions::print_legal_comment` doesn't make any effect
When I try to print legal comments only with using oxc, I found, with passing

```rust
CodegenOptions {
  comments: false,
  legal_comments: LegalComment::Inline,
  ..
}
```

won't print legal/any comment.

Image

The field is not used.

https://github.com/oxc-project/oxc/blob/a86cbb38812085fb2114793b43a0ebc0f98835c1/crates/oxc_codegen/src/lib.rs#L112

Image

Context:

- Rolldown follows esbuild's way to handle comments.
- Esbuild only preserve legal and annoatation comments.
- Rolldown intend to ensure this behavior too.

Boshen:
fixes #10948

This PR changes how comment options are interpreted.

`CodegenOptions::comments`, `CodegenOptions::annotation_comments` and `CodegenOptions::legal_comments` are now independent of each other.

### `CodegenOptions::comments`

Controls printing of normal comments. At present, only some leading comments are preserved. It does not control legal and annotation comments.

### `CodegenOptions::annotation_comments`

Controls printing of annotation comments

* jsdoc: `/** jsdoc */`                                                 
* pure: `/* #__PURE__ */` and `/* #__NO_SIDE_EFFECTS__ */`              
* webpack: `/* webpackChunkName */`                                     
* vite: `/* @vite-ignore */`                                            
* coverage: `v8 ignore`, `c8 ignore`, `node:coverage`, `istanbul ignore`

### `CodegenOptions::legal_comments`

Controls printing of legal comments

* starts with `//!` or `/*!`.                   
* contains `/* @license */` or `/* @preserve */`

---

Comment generation is also changed:

codegen first builds a comments map by keeping only the needed comments.
Printing functions will then print out comments in their respective positions.

---

This PR reveals some incorrect comment interpretation:  

* https://github.com/oxc-project/oxc/issues/10998

### DIFF
--- a/crates/oxc_ast/src/ast/comment.rs
+++ b/crates/oxc_ast/src/ast/comment.rs
@@ -49,15 +49,14 @@ pub enum CommentAnnotation {
     #[default]
     None = 0,
 
-    /// `/** jsdoc */`
-    /// <https://jsdoc.app>
-    Jsdoc = 1,
-
     /// Legal Comment
     /// e.g. `/* @license */`, `/* @preserve */`, or starts with `//!` or `/*!`.
-    ///
     /// <https://esbuild.github.io/api/#legal-comments>
-    Legal = 2,
+    Legal = 1,
+
+    /// `/** jsdoc */`
+    /// <https://jsdoc.app>
+    Jsdoc = 2,
 
     /// `/* #__PURE__ */`
     /// <https://github.com/javascript-compiler-hints/compiler-notations-spec>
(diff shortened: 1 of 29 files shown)


Boshen:
Tested in Rolldown, things are working as intended.

PR #11056, opened 2025-05-15
perf(ast): re-order struct fields to reduce padding

overlookmotel:
All our AST structs are `#[repr(C)]`, which gives us guaranteed stable memory layouts. This is essential for raw transfer, and helpful for other purposes too.

However, the downside is that we lose out on the automatic field-reordering that default `#[repr(Rust)]` provides, where structs are packed to minimize padding bytes.

This PR implements a similar field-reordering scheme, to get the same effect.

* `oxc_ast_tools` determines optimal field order for compact memory layout.
* `oxc_ast_tools` generates a file (`structs.rs`) which details the desired field order for each type.
* `#[ast]` proc macro follows the instructions in `structs.rs` and re-orders the fields.

This reduces the size of various AST types by 8 bytes. No type now has more than 7 padding bytes (previously e.g. `Function` had 15 bytes of padding).

Possible future improvements:

1. Get Rust Analyser to display original field order.
2. Remove `syn` dependency from `oxc_ast_macros` crate, to improve compile times. It should be possible to do all the AST manipulation using just `proc_macro` crate, by generating more code in `oxc_ast_tools`, so the macro's logic can be simpler.

Note: This PR doesn't add `syn` dependency, it was already there. But it makes it easier to remove it.

PR #11046, opened 2025-05-14
perf(parser): use 8 bits for each `Token` flag

overlookmotel:
Use 8 bits for each of the boolean flags in `Token`, so they can be read/written as `bool`s, avoiding bitwise operations.

PR #11062, opened 2025-05-15
perf(ast): reduce size of `Comment` to 16 bytes

camchenry:
I noticed that most of `Comment` consisted of enums which were only 1 byte in size, but only used a few bits in each byte. There are few enough fields that we can actually store all of them in a single `u16` bit flag, which reduces the size of `Comment` from 24 bytes to 16 bytes.

### DIFF
--- a/crates/oxc_ast/src/ast/comment.rs
+++ b/crates/oxc_ast/src/ast/comment.rs
@@ -81,6 +81,22 @@ pub enum CommentAnnotation {
     CoverageIgnore = 7,
 }
 
+/// State of newlines around a comment.
+#[ast]
+#[generate_derive(CloneIn, ContentEq)]
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq)]
+pub enum CommentNewlines {
+    /// No newlines before or after
+    #[default]
+    None = 0,
+    /// Preceded by a newline
+    Leading = 1,
+    /// Followed by a newline
+    Trailing = 2,
+    /// Preceded and followed by a newline
+    LeadingAndTrailing = 3,
+}
+
 /// A comment in source code.
 #[ast]
 #[generate_derive(CloneIn, ContentEq, ESTree)]
@@ -105,14 +121,10 @@ pub struct Comment {
     #[estree(skip)]
     pub position: CommentPosition,
 
-    /// Whether this comment has a preceding newline.
+    /// Whether this comment has newlines around it.
     /// Used to avoid becoming a trailing comment in codegen.
     #[estree(skip)]
-    pub preceded_by_newline: bool,
-
-    /// Whether this comment has a tailing newline.
-    #[estree(skip)]
-    pub followed_by_newline: bool,
+    pub newlines: CommentNewlines,
 
     /// Comment Annotation
     #[estree(skip)]
@@ -129,8 +141,7 @@ impl Comment {
             attached_to: 0,
             kind,
             position: CommentPosition::Trailing,
-            preceded_by_newline: false,
-            followed_by_newline: false,
+            newlines: CommentNewlines::None,
             annotation: CommentAnnotation::None,
         }
     }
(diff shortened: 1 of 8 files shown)


camchenry:
* **#11062** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11062).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camchenry:
@overlookmotel I don't know if this conflicts too much with https://github.com/oxc-project/oxc/pull/11056, but I had started working on this a bit last night before I saw your PR. I think this should still be valid since this both removes a lot of the padding and also reduces the overall size of this struct. Not sure if it'll affect perf at all though, I'm just trying some things at this point.

crates/oxc_ast/src/ast/comment.rs
```diff
@@ -97,86 +98,134 @@ pub struct Comment {
...
+            | Self::WEBPACK_ANNOTATION.bits()
+            | Self::VITE_ANNOTATION.bits()
+            | Self::COVERAGE_IGNORE_ANNOTATION.bits();
+    }
+}
 
-    /// Whether this comment has a tailing newline.
-    #[estree(skip)]
-    pub followed_by_newline: bool,
+/// Dummy type to communicate the content of `RegExpFlags` to `oxc_ast_tools`.
```
overlookmotel, line 147: I think this comments means to say `CommentFlags`!

overlookmotel, line 147:
```suggestion
/// Dummy type to communicate the content of `CommentFlags` to `oxc_ast_tools`.
```

crates/oxc_ast/src/ast/comment.rs
```diff
@@ -97,86 +98,134 @@ pub struct Comment {
...
+    #[derive(Debug, Default, Clone, Copy, Eq, PartialEq)]
+    pub struct CommentFlags: u16 {
+        /// The comment is a line comment.
+        const LINE = 1 << 0;
+        /// The comment is a block comment.
+        const BLOCK = 1 << 1;
+        /// The comment is a leading comment.
+        const LEADING = 1 << 2;
+        /// The comment is a trailing comment.
+        const TRAILING = 1 << 3;
```
overlookmotel, line 117: Aren't `LINE` and `BLOCK` mutually exclusive, and therefore only 1 bit needed for both? If `BLOCK` bit is set it's a block comment, if it's not set then it's a line comment.

Ditto `LEADING` / `TRAILING`.

camchenry: In the formatter there is some concept of `InlineBlock`, but generally speaking I think you are correct

overlookmotel (commented):
We only need to lose 1 byte to get this type down to 16 bytes.

So rather than doing so much with the `CommentFlags` bitflag set, how about just combining `preceded_by_newline` and `followed_by_newline` into 1 byte?

That's simple enough that you could do it with an `enum`, and avoid complication of the `bitflags!` macro.

```rs
enum CommentNewlines {
    None,
    Leading,
    Trailing,
    LeadingAndTrailing,
}
```

Generated assembly is sometimes more efficient if a struct has no padding, so using all the available bytes can actually perform better than aggressively packing everything into the minimum number of bits. e.g. #11046 got a small perf gain from using *more* bits.

The other advantage is that `preceded_by_newline` and `followed_by_newline` are skipped in ESTree AST, so that avoids writing a custom ESTree serializer, which would be necessary with your current approach. I'm keen to avoid custom serializers as much as we can - they're a pain, and if the type changes later on, they have to be kept in sync.

camchenry:
> Generated assembly is sometimes more efficient if a struct has no padding, so using all the available bytes can actually perform better than aggressively packing everything into the minimum number of bits. e.g. #11046 got a small perf gain from using _more_ bits.

Interesting! Seems worth a try.



> The other advantage is that `preceded_by_newline` and `followed_by_newline` are skipped in ESTree AST, so that avoids writing a custom ESTree serializer, which would be necessary with your current approach. I'm keen to avoid custom serializers as much as we can - they're a pain, and if the type changes later on, they have to be kept in sync.

yeah I was just starting to figure out the serializer part. since we only need to lose 1 byte here, I'm leaning towards dropping the bitflags and keeping the enums, but combining them like you suggested. should hopefully make it a little easier to maintain too.

overlookmotel:
I really hate the `bitflags!` macro! Half the time it doesn't do what we really need anyway.

What we really need is a way to pack enums together. e.g.:

```rs
pub enum Foo {
    A,
    B,
    C,
    D,
}

pub enum Bar {
    E,
    F,
}

#[pack]
pub struct FooAndBar {
    pub foo: Foo,
    pub bar: Bar,
}
```

`#[pack]` would convert `FooAndBar` to something like:

```rs
pub struct FooAndBar(FooAndBarInner);

enum FooAndBarInner {
    A_and_E,
    B_and_E,
    C_and_E,
    D_and_E,
    A_and_F,
    B_and_F,
    C_and_F,
    D_and_F,
}

impl FooAndBar {
    pub fn foo(&self) -> Foo {
        match self {
            Self::A_and_E | Self::A_and_F => Foo::A,
            Self::B_and_E | Self::B_and_F => Foo::B,
            Self::C_and_E | Self::B_and_F => Foo::C,
            Self::D_and_E | Self::B_and_F => Foo::D,
        }
    }

    pub fn bar(&self) -> Bar {
        match self {
            Self::A_and_E | Self::B_and_E | Self::C_and_E | Self::D_and_E => Bar::E,
            Self::A_and_F | Self::B_and_F | Self::C_and_F | Self::D_and_F => Bar::F,
        }
    }
}
```

I've not located a crate which does this, sadly.

camchenry:
Looks like this gets us about the same result in codspeed (+1% in some lexer benchmarks). Running locally, I see a ~1% perf improvement in the parser (on my M1 Air laptop) (numbers are reversed, because I ran in the opposite branch order):

image

overlookmotel (approved):
Great!

Let's merge this.

However, my idea of using an enum instead of `bitflags!` for `CommentNewlines` was misjudged. I hadn't taken into account the complexity of the setters for the 2 flags. `bitflags!` would be simpler, and probably more performant - compiler is surprisingly bad at optimizing operations on fieldless enums.

Here's a comparison between the enum-based implementations and code like what `bitflags!` would produce: https://godbolt.org/z/qT817zo18

The most important ones to compare are the getters and the "real world usage" methods e.g. `set_followed_by_newline_true`.

So, although I do hate `bitflags!` it's still better than what I came up with!

I don't know if you have time/inclination to try making that change in a follow-up PR? Really sorry I sent you on a wild goose-chase with my bitflag-hating. You were on the right track to start with, and I led you astray...

overlookmotel:
Pushed a commit to remove the change to the codegen for raw transfer. That change was I think left over from earlier version where used a `u16` for flags.

overlookmotel:
### Merge activity

* **May 15, 3:26 PM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Issue #11029, opened 2025-05-14
A case playground rejects, but `oxc-parser` successful parsed


fisker:
```ts
class A {
    @decorator()
    abstract method(): Array<string>
}
```

This case should be invalid, and the playground throws as expected, but `parseAsync` `parseSync` from `oxc-parser` all parsed without error.

My `oxc-parser` version is `0.69.0`, I checked the commit history, but didn't find related change.

Is it fixed or fixed accidentally? Should I wait for a newer version of `oxc-parser`?

leaysgur:
> playground throws as expected

It may be a bit tricky, so let me explain.

With the current implementation, `oxc-parser`.`parse(Sync|Async)()` on this input won't throw an exception that can be `try-catch`ed.

Instead, if you set the `showSemanticErrors` option, it will be reported in `errors` array.

In the playground, this option is the **Check Syntax Errors** toggle in the sidebar and just displaying them.

So, you may need to set `showSemanticErrors` and manually throw if `errors` is not empty.

(However, AFAIK, it is important to note that `oxc-parser` and `@typescript-eslint/typescript-estree` do not behave exactly the same in some cases.)

leaysgur:
NOTE: About current playground implementation.

- `Check Syntax Errros` toggle controls `options.run.syntax`
https://github.com/oxc-project/playground/blob/f50c40cf179f15f8d59c5af273b6475576179bc6/src/components/sidebar/Parse.vue#L87
- If `options.syntax` is `true`, `getDiagnostics()` returns parser errors + semantic errors
https://github.com/oxc-project/oxc/blob/c1e6c9bfa2ff505dcc87c98495c7fb78185a217d/napi/playground/src/lib.rs#L148-L150
- If this is `false`, parser errors are simply discarded... 🤔 
- If `getDiagnostics()` is not empty, errors are displayed
https://github.com/oxc-project/playground/blob/f50c40cf179f15f8d59c5af273b6475576179bc6/src/components/output/DiagnosticPanel.vue#L9

fisker:
Thanks for the explanation, in this case, shouldn't it be a syntax error?

And `showSemanticErrors: true` seems to produce some weird errors.

For example


```js
(-a) ** b;
```

This code should be valid,  but it throws `SyntaxError: Unexpected exponentiation expression`.

fisker:
Ah, it seems the `preserveParens: false`'s fault.

```
[...]

PR #11065, opened 2025-05-16
fix(napi/playground): Always display syntax errors

leaysgur:
Related #11029

Currently, if the `Check Syntax Errors` option is OFF, it appears that no errors are found during parsing.

image

But actually the parser returns `errors`, this is the gap between playground and parser.

image

---

Does this change make sense?
If so, I will update the playground label from `Check Syntax Errors` to `Check Semantic Errors` later.


Boshen:
Makes sense.

PR #11068, opened 2025-05-16
fix(parser): parse index signature with multiple parameter

### LINKED ISSUE #11052: parser: An index signature must have exactly one parameter.(1096)
The following code typescript reports an error as `An index signature must have exactly one parameter.`

```ts
interface A {
  [A: number, B: number]: number;
}
```

Whereas Oxc reports an error as `Expected ']' but found ','`.

Current `TSIndexSignature`'s parameters is a `Vec`, which means it can store more than one parameter, but it wouldn't parse when it has more than one parameter, so that Vec here is unnecessary; we should change it to `TSIndexSignatureName` instead.

https://github.com/oxc-project/oxc/blob/c1e6c9bfa2ff505dcc87c98495c7fb78185a217d/crates/oxc_ast/src/ast/ts.rs#L1007-L1013

If we need to consider aligning with `TypeScript` (also the same as typescript-eslint), we should make it parse and then throw an error in Semantic instead.

Also, I found that the parameters printing is incorrect in codegen; the separator should be `,` not `|`.
https://github.com/oxc-project/oxc/blob/c1e6c9bfa2ff505dcc87c98495c7fb78185a217d/crates/oxc_codegen/src/gen.rs#L3490-L3496

Personally, I prefer to change the AST to only accept one parameter because multiple parameters means the code is incorrect and make it parse correctly is meaningless, the AST needs to deal with potential multiple parameters, moreover we can't assume only one parameter exists in codegen (ditto for formatted), we must iterate over `parameters` and print all of them, which increases unnecessary checks.

Yes, a lot of downside to parse it correctly.

Boshen:
fixes #11052

### DIFF
--- a/crates/oxc_codegen/examples/codegen.rs
+++ b/crates/oxc_codegen/examples/codegen.rs
@@ -2,8 +2,9 @@
 use std::path::Path;
 
 use oxc_allocator::Allocator;
+use oxc_ast::ast::Program;
 use oxc_codegen::{Codegen, CodegenOptions};
-use oxc_parser::{ParseOptions, Parser, ParserReturn};
+use oxc_parser::{ParseOptions, Parser};
 use oxc_span::SourceType;
 use pico_args::Arguments;
 
@@ -25,8 +26,8 @@ fn main() -> std::io::Result<()> {
     let mut allocator = Allocator::default();
 
     let printed = {
-        let Some(ret) = parse(&allocator, &source_text, source_type) else { return Ok(()) };
-        codegen(&ret, minify)
+        let program = parse(&allocator, &source_text, source_type);
+        codegen(&program, minify)
     };
     println!("First time:");
     println!("{printed}");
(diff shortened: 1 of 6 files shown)


Boshen:
* **#11068** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11068?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11068?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11068).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #11072, opened 2025-05-16
docs(allocator/vec): correct safety comments for `RawVec::append_elements`

overlookmotel:
Follow-on after #10884.

The original safety conditions aren't required any more, because of the change made to `Vec::reserve` in #10884 - `reserve` will panic if `other.len` or `self.len + other.len` exceeds `u32::MAX`. So nothing can go wrong if these conditions aren't satisfied.

Which is lucky, because `append` doesn't do anything to make sure `self.len + other.len <= u32::MAX`!

`append_elements` originally had no safety docs at all (naughty bumpalo!). So add some.

PR #11071, opened 2025-05-16
refactor(allocator/vec): clarify comment

overlookmotel:
Follow-on after #10884. Clarify comment as to why `as u32` is valid here.

PR #11079, opened 2025-05-16
perf(allocator/vec): remove `SetLenOnDrop`

overlookmotel:
Simplify `Vec::truncate` and `Vec::extend_with` methods.

Unlike standard library, our `Vec<'a, T>` requires that `T` is not `Drop`. Therefore we don't need to worry about dropping elements.

There are likely many more places where we can make similar simplifications (optimizations). I've only done these 2 now in order to get rid of `SetLenOnDrop`. I wanted to get rid of it because it requires direct access to `self.buf.len`, which I'd like to make private in a later PR.

### DIFF
--- a/crates/oxc_allocator/src/vec2/mod.rs
+++ b/crates/oxc_allocator/src/vec2/mod.rs
@@ -955,20 +955,11 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     /// [`drain`]: #method.drain
     pub fn truncate(&mut self, len: usize) {
         let current_len = self.buf.len_usize();
-        unsafe {
-            let mut ptr = self.as_mut_ptr().add(current_len);
-            // Set the final length at the end, keeping in mind that
-            // dropping an element might panic. Works around a missed
-            // optimization, as seen in the following issue:
-            // https://github.com/rust-lang/rust/issues/51802
-            let mut local_len = SetLenOnDrop::new(&mut self.buf.len);
-
-            // drop any extra elements
-            for _ in len..current_len {
-                local_len.decrement_len(1);
-                ptr = ptr.offset(-1);
-                ptr::drop_in_place(ptr);
-            }
+        if len < current_len {
+            // SAFETY: `len` is less than current len, so cannot be greater than `u32::MAX`,
+            // and cannot be greater than `self.capacity()`.
+            // We are shrinking `len`, so there are no concerns about initialized elements.
+            unsafe { self.set_len(len) };
         }
     }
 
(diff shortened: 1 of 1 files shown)


overlookmotel:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11079?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#11079** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11079?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11079?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#11074** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11074?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11073** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11073?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11072** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11072?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11071** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11071?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11079).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -2081,66 +2072,34 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
...
             if n > 0 {
                 // We can write the last element directly without cloning needlessly
                 ptr::write(ptr, value.last());
-                local_len.increment_len(1);
             }
 
-            // len set by scope guard
+            // `n` and `self.len() + n` must both be `<= u32::MAX`, otherwise `self.reserve(n)` above
+            // would have panicked. So `n as u32` cannot truncate `n`, and `len + n` cannot wrap.
+            #[expect(clippy::cast_possible_truncation)]
```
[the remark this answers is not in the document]
overlookmotel: Oh do shut up about `#[expect]`, you doofus!

crates/oxc_allocator/src/vec2/mod.rs
```diff
@@ -2081,66 +2072,34 @@ impl<'bump, T: 'bump> Vec<'bump, T> {
     fn extend_with<E: ExtendWith<T>>(&mut self, n: usize, mut value: E) {
         self.reserve(n);
 
+        // std library version of this method updates `len` after writing each element,
+        // because `value.next()` or `value.last()` could panic.
+        // If that happens, `len` needs to contain all the elements written so far,
+        // so they get dropped when the `Vec` is dropped.
+        // But our `Vec` requires that `T` is not `Drop`, so we don't need to worry about that.
```
Dunqing, line 2079: What does this mean? We don't need to worry about the `len` being incorrect, or is it due we don't have a way to deal with the `SetLenOnDrop::drop` thing, so we don't need to worry about it?

Dunqing (approved):
Looks good! Your comments let me know better what `SetLenOnDrop` is going on.

crates/oxc_allocator/src/vec2/mod.rs
overlookmotel: Oh damn it! The way I wrote this comment is obviously not very clear.

### What stdlib does

stdlib's `Vec` allows the type it contains `T` to be `Drop`. When the `Vec` is dropped, it must drop the individual `T`s it contains first. This is performed in the `Drop` impl for `Vec`.

`Vec::resize` takes a value and clones it multiple times, inserting each clone into the `Vec`.

The problem is that `Clone::clone` can panic (because it can be user-defined).

Let's consider what happens if `clone` panics on the 2nd call if you don't set `len` after each turn of the loop:

* Our `Vec` is a `Vec<String>`.
* Initial length of `Vec` is 0.

1. `String` is cloned and added to the `Vec`.
2. `len` of the `Vec` is still 0 (`len` is only incremented in a local var inside `extend_with`).
3. `extend_with` tries to clone a 2nd value, but `clone` panics.
4. Stack unwinds, and all objects on stack are dropped.
5. The `Vec` is dropped.
6. `Vec::drop` must drop all the `String`s it owns.
7. It checks it's `len` field, but it's 0, so it concludes "I don't own anything, nothing to do".
8. The first `String` which was inserted into the `Vec` in step 2 is leaked -> memory leak.

So the solution std lib takes is to always set the `len` field after each element is added to the `Vec`. `len` is always up to date. This ensures that if there's a panic, `Vec::drop` knows exactly how many elements it owns, and will drop them all.

It does this via the `SetLenOnDrop` structure, which itself has a `Drop` impl which sets the `len` field when it's dropped. It's pretty complicated!

### Why we can do it simpler

In our case, we require that `T` in `Vec<T>` is not `Drop`, and `Vec` itself has no `Drop` implementation. So nothing ever needs to be dropped. That's what I mean about "we don't need to worry about it".

Consequently all the complication with `SetLenOnDrop` is unnecessary, and we can just set the `len` field once at the end. If `clone` panics, then `len` field will be left in an inaccurate state, but that doesn't matter as nothing reads it anyway.

overlookmotel:
### Merge activity

* **May 17, 6:15 AM EDT**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

overlookmotel:
I'll fix the Miri fail. It appears not to be due to Miri flagging UB, but due to lint errors which have been added on nightly.

overlookmotel:
Mire fail was unrelated. Fixed in #11092.

crates/oxc_allocator/src/vec2/mod.rs
Dunqing: Thank you for taking the time to answer my question!

> Oh damn it! The way I wrote this comment is obviously not very clear.

Not at all, the comment is good! I don't understand due to my lack of Rust low-level knowledge.

Issue #10033, opened 2025-03-25
minifier: remove unused variable declaration


sapphi-red:
```js
const FOO = 0
// replace with

const BAR = something()
// replace with
something()
```

To keep the minifier idempotent, this feature requires updating the symbol table when dropping references in compression.

PR #11091, opened 2025-05-17
wip(minifier): update semantics

sapphi-red:
I had this commit locally for a while but forgot to push it.

This PR makes the minifier to update the semantics information when removing a reference. This is needed for #10033.

@Boshen What do you think of this approach? If it looks good, I'll rebase this on main branch and add a test similar to `tasks/transform_checker`.


Boshen:
I haven't thought about updating semantics in the minifier yet, will do when the time is ready. I'll leave this as draft for now.

PR #11104, opened 2025-05-17
docs(linter): normalize rule docs format

ulrichstark:
Most of these changes were done automatically using find & replace. This brings many rule docs closer to the desired format of #6050. Also fixed minor issues in some docs discovered while working on this.


camc314 (approved):
thank you for the PR, please try to split these up into smaller chunks in future as it means it's easier and quicker to review

ulrichstark:
> thank you for the PR, please try to split these up into smaller chunks in future as it means it's easier and quicker to review

@camc314 yes true, sorry. The PR got bigger than anticipated while working on it. I will make more selective changes in the future. Thanks for your follow up changes!

PR #11142, opened 2025-05-19
refactor(regular_expression): Refactor `regexp-modifiers` support

leaysgur:
When I first implemented this proposal, it was Stage 3 and the specs were not even fixed.
But for now, this spec is in Stage 4 and has been cleared up, so I've updated it.

In the process, I have done some refactoring.

- Shorten code by `bitflags` like `RegExpFlags`
  - This requires some lines to make `ast_tools` works
- Update readme

NOTE: The basic behavior remains unchanged.


leaysgur:
I checked the `cal.com.tsx` file and all the `RegExp` usage. But none of them pass `parse_modifiers()`.

Regression is accidental...? or Am I missing something? 🤔

Boshen:
Nice!

PR #11184, opened 2025-05-20
feat(linter): add autofix to eslint/no-unneeded-ternary

huangtiandi1999:
### Auto-fix Example
```js
let a = x === 2 ? true : false;
// let a = x === 2;
```

```js
foo() ? false : true;
// !foo();
```

```typescript
foo ? foo : bar as any
// foo || (bar as any)
```

```typescript
let a = {} satisfies User ? true : false
// let a = !!({} satisfies User)
```


camc314 (approved):
thanks!

Issue #11145, opened 2024-06-21
Custom arena allocator


overlookmotel:
I feel we should investigate replacing Bumpalo with a custom allocator. Notes on why, and some ideas:

### Make it faster - remove alignment calculations

Oxc's AST has a property which is unusual. Every single AST type is aligned on 8 (because they all contain a `Vec`, `Box`, `Atom` or `&str`).

We can exploit this property to remove alignment calculations when allocating into the arena. The bump pointer is always aligned on 8 after the last allocation, so no need to re-align each time you allocate.

This will only remove a couple of instructions (and possibly one branch), but given that allocating is a very hot path, it might well make a measurable difference.

Bumpalo is designed for a general use case, and so cannot perform this optimization.

NB: There are 2 exceptions to the "everything aligned on 8" rule:

1. String data is aligned on 1. Either store strings in a separate memory space, or at start of arena chunks (see below), or align them on 8 manually.
2. `BooleanLiteral` is aligned on 4. Easy fix by making it `#[repr(align(8)]`.

Reference: [ser_raw](https://docs.rs/ser_raw/latest/ser_raw/storage/struct.AlignedVec.html) performs this optimization.

### Make it faster - alternative design

[Blink allocator](https://github.com/zakarumych/blink-alloc) [claims to be much faster](https://github.com/zakarumych/blink-alloc/blob/main/BENCHMARKS.md) than Bumpalo. There was an attempt to replace Bumpalo with Blink in Oxc last year https://github.com/oxc-project/oxc/pull/106, but it didn't show performance improvements. Nonetheless, I think it's worthwhile investigating why and whether anything we can learn from Blink.

I have some questions myself about Bumpalo's design. I don't understand why it uses double-indirection (pointer to a pointer) to get to the current chunk cursor. I wonder if reducing indirection could yield a speed boost.

Some other possible optimizations here: https://github.com/fitzgen/bumpalo/issues/234

### Fill chunks from both ends

* Fill from end for AST nodes (bumping downwards, same as Bumpalo does).
* Fill from start for strings (bumping upwards).

oxc-project/backlog#37

This would also have the nice property of grouping all string data together in one place in memory, which is probably advantageous for AST transfer.

### Align chunks on 4 GiB

[...]

PR #11198, opened 2025-05-20
refactor(allocator): introduce `Alloc` trait

overlookmotel:
First step towards replacing `bumpalo` with our own allocator (#11145).

Introduce `Alloc` trait and implement it for `bumpalo::Bump`. A later PR will parameterize `Vec` with `Alloc`, instead of a static dependency on `Bump`.

We can later on replace `Bump` with our own `Arena` type. `Arena` will also implement `Alloc`, so the two will be able to be swapped with a minimum of fuss.

### Why do we need a trait at all?

Why couldn't `Vec` just statically depend on `Arena`, same as it now does on `Bump`?

Rationale for a trait here is: My intent is to have 2 different variants of `Arena` to support filling allocator chunks from both ends (string data fills from the start, all other types fill from the end). This will have 3 advantages: 

1. More tightly packed data. `&str` / `Atom` is the only type in the AST which is not aligned on 8. Strings currently often tend to end up with unnecessary padding bytes around them.
2. Packing all string data into a single immutable stretch of memory has potential to speed up all string operations (https://github.com/oxc-project/backlog/issues/46#issuecomment-2603153807).
3. It should unblock the biggest current perf bottleneck in raw transfer, by allowing all strings in the AST to be converted to UTF-16 in a single shot (as if they were just 1 long string), rather than converting each string individually.

`String` is a wrapper around `Vec`, and it's preferable not to have to implement `Vec` twice for front-filling (strings) and back-filling (all other types). `Vec<T, A: Alloc>` is a simpler solution.

### DIFF
--- a/crates/oxc_allocator/src/alloc.rs
+++ b/crates/oxc_allocator/src/alloc.rs
@@ -1,0 +1,183 @@
+// All methods just delegate to `Bump`'s methods
+#![expect(clippy::inline_always)]
+
+use std::{
+    alloc::{Layout, handle_alloc_error},
+    ptr::NonNull,
+};
+
+use allocator_api2::alloc::Allocator;
+use bumpalo::Bump;
+
+/// Trait describing an allocator.
+///
+/// It's a simpler version of `allocator_api2`'s [`Allocator`] trait.
+///
+/// The difference between these methods and [`Allocator`]'s versions of them are:
+///
+/// * `shrink` and `grow` return a pointer and panic/abort if allocation fails,
+///   instead of returning `Result::Err`.
+/// * All methods return a `NonNull<u8>`, instead of `NonNull<[u8]>`.
+pub trait Alloc {
+    /// Allocate space for an object with the given [`Layout`].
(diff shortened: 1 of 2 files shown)


overlookmotel:
* **#11198** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11198).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing (approved):

Dunqing:
Wow, that's a good idea to improve `String`

PR #11202, opened 2025-05-21
chore(ci): test big endian architecture (s390x-unknown-linux-gnu)

### LINKED ISSUE #11155: Test in CI on big-endian arch
#11119 demonstrated that we have code which relies on the endianness of the system, and that it's not tested in our CI.

There's more code that depends on endianness in our crates, and also in dependencies like [compact_str](https://crates.io/crates/compact_str) (which doesn't have any testing for big endian in its own CI).

Raw transfer currently does not support big endian, only because I'm not sure how to test it.

It would be great if we could add a big endian arch to CI. It'd be ideal if we ran both:

1. `cargo test`
2. `cd napi/parser; pnpm build && pnpm test;`

These probably wouldn't need to run on all PRs, maybe only on main. Tests are quite quick to run, so I doubt would be a big slow-down to CI, unless cross-compilation imposes a large cost.

Boshen:
closes #11155

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,19 @@ jobs:
         working-directory: ${{ env.DEV_DRIVE_WORKSPACE }}
         shell: bash
 
+  test-big-endian:
+    name: Test big-endian # s390x-unknown-linux-gnu is a big-endian
+    if: ${{ github.ref_name == 'main' }}
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Boshen:
### Merge activity

* **May 21, 3:12 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

Issue #11194, opened 2025-05-20
parser: use checkpoints instead of token lookahead


camchenry:
In order to remove `Lexer::lookahead` and the `VecDeque` there, we need to remove any usage of it in the parser. Namely, that comes through these methods:

- `ParserImpl::peek_token`
- `ParserImpl::peek_kind`
- `ParserImpl::peek_at`
- `ParserImpl::nth`
- `ParserImpl::nth_at`
- `ParserImpl::nth_kind`

So, in order to remove the `lookahead` field from the lexer, we'll need to replace usage of these methods with `checkpoint`/`rewind` instead.

Boshen:
An example of replacing lexer token peek with parser peek: 

https://github.com/oxc-project/oxc/blob/699ec645f18c1b2fffac0d66c9bfaf148317bbc6/crates/oxc_parser/src/js/statement.rs#L131-L138 -> https://github.com/microsoft/TypeScript/blob/81c951894e93bdc37c6916f18adcd80de76679bc/src/compiler/parser.ts#L7337-L7342

Boshen:
After removing `Lexer::lookahead`, it is worth a try to find where `Parser::lookahead` and `Parser::try_parse` are called unnecessarily and reduce these calls.

leaysgur:
We already have `ParserImpl::lookahead()` and it does not depend on `Lexer::lookahead`.

So this issue can be solved by following 3 steps below.

- Rewrite all `ParserImpl::peek_xxx()` and `ParserImpl::nth_xxx()` usage
  - Find files that use these deprecated methods (see list below or run `rg '\.(peek_(token|at|kind)|nth(_(at|kind))?)'`)
  - Check the `microsoft/TypeScript` repo and find the relevant logic in `parser.ts`
  - Update our parser logic with `ParserImpl::lookahead()` or something that does NOT depend on `Lexer::lookahead`
- Add "special check before `ParserImpl::lookahead` to reduce the need for a lookahead"
  - If needed
- Remove `Lexer::lookahead`
  - Revive #11206 
  - (This should be merged last, parser perf is regressed without updating its usage)

### Files to update

- [ ] js/arrow.rs @camchenry 
- [ ] js/binding.rs
- [ ] js/class.rs
- [ ] js/declaration.rs
- [ ] js/expression.rs
- [ ] js/function.rs
- [ ] js/module.rs
- [ ] js/object.rs
- [ ] js/statement.rs
- [ ] jsx/mod.rs
[...]

PR #11206, opened 2025-05-21
perf(parser): Remove `Lexer::lookahead: VecDeque`

leaysgur:
Part of #11194

I found that `ParserImpl::lookahead` was already implemented and it did not seem to depend on `Lexer::lookahead`.
So, I replaced all the `ParserImpl::peek_xxx()` methods and removed `Lexer::lookahead` as a first step.

### DIFF
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -64,7 +64,7 @@ impl<'a> ParserImpl<'a> {
     /// Peek next token, returns EOF for final peek
     #[inline]
     pub(crate) fn peek_token(&mut self) -> Token {
-        self.lexer.lookahead(1)
+        self.nth(1)
     }
 
     /// Peek next kind, returns EOF for final peek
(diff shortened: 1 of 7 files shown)


Boshen:
This is a good start, but you need to change all the parser methods that use token peek to how typescript parses things, otherwise it's not going to work.

leaysgur:
> change all the parser methods that use token peek to how typescript parses things

Yes, I was thinking of doing that as a next step.

And please let me confirm about current benchmark results?

- Lexer improvement is expected ✨
- Parser regression is also expected for now

But it will be improved in the future by following TS logic like:

- Support
  - > typescript parser always have a special check before Parser::lookahead to reduce the need for a lookahead
- Reduce redundunt, nested `peek_xxx` calls
- etc...

Am I missing something?

Boshen:
You are correct, please continue 😁 @camchenry seems to be working on arrow.rs so try other files first.

leaysgur:
Closing this for now, since we know what needs to be done. (And this cannot be merged soon.)

Replacing methods that depend on `Lexer::lookahead' can be done without this PR.
And when all the work is done, we do the same thing again, without tweaking `src/cursor.rs`.

PR #11207, opened 2025-05-21
perf(parser): remove redundant checks

ulrichstark:
1. `is_number` in `class.rs` can be removed, because `is_identifier_or_keyword` implies `is_literal_property_name` implying `is_number`
2. ~~Avoids checking for `is_reserved_keyword` one way and then the other way.~~
3. `is_all_keyword` in `is_identifier_or_keyword` can be removed, because `is_literal_property_name` implies `is_identifier_name` implying `is_all_keyword`

Let's see if those changes yield any improvements in the benchmark or if the compiler was smart enough to optimize that.

Edit: removed the second change because it's not worth the sacrificed readability

### DIFF
--- a/crates/oxc_parser/src/js/class.rs
+++ b/crates/oxc_parser/src/js/class.rs
@@ -239,11 +239,7 @@ impl<'a> ParserImpl<'a> {
         }
 
         let kind = self.cur_kind();
-        if kind.is_identifier_or_keyword()
-            || kind.is_number()
-            || kind == Kind::Star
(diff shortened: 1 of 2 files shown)


ulrichstark:
Doesn't look like there are any substantial improvements. Feel free to close my PR or pick the changes you want. Two of my three changes are simple removals and make the code simpler. `is_identifier` was honestly more readable before I changed it, but I wanted to try it anyway for potentially better performance.

PR #11208, opened 2025-05-21
fix(lexer): restore debug assert in `Lexer::finish_next`

overlookmotel:
#11204 removed a `debug_assert!` from `Lexer::finish_next` which checks that `end` of a `Token` is always >= `start`.

I assume this was unintentional, as this check is useful for making sure lexer is operating correctly in conformance tests. Restore it.


overlookmotel:
@Boshen I *assume* removing it was unintentional?

Boshen:
It's checked here https://github.com/oxc-project/oxc/blob/6e9de84808694c31dd584ab719f28ad8c06f8a95/crates/oxc_parser/src/lexer/token.rs#L99

overlookmotel:
Oh sorry I completely missed that.

PR #11232, opened 2025-05-22
refactor(parser): Remove Lexer peeking for jsx

leaysgur:
Part of #11194 

Diffs in the snapshots are expected and closer to the TS.

### DIFF
--- a/crates/oxc_parser/src/jsx/mod.rs
+++ b/crates/oxc_parser/src/jsx/mod.rs
@@ -8,7 +8,10 @@ use crate::{Context, ParserImpl, diagnostics, lexer::Kind};
 
 impl<'a> ParserImpl<'a> {
     pub(crate) fn parse_jsx_expression(&mut self) -> Expression<'a> {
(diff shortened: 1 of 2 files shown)


Boshen (commented):
tsc parser `parseJsxChild` function seems a lot more optimized than our code, feel free to follow up with more PRs to have less rewinds.

leaysgur:
That might be the case.

But from the beginning, TSC did not use a `parseJsxExpression()` as an entry point for parsing JSX, so we may need to rewrite this from the ground up. 🙃

---

Oh My GitHub... 😇

Boshen:
### Merge activity

* **May 22, 10:07 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

camchenry:
I re-ran the failed jobs, hopefully that should fix it and allow it to merge

PR #11235, opened 2025-05-22
perf(linter): avoid iterating lines twice if blank lines are skipped

ulrichstark:
Both rules `eslint/max-lines` and `eslint/max-lines-per-function` previously did count all lines and then count all blank lines again to subtract them if config parameter skipBlankLines was enabled. Just count all non-blank lines if skipBlankLines is enabled to avoid one redundant iteration over all lines.


ulrichstark:
Sadly I don't see any improvements in the benchmarks because both rules run with default config parameters => disabled skipBlankLines.

PR #11252, opened 2025-05-23
feat(ast): add parenthesized info to `AssignmentExpression` and `UpdateExpression`

### LINKED ISSUE #10899: No error for invalid left-hand side in assignment when using `preserveParens: false`
Code:
```js
({}) = x;
```

With `preserveParens: true`: "Cannot assign to this expression" [Playground](https://playground.oxc.rs/#eNo9T7tqxEAM/JVDVQIukkAah3QhdSDtNfLe+FhYS0Za3wPjf7/dM3alGZiXZgrU0su8vB6+D7evo1BDSu1MNkk9fpfMN2qzTWgoRckb9qAjdnIfOk0by8bivdpAbc/JsTQ0sjmsJo6Ggi74Y4P4aimCGv0UFLz7V0dmO6P0Evzj7f2zLIyuiTNOPwiJjXPUmrR1BT3hjOd8CHcJ/zpZwMDj3jZEiX3c+oJKNk2/Sa/VdIF16uW3NXBZHr3gZ2U=)

With `preserveParens: false`: AssignmentExpression.left.type == "ObjectPattern" [Playground](/Ovj87tUjK6JMw5/CImNc9RXVNADjrj3h3CTsNPBAjrul4bF0kWJbVzzgko2Tf9Jz/PRCdaol+cW4DTdABqDZ7A=)

Note: TS-ESLint produce an AST with AssignmentExpression.left.type == "ObjectExpression", but I do think erroring is better and more consistent

Boshen:
fixes #10899
fixes #10939

To allow

* `(a)++`
* `({}) = {}`
* `(fn) = function () {};` See https://oxc.rs/docs/learn/ecmascript/grammar.html#parenthesized-expression

in codegen and formatter.

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -686,6 +686,8 @@ pub struct UpdateExpression<'a> {
     pub operator: UpdateOperator,
     pub prefix: bool,
     pub argument: SimpleAssignmentTarget<'a>,
+    #[estree(skip)]
+    pub argument_parenthesized: Option<Span>,
 }
 
 /// `typeof` in `typeof a === "string"`
@@ -768,6 +770,8 @@ pub struct AssignmentExpression<'a> {
     pub operator: AssignmentOperator,
     pub left: AssignmentTarget<'a>,
     pub right: Expression<'a>,
+    #[estree(skip)]
+    pub left_parenthesized: Option<Span>,
 }
 
 inherit_variants! {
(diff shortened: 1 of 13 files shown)


overlookmotel:
I'm not sure what you mean by "play nicely with". If you mean don't include these fields in `AstBuilder` methods, mark them `#[builder(default)]`.

But... this change increases the size of `AssignmentExpression` and `UpdateExpression` by 16 bytes, and both are very common in AST.

Couldn't we instead add a variant to `SimpleAssignmentTarget`?

```rs
enum SimpleAssignmentTarget<'a> {
    ParenthesisedAssignmentTarget(Box<'a, ParenthesisedAssignmentTarget<'a>>),
    // ...
}

struct ParenthesisedAssignmentTarget<'a> {
    span: Span,
    target: SimpleAssignmentTarget<'a>,
}
```

This would cost nothing where left side is not parenthesised (99% of cases).

PR #11294, opened 2025-05-25
fix(linter): make `jest/no-large-snapshots` error easier to comprehend

ulrichstark:
> @camc314 as a followup, we could change the span to perhaps just point to toThrowErrorMatchingInlineSnapshot rather than the whole snapshot to help the error be more comprehensible

https://github.com/oxc-project/oxc/pull/11291#discussion_r2106288733


camc314 (commented):
thanks 🙏

PR #11298, opened 2025-05-26
refactor(parser): Remove Lexer peeking for js/expression

leaysgur:
Part of #11194 

- Snapshot diffs closer to TS
- `next_token_is_identifier_or_keyword_or_literal_on_same_line` is based on:
  - https://github.com/microsoft/TypeScript/blob/main/src/compiler/parser.ts#L7150

### DIFF
--- a/crates/oxc_parser/src/js/expression.rs
+++ b/crates/oxc_parser/src/js/expression.rs
@@ -734,12 +734,15 @@ impl<'a> ParserImpl<'a> {
             }
 
             let mut question_dot = false;
-            let is_property_access = if allow_optional_chain && self.at(Kind::QuestionDot) && {
-                let peek_kind = self.peek_kind();
-                peek_kind == Kind::LBrack
-                    || peek_kind.is_identifier_or_keyword()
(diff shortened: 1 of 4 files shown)


crates/oxc_parser/src/js/expression.rs
```diff
@@ -1427,40 +1433,44 @@ impl<'a> ParserImpl<'a> {
...
+    ) -> bool {
+        self.bump_any();
+
+        // NOTE: This check implementation is based on TypeScript parser.
+        // But TS does not have this exception.
+        // This is needed to pass parser_babel test to parse:
+        // `for (await of [])` with `sourceType: script`
+        if !self.is_ts && is_await && self.at(Kind::Of) {
+            return false;
+        }
```
leaysgur, line 1470:
```rs
if matches!(
    peek_token.kind(),
    Kind::Of | Kind::LParen | Kind::LBrack | Kind::Slash | Kind::RegExp
) {
```

The original code covered these.
But even with original code, the test did not fail when I erased all but the `Of`.

PR #11308, opened 2025-05-26
refactor(parser): Remove lexer lookahead in JS let declaration parsing

therewillbecode:
Part of #11194.

Removes usage of lexer lookahead methods in favour of using the parser lookaheads for parsing `let` declarations.


camchenry (approved):

Boshen (commented):
This is on a hot path, but none of our benchmarks exercise this hot path fully.

Tracking `SyntaxKind.LetKeyword` I found 

```
            case SyntaxKind.LetKeyword:
                if (isLetDeclaration()) {
                    return parseVariableStatement(getNodePos(), hasPrecedingJSDocComment(), /*modifiers*/ undefined);
                }
                break;
```

So maybe a little refactoring to make the code read better in a follow up PR.

PR #11310, opened 2025-05-26
feat(linter): use `itoa` for integer-to-str in eslint rules

shulaoda:


Boshen:
@Sysix please note codspeed benchmark doesn't exercise these code paths, don't make the judgement too early!

shulaoda:
> @shulaoda  please note codspeed benchmark doesn't exercise these code paths, don't make the judgement too early!
> 
> 
> 
> But please start from optimizing hot paths, these cold paths don't make too much of a difference.

Actually, I was referring to the benchmark of itoa itself. After submitting this PR, I suddenly wanted to verify whether I had done the right thing. So I added a benchmark for formatting using itoa on top of the existing itoa benchmark. I found only a very slight improvement, which made me think that the changes might not be very meaningful.

PR #11312, opened 2025-05-27
chore(ci): use warm cache in wasi build

Boshen:


Boshen:
Most of the cache are shared expect `napi`.

PR #11315, opened 2025-05-27
fix(eslint/no-fallthrough): fix allowEmptyCase invalid

ityuany:
## Problem Description

The `allowEmptyCase` configuration was not working properly. When set to `true`, empty case statements would still trigger fallthrough errors.

### 1. Refactored Logic
- Fixed the `is_fallthrough` judgment logic to comply with ESLint official specification
- Properly handles `allowEmptyCase` configuration in all scenarios

### 2. Configuration Behavior
- **`allowEmptyCase: true`**: Empty cases are always allowed to fall through regardless of layout
- **`allowEmptyCase: false` (default)**: Only allows empty cases when they are on the same line or consecutive lines

### 3. Added Test Cases
Added comprehensive test cases to verify `allowEmptyCase` configuration:
- Empty case behavior under different layouts
- Ensures non-empty case behavior is unaffected

## Testing

✅ All existing tests pass
✅ New test cases verify the fix
✅ Manual testing confirms configuration works

## Reference

ESLint documentation: https://eslint.org/docs/latest/rules/no-fallthrough#allowemptycase

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -332,9 +332,20 @@ impl Rule for NoFallthrough {
                 continue;
             }
             let is_illegal_fallthrough = {
-                let is_fallthrough = !case.consequent.is_empty()
-                    || (!self.0.allow_empty_case
-                        && Self::has_blanks_between(ctx, case.span.start..next_case.span.start));
+                let is_case_empty = case.consequent.is_empty();
+                let is_fallthrough = if is_case_empty {
+                    // For empty cases, check allowEmptyCase option
+                    if self.0.allow_empty_case {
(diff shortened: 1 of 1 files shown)


camc314 (commented):
thanks for the PR, but i'm not sure what this is fixing, your changed logic appears to have the same logic before, it also doesn't fix #6417 and the additional tests you added appear to pass (as expected) on `main` ?

ityuany:
> thanks for the PR, but i'm not sure what this is fixing, your changed logic appears to have the same logic before, it also doesn't fix #6417 and the additional tests you added appear to pass (as expected) on `main` ?感谢您的PR，但我不确定这是什么修复，您更改的逻辑似乎具有相同的逻辑，它也没有修复＃6417，并且您添加的其他测试似乎在MAIN上（预期）通过（预期）？

see  https://github.com/oxc-project/oxc/issues/6417#issuecomment-2586087954

PR #11323, opened 2025-05-27
perf(parser): simplify getting span of identifiers and literals

overlookmotel:
Small optimization. For identifiers and literals, the span of the AST node is the same as the span of the `Token`. So just get the whole span from the `Token` in one access, and avoid the dance of assembling the span by calling `start_span` and then `end_span`.


camchenry (approved):

Boshen:
### Merge activity

* **May 27, 3:18 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.

PR #11324, opened 2025-05-27
test(linter): `eslint/no-duplicate-imports` shouldn't report the same span

ulrichstark:
Adds a test for a bug that was unintentionally fixed by #11320.
I also confirmed that the test would have failed before my original PR.

Context: https://github.com/oxc-project/oxc/pull/11320#issuecomment-2912286528


camc314 (approved):
thank you

PR #11332, opened 2025-05-27
perf(parser): remove lexer lookahead

### LINKED ISSUE #11194: parser: use checkpoints instead of token lookahead

camchenry:
- Closes #11194 

Removes the lookahead `VecDeque` in the lexer and all of the bookkeeping code depending on it.

### DIFF
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -61,49 +61,6 @@ impl<'a> ParserImpl<'a> {
         self.lexer.get_template_string(self.token.start())
     }
 
-    /// Peek next token, returns EOF for final peek
-    #[inline]
-    pub(crate) fn peek_token(&mut self) -> Token {
-        self.lexer.lookahead(1)
-    }
-
-    /// Peek next kind, returns EOF for final peek
-    #[inline]
-    #[expect(dead_code)]
-    pub(crate) fn peek_kind(&mut self) -> Kind {
-        self.peek_token().kind()
-    }
-
-    /// Peek at kind
-    #[inline]
-    #[expect(dead_code)]
-    pub(crate) fn peek_at(&mut self, kind: Kind) -> bool {
-        self.peek_token().kind() == kind
-    }
-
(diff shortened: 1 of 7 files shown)


camchenry:
Similar to before, this speeds up the lexer (~3-4%) but slows down the parser (~3-4%) which nets out to being slower than before. However, once we work on https://github.com/oxc-project/oxc/issues/11334 and other small perf improvements, we should be able to recover this performance and do even better in the long run.

crates/oxc_parser/src/cursor.rs
```diff
@@ -384,7 +342,13 @@ impl<'a> ParserImpl<'a> {
...
                 first = false;
             } else {
-                if !trailing_separator && self.at(separator) && self.peek_at(close) {
+                if !trailing_separator
+                    && self.at(separator)
+                    && self.lookahead(|p| {
+                        p.bump_any();
+                        p.at(close)
+                    })
+                {
```
overlookmotel, line 351: I suggest splitting this change out into a separate PR. Presumably these lines is where the parser perf regression is coming from. Removing the `VecDeque` from the lexer should be a gain in both lexer and parser benchmarks.

Then in this block of code, I think we can refactor to avoid using lookahead at all. I think that should be possible since we're immediately bumping on to next token anyway in the next line `self.expect(separator)`. If we can do that, hopefully it'll remove the perf regression.

But it'd be useful to benchmark that in isolation without the change of removing the `VecDeque` also in the mix - hence why I think it'd be better in a separate PR.

PR #11349, opened 2025-05-28
perf(parser): remove lexer lookahead

Boshen:


camchenry:
Third times the charm maybe? 😄

Boshen:
Performance regression after we remove code? 👎

PR #11356, opened 2025-05-28
refactor(parser): rewrite import/export specifier parsing

camchenry:
- part of https://github.com/oxc-project/oxc/issues/11334

Rewrites the import/export specifier parsing to be much closer to the TS compiler reference version. The diagnostics should be nearly the same as before, but with a slight improvement for a few cases where we had "unexpected token" before and now we know it should be "identified expected." In theory this might restore the performance somewhat since we don't do any lookahead in this version. However, the initial version of this has some extra allocations there weren't there previously I think, so it may not be much faster.


camchenry:
Looks like this gets us back around 1% of the parser performance, so probably a worthwhile improvement.

Boshen:
Great work!

PR #11358, opened 2025-05-28
perf(parser): speed up simple lookaheads by introducing `Lexer::peek_token`

ulrichstark:
This is my proof of concept function to implement @overlookmotel's tip:

> We could also introduce a more slimmed-down (cheaper) version of lookahead when all you need is to get the next 1 or 2 tokens from lexer and then reset its pointer and `errors_pos` back to where they were before. Current `lookahead` does quite a lot of work which isn't required in these cases.

source: https://github.com/oxc-project/oxc/issues/11334#issuecomment-2914341490

### DIFF
--- a/crates/oxc_parser/src/js/class.rs
+++ b/crates/oxc_parser/src/js/class.rs
@@ -222,7 +222,7 @@ impl<'a> ParserImpl<'a> {
         );
 
         // static { block }
-        if self.at(Kind::Static) && self.lookahead(Self::next_token_is_open_brace) {
+        if self.at(Kind::Static) && self.lexer.peek_token().kind() == Kind::LCurly {
             for decorator in decorators {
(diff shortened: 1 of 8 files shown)


ulrichstark:
Looking great 🎉

image

Boshen:
This looks very promising! Let's wait until we fix all the code around parser peek before going into this.

ulrichstark:
Just rebased onto latest master to include #11356 in my PR.

ulrichstark:
- Rebased again
- Changed api from Lexer::lookahead_token taking predicate to Lexer::peek_token returning peeked token
- Found three more cases to use Lexer::peek_token

Boshen:
I'll take a look at the remaining token peeks before this PR.

Boshen:
Can't resist the 2%!

PR #11363, opened 2025-05-28
refactor(parser): Add some is_ts fast checks before doing lookaheads

therewillbecode:
Add some is_ts fast checks before doing lookaheads.

Relates to #11334 .

### DIFF
--- a/crates/oxc_parser/src/ts/statement.rs
+++ b/crates/oxc_parser/src/ts/statement.rs
@@ -477,11 +477,18 @@ impl<'a> ParserImpl<'a> {
     }
 
     pub(crate) fn at_start_of_ts_declaration(&mut self) -> bool {
+        if !self.is_ts {
+            return false;
+        }
         self.lookahead(Self::at_start_of_ts_declaration_worker)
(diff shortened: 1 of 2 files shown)


crates/oxc_parser/src/ts/statement.rs
```diff
@@ -477,11 +477,18 @@ impl<'a> ParserImpl<'a> {
...
     pub(crate) fn at_start_of_ts_declaration(&mut self) -> bool {
+        if !self.is_ts {
+            return false;
+        }
         self.lookahead(Self::at_start_of_ts_declaration_worker)
     }
 
     /// Check if the parser is at a start of a ts declaration
     fn at_start_of_ts_declaration_worker(&mut self) -> bool {
+        if !self.is_ts {
```
leaysgur, line 488: And this is exactly what is called by `at_start_of_ts_declaration` above.

leaysgur (changes requested):
I don't think all these checks are necessary, many of them are already guarded by the caller.

Would you please check again?

PR #11377, opened 2025-05-29
perf(parser): avoid checkpoint when parsing left curly in jsx

ulrichstark:
This requires `parse_jsx_expression_container` and `parse_jsx_spread_child` not to expect LCurly because that's already bumped by the caller. Is that okay? Do I need to update the doc comments of both functions? Let's see if there are improvements in the benchmark...

Part of #11334

### DIFF
--- a/crates/oxc_parser/src/jsx/mod.rs
+++ b/crates/oxc_parser/src/jsx/mod.rs
@@ -264,18 +264,16 @@ impl<'a> ParserImpl<'a> {
                 self.unexpected()
             }
             Kind::LCurly => {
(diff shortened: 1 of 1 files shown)


leaysgur:
FYI: I also tried to eliminate the `rewind()` for `{expr` and `{...expr` in #11314 (https://github.com/oxc-project/oxc/commit/a12ff218e6d93be84a90ba1f8bcff25181b4b384), but there was no visible improvement. 😓

Boshen (commented):
This is a really good change. I added a small change to avoid to `at` + `bump` check.

PR #11378, opened 2025-05-29
perf(parser): use const generics to speed up parsing

magic-akari:
Since `is_ts` never changes during parsing, we can convert it into a const generic parameter to test for potential speed improvements.


Boshen:
image

Let me check later.

PR #11362, opened 2025-05-28
refactor(ast/estree): re-order fields in visitation order

overlookmotel:
Re-order fields in ESTree and TS-ESTree ASTs to be in visitation order (which is the order in which they're defined in Rust types).

This does not align with either Acorn or TS-ESLint, but for the reasons discussed in https://github.com/oxc-project/oxc/issues/9705#issuecomment-2859250834, I feel this isn't a concern. The field order doesn't matter much per se, only that it's consistent, and this change has other benefits - it will make it simpler to build a visitor which visits fields in the correct order, and allows adding other parsers to `acorn-test262` in order to produce test cases for decorators.

Visitation order aligns with TS-ESLint except for the following types:

* `ExportSpecifier`
* `TSImportType`
* `TSIndexedAccessType`
* `TSMethodSignature`
* `TSPropertySignature`
* `TSTypePredicate`

In these 6 cases, TS-ESLint's `visitorKeys` has the wrong visitation order, because it does not visit fields in source order. I will open an issue and hopefully they can fix.

This PR also bumps `acorn-test262` to include https://github.com/oxc-project/acorn-test262/pull/37 which re-orders the fields in fixture snapshots to match the new field order here.

PR #11385, opened 2025-05-30
fix(estree): Use consistent field order in serialization

leaysgur:
Part of #9705 , follow up #11362 

This change will generate the same AST shapes in all locations, regardless of `visitor-keys`.


overlookmotel (approved):
Great! Thank you.

PR #11387, opened 2025-05-30
perf(linter): change `Arc<OsStr>` to `Box<OsStr>` in linter runtime

Boshen:
Given our conditions:

* Paths are relatively small strings
* There's about 1k - 10k paths in total for most apps
* mimalloc is used

`Arc::clone` + `Arc::drop` should be slower than `String::clone` + `String::drop`.


Boshen:
Nope, too much clone happened with this change, leading to performance regression.

PR #11394, opened 2025-05-30
docs(linter): clarify jsdoc/check-tag-names configuration

Ragnar-Oock:
Explain how custom tags can be registered to avoid false positives when using them in comments.

related issue #11392


camc314 (approved):
thanks

PR #11408, opened 2025-05-31
refactor(mangler): use external allocator

camchenry:
related to https://github.com/oxc-project/oxc/issues/11347

### DIFF
--- a/crates/oxc_mangler/src/lib.rs
+++ b/crates/oxc_mangler/src/lib.rs
@@ -37,6 +37,26 @@ pub struct MangleOptions {
 
 type Slot = usize;
 
+/// Enum to handle both owned and borrowed allocators. This is not `Cow` because that type
+/// requires `ToOwned`/`Clone`, which is not implemented for `Allocator`. Although this does
+/// incur some pointer indirection on each reference to the allocator, it allows the API to be
+/// more ergonomic by either accepting an existing allocator, or allowing an internal one to
+/// be created and used temporarily automatically.
+enum TempAllocator<'t> {
+    Owned(Allocator),
+    Borrowed(&'t Allocator),
+}
+
+impl TempAllocator<'_> {
+    /// Get a reference to the allocator, regardless of whether it's owned or borrowed
+    fn as_ref(&self) -> &Allocator {
+        match self {
+            TempAllocator::Owned(allocator) => allocator,
+            TempAllocator::Borrowed(allocator) => allocator,
+        }
+    }
+}
+
 /// # Name Mangler / Symbol Minification
 ///
 /// ## Example
@@ -155,15 +175,78 @@ type Slot = usize;
 /// - slot 1: `top_level_b`, `foo_a`, `bar_a`
 /// - slot 2: `foo`
 /// - slot 3: `bar`
-#[derive(Default)]
-pub struct Mangler {
+pub struct Mangler<'t> {
     options: MangleOptions,
+    /// An allocator meant to be used for temporary allocations during mangling.
+    /// It can be cleared after mangling is done, to free up memory for subsequent
+    /// files or other operations.
+    temp_allocator: TempAllocator<'t>,
 }
 
-impl Mangler {
+impl Default for Mangler<'_> {
+    fn default() -> Self {
+        Self {
+            options: MangleOptions::default(),
+            temp_allocator: TempAllocator::Owned(Allocator::default()),
+        }
+    }
+}
+
+impl<'t> Mangler<'t> {
     #[must_use]
     pub fn new() -> Self {
         Self::default()
+    }
+
+    /// Creates a new `Mangler` using an existing temporary allocator. This is an allocator
+    /// that can be reset after mangling and is only used for temporary allocations during
+    /// the mangling process. This makes processing multiple files at once much more efficient,
+    /// because the same memory can be used for mangling each file.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use oxc_allocator::Allocator;
+    /// use oxc_mangler::Mangler;
+    /// use oxc_parser::Parser;
+    /// use oxc_span::SourceType;
+    ///
+    /// let allocator = Allocator::default();
+    /// let mut temp_allocator = Allocator::default();
+    /// let source = "function myFunction(param) { return param + 1; }";
+    ///
+    /// let parsed = Parser::new(&allocator, source, SourceType::mjs()).parse();
+    /// let mangled_symbols = Mangler::new_with_temp_allocator(&temp_allocator)
+    ///     .build(&parsed.program);
+    ///
+    /// // Reset the allocator to free temporary memory
+    /// temp_allocator.reset();
+    /// ```
+    ///
+    /// Processing multiple files:
+    ///
+    /// ```rust
+    /// # use oxc_allocator::Allocator;
+    /// # use oxc_mangler::Mangler;
+    /// # use oxc_parser::Parser;
+    /// # use oxc_span::SourceType;
+    /// let allocator = Allocator::default();
+    /// let mut temp_allocator = Allocator::default();
+    /// let files = ["function foo() {}", "function bar() {}"];
+    ///
+    /// for source in files {
+    ///     let parsed = Parser::new(&allocator, source, SourceType::mjs()).parse();
+    ///     let mangled_symbols = Mangler::new_with_temp_allocator(&temp_allocator)
+    ///         .build(&parsed.program);
+    ///     temp_allocator.reset(); // Free memory between files
+    /// }
+    /// ```
+    #[must_use]
+    pub fn new_with_temp_allocator(temp_allocator: &'t Allocator) -> Self {
+        Self {
+            options: MangleOptions::default(),
+            temp_allocator: TempAllocator::Borrowed(temp_allocator),
+        }
     }
 
     #[must_use]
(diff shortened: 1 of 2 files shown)


camchenry:
@overlookmotel I thought I'd give this a try, since the mangler benchmark keeps bothering me in my PRs. How much does it matter whether we use two separate allocators or the same allocator? I've used a mix of both here, since I wasn't sure. I use two separate allocators in the benchmark but just one in the compiler.

crates/oxc/src/compiler.rs
```diff
@@ -279,8 +280,13 @@ pub trait CompilerInterface {
...
 
-    fn mangle(&self, program: &mut Program<'_>, options: MangleOptions) -> Scoping {
-        Mangler::new().with_options(options).build(program)
+    fn mangle(
+        &self,
+        allocator: &Allocator,
+        program: &mut Program<'_>,
+        options: MangleOptions,
+    ) -> Scoping {
+        Mangler::new(allocator).with_options(options).build(program)
```
overlookmotel: This allocator is just used as temporary "scratch space". It's quite confusing having multiple allocators in play, so I suggest `temp_allocator: &Allocator` would be clearer.

A doc comment explaining the purpose of this allocator would also be helpful.

Style nit: Usually we pass `Allocator` as the last parameter to methods.

crates/oxc_mangler/src/lib.rs
```diff
@@ -155,15 +155,15 @@ type Slot = usize;
 /// - slot 1: `top_level_b`, `foo_a`, `bar_a`
 /// - slot 2: `foo`
 /// - slot 3: `bar`
-#[derive(Default)]
-pub struct Mangler {
+pub struct Mangler<'m> {
     options: MangleOptions,
+    allocator: &'m Allocator,
```
overlookmotel: Similar to comment above. I suggest:

```suggestion
    temp_allocator: &'t Allocator,
```

(`'t` short for "temp")

crates/oxc_mangler/src/lib.rs
```diff
@@ -300,13 +298,13 @@ impl Mangler {
             &keep_name_symbols,
             total_number_of_slots,
             &slots,
-            &allocator,
+            self.allocator,
```
overlookmotel: `tally_slot_frequencies` already takes `&self`, so no need to pass the `Allocator` here - it can access it via `self.temp_allocator`.

overlookmotel (commented):
> How much does it matter whether we use two separate allocators or the same allocator? I've used a mix of both here, since I wasn't sure. I use two separate allocators in the benchmark but just one in the compiler.

This is the tricky question!

We're using this new allocator as temporary "scratch space" only. None of the data in it needs to persist after mangling is finished.

If you use a different `Allocator` for each AST, and keep them around for a longish period (as Rolldown does), ideally you don't want to end up hanging on to a big chunk of ephemeral data too.

Therefore it's preferable to use a separate temp allocator, and not put this temp data into the AST allocator.

Additionally, if you have a separate temp allocator, you can reuse it for every AST. This will have a perf benefit as you're constantly reusing the same chunks of memory, which will likely stay warm in the cache, reducing cache misses (as you're aware - it was your PR speeding up the linter some months ago that really turned me on to the huge perf gain of efficient cache usage).

In the longer term, we could also adopt this pattern in transformer, minifier etc - like the mangler, they also create lots of temp structures, which can be discarded as soon as the pass is done. So one "temp allocator" could be used over and over again throughout all parts of the pipeline - and then reused *again* for the next file. https://github.com/oxc-project/backlog/issues/121

But... let's not get ahead of ourselves!

Having to pass around multiple allocators everywhere is unergonomic to say the least, and very likely users won't understand it, and will use the same allocator for everything. To make correct usage ergonomic and obvious, we probably need a single type (`AllocatorPool`?) which doles out multiple allocators, and handles all this for you.

But I think we should say that's out of scope for now. Mainly right now we're trying to fix the crazy benchmark variance.

So...

As a first step, I suggest not changing the existing external APIs. `Mangler::new` could remain as it was - not taking an `&Allocator` as param. It would create an `Allocator` internally and then pass it to `Mangler::new_with_temp_allocator`.

It's `Mangler::new_with_temp_allocator` which we'd use in the benchmarks.

`Mangler::new_with_temp_allocator` can be a public API, but for "advanced usage". If we document how it works, and its perf advantages, users who really want to push on perf can take advantage of it, but we don't impose confusing APIs on most users who are likely unfamiliar with custom allocators, and for whom probably 1 `Allocator` is already quite weird, let alone 2!

Later on we can try to find a better way to make "high perf" option the default, but without imposing wacky APIs.

What do you think? I'm not completely set on this view, but it feels like a somewhat pragmatic approach.

overlookmotel:
PS Sorry it's taken me so long to come back on this PR. It didn't show up in my notifications, for some reason.

camchenry:
> Having to pass around multiple allocators everywhere is unergonomic to say the least, and very likely users won't understand it, and will use the same allocator for everything. To make correct usage ergonomic and obvious, we probably need a single type (`AllocatorPool`?) which doles out multiple allocators, and handles all this for you.

I think this makes a lot of sense to me: it's sort of what I was thinking we should implement as I was working on this. I don't really expect users to be able to normally manage all of the allocators correctly to maximize performance.

I'll keep the external APIs the same and then we can change it in the future once we have a better plan for managing allocators automatically.

camchenry:
@overlookmotel Thanks for the thorough review! I think this has a much more minimal diff now. The main difference is the new `new_with_temp_allocator` API, and the addition of a `temp_allocator` field that supports both owned allocators and borrowed allocators. While I don't like the extra indirection for the allocator, it seemed the easiest way to support both something like `Mangler::new()` and `Mangler::new_with_temp_allocator(allocator)` without doing more rearchitecting.

crates/oxc_mangler/src/lib.rs
```diff
@@ -155,17 +175,80 @@ type Slot = usize;
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// use oxc_allocator::Allocator;
+    /// use oxc_mangler::Mangler;
+    /// use oxc_parser::Parser;
+    /// use oxc_span::SourceType;
+    ///
+    /// let allocator = Allocator::default();
+    /// let temp_allocator = Allocator::default();
```
overlookmotel:
```suggestion
    /// let mut temp_allocator = Allocator::default();
```

crates/oxc_mangler/src/lib.rs
```diff
@@ -155,17 +175,80 @@ type Slot = usize;
...
+    ///
+    /// Processing multiple files:
+    ///
+    /// ```rust
+    /// # use oxc_allocator::Allocator;
+    /// # use oxc_mangler::Mangler;
+    /// # use oxc_parser::Parser;
+    /// # use oxc_span::SourceType;
+    /// let allocator = Allocator::default();
+    /// let temp_allocator = Allocator::default();
```
overlookmotel:
```suggestion
    /// let mut temp_allocator = Allocator::default();
```

crates/oxc_mangler/src/lib.rs
```diff
@@ -387,12 +471,12 @@ impl Mangler {
...
         slots: &[Slot],
-        allocator: &'a Allocator,
     ) -> Vec<'a, SlotFrequency<'a>> {
         let root_scope_id = scoping.root_scope_id();
         let mut frequencies = Vec::from_iter_in(
-            repeat_with(|| SlotFrequency::new(allocator)).take(total_number_of_slots),
-            allocator,
+            repeat_with(|| SlotFrequency::new(self.temp_allocator.as_ref()))
+                .take(total_number_of_slots),
+            self.temp_allocator.as_ref(),
```
overlookmotel: Probably preferable to hoist `let temp_allocator = self.temp_allocator.as_ref();` to top of the function. Not sure if compiler is smart enough to do that for us - it may end up executing `self.temp_allocator.as_ref()` repeatedly for every call to the `repeat_with` closure.

overlookmotel (commented):
Thanks for taking my feedback on board.

A little bit annoying that we need `TempAllocator`, but I can't see a good way around it with current API. I'm not a fan of this aspect of the builder pattern where you have to initialize all fields at the start. We [have a workaround for this in parser](https://github.com/oxc-project/oxc/blob/9b475ad05b750f99762d63094174be6f6fc3c0eb/crates/oxc_parser/src/lib.rs) with an external `Parser` type, and then `ParserImpl`, allowing creating the main type in `build` method, instead of `new`. But that's a bit crap too! 🤷

`TempAllocator::as_ref` is only called a few times, so I think not a practical concern.

Apart from small comments below, all looks good!

camchenry:
> A little bit annoying that we need `TempAllocator`, but I can't see a good way around it with current API. I'm not a fan of this aspect of the builder pattern where you have to initialize all fields at the start. We [have a workaround for this in parser](https://github.com/oxc-project/oxc/blob/9b475ad05b750f99762d63094174be6f6fc3c0eb/crates/oxc_parser/src/lib.rs) with an external `Parser` type, and then `ParserImpl`, allowing creating the main type in `build` method, instead of `new`. But that's a bit crap too! 🤷

Good to know, I wasn't sure my intuition here was correct, but it sounds like this type is mostly unavoidable for now. I don't think it's a big deal either way.

overlookmotel:
Hope this sorts out the wavy benchmarks. We'll find out!

PR #11409, opened 2025-05-31
refactor(mangler): use bump allocator for internal storage

camchenry:


camchenry:
* **#11409** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11409?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11409?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#11408** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11408?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11409).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):
> ⚡	`mangler[RadixUIAdoptionSection.jsx]`	15.1 µs	14.2 µs	+6.57%

Nice! I am *guessing* that this isn't just noise. `RadixUIAdoptionSection.jsx` is the smallest test fixture, so it makes sense it moves the needle most here, as the allocations are a larger slice of the overall workload. Personally I think it's the most important benchmark we have as it's most representative of real-world code. Most people don't write 50,000-line files like `checker.ts`!

It's a shame that `FixedBitset` doesn't support custom allocators, or we could put them into the temp arena too. They're ideal candidates for putting in an arena as they never grow (high cost of growth is the only perf downside of arena-allocated data structures).

Probably we could write our own version of `FixedBitset` which allocates in an arena. We don't use most of `FixedBitset`'s APIs, and I don't think the APIs we do use benefit from SIMD-acceleration, so our version could be a lot simpler. That could be fun! But, to be honest, mangler is so fast anyway that it's not the first place we should be looking for optimizations.

camchenry:
> It's a shame that `FixedBitset` doesn't support custom allocators, or we could put them into the temp arena too. They're ideal candidates for putting in an arena as they never grow (high cost of growth is the only perf downside of arena-allocated data structures).

We could at least put this into an arena-owned `Vec` if our Vec implementation supported `resize_with` by the way: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize_with. The mangler is currently using this method with the `FixedBitSet` Vec. I didn't want to implement it in this PR stack but that's a future improvement we can make here as well.

camchenry:
Perf improvement appears to be fairly consistent after refactoring as well: 

⚡ 	mangler[RadixUIAdoptionSection.jsx] 	15 µs 	14.2 µs 	+5.59%

overlookmotel:
> We could at least put this into an arena-owned `Vec` if our Vec implementation supported `resize_with`

I don't think we can, unfortunately. `oxc_allocator::Vec` does not allow storing types which own heap memory (are `Drop`), because `Vec` does nothing on drop. So it'd leak memory if it contained types which rely on `drop` being called to free the memory they own. This is by design - it makes dropping an AST very fast. But consequently, putting `FixedBitSet`s in arena would require a new impl which stores its data in arena, instead of on the heap, so it doesn't require `Drop`.

That said, we should implement `Vec::resize_with`. I'm not sure why we don't have it.

camchenry:
> I don't think we can, unfortunately. `oxc_allocator::Vec` does not allow storing types which own heap memory (are `Drop`), because `Vec` does nothing on drop. So it'd leak memory if it contained types which rely on `drop` being called to free the memory they own. This is by design - it makes dropping an AST very fast. But consequently, putting `FixedBitSet`s in arena would require a new impl which stores its data in arena, instead of on the heap, so it doesn't require `Drop`.

Makes sense, I thought it was simply the lack of `FixedBitSet::with_capacity_in` or a similar function, thanks for the explanation.

camchenry:
### Merge activity

* **Jun 8, 10:39 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jun 8, 10:49 PM UTC**: `graphite-app` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jun 8, 10:58 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #11416, opened 2025-05-31
perf(parser): optimize around `eat_decorators`

ulrichstark:
- move `Kind::At` check to call site of `eat_decorators` like other calls
- remove surrounding `try_parse` calls surrounding calls of `eat_decorators`

I'm not sure about the removed `try_parse`...
Tests are still green and no diffs in snapshots.
Seems like it didn't break anything 🤞

### DIFF
--- a/crates/oxc_parser/src/js/function.rs
+++ b/crates/oxc_parser/src/js/function.rs
@@ -81,7 +81,9 @@ impl<'a> ParserImpl<'a> {
 
     fn parse_formal_parameter(&mut self) -> FormalParameter<'a> {
         let span = self.start_span();
-        self.eat_decorators();
+        if self.at(Kind::At) {
+            self.eat_decorators();
+        }
         let modifiers = self.parse_parameter_modifiers();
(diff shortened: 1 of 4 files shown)


Boshen:
Seems like a good change? It removes a function call from these hot paths.

ulrichstark:
> Seems like a good change? It removes a function call from these hot paths.

Okay, just reopened. Feel free to merge or close after reviewing. I closed this PR, because of following reasons in descending importance:
- No real visible benchmark improvements
- I'm not sure if the removed `has_trailing_decorator` value or `try_parse` calls might have broken something the conformance snapshots don't cover
- More lines of code

ulrichstark:
Just rebased to trigger CI.

Boshen:
> No real visible benchmark improvements

Our benchmark files don't have any decorators :-/

> I'm not sure if the removed has_trailing_decorator value or try_parse calls might have broken something the conformance snapshots don't cover

it's in the tsc code ... but it should be removed if no test failures.

PR #11464, opened 2025-06-03
refactor(parser): Reduce backtracking during type alias parsing

therewillbecode:
- Replace use of lookahead when parsing type alias declarations in favour of checkpoint and rewinds.
- Remove unnecessary backtracking when parsing type alias declarations.

Relates to  #11334  as part of reducing unnecessary backtracking in the parser.

### DIFF
--- a/crates/oxc_parser/src/ts/statement.rs
+++ b/crates/oxc_parser/src/ts/statement.rs
@@ -126,17 +126,12 @@ impl<'a> ParserImpl<'a> {
         let params = self.parse_ts_type_parameters();
         self.expect(Kind::Eq);
 
-        let at_intrinsic = self.at(Kind::Intrinsic);
         let checkpoint = self.checkpoint();
-        let is_next_token_dot = self.is_next_token_dot();
-        // self.bump_any();
-        //self.at(Kind::Dot)
-        self.rewind(checkpoint);
-        let ty = if at_intrinsic && !is_next_token_dot {
+        let ty = if self.at(Kind::Intrinsic) && !self.is_next_token_dot() {
             let span = self.start_span();
-            self.bump_any();
             self.ast.ts_type_intrinsic_keyword(self.end_span(span))
         } else {
+            self.rewind(checkpoint);
             self.parse_ts_type()
         };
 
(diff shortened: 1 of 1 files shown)


crates/oxc_parser/src/ts/statement.rs
```diff
@@ -126,11 +126,12 @@ impl<'a> ParserImpl<'a> {
         let params = self.parse_ts_type_parameters();
         self.expect(Kind::Eq);
 
-        let ty = if self.at(Kind::Intrinsic) && !self.lookahead(Self::is_next_token_dot) {
+        let checkpoint = self.checkpoint();
+        let ty = if self.at(Kind::Intrinsic) && !self.is_next_token_dot() {
             let span = self.start_span();
-            self.bump_any();
             self.ast.ts_type_intrinsic_keyword(self.end_span(span))
```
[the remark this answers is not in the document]
therewillbecode: `bump_any` is called inside `is_next_token_dot`. With this PR's changes `bump_any` is now called before `start_span`. However if the order of calling `bump_any` and `start_span` mattered, I would expect to see conformance tests failing.

crates/oxc_parser/src/ts/statement.rs
```diff
@@ -126,11 +126,12 @@ impl<'a> ParserImpl<'a> {
         let params = self.parse_ts_type_parameters();
         self.expect(Kind::Eq);
 
-        let ty = if self.at(Kind::Intrinsic) && !self.lookahead(Self::is_next_token_dot) {
+        let checkpoint = self.checkpoint();
+        let ty = if self.at(Kind::Intrinsic) && !self.is_next_token_dot() {
```
Boshen, line 130: This is not ideal. We hid token bump inside `is_next_token_dot`.

crates/oxc_parser/src/ts/statement.rs
```diff
@@ -126,11 +126,12 @@ impl<'a> ParserImpl<'a> {
...
             let span = self.start_span();
-            self.bump_any();
             self.ast.ts_type_intrinsic_keyword(self.end_span(span))
         } else {
+            self.rewind(checkpoint);
```
Boshen, line 134: This is the hot path we wish to remove, not the cold path on the `intrinsic` keyword.

Boshen (changes requested):

crates/oxc_parser/src/ts/statement.rs
therewillbecode: Just curious, how did you know its a hot path?

Issue #11490, opened 2025-06-04
Make `AstKind` consistent


overlookmotel:
To support printing comments in formatter and codegen, we are discussing adding a `CommentNodeId` field to every AST type (#11213).

However, what we *really* want is standard `NodeId` on every AST node.

### Why?

At present, when you have an `AstNode`, via `AstKind` you can get the actual AST type (e.g. `BinaryExpression`). However you can't go in the opposite direction e.g. start with a `BinaryExpression`, obtain its `NodeId`, and use `Semantic` to e.g. get its parent node.

If we store `NodeId` in all AST structs, we'll be able to go in both directions, which allows moving around and querying the AST much more easily.

* https://github.com/oxc-project/oxc/discussions/5689
* https://github.com/oxc-project/oxc/issues/11048

### What does `AstKind` have to do with this?

The main blocker to achieving this is that `AstKind` is rather inconsistent:

* Most AST types which are structs have a corresponding `AstKind`, but some don't.
* Most AST types which are enums do *not* have a corresponding `AstKind`, but a few do!

Only types which have an `AstKind` get a `NodeId`. So, to align `NodeId` in `Semantic` with `NodeId`s stored in AST structs, we need to standardize what types have `AstKind`s.

### What we need to do

We need to:

* Add `AstKind`s for all AST structs (e.g. `StaticMemberExpression`).
* Remove `AstKind`s for all AST enums (e.g. `MemberExpression`).

### How to do it

`oxc_ast_tools` contains 2 lists of the exceptions to the above rules:

https://github.com/oxc-project/oxc/blob/7e884511bda981f85ff3472f748134ab60c3aec8/tasks/ast_tools/src/generators/ast_kind.rs#L28-L83

We need to reduce both these lists to zero.

We can approach this goal in small steps.

e.g. for `MemberExpression`:

* Add `AstKind`s for `ComputedMemberExpression`, `StaticMemberExpression` and `PrivateFieldExpression` (by removing them from `STRUCTS_BLACK_LIST`).
* Remove `AstKind` for `MemberExpression` (by removing it from `ENUMS_WHITE_LIST`).
* Run `cargo run -p oxc_ast_tools` to regenerate the definition of `AstKind` and `Visit` etc.
* Update all code which used the old `AstKind::MemberExpression` to use the new `AstKind::StaticMemberExpression` etc instead.

The changes will be mostly in the linter, which is the heaviest user of `AstKind`.

e.g. in `eslint/no-caller` rule:

[...]

PR #11535, opened 2025-06-06
refactor(ast): get jsx types out of AstKind exceptions

ulrichstark:
This is my (not so confident) attempt to do #11490 for all jsx types.

1. removed all jsx types from STRUCTS_BLACK_LIST and ENUMS_WHITE_LIST
2. regenerated ast code
3. added missing arms to AstKind::debug_name
4. removed now broken is_jsx function, because it was unused
5. fixed all compile errors
6. fixed all test errors
7. regenerated snapshots (different node ids and decreased prettier conformance)

All tests are green 🥳🎉


ulrichstark:
Less code and even some nice performance improvements 😎
image

overlookmotel:
> ⚡	`linter[RadixUIAdoptionSection.jsx]`	2.8 ms	2.6 ms	+7.92%
> ⚡	`semantic[RadixUIAdoptionSection.jsx]`	72.4 µs	69.8 µs	+3.64%

Banger!

Thanks so much for getting on this. It's a large PR, so will take me a while to get through reviewing it. But, on the face of it, looks great.

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
```diff
@@ -55,11 +55,8 @@ impl Rule for NoMultiStr {
     }
 }
 
-fn is_within_jsx_attribute_item(id: NodeId, ctx: &LintContext) -> bool {
-    if matches!(ctx.nodes().parent_kind(id), Some(AstKind::JSXAttributeItem(_))) {
-        return true;
-    }
-    false
+fn is_within_jsx_attribute(id: NodeId, ctx: &LintContext) -> bool {
+    matches!(ctx.nodes().parent_kind(id), Some(AstKind::JSXAttribute(_)))
```
overlookmotel, line 59: This changes behavior slightly. Previously it would have matched `JSXSpreadAttribute` too. But I'm not sure if that's a problem.

crates/oxc_linter/src/rules/eslint/no_useless_escape.rs
```diff
@@ -152,11 +152,8 @@ impl Rule for NoUselessEscape {
     }
 }
 
-fn is_within_jsx_attribute_item(id: NodeId, ctx: &LintContext) -> bool {
-    if matches!(ctx.nodes().parent_kind(id), Some(AstKind::JSXAttributeItem(_))) {
-        return true;
-    }
-    false
+fn is_within_jsx_attribute(id: NodeId, ctx: &LintContext) -> bool {
+    matches!(ctx.nodes().parent_kind(id), Some(AstKind::JSXAttribute(_)))
```
overlookmotel, line 156: Same as previous comment.

crates/oxc_linter/src/rules/nextjs/no_duplicate_head.rs
```diff
@@ -106,7 +106,7 @@ impl Rule for NoDuplicateHead {
             }
 
             if !matches!(
-                nodes.ancestor_ids(reference.node_id()).nth(2).map(|node_id| nodes.kind(node_id)),
+                nodes.ancestor_ids(reference.node_id()).nth(1).map(|node_id| nodes.kind(node_id)),
```
overlookmotel: I'm not 100% sure, but can this be shortened by using `.parent_kind()`?

overlookmotel:
Actually, it wasn't so long. I've reviewed. From what I can see, looks good. But I'm not the best person to review the linter changes as I'm not so familiar with the linter.

1. @camc314 Could you possibly take a look at the linter changes (and not just the ones I've commented on)? As we're planning to drop Oxlint 1.0 next week, and make a big song and dance about it, it would be a really unfortunate moment to break things!

2. The reduced conformance pass rate in formatter suggests something may be wrong there. I know literally *nothing* about the formatter. @Dunqing Can you advise what we should do about that?

Note: The perf boost in Semantic I'm guessing is due to producing less `AstNode`s (the snapshot changes with lower `NodeId`s suggests that). I guess the boost in linter is just due to simpler code (less branches and lookups), but I'm amazed it has such a large effect. Bravo!

@ulrichstark Just to flag please don't get your hopes up that every further step we take on adding/removing `AstKind`s will get a similar perf boost. In some cases it'll probably result in more `AstNode`s, in others less. So perf may go up or down. But this is certainly a very pleasant first step down the road!

overlookmotel:
Not sure what's going on with the "Check links" CI fail. I've re-run it twice and it keeps failing on https://babeljs.io/docs/babel-plugin-transform-typescript, which seems to be a valid link.

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
camc314: :+1: we should restore the previous behaviour I think.

and possibly add a test for this case?

tasks/prettier_conformance/snapshots/prettier.ts.snap.md
```diff
@@ -9,7 +9,7 @@ ts compatibility: 186/573 (32.46%)
 | jsx/comments/eslint-disable.js | 💥 | 0.00% |
 | jsx/comments/in-attributes.js | 💥 | 41.67% |
 | jsx/comments/in-end-tag.js | 💥 | 32.18% |
-| jsx/comments/in-tags.js | 💥 | 53.33% |
+| jsx/comments/in-tags.js | 💥 | 48.89% |
```
camc314, line 12: these prettier changes are possible worth investigating, but we could just do it in a follow up pr

camc314 (commented):
linter changes look good excluding those two @overlookmotel caught

Dunqing:
There's no need to worry about the Formatter; I've created #11511 to revamp the Formatter core. I will revisit all `AstKind` use cases in Formatter after that work is done.

Boshen:
@camc314 merge after oxlint 1.0

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
overlookmotel: Reason I wasn't sure if the change matters is that I'm not sure if a `StringLiteral` inside a `JSXSpreadAttribute` is relevant to the rule.

```jsx
<Foo {..."bar"} />
```

That is transformed to:

```js
import _objectSpread from "@oxc-project/runtime/helpers/objectSpread2";
import { jsx as _jsx } from "react/jsx-runtime";
/* @__PURE__ */ _jsx(Foo, _objectSpread({}, "bar"));
```

And `_objectSpread({}, "bar")` evaluates to `{ '0': 'b', '1': 'a', '2': 'r' }`!

So in this bizarre case, should the rule produce an error or not? I suspect probably it should, in which case the change in this PR is correct. But I doubt this code would ever appear anywhere anyway, so it's rather academic.

camc314: ah so this is confusing:
`AstKind::JSXAttributeItem` points to `JSXAttributeItem`

https://github.com/oxc-project/oxc/blob/12690a10bfb080ae64a2782ac43dbb65ff4cc0f3/crates/oxc_ast/src/ast/jsx.rs#L308-L313

but we've already got `AstKind::JSXSpreadAttribute`

https://github.com/oxc-project/oxc/blob/62c3a4a5764ddf8a55bc337a7ef49aaedc8e7dfc/crates/oxc_ast/src/generated/ast_kind.rs#L304

so tldr is that this change keeps the same behaviour... but looks like there are some bugs in the impl of this rule (e.g. it doesn't handle parenthesised expressions correctly)

for reference, this is how eslint checks:

https://github.com/eslint/eslint/blob/46eea6d1cbed41d020cb76841ebba30710b0afd0/lib/rules/no-multi-str.js#L45-L47

This change is ok, we can fix the bugs in the rule in a different PR

ulrichstark:
Just did a commit to shorten the call to get the parent node kind. The other feedback around the two instances of is_within_jsx_attribute seems to be resolved? Is more work required from my side or are we good and wait for the oxlint 1.0 release?

crates/oxc_linter/src/rules/eslint/no_useless_escape.rs
camc314: good catch. i've checked and this change is ok

camc314:
> Is more work required from my side or are we good and wait for the oxlint 1.0 release?

nope, i don't think so. I've checked both changes and they look fine. thanks for your work on this!

crates/oxc_linter/src/rules/eslint/no_multi_str.rs
overlookmotel: Ah so I was wrong in the first place. I hadn't clocked that there was already `AstKind::JSXSpreadAttribute`, so if parent was previously `AstKind::JSXAttributeItem`, now it can only be `AstKind::JSXAttribute`. So this change does not alter behavior after all!

This is one reason why I think making what types do and don't have `AstKind`s consistent - it's confusing as hell!

@ulrichstark Just for future reference, it might have been preferable to tackle each individual type in a separate PR, as it would have made it easier to reason about. If you intend to continue and tackle some more, we can look at adding you to Graphite so you can use "stacked PRs". Graphite makes it much easier to break up changes into multiple PRs without getting merge conflicts.

crates/oxc_semantic/src/checker/typescript.rs
```diff
@@ -556,7 +556,10 @@ pub fn check_jsx_expression_container(
     ctx: &SemanticBuilder<'_>,
 ) {
     if matches!(container.expression, JSXExpression::EmptyExpression(_))
-        && matches!(ctx.nodes.parent_kind(ctx.current_node_id), Some(AstKind::JSXAttributeItem(_)))
+        && matches!(
+            ctx.nodes.parent_kind(ctx.current_node_id),
+            Some(AstKind::JSXAttribute(_) | AstKind::JSXSpreadAttribute(_))
+        )
```
overlookmotel: Given the conversation above, shouldn't this change just be:

```diff
- && matches!(ctx.nodes.parent_kind(ctx.current_node_id), Some(AstKind::JSXAttributeItem(_)))
+ && matches!(ctx.nodes.parent_kind(ctx.current_node_id), Some(AstKind::JSXAttribute(_)))
```

?

ulrichstark: You are correct. I can't think of any case in which JSXExpressionContainer is inside a JSXSpreadAttribute. So only checking for JSXAttribute makes more sense.

overlookmotel:
Nice work @ulrichstark!

PR #11617, opened 2025-06-11
refactor(ast): don't generate AstKind for ForStatementInit

ulrichstark:
Part of #11490

- two lint rules needed change
- formatter needed change (conformance went slightly up)


Dunqing (approved):
Looks good. Thanks!

PR #11651, opened 2025-06-12
fix(linter): add missing menuitemradio and menutitemcheckbox roles

dflynn15:
# Why?

In testing `oxlint` in a few repositories, I noticed that this rule was missing two roles to check for in the required section. This has been working in the ESLint version due to its use of `aria-query` to help ensure all roles and required aria props were met.

## Short-term

This patches the missing instances and ensures the rule is working correctly.

## Long-term

I'm putting a PoC together to refactor `globals.rs` and the `jsx_a11y` rules to be fully compliant with WAI-ARIA 1.2 and present a possible _tiny_ speed boost that will help me also write the remaining rules I depend on preventing me from adopting `oxlint` at this time.


camc314 (approved):
thank you!

Issue #11668, opened 2025-06-13
linter: `typescript/prefer-function-type` not match some case


CoffeeChaton:
### What version of Oxlint are you using?

 1.1.0

### What command did you run?

`oxlint`

### What does your `.oxlintrc.json` config file look like?

```jsonc
{
    "$schema": "./node_modules/oxlint/configuration_schema.json",
    "ignorePatterns": [
        "node_modules"
    ],
    "rules": {
        // https://oxc.rs/docs/guide/usage/linter/rules/typescript/prefer-function-type.html#typescript-prefer-function-type
        "typescript/prefer-function-type": "error"
    }
}

```

### What happened?

```ts
export class CC {
    public foo() {
        //
    }
}

export function FN<T extends CC>(resultClass: { new(): T }): T {
    //                                        ^^^^^^^^^^^^
    //   8:49  error  Type literal only has a call signature, you should use a function type instead  @typescript-eslint/prefer-function-type

    // eslint is match   // https://typescript-eslint.io/rules/prefer-function-type/
    // oxlint not match // https://oxc.rs/docs/guide/usage/linter/rules/typescript/prefer-function-type.html#typescript-prefer-function-type
    return new resultClass();
}

```

`typescript-eslint` can find it, and auto-fix
`oxlint` not find it , and `eslint-plugin-oxlint` disabled `typescript-eslint` diagnostics.

https://github.com/CoffeeChaton/oxlint--typescript--prefer-function-type

PR #11672, opened 2025-06-13
refactor(linter): prefer-function-type cleanup

braddunbar:
I started looking into #11668 and as a way of getting familiar with things, I refactored a little bit to better understand what's happening in this file. Mostly, I'm reducing nesting with let-else and early returns. Commit by commit it's super straightforward.

This is my first contribution here, so please let me know what I've missed and how I can fix it! 🙏


camc314 (approved):
this is great, thanks for using and contributing

PR #11681, opened 2025-06-13
fix(linter/consistent-function-scoping): allow functions in TS modules/namespaces

DonIsaac:
## What This PR Does
Fixes a false positive in `unicorn/consistent-function-scoping` on non-exported functions in TypeScript `module` and `namespace` nodes.

TypeScript modules and namespaces are valid ways to encapsulate code.
```ts
export namespace Foo {
  // public-facing API
  export function doThing() {
     let x = privateUtility();
     return x + 2;
  }

  // private function only available to functions within this namespace.
  // this used to be reported.
  function privateUtility() {
     return 1;
  }
}
```


camc314 (approved):
thank you!

PR #11710, opened 2025-06-14
feat(playground): Allow specifying a JSON string as the linter config

nrayburn-tech:
When creating bug reports it would be helpful to have a link to a reproduction that just worked in the browser. By adding the ability to config Oxlint here, simple reproductions can be done within the Playground and URLs shared from it.

This would be a lot better if the `config` was a JSON object instead of a string, but I don't know how to do that only on the Rust side.

An option would be to keep the Rust side a string, but on the browser side allow it to be an object. Whenever the object changes, we serialize that to a string before sending it to the Rust side. If there's interest in this, then I think I can make this change in the playground repo.

Example config and screenshots below.

```json
"linter": {},
```
image

```json
"linter": {
  "config": "{\"plugins\":[\"typescript\"],\"categories\":{\"correctness\":\"error\"},\"rules\":{\"no-debugger\":\"warn\"}}"
},
```
image

### DIFF
--- a/crates/oxc_linter/src/config/oxlintrc.rs
+++ b/crates/oxc_linter/src/config/oxlintrc.rs
@@ -149,6 +149,17 @@ impl Oxlintrc {
         Ok(config)
     }
 
+    pub fn from_string(json_string: &str) -> Result<Self, OxcDiagnostic> {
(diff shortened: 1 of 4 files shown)


camc314 (approved):
thank yoi!

PR #11714, opened 2025-06-15
refactor(language_server): Use rule name directly from OxcCode instead of parsing out of the stringified version of OxcCode

nrayburn-tech:


camc314 (approved):
thanks!

@Sysix mind havign a quick glance?

Sysix (approved):

Issue #10307, opened 2025-04-07
AST nodes with loc and/or range fields.


knightedcodemonkey:
It would be great to have AST nodes with a `loc` and/or `range` fields.

Here's some context: https://github.com/oxc-project/oxc/issues/10139#issuecomment-2783567533

Possibly dependent upon https://github.com/oxc-project/oxc/issues/9705 being completed first.

alexander-akait:
@overlookmotel yeah, I understand this, I wanted to do a fairly powerful testing through many webpack tests and places, but without these things, especially without `loc`, it is not possible, will be wait updates

overlookmotel:
More testing would be very welcome! Thank you.

On TS-ESLint alignment, in 2 weeks, we've gone from 1% of tests passing to 75%. So hopefully it won't be too long until #9705 is complete, and then we can look at this. By all means please get involved in the effort on #9705 if you'd like to. More hands will get us to the finish line faster.

overlookmotel:
#11728 will cover adding `range`.

### Next steps - `range`

After it's merged, we'll need to:

* [ ] Adapt ESTree conformance testing to check `range` field is present and correct in all AST nodes (it'd be easy to forget it in some of the types which have custom `ESTree` implementations).
* [x] Support `range` field in raw transfer too.
* [x] Update TS type definitions to include `range` field.

### `loc`

Unfortunately I think supporting `loc` is more tricky, due to UTF-8 / UTF-16 offsets.

Here are the steps I think we'd need to take to support it:

* [x] Move `start`, `end` and `range` to be the last properties in each node object.
* [x] Adapt `acorn-test262` to do the same, so snapshots in conformance test have these properties in right order.
* [ ] Remove the extra AST pass converting offsets UTF-8 -> UTF-16. Convert during serialization instead.
* [ ] Do the same in raw transfer.
* [ ] Alter the [offsets translation table](https://github.com/oxc-project/oxc/blob/80c87d47b33cfb2afbb802af9c9d8e511275859f/crates/oxc_ast_visit/src/utf8_to_utf16.rs) to also represent line breaks.
* [ ] Use the new translation table to also convert UTF-8 offset -> line + column during serialization.

The last 4 steps are non-trivial.

I have only a vague idea of what a new translation table would look like. Ideally we probably want a single translation table that performs both UTF-8 to UTF-16 offset conversion, *and* UTF-8 offset to line / column conversion - rather than 2 separate tables.

### Why do location properties have to be last?

[...]

Issue #6458, opened 2024-10-11
Get line number using oxc-parser


azat-io:
Hello!

Is there any way to get real line number of Node?
Can we add some option to `parseAsync` func to add information about it?

For example, I parse JS and read comments:

```ts
await oxc.parseAsync(source, { sourceFilename: 'test.js' })
return ast.comments
``` 

I get the following result:

```ts
[
  {
    value: ' Comment',
    type: 'Line',
    start: 2,
    end: 10,
  },
]
```

I need to get line number of this comment.

leaysgur:
It's planned, but at the moment we can't get them, even on the Rust side.

Ongoing issue: https://github.com/oxc-project/oxc/issues/959

As referenced there, the only way seems to be to calculate the position from the original file for now.

IWANABETHATGUY:
You can refer https://github.com/oxc-project/oxc/issues/5493

overlookmotel:
Closing, not because it's an unreasonable request, but because I think the answer is covered by #959 and #5493. If I've missed something that's not covered by those two, please feel free to come back.

bacarybruno:
Hello 👋
Now that https://github.com/oxc-project/oxc/pull/8687 is merged, can you please tell us what is the recommended way to get that information on `oxc-parser` ? 🙏

bacarybruno:
Hello @overlookmotel 👋
Now that https://github.com/oxc-project/oxc/pull/8687 is merged, can you please tell us what is the recommended way to get that information on oxc-parser ? 🙏

PR #11728, opened 2025-06-15
feat(napi/parser)!: add `range` option

bacarybruno:
Heyo team 👋!

I took a shot at https://github.com/oxc-project/oxc/issues/10307 to add `loc` and `range` fields. I followed this suggestion https://github.com/oxc-project/oxc/issues/10307#issuecomment-2786140446 by making the fields configurable.

I [needed to get this information](https://github.com/oxc-project/oxc/issues/6458#issuecomment-2774321478) from oxc-parser for different use cases including writing an eslint parser. Is this gets merged I would also need/like to add the a `tokens` option (to be discussed).

The JSDoc comments are based on https://typescript-eslint.io/packages/typescript-estree/#parsecode-options

P.S: I'm new to Rust and this is my first contrib on this repo, so feel free to suggest improvements. That would be really appreciated!


tasks/ast_tools/src/derives/estree.rs
```diff
@@ -435,9 +437,37 @@ impl<'s> StructSerializerGenerator<'s> {
 
         let field_name_ident = field.ident();
 
+        // Special case for `Span` fields - add conditional range and loc serialization
+        if let TypeDef::Struct(inner_struct_def) = field.type_def(self.schema) {
+            if inner_struct_def.name() == "Span" && field.name() == "span" {
+                self.stmts.extend(quote! {
+                    if state.range() {
+                        let range = oxc_estree::Range::new(#self_path.#field_name_ident.start, #self_path.#field_name_ident.end);
+                        state.serialize_field("range", &range);
```
overlookmotel: I don't think we need the `Range` type. Generating this code would also work:

```rs
state.serialize_field("range", &[#self_path.span.start, #self_path.span.end]);
```

Also, small nit: Can we put `range` after `start` and `end`, instead of before?

crates/oxc_estree/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #[cfg(feature = "serialize")]
-mod serialize;
+pub mod serialize;
+
```
overlookmotel: If we get rid of `Range` type, can revert this change, which will fix the clippy errors.

crates/oxc_estree/src/serialize/mod.rs
```diff
@@ -94,12 +105,21 @@ pub type PrettyFixesTSSerializer = ESTreeSerializer<ConfigFixesTS, PrettyFormatt
...
+    pub loc: bool,
+}
+
 /// ESTree serializer.
 pub struct ESTreeSerializer<C: Config, F: Formatter> {
     buffer: CodeBuffer,
     formatter: F,
     trace_path: NonEmptyStack<TracePathPart>,
     fixes_buffer: CodeBuffer,
+    options: RuntimeOptions,
```
overlookmotel: It'd be preferable to store the options in `Config` types, instead of in the `ESTreeSerializer`.

```rs
pub trait Config {
    fn ranges(&self) -> bool;
    // ...
}

pub struct ConfigTS {
    ranges: bool,
}

impl Config for ConfigTS {
    #[inline]
    fn ranges(&self) -> bool {
        self.ranges
    }
    // ...
}

// Ditto the other `Config` types

impl<C: Config, F: Formatter> ESTreeSerializer<C, F> {
    #[inline(always)]
    fn ranges(&self) -> bool {
        self.config.ranges()
    }
}
```

The advantage is that then you can create a `Config` which statically defines the value of `ranges`. We can use that to speed up the conformance runner by using a custom `Config`:

```rs
struct ConformanceConfig;

impl Config for ConformanceConfig {
    #[inline(always)]
    fn ranges(&self) -> bool {
        false
    }
    // ...
}
```

With that config, compiler should remove all the `if self.ranges()` branches from `ESTree` impls.

overlookmotel (commented):
Thanks very much for tackling this. Given that you say you're new to Rust, very impressive that you managed to figure out the codegen as well as everything else!

I've made a few comments below, but the larger problem is that, while the implementation for `ranges` is good, I don't think `loc` is going to work correctly.

### `loc`

The problem is UTF-8 vs UTF-16 offsets. Rust works with UTF-8 strings, whereas JS works in UTF-16. So there's an AST pass that happens before serialization which converts all the offsets from UTF-8 to UTF-16:

https://github.com/oxc-project/oxc/blob/ea6ce9de58d1621eacf3149eba7f9439d69234bc/napi/parser/src/lib.rs#L91-L92

https://github.com/oxc-project/oxc/blob/ea6ce9de58d1621eacf3149eba7f9439d69234bc/crates/oxc_napi/src/lib.rs#L12-L59

During serialization, you're using `Rope` to convert offsets to line/column. But `Rope` expects UTF-8 offsets, and by this point we've already converted them to UTF-16.

If the whole source text is ASCII, then there's no difference, but I think you'll find `loc` is incorrect for source text like this:

```js
[
"🍄🍄🍄",
123
]
```

Solving this problem is not so easy (especially making it work *and* performant).

I propose:

1. Scale back the scope of this PR to only add `range`s.
2. I'll write up on #10307 the steps I think we'd need to take to support `loc` too, and we can discuss.

### Custom serializers

This point applies to `range` too: `range` field also needs to be added manually in all the hand-written `ESTree` impls in `serialize` directory.

We'll also need to support it in raw transfer, but you can leave that to me - raw transfer is a bit of a minefield.

Ditto we'll need conformance testing for `range`, but I can handle that.

### Tests

In absence of conformance test coverage, could you please add some tests for the `range` option to https://github.com/oxc-project/oxc/blob/main/napi/parser/test/parse.test.ts.

No need for anything fancy, just 3 tests which check the output with `ranges: false`, `ranges: true`, and `ranges: undefined`.

overlookmotel:
I see you're making changes after my feedback, so marking as draft for now. Please let me know when it's ready for review again.

And thank you for your efforts!

napi/parser/src/types.rs
```diff
@@ -94,6 +101,11 @@ pub struct EcmaScriptModule {
 pub struct Span {
     pub start: u32,
     pub end: u32,
+    /// An array of two numbers.
+    /// Both numbers are a 0-based index which is the position in the array of source code characters.
+    /// The first is the start position of the node, the second is the end position of the node.
+    #[napi(ts_type = "[number, number]")]
+    pub range: Option<Vec<i32>>,
```
bacarybruno: Napi doesnt support `Option<[i32; 2]>`. Is there a better way to do this or is it good enough?

tasks/ast_tools/src/generators/raw_transfer.rs
```diff
@@ -477,6 +477,7 @@ fn generate_primitive(primitive_def: &PrimitiveDef, code: &mut String, schema: &
         "Atom" => return,
         // Dummy type
         "PointerAlign" => return,
+        "ArrayType" => return,
```
bacarybruno: I had to modify raw_tranfer generators but not sure about the right return value (or how it works in general)

tasks/ast_tools/src/generators/raw_transfer_lazy.rs
```diff
@@ -594,6 +594,7 @@ fn generate_primitive(primitive_def: &PrimitiveDef, code: &mut String, schema: &
         "Atom" => return,
         // Dummy type
         "PointerAlign" => return,
+        "ArrayType" => return,
```
bacarybruno: https://github.com/oxc-project/oxc/pull/11728/files#r2153722471

tasks/ast_tools/src/parse/parse.rs
```diff
@@ -226,6 +226,7 @@ impl<'c> Parser<'c> {
             // TODO: Remove the need for this by adding
             // `#[cfg_attr(target_pointer_width = "64", repr(align(8)))]` to all AST types
             "PointerAlign" => primitive("PointerAlign"),
+            "ArrayType" => primitive("ArrayType"),
```
bacarybruno: Not sure about the name `ArrayType`. Open to suggestion or maybe there is abeeter way to do it?

bacarybruno: Renamed to RangeType, but still

bacarybruno:
Hello @overlookmotel yes sorry for the noise. The PR is now ready on my side 🙏

tasks/ast_tools/src/derives/estree.rs
bacarybruno: ✅

crates/oxc_estree/src/lib.rs
bacarybruno: ✅

crates/oxc_estree/src/serialize/mod.rs
bacarybruno: ✅

crates/oxc_span/src/span.rs
```diff
@@ -85,6 +85,10 @@ pub struct Span {
     /// The zero-based end offset of the span. This may be equal to [`start`](Span::start) if
     /// the span is empty, but should not be less than it.
     pub end: u32,
+
+    /// A [number, number] which indicates the start/end index of the node in the file contents.
+    pub range: Option<[i32; 2]>,
```
overlookmotel: We definitely don't want to do this. Almost every AST node contains `Span`, so this increases the size of every AST type by 16 bytes. And it's redundant information - `range[0]` is the same as `start`, `range[1]` is the same as `end`.

`range` shouldn't exist at all in Rust AST types, but only be added to JSON during serialization for JS side.

I assume you've done this to work around weirdness with codegen etc?

crates/oxc_parser/src/lib.rs
```diff
@@ -221,6 +221,12 @@ pub struct ParseOptions {
     ///
     /// [`V8IntrinsicExpression`]: oxc_ast::ast::V8IntrinsicExpression
     pub allow_v8_intrinsics: bool,
+
+    /// Controls whether the `range` information is included to each node.
+    /// The `range` property is a [number, number] which indicates the start/end index of the node in the file contents.
+    ///
+    /// Default: `false`
+    pub range: bool,
```
overlookmotel: Similar to comment above, there shouldn't be a `range` option for Rust parser. We want it to be an option on JS side (`oxc-parser` NPM package), but that option would only affect the serializer, not the underlying parser.

overlookmotel (commented):
Please see comments below. This is getting complicated!

As I mention below, adding `range` field to `Span` is not the way to go. Though I completely understand why you may be struggling with the codegen.

Can I ask: Could you please:

1. Remove `range` field from `Span`?
2. Remove any other extraneous changes? (e.g. I think `RangeArray` can go if there's no `range` field in `Span`, I'm not sure if `pub mod serialize;` is necessary, `Cargo.toml` change probably not necessary etc).

Don't worry if tests don't pass, it won't compile, or the TS types don't include `range` etc. It'll be much easier for me to review without so much noise in the diff, and I can help with whatever problems remain.

Sorry if this is becoming more of a gruelling task than you'd expected!

napi/parser/src/types.rs
overlookmotel: Not sure how we're going to handle non-AST types like this. Maybe module record doesn't need `range` fields anyway? It's Oxc-specific, so we don't have to conform to any other standard here.

I suggest get rid of this for now, and we'll figure it out later.

crates/oxc_span/src/span.rs
bacarybruno: Yes I did it to have it in oxc-types. Is there another way to do it?

crates/oxc_parser/src/lib.rs
bacarybruno: It's clearer, thanks. That will simplify the code a lot. I'll do the update 🙏

crates/oxc_span/src/span.rs
overlookmotel: Yes, we can add it in the codegen for the TS type defs (`ast_tools/src/generators/typescript.rs`). That should be fairly simple.

crates/oxc_estree/src/serialize/config.rs
```diff
@@ -4,8 +4,16 @@ pub trait Config {
     const INCLUDE_TS_FIELDS: bool;
     /// `true` if should record paths to `Literal` nodes that need fixing on JS side
     const FIXES: bool;
+    /// `true` if `range` field should be emitted
+    const RANGES: bool;
```
overlookmotel: It'd be preferable if we didn't have this const. Instead `ConfigJS`, `ConfigTS` etc can have a field `ranges: bool`:

```rs
struct ConfigTS {
    ranges: bool,
}

impl Config for ConfigTS {
    #[inline(always)]
    fn new(ranges: bool) -> Self {
        Self { ranges }
    }

    #[inline(always)]
    fn ranges(&self) -> bool {
        self.ranges
    }
    // ...
}

// Same for the other 3 `Config` types
```

Then we only need the 4 x `Config` impls that we already have, rather than 8.

I'm intending (later on) to also replace `const INCLUDE_TS_FIELDS` with a runtime option so we can reduce it to 2 x `Config` impls. So don't want to go in the other direction!

I'm not sure if you're familiar with the difference. It's somewhat subtle. The trade off is: The const approach may be very marginally faster, but it bloats the size of the binary.

napi/parser/test/parse.test.ts
```diff
@@ -682,6 +682,26 @@ describe('parse', () => {
...
+      const ret = parseSync('test.js', '(x)', { range: false });
+      expect(ret.program.body[0].start).toBe(0);
+      expect(ret.program.body[0].range).toBeUndefined();
+    });
+
+    it('should include range by default', () => {
+      const ret = parseSync('test.js', '(x)', { range: true });
+      expect(ret.program.body[0].start).toBe(0);
+      expect(ret.program.body[0].range).toEqual([0, 3]);
+    });
```
overlookmotel: This test should be for what happens when no `range` option is provided. So `parseSync` should be called with an empty options object.

In my opinion, the option should default to `false`.

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -168,10 +168,14 @@ impl ESTree for ProgramConverter<'_, '_> {
...
             if S::INCLUDE_TS_FIELDS { get_ts_start_span(program) } else { program.span.start };
 
+        let range = serializer.range();
         let mut state = serializer.serialize_struct();
         state.serialize_field("type", &JsonSafeString("Program"));
         state.serialize_field("start", &span_start);
         state.serialize_field("end", &program.span.end);
+        if range {
+            state.serialize_field("range", &[span_start, program.span.end]);
+        }
```
overlookmotel: If `range` is last field (which is what you have in the codegen), it should be last here too, for consistency.

overlookmotel:
I know this is still a work in progress, so apologies for making comments prematurely. I just wanted to give feedback while I had the chance, as I'm going to be away from keyboard over the weekend.

napi/parser/test/parse.test.ts
bacarybruno: Yes, fixed locally. I'll include it in the next round of commits. 👍
By the way, thanks for taking the time to share your feedback on this PR. It's really valuable and appreciated.
I know it's taking up your time for a task you could have done more efficiently.

crates/oxc_ast/src/serialize/mod.rs
bacarybruno: Updated ✅

napi/parser/test/parse.test.ts
bacarybruno: Updated ✅

crates/oxc_estree/src/serialize/config.rs
bacarybruno: Updated ✅

crates/oxc_parser/src/lib.rs
bacarybruno: Updated ✅

bacarybruno:
@overlookmotel I think I've gotten closer to what you had in mind. I'll continue when you get back from the weekend. Have a good weekend and thanks again for your help 🙏
I'k keeping the PR in draft in the meantime

overlookmotel (approved):
I hope this doesn't come across as rude, but I'm going away tomorrow for a week, and didn't want to leave your PR hanging any longer, so I've pushed some commits to finish it off. Going to merge it now.

We have some follow-up work to complete after this is merged. I'll make a comment on #10307 in case you want to continue on with this work (please do!)

napi/parser/test/parse.test.ts
overlookmotel: > Yes, fixed locally. I'll include it in the next round of commits. 👍 By the way, thanks for taking the time to share your feedback on this PR. It's really valuable and appreciated. I know it's taking up your time for a task you could have done more efficiently.

Not at all! Really appreciate your work on this. And as I said above, as you've mentioned that you're a Rust "newb", it's really impressive you've managed to get your head around all this. Thank you.

PR #11755, opened 2025-06-16
perf(parser): optimize code around parsing delimited list and object

ulrichstark:
parse_object_expression:
- remove unnecessary `bump(Kind::Comma)`

parse_delimited_list:
- avoid `first` variable and related check by unrolling first iteration of loop
- avoid `trailing_separator` variable by restructuring code
- avoid mostly wasted call to `start_span()` (only possible if `separator` is always one character)

parse_object_literal_element
- avoid redundant kind check when parsing getter and setter


camchenry (approved):

Boshen:
Nice! This is an extremely hot path.

PR #11766, opened 2025-06-17
refactor(ast): add `AstKind` for `ComputedMemberExpression`

camchenry:
- part of https://github.com/oxc-project/oxc/issues/11490


camchenry:
* **#11767** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11767?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#11888](https://github.com/oxc-project/oxc/pull/11888) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11888?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#11889](https://github.com/oxc-project/oxc/pull/11889) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11889?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#11766** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11766).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_formatter/src/parentheses/expression.rs
```diff
@@ -531,6 +536,9 @@ fn update_or_lower_expression_needs_parens(span: Span, parent: &AstNodes<'_>) ->
     if let AstNodes::MemberExpression(member_expr) = parent {
         return member_expr.object().get_inner_expression().span() == span;
     }
+    if let AstNodes::ComputedMemberExpression(computed_member_expr) = parent {
+        return computed_member_expr.object.without_parentheses().span() == span;
```
camc314: is this right? L 537 uses `get_inner_expression` which skips non null assertion/`as`/`satifies` ect

crates/oxc_ast/src/ast_impl/js.rs
```diff
@@ -668,6 +668,22 @@ impl<'a> ComputedMemberExpression<'a> {
...
+    /// Returns the static property name of this member expression, if it has one, along with the source code [`Span`],
+    /// or `None` otherwise.
+    /// If you don't need the [`Span`], use [`ComputedMemberExpression::static_property_name`] instead.
+    pub fn static_property_info(&self) -> Option<(Span, &'a str)> {
+        match &self.expression {
+            Expression::StringLiteral(lit) => Some((lit.span, lit.value.as_str())),
+            Expression::TemplateLiteral(lit)
+                if lit.expressions.is_empty() && lit.quasis.len() == 1 =>
+            {
+                Some((lit.span, lit.quasis[0].value.raw.as_str()))
```
overlookmotel: Please amend to use `cooked`, in line with #11829. This wasn't your mistake, but I just noticed the bug while reviewing this.

crates/oxc_formatter/src/parentheses/expression.rs
overlookmotel: @Dunqing ?

camchenry: I think this was my mistake. This should just be the same as the line above, but with the new `AstKind` variant.

overlookmotel:
I've started working through this. Immediately got sidetracked by `static_property_info` and `TemplateElement`s (#11879), but now onwards...

Have taken liberty of rebasing this and #11767 on latest main to remove merge conflicts.

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
```diff
@@ -181,24 +181,19 @@ fn get_property_assignment<'a>(
...
-                        .is_some_and(|expression| expression.content_eq(node_expr))
-                    {
-                        return None;
-                    }
-                    return None;
-                }
-            }
-            _ => {}
+        if let AstKind::AssignmentExpression(_) = parent.kind() {
+            return Some(parent);
```
overlookmotel, line 195: I'm unsure about this change. But I suspect the original implementation may be buggy already.

I assume the purpose of the loop walking down ancestry chain is to find patterns like this:

```js
// Make `array.push()` a no-op
[ Array.prototype.push ] = [ () => {} ];
```

But if that *is* the purpose:

1. There are plenty of other nodes beyond `ComputedMemberExpression` which should also break the loop. e.g. `CallExpression` in `foo(Array.prototype.push).x = 1`.
2. It's not checking if the `MemberExpression` is on left or right of the `=`. e.g. does `x = Array.prototype.push` incorrectly trigger a warning?

But maybe I've completely misunderstood. Regardless, I don't think we should block this PR on it. But I think worth investigating afterwards.

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
```diff
@@ -215,51 +210,42 @@ fn get_prototype_property_accessed<'a>(
...
 
-    if is_computed_member_expression_matching(grandparent_node, prop_access_expr) {
-        prototype_node = Some(grandparent_node);
-    }
+            if let AstKind::ChainExpression(_) = grandparent_node.kind() {
+                prototype_node = Some(grandparent_node);
+                if let Some(grandparent_parent) = ctx.nodes().parent_node(grandparent_node.id()) {
+                    prototype_node = Some(grandparent_parent);
+                }
+            }
```
overlookmotel, line 226: Isn't the parent of a `ComputedMemberExpression` always a `MemberExpression`?

So `if let AstKind::ChainExpression(_) = grandparent_node.kind()` will never match.

#11767 will remove `AstKind::MemberExpression`, which will fix it again. But it's worrying that this didn't cause any tests to fail. I suggest adding a test which fails due to this problem, fix it in this PR, and then we can make sure that test still passes after #11767.

---

Also, I think the code related to `ChainExpression` which is in both match arms could be de-duplicated by moving it to after the `match` (or maybe not worthwhile doing that if #11767 simplifies it again).

crates/oxc_linter/src/rules/eslint/prefer_promise_reject_errors.rs
```diff
@@ -184,12 +181,11 @@ fn check_reject_in_function(
...
-            continue;
-        };
-
-        if let AstKind::CallExpression(call_expr) = node.kind() {
-            check_reject_call(call_expr, ctx, allow_empty_reject);
+        for parent in ctx.nodes().ancestor_kinds(parent.id()) {
+            if let AstKind::CallExpression(call_expr) = parent {
+                check_reject_call(call_expr, ctx, allow_empty_reject);
+                break;
+            }
```
overlookmotel, line 188: Doesn't this change the behavior? Previously would only call `check_reject_call` for a `CallExpression` which is the direct parent of the `MemberExpression`, now it will call `check_reject_call` for *any* ancestor of the `MemberExpression`.

I don't know this rule well enough to judge whether that's a good change or not, but as this PR is intended to be pure refactor, it'd be preferable to move behavioral changes into a separate PR, and add tests.

crates/oxc_linter/src/rules/jest/no_alias_methods.rs
```diff
@@ -106,43 +106,39 @@ impl Rule for NoAliasMethods {
...
+        };
+        let Some(jest_fn_call) = parse_expect_jest_fn_call(call_expr, jest_node, ctx) else {
+            return;
+        };
+        let Some(matcher) = jest_fn_call.matcher() else {
+            return;
+        };
+        let Some(alias) = matcher.name() else {
+            return;
+        };
```
overlookmotel, line 121: The changes to this file are unrelated refactoring and don't change anything to do with `MemberExpression`, right?

(if that's correct, just for future reference, it would have been preferable to do this in a separate PR - the diff of this PR is hardly minimal already!)

crates/oxc_linter/src/rules/oxc/bad_array_method_on_arguments.rs
```diff
@@ -60,57 +57,50 @@ impl Rule for BadArrayMethodOnArguments {
...
+        let member_expr = parent.kind();
+        let Some(grandparent) = ctx.nodes().parent_node(parent.id()) else {
             return;
         };
+        if matches!(member_expr, AstKind::ComputedMemberExpression(_)) {
+            let great_grandparent = ctx.nodes().parent_kind(grandparent.id());
+            let Some(AstKind::CallExpression(_)) = great_grandparent else {
+                return;
+            };
+        } else if !matches!(member_expr, AstKind::ComputedMemberExpression(_)) {
```
overlookmotel, line 78: I think the 2nd check is redundant. Could just be `} else {`.

crates/oxc_linter/src/rules/oxc/bad_array_method_on_arguments.rs
```diff
@@ -60,57 +57,50 @@ impl Rule for BadArrayMethodOnArguments {
...
-                    _ => {}
+            AstKind::ComputedMemberExpression(expr) => {
+                let Some(name) = expr.static_property_name() else {
+                    return;
+                };
+                if ARRAY_METHODS.binary_search(&name.as_str()).is_ok() {
+                    ctx.diagnostic(bad_array_method_on_arguments_diagnostic(
+                        name.as_str(),
+                        expr.span,
+                    ));
```
overlookmotel, line 100: I think this block can be deduplicated further:

```rs
let (name, span) = match member_expr {
    AstKind::MemberExpression(member_expr) => {
        let MemberExpression::StaticMemberExpression(expr) = member_expr else {
            return;
        };
        (expr.property.name, expr.span)
    }
    AstKind::ComputedMemberExpression(expr) => {
        let Some(name) = expr.static_property_name() else {
            return;
        };
        (name, expr.span)
    }
    _ => unreachable!(),
};

if ARRAY_METHODS.binary_search(&name.as_str()).is_ok() {
    ctx.diagnostic(bad_array_method_on_arguments_diagnostic(&name, span));
}
```

It could probably be combined with previous `if` / `else` too to reduce branching.

Note: `unreachable!()` will likely cause compiler to backtrack and prove to itself that `member_expr` can only be `AstKind::MemberExpression` or `AstKind::ComputedMemberExpression`, so it'll remove the unreachable branch.

crates/oxc_linter/src/rules/oxc/uninvoked_array_callback.rs
```diff
@@ -67,28 +67,51 @@ impl Rule for UninvokedArrayCallback {
...
-            MemberExpression::PrivateFieldExpression(expr) => expr.field.span,
-        };
-        ctx.diagnostic(uninvoked_array_callback_diagnostic(property_span, new_expr.span));
+                ctx.diagnostic(uninvoked_array_callback_diagnostic(
+                    computed_member_expr.expression.span(),
+                    new_expr.span,
+                ));
+            }
+            _ => {}
+        }
```
overlookmotel, line 114: Some of the repeated code could be dedupllicated by moving it to below the match.

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
overlookmotel: I changed my mind. Going to merge this now, but we should check after both PRs are merged that this didn't break anything.

overlookmotel (approved):
I've made some comments below. Many relate to refactoring to deduplicate code which appears in both the branches for `MemberExpression` and `ComputedMemberExpression`.

But 3 of the comments relate to where I think the changes in this PR may alter behavior, and might have introduced bugs. Our test coverage in linter tends to be a little sparse, compared to the wide variety of wacky patterns which are legal JavaScript! So it's hard to rely on the tests.

#11767 is stacked on top of this and presumably touches much of the same code, so making changes in this PR would be a pain. And some of my comments may be rendered obsolete by #11767 anyway.

So I'm going to merge this now. But @camchenry could you possibly try to address any of my comments which are valid after these 2 PRs are merged?

overlookmotel:
### Merge activity

* **Jun 24, 7:01 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jun 24, 7:03 PM UTC**: `overlookmotel` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jun 24, 7:11 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

overlookmotel:
By the way, I hope you don't take my litany of comments as criticism. These `AstKind` changes are surprisingly subtle and brain-bending (well I find them so anyway). I think it's nigh on impossible to get through the big ones like `MemberExpression` bug-free without at least 2 pairs of eyes on it. Maybe even 2 pairs is not enough...

overlookmotel:
Side note: `ctx.nodes().parent_node(node_id)` returning an `Option` is a massive pain. Every node except `Program` has a parent, and generally you know the node you're getting the parent of is not `Program`, so all the `let Some(parent) = ...` is just boilerplate noise in the code, and introduces unnecessary branches all over the place.

Should we make `Program` have itself as its parent, so `parent_node` can return `AstNode` not `Option<AstNode>`?

crates/oxc_linter/src/rules/eslint/no_extend_native.rs
camchenry: No, you're absolutely right, and ESLint doesn't catch the array assignment target case as-is either. I tested it in Firefox and this definitely does work to assign to the prototype though. I've implemented a fix for the assignment expression so that it ensures it is on the **left** side. I've also changed it so that _any_ AST kind other than a few allowed ones (computed members, static members, etc.) will stop iteration. That way, it's more of an "allow list" rather than a "block list".

https://github.com/oxc-project/oxc/pull/11888

Screenshot 2025-06-24 at 11 56 34 PM

crates/oxc_linter/src/rules/eslint/prefer_promise_reject_errors.rs
camchenry: Reverted to old behavior and added a test case in https://github.com/oxc-project/oxc/pull/11889

crates/oxc_linter/src/rules/jest/no_alias_methods.rs
camchenry: Woops, sorry about that! I was deep into debugging the jest fn parsing logic and I had rearranged/refacotred things and reverted all of it, but apparently I forgot to revert this.

crates/oxc_linter/src/rules/oxc/bad_array_method_on_arguments.rs
camchenry: This should be fixed in #11767, as we don't need this code anymore. I think it was some clippy lint rule that made me write it initially.

crates/oxc_linter/src/rules/oxc/bad_array_method_on_arguments.rs
camchenry: This was simplified in #11767.

crates/oxc_linter/src/rules/oxc/uninvoked_array_callback.rs
camchenry: This was simplified in #11767.

PR #11795, opened 2025-06-18
refactor(transformer): replace TransformCtx with TransformState 

### LINKED ISSUE #11742: Move `TransformCtx` into `TraverseCtx`
## The problem

Currently we have `TransformCtx` which contains shared state between all transforms. It's stored in various transforms' `self` as a reference `&'ctx TransformCtx<'a>`.

Because transforms only have an immutable reference to `TransformCtx`, all mutable state has to be wrapped in a `RefCell`. e.g. the shared stack `VarDeclarations` introduced in oxc-project/oxc#6170. `RefCell` is somewhat costly.

## Proposed solution

Instead, move `TransformCtx` to live within `TraverseCtx`. `TraverseCtx` is a singleton and is passed to every visitor method in all transforms as a mutable `&mut TraverseCtx<'a>`. So there only needs to be one instance of `TransformCtx`, and we can get rid of the `'ctx` lifetime too.

But we want to avoid leaking the transformer's internals into `Traverse`, as `Traverse` is meant to be a tool which can be used for other purposes too (e.g. minifier).

## How?

I propose that we parameterize `TraverseCtx` so it can hold arbitrary state:

```rs
pub trait Traverse<'a, State> {
    fn enter_statements(&mut self, stmts: &mut Vec<'a, Statement<'a>>, ctx: &mut TraverseCtx<'a, State>) {}
    fn exit_statements(&mut self, stmts: &mut Vec<'a, Statement<'a>>, ctx: &mut TraverseCtx<'a, State>) {}
    // etc...
}

pub struct TraverseCtx<'a, State> {
    pub ancestry: TraverseAncestry<'a>,
    pub scoping: TraverseScoping,
    pub ast: AstBuilder<'a>,
    pub state: State,
}

// The new name for `TransformCtx`
struct TransformState<'a> {
    errors: Vec<OxcDiagnostic>,
    pub trivias: Trivias,
    var_declarations: SparseStack<Vec<'a, VariableDeclarator<'a>>>,
    // ... etc ...
}

type TransCtx<'a> = TraverseCtx<'a, TransformState<'a>>;

struct Common;

impl<'a> Traverse<'a, TransformState<'a>> for Common {
    fn enter_statements(&mut self, _stmts: &mut Vec<'a, Statement<'a>>, ctx: &mut TransCtx<'a>) {
        ctx.state.var_declarations.push(None);
    }

    fn exit_statements(&mut self, stmts: &mut Vec<'a, Statement<'a>>, ctx: &mut TransCtx<'a>) {
        if let Some(declarations) = ctx.state.var_declarations.pop() {
            // Insert `var` declaration statement
        }
    }
}
```

## Preconditions

We currently store `AstBuilder` in both `TransformCtx` and `TraverseCtx`. We will need switch all uses of the copy in `TransformCtx` over to `TraverseCtx` i.e. `self.ctx.ast.null_literal(SPAN)` -> `ctx.ast.null_literal(SPAN)`.

We can then remove `AstBuilder` from `TransformCtx`.

We will need to do this anyway once we have `NodeId`s on all AST nodes, and need `AstBuilder` to be stateful to maintain a counter for next Node ID - which requires there to be only 1 copy of `AstBuilder` in the transformer.

Boshen:
closes #11742

### DIFF
--- a/crates/oxc_transformer/src/common/arrow_function_converter.rs
+++ b/crates/oxc_transformer/src/common/arrow_function_converter.rs
@@ -103,8 +103,8 @@ use oxc_syntax::{
 };
 use oxc_traverse::{Ancestor, BoundIdentifier, Traverse};
 
+use crate::context::{TransformState, TraverseCtx};
 use crate::{EnvOptions, utils::ast_builder::wrap_expression_in_arrow_function_iife};
-use crate::{context::TraverseCtx, state::TransformState};
 
 type FxIndexMap<K, V> = IndexMap<K, V, FxBuildHasher>;
 
--- a/crates/oxc_transformer/src/common/computed_key.rs
+++ b/crates/oxc_transformer/src/common/computed_key.rs
@@ -4,11 +4,11 @@ use oxc_ast::ast::Expression;
 use oxc_semantic::SymbolFlags;
 
 use crate::{
-    context::{TransformCtx, TraverseCtx},
+    context::{TransformState, TraverseCtx},
     utils::ast_builder::create_assignment,
 };
 
-impl<'a> TransformCtx<'a> {
+impl<'a> TransformState<'a> {
     /// Check if temp var is required for `key`.
     ///
     /// `this` does not have side effects, but in this context, it needs a temp var anyway, because `this`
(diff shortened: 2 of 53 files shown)


Boshen:
```
error[E0502]: cannot borrow `*ctx` as mutable because it is also borrowed as immutable
   --> crates/oxc_transformer/src/proposals/explicit_resource_management.rs:762:22
    |
762 |         let callee = ctx.state.helper_load(Helper::UsingCtx, ctx);
    |                      ---------^-----------^^^^^^^^^^^^^^^^^^^^^^^
    |                      |         |
    |                      |         immutable borrow later used by call
    |                      mutable borrow occurs here
    |                      immutable borrow occurs here
```

PR #11796, opened 2025-06-18
feat(minifier): remove unused variable declaration

### LINKED ISSUE #10033: minifier: remove unused variable declaration

Boshen:
closes #10033


Boshen:
I'll resolve the performance regression when I finalize the algorithm.

PR #11843, opened 2025-06-21
refactor(ast): Remove AstKind from `TSTypeName` node

therewillbecode:
This PR is part of the ongoing work in #11490 

Removes AstKind from `TSTypeName` nodes by removing the entry from the list of enum exceptions in ast_tools.


crates/oxc_language_server/src/snapshots/fixtures_linter_issue_9958@issue.ts.snap
```diff
@@ -15,6 +15,32 @@ tags: None
 fixed: None
 
 
+code: "eslint(no-unused-vars)"
```
therewillbecode, line 18: Can someone advise on whether this snapshot change is okay?

overlookmotel:
Out of interest, did you close this because of the perf drop, or because it didn't work?

Issue #11582, opened 2025-06-10
parser: Improve TS error story


leaysgur:
### Background

<details>

> Boshen:
> I want to improve ts error story, but I have trouble differentiating ts syntax errors vs type errors in https://github.com/oxc-project/oxc/blob/main/tasks/coverage/snapshots/parser_typescript.snap
>
> When it says:
> ```
> Expect Syntax Error: tasks/coverage/typescript/tests/cases/compiler/ClassDeclaration24.ts
> ```
>
> the error file tsc produces may contain type errors, which we should exclude.
>
> I thought I could bisect https://github.com/microsoft/TypeScript/blob/main/src/compiler/diagnosticMessages.json and split the error codes, but the error codes seem pretty random.
>
> Turning off `noCheck` doesn't really help, because it just turns off checker.ts, and checker.ts may produce syntax errors as well.
> e.g. `class any {}` (from ClassDeclaration24.ts)
> https://discord.com/channels/1079625926024900739/1080712024683708466/1380227952322674718

</details>

### The problem

In `parser_typescript.snap`, we sometimes get "Expect Syntax Error" lines, but it is not clear how to fix them.

Here is how the message appears:

- A file `xxx.errors.txt` exists inside `tests/baselines/reference` in the TypeScript tests
- && Our parser does NOT report error  
  - NOTE: `oxc_semantic` is disabled in the `parser_typescript.snap` run

So the message means: "The parser should have reported an syntactic error, but it did not."

However, a file named `xxx.errors.txt` can exist for several reasons:

1. The code is not valid JS or TS syntax
   - e.g. `if {`, `async class X {}`, `const a;`, `class any {}`, `class X { private private x = 1 }`
2. The code is syntactically fine, but has only type errors

Case 2 is a problem for us. OXC should not treat these as parser errors, so the "Expect Syntax Error" message is confusing.

We would like to filter such cases out. But how?

TSC itself reports 2 main kinds of errors:

1. `tsProgram.getSyntacticDiagnostics()`
2. `tsProgram.getSemanticDiagnostics()`

The first case represents a critical syntax error, and OXC also wants to treat this as a parser error.

The second case is somewhat tricky, while TSC claims it's a semantic error, but we wants to treat many cases as parser errors instead.
(e.g. `class any {}`, `const a;`)

Therefore, we need to look at all errors reported by TSC and decide which ones should become parser errors in OXC.

[...]

PR #11858, opened 2025-06-23
refactor(coverage/parser_typescript): Improve `should_fail` check by extracting diagnostic codes from `.errors.txt`

leaysgur:
Part of #11582 

- 👉🏻 Check error codes in coverage tests
- Fill more error codes 
- Dump currently reviewed error codes somewhere

### DIFF
--- a/tasks/coverage/snapshots/semantic_typescript.snap
+++ b/tasks/coverage/snapshots/semantic_typescript.snap
@@ -1,8 +1,8 @@
 commit: 81c95189
 
 semantic_typescript Summary:
-AST Parsed     : 6537/6537 (100.00%)
-Positive Passed: 2848/6537 (43.57%)
+AST Parsed     : 6540/6540 (100.00%)
+Positive Passed: 2848/6540 (43.55%)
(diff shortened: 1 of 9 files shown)


tasks/coverage/snapshots/semantic_typescript.snap
```diff
@@ -20478,6 +20478,14 @@ Scope children mismatch:
 after transform: ScopeId(0): [ScopeId(1)]
 rebuilt        : ScopeId(0): []
 
+semantic Error: tasks/coverage/typescript/tests/cases/compiler/moduleAndInterfaceSharingName2.ts
```
leaysgur, line 20481: I'm not sure how to understand these diffs.

Should I apply error code filtering only to the `parser_typescript` flow?

Boshen: yes, don't worry about semantic.

leaysgur: Thanks for confirming!

Later (maybe tomorrow), I will adjust it to behave the same as the original.

PR #11874, opened 2025-06-24
feat(bench): add conditional job to run walltime benchmarks

camchenry:
- towards https://github.com/oxc-project/backlog/issues/5
- Co-authored by Copilot

This adds a sorely missing benchmarking tool: a walltime benchmark. This is strongly based on [rolldown](https://github.com/rolldown/rolldown/blob/main/.github/workflows/benchmark-rust.yml)'s implementation of the benchmark. This eliminate machine variance by running both sets of benchmarks on the same machine in the same step. I've set the sample size and warm-up times higher than default to try and ensure that we get really clean statistical data when we do run the benchmark. The benchmark is also only run conditionally via the `run-bench` label, so that we only incur the cost when it's necessary. This will provide us with much more accurate information about things like branch prediction and cache misses that CodSpeed does not give us.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -1,0 +1,106 @@
+name: Benchmark PR
+
+permissions:
+  pull-requests: write
+  contents: read
+
+on:
+  pull_request:
+    types: labeled
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
+    name: Walltime Benchmark
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    if: contains(github.event.pull_request.labels.*.name, 'run-bench')
+
+    steps:
+      - name: Checkout PR Branch
+        uses: taiki-e/checkout-action@b13d20b7cda4e2f325ef19895128f7ff735c0b3d # v1.3.1
+
+      - name: Setup Rust
+        uses: oxc-project/setup-rust@cd82e1efec7fef815e2c23d296756f31c7cdc03d # v1.0.0
+        with:
+          tools: critcmp
+          cache-key: rust-benchmark
+          save-cache: ${{ github.ref_name == 'main' }}
+
+      - name: Run Benchmarks on PR Branch
+        shell: bash
+        run: |
+          echo "Running benchmarks on PR branch..."
+          cargo bench -p oxc_benchmark --bench lexer -- --save-baseline pr --sample-size 250 --warm-up-time 5 --measurement-time 15
+          cargo bench -p oxc_benchmark --bench parser -- --save-baseline pr --sample-size 250 --warm-up-time 5 --measurement-time 15
+
+      - name: Checkout Target Branch
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          clean: false
+          ref: ${{ github.event.pull_request.base.ref }}
+          persist-credentials: false
+
+      - name: Run Benchmarks on Target Branch
+        shell: bash
+        run: |
+          echo "Running benchmarks on target branch..."
+          cargo bench -p oxc_benchmark --bench lexer -- --save-baseline target --sample-size 250 --warm-up-time 5 --measurement-time 15
+          cargo bench -p oxc_benchmark --bench parser -- --save-baseline target --sample-size 250 --warm-up-time 5 --measurement-time 15
+
+      - name: Verify Baselines
+        shell: bash
+        run: |
+          echo "Verifying benchmark baselines were created..."
+          critcmp --baselines
+          if ! critcmp --baselines | grep -q "pr"; then
+            echo "Error: PR baseline not found"
+            exit 1
+          fi
+          if ! critcmp --baselines | grep -q "target"; then
+            echo "Error: Target baseline not found"
+            exit 1
+          fi
+          echo "✅ Both baselines found successfully"
+
+      - name: Compare Benchmark Results
+        id: bench_comparison
+        shell: bash
+        run: |
+          echo "## 🚀 Benchmark Results" > output
+          echo "" >> output
+
+          echo "\`\`\`" >> output
+          if critcmp target pr >> output 2>/dev/null; then
+            echo "Benchmark comparison completed"
+          else
+            echo "No significant benchmark differences found." >> output
+          fi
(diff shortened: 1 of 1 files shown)


camchenry:
* **#11874** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11874?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#11886](https://github.com/oxc-project/oxc/pull/11886) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11886?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#11912](https://github.com/oxc-project/oxc/pull/11912) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11912?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/11874?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=11874).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Hooray! It'll be great to have wallclock benchmarks. I'm afraid I don't have time today to review in full, and I'm away until next week, but a couple of quick thoughts:

* I'm not sure a set sample size of 200 is appropriate for all benchmarks. `parser/RadixUIAdoptionSection.jsx` bench is so fast that 200 iterations will probably not be enough to average out wide variance.
* Have you audited `peter-evans/create-or-update-comment` action? Just asking because we give it the keys to the kingdom `secrets.GITHUB_TOKEN`.

camc314:
> * Have you audited `peter-evans/create-or-update-comment` action? Just asking because we give it the keys to the kingdom `secrets.GITHUB_TOKEN`.

looks like rolldown does something similar using the peter evans action

https://github.com/rolldown/rolldown/blob/a7332057c88f054c7fe93b1123b4b33f17a501e5/.github/workflows/benchmark-rust.yml#L0-L1

camchenry:
> I'm not sure a set sample size of 200 is appropriate for all benchmarks. parser/RadixUIAdoptionSection.jsx bench is so fast that 200 iterations will probably not be enough to average out wide variance.

Based on the [description of the analysis process](https://bheisler.github.io/criterion.rs/book/analysis.html) for Criterion, I think each "sample" actually consists of many iterations of the function. Looking at the latest run, you can see that it actually ran the parser ~442,000 times to get the statistics:

```
Benchmarking parser/RadixUIAdoptionSection.jsx
Benchmarking parser/RadixUIAdoptionSection.jsx: Warming up for 5.0000 s
Benchmarking parser/RadixUIAdoptionSection.jsx: Collecting 200 samples in estimated 5.0313 s (442k iterations)
Benchmarking parser/RadixUIAdoptionSection.jsx: Analyzing
parser/RadixUIAdoptionSection.jsx
                        time:   [11.351 µs 11.357 µs 11.363 µs]
Found 9 outliers among 200 measurements (4.50%)
  1 (0.50%) low mild
  3 (1.50%) high mild
  5 (2.50%) high severe
```

I think the number of iterations per sample is computed dynamically based on some estimates and statistical calculations to make the results significant. But suppose that each sample runs the

For example, if we suppose that Criterion runs `parser/RadixUIAdoptionSection.jsx` 1,000 times in ~10ms (~10 microseconds per run) and does that for each sample:

1. On the first sample, it will run the parser 1,000 times and then compute the average time and use that as the time for that sample.
2. On the second sample, it will run the parser 2,000 times and repeat.
3. On the third sample, it will run the parser 3,000 times and so on.
4. On the Nth sample (N = 200), it will run the parser 200,000 times and use the average time as the 200th sample time.
5. Then, it finally uses those 200 sample times as the data points for doing a [t-test](https://en.wikipedia.org/wiki/Student%27s_t-test) to determine statistical significance.

We have some additional things we can tweak to reduce variance, but with tradeoffs:

- **Increase sample size / measurement time**: https://docs.rs/criterion/latest/criterion/struct.BenchmarkGroup.html#method.measurement_time. This increases the number of samples that are collected, which makes the confidence interval for the true mean smaller, which makes it easier to determine statistical significance for small changes in performance. Of course, this comes at the cost of time. We could potentially have two variations of the CI workflow: a normal one with a moderate sample size, and a high accuracy run where we dramatically bump up the sample size to ensure we are getting an accurate result for changes which we think are real but small (and therefore hard to measure).
- **Change the confidence level**: https://docs.rs/criterion/latest/criterion/struct.BenchmarkGroup.html#method.confidence_level. By default this is set to 95%, which means that the true mean will be captured in the indicated intervals 95% of the time (5% of the time we will report a difference when there is none). We could increase this to 99% or 99.9%, but it will make the reported time intervals wider. To make the intervals smaller, we have to collect more samples.

The given intervals currently for these small benchmarks are pretty tight already though: we can say with 95% confidence that the true time it takes to run the parser on `RadixUIAdoptionSection.jsx` is 11.357 microseconds, plus or minus 6 nanoseconds!

camchenry:
> * Have you audited `peter-evans/create-or-update-comment` action? Just asking because we give it the keys to the kingdom `secrets.GITHUB_TOKEN`.

I have a high trust level for this action because the author is a GitHub employee and so technically a co-worker 😄 But also I've read over the source code and it is pretty short, trivial usage of the GitHub API via Octokit. Pinning the version here should ensure it doesn't change out without us knowing. The risk level is also somewhat reduced by the fact that we're not giving it total access to everything, it would at least be limited to only editing PR metadata/comments and the ability to create/delete milestones/labels: https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests.

.github/workflows/benchmark-pr.yml
```diff
@@ -0,0 +1,111 @@
+name: Benchmark PR
+
+permissions:
+  pull-requests: write
+  contents: read
+
+on:
+  pull_request:
+    types: [labeled, synchronize]
```
Boshen: `synchronize` is probably not needed, that's a job per PR at least.

camchenry:
Now that we've run this more times, I'm not sure this will be able to pick up subtle variations in performance. It could probably detect a 20-30% change fairly easily, but I feel like 2% performance wins will be hard to measure. I'm not sure it's worth the effort of maintaining right now.

PR #11875, opened 2025-06-24
refactor(coverage/typescript): Check error codes only for parser suites

leaysgur:
Follow up #11858 

Only parser-related tests 

- `parser_typescript.snap`
- `estree_typescript.snap`

will check for error codes.

Or..., should I just skip it only for `semantic_typescript.snap`?


Boshen:
> should I just skip it only for semantic_typescript.snap

This is better. Let's avoid the weird conditions in this PR.

leaysgur:
#11902

PR #11887, opened 2025-06-24
feat(linter): Add id-length rule from eslint

nrayburn-tech:


crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// function foobar({ prop }) { }
+    /// function foobaz({ a: prop }) { }
+    /// const { a: property } = {};
+    /// ({ prop: obj.longName } = {});
+    /// const data = { "x": 1 };  // excused because of quotes
+    /// data["y"] = 3;  // excused because of calculated property access
+    /// ```
+    IdLength,
+    eslint,
+    nursery, // TODO: change category to `correctness`, `suspicious`, `pedantic`, `perf`, `restriction`, or `style`
```
nrayburn-tech: Not really sure where this best fits, feel free to update it.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
+        match node.kind() {
+            AstKind::BindingIdentifier(ident) => {
+                let ident_name = ident.name.to_string();
+
+                if let Some(parent_node) = ctx.nodes().parent_node(node.id()) {
+                    match parent_node.kind() {
+                        AstKind::ImportSpecifier(import_specifier) => {
+                            // TODO: Is there a better way to do this check?
+                            if import_specifier.imported.name().eq(&import_specifier.local.name) {
```
nrayburn-tech: Not sure if this is the best way. I would have to do some testing to see, but I think the problem was that I needed to be able to differentiate between `import { x } from 'xyz';` and `import { x as y } from 'xyz';` or something like that. The former shouldn't error, the latter _might_ error.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+                                return;
+                            }
+                            // TODO: Is there a better way to do this check?
+                            if !import_specifier.local.content_eq(ident) {
```
nrayburn-tech: Not sure if this is the best way.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+                                return;
+                            }
+                        }
+                        AstKind::ObjectPattern(object_pattern) => {
+                            // TODO: Is there a better way to do this check?
+                            let binding_property_option =
+                                object_pattern.properties.iter().find(|x| x.span.eq(&ident.span));
```
nrayburn-tech: Not sure if this is the best way.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+                                return;
+                            }
+
+                            if let Some(parent_parent_node) =
+                                ctx.nodes().parent_node(parent_node.id())
+                            {
+                                match parent_parent_node.kind() {
+                                    AstKind::ObjectPattern(object_pattern) => {
+                                        // TODO: Is there a better way to do this check?
+                                        let binding_property_option = object_pattern
```
nrayburn-tech: Not sure if there is a better way.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+        ), // {				"ecmaVersion": 2025,			},
+        (
+            "export * from 'foo.json' with { type: 'json' }",
+            Some(serde_json::json!([{ "min": 1, "max": 3, "properties": "always" }])),
+        ), // {				"ecmaVersion": 2025,			},
+        (
+            "export { default } from 'foo.json' with { type: 'json' }",
+            Some(serde_json::json!([{ "min": 1, "max": 3, "properties": "always" }])),
+        ), // {				"ecmaVersion": 2025,			},
+                                                                                      // TODO:
```
nrayburn-tech, line 542: I wasn't sure how to fix these without hardcoding the `with` and `type` properties. They can probably just be fixed later, I expect they're not frequent errors.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+    Never,
+}
+
+impl PropertyKind {
+    pub fn from(raw: &str) -> Self {
+        if raw == "never" { PropertyKind::Never } else { PropertyKind::default() }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
```
nrayburn-tech: I tried to add `JsonSchema` to this, so that the config could be added to declare_oxc_lint!, but `Regex` doesn't have an implement `JsonSchema` and I wasn't sure the best way to do that. Didn't feel appropriate to do that here.

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,573 @@
...
+            min: object.and_then(|map| map.get("min")).and_then(Value::as_u64).unwrap_or(2),
+            properties: object
+                .and_then(|map| map.get("properties"))
+                .and_then(Value::as_str)
+                .map(PropertyKind::from)
+                .unwrap_or_default(),
+        }
+    }
+
+    fn run<'a>(&self, node: &AstNode<'a>, ctx: &LintContext<'a>) {
```
nrayburn-tech, line 382: I left this method linear so that it would be easy to read. It could probably be refactored into some smaller functions.

nrayburn-tech:
This test is failing for me locally as well. I don't understand what the problem is or how to resolve it though.

```
failures:

---- size_asserts stdout ----

thread 'size_asserts' panicked at crates/oxc_linter/src/lib.rs:63:5:
  left: 72
 right: 16


failures:
    size_asserts
```

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,593 @@
...
+    Never,
+}
+
+impl PropertyKind {
+    pub fn from(raw: &str) -> Self {
+        if raw == "never" { PropertyKind::Never } else { PropertyKind::default() }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
```
Sysix: please use `impl Default for IdLength`, currently fixing bugs where rules are getting activated without a configuration (when enabling the category):
https://github.com/oxc-project/oxc/pull/11908

crates/oxc_linter/src/rules/eslint/id_length.rs
Sysix: I do not think this is fixable. I can only think  `i` for `index`.

Sysix:
> I don't understand what the problem is or how to resolve it though.

@nrayburn-tech you need to `Box<>` your configuration because of the Vec and RegEx.
You can find an example here: https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/no_restricted_imports.rs

crates/oxc_linter/src/rules/eslint/id_length.rs
nrayburn-tech: I thought I removed the fixable part, unless I'm missing something somewhere. My comment is that I'm not sure what category this rule should fall under.

Sysix: oh my bad. I would put it into "style" :)

crates/oxc_linter/src/rules/eslint/id_length.rs
```diff
@@ -0,0 +1,629 @@
...
+                                        let binding_property_option = object_pattern
+                                            .properties
+                                            .iter()
+                                            .find(|x| x.key.content_eq(property_key));
+
+                                        if let Some(binding_property) = binding_property_option {
+                                            match &binding_property.value.kind {
+                                                BindingPatternKind::BindingIdentifier(_)
+                                                | BindingPatternKind::ObjectPattern(_) => {
+                                                    return;
```
Sysix: in line 248 there is an extra `if self.properties == PropertyKind::Never {`. 
Can both `AstKind::ObjectPattern` checks combined into one helper function?

nrayburn-tech: Created a helper function for this. Still need the `self.properties` check in the one spot, but not the other so that logic was retained.

crates/oxc_linter/src/rules/eslint/id_length.rs
nrayburn-tech: Broke the method up into 3 smaller methods based on the top level AST node kinds. Might still be able to be refactored more, but I'm not seeing anything myself.

Sysix (approved):

PR #11921, opened 2025-06-26
docs(linter): fix prefer-string-replace-all example

zettca:
`prefer-string-replace-all` has an incorrect "incorrect" example 😄 
(copy-pasted from `no-array-reduce`)


camc314 (commented):
thanks!

PR #11937, opened 2025-06-27
feat(linter): add eslint/arrow-body-style rule

huangtiandi1999:
Rule detail: https://eslint.org/docs/latest/rules/arrow-body-style


camc314 (approved):
thank you!

PR #12051, opened 2025-07-03
fix(linter): Fix inconsistent behavior in `no-duplicate-imports` rule

huangtiandi1999:
Currently, the `no-duplicate-imports` rule in Oxlint does not report an error for the following code:
```js
import b from 'foo';
import { a } from 'foo';
```


mesqueeb:
@huangtiandi1999 did you also look at the rule for `import/no-duplicates` ? this one is very similar as `eslint/no-duplicate-imports` I imagine the behaviour should be kept the same at all times.

PR #12066, opened 2025-07-04
feat(transformer): support styled components plugin

### LINKED ISSUE #11876: transformer: babel-plugin-styled-components
It would be nice to have a [plugin with styled-components](https://styled-components.com/docs/tooling) support like those for [babel](https://github.com/styled-components/babel-plugin-styled-components) or [swc](https://github.com/swc-project/plugins/tree/main/packages/styled-components). Maybe it will be possible to make a wrapper for the swc plugin, like [rsbuild](https://github.com/rspack-contrib/rsbuild-plugin-styled-components) did.

This plugin allows, for example, to display the name of the components in classes, which greatly simplifies debugging via developer tools in the browser.

Dunqing:
close: #11876 

This plugin adds comprehensive support for styled-components with server-side rendering, style minification, and enhanced debugging capabilities. The implementation is ported from the [official Babel plugin](https://github.com/styled-components/babel-plugin-styled-components) to ensure compatibility and feature parity.

## Current Limitations

⚠️ **Import Support**: This plugin currently only supports styled-components imported via ES modules (`import` statements). CommonJS imports using `require("styled-components")` are not yet supported.

### Options:
 **✅ Fully Supported:**
 - `displayName`: Adds display names for debugging
 - `fileName`: Controls filename prefixing in display names
 - `ssr`: Adds unique component IDs for server-side rendering
 - `transpileTemplateLiterals`: Converts template literals to function calls
 - `minify`: Minifies CSS content in template literals
 - `namespace`: Adds namespace prefixes to component IDs
 - `meaninglessFileNames`: Controls which filenames are considered meaningless

 **⚠️ Partially Supported:**
 - `pure`: Only supports call expressions, not tagged template expressions (bundler limitation)

 **❌ Not Yet Implemented:**
 - `cssProp`: JSX css prop transformation
 - `topLevelImportPaths`: Custom import path handling

## Testing

The test suite is adapted from the [official Babel plugin tests](https://github.com/styled-components/babel-plugin-styled-components) with modifications to integrate with our testing infrastructure. 

**Note**: Some test outputs differ from the original due to our different hashing algorithm for component IDs. This is intentional and follows the same approach used by SWC's implementation.


Dunqing:
* **#12066** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12066?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#12155](https://github.com/oxc-project/oxc/pull/12155) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12155?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#12169](https://github.com/oxc-project/oxc/pull/12169) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12169?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12066).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/plugins/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::{
+    context::{TransformCtx, TraverseCtx},
+    plugins::styled_components::StyledComponents,
+    state::TransformState,
+};
+
+pub struct Plugins<'a, 'ctx> {
+    styled_components: StyledComponents<'a, 'ctx>,
+    options: PluginsOptions,
+}
```
overlookmotel, line 18: Is it necessary to store `PluginsOptions` here as well as inside `StyledComponents`? The clone isn't so cheap.

How about either of these?

```rs
pub struct Plugins<'a, 'ctx> {
    styled_components: StyledComponents<'a, 'ctx>,
    is_styled_components_enabled: bool,
}

// In `enter_program`
if self.is_styled_components_enabled {
    self.styled_components.enter_program(node, ctx);
}
```

or:

```rs
pub struct Plugins<'a, 'ctx> {
    styled_components: Option<StyledComponents<'a, 'ctx>>,
}

// In `enter_program`
if let Some(style_components) = &mut self.styled_components {
    styled_components.enter_program(node, ctx);
}
```

(personally I prefer the latter)

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -0,0 +1,1154 @@
...
+
+    /// Transpiles styled-components tagged template literals to a smaller representation
+    /// than what Babel normally creates, helping to reduce bundle size.
+    ///
+    /// Converts `styled.div\`width: 100%;\`` to `styled.div(['width: 100%;'])`, which is
+    /// more compact than the standard Babel template literal transformation.
+    ///
+    /// Default: `true`
+    #[serde(default = "default_as_true")]
+    pub transpile_template_literals: bool,
```
overlookmotel, line 118: Should we make this option default to `false`? We don't downlevel beyond ES6, so template literals are more compact.

```js
styled.div`width: 100%;`
styled.div(['width: 100%;'])
```

overlookmotel:
I've pushed a commit to enable this plugin in benchmarks, so we can see what impact it has.

overlookmotel:
Not too bad!

bench

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -0,0 +1,1154 @@
...
+    }
+
+    fn enter_call_expression(&mut self, call: &mut CallExpression<'a>, ctx: &mut TraverseCtx<'a>) {
+        if (self.options.display_name || self.options.ssr)
+            // Only transform call expression that is not a part of a member expression
+            // or a callee of another call expression.
+            && !matches!(
+                ctx.parent(),
+                Ancestor::CallExpressionCallee(_) | Ancestor::StaticMemberExpressionObject(_)
+            )
```
overlookmotel, line 290: Should `ComputedMemberExpression` be transformed?

overlookmotel (commented):
I've tried to review this properly, but it's hard because I don't understand what any of it is doing! I know we're keen to get this merged ASAP, but afterwards could you possibly add some comments, and I'll read through it again?

Please feel free to merge in the meantime. I'll do a few follow-up PRs with suggested refactoring and perf, and my comments below could also be addressed in follow-ups.

Another thought: Are we able to disable this plugin on files which aren't JSX/TSX? Or does it need to run on JS files too? The loop through all top-level statements is probably quite costly for long files, and it'd be ideal to skip it if we can.

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -0,0 +1,1154 @@
...
+            while num != 0 {
+                result.push(BASE36_BYTES[(num % 36) as usize]);
+                num /= 36;
+            }
+
+            ctx.ast.allocator.alloc_str(
+                // SAFETY: the bytes are valid UTF-8 as they are ASCII characters.
+                unsafe { std::str::from_utf8_unchecked(&result) },
+            )
+        }
```
overlookmotel, line 616: Is there a reason for using base 36 here? Couldn't we use base 64 (A-Z, a-z, 0-9, _, $)?

Division by numbers which aren't a power of 2 is quite expensive, and it'd give a larger range of possibilities in 6 chars (less chance of collisions).

Also, building the string with `InlineString` would be be more performant than `Vec`. Like in mangler:

https://github.com/oxc-project/oxc/blob/fe843806ece40a19adca3a7c6366abfb6f8662bb/crates/oxc_mangler/src/base54.rs

Dunqing: I guess we can use base64. Using `base36` is to align with its Babel plugin.

https://github.com/styled-components/babel-plugin-styled-components/blob/4e2eb388d9c90f2921c306c760657d059d01a518/src/utils/hash.js#L54-L55. 

Ok, I will follow this kind of optimization.

overlookmotel: It's not a big thing! Was just asking if there was a reason because base36 is unusual. Odd choice from `styled-components` authors!

crates/oxc_transformer/src/plugins/mod.rs
Dunqing: Yes, I thought of this and agree with that second way.

Dunqing:
> Not too bad!
> 
> bench

I guess we don't have a benchmark that can cover all the features that this plugin has, so we don't see much performance change.

Dunqing:
> Another thought: Are we able to disable this plugin on files which aren't JSX/TSX? Or does it need to run on JS files too? The loop through all top-level statements is probably quite costly for long files, and it'd be ideal to skip it if we can.

It needs to run in all files. One possible case that we need to consider is that they would like to define all of the styled components in a utility file (.js or .ts), and use them in the JSX/TSX files.

Another serious performance hit if we need to support CJS, we need to go through the whole AST to find whether there is a `const variable =  require('styled-components')`. This is the reason I don't support it at this stage.

crates/oxc_transformer/src/plugins/styled_components.rs
Dunqing: I went through the styled-components documentation and some articles about it, but I don't see any section that describes which way is better for performance. And I don't want to dig into the implementation, so I would prefer to keep this as-is to maintain compatibility.

crates/oxc_transformer/src/plugins/styled_components.rs
Dunqing: Aha, it should be, I miss to support `ComputedMemberExpression`. I will do it in the follow-up

Dunqing:
Consider this stack of PR to be huge, so I am merging this stack and following your suggestion and finishing some missing work in the follow-up PRs later.

overlookmotel:
> Another serious performance hit if we need to support CJS, we need to go through the whole AST to find whether there is a `const variable = require('styled-components')`. This is the reason I don't support it at this stage.

Argh! Maybe we need to include `require` in the module record that parser produces, and use that in transformer.

I suppose we could more easily support `require('styled-components')` at top level only - which is the common use case.

crates/oxc_transformer/src/plugins/styled_components.rs
overlookmotel: @Dunqing I realized now that maybe there *is* a reason. Maybe class names are case insensitive in some browsers?

PR #12078, opened 2025-07-04
feat(minifier): implement `Number` known methods

7086cmd:
My vacation is near, so I may have some time resuming my contributions. Here it is just implementing some functionalities that may never be used in actual usage :-)

I'm not sure if the minsize result is accurate, since there may be a problem with my `minsize` locally. I'll check my local `minsize` later, and I'll use the CI as a test this time.

### DIFF
--- a/crates/oxc_minifier/src/peephole/replace_known_methods.rs
+++ b/crates/oxc_minifier/src/peephole/replace_known_methods.rs
@@ -1065,9 +1065,7 @@ impl<'a> PeepholeOptimizations {
 /// Port from: <https://github.com/google/closure-compiler/blob/v20240609/test/com/google/javascript/jscomp/PeepholeReplaceKnownMethodsTest.java>
 #[cfg(test)]
 mod test {
-    use cow_utils::CowUtils;
     use oxc_syntax::es_target::ESTarget;
-    use std::borrow::Cow;
 
     use crate::{
         CompressOptions,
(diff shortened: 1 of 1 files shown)


crates/oxc_minifier/src/peephole/replace_known_methods.rs
```diff
@@ -1640,46 +1687,43 @@ mod test {
     #[test]
     #[ignore]
```
Boshen: These tests are still ignored.

7086cmd: Yes, I haven't implemented `Math.pow` (nor does the PR title include a change related to `Math.pow`). Hereby, I just modified the test caller, which should be `test_value` when it comes with expressions instead of statements.

I don't think the `Math.pow` functionality holds continuity and consistency with this PR, so I may prefer implementing it later.

7086cmd: Well, I reverted changes related to `Math.pow` tests, and I will fix it in the next turn to reduce confusion.

7086cmd:
Wait, how can altering testing function produces non-related benchmark degrade dramatically?!

overlookmotel:
> Wait, how can altering testing function produces non-related benchmark degrade dramatically?!

If you're referring to the mangler benchmarks, you can ignore them. They display a crazy amount of random variance. We've tried to fix them, but not had luck as yet.

PR #12084, opened 2025-07-04
feat(linter): implement react/display-name rule

taearls:
I implemented the `react/display-name` rule. I categorized it as style because it's more idiomatic to add displayNames and helps with debugging, but it does not affect functionality and is not required for React to work. 

In order to get all the tests to pass, I needed to add a `version` and a `component_wrapper_functions` setting to the `react.rs` module. This resulted in some snapshot updates.

All the options available for this rule are supported with this change. However, I commented out [one specific test](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/display-name.js#L461) from the `eslint-plugin-react` test suite because when it's enabled, it throws an unexpected token error. It seems like their test suite is expecting Flow rather than TypeScript. I figure that it's out of scope for this project to support Flow, so I didn't bother with this test.

I added some utility functions to the react utils file that were useful for checking for the presence of jsx or a HOC.

I tried to use CompactStr and FxHashSet rather than vec and String where possible. I also tried to make this implementation as simple as possible, but there are a lot of use cases to account for. 

I'm open to any feedback you have!


taearls:
@camc314 I fixed the build on this branch and rebased off of main. this is ready for review whenever you get a chance!

crates/oxc_linter/src/snapshots/react_display_name.snap
```diff
@@ -0,0 +1,423 @@
+---
+source: crates/oxc_linter/src/tester.rs
+---
+  ⚠ eslint-plugin-react(display-name): Component definition is missing display name
+   ╭─[display_name.tsx:2:18]
+ 1 │     
+ 2 │ ╭─▶                     var Hello = createReactClass({
```
camc314: If possible, it would be great if we could just report on the `var Hello`? Feel free to do it in a different PR, i know it's taken me a while to get to this.

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2672 @@
...
+    fn run_once(&self, ctx: &LintContext) {
+        let mut tracker = ComponentTracker::new();
+        let mut version_cache = VersionCache::new();
+        let ignore_transpiler_name = self.0.ignore_transpiler_name;
+        // Only check context objects if React version is >= 16.3.0
+        let check_context_objects =
+            self.0.check_context_objects && version_cache.get_context_objects_compatible(ctx);
+
+        // Single pass: collect React components and process displayName assignments
+        for node in ctx.nodes() {
```
camc314: looping over all nodes is slow, can we not do something like

```
fn run(node) {
    if is_react_component(node) {
        // look at the components symbol id, get it's references and look to see whether displayname is applied

Comment (taearls) [reply crates/oxc_linter/src/rules/react/display_name.rs]: yeah that makes sense I'll look into that

Comment (taearls) [reply crates/oxc_linter/src/rules/react/display_name.rs]: I'm still working through this. I think the only tricky part is dealing with anonymous components that don't have a name.

for example, this test case with no identifier for a variable or function expects to fail:
```
module.exports = () => {
    return <div>Hello {props.name}</div>;
}
```

File: crates/oxc_linter/src/snapshots/react_display_name.snap (hunk 1 above)
Comment (taearls) [reply crates/oxc_linter/src/snapshots/react_display_name.snap]: I believe I've addressed this. I only updated the snapshots for the variable declarations. I wasn't sure if we wanted to also only highlight only names for other AstKinds as well. 

iirc I think eslint will highlight the full component in typical function component declarations so I didn't want to touch that, but let me know if you'd like that changed as well!

File: crates/oxc_linter/src/rules/react/display_name.rs (hunk 2 above)
Comment (taearls) [reply crates/oxc_linter/src/rules/react/display_name.rs]: Ok I believe I've figured out a way to cover all the test cases without iterating the nodes. take a look when you get a chance and let me know what you think!

**Comment by connorshea** [comment]
This'll be great to have, thank you!

**Comment by taearls** [comment]
@camc314 when you get a chance could you give this PR another look? Thanks!

File: crates/oxc_linter/src/rules/react/display_name.rs [hunk 3]
```diff
@@ -0,0 +1,2402 @@
...
+    /// ```
+    ///
+    /// Examples of **correct** code for this rule:
+    /// ```jsx
+    /// const MyComponent = () => <div>Hello</div>;
+    /// MyComponent.displayName = 'MyComponent';
+    /// ```
+    DisplayName,
+    react,
+    style,
```
camchenry: I think `pedantic` might be more appropriate? I can see the argument for `style`, but since this actually has an effect on the runtime behavior, that makes more sense to me.

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+    Function,
+    Class,
+    CreateReactClass,
+    ObjectProperty,
+}
+
+#[derive(Debug)]
+struct ReactComponentInfo {
+    span: Span,
+    _component_type: ComponentType,
```
camchenry: Any reason we need the underscore here?


```suggestion
    component_type: ComponentType,
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+        {
+            for &reference_id in module_reference_ids {
+                let reference = ctx.scoping().get_reference(reference_id);
+                let node = ctx.nodes().get_node(reference.node_id());
+
+                // Walk up to find if this is part of module.exports assignment
+                for ancestor_id in ctx.nodes().ancestor_ids(node.id()) {
+                    let ancestor = ctx.nodes().get_node(ancestor_id);
+
+                    if let AstKind::AssignmentExpression(assign) = ancestor.kind() {
```
camchenry: `ancestor_kinds` could be used here directly if we don't need `ancestor_id` or the node specifically:


```suggestion
                for ancestor_kind in ctx.nodes().ancestor_kinds(node.id()) {
                    if let AstKind::AssignmentExpression(assign) = ancestor_kind {
```

There's a number of places this pattern occurs, so some advice applies to all of them. Fewer `.kind()` calls is also good for performance (marginally).

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+        let ref_node = ctx.nodes().get_node(reference.node_id());
+
+        // Walk up to find the assignment expression
+        for ancestor_id in ctx.nodes().ancestor_ids(ref_node.id()) {
+            let ancestor = ctx.nodes().get_node(ancestor_id);
+
+            if let AstKind::AssignmentExpression(assign) = ancestor.kind()
+                && let Expression::CallExpression(call) = &assign.right
+                && let Some(callee_name) = call.callee_name()
+                && (callee_name == "createContext" || callee_name.ends_with(".createContext"))
```
camchenry: This code appears in a few places: what do you think about turning it into a little shared function?


```suggestion
                && is_create_context_call(call)
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+    obj_expr: &ObjectExpression,
+    ignore_transpiler_name: bool,
+) -> bool {
+    for prop in &obj_expr.properties {
+        if let ObjectPropertyKind::ObjectProperty(obj_prop) = prop
+            && let PropertyKey::StaticIdentifier(ident) = &obj_prop.key
+            && obj_prop.method
+            && is_react_component_name(&ident.name)
+            && matches!(&obj_prop.value, Expression::FunctionExpression(f) if function_contains_jsx(f))
+            && ignore_transpiler_name
```
camchenry: If the order of these isn't important, these could be ordered to put the fast checks first (just move up the `ignore_transpiler_name`):


```suggestion
        if let ObjectPropertyKind::ObjectProperty(obj_prop) = prop
            && let PropertyKey::StaticIdentifier(ident) = &obj_prop.key
            && obj_prop.method
            && ignore_transpiler_name
            && is_react_component_name(&ident.name)
            && matches!(&obj_prop.value, Expression::FunctionExpression(f) if function_contains_jsx(f))

```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+    if direct_jsx {
+        return Some((ComponentType::Function, false)); // Not a HOF, directly contains JSX
+    }
+
+    // Check if this is a Higher-Order Function that returns a function with JSX
+    if let Some(innermost) = find_innermost_function_with_jsx(expr, ctx) {
+        // Check if the innermost function has a name
+        let inner_has_name = match innermost {
+            crate::utils::InnermostFunction::Function(func) => func.id.is_some(),
+            crate::utils::InnermostFunction::ArrowFunction(_) => false,
```
camchenry: Stylistically I'd probably recommend importing these qualified types:


```suggestion
            InnermostFunction::Function(func) => func.id.is_some(),
            InnermostFunction::ArrowFunction(_) => false,
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+
+/// Handle anonymous export default components
+fn is_anonymous_export_component<'a>(
+    export: &oxc_ast::ast::ExportDefaultDeclaration<'a>,
+    _ctx: &LintContext<'a>,
+    _version_cache: &mut VersionCache,
+    ignore_transpiler_name: bool,
+) -> Option<ReactComponentInfo> {
+    match &export.declaration {
+        oxc_ast::ast::ExportDefaultDeclarationKind::ArrowFunctionExpression(func) => {
```
camchenry: Can import this type to avoid qualifying it:


```suggestion
        ExportDefaultDeclarationKind::ArrowFunctionExpression(func) => {
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+        }
+    }
+    false
+}
+
+/// Handle anonymous export default components
+fn is_anonymous_export_component<'a>(
+    export: &oxc_ast::ast::ExportDefaultDeclaration<'a>,
+    _ctx: &LintContext<'a>,
+    _version_cache: &mut VersionCache,
```
camchenry: Looks like these are unused, can we remove them?


```suggestion
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+        {
+            return true;
+        }
+    }
+    false
+}
+
+/// Handle anonymous export default components
+fn is_anonymous_export_component<'a>(
+    export: &oxc_ast::ast::ExportDefaultDeclaration<'a>,
```
camchenry:
```suggestion
    export: &ExportDefaultDeclaration<'a>,
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+    }
+
+    None
+}
+
+/// Handle module.exports assignments
+fn is_module_exports_component<'a>(
+    assign: &oxc_ast::ast::AssignmentExpression<'a>,
+    _ctx: &LintContext<'a>,
+    _version_cache: &mut VersionCache,
```
camchenry: Could these be removed if they're unused?


```suggestion
```

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+            });
+        }
+    }
+
+    None
+}
+
+/// Handle module.exports assignments
+fn is_module_exports_component<'a>(
+    assign: &oxc_ast::ast::AssignmentExpression<'a>,
```
camchenry:
```suggestion
    assign: &AssignmentExpression<'a>,
```

crates/oxc_linter/src/utils/react.rs
```diff
@@ -351,3 +352,146 @@ pub fn is_state_member_expression(expression: &StaticMemberExpression<'_>) -> bo
...
+    // Check component wrapper functions from settings
+    ctx.settings().react.is_component_wrapper_function(callee_name)
+}
+
+/// Finds the innermost function with JSX in a chain of HOC calls
+#[derive(Debug)]
+pub enum InnermostFunction<'a> {
+    Function(&'a Function<'a>),
+    #[expect(dead_code, reason = "False positive: field is read in pattern matching")]
+    ArrowFunction(&'a ArrowFunctionExpression<'a>),
```
camchenry: Are you sure this is still necessary? I would have expected that the compiler would detect this correctly, but perhaps not.

camchenry (commented):
Truly sorry this hasn't been reviewed earlier. I think this looks pretty good overall. The logic seems generally sound to me, and if the tests are passing then I think this rule is in good shape.

I've left some comments to improve the style a little bit and also be slightly more idiomatic for our codebase, but I don't think there's anything major that needs to be addressed. I believe if the merge conflicts are resolved as well, we can try to get this merged.

@camc314 Do you mind taking a look at this as well and see if you agree?

taearls:
> Truly sorry this hasn't been reviewed earlier. I think this looks pretty good overall. The logic seems generally sound to me, and if the tests are passing then I think this rule is in good shape.
> 
> I've left some comments to improve the style a little bit and also be slightly more idiomatic for our codebase, but I don't think there's anything major that needs to be addressed. I believe if the merge conflicts are resolved as well, we can try to get this merged.
> 
> @camc314 Do you mind taking a look at this as well and see if you agree?

Thanks for your feedback! I'll try to address it this week.

crates/oxc_linter/src/utils/react.rs
```diff
@@ -351,3 +352,146 @@ pub fn is_state_member_expression(expression: &StaticMemberExpression<'_>) -> bo
...
+            if let Statement::ReturnStatement(ret_stmt) = stmt
+                && let Some(expr) = &ret_stmt.argument
+                && contains_jsx(expr)
+            {
+                return true;
+            }
+        }
+    }
+    false
+}
```
camc314, line 473: this function won't work properly as soon as the return is in a block statment.
e.g.
```
function Foo() {
    if (true) {
        return <div></div>
    }
}

File: crates/oxc_linter/src/utils/react.rs [hunk 16]
```diff
@@ -351,3 +352,146 @@ pub fn is_state_member_expression(expression: &StaticMemberExpression<'_>) -> bo
...
+                None
+            }
+        }
+        _ => None,
+    }
+}
+
+/// Checks if an expression contains JSX elements
+pub fn contains_jsx(expr: &Expression) -> bool {
+    match expr {
```
camc314: this breaks down, e.g. what about if you;ve got a type cast?

it's probably better to use the visitor pattern here

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+                return Some(ReactComponentInfo {
+                    span,
+                    _component_type: ComponentType::Class,
+                    is_context: false,
+                    name: Some(CompactStr::from(name.name.as_str())),
+                });
+            }
+
+            // Check if extends React.Component
+            let extends_react_component = class.super_class.as_ref().is_some_and(|super_class| {
```
camc314: This code is exactly the same as L 884, it can be extracted out into a funciton

crates/oxc_linter/src/rules/react/display_name.rs
```diff
@@ -0,0 +1,2402 @@
...
+    ctx: &LintContext,
+) -> bool {
+    if !reference.is_read() {
+        return false;
+    }
+
+    let reference_node = ctx.nodes().get_node(reference.node_id());
+
+    // Walk up to find if this is part of: Component.displayName = ...
+    for ancestor_id in ctx.nodes().ancestor_ids(reference_node.id()) {
```
camc314: this parent walking is unneccessary, and it likley means we'll go further up the tree than we need to, hurting perf. Something like:
```
ctx.scoping().get_resolved_references(symbol_id).any(|reference| {
        if !reference.is_read() {
            return false;
        }

        let reference_node = ctx.nodes().get_node(reference.node_id());

        if !matches!(reference_node.kind(), AstKind::IdentifierReference(_)) {
            return false;
        }

        let mut parents = iter_outer_expressions(ctx.nodes(), reference_node.id());

        if let Some(next) = parents.next() {
            if let AstKind::StaticMemberExpression(me) = next {
                if me.property.name != "displayName" {
                    return false;
                }
            }
        }

        if let Some(next) = parents.next() {
            if let AstKind::AssignmentExpression(assign) = next {
                if assign.left.span().contains_inclusive(reference_node.span()) {
                    return true;
                }
            }
        }

        false
    })
```

should work instead

crates/oxc_linter/src/rules/react/display_name.rs
camc314: it's impossible for `callee_name` to end with `.createContext`

The only case would be if you had code like:

```
foo["bar.createContext"]()
```

but we don't care about that case since you're not accessing the createContext property, you're accessing some other property.

This is the same for the other `ends_with(".createContext")` checks in this file

crates/oxc_linter/src/utils/react.rs
camc314: I looked at this, rust is correct here, you never read the value of the enum member. This can just be simplified to `ArrowFunction`, and we can avoid passing some extra data around

crates/oxc_linter/src/rules/react/display_name.rs
camc314: It's unused code. I'm guessing it's prefixed with an `_` so that the compiler doesn't report a dead code warning.

We should just remove this struct property.

camc314:
Thanks for working so hard on this, it's a complex rule. Apologies for the delay in getting to it

> @camc314 Do you mind taking a look at this as well and see if you agree?

Yeah, I agree - thank you for reviewing.

taearls:
> Thanks for working so hard on this, it's a complex rule. Apologies for the delay in getting to it
> 
> > @camc314 Do you mind taking a look at this as well and see if you agree?
> 
> Yeah, I agree - thank you for reviewing.

thanks for your feedback, both of you. I appreciate it. in hindsight I should've been more thorough in reviewing my own code before opening this PR and that's on me. 

I resolved the merge conflicts. I'll start working through each piece of feedback today. It might be a few days before this is ready for re-review.

taearls:
I have addressed all the feedback and rebased onto main. thanks for your patience! this is ready for another review when you get a chance @camc314 @camchenry

taearls:
@camc314 @camchenry when you can, could I get a review on this? I think it's ready to go.

camchenry:
Hi @taearls. Sorry for the very long timeline in this PR, but we are going to go ahead and merge this! Thank you for all of your work on this. There are some merge conflicts currently, but I've pushed up fixes for those to a separate PR which we'll merge directly: https://github.com/oxc-project/oxc/pull/18426.

taearls:
> Hi @taearls. Sorry for the very long timeline in this PR, but we are going to go ahead and merge this! Thank you for all of your work on this. There are some merge conflicts currently, but I've pushed up fixes for those to a separate PR which we'll merge directly: https://github.com/oxc-project/oxc/pull/18426.

I understand! I get that there have been higher priorities. Thanks for reviewing and accepting my PR!

PR #12087, opened 2025-07-05
refactor(semantic): remove Option from parent_* methods

### LINKED ISSUE #11946: semantic: Remove `Option` from `parent_*` methods
Originally suggested by @overlookmotel. Currently, all of the methods relating to getting a node's parent ID/kind/node return `Option<&AstNode>`. However, we know that all AST nodes should have a parent, with the exception of `Program`. In this case, we could have `Program` be self-referential: it will return its parent ID as itself, and then we can remove `Option` from the return type of all methods fetching parent data. This would remove a lot of boilerplate code in the linter and elsewhere. Could potentially improve performance as well by removing lots of unnecessary branches which means fewer instructions.

ulrichstark:
Closes #11946

> @camchenry in #11946:
> Originally suggested by @overlookmotel. Currently, all of the methods relating to getting a node's parent ID/kind/node return Option<&AstNode>. However, we know that all AST nodes should have a parent, with the exception of Program. In this case, we could have Program be self-referential: it will return its parent ID as itself, and then we can remove Option from the return type of all methods fetching parent data. This would remove a lot of boilerplate code in the linter and elsewhere. Could potentially improve performance as well by removing lots of unnecessary branches which means fewer instructions.

This required updating logic in multiple cases, because parent of Program is now Program itself. I managed to get all tests to pass, but it's very likely that I have broken something untested. Broken logic will surface as incorrect results, stack overflows or infinite loops.

Edit:
Seems like semantic benchmarks have improved a bit (+0.3%) and linter benchmarks have slightly regressed (-0.1%). I suspect the now required additional checks in ancestors, ancestor_kinds and ancestor_ids for the slight slowdown of the linter benchmarks.

### DIFF
--- a/crates/oxc_linter/src/ast_util.rs
+++ b/crates/oxc_linter/src/ast_util.rs
@@ -204,7 +204,7 @@ pub fn get_enclosing_function<'a, 'b>(
         {
             return Some(current_node);
         }
-        current_node = semantic.nodes().parent_node(current_node.id())?;
+        current_node = semantic.nodes().parent_node(current_node.id());
     }
 }
 
@@ -222,11 +222,10 @@ pub fn outermost_paren<'a, 'b>(
     let mut node = node;
 
     loop {
-        if let Some(parent) = semantic.nodes().parent_node(node.id()) {
-            if let AstKind::ParenthesizedExpression(_) = parent.kind() {
-                node = parent;
-                continue;
-            }
+        let parent = semantic.nodes().parent_node(node.id());
+        if let AstKind::ParenthesizedExpression(_) = parent.kind() {
+            node = parent;
+            continue;
         }
 
         break;
@@ -652,7 +651,7 @@ pub fn is_default_this_binding<'a>(
 
     let mut current_node = node;
     loop {
-        let parent = semantic.nodes().parent_node(current_node.id()).unwrap();
+        let parent = semantic.nodes().parent_node(current_node.id());
         let parent_kind = parent.kind();
         match parent_kind {
             AstKind::ChainExpression(_)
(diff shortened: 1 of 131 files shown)


overlookmotel:
We might be able to improve perf of `AstNodes::ancestor_ids` a little by using a custom iterator, instead of `std::iter::successors`.

Also, `node_id == NodeId::DUMMY` would be a cheaper check than `parent_id == node_id`. The `NodeId` of `Program` in practice is always `NodeId::DUMMY` (0), but we don't have a static guarantee of that at present. But we could get that guarantee by adding `assert!(self.parent_ids.is_empty())` to the top of `AstNodes::add_program_node` (or maybe just `debug_assert!`).

But anyway, the main reason I suggested this change was to remove tons of boilerplate code from linter, rather than the perf impact.

camchenry (approved):
I think the improvements in readability of code in the linter alone are worth it. I think the change in the benchmarks is reasonable, and probably within what I would consider noise. Changes >=1% are a better signal typically.

I think my only hesitation is whether it makes sense change this API contract and whether we feel like it is confusing to not return `Option` on these functions. Would we ever want to return `None` in the future?

camchenry:
> Also, `node_id == NodeId::DUMMY` would be a cheaper check than `parent_id == node_id`. The `NodeId` of `Program` in practice is always `NodeId::DUMMY` (0), but we don't have a static guarantee of that at present. But we could get that guarantee by adding `assert!(self.parent_ids.is_empty())` to the top of `AstNodes::add_program_node` (or maybe just `debug_assert!`).

We should do this in this PR, or at least as a follow-up PR.

PR #12095, opened 2025-07-06
perf(parser): optimize around `parse_return_type`

ulrichstark:
- Removed `has_return_colon` because having both `return_type` and `has_return_colon` on `ArrowFunctionHead` was redundant
- Avoid redundant `self.is_ts` checks by moving only required ones to call site of `parse_ts_return_type_annotation`
- Removed unnecessary parameters `return_kind` and `is_type` from `parse_return_type` and their corresponding checks

Benchmark results:
```
parser[react.development.js] : +0.15%
parser[binder.ts]            : +0.05%
parser[cal.com.tsx]          : +0.01%
```

I expected more improvements on the benchmarks, but less code is also worth the change on its own.

### DIFF
--- a/crates/oxc_parser/src/js/arrow.rs
+++ b/crates/oxc_parser/src/js/arrow.rs
@@ -12,7 +12,6 @@ struct ArrowFunctionHead<'a> {
     return_type: Option<Box<'a, TSTypeAnnotation<'a>>>,
     r#async: bool,
     span: u32,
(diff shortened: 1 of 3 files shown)


Boshen:
@ulrichstark It seems like you are getting good with the codebase and the parser, may I invite you to work on the minifier as well? Contact me on discord if you are intrested, and I'll get you acquainted.

PR #12185, opened 2025-07-10
fix(transformer/object-rest-spread): arrow function expressions with object rest patterns

### LINKED ISSUE #12184: transformer bug involving object rest spread and arrow function transformation
https://playground.oxc.rs/#eNpVkMFqwzAMhl8l6LRBCdtgl0J3Gr1ubHsBx1GKwbaCJHcNIe8+OaGBnSxZ8v9/v2fwcARPWbQZcnNqHua2bR1fZHlsTm9NreAABMcZuOR6yJTV3eCoXPAAMWS91+JpxL2ZUkfx3im7LANx2i6WA4yOBbkKuhjp9wu1cP4oKqHHc8leA5nd9npktN0rfjrGLLsDFfZ4DhGzS+YLiqKtys2At9nPVHkgUV8igplW2tXU6h1po1CLihYFUF6enl9NIwhFp9i/o4+OXQUy78FFqfyeerzg+iPm30X8Xi2TG/eIKeQwhE3eU6op/glkZYpnC18XrsgdieFu82X5Axr9kgw=

incorrect transformation as `args` is supposed to be returned but it's just left with an empty statement that does nothing `args;`

cod1r:
close: https://github.com/oxc-project/oxc/issues/12184

I couldn't figure out where **exactly** to put tests regarding object rest spread and arrow function transformation.

Probably somewhere underneath tasks/transform_conformance


Dunqing:
You could put a test at https://github.com/oxc-project/oxc/blob/224775c056708f0df74e94903f3f2bf2fc0bb2a9/tasks/transform_conformance/tests/babel-plugin-transform-object-rest-spread/test/fixtures.

Dunqing:
Have you found that your test fails? Generally, we use `babelHelpers.helper` instead of injecting a helper function in tests. By the way, you can run `just test-transform --filter TEST-PATH` to test what code it transformed.

cod1r:
Oh ye sry, I was thinking there'd be some way to get the right output but didn't know how to get it. I just thought I'd use the babel playground and get it there by pasting in the input.js contents 😄 .

Dunqing:
> Oh ye sry, I was thinking there'd be some way to get the right output but didn't know how to get it. I just thought I'd use the babel playground and get it there by pasting in the input.js contents 😄 .

You can get a correct output from the `Babel` playground, but you should add `@babel/plugin-external-helper`. See [Babel Playground](https://babeljs.io/repl#?browsers=&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=MYewdgzgLgBAZmGBeGAKA3jAdDghgJwHMIYBfASmQD4YDiBuIA&forceAllTransforms=false&modules=false&shippedProposals=true&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.28.0&externalPlugins=%40babel%2Fplugin-transform-object-rest-spread%407.28.0%2C%40babel%2Fplugin-external-helpers%407.27.1&assumptions=%7B%7D)

crates/oxc_transformer/src/es2018/object_rest_spread.rs
```diff
@@ -576,6 +576,18 @@ impl<'a> ObjectRestSpread<'a, '_> {
...
+                .statements
+                .iter_mut()
+                .rev()
+                .find(|s| matches!(s, Statement::ExpressionStatement(_)));
+            if let Some(expr_stmt) = expr_stmt {
+                if let Statement::ExpressionStatement(inner_expr_stmt) = expr_stmt {
+                    let expression = Some(inner_expr_stmt.expression.take_in(ctx.ast));
+                    *expr_stmt = ctx.ast.statement_return(SPAN, expression);
+                }
+            }
```
Dunqing: The above `Self::has_nested_object_rest` check has determined that this arrow function needs to be transformed. That is, we can move this whole branch to that and place it before `Self::replace_rest_element`, so that it should only have one statement in it.

Then you can simplify it like:

```rs
// `({ ...args }) => { args }`
if arrow.expression {
    arrow.expression = false;

    debug_assert!(arrow.body.statements.len() == 1);

    let Statement::ExpressionStatement(stmt) = arrow.body.statements.pop().unwrap()
    else {
        unreachable!(
            "`arrow.expression` is true, which means it has only one ExpressionStatement."
        );
    };
    let return_stmt =
        ctx.ast.statement_return(stmt.span, Some(stmt.unbox().expression));
    arrow.body.statements.push(return_stmt);
}
```

Dunqing (approved):

Dunqing:
There is a conflict in `oxc.snap.md`, can you run `just test-transform` to update it?

Dunqing (approved):
Thanks!

PR #12197, opened 2025-07-10
docs(transformer/styled-components): add comments about `CSSMinifier`

Dunqing:
Added some comments to describe `CSSMinifier`, and clarify why `quasis.set_len(expressions.len() + 1);` is safe.

Also, there is a `assert` in codegen to ensure that `quasis` is always one greater than `expressions`

https://github.com/oxc-project/oxc/blob/c003ebad60970d27aae48b488ff361ef2b612379/crates/oxc_codegen/src/gen.rs#L2091-L2092

In addition, I found that we can use `truncate()` instead of `set_len`. Both are the same, but `truncate` is a safe function.

### DIFF
--- a/crates/oxc_transformer/src/plugins/styled_components.rs
+++ b/crates/oxc_transformer/src/plugins/styled_components.rs
@@ -416,8 +416,15 @@ impl<'a> StyledComponents<'a, '_> {
             quais.value.raw = new_raw;
         }
 
-        // SAFETY: `expressions` is always one less than `quasis`.
+        // SAFETY:
+        // This is safe because template literal has ensured that `quasis` always
+        // has one more element than `expressions`, and the `CSSMinifier` guarantees that
+        // once a expression is removed, the corresponding quasi will also be removed.
+        // Therefore, the length of `quasis` will always be one more than `expressions`.
+        // So we can safely set the length of `quasis` to `expressions.len() + 1`.
         unsafe {
+            // Set the length of `quasis` to `expressions.len() + 1` to truncate the quasis that
+            // have been minified out.
             quasis.set_len(expressions.len() + 1);
         }
     }
@@ -838,6 +845,8 @@ fn is_valid_styled_component_source(source: &str) -> bool {
     )
 }
 
+/// A CSS minifier that is specifically designed to minify CSS code within
+/// styled-components template literals.
 pub struct CssMinifier<'a> {
     ast: AstBuilder<'a>,
 }
@@ -850,6 +859,17 @@ impl<'a> CssMinifier<'a> {
         Self { ast }
     }
 
+    /// Minifies the CSS code within a series of template literal `quasis`.
+    ///
+    /// This function takes an array of `TemplateElement` (quasis) and
+    /// first injects unique placeholders for the expressions between them.
+    /// Then, passes the processed CSS string to the [CssMinifier::minify_css].
+    ///
+    /// ### Returns:
+    ///
+    /// A tuple containing:
+    /// First: A vector of `Atom` containing the minified CSS strings.
+    /// Second: A set of indices representing which expressions were kept.
     pub fn minify_quasis(
         quasis: &[TemplateElement<'a>],
         ast: AstBuilder<'a>,
@@ -865,6 +885,19 @@ impl<'a> CssMinifier<'a> {
         minifier.minify_css(css)
     }
 
+    /// Injects unique placeholders into a series of `quasis` to represent expressions.
+    ///
+    /// This is a key step for minification. By replacing expressions with placeholders,
+    /// we can treat the entire template literal as a single block of CSS, which simplifies
+    /// the minification process. The placeholders are designed to be unique and unlikely
+    /// to appear in regular CSS.
+    ///
+    /// # Example
+    ///
+    /// Given quasis from `` `width: ${width}px; color: ${color};` ``, which are
+    /// `["width: ", "px; color: ", ";"]`, and expressions `[width, color]`,
+    /// this function will produce the string:
+    /// `"width: __PLACEHOLDER_0__px; color: __PLACEHOLDER_1__;"`
     fn inject_unique_placeholders(&self, quasis: &[TemplateElement]) -> &str {
         let estimated_capacity: usize = quasis.iter().map(|s| s.value.raw.len()).sum::<usize>()
             + (quasis.len() - 1)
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#12212** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12198** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12198?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12211** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12202** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12200** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12199** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12199?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12197** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12197?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12197?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12197).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -412,8 +412,15 @@ impl<'a> StyledComponents<'a, '_> {
             quais.value.raw = new_raw;
         }
 
-        // SAFETY: `expressions` is always one less than `quasis`.
+        // SAFETY:
+        // This is safe because template literal has ensured that `quasis` always
+        // has one more element than `expressions`, and the `CSSMinifier` guarantees that
```
overlookmotel, line 421: What makes it true that "template literal has ensured that `quasis` always has one more element than `expressions`"? How is that ensured?

If in some other transform, I do `expressions.push(...)` without corresponding `quasis.push(...)` by accident, this invariant is broken, and `quasis.set_len(expressions.len() + 1)` is UB. Safety invariants should be local and provable. "SAFETY: There are no bugs in the rest of the codebase" isn't very convincing!

> Also, there is a `assert` in codegen to ensure that `quasis` is always one greater than `expressions`
> 
> https://github.com/oxc-project/oxc/blob/c003ebad60970d27aae48b488ff361ef2b612379/crates/oxc_codegen/src/gen.rs#L2091-L2092

1. It's `debug_assert!` not `assert!`. That's no good as a safety guarantee.
2. It's in codegen. Even if it was `assert!`, it's too late. We's already invoke UB before we get to that.

> In addition, I found that we can use `truncate()` instead of `set_len`. Both are the same, but `truncate` is a safe function.

`truncate` would be fine - it might panic if some other code has broken the rule, but won't cause UB.

Or we could add an assertion at top of the function `assert!(quasis.len() == expressions.len() + 1)`, and then use unsafe `set_len`.

But... this PR is mainly about adding comments. Let's leave it as is here, and address it in a separate PR.

overlookmotel: Personally, I'd quite like to make this a type system invariant by changing `TemplateLiteral` to:

```rs
pub struct TemplateLiteral<'a> {
    pub span: Span,
    pub lead: Vec<'a, TemplateElementAndExpression<'a>>,
    pub tail: TemplateElement<'a>,
}

pub struct TemplateElementAndExpression<'a> {
    quasi: TemplateElement<'a>,
    expression: Expression<'a>,
}
```

Then we can stop worrying about this! Rust will prevent us making mistakes.

overlookmotel (approved):
Apart from the typo Graphite's spotted (it *finally* got something right), all looks good. Thank you for documenting this.

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -903,6 +940,59 @@ impl<'a> CssMinifier<'a> {
...
+    /// 2. It iterates through each byte of the CSS string:
+    ///   - If it encounters a string quote (`"` or `'`), it toggles the `in_string` state.
+    ///   - If it finds a placeholder (starting with `_` and followed by
+    ///     [CssMinifier::PLACEHOLDER_PREFIX]), it tries to parse the placeholder and
+    ///     keeps track of its index.
+    ///   - It skips comments (both block and line comments) and whitespace, compressing them
+    ///    as needed.
+    ///   - It handles escaped newlines by replacing them with a space.
+    /// 3. It collects the minified CSS parts into `new_raws` and tracks which
+    ///   expressions were kept in a `remaining_expression_indexes`.
```
overlookmotel: The nittiest nit of all time!

```suggestion
    /// 1. It initializes state variables to track whether it's inside a string,
    ///    the current string character, and the depth of parentheses.
    /// 2. It iterates through each byte of the CSS string:
    ///   - If it encounters a string quote (`"` or `'`), it toggles the `in_string` state.
    ///   - If it finds a placeholder (starting with `_` and followed by
    ///     [CssMinifier::PLACEHOLDER_PREFIX]), it tries to parse the placeholder and
    ///     keeps track of its index.
    ///   - It skips comments (both block and line comments) and whitespace, compressing them
    ///     as needed.
    ///   - It handles escaped newlines by replacing them with a space.
    /// 3. It collects the minified CSS parts into `new_raws` and tracks which
    ///    expressions were kept in a `remaining_expression_indexes`.
```

Dunqing: Sharp eyes!

overlookmotel: I'm a pedant through and through!

crates/oxc_transformer/src/plugins/styled_components.rs
overlookmotel: @Dunqing Ping just to make sure you see this. I had to mark it as "resolved" to merge the PR.

Dunqing: The new type looks good, the current one is probably to align with the `ESTree`. I will add that assertion later.

overlookmotel: #12213

PR #12225, opened 2025-07-11
feat(ast): compute max value that can be mapped to `AstType`

camchenry:
- part of https://github.com/oxc-project/oxc/issues/12223

This adds an automatically generated `usize` which represents the largest variant that is contained in `AstKind`/`AstType`. In following PRs, this is used to create a set of bits that is exactly sized to the number of different AST node types there are.

PR #12226, opened 2025-07-11
feat(semantic): add set of `AstType` that exist in AST

camchenry:
- part of https://github.com/oxc-project/oxc/issues/12223

Adds a bitset that we compute as we push nodes. Since we have to call `add_node` for every AST node anyway, adding a small bit of code here to insert a single bit into the set should be very cheap. This enables us to do optimizations later on in the linter to quickly check if a file has any of a specific node kind. I've added some debug assertions to check that the bounds checks are respected so that we can do an `insert_unchecked` call. Not sure if this is good enough or not though.

### DIFF
--- a/crates/oxc_semantic/src/node.rs
+++ b/crates/oxc_semantic/src/node.rs
@@ -1,7 +1,8 @@
 use std::iter::FusedIterator;
 
+use fixedbitset::FixedBitSet;
 use oxc_allocator::{Address, GetAddress};
-use oxc_ast::{AstKind, ast::Program};
+use oxc_ast::{AstKind, ast::Program, ast_kind::AST_TYPE_MAX};
 use oxc_cfg::BlockNodeId;
 use oxc_index::{IndexSlice, IndexVec};
 use oxc_span::{GetSpan, Span};
@@ -96,12 +97,27 @@ impl GetAddress for AstNode<'_> {
 }
 
 /// Untyped AST nodes flattened into an vec
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct AstNodes<'a> {
     program: Option<&'a Program<'a>>,
     nodes: IndexVec<NodeId, AstNode<'a>>,
+    /// Stores a set of bits of a fixed size, where each bit represents a single [`AstKind`]. If the bit is set (1),
+    /// then the AST contains at least one node of that kind. If the bit is not set (0), then the AST does not contain
+    /// any nodes of that kind.
+    node_kinds_set: FixedBitSet,
     /// `node` -> `parent`
     parent_ids: IndexVec<NodeId, NodeId>,
+}
+
+impl Default for AstNodes<'_> {
+    fn default() -> Self {
+        Self {
+            program: None,
+            nodes: IndexVec::new(),
+            node_kinds_set: FixedBitSet::with_capacity(AST_TYPE_MAX as usize + 1),
+            parent_ids: IndexVec::new(),
+        }
+    }
 }
 
 impl<'a> AstNodes<'a> {
@@ -210,6 +226,12 @@ impl<'a> AstNodes<'a> {
         let node_id = self.parent_ids.push(parent_node_id);
         let node = AstNode::new(kind, scope_id, cfg_id, flags, node_id);
         self.nodes.push(node);
+        debug_assert!((kind.ty() as usize) < self.node_kinds_set.len());
+        // SAFETY: `AstKind` maps exactly to `AstType`, and there should be exactly
+        // enough bits to insert it into `node_kinds`, so we can skip a bounds check here.
+        unsafe {
+            self.node_kinds_set.insert_unchecked(kind.ty() as usize);
+        }
         node_id
     }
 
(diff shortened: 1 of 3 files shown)


camchenry:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12226?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#12231** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12229** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12229?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12228** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12227** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12226** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12225** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12225?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12226).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_semantic/src/node.rs
```diff
@@ -96,14 +97,29 @@ impl GetAddress for AstNode<'_> {
...
     /// `node` -> `parent`
     parent_ids: IndexVec<NodeId, NodeId>,
 }
 
+impl Default for AstNodes<'_> {
+    fn default() -> Self {
+        Self {
+            program: None,
+            nodes: IndexVec::new(),
+            node_kinds_set: FixedBitSet::with_capacity(AST_TYPE_MAX + 1),
```
overlookmotel: Why `+ 1`?

camchenry: `AST_TYPE_MAX` is the largest index/offset. If for example, `AstType` only contained a single variant:

```rust
enum AstType {
  Program = 0
}
```

Then we would want to create a bitset with `0 + 1` = 1 bit in it.

overlookmotel:
I have a branch I never got around to polishing and pushing containing a fixed-size bitset (genuinely fixed size - size is `const`, and determined at compile time).

It's a generalization of this thing:

https://github.com/oxc-project/oxc/blob/242e7210aa37e6eef14457b28aebbc31035d3c28/tasks/ast_tools/src/schema/derives.rs

That might be useful here as it stores the data on stack instead of heap, so should be a bit cheaper, and it encapsulates the unsafe code.

Here is a cut-down version, specialized for `AstType`:

```rs
const USIZE_BITS: usize = usize::BITS as usize;

/// Number of bytes required for bit set which can represent all [`AstType`]s.
const NUM_USIZES: usize = (AST_TYPE_MAX + 1).div_ceil(USIZE_BITS);

/// Bit set with a bit for each [`AstType`].
pub struct AstTypesBitset([usize; NUM_USIZES]);

impl AstTypesBitset {
    /// Create empty [`AstTypesBitset`] with no bits set.
    pub const fn new() -> Self {
        Self([0; NUM_USIZES])
    }

    /// Returns `true` if bit is set for provided [`AstType`].
    pub const fn has(self, ty: AstType) -> bool {
        let (index, mask) = Self::index_and_mask(ty);
        (self.0[index] & mask) != 0
    }

    /// Set bit for provided [`AstType`].
    pub const fn set(&mut self, ty: AstType) {
        let (index, mask) = Self::index_and_mask(ty);
        self.0[index] |= mask;
    }

    /// Returns `true` if any bit is set in both `self` and `other`.
    pub fn intersects(&self, other: &Self) -> bool {
        let mut intersection = 0;
        for (&a, &b) in self.0.iter().zip(other.0.iter()) {
            intersection |= a & b;
        }
        intersection != 0
    }

    /// Get index and mask for an [`AstType`].
    /// Returned `index` is guaranteed not to be out of bounds of the array.
    const fn index_and_mask(ty: AstType) -> (usize, usize) {
        let n = ty as usize;
        let index = n / USIZE_BITS;
        let mask = 1usize << (n % USIZE_BITS);
        (index, mask)
    }
}
```

Note: The unsafe code may not even be necessary. Maybe compiler can prove from its knowledge of `AstType`'s variants that the index operations cannot be out of bounds, so `self.0[index]` produces same code as `self.0.get_unchecked(index)`. You could check that in Godbolt.

crates/oxc_semantic/src/node.rs
overlookmotel: Oh I get it. `AST_TYPE_MAX` is the max value, not the number of values.

overlookmotel:
Yes, it looks like compiler is smart enough to remove the bounds checks. No unsafe needed. https://godbolt.org/z/EjPnEde3G

I've edited the code above to remove the unsafe.

camchenry:
Closing in favor of https://github.com/oxc-project/oxc/pull/13137, which incorporates some of the above changes.

PR #12256, opened 2025-07-13
perf(transformer/styled-components): remove quasis and expressions in batch

Dunqing:
Remove `quasis` and `expressions` in batch instead of removing them one by one, so that it would be more efficient. Considering that the expression's removal is unlikely to be common, I add a `has_deletion_marker` flag to avoid unnecessary `retain` calls, so that we can keep the performance as before in this situation.

### DIFF
--- a/crates/oxc_transformer/src/plugins/styled_components.rs
+++ b/crates/oxc_transformer/src/plugins/styled_components.rs
@@ -868,6 +868,10 @@ fn is_valid_styled_component_source(source: &str) -> bool {
 /// ```
 fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a>) {
     const NOT_IN_STRING: u8 = 0;
+    /// `Span` used as a sentinel indicating quasi should be removed.
+    /// Source text is limited to max `u32::MAX` bytes, so it's impossible for a `TemplateElement`
+    /// to have this span, because it's always followed by (at minimum) a '`'.
+    const REMOVE_SENTINEL: Span = Span::new(u32::MAX, u32::MAX);
 
     let TemplateLiteral { quasis, expressions, .. } = lit;
 
@@ -883,6 +887,8 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
     let mut string_quote = NOT_IN_STRING;
     // Parentheses depth. `((x)` -> 1, `(x))` -> -1.
     let mut paren_depth: isize = 0;
+    // A state that indicates whether we should delete quasi and expression.
+    let mut has_remove_sentinel = false;
 
     // Current minified quasi being built
     let mut output = Vec::new();
@@ -890,8 +896,7 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
     // TODO: What about `cooked`? Shouldn't we alter that too?
     let mut quasi_index = 0;
     while quasi_index < quasis.len() {
-        let quasi = &quasis[quasi_index];
-        let mut bytes = quasi.value.raw.as_str().as_bytes();
+        let mut bytes = quasis[quasi_index].value.raw.as_str().as_bytes();
 
         if quasi_index > 0 {
             if let Some(is_block_comment) = comment_type {
@@ -909,12 +914,12 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
                 // styled.div`color: /* ${c} */ blue`
                 // //         ^^^^^^^^^^^^^^^^^^^^^^
                 // ```
-                quasis[quasi_index - 1].span.end = quasi.span.end;
+                quasis[quasi_index].span.start = quasis[quasi_index - 1].span.start;
+                // Mark previous quasi's span as `REMOVE_SENTINEL`, so that it can be removed after all quasis are processed.
+                quasis[quasi_index - 1].span = REMOVE_SENTINEL;
+                quasi_index += 1;
 
-                // Remove this quasi and the preceding expression.
-                // TODO: Remove scopes, symbols, and references for removed `Expression`.
-                quasis.remove(quasi_index);
-                expressions.remove(quasi_index - 1);
+                has_remove_sentinel = true;
 
                 // Find end of comment
                 let start_index = if is_block_comment {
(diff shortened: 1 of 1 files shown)


Dunqing:
* **#12256** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12240** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12224** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12256).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -911,11 +914,10 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
...
                 quasis[quasi_index - 1].span.end = quasi.span.end;
-
-                // Remove this quasi and the preceding expression.
-                // TODO: Remove scopes, symbols, and references for removed `Expression`.
-                quasis.remove(quasi_index);
-                expressions.remove(quasi_index - 1);
+                // Mark previous quasi for deletion, it will be deleted all of quasis are processed.
+                quasis[quasi_index - 1].value.raw = QUASI_DELETION_MARKER;
+                has_deletion_marker = true;
+                quasi_index += 1;
```
overlookmotel: String comparisons are somewhat expensive. Could we use a different marker for "remove this"? How about `quasis[quasi_index - 1].span = Span::new(u32::MAX, u32::MAX)`?

Source text is limited to max `u32::MAX` bytes, so it's impossible for a `TemplateElement` to have that span, because it's always followed by (at minimum) a `` ` ``. So it's a safe value to use as a sentinel.

overlookmotel, line 920: All branches now increment `quasi_index`, so we can remove all 3 x `quasi_index += 1;` statements and instead make the main loop:

```rs
for quasi_index in 0..quasis.len() {
  // ...
}
```

overlookmotel (commented):
As you say, this case is uncommon, but yes using `retain` is ideal to avoid poor perf in odd cases where there are tons of quasis and many get deleted. 👍

If we do #12213, then probably we'll be able to make the main loop use `retain`, and then we won't need the extra loops to remove nodes at the end.

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -911,11 +914,10 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
                 // //         ^^^^^^^^^^^^^^^^^^^^^^
                 // ```
                 quasis[quasi_index - 1].span.end = quasi.span.end;
-
-                // Remove this quasi and the preceding expression.
-                // TODO: Remove scopes, symbols, and references for removed `Expression`.
-                quasis.remove(quasi_index);
-                expressions.remove(quasi_index - 1);
+                // Mark previous quasi for deletion, it will be deleted all of quasis are processed.
+                quasis[quasi_index - 1].value.raw = QUASI_DELETION_MARKER;
```
overlookmotel: It's *this* quasi which should be deleted, not the previous one.

crates/oxc_transformer/src/plugins/styled_components.rs
```diff
@@ -911,11 +914,10 @@ fn minify_template_literal<'a>(lit: &mut TemplateLiteral<'a>, ast: AstBuilder<'a
                 // //         ^^^^^^^^^^^^^^^^^^^^^^
                 // ```
                 quasis[quasi_index - 1].span.end = quasi.span.end;
-
-                // Remove this quasi and the preceding expression.
-                // TODO: Remove scopes, symbols, and references for removed `Expression`.
```
overlookmotel, line 916: This comment is still relevant. Don't delete me!

crates/oxc_transformer/src/plugins/styled_components.rs
overlookmotel: Weird that all tests still passing with this mistake. Don't the tests cover this? Or am I mistaken?

crates/oxc_transformer/src/plugins/styled_components.rs
Dunqing: I've tried a similar way to remove `quasi_index += 1` before you reviewing, however, this causes the diff to be hard to review, so I reverted the changes and prefer to leave it to a follow-up PR

Dunqing: Sounds good, I always follow the best performance way!

Dunqing: Sounds good, I always follow the best performance way!

overlookmotel: OK cool. Probably a good idea to define a const for the sentinel span so it's clear what's going on.

```rs
/// `Span` used as a sentinel indicating quasi should be removed.
/// Source text is limited to max `u32::MAX` bytes, so it's impossible for a `TemplateElement`
/// to have this span, because it's always followed by (at minimum) a '`'.
const REMOVE_SENTINEL: Span = Span::new(u32::MAX, u32::MAX);
```

crates/oxc_transformer/src/plugins/styled_components.rs
Dunqing: This is due to a shift in the removal strategy. Previously, the `quasi` would be removed from the `quasis`, preventing access from anywhere. However, now it's different, the `span` of `quasi` is marked as `REMOVE_SENTINEL`, so you still can access that `quais`, which causes dealing with assigning the new output to become difficult.

The new handling of the way is to mark the previous quasi as removed, so the new output will be assigned to the current `quasi` in:

https://github.com/oxc-project/oxc/blob/b20f2ab55bfcd2d7780d8c2e443bf1306c8fb8d8/crates/oxc_transformer/src/plugins/styled_components.rs#L957-L965

This may be hard to understand, but from both behaviours, there is no difference.

overlookmotel: I've got it now! I'll update the comments in a follow-on PR to reflect this change in the logic.

overlookmotel:
### Merge activity

* **Jul 15, 9:55 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jul 15, 9:55 AM UTC**: `overlookmotel` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jul 15, 10:04 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #12329, opened 2025-07-16
perf(linter): reduce iterations when collecting directories for nested configs

overlookmotel:
Small perf optimization.

Presumably all paths will usually have the same prefix e.g.:

* `/Users/bozo/projects/best-project-ever`
* `/Users/bozo/projects/best-project-ever/src/components`
* `/Users/bozo/projects/best-project-ever/src/server`

When finding directories to search for nested configs, stop walking up to parent directories when find a directory that's already in the `HashSet`. We know that dir's parent, grandparent, etc will already be in the set too, so no need to check them again.

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -445,7 +445,10 @@ impl LintRunner {
             while let Some(dir) = current {
                 // NOTE: Initial benchmarking showed that it was faster to iterate over the directories twice
                 // rather than constructing the configs in one iteration. It's worth re-benchmarking that though.
-                directories.insert(dir);
+                let inserted = directories.insert(dir);
+                if !inserted {
+                    break;
+                }
                 current = dir.parent();
             }
         }


overlookmotel:
* **#12329** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12329?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12329?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12329).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
I *assume* this'd be an improvement, but we don't have benchmarks for oxlint in CI. @camc314 How do you normally test changes like this to make sure they do have the desired effect?

overlookmotel:
@camc314 Just a nudge on this one. Only reason I questioned whether it'd be a perf improvement is because of the comments above the code this PR changes. But "logically" it seems quite obvious that avoiding pointlessly iterating more than you need to should be a gain.

camchenry (approved):
This looks correct to me, I don't believe this should result in any configuration differences. This will only break out of the loop if it's a path that we've already processed (and therefore already know its nearest config).

I ran a benchmark on this locally. I used the latest oxlint 1.9 and built this PR with `cargo build -p oxlint --release` and compared the two on my M1 laptop. The project used for testing is the https://github.com/microsoft/fluentui repository at `b5fa2f0ae70bc472c0603a3927ac68e65b702525`. My local copy replaces every `.eslintrc` file with an `.oxlintrc.json` file containing `{}`. In total, there are 260 `.oxlintrc.json` files at various levels of the hierarchy (according to `find . | rg oxlintrc | wc -l`), and there are about 14836 lintable JS/TS files. These were the results:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `oxlint-dev . --silent` | 404.0 ± 22.8 | 375.0 | 463.5 | 1.00 |
| `oxlint . --silent` | 444.2 ± 22.7 | 415.4 | 497.4 | 1.10 ± 0.08 |

On average, this PR is 10% faster than the current oxlint 1.9.0 version on the `fluentui` project (with eslintrc files replaced with oxlintrc). The fact that the minimum times are significantly different also gives me some confidence that this really does significantly reduce the work that we have to do in large monorepos with nested configs.

camchenry:
For comparison, I also tested on `microsoft/vscode` (without any changes, just using the code as-is) to see how it fairs on a project without any nested configs. The difference was not significant, depending on the order that I ran the benchmarks one or the other could be faster.

overlookmotel:
@camchenry Thanks very much for delving into this. Holy smokes! It had never occurred to me that a project might have 260 config files, or that this change would have anything more than a microscopic perf benefit.

overlookmotel:
### Merge activity

* **Jul 30, 10:05 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jul 30, 10:05 AM UTC**: `overlookmotel` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jul 30, 10:13 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #12385, opened 2025-07-18
test(coverage/typescript): Update `5xxx` error codes handling

leaysgur:
Part of #11582

### DIFF
--- a/tasks/coverage/src/typescript/mod.rs
+++ b/tasks/coverage/src/typescript/mod.rs
@@ -57,6 +57,8 @@ impl<T: Case> Suite<T> for TypeScriptSuite<T> {
             "importingExportingTypes.ts",
             // This is just a binary file
             "TransportStream.ts",
+            // Emits TS5052 (complains compilerOptions are invalid, which we do not support), also implies TS2564 but not emitted
+            "esDecorators-emitDecoratorMetadata.ts",
         ]
         .iter()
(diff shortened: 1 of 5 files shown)


tasks/coverage/src/typescript/mod.rs
```diff
@@ -57,6 +57,8 @@ impl<T: Case> Suite<T> for TypeScriptSuite<T> {
             "importingExportingTypes.ts",
             // This is just a binary file
             "TransportStream.ts",
+            // Emits TS5052 (complains compilerOptions are invalid, which we do not support), also implies TS2564 but not emitted
+            "esDecorators-emitDecoratorMetadata.ts",
```
leaysgur, line 61: FYI: 

- Test file: https://github.com/microsoft/TypeScript/blob/2778e84ed88894156e2c5a17d74a25c73cc96ffe/tests/cases/conformance/esDecorators/esDecorators-emitDecoratorMetadata.ts
- Error file: https://github.com/microsoft/TypeScript/blob/2778e84ed88894156e2c5a17d74a25c73cc96ffe/tests/baselines/reference/esDecorators-emitDecoratorMetadata(target%3Desnext).errors.txt

Error file only reports TS5052 for `compilerOptions`.

But this code should report error TS2564 without `compilerOptions`.

```ts
class C {
  // Property 'y' has no initializer and is not definitely assigned in the constructor.
  y: number;
}
```

PR #12446, opened 2025-07-22
fix(linter): `unicorn/catch-error-name` wasn't using the ignore property

### LINKED ISSUE #12430: linter: unicorn/catch-error-name

1.7.0

### What command did you run?

oxlint --max-warnings=0

```jsonc
{
  "rules": {
    "unicorn/catch-error-name": [
      "error",
      {
        "ignore": [
          "^error\\d*$"
        ]
      }
    ]
  }
}
```

### What happened?

oxlint complains with `eslint-plugin-unicorn(catch-error-name): The catch parameter "error2" should be named "error"` with code bellow
```typescript
try {
  // some codes
} catch (error2: unknown) {
  // some codes
}
```

this makes problem in nested try { }  catch { }
```typescript
try {
  // some codes
} catch (error: unknown) {
  try {
    // some codes
  } catch (error2: unknown) {
    // some codes
  }
}
```

unicorn in eslint auto fix this by appending _ to the second catch error name (https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/catch-error-name.md#tip) that still not works in oxlint and erros with ` eslint-plugin-unicorn(catch-error-name): The catch parameter "error_" should be named "error"`

imranbarbhuiya:
Allowed `regex` in `ignore` vec. Also changed `ignored` to `ignore` to match docs and unicorn config

fixes: #12430

the regex build will silently fail if its invalid, should I log it? if yes how to do it as using any print related function causes clippy to fail

### DIFF
--- a/crates/oxc_linter/src/rules/unicorn/catch_error_name.rs
+++ b/crates/oxc_linter/src/rules/unicorn/catch_error_name.rs
@@ -1,3 +1,4 @@
+use lazy_regex::Regex;
 use oxc_ast::{
     AstKind,
     ast::{Argument, BindingPatternKind, Expression},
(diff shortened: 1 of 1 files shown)


camc314 (approved):
thanks!

PR #12463, opened 2025-07-23
fix(transformer/top-level-statements): should not inject statements after non-import statement

### LINKED ISSUE #12460: transform: jsx refresh injects variable after call for HOCs
For the following input, Oxc injects `var _s = $RefreshSig$()` after `import { bar } from './bar'` even though `_s` is called on the right hand of `BazComponent =`.
```jsx
import { observer } from 'mobx-react-lite'
import { useFoo } from './useFoo'

export const BazComponent = observer(function BazComponent_() {
  const foo = useFoo()
  return (
    <>
      {foo}
      {bar}
    </>
  )
})

import { bar } from './bar'
```
[reproduction using transform function exposed from rolldown](https://stackblitz.com/edit/github-xv4yss-ervsp63v?file=package.json,build.mjs)
[swc playground](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D)

Dunqing:
* close https://github.com/oxc-project/oxc/issues/12460

Injecting top-level statements after the last import statement is unsound because once there is an import statement in the last line, it would be broken if other code relies on the injected statements. Learn more to see the following example.

I suppose the reason we did this is that we had a test case that misled us. I overrode that test with the expected output.

For example: 

Input:

```js
import { observer } from 'mobx-react-lite'
import { useFoo } from './useFoo'

export const BazComponent = observer(function BazComponent_() {
  const foo = useFoo()
  return (
    <>
      {foo}
      {bar}
    </>
  )
})

import { bar } from './bar'
```

Before output:
```js
import { observer } from "mobx-react-lite";
import { useFoo } from "./useFoo";
export const BazComponent = _s(observer(_c = _s(function BazComponent_() {
        _s();
        const foo = useFoo();
        return /* @__PURE__ */ _jsxs(_Fragment, { children: [foo, bar] });
}, "useFoo{foo}", false, function() {
        return [useFoo];
})), "useFoo{foo}", false, function() {
        return [useFoo];
});
_c2 = BazComponent;
import { bar } from "./bar";
import { Fragment as _Fragment, jsxs as _jsxs } from "react/jsx-runtime";
var _s = $RefreshSig$();
var _c, _c2;
$RefreshReg$(_c, "BazComponent$observer");
$RefreshReg$(_c2, "BazComponent");

```

After output:

```js
import { observer } from "mobx-react-lite";
import { useFoo } from "./useFoo";
import { Fragment as _Fragment, jsxs as _jsxs } from "react/jsx-runtime";
var _s = $RefreshSig$();
export const BazComponent = _s(observer(_c = _s(function BazComponent_() {
        _s();
        const foo = useFoo();
        return /* @__PURE__ */ _jsxs(_Fragment, { children: [foo, bar] });
}, "useFoo{foo}", false, function() {
        return [useFoo];
})), "useFoo{foo}", false, function() {
        return [useFoo];
});
_c2 = BazComponent;
import { bar } from "./bar";
var _c, _c2;
$RefreshReg$(_c, "BazComponent$observer");
$RefreshReg$(_c2, "BazComponent");
```

The difference is that `import { Fragment as _Fragment, jsxs as _jsxs } from "react/jsx-runtime"; var _s = $RefreshSig$();` has injected after `import { useFoo } from "./useFoo";`

### DIFF
--- a/crates/oxc_transformer/src/common/top_level_statements.rs
+++ b/crates/oxc_transformer/src/common/top_level_statements.rs
@@ -74,12 +74,12 @@ impl<'a> TopLevelStatementsStore<'a> {
             return;
         }
 
-        // Insert statements after any existing `import` statements
+        // Insert statements before the first non-import statement.
         let index = program
             .body
             .iter()
-            .rposition(|stmt| matches!(stmt, Statement::ImportDeclaration(_)))
-            .map_or(0, |i| i + 1);
+            .position(|stmt| !matches!(stmt, Statement::ImportDeclaration(_)))
+            .unwrap_or(program.body.len());
 
         program.body.splice(index..index, stmts.drain(..));
     }
--- a/tasks/coverage/snapshots/semantic_typescript.snap
+++ b/tasks/coverage/snapshots/semantic_typescript.snap
@@ -19242,7 +19242,7 @@ rebuilt        : []
 semantic Error: tasks/coverage/typescript/tests/cases/compiler/jsxPartialSpread.tsx
 Symbol reference IDs mismatch for "Select":
 after transform: SymbolId(0): [ReferenceId(2), ReferenceId(3)]
-rebuilt        : SymbolId(0): [ReferenceId(3)]
+rebuilt        : SymbolId(3): [ReferenceId(3)]
 Unresolved references mismatch:
 after transform: ["Parameters", "Partial"]
 rebuilt        : []
--- a/tasks/transform_conformance/overrides/babel-plugin-transform-react-jsx/test/fixtures/autoImport/after-polyfills-2/output.mjs
+++ b/tasks/transform_conformance/overrides/babel-plugin-transform-react-jsx/test/fixtures/autoImport/after-polyfills-2/output.mjs
@@ -1,0 +1,12 @@
+// The reason to override this test case:
+// We generate the import statements before the first non-import statement,
+// while the original test case expects them to be after all import statements.
+
+import { jsx as _jsx } from "react/jsx-runtime";
+ReactDOM.render(
+	/* @__PURE__ */ _jsx("p", { children: "Hello, World!" }),
+	document.getElementById("root"),
+);
+import "react-app-polyfill/ie11";
+import "react-app-polyfill/stable";
+import ReactDOM from "react-dom";
--- a/tasks/transform_conformance/snapshots/oxc.snap.md
+++ b/tasks/transform_conformance/snapshots/oxc.snap.md
@@ -1,6 +1,6 @@
 commit: 1d4546bc
 
-Passed: 180/299
+Passed: 180/300
 
 # All Passed:
 * babel-plugin-transform-class-static-block
@@ -502,7 +502,17 @@ after transform: [ReferenceId(0), ReferenceId(1), ReferenceId(4), ReferenceId(9)
 rebuilt        : [ReferenceId(5)]
 
 
-# babel-plugin-transform-react-jsx (44/46)
+# babel-plugin-transform-react-jsx (44/47)
+* refresh/import-after-component/input.js
+Missing ScopeId
+Missing ReferenceId: "useFoo"
+Scope children mismatch:
+after transform: ScopeId(0): [ScopeId(1), ScopeId(2)]
+rebuilt        : ScopeId(0): [ScopeId(1), ScopeId(2), ScopeId(3)]
+Symbol reference IDs mismatch for "useFoo":
+after transform: SymbolId(1): [ReferenceId(1), ReferenceId(7)]
+rebuilt        : SymbolId(1): [ReferenceId(6), ReferenceId(11), ReferenceId(12)]
+
 * refresh/react-refresh/includes-custom-hooks-into-the-signatures-when-commonjs-target-is-used/input.jsx
 x Output mismatch
 
--- a/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/refresh/import-after-component/input.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/refresh/import-after-component/input.js
@@ -1,0 +1,14 @@
+import { observer } from "mobx-react-lite";
+import { useFoo } from "./useFoo";
+
+export const BazComponent = observer(function BazComponent_() {
+	const foo = useFoo();
+	return (
+		<>
+			{foo}
+			{bar}
+		</>
+	);
+});
+
+import { bar } from "./bar";
--- a/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/refresh/import-after-component/output.js
+++ b/tasks/transform_conformance/tests/babel-plugin-transform-react-jsx/test/fixtures/refresh/import-after-component/output.js
@@ -1,0 +1,21 @@
+import { observer } from "mobx-react-lite";
+import { useFoo } from "./useFoo";
+import { Fragment as _Fragment, jsxs as _jsxs } from "react/jsx-runtime";
+var _s = $RefreshSig$();
+
+export const BazComponent = _s(observer(_c = _s(function BazComponent_() {
+  _s();
+  const foo = useFoo();
+  return /* @__PURE__ */ _jsxs(_Fragment, { children: [foo, bar] });
+}, "useFoo{foo}", false, function() {
+  return [useFoo];
+})), "useFoo{foo}", false, function() {
+  return [useFoo];
+});
+_c2 = BazComponent;
+
+import { bar } from "./bar";
+
+var _c, _c2;
+$RefreshReg$(_c, "BazComponent$observer");
+$RefreshReg$(_c2, "BazComponent");


Dunqing:
* **#12463** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12463?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12463?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12463).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
I'm not entirely sure about this.

### Polyfills

The test case that's overridden here does have a rationale for why it is as it is. I don't think it's misled us exactly.

It links to this Babel issue: https://github.com/babel/babel/issues/12522

If an import low down in the file is a polyfill, the polyfill should execute before `import ... from "react/jsx-runtime";`, so that the polyfill is applied before `react` is loaded. React may need some APIs which the polyfill provides.

So this PR as it is fixes one problem, but may create another.

Strictly speaking, I think the correct behavior should be to treat `import` statements and other statements differently:

* `import` statements are added after last existing `import` statement.
* All other statements are added at very top of file (or after initial block of `import`s at top of file, if we prefer).

Or:

* Move all existing `import` statements up to top of file (maintaining original order).
* Insert new statements (whatever they are) after all the `import`s.

The latter might make for more performant code, because browser will find all the `import`s quickly and can start fetching them before it's finished parsing the rest of the file. But bundlers presumably already do this, so maybe it's not the job of transformer to perform such optimizations. But if it's easier for us to do it like that, I don't think it'll hurt.

### CommonJS

We're probably not handling CommonJS correctly either. Presumably any `require`s we add should be after existing `require`s. I'm not sure if that's tractable though, because `require` can appear anywhere. From the Babel issue, it sounds like they decided it's not, and ignore that problem.

### What to do?

I can see 2 options here:

1. Try one of the suggestions above (or something else that has same effect).
2. Decide that polyfills are a thing of the past, and that we can safely ignore this problem! In which case this PR is good to go.

@Dunqing what do you think?

Boshen:
### Merge activity

* **Jul 24, 12:49 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jul 24, 12:49 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jul 24, 1:06 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

Dunqing:
> ### What to do?
> I can see 2 options here:
> 
> 1. Try one of the suggestions above (or something else that has same effect).
> 2. Decide that polyfills are a thing of the past, and that we can safely ignore this problem! In which case this PR is good to go.
> 
> @Dunqing what do you think?

Wow, thank you for a deeper understanding of this.

First of off, I need to point out that the example code of your linked issue is not the same as the test being overridden.

**Example of https://github.com/babel/babel/issues/12522:**
``` js
import 'react-app-polyfill/ie11';
import 'react-app-polyfill/stable';
import ReactDOM from 'react-dom';

ReactDOM.render(
    <p>Hello, World!</p>,
    document.getElementById('root')
);
```

**The overridden test:**
```js
// https://github.com/babel/babel/issues/12522

ReactDOM.render(
    <p>Hello, World!</p>,
    document.getElementById('root')
);

// Imports are hoisted, so this is still ok
import 'react-app-polyfill/ie11';
import 'react-app-polyfill/stable';
import ReactDOM from 'react-dom';
```

There is an essential difference between them! The first case we can handle well in terms of this PR solution. But the second one, yes, we will meet a problem with that, but in my opinion, users always know the polyfill should always be put on the first line, so why I am saying that I prefer to do nothing for now.

Putting this case aside, I think the first option is nice to do, but I prefer option(2) for now. The following is what I thought on this:

1. As I said above, polyfill should always be put on the first line is common practice, so the current logic can handle this well.
2. Normally, imports always come before non-import statements, so this change has no impact whatsoever and accounts for even worst-case scenarios (issue this PR linked).
3. We always have a workaround for the problem, see https://github.com/oxc-project/oxc/issues/12460#issuecomment-3105613349.
4. We have the same output as [SWC](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D). Since SWC is a mature transformer, to some extent, we follow the same standard.

Anyway, don't get me wrong. I don't disagree with you!

overlookmotel:
> First of off, I need to point out that the example code of your linked issue is not the same as the test being overridden.

True. But the test case *does* relate to that issue. The babel authors amended the original test case to cover the hardest case where polyfills are imported later in the file.

Yes, it's common practice to put imports of polyfills at the top of the file. So this import order problem only occurs in an edge case where polyfills are lower down, and as you say there's a workaround. But still... this edge case is legal, and I think in an ideal world, we'd handle it correctly like Babel does.

> 4. We have the same output as [SWC](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D). Since SWC is a mature transformer, to some extent, we follow the same standard.

You are right that SWC does not handle it correctly in the presence of polyfills: [SWC playground](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D).

But personally I don't think we should follow SWC by default. We know SWC takes shortcuts and has edge case bugs in many places, and I've heard people complain about that. So I think we should be aiming to be *more* correct than SWC. In particular, some big companies care a great deal about reliability and correctness, and it's a selling point of Oxc over SWC.

So... TLDR:

In practice, the polyfill problem is going to be rare. So I agree with your arguments. I think it was right to merge this PR - it solves more problems than it creates.

I'll create a backlog issue about the polyfill edge case. We may want to address it further down the line when we have time.

Dunqing:
> > First of off, I need to point out that the example code of your linked issue is not the same as the test being overridden.
> 
> True. But the test case _does_ relate to that issue. The babel authors amended the original test case to cover the hardest case where polyfills are imported later in the file.
> 
> Yes, it's common practice to put imports of polyfills at the top of the file. So this import order problem only occurs in an edge case where polyfills are lower down, and as you say there's a workaround. But still... this edge case is legal, and I think in an ideal world, we'd handle it correctly like Babel does.
> 
> > 4. We have the same output as [SWC](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D). Since SWC is a mature transformer, to some extent, we follow the same standard.
> 
> You are right that SWC does not handle it correctly in the presence of polyfills: [SWC playground](https://play.swc.rs/?version=1.12.3&code=H4sIAAAAAAAAA0vOzysuUXD29%2FEPig9w9HENCXFVsFWIjrXmSgbLBLk6OofEO%2Fv7Bvj7ufqFAOU0NBVs7RSqa60B39m%2B7DsAAAA%3D&config=H4sIAAAAAAAAA1VQSQ6DMAy88wrkM9f20D%2F0EVFwaFA22aEqQvy9IRAIt3gWzzhL07YwsoRXu6RnGoIgRjrnhPDsovglBFBawZJ0iNAVduSNijRhRtadgChowJhN%2FDjUYLxnTJgShrEISThWnmwdSShkrIAE9fhF44NFF4%2B87iIJFSF%2F6iKpyq2Q1U6ruQ6R3obk4nuhTSrcYM6eTdmVN4H1%2FZTJ47%2FiHHA%2F8wmXqISdi0Hzuzhzx%2FUP271TS3sBAAA%3D).
> 
> But personally I don't think we should follow SWC by default. We know SWC takes shortcuts and has edge case bugs in many places, and I've heard people complain about that. So I think we should be aiming to be _more_ correct than SWC. In particular, some big companies care a great deal about reliability and correctness, and it's a selling point of Oxc over SWC.
> 
> So... TLDR:
> 
> In practice, the polyfill problem is going to be rare. So I agree with your arguments. I think it was right to merge this PR - it solves more problems than it creates.
> 
> I'll create a backlog issue about the polyfill edge case. We may want to address it further down the line when we have time.

Thanks!

overlookmotel:
https://github.com/oxc-project/backlog/issues/169

Issue #12437, opened 2025-07-21
Linter plugins: Various changes


overlookmotel:
A short "shopping list" of potential changes/optimizations to JS linter plugins.

The first is a potential edge case bug. The rest are not urgent or important, but I just wanted to note them down somewhere.

* [ ] `ExternalPluginStore` should not contain `plugin_names`.
  * Plugin name to plugin ID should be resolved on a per-config basis
    (because 2 plugins can have same name but be different plugins).
  * Configs should store plugin IDs.
* [ ] Remove `registered_plugin_paths` from `ExternalPluginStore`. It's only needed while building configs, but never used again after that stage.
* [ ] `external_rules_for_override` in `OxlintRules::override_rules` could be a `FixedBitSet`?
  There probably aren't that many external rules, so `FxHashMap<ExternalRuleId, AllowWarnDeny>` is inefficient.
* [ ] Use only 1 `Resolver` for all nested configs - creating a `Resolver` is not cheap, and if there's only one, they'll all use the same cache.
* [x] When JS plugin creates an error, code to get severity for rule is not optimal:
  * It has to search through `external_rules` to find the entry with `RuleId`
  * #12482
* [x] JS class for `Context`. #12440
* [x] Recycle `Context` objects. #12475
* [x] Add `id` property (rule name) to `Context`. #12480
* [x] Improve perf of merging multiple visitors when many rules. #12472

PR #12481, opened 2025-07-23
feat(napi/oxlint): add `getFilename` + `getPhysicalFilename` methods to `Context`

overlookmotel:
Add `getFilename` and `getPhysicalFilename` methods to `Context`, to match ESLint's API.

PR #12482, opened 2025-07-23
refactor(napi/oxlint): diagnostics communicate which rule via rule index, not rule ID

overlookmotel:
Related to #12437.

Diagnostics sent from JS to Rust specify which rule the diagnostic relates to via an index into `external_rules`, instead of the `ExternalRuleId`. This allows removing the loop over `external_rules` on Rust side to find the relevant rule.

### DIFF
--- a/crates/oxc_linter/src/external_plugin_store.rs
+++ b/crates/oxc_linter/src/external_plugin_store.rs
@@ -79,11 +79,13 @@ impl ExternalPluginStore {
         })
     }
 
-    pub fn resolve_plugin_rule_names(&self, external_rule_id: u32) -> Option<(&str, &str)> {
-        let external_rule = self.rules.get(ExternalRuleId::from_raw(external_rule_id))?;
+    pub fn resolve_plugin_rule_names(
+        &self,
+        external_rule_id: ExternalRuleId,
+    ) -> (/* plugin name */ &str, /* rule name */ &str) {
+        let external_rule = &self.rules[external_rule_id];
         let plugin = &self.plugins[external_rule.plugin_id];
-
-        Some((&plugin.name, &external_rule.name))
+        (&plugin.name, &external_rule.name)
     }
 }
 
--- a/napi/oxlint2/src-js/index.js
+++ b/napi/oxlint2/src-js/index.js
@@ -51,18 +51,18 @@ async function loadPluginImpl(path) {
 
   // TODO: Use a validation library to assert the shape of the plugin, and of rules
   const pluginName = plugin.meta.name;
-  let ruleId = registeredRules.length;
+  const offset = registeredRules.length;
   const ruleNames = [];
-  const ret = { name: pluginName, offset: ruleId, ruleNames };
 
   for (const [ruleName, rule] of Object.entries(plugin.rules)) {
     ruleNames.push(ruleName);
-    registeredRules.push({ rule, context: new Context(ruleId, `${pluginName}/${ruleName}`) });
-    ruleId++;
+    registeredRules.push({ rule, context: new Context(`${pluginName}/${ruleName}`) });
   }
 
-  return JSON.stringify({ Success: ret });
+  return JSON.stringify({ Success: { name: pluginName, offset, ruleNames } });
 }
+
+let setupContextForFile;
 
 /**
  * Context class.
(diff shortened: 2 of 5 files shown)


overlookmotel:
* **#12481** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12497** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12482** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12482).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_linter/src/external_plugin_store.rs
```diff
@@ -79,8 +79,11 @@ impl ExternalPluginStore {
         })
     }
 
-    pub fn resolve_plugin_rule_names(&self, external_rule_id: u32) -> Option<(&str, &str)> {
-        let external_rule = self.rules.get(ExternalRuleId::from_raw(external_rule_id))?;
+    pub fn resolve_plugin_rule_names(
+        &self,
+        external_rule_id: ExternalRuleId,
+    ) -> Option<(&str, &str)> {
+        let external_rule = self.rules.get(external_rule_id)?;
```
camc314: if the argument here is `ExternalRuleId`, I don't think we need to use `.get` since we're indexing into the index vec with the correct key

camc314: the reason this returned an option originally is because you could pass in an arbitary `u32` value which might not be in the correct range.

(i guess you could pass in an arbitary value for `ExternalRuleId`, but it's a lot less likley

napi/oxlint2/src-js/index.js
```diff
@@ -51,41 +51,39 @@ async function loadPluginImpl(path) {
...
-  let ruleId = registeredRules.length;
+  const offset = registeredRules.length;
   const ruleNames = [];
-  const ret = { name: pluginName, offset: ruleId, ruleNames };
 
   for (const [ruleName, rule] of Object.entries(plugin.rules)) {
     ruleNames.push(ruleName);
-    registeredRules.push({ rule, context: new Context(ruleId, `${pluginName}/${ruleName}`) });
-    ruleId++;
+    registeredRules.push({ rule, context: new Context(`${pluginName}/${ruleName}`) });
```
camc314, line 59: unrelated, but just noticed, should we re-use the string for the `pluginName/ruleName` rather than creating it? might need benching

napi/oxlint2/src-js/index.js
```diff
@@ -123,6 +121,14 @@ class Context {
...
     return this.physicalFilename;
   }
+
+  static {
+    setupContextForFile = (context, ruleIndex, filePath) => {
+      context.#ruleIndex = ruleIndex;
+      context.filename = filePath;
+      context.physicalFilename = filePath;
+    };
+  }
```
camc314: out of interest is static faster than a non-static method directly on the class??

camc314: lgtm

overlookmotel: Making `setupContextForFile` a normal class method would be simpler, but that'd expose the method to user, which we don't want to do.

We have to have this function inside the class body because it's not possible to set `#ruleIndex` private property from outside the class.

I'll add a comment to clarify the rationale here.

napi/oxlint2/src-js/index.js
overlookmotel: How do you mean? This is when rule is registered, which only happens once. So I think each string is only created once.

We could do this:

```js
const pluginNameWithSlash = `${pluginName}/`;
for (const [ruleName, rule] of Object.entries(plugin.rules)) {
  ruleNames.push(ruleName);
  registeredRules.push({ rule, context: new Context(`${pluginNameWithSlash}${ruleName}`) });
}
```

But I'm not sure if that'd be a perf improvement or not. If it is, I doubt the difference would be measurable.

camc314: probably won't have a perf diff, but i more meant changing `plugin.rules shape to be like
```
type Rules = { [ruleName: string]: { fullName: `${pluginName}/${ruleNme}`, rule: any } }
```

so we only construct `fullName` once. not important to optimize right now though.

Currently we construct `${pluginNameWithSlash}${ruleName}`` num files * num rules times, but we could just change it to construct a max of `num rules` times.

crates/oxc_linter/src/external_plugin_store.rs
overlookmotel: Good point. While there's no real difference between `u32` and `ExternalRuleId`, the ID now comes from within Rust code, rather than being passed from JS, so we can have more confidence that's it's valid.

I've pushed a commit to make `resolve_plugin_rule_names` infallible.

napi/oxlint2/src-js/index.js
overlookmotel: But that's what I'm saying. We only construct it `num rules` times. This code is in `loadPlugin`, not `lintFile`.

PR #12544, opened 2025-07-27
fix(vscode): fix statusbar icon order

fehmer:
Change the order of the status bar to be icon first then text.

I don't know if there is a reason behind this or an oversight. Other plugins like typescript and prettier show the icon on the left and the text on the right:
image


Sysix:
thank you!

Issue #11045, opened 2025-05-14
parser: Illegal instruction in aarch64


Arlanthir:
I'm trying to run oxc-parser in an `aarch64` machine, running on Amazon EC2 (with ubuntu).

It fails with the error:
```bash
Illegal instruction (core dumped)
```

The following minimal code produces the error:

package.json
```json
{
  "name": "testoxc",
  "type": "module",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "main": "node index.js"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "oxc-parser": "^0.69.0"
  }
}
```

index.js (copied from the npm package instructions):
```js
import oxc from 'oxc-parser';

const code = 'const url: String = /* 🤨 */ import.meta.url;';

// File extension is used to determine which dialect to parse source as.
const filename = 'test.tsx';

const result = oxc.parseSync(filename, code);
// or `await oxc.parseAsync(filename, code)`

// An array of errors, if any.
console.log(result.errors);

// AST and comments.
console.log(result.program, result.comments);

// ESM information - imports, exports, `import.meta`s.
console.log(result.module);
```

It resolves and installs `node_modules/@oxc-parser/binding-linux-arm64-gnu`.

The same code works fine on my arm64 (macOS) machine (via `binding-darwin-arm64`).

Testing various versions, I discovered that it stopped working in version `0.60.0` (it worked with `0.58.1`.

Arlanthir:
This is still reproducible in 0.70.0, in case it was supposed to be fixed.

Brooooooklyn:
@Arlanthir can you run `sudo lscpu` on your instance and give us back the output information?

Arlanthir:
> [@Arlanthir](https://github.com/Arlanthir) can you run `sudo lscpu` on your instance and give us back the output information?

Of course, here you go:

```bash
Architecture:             aarch64
  CPU op-mode(s):         32-bit, 64-bit
  Byte Order:             Little Endian
CPU(s):                   2
  On-line CPU(s) list:    0,1
Vendor ID:                ARM
  Model name:             Cortex-A72
    Model:                3
    Thread(s) per core:   1
    Core(s) per socket:   2
    Socket(s):            1
    Stepping:             r0p3
    BogoMIPS:             166.66
    Flags:                fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
Caches (sum of all):
[...]

PR #12555, opened 2025-07-28
test(parser): track number of allocations

camchenry:
- related: https://github.com/oxc-project/backlog/issues/5

This PR adds a reliable method of tracking the number of allocations made while running the parser. This includes both arena allocations (via a new feature), as well as system allocations (by creating a new global allocator). Reductions in these numbers should correlate to real-world performance improvements that aren't quantified in CodSpeed currently. This will also help prevent regressions where we accidentally allocate lots more memory than we expect.

Originally I had included total system bytes allocated, but it was tricky to get this number to match exactly between platforms. I opted not try and make this perfect but instead use the total number of allocations which is a good proxy for bytes anyway.


camchenry:
* **#12555** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12555?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#12701](https://github.com/oxc-project/oxc/pull/12701) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12701?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#12837](https://github.com/oxc-project/oxc/pull/12837) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12837?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12555).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_allocator/src/allocator.rs
```diff
@@ -571,6 +597,12 @@ impl Allocator {
         bytes
     }
 
+    /// Returns the total number of allocations that have been made in this [`Allocator`] instance.
+    #[cfg(feature = "track_allocations")]
+    pub fn num_alloc(&self) -> usize {
+        NUM_ALLOC.load(Relaxed)
```
overlookmotel: I've never entirely got my head around atomic ordering, but I think `Relaxed` here may not be sufficient. You want to ensure all the increments have been committed before reading the total. For the same reason, I'm not sure if `Relaxed` where the increments occurs needs to be something stronger.

Or maybe sleeping for a short time before loading the total count would work, to allow everything to "catch up" first.

overlookmotel: Actually, maybe this doesn't matter here because parser always runs on a single thread anyway. But it might matter more on the `#[global_allocator]` tracker.

camchenry: I found this to be a good overview: https://doc.rust-lang.org/nomicon/atomics.html. I think I might change this to be sequentially consistent (`SeqCst`), just to be safe. Not that I really think it matters too much, but I stole this code from the docs mostly anyway and I think having it be thread-safe by default is a good idea.

tasks/allocs/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    let mut allocator = Allocator::default();
+
+    let options = ParseOptions { parse_regular_expression: true, ..ParseOptions::default() };
+
+    // Warm-up by parsing each file first, and then measuring the actual allocations.
+    // TODO: For some reason, this causes some variance in the total system allocations, specifically
+    // for checker.ts and antd.js. I tried to track this down but wasn't able to figure out why.
+    for file in files.files() {
+        Parser::new(&allocator, &file.source_text, file.source_type).with_options(options).parse();
+    }
```
overlookmotel: Interesting! Could it be that without the warm-up, the arena `Allocator` will grow multiple times while parsing, and each time it grows, that's a call to system allocator?

The fixtures run in this order:

https://github.com/oxc-project/oxc/blob/e836e552f347c9d8f6beef7550e2f952dfc0b5e4/tasks/common/src/test_file.rs#L61-L76

`checker.ts` and `antd.js` are the largest files, so `Allocator` would need to grow for `checker.ts`, but then the rest are probably small enough to fit into the `Allocator` without resizing it, except for `antd.js`, which requires growing the `Allocator` again.

I switched `antd.js` to be first in the `complicated_urls` list and sure enough, even with the warm-up commented out, the number of system allocations stayed the same for `checker.ts`. This seems to confirm `Allocator` growth is the cause.

There's no problem with doing a warm-up round, but if you agree this is the likely cause, it'd probably be worth updating the comment.

tasks/allocs/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        unsafe {
+            mimalloc_safe::MiMalloc.dealloc(ptr, layout);
+        }
+    }
+
+    unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+        unsafe { mimalloc_safe::MiMalloc.alloc_zeroed(layout) }
+    }
```
overlookmotel: Increment `NUM_ALLOC` here too?

tasks/allocs/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
+        unsafe { mimalloc_safe::MiMalloc.realloc(ptr, layout, new_size) }
+    }
```
overlookmotel: Increment `NUM_ALLOC` here too?

It might even be worthwhile having a separate `NUM_REALLOC` counter, because `realloc` is particularly expensive. If the allocation can't grow in place, `realloc` results in copying all the bytes from old allocation into the new allocation. If the allocation is large, that's *really* costly.

This would catch when we haven't pre-allocated enough space in a `Vec` for its contents and the `Vec` has to grow, for example.

You can check if the reallocation occurred in place (cheap) or required memcopy (expensive) by comparing the old `ptr` with the new pointer returned by `realloc` - if they're the same, reallocation occurred in place, if not `layout.size()` bytes had to be copied. *But...* that behavior may well be non-deterministic, or at least differ across platforms.

overlookmotel (commented):
This is great.

In addition to the comments below:

There are also many arena allocations that occur without going through the `Allocator` APIs. Notably, `Vec` performs allocations and reallocations directly via `Bump`, by going through the `Alloc` trait (`alloc.alloc()` and `alloc.grow()` calls).

https://github.com/oxc-project/oxc/blob/5f50bc3126b3353c4a284a8a8049165b267bed40/crates/oxc_allocator/src/vec2/raw_vec.rs#L93-L115

https://github.com/oxc-project/oxc/blob/5f50bc3126b3353c4a284a8a8049165b267bed40/crates/oxc_allocator/src/vec2/raw_vec.rs#L742-L766

These allocations / reallocations aren't getting counted, and they're a significant proportion of arena allocations.

So the ideal would be to increment `num_alloc` in `Alloc::alloc` + `Alloc::grow` as well. But problem is that in those methods you only have access to the `Bump`, not the `Allocator`.

A terrible hack would be:

```rs
impl Alloc for Bump {
    #[inline(always)]
    fn alloc(&self, layout: Layout) -> NonNull<u8> {
        #[cfg(feature = "track_allocations")]
        unsafe {
            use crate::Allocator;
            use std::{
                mem::offset_of,
                ptr,
                sync::atomic::{AtomicUsize, Ordering},
            };
            #[expect(clippy::cast_possible_wrap)]
            const OFFSET: isize = (offset_of!(Allocator, num_alloc) as isize)
                - (offset_of!(Allocator, bump) as isize);
            let num_alloc_ptr = ptr::from_ref(self).byte_offset(OFFSET).cast::<AtomicUsize>();
            let num_alloc = num_alloc_ptr.as_ref().unwrap_unchecked();
            num_alloc.fetch_add(1, Ordering::SeqCst);
        }

        self.alloc_layout(layout)
    }
}
```

This is completely unsound! It relies on `Bump` always being wrapped inside an `Allocator`, which can't be statically proven. But in practice we know that's always the case in Oxc - we never use `Bump` on its own. This unsoundness would be unacceptable in production code, but as it's behind a cargo feature which is only used in this internal tool, I think it'd be OK.

When I finally finish up replacing `Bump` with our own custom allocator ([WIP](https://github.com/oxc-project/oxc/tree/05-16-feat_allocator_replace_bumpalo_)) we could build allocation counting into it, and remove this hack.

---

Note: Counting the number and total byte size of arena *reallocations* separately would be a really good measure. Allocation in arena is very cheap, but reallocation is *way* more expensive (see comment below). Of the stats for arena allocations, that's one of the measures which should be a priority to reduce.

overlookmotel:
> Originally I had included total system bytes allocated, but it was tricky to get this number to match exactly between platforms.

Did you check that this is still the case once you added the warmup? If it still differs across platforms, I'd be interested to know by how much, and on what platforms. And is there also variance on the *same* platform across multiple runs?

I'd have broadly expected this measure to be deterministic. If it's not, the reason why might also shed some light on some of the variance we have in our benchmarks. It seems to relate to non-deterministic allocation patterns in `HashMap`s, but I've never understood where that's coming from.

tasks/allocs/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+// is the same across different platforms for the purposes of tracking allocations.
+#[expect(unsafe_code, clippy::undocumented_unsafe_blocks)]
+unsafe impl GlobalAlloc for TrackedAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ret = unsafe { mimalloc_safe::MiMalloc.alloc(layout) };
+        if !ret.is_null() {
+            NUM_ALLOC.fetch_add(1, SeqCst);
+        }
+        ret
+    }
```
overlookmotel: Is mimalloc workable on all platforms we expect people to develop Oxc on? I seem to remember seeing elsewhere in codebase that we gate its use on some platforms - though I'm not sure why.

Might be simpler just to delegate to `System` allocator, rather than `mimalloc`?

tasks/allocs/src/lib.rs
camchenry: Ah now that you mention this, I feel like this must be it. That matches exactly the behavior I was seeing. Maybe it would be best to just do a very large allocation up front? I think I'll leave it this way for now though, as it should ensure it works as expected even if we change the files later.

tasks/allocs/src/lib.rs
camchenry: Yeah I think treating this separately feels right to me. Since the whole point of this is to catch subtle allocations and costly operations, this feels like the right thing to do.



> You can check if the reallocation occurred in place (cheap) or required memcopy (expensive) by comparing the old `ptr` with the new pointer returned by `realloc` - if they're the same, reallocation occurred in place, if not `layout.size()` bytes had to be copied. _But..._ that behavior may well be non-deterministic, or at least differ across platforms.

I'll give this a try at least. If nothing else, the separate realloc count should be a good start.

camchenry:
> > Originally I had included total system bytes allocated, but it was tricky to get this number to match exactly between platforms.
> 
> Did you check that this is still the case once you added the warmup? If it still differs across platforms, I'd be interested to know by how much, and on what platforms. And is there also variance on the _same_ platform across multiple runs?

Yes, it would vary between platforms, even if I switched to `mimalloc` and also added the warmup. Not by much, and not on all files, but it still varied enough to cause CI to fail. The variance was on the order of ~0.05% of bytes allocated. For example, RadixUIAdoptionSection.jsx was allocating 724 bytes in the system allocator on my local machine, but parsing that file same file in CI allocated 756 bytes (an extra 32 bytes) in the system allocator.  Interestingly, not all files exhibited variance, some were exactly the same in CI. I wonder if it's a particular language feature that is causing the issue. 

> I'd have broadly expected this measure to be deterministic. If it's not, the reason why might also shed some light on some of the variance we have in our benchmarks. It seems to relate to non-deterministic allocation patterns in `HashMap`s, but I've never understood where that's coming from.

I also expected it to be the same, given that it's the same input, and the same parsing code, but evidently there are other factors I'm not considering. The number of allocations appears to be affected by the platform it runs on, the order that files are parsed in, and the allocator being used as well.

tasks/allocs/src/lib.rs
camchenry: I did consider this, but I found that the number of system allocations increased when I defaulted to the system allocator and there was more variance in the number of allocations. When I switched to mimalloc, it stabilized the number of allocations across my local machine and CI, but still kept some variance in the number of bytes allocated. Maybe if we can track down all the sources of non-determinism we could default back to the system allocator.

Regarding platform support: mimalloc is well supported across macOS, Linux, and Windows as far as I can tell. There were some issues with executing illegal instructions on some platforms (like Raspberry Pi), but I think those are fixed now: https://github.com/oxc-project/oxc/issues/11045, https://github.com/oxc-project/oxc/pull/11743.

camchenry:
> So the ideal would be to increment `num_alloc` in `Alloc::alloc` + `Alloc::grow` as well. But problem is that in those methods you only have access to the `Bump`, not the `Allocator`.
> 
> A terrible hack would be: [...]
> This is completely unsound! It relies on `Bump` always being wrapped inside an `Allocator`, which can't be statically proven. But in practice we know that's always the case in Oxc - we never use `Bump` on its own. This unsoundness would be unacceptable in production code, but as it's behind a cargo feature which is only used in this internal tool, I think it'd be OK.

I think I can live with this, if you're okay with a little bit of extra feature-gated code in these methods. I think the effort will be worth it, as this will help us unlock a new avenue of performance optimization as well as preventing memory/performance regressions.

> When I finally finish up replacing `Bump` with our own custom allocator ([WIP](https://github.com/oxc-project/oxc/tree/05-16-feat_allocator_replace_bumpalo_)) we could build allocation counting into it, and remove this hack.

This sounds like a great path forward to me.

tasks/allocs/src/lib.rs
overlookmotel: I think doing the warm-up run is a good solution. As you say, it's resilient to changing the files later on.

Only change I'd suggest is updating the "but wasn't able to figure out why" comment.

tasks/allocs/src/lib.rs
overlookmotel: > I did consider this, but I found that the number of system allocations increased when I defaulted to the system allocator and there was more variance in the number of allocations. When I switched to mimalloc, it stabilized the number of allocations across my local machine and CI, but still kept some variance in the number of bytes allocated. Maybe if we can track down all the sources of non-determinism we could default back to the system allocator.

Very interesting. Why would there be a different number of allocations depending on the allocator? The code requesting allocations is the same. Doesn't make sense to me! But there's clearly something I'm missing.

But how the hell would we figure out where the difference is coming from? `#[track_caller]` on allocation functions, log every call, and then diff the logs? I have no idea at all if that'd work.

And just to clarify: When you say "there was more variance in the number of allocations", do you mean there was variance between different runs *on the same machine*, or there was a difference between local and CI?

> Regarding platform support: mimalloc is well supported across macOS, Linux, and Windows as far as I can tell. There were some issues with executing illegal instructions on some platforms (like Raspberry Pi), but I think those are fixed now: #11045, #11743.

👍 Red herring. Sorry!

overlookmotel: > but still kept some variance in the number of bytes allocated

Were you logging the number of bytes *requested* or delivered? I know mimalloc (and possibly other allocators too) may give you a slightly larger allocation than you asked for (e.g. rounding up to multiple of 16 bytes), and I'd expect that not to be 100% deterministic. But the number of bytes requested *should* (if I understand right) be deterministic in single-threaded code.

Unless Rust compiler is non-deterministic, so every time you compile the code, you get a slightly different binary, which will behave slightly differently. That seems quite plausible, if unfortunate.

overlookmotel:
> I think I can live with this, if you're okay with a little bit of extra feature-gated code in these methods. I think the effort will be worth it, as this will help us unlock a new avenue of performance optimization as well as preventing memory/performance regressions.

I can live with monstrous unsoundness, if you can! 😄 I agree it's a metric we can likely leverage to make improvements on perf.

overlookmotel:
Oh, just one more thing. There's an annoyance with cargo features.

We run tests with `--all-features`, and we don't really want this code active in tests. If nothing else, it'll slow them down due to all the atomic ops in `Allocator`, and if we're adding the terrible unsound hack to `Alloc` trait, then we *definitely* don't want that code running in any circumstances other than `oxc_allocs`.

The crappy "solution" we've found for feature-gating certain code that's only used for oxlint JS plugins is to have 2 features called `oxlint2` and `disable_oxlint2`. All the code which we want compiled for `napi/oxlint2` is gated with:

```rs
#[cfg(all(feature = "oxlint2", not(feature = "disable_oxlint2")))]
fn gated() { /* ... */ }
```

So `gated` is not compiled with `--all-features`. You have to just enable `oxlint2` feature by itself.

This is crap. But it's the best we've come up with so far. So unfortunately I suggest using the same hack here in `oxc_allocator` crate - `track_allocations` and `disable_track_allocations` features.

To double-protect against hack code being compiled in production builds, I'd recommend adding `disable_track_allocations` to default features in `oxc_allocator` and then disabling default features for `oxc_allocator` in `oxc_allocs`.

camchenry:
> This is crap. But it's the best we've come up with so far. So unfortunately I suggest using the same hack here in `oxc_allocator` crate - `track_allocations` and `disable_track_allocations` features.
> 
> To double-protect against hack code being compiled in production builds, I'd recommend adding `disable_track_allocations` to default features in `oxc_allocator` and then disabling default features for `oxc_allocator` in `oxc_allocs`.

Done. However, I wasn't able to add `disable_track_allocations` as a default feature, as you can't override the default features in `oxc_allocs` then apparently:

```
error: failed to load manifest for workspace member `oxc/tasks/allocs`
Caused by:
  error inheriting `oxc_allocator` from workspace root manifest's `workspace.dependencies.oxc_allocator`

Caused by:
  `default-features = false` cannot override workspace's `default-features`
```

tasks/allocs/src/lib.rs
camchenry: > Were you logging the number of bytes _requested_ or delivered? I know mimalloc (and possibly other allocators too) may give you a slightly larger allocation than you asked for (e.g. rounding up to multiple of 16 bytes), and I'd expect that not to be 100% deterministic. But the number of bytes requested _should_ (if I understand right) be deterministic in single-threaded code.

Just the number of bytes requested: `layout.size()`.

> But how the hell would we figure out where the difference is coming from? `#[track_caller]` on allocation functions, log every call, and then diff the logs? I have no idea at all if that'd work.

Possibly. It is definitely possible to enable debug symbols and set a breakpoint on the global allocator function itself and trace back to callers that way, but it is tedious.

> And just to clarify: When you say "there was more variance in the number of allocations", do you mean there was variance between different runs _on the same machine_, or there was a difference between local and CI?

Thankfully, just variance on number of allocations between local and CI. The number of bytes allocated and total number of allocations seems to be stable on the same machine as far as I can tell (from my `n=1` sample size).

tasks/allocs/src/lib.rs
camchenry: Okay I'm tracking reallocs separately now. It appears the grow-in-place behavior is very platform dependent (or at least allocator dependent). mimalloc seems to always return a different pointer? And I get different results for the macOS system allocator and system allocator in CI with the `ubuntu-latest` image.

Example:
image

tasks/allocs/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "oxc_allocs"
```
Boshen: `allocs` is too vague, maybe `oxc_track_memory_allocations` and `track_memory_allocations` for the directory name.

You can add a just command to shorten typing: `just allocs`.

tasks/allocs/allocs_parser.snap
Boshen, line 1: 👍 really nice!

tasks/allocs/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+
+use oxc_allocator::Allocator;
+use oxc_parser::{ParseOptions, Parser};
+use oxc_tasks_common::{TestFiles, project_root};
+
+use std::alloc::{GlobalAlloc, Layout};
+use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
+
+#[global_allocator]
+static GLOBAL: TrackedAllocator = TrackedAllocator;
```
Boshen: I wonder whether we can get https://docs.rs/dhat/latest/dhat working as well. Just a thought though, no changes required.

tasks/allocs/Cargo.toml
camchenry: Fair. I added the `just allocs` alias, but renamed the task to `oxc_track_memory_allocations`.

tasks/allocs/src/lib.rs
camchenry: Hm, I'll have to think about this. The allocation backtraces I think are the most useful feature we could take from this. That would make it much easier to find where we are allocating without needing to debug.

Boshen:
### Merge activity

* **Aug 6, 5:37 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Aug 6, 5:37 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Aug 6, 5:43 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #12576, opened 2025-07-28
refactor(napi/oxlint): add types to `mergeVisitFns`

overlookmotel:
Add types to `mergeVisitFns`.

PR #12577, opened 2025-07-28
refactor(napi/oxlint): add more types

overlookmotel:
Add more types to `visitor.ts`.

This is quite tricky, because the type of `compiledVisitor` changes over time.

Before calling `finalizeCompiledVisitor`, it's:

```
VisitFn | VisitFn[] | { enter: VisitFn | VisitFn[] | null, exit: VisitFn | VisitFn[] | null } | null
```

but after calling `finalizeCompiledVisitor`, all the arrays have been removed, so it's:

```
VisitFn | { enter: VisitFn | null, exit: VisitFn | null } | null
```

And which of `VisitFn` or `{ enter: VisitFn | null, exit: VisitFn | null }` it is depends on the index into `compiledVisitor`.

I've done my best, but this is quite an imperfect solution to these vagaries.

### DIFF
--- a/napi/oxlint2/src-js/visitor.ts
+++ b/napi/oxlint2/src-js/visitor.ts
@@ -72,13 +72,21 @@
 // for objects created by user code in visitors. If ephemeral user-created objects all fit in new space,
 // it will avoid full GC runs, which should greatly improve performance.
 
-import type { Node, VisitFn } from './types.ts';
+import type { CompiledVisitorEntry, EnterExit, Node, VisitFn, Visitor } from './types.ts';
 
 // TODO(camc314): we need to generate `.d.ts` file for this module.
 // @ts-expect-error
 import { LEAF_NODE_TYPES_COUNT, NODE_TYPE_IDS_MAP, NODE_TYPES_COUNT } from '../dist/parser/generated/lazy/types.cjs';
 
 const { isArray } = Array;
+
+// Types for temporary state of entries of `compiledVisitor`
+// between calling `initCompiledVisitor` and `finalizeCompiledVisitor`.
+type CompilingLeafVisitorEntry = VisitFn | VisitFn[] | null;
+type CompilingNonLeafVisitorEntry = {
+  enter: VisitFn | VisitFn[] | null;
+  exit: VisitFn | VisitFn[] | null;
+} | null;
 
 // Compiled visitor used for visiting each file.
 // Same array is reused for each file.
@@ -87,7 +95,7 @@ const { isArray } = Array;
 // not "holey" (hash map). This is critical, as looking up elements in this array is a very hot path
 // during AST visitation, and holey arrays are much slower.
 // https://v8.dev/blog/elements-kinds
-export const compiledVisitor: any[] = [];
+export const compiledVisitor: CompiledVisitorEntry[] = [];
 
 for (let i = NODE_TYPES_COUNT; i !== 0; i--) {
   compiledVisitor.push(null);
@@ -125,15 +133,15 @@ let hasActiveVisitors = false;
 //
 // `enterExitObjectCacheNextIndex` is the index of first object in cache which is currently unused.
 // It may point to the end of the cache array.
-const enterExitObjectCache: any[] = [];
+const enterExitObjectCache: EnterExit[] = [];
 let enterExitObjectCacheNextIndex = 0;
 
-function getEnterExitObject() {
+function getEnterExitObject(): EnterExit {
   if (enterExitObjectCacheNextIndex < enterExitObjectCache.length) {
     return enterExitObjectCache[enterExitObjectCacheNextIndex++];
   }
 
-  const enterExit: any = { enter: null, exit: null };
+  const enterExit: EnterExit = { enter: null, exit: null };
   enterExitObjectCache.push(enterExit);
   enterExitObjectCacheNextIndex++;
   return enterExit;
@@ -149,10 +157,10 @@ function getEnterExitObject() {
 //
 // `visitFnArrayCacheNextIndex` is the index of first array in cache which is currently unused.
 // It may point to the end of the cache array.
-const visitFnArrayCache: any[][] = [];
+const visitFnArrayCache: VisitFn[][] = [];
 let visitFnArrayCacheNextIndex = 0;
 
-function createVisitFnArray(visit1: any, visit2: any): any[] {
+function createVisitFnArray(visit1: VisitFn, visit2: VisitFn): VisitFn[] {
   if (visitFnArrayCacheNextIndex < visitFnArrayCache.length) {
     const arr = visitFnArrayCache[visitFnArrayCacheNextIndex++];
     arr.push(visit1, visit2);
@@ -188,7 +196,7 @@ export function initCompiledVisitor(): void {
  *
  * @param visitor - Visitor object
  */
-export function addVisitorToCompiled(visitor: any): void {
+export function addVisitorToCompiled(visitor: Visitor): void {
   if (visitor === null || typeof visitor !== 'object') {
     throw new TypeError(
       'Visitor returned from `create` method must be an object',
@@ -218,9 +226,11 @@ export function addVisitorToCompiled(visitor: any): void {
       throw new Error(`Unknown node type '${name}' in visitor object`);
     }
 
-    const existing = compiledVisitor[typeId];
+    const existing = compiledVisitor[typeId] as CompilingNonLeafVisitorEntry;
     if (typeId < LEAF_NODE_TYPES_COUNT) {
-      // Leaf node - store just 1 function, not enter+exit pair
+      // Leaf node - store just 1 function, not enter+exit pair.
+      // Note: `existing` here is actually `CompilingLeafVisitorEntry`, but can't tell TS compiler that
+      // without adding some code which would add (small) runtime cost.
       if (existing === null) {
         compiledVisitor[typeId] = visitFn;
       } else if (isArray(existing)) {
@@ -236,9 +246,9 @@ export function addVisitorToCompiled(visitor: any): void {
         }
       } else {
         // Same as above, enter visitor is put to front of list to make sure enter is called before exit
-        compiledVisitor[typeId] = isExit
-          ? [existing, visitFn]
-          : [visitFn, existing];
+        (compiledVisitor[typeId] as CompilingLeafVisitorEntry) = isExit
+          ? [existing as unknown as VisitFn, visitFn]
+          : [visitFn, existing as unknown as VisitFn];
         mergedLeafVisitorTypeIds.push(typeId);
       }
     } else {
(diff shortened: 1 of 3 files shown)


overlookmotel:
* **#12589** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12588** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12577** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12577?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12577?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12576** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12576?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12575** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12575?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12574** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12574?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12577).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

napi/oxlint2/src-js/visitor.ts
```diff
@@ -239,8 +253,8 @@ export function addVisitorToCompiled(visitor: any): void {
       } else {
         // Same as above, enter visitor is put to front of list to make sure enter is called before exit
         compiledVisitor[typeId] = isExit
-          ? [existing, visitFn]
-          : [visitFn, existing];
+          ? [existing as VisitFn, visitFn]
+          : [visitFn, existing as VisitFn];
```
camc314: i think we should be able to massage the types so we don't have to cast here. Let's merge this as is, and i'll take a look later

overlookmotel: This is rather annoying.

Ideally want some way of telling typechecker at the top of this block that "`existing` is definitely `LeafCompiledVisitorEntry` here". Then the rest of the types could be inferred.

https://github.com/oxc-project/oxc/blob/ba55e92ac2c5853a6c06647277d5c342eae2dc81/napi/oxlint2/src-js/visitor.ts#L238-L258

I guess this would do it:

```js
const existing = compiledVisitor[typeId];
if (typeId < LEAF_NODE_TYPES_COUNT) {
  const existingLeaf = existing as LeafCompiledVisitorEntry;
  // ... use `existingLeaf` in the rest of this block
}
```

But this introduces another variable purely to satisfy TS. At runtime, this extra var is pointless, and may have a (small) runtime cost.

Is there anyway to communicate the type constraint to TS without generating more runtime code?

draphikas: What about creating a type guard function like : 

```ts
function isLeafNode(_existing, typeId): _existing is LeafCompiledVisitorEntry {
    return typeId < LEAF_NODE_TYPES_COUNT;
}
```

here is playground to demonstrate : [Playground](https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBDAnmYcBqBLAzhmAxAOzgF44AKAiAE2AC44A5a4AShID44A3CDKgbgBQg0JFhwMBGMCgAzAIYBjVAGEIAWzAYANsCoBRKTP0hccAN6C4cYEaj1MOfEQA+cAgFdt2oddC4HbFxCODdPbyEAX2FRaHhJaTklVENEkzNLPztApxC3R2CCAG0AXVD3Lx8rG1MYHMLygudS8vCq6MEkFDgAGWB5WTVNHT0m6FSoRBJ0IOdG2cIWsMqhLtQmAj6Boa1dKjGoCanSVONa1pXO5FUNXdHZ8alJ6a3B25H9h8OnqbcN152HwORyEglkHgIihgGAgRGwryYNDIAH1-FhoQQAOb0CEAa0oAHcCAAaBDXACSVHonnUACMZCx6KjTOjJJiJFhev03sM9sCfhZqtYoMAYB4oEQ1pS4AAeXr6ACCeGRDAA8gARfTIgAqAE0AAr6ADKyJUqoAqgxtVFhIpYej5UqVRqtXrDSazZbtdMAMygu0EB123n3JzQeiAvlfI4tUilUG6eBSqkVOkyaYARn99vgaIx7NIwbunzDUCKyZKoIwsnI8O5iOAZDzbNJyZYbEy1hq2Hzvi7AHp+3AAHoAfiFEhrTZZ+ZIxFIbQ7E673dZWL7K+sg5H45XkRs2iwqE7m9XveXA6HY4n0X3wEPx4nzfXE+3O8E0SAA)

overlookmotel: I've found a solution to this, but it's not very pleasing, because it involves creating pointless intermediate variables, which are a (small) cost at runtime.

overlookmotel: @draphikas  That's better! Avoids type assertions, and the types are based on the `typeId < LEAF_NODE_TYPES_COUNT` check, which needs to happen anyway.

But... this still has a runtime cost because there's an additional function call.

If the code goes through a minifier, it could inline the `isLeafNode` function, which would reduce the runtime cost to zero. But unfortunately Oxc minifier, which we'll be using on this code, [doesn't manage to inline the function at present](https://playground.oxc.rs/#eNqFklFr2zAQx7+K0FMCedgGffHWwWgTGJSkrN1gzCMo9jmIyXfmJKcJab77TkrspJCkL7bk+9//fnfnrS50pgtCH9TD+NtkPp3dj+fPvx/HT/O72c/ps7pVN59zzBHWDXFQYdOA+mW9DROU2ACphExN5TlUt1/Vimx5KrcYgCtTgBrH03htg9rmqBTEa9Y7vXanP3/ljK1z4iIq0b8j2sVqieoBTHVHdWMdlElILEV5I5hXHVLylPBK/hH+bdYF+UWn12t10tyqFotgCZX1URcHO5jLGHywuMxUi/+QXnCU9vC9zISmXgAPM9WLJPMyQJo9Q2gZDxbqy/nFd6M9LLKEyrQuqI5vcGQ6V+gMoA8c6RKBrdTgpMHOq8saDveyHjXXTrS5Tj+FYJ0EkILqgxLSI00622puMb78BoNZ6yxwCyPt5H/UWWWcl4svqIEu4jf1glx3C2zQV8T1/sNupBvDHjg6Gufo5UcqP2uDtyVMDjPpshsG0a7g0TCg7y1i8WQh577A3jMYXoKQafCfPny8kR6sJ2cClPdQOMMm2otTIpf0Qqa2hNQgoFk4eKKWC6hN01erLdrK7u0L2ZAwHVAkZnDppPWjneyM3EQai/IV8IJ8H9/t/gPis33q).

Is there any way to do the same thing without a separate function? Are inline type guards a thing?

---

2 bits of background:

1. This code is extremely performance-sensitive. Really want to avoid *any* unnecessary overhead, no matter how small.

2. This is my very first foray into TypeScript! I've been strictly plain JS up until today (I know, very old skool). So really appreciate the help.

draphikas: @overlookmotel I knew you were going to talk about the runtime cost of this function.

The `typeId < LEAF_NODE_TYPES_COUNT` check is not related to `existing` directly in types context.
Unfortunately, Typescript don't know that if `typeId < LEAF_NODE_TYPES_COUNT` is true that means the node is a leaf node.
The type guard function I showed you is a hack to enforce `existing` type without using it and infer the rest.

I don't actually know how to do that without creating this intermediate function.

But I agree that it's better if we can avoid type assertions, t can introduce some unexpected errors.

I let you decide if the performance cost of that function is more important that the code's maintainability before Oxc Minifier would support Inline function or the type assertions overhead.

overlookmotel: Thank you for coming back. I wasn't trying to attack your solution. It's ingenious.

It's more that this is my first experience of TypeScript, and I'm disappointed to find that there's a trade-off at all. I thought the whole point was that types are a compile-time only thing, and don't have a runtime cost. That cost here is tiny, but still, it's annoying...

I expected there'd be something like this:

```ts
const existing = compiledVisitor[typeId];
if (typeId < LEAF_NODE_TYPES_COUNT) {
  // Next line is completely erased by compiler
  assume existing is LeafCompiledVisitorEntry;
}
```

draphikas: Thanks !
Typescript can be frustrating sometimes. You need to do some weird things to achieve what you want with types.

The use of intermediate variables is probably the best trade-off for performance here.

draphikas: Just a thought
If we move down `const existing = compiledVisitor[typeId];`, we can use type assertion without creating extra variable :

Insted of : 
```ts
const existingEntry = compiledVisitor[typeId];
if (typeId < LEAF_NODE_TYPES_COUNT) {
  // Leaf node - store just 1 function, not enter+exit pair
  const existing = existingEntry as CompilingLeafVisitorEntry;
} else {
  // Not leaf node - store enter+exit pair
  const existing = existingEntry as CompilingNonLeafVisitorEntry;
}
```
We can do : 
```ts
if (typeId < LEAF_NODE_TYPES_COUNT) {
  // Leaf node - store just 1 function, not enter+exit pair
  const existing = compiledVisitor[typeId] as CompilingLeafVisitorEntry;
} else {
  // Not leaf node - store enter+exit pair
  const existing = compiledVisitor[typeId] as CompilingNonLeafVisitorEntry;
}
```
It's just some extra bytes but no more runtime cost

overlookmotel: That's better, yes. It's still *slightly* less optimal as the original as it produces more bytecode - but that's a tiny consideration.

Also found this method which Oxc minifier *does* manage to completely remove, because the assertion functions are completely empty:

```ts
function assertIsLeaf(value: unknown): asserts value is CompilingLeafVisitorEntry {}
function assertIsNonLeaf(value: unknown): asserts value is CompilingNonLeafVisitorEntry {}

const existing = compiledVisitor[typeId];
if (typeId < LEAF_NODE_TYPES_COUNT) {
  assertIsLeaf(existing);
  // ...
} else {
  assertIsNonLeaf(existing);
  // ...
}
```

But right now, we've not minifying this code.

Being something of a fanatic, I've opted in the end for a zero compromise approach. Latest commit makes the introduction of types not change the runtime code in *any* way, at the cost of some ugly type assertions.

Once we are minifying this code we can return to this and hopefully restore sanity. If Oxc minifier could inline it, @draphikas's type guard function seems like the best solution to me.

Thanks all for the ideas and for guiding a TS newbie!

camc314:
### Merge activity

* **Jul 29, 10:40 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Jul 29, 10:40 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Jul 29, 10:50 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #12927, opened 2025-08-09
fix(linter): fix no-fallthrough rule, when the default condition is not last

### LINKED ISSUE #12793: linter: no-fallthrough has different from eslint

1.9.0

### What command did you run?

_No response_

```jsonc
{
  "no-fallroughness": "error"
}
```

### What happened?

When the code is as follows, oxlint will error, but eslint will not error.

```js
const a  = 'a';

switch(a) {
  case 'a':
  default:
    console.log('value is a or defualt');
    break;
  case 'b':
    console.log('value is b');
    break;
}
```

[oxlint playground](https://playground.oxc.rs/#eNp1kMFOwzAMhl8l8mWbNBAgcSnihnYFwbUXN3W7iDSunGRjqvruOK22C+IUO7Y///4nsFCB5RCTQWNezQY3L3WoQzy7ZI9b3JmpDsZYjFRqVUla6jD7tMRa0mH2dO+5325O6DMZFxXGUhoz+rTZKbK0NkL4vcQrr1l5/zGaP4Mz7IGhmkByKE+8hIQ/UCXJtAfvQrrG0fJIt+QyNOyvWRIMsWMZoOrQR5r3MKJEkkJE7/n8SSlLeM8pupYOOdjkWPet46OQ9p7oA4VCXH8VUZavCD2mc73aOtWq01OsoVIXIfCd7vPpKJz7Y/msgURYaphnUMRN2MpJKD3pQUDx6eHxWU93ahImat/IehQsqlTA9QjLLfW0+EIBG09fnMXSgONN5OCC69yCXwZCEvYHvbgMnUgajmraCiyCdDnGBPMvuyy8lQ==)

[eslint playground](https://eslint.org/play/#eyJ0ZXh0IjoiY29uc3QgYSAgPSAnYSc7XG5cbnN3aXRjaChhKXtcbiAgY2FzZSAnYSc6XG4gIGRlZmF1bHQ6XG4gICAgY29uc29sZS5sb2coJ3ZhbHVlIGlzIGEgb3IgZGVmdWFsdCcpO1xuICAgIGJyZWFrO1xuICBjYXNlICdiJzpcbiAgICBjb25zb2xlLmxvZygndmFsdWUgaXMgYicpO1xuICAgIGJyZWFrO1xufSIsIm9wdGlvbnMiOnsicnVsZXMiOnsibm8tZmFsbHRocm91Z2giOlsiZXJyb3IiXX0sImxhbmd1YWdlT3B0aW9ucyI6eyJzb3VyY2VUeXBlIjoibW9kdWxlIiwicGFyc2VyT3B0aW9ucyI6eyJlY21hRmVhdHVyZXMiOnt9fX19fQ==)

lambda47:
In the previous implementation, if the default condition was not the last one, the order of cfg_ids was incorrect

fixes #12793

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -471,7 +471,7 @@ fn get_switch_semantic_cases(
                 (cfg_ids, conds, exit)
             } else {
                 if has_default {
-                    cfg_ids.push(target)
+                    cfg_ids.push(target);
                 }
                 (cfg_ids, conds, Some(target))
             }
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -477,11 +477,7 @@ fn get_switch_semantic_cases(
             }
         });
 
-    let (default, exit) = if has_default {
-        (exit, None)
-    } else {
-        (None, exit)
-    };
+    let (default, exit) = if has_default { (exit, None) } else { (None, exit) };
     cfg_ids.reverse();
     (cfg_ids, FxHashMap::from_iter(tests), default, exit)
 }
--- a/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
@@ -437,12 +437,12 @@ fn get_switch_semantic_cases(
     let cfg = ctx.cfg();
     let graph = cfg.graph();
     let has_default = switch.cases.iter().any(SwitchCase::is_default_case);
-    let (tests, exit) = graph
+    let (mut cfg_ids, tests, exit) = graph
         .edges_directed(node.cfg_id(), Direction::Outgoing)
-        .fold((Vec::new(), None), |(mut conds, exit), it| {
+        .fold((Vec::new(), Vec::new(), None), |(mut cfg_ids, mut conds, exit), it| {
             let target = it.target();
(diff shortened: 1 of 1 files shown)


camc314:
thanks for the contribution, can we add a test case for this behaviour.

lambda47:
I add the test case for it.

crates/oxc_linter/src/rules/eslint/no_fallthrough.rs
```diff
@@ -482,6 +483,7 @@ fn get_switch_semantic_cases(
     } else {
         (None, exit)
     };
+    cfg_ids.sort_by_key(|cfg_id| nodes.iter().position(|it| it.cfg_id() == *cfg_id));
```
camc314: did the original solution of pushing cfg ids to a vec work? I think that solution is preferable as this could be pretty slow if cfg_ids is large, as it's going to have to loop over all nodes fo find the right one

lambda47: The original implementation, the order of cfg_ids is dependent on the order of edges returned by edges_directed function. It is dependent on the edges_directed’s internal implementation. I think it’s not good, but it quick. So I decide to use the original implementation.

camc314: I think this is likely also why there's a perf regression

camc314: Screenshot 2025-08-10 at 12 40 46

camc314 (approved):
thanks!

camc314 (approved):
thank you!

Issue #4240, opened 2024-07-13
ast-codegen: support `#[visit(enter_before)]` attribute


Dunqing:
We have discussed this at https://github.com/oxc-project/oxc/pull/4147#issuecomment-2220216905. But we do need this at the moment.

In semantic builder, We need `enter_scope` before `enter_node` in `visit_function` and `visit_class`.

https://github.com/oxc-project/oxc/blob/2eb37d52f1250dfc1348379cbf9d3fb3c70b7961/crates/oxc_semantic/src/builder.rs#L1423-L1450

We want to treat `FunctionExpression` and `FunctionDeclaration` differently and the class is the same too.

**FunctionExpression**: create a binding for function `id` to the current scope
**FunctionDeclaration**: create a binding for function `id` to the parent scope

To add the binding to the parent scope, we must need `enter_scope` before `enter_node`.

I tried another way to get around it (Same as `TypeScript`). But I find it a bit confusing. See: https://github.com/oxc-project/oxc/pull/4195#pullrequestreview-2175953763

Dunqing:
Ah. Haha, this is also a bit confusing for visit orders.  I thought about it again. We can defer binding `FunctionExpression` and `ClassExpression` and bind them when we visit `BindingIdentifier`.

overlookmotel:
@Dunqing Just to throw in my 2 cents, I wonder if `FunctionExpression` and `FunctionDeclaration` should be separate types (and each type would have `Function` as a field within it). The 2 have various significant differences:

1. One is a `Statement`, the other an `Expression`.
2. `FunctionExpression`'s `name` field should be an `Option`, whereas it's compulsory in `FunctionDeclaration`.
3. Scope should be entered before `name` for `FunctionExpression` and after name for `FunctionDeclaration`.

We could also have a separate type for `ObjectMethod` (whose `name` isn't a binding at all).

At present, it's possible to create non-sensical AST nodes (e.g. a `FunctionDeclaration` with no name) - it'd be good to make this impossible through the type system.

Dunqing:
> @Dunqing Just to throw in my 2 cents, I wonder if `FunctionExpression` and `FunctionDeclaration` should be separate types (and each type would have `Function` as a field within it). The 2 have various significant differences:
> 
> 1. One is a `Statement`, the other an `Expression`.
> 2. `FunctionExpression`'s `name` field should be an `Option`, whereas it's compulsory in `FunctionDeclaration`.
[...]

PR #12930, opened 2025-08-09
refactor(ast_tools/formatter): compile list of no following nodes as `TypeId`s

overlookmotel:
Follow-on after #12864. Pure refactor - does not alter generated code.

There's a couple of things going on in this PR.

Firstly, perf:

* Prefer comparing `TypeId`s over string comparison, because it's cheaper.
* Compile a single `Vec` of all types which have no following node at the start (including those listed in `AST_NODE_WITHOUT_FOLLOWING_NODE_LIST`), rather than checking 2 `Vec`s each time in `generate_struct_impls`.

Secondly, the code previously was relying on the fact that almost all of `Statement`s variants have the same name as the types those variants contain e.g.:

```rs
pub enum Statement<'a> {
    BlockStatement(Box<'a, BlockStatement<'a>>),
    // ...
}
```

But that the names match is a bit of a co-incidence. This PR makes it so we don't rely on that co-incidence, and instead makes the "no following node" list from the actual *types* of the variants.

This change reveals a side-effect of the previous behavior which may or may not be intentional. The only variants of `Statement` where the name of the variant and name of the *type* of the variant don't match are `Function` and `Class`.

To maintain the same output as before, I've added an exclude list `AST_NODE_WITH_FOLLOWING_NODE_LIST` containing these 2 types.

Presumably the reason why these 2 need an exclusion is because `Function` and `Class` can be either a `Statement` or an `Expression`. I don't know if this may be problematic and might need logic elsewhere to handle them differently, depending on the context?

Note: `TypeDef::innermost_type` is to get the `TypeId` of the inner type of a variant (e.g. `BlockStatement`), rather than the type which is the actual enum variant (e.g. `Box<BlockStatement>`).

@Dunqing I don't know the formatter at all, so these changes may be unhelpful. Feel free to close or modify this PR if so.

### DIFF
--- a/tasks/ast_tools/src/generators/formatter/ast_nodes.rs
+++ b/tasks/ast_tools/src/generators/formatter/ast_nodes.rs
@@ -9,7 +9,7 @@ use crate::{
     Codegen, Generator,
     generators::define_generator,
     output::{Output, output_path},
-    schema::{Def, EnumDef, FieldDef, Schema, StructDef, TypeDef, VariantDef},
+    schema::{Def, EnumDef, FieldDef, Schema, StructDef, TypeDef, TypeId},
 };
 
 pub fn get_node_type(ty: &TokenStream) -> TokenStream {
@@ -23,13 +23,15 @@ const FORMATTER_CRATE_PATH: &str = "crates/oxc_formatter";
 const AST_NODE_WITHOUT_FOLLOWING_NODE_LIST: &[&str] =
     &["AssignmentExpression", "FormalParameters", "StaticMemberExpression"];
 
+const AST_NODE_WITH_FOLLOWING_NODE_LIST: &[&str] = &["Function", "Class"];
+
 pub struct FormatterAstNodesGenerator;
 
 define_generator!(FormatterAstNodesGenerator);
 
 impl Generator for FormatterAstNodesGenerator {
     fn generate(&self, schema: &Schema, _codegen: &Codegen) -> Output {
-        let all_statement_variants_names = get_all_statement_variants_names(schema);
+        let no_following_node_type_ids = get_no_following_node_type_ids(schema);
 
         let impls = schema
             .types
@@ -38,7 +40,7 @@ impl Generator for FormatterAstNodesGenerator {
                 TypeDef::Struct(struct_def)
                     if struct_def.visit.has_visitor() && !struct_def.builder.skip =>
                 {
-                    Some(generate_struct_impls(struct_def, &all_statement_variants_names, schema))
+                    Some(generate_struct_impls(struct_def, &no_following_node_type_ids, schema))
                 }
                 TypeDef::Enum(enum_def) if enum_def.visit.has_visitor() => {
                     Some(generate_enum_impls(enum_def, schema))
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#12930** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12930?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12930?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12929** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/12929?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=12930).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camc314 (approved):

Dunqing (approved):
Thanks for looking into this and improving it! I haven't started to polish this because those generator files are still constantly changing, so putting effort into it might be wasted at some point. Anyway, your changes are very good, which could let me make full use of existing APIs to achieve my desired functionality.

> Presumably the reason why these 2 need an exclusion is because `Function` and `Class` can be either a `Statement` or an `Expression`. I don't know if this may be problematic and might need logic elsewhere to handle them differently, depending on the context?


Oh, this is a potential problem that I wasn't aware of before. Now, whether we have or do not have a following node depends on whether printing comments relies on. Currently, the class printing implementation isn't complete, so I am not sure whether `Class` or `Function` should be excluded or not. I will revisit this once the `Class` implementation is complete.

Dunqing:
### Merge activity

* **Aug 9, 2:03 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Aug 9, 2:03 PM UTC**: `Dunqing` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Aug 9, 2:09 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

overlookmotel:
> Oh, this is a potential problem that I wasn't aware of before. Now, whether we have or do not have a following node depends on whether printing comments relies on. Currently, the class printing implementation isn't complete, so I am not sure whether `Class` or `Function` should be excluded or not. I will revisit this once the `Class` implementation is complete.

In my opinion we should have separate AST types for `FunctionExpression` and `FunctionDeclaration`, like ESTree and Babel (ditto for classes). Maybe if a single `Function` type causes problems for formatter, it'll be the spur to finally make that change!

Dunqing:
> In my opinion we should have separate AST types for `FunctionExpression` and `FunctionDeclaration`, like ESTree and Babel (ditto for classes). Maybe if a single `Function` type causes problems for formatter, it'll be the spur to finally make that change!

Yes!! We've discussed this many times, please see https://github.com/oxc-project/oxc/issues/4240#issuecomment-2227160460. This is what we really want to change, but it hasn't happened yet. Hope it is not too far.

PR #13012, opened 2025-08-12
feat(tasks): track allocations for semantic

camchenry:
Now that we've tested the allocation tracking for parsing for a while and it has been fairly stable, I'd like to start tracking allocations for Semantic as well. Tracking the arena allocations for semantic is non-trivial unfortunately due to the usage of the allocator internally. Once we can supply an allocator to `SemanticBuilder` it will be easy to start tracking allocations.

### DIFF
--- a/tasks/track_memory_allocations/src/lib.rs
+++ b/tasks/track_memory_allocations/src/lib.rs
@@ -11,7 +11,8 @@ use mimalloc_safe::MiMalloc;
 
 use oxc_allocator::Allocator;
 use oxc_parser::{ParseOptions, Parser};
-use oxc_tasks_common::{TestFiles, project_root};
+use oxc_semantic::SemanticBuilder;
+use oxc_tasks_common::{TestFile, TestFiles, project_root};
 
 use std::alloc::{GlobalAlloc, Layout};
 use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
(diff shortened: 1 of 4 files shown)


camchenry:
* **#13090** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13012** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13012?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=13012).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Number of arena reallocations happening in `SemanticBuilder` could be really interesting here.

Semantic uses a separate `Allocator`. You'd need to get access to it somehow to find out how many allocations it makes.

camchenry:
> Number of arena reallocations happening in `SemanticBuilder` could be really interesting here.
> 
> Semantic uses a separate `Allocator`. You'd need to get access to it somehow to find out how many allocations it makes.

It looks like this would be a pretty non-trivial change. I think we'd need to update the public API to match `Parser` where it accepts an allocator reference and stores it internally? For now, I'm just going to set this up to track system allocations for now so we at least have some information available.

tasks/track_memory_allocations/src/lib.rs
```diff
@@ -73,16 +74,108 @@ fn test() {
...
+        allocator.reset();
+        reset_global_allocs();
+
+        let ret = Parser::new(&allocator, &file.source_text, file.source_type)
+            .with_options(options)
+            .parse();
+
+        parser_stats.push(record_stats(file, &allocator));
+
+        let program = ret.program.clone_in(&scratch_allocator);
```
overlookmotel: Can we skip cloning `Program` here?

The cloned `Program` will be slightly different from the original - e.g. a `Vec` with len 1, cap 4 will be compacted in the clone to len 1, cap 1. Working with that compacted AST *might* alter the pattern of heap allocations slightly. By my logic, it *shouldn't* make any difference, but there's other things we're seeing with heap allocations that I can't explain (different behavior on different platforms), so obviously my logic is wrong somehow. So I think its preferable to stick as close to "real world" usage as we can.

A more obvious reason to remove the clone is that I don't think it does anything useful, since we're not able to measure arena allocations in semantic at present. And if we could, it'd be a different allocator anyway.

overlookmotel: Also, I don't think you're `reset`-ing `scratch_allocator`.

tasks/track_memory_allocations/src/lib.rs
```diff
@@ -73,16 +74,108 @@ fn test() {
...
+
+        allocator.reset();
+        reset_global_allocs();
+
+        // TODO: Pass allocator into SemanticBuilder when it supports using an external allocator
+        // instead of the internal allocator.
```
overlookmotel: For reasons mentioned in https://github.com/oxc-project/oxc/pull/13034#issuecomment-3183827022, I don't think it's likely we'll be able to easily make `Scoping` use an external allocator. So I think the point here is more that we need to find a way to get stats from `Scoping`'s existing internal allocator.

I imagine we can find a way to do that. #13042 stack makes some changes to `Scoping` and its `Allocator`. Once that's landed, we may be able to build on it to get the stats we need out of it.

tasks/track_memory_allocations/src/lib.rs
camchenry: This was some laziness on my part, I was running into borrow checking issues and I tried to clone my way into some productivity. Looking at this with fresh eyes, I realized it's because we were resetting the allocator between parser/semantic which will obviously clear memory for the program, so we don't want to do that. I updated so that we can compute the stats by taking diffs from the previous allocation state and only doing a reset at the beginning of each loop.

overlookmotel: Ah the dastardly borrow-checker, always peeing on our parade.

Building `Semantic` doesn't make any allocations into to AST `Allocator`, does it? Does it??? If it does, you've discovered something we should address.

camchenry: > Building `Semantic` doesn't make any allocations into to AST `Allocator`, does it? Does it??? If it does, you've discovered something we should address.

No, the borrow checker was 100% correct in this case: I was trying to reset the allocator's memory while the parser had an immutable reference to the allocator and we were using a value returned by the parser (i.e., the `program`).

I don't think there are any allocations being made into the AST allocator.

camchenry:
Going to close for now and will revisit at a later point.

Issue #11723, opened 2025-06-15
oxc-parser: AST issue for `type Foo = intrinsic.bar` with rawTransfer enabled


ArnaudBarre:
body[0].typeAnnotation.typeName.left.name is `"\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000"` instead of "intrinsic". The issue only appears without raw transfer

This seems to be a regression from 0.72

Issue also happens for `import { type as as } from "./mod.js";`

overlookmotel:
Just to check: Do you mean this was working in 0.72 and it broke after that? Or it worked in 0.71 and got broken in 0.72?

I can replicate this locally. I'm wondering why didn't this get spotted in our conformance tests? Maybe they're just too obscure, and TS doesn't have test cases for them.

ArnaudBarre:
I just tested and it works with 72.2

overlookmotel:
How did you find these bizarre cases?

ArnaudBarre:
Running against `babel/packages/babel-parser/test/fixtures/typescript/types/intrinsic-identifier/input.ts` from `oxc/tasks/coverage`

overlookmotel:
Have you run it on all of `babel/packages/babel-parser/test/fixtures/typescript`? Or even `babel/packages/*/test/**`?

ArnaudBarre:
on `oxc/tasks/coverage/{babel,acorn-test262,typescript,test262}/**/*.{js,jsx,ts,tsx}` 🤪

overlookmotel:
> on `oxc/tasks/coverage/{babel,acorn-test262,typescript,test262}/**/*.{js,jsx,ts,tsx}` 🤪

OK great! And these are the only problems you found?

ArnaudBarre:
Yes the only one with 0.73

overlookmotel:
> Yes the only one with 0.73

Praise be! Thanks for testing it so thoroughly.

PR #13016, opened 2025-08-12
[POC] feat(allocator): Remove &A: Alloc from Vec

aapoalas:
A proof of concept of what it would look like if the `&A: Alloc` reference was made `PhantomData` in `Vec` like it is in `Box`.

This drops 8 bytes of data from each vector, slimming down the AST not insignificantly. Much of the code around the AST building now just passes an extra `self.ast.allocator` reference around (here some type shenanigans lead it to being `self.ast.allocator.bump()` though), which is perhaps mildly annoying but not too different from what was already happening since the allocator was being used for eg. `Box/Vec::new_in` already.

These changes are enough to get the allocator, parser, and semantic crates working. Next up linter, minifier and transformer together give out 381 errors :)

### DIFF
--- a/crates/oxc_allocator/src/allocator.rs
+++ b/crates/oxc_allocator/src/allocator.rs
@@ -591,7 +591,7 @@ impl Allocator {
     /// allocator.alloc(123u64);
     ///
     /// // `vec` has to grow beyond it's initial capacity
-    /// vec.extend([1, 2, 3, 4]);
+    /// vec.extend_desugared(([1, 2, 3, 4]).into_iter(), allocator.bump());
     ///
     /// // `vec` takes up 32 bytes, and `123u64` takes up 8 bytes = 40 total.
     /// // But there's an additional 16 bytes consumed for `vec`'s original capacity of 2,
@@ -621,7 +621,7 @@ impl Allocator {
     // `#[inline(always)]` because it's a no-op
     #[expect(clippy::inline_always)]
     #[inline(always)]
-    pub(crate) fn bump(&self) -> &Bump {
+    pub fn bump(&self) -> &Bump {
         &self.bump
     }
 
--- a/crates/oxc_allocator/src/vec.rs
+++ b/crates/oxc_allocator/src/vec.rs
@@ -101,13 +101,13 @@ impl<'alloc, T> Vec<'alloc, T> {
     ///
     /// // These are all done without reallocating...
     /// for i in 0..10 {
-    ///     vec.push(i);
+    ///     vec.push(i, arena.bump());
     /// }
     /// assert_eq!(vec.len(), 10);
     /// assert_eq!(vec.capacity(), 10);
     ///
     /// // ...but this may make the vector reallocate
-    /// vec.push(11);
+    /// vec.push(11, arena.bump());
     /// assert_eq!(vec.len(), 11);
     /// assert!(vec.capacity() >= 11);
     ///
@@ -135,7 +135,7 @@ impl<'alloc, T> Vec<'alloc, T> {
         let hint = iter.size_hint();
         let capacity = hint.1.unwrap_or(hint.0);
         let mut vec = InnerVec::with_capacity_in(capacity, allocator.bump());
-        vec.extend(iter);
+        vec.extend_desugared(iter, allocator.bump());
         Self(vec)
     }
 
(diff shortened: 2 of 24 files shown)


Boshen:
Perhaps a trickery to https://github.com/oxc-project/oxc/blob/main/crates/oxc_allocator/src/accessor.rs can improve the API?

overlookmotel:
We have intended to reduce size of `Vec` to 16, but using a different method: https://github.com/oxc-project/backlog/issues/174. Sorry, this was originally included in #9706, but that issue was closed when we completed the first batch of work (reducing `Vec` from 32 bytes to 24).

The approach you've taken in this PR is likely more efficient than https://github.com/oxc-project/backlog/issues/174, because there's no need to "decode" pointer to `Allocator` from other data in the `Vec`. However, I see a couple of disadvantages:

### 1. Ergonomics

It's a pain to have to pass `&Allocator` to APIs like `Vec::push`, `Vec::extend` etc.

### 2. `Allocator` confusion

Probably more importantly, how do we ensure that the `&Allocator` matches the `Vec`?

If you create a `Vec` with `let vec = Vec::new_in(&alloc1)`, nothing prevents you calling `vec.push(value, &alloc2)`, as long as `alloc2` lives as long as `alloc1`. But this isn't really what you want. Which arena the `Vec`'s data ends up in is quite unpredictable:

1. If there's spare capacity in `vec`, the data remains in `alloc1`.
2. If `vec` is full to capacity, the data is moved into `alloc2` (but the *contents* of the `Vec` may still include data which is in `alloc1`).

I don't think we can end up with UB in this way - borrow checker will prevent that - but it's still rather messy.

There are parts of codebase where we have multiple `Allocator`s in play at the same time (e.g. `oxc_semantic` and `oxc_mangler`), and I'm concerned that this change may make it easier to use the wrong one.

#### Raw transfer

There's one constraint beyond Rust's standard ones. "Raw transfer" relies on all data for an AST being in a single arena. The arena gets shared with JS as a `Uint8Array`. JS side only has access to this 1 arena, so any data outside it isn't accessible.

Raw transfer underlies linter JS plugins (coming soon). We've already had a few problems with AST referencing string data outside of the arena breaking raw transfer (#11723). Unfortunately, because our requirements go further than Rust's - we need to guarantee the *location* of data, as well as how long it lives - the borrow-checker can't help us enforce our requirements.

The "blast radius" of such bugs at present is quite limited. We just get nonsense data on JS side.

But once we have linter plugins done, the next step is to apply the same techniques to support JS transform plugins. At this point we get into *really* tricky territory. JS will be writing data into a `Uint8Array` which is essentially a view into Rust's memory. Basically, it's unsafe Rust, written in JS! If the data in the arena is unexpected (like pointing to data outside of that arena), we could easily end up with UB.

I believe we *can* make this sound, but to do so we're going to have to "lock down" our string types (currently `Atom`) to make it a guarantee that `Atom` points to data in the arena. The ability to create an `Atom` from a `&'static str` will have to go, for example.

Anyway... my point is that we already have quite a tricky problem to solve here, and I'm concerned that making it easier for `Vec` to reference data in another arena expands the problem space significantly.

#### Branded lifetimes

Only way I'm aware of to "lock" a `Vec` to a specific `Allocator` is branded lifetimes. Personally I don't rule that out, and to completely guarantee soundness of JS transform plugins, it may transpire that we need to go that route. However, as you're aware, that is not for the faint-hearted, so I had hoped to defer that until later.

overlookmotel:
> These changes are enough to get the allocator, parser, and semantic crates working.

It would be really interesting to see the effect on benchmarks at this point.

Could you possibly disable all the benchmarks except parser and semantic? Or remove this line:

https://github.com/oxc-project/oxc/blob/c58a699e1907680085751c368730ebd55585691a/.github/workflows/benchmark.yml#L47

Either of those I *think* should allow parser and semantic benches to run, even if the rest won't compile.

overlookmotel:
By the way, my comments above weren't meaning to say "we shouldn't do this". Just asking questions...

Boshen:
The idea is 👍 but appending the allocator everywhere is 👎

aapoalas:
Oops sorry, I had totally forgotten about this. If you want to, I can still try those benchmarks out.

overlookmotel:
If the code compiles already and so it wouldn't take too long, yes it'd be great to see what parser and semantic benchmarks look like after this change.

PR #13034, opened 2025-08-12
refactor!(semantic): add external allocator for semantic

camchenry:


camchenry:
I'm not really actively working on this. I just had some WIP I wanted to save. Was curious what it would take to change the API. Seems like a big change.

overlookmotel:
Just FYI: I agree that it'd be ideal if `Scoping` didn't own its own `Allocator`. The problem is that Rolldown relies on `Scoping` owning its own data, and not having a lifetime. I don't fully understand the reason why, but maybe something to do with being able to move/share `Scoping` across threads.

PR #13081, opened 2025-08-14
fix(linter/array-callback-return): fix handling of default case in switch statements for array-callback-return rule

lambda47:
fix handling of default case in switch statements for array-callback-return rule

- fixes(#13075)
- Consider cases following the default case in switch statements


camc314 (approved):
thanks!

PR #13137, opened 2025-08-16
feat(semantic): add ability to lookup if AST contains any node kinds

camchenry:
I decided to revive this stack: https://github.com/oxc-project/oxc/pull/12227 and merge all of the changes into one PR for slightly easier review and changes.

This PR adds a new `AstTypesBitset` struct which is a set of bits that has a fixed size known at compile time and can be used to store AST types as a set. The aim of this structure is to making checks like "does this file have a for statement?" possible to answer without doing any additional iterations over the AST.

PR #13138, opened 2025-08-16
perf(linter): skip rules that do not have any relevant node types

camchenry:
- part of https://github.com/oxc-project/oxc/issues/12223
- supersedes stack: https://github.com/oxc-project/oxc/pull/12228

This PR adds the ability to add node type information to indicate what node types a lint rule acts on. The lint rule runner can use this information to optimize running lint rules by skipping nodes, files, or rules that don't apply.

This is a simple version that is focused on creating the framework for automatically generating the rule runner code. This is not the end of optimizations. Currently only 102 rules have node type information, while 471 rules have no node type information and will not be skipped. Adding node type information for more rules will improve performance. We may also be able to skip more nodes and files by being clever about how we use this information. For example, filtering out any rules which do not run on any node type, if the file contains no relevant node types for the rule.

- Adds a new `RuleRunner` trait which includes information needed for indicating whether a rule has node type info or not
- Adds a `oxc_linter_codegen` task which is used to generate the implementation of `RuleRunner` trait for all lint rules. This is mostly written by AI, so worth a bit of skepticism, but I've tried to manually verify the results and it looks good (just lacking more complex detection).
  - Currently, only rules that contain a single top-level `if` statement (or chain of `else if`), or a single `match` statement will have their node type info generated. This should give us a high level of confidence, since these patterns are relatively easy to verify and identify, but it's only a subset of all rules. Next steps will be to add support for `let...else` 
- Adjusts the main linting loop to use the node type information to skip rules for node types that don't apply
- Updates CI to include a check for any uncommitted linter codegen changes

---

## Benchmarks

In the latest version of this PR, CodSpeed shows an up to +13% increase in linter performance across large and small files. In the real-world, I've seen a ~5-15% improvement in performance, but in some cases close to zero for very small files. In either case, there appears to be little downside and a lot of potential upside here.

Screenshot 2025-08-24 at 9 29 30 PM
Screenshot 2025-08-24 at 9 30 24 PM
Screenshot 2025-08-24 at 9 34 52 PM


camchenry:
* **#13138** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13138?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#13287](https://github.com/oxc-project/oxc/pull/13287) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13287?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#13573](https://github.com/oxc-project/oxc/pull/13573) <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13573?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13137** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=13138).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
What an unbelievable performance improvement!

Boshen:
Insane!

crates/oxc_semantic/src/ast_types_bitset.rs
```diff
@@ -66,6 +68,55 @@ impl AstTypesBitset {
...
+    type Item = AstType;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while self.current_word == 0 {
+            self.word_index += 1;
+            if self.word_index >= self.bitset.0.len() {
+                return None;
+            }
+            self.current_word = self.bitset.0[self.word_index];
+        }
```
overlookmotel: Storing `current_word` in the iterator and mutating it is a nice solution.

Nit: This change makes it resilient against the iterator being abused by calling `next` a ridiculously large number of times. Previously `word_index` would wrap around back to 0 if you called it `usize::MAX` times.

```suggestion
        while self.current_word == 0 {
            let word_index = self.word_index + 1;
            // `==` would be sufficient, but `>=` should remove bounds check below
            if word_index >= self.bitset.0.len() {
                return None;
            }
            self.word_index = word_index;
            self.current_word = self.bitset.0[word_index];
        }
```

You can then also implement:

```rs
impl FusedIterator for AstTypesIter<'_> {}
```

Doubt that'll make any perf difference to current use, but it's in general a good idea.

crates/oxc_semantic/src/ast_types_bitset.rs
```diff
@@ -66,6 +68,55 @@ impl AstTypesBitset {
...
+        // isolate lowest set bit
+        let tz = self.current_word.trailing_zeros() as usize; // safe: current_word != 0
```
overlookmotel: `NonZeroUsize::trailing_zeros` is cheaper than `usize::trailing_zeros` on some platforms.

```suggestion
        // SAFETY: `self.current_word` cannot be 0 or we'd have already exited in the loop above
        let current_word = unsafe { NonZeroUsize::new_unchecked(self.current_word) };
        let tz = current_word.trailing_zeros() as usize;
```

You can also do this without unsafe code by altering the loop above to use safe `NonZeroUsize::new` method for the `self.current_word == 0` check.

crates/oxc_semantic/src/ast_types_bitset.rs
```diff
@@ -66,6 +68,55 @@ impl AstTypesBitset {
...
+        let global_bit_index = self.word_index * USIZE_BITS + tz;
+        // clear that bit
+        self.current_word &= self.current_word - 1;
+        if global_bit_index > AST_TYPE_MAX as usize {
+            return self.next();
+        }
```
overlookmotel: Is it ever possible for this branch to be taken? None of the other methods of `AstTypesBitset` allow setting a bit which corresponds to a value above `AST_TYPE_MAX`. I think it can be removed.

overlookmotel (commented):
Superb!

Probably you intend to do this anyway, but I think need to use `syn` to parse the rule files. The regex may be, for example, missing the `else` branch in patterns like this:

```rs
if let AstKind::Function(func) = node.kind() {
    // ...
} else if let AstKind::ArrowFunctionExpression(arrow) = node.kind() {
    // ...
}
```

If it's currently missing out some types, that may slow things down a little.

On the other hand, some patterns aren't currently being caught by the regex, like this one:

```rs
let Some(ss) = node.kind().as_switch_statement() else { return };
```

(from [no_duplicate_case rule](https://github.com/oxc-project/oxc/blob/4f8506e0611000d5d27af2ae09c476d8fdc34695/crates/oxc_linter/src/rules/eslint/no_duplicate_case.rs#L83-L94))

Matching those would take more rules out of the "any type" category, and speed things up even more!

---

I've made a few small comments about the `AstTypesIter` iterator, but that's just fiddling at the edges.

---

Edit 7/9/25: This PR has been scaled back a lot from the original. The first version showed ~90% perf improvement. For anyone wondering, that's the reason for all our "amazing perf!" comments.

crates/oxc_semantic/src/ast_types_bitset.rs
camchenry: It shouldn't be. I'm going to remove this.

camchenry:
Thanks @overlookmotel for the early feedback. I've shifted this PR to be a simple initial version of the optimization now that we've proven this can actually have double digit performance impact. This will likely be the biggest PR since it adds the framework (CI, codegen, etc.) for doing this optimization. But I'm hoping that we can break the rest of this work into smaller PRs so we can better quantify the impact of each change.

crates/oxc_semantic/src/ast_types_bitset.rs
```diff
@@ -66,6 +70,59 @@ impl AstTypesBitset {
...
+            self.current_word = self.bitset.0[word_index];
+        }
+        // isolate lowest set bit
+        // SAFETY: `self.current_word` cannot be 0 or we'd have already exited in the loop above
+        let current_word = unsafe { NonZeroUsize::new_unchecked(self.current_word) };
+        let tz = current_word.trailing_zeros() as usize;
+        let global_bit_index = self.word_index * USIZE_BITS + tz;
+        // clear that bit
+        self.current_word &= self.current_word - 1;
+        // SAFETY: `global_bit_index` is always less than AST_TYPE_MAX which is a u8
```
overlookmotel: This safety comment isn't quite correct, and a little sparse.

```suggestion
        // SAFETY: `global_bit_index` is `<= AST_TYPE_MAX`.
        // `AstType` is `#[repr(u8)]`, and its discriminants cover the range `0..=AST_TYPE_MAX`,
        // so any `u8` value `<= AST_TYPE_MAX` is a valid `AstType`.
```

I'd also suggest adding a debug assertion `debug_assert!(global_bit_index <= AST_TYPE_MAX as usize)`.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fs, io,
+    path::Path,
+};
+
+use convert_case::{Case, Casing};
+use oxc_ast::AstType;
+use oxc_semantic::AstTypesBitset;
+use syn::{Expr, ExprIf, ExprMatch, File, Pat, Path as SynPath, Stmt}; // keep syn in scope for parse_file used elsewhere
```
overlookmotel: Style nit: Could you possibly follow the import order style outlined in #12723?

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+        }
+    }
+    // Sort deterministically by full path
+    rule_entries.sort_by(|a, b| a.0.cmp(&b.0));
+
+    let mut out = String::new();
+    out.push_str("// Auto-generated code, DO NOT EDIT DIRECTLY!\n");
+    out.push_str("// To regenerate: `cargo run -p oxc_linter_codegen`\n\n");
+    out.push_str("#![allow(clippy::needless_pass_by_value)]\n\n");
+    out.push_str("use oxc_semantic::AstTypesBitset;\n");
```
overlookmotel: Nit. Otherwise `rustfmt` reverses the order of the `use` statements.

```suggestion
    out.push_str("use oxc_semantic::AstTypesBitset;\n\n");
```

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    out.push_str("use crate::rule::RuleRunner;\n\n");
+    let mut needs_ast_type_import = false;
+    let mut impls_buf = String::new();
+
+    // Build a set of valid AstType variant names to validate any textual fallbacks.
+    let _allowed_asttype_variants = get_asttype_variant_names().unwrap_or_default();
```
overlookmotel: This appears to be dead code. Can it be removed?

Presumably if the rule's code contains non-existent `AstKind`s e.g. `AstKind::Monkey` then codegen will generate code like `const NODE_TYPES = &AstTypesBitset::from_types(&[AstType::Monkey])` which will fail to compile, rather than result in a bug. So maybe it's not necessary to check validity of names here anyway?

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+
+    // Resolve variant names to AstType values by reading the generated AstType enum to get discriminants.
+    let Ok(root) = project_root::get_project_root() else { return None };
+    let ast_kind_path = root.join("crates/oxc_ast/src/generated/ast_kind.rs");
+    let Ok(ast_kind_contents) = fs::read_to_string(&ast_kind_path) else { return None };
+    let Ok(ast_file) = syn::parse_file(&ast_kind_contents) else { return None };
+    let mapping = build_asttype_name_to_index_map(&ast_file);
+    if mapping.is_empty() {
+        return None;
+    }
```
overlookmotel: If I've understood right, `detect_top_level_node_types` is called once for each lint rule file. So for each lint file, you're reading and parsing `ast_kind.rs` to create mapping from type name to `AstType`. It'd be better to do this work only once, and pass the mapping into this function as a param.

Personally, I'd avoid using `AstType` and `AstTypesBitset` in the codegen. As mentioned above, I think can get away with just dealing with type names as strings without validating them or doing other clever stuff! But that's my personal opinion, you may feel differently.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+        }
+        write!(
+            impls_buf,
+            "impl RuleRunner for {full_type_path} {{\n    const NODE_TYPES: &AstTypesBitset = &{node_types_init};\n    const ANY_NODE_TYPE: bool = {any_node_type};\n\n    #[inline] fn types_info(&self) -> (&'static AstTypesBitset, bool) {{ (Self::NODE_TYPES, Self::ANY_NODE_TYPE) }}\n}}\n\n"
+        ).unwrap();
+    }
+
+    if needs_ast_type_import {
+        out.push_str("use oxc_ast::AstType;\n\n");
+    }
```
overlookmotel: Conditionally adding this `use` statement looks like an unnecessary complication. Can't we assume that we'll be able to detect the types that at least *one* rule acts on? If not, this whole effort wouldn't have much value!

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+                let Stmt::Expr(Expr::If(ifexpr), _) = stmt else { return None };
+                let mut variants = BTreeSet::new();
+                if !collect_if_chain_variants(ifexpr, &mut variants) {
+                    return None;
+                }
+                return Some(Self { variants });
+            }
+        }
+        None
+    }
```
overlookmotel: Don't we need to also check that the trait being implemented is called `Rule`?

Also, I think this is assuming that each file only contains 1 rule. Can we rely on that?

If a file contained something like this:

```rs
impl Rule for SomeRule {
        if let AstKind::Function(func) = node.kind() {
            // ...
        }
    }
}

impl Rule for SomeOtherRule {
        if let AstKind::Class(class) = node.kind() {
            // ...
        }
    }
}
```

...then I think `SomeOtherRule` will be incorrectly identified as acting on `Function` nodes, not `Class` nodes.

I'd suggest moving more logic into `detect_top_level_node_types` - break the file up into `impl`s, find the *right* `impl` for the specific rule, find `run` method. Then pass just the body of `run` method into both `IfElseKindDetector` and `MatchKindDetector`.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+            }
+        }
+        Pat::Path(ppath) => {
+            if let Some(variant) = astkind_variant_from_path(&ppath.path) {
+                out.insert(variant);
+                true
+            } else {
+                false
+            }
+        }
```
overlookmotel: I don't think this `Pat::Path` branch can ever be taken. `if let AstKind::Function = node.kind()` is not valid syntax - all `AstKind` variants have a field, so it'd need to be `AstKind::Function(_)` (`Pat::TupleStruct`).

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+fn collect_if_chain_variants(ifexpr: &ExprIf, out: &mut BTreeSet<String>) -> bool {
+    // Expect condition to be `if let AstKind::Xxx(..) = node.kind()`.
+    if !extract_variants_from_if_condition(&ifexpr.cond, out) {
+        return false;
+    }
+    // Walk else-if chain.
+    if let Some((_, else_branch)) = &ifexpr.else_branch {
+        match &**else_branch {
+            Expr::If(nested) => collect_if_chain_variants(nested, out),
+            _ => true, // plain else { .. } is fine
```
overlookmotel: I think this should return `false`. If the body of `run` method contains something like this:

```rs
if let AstKind::Function(func) = node.kind() {
    // ...
} else {
    // ...
}
```

...then the rule may act on *any* AST node.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+    }
+}
+
+fn arm_pat_has_wildcard(pat: &Pat) -> bool {
+    match pat {
+        Pat::Wild(_) => true,
+        Pat::Or(orpat) => orpat.cases.iter().any(arm_pat_has_wildcard),
+        _ => false,
+    }
+}
```
overlookmotel: It might be safer to take a whitelist approach rather than blacklist. Are we correctly handing (weird) patterns like?

```rs
match node.kind() {
    kind @ AstKind::Function(_) => { /* ... */ }
    _ => {}
}
```

```rs
macro_rules! some_types {
    () => {
        AstKind::Function(_) | AstKind::Class(_)
    };
}

match node.kind() {
    some_types!() => { /* ... */ }
    _ => {}
}
```

Probably best to bail out on anything we don't understand, rather than assuming that we'll never find such code.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,438 @@
...
+            "impl RuleRunner for {full_type_path} {{\n    const NODE_TYPES: &AstTypesBitset = &{node_types_init};\n    const ANY_NODE_TYPE: bool = {any_node_type};\n\n    #[inline] fn types_info(&self) -> (&'static AstTypesBitset, bool) {{ (Self::NODE_TYPES, Self::ANY_NODE_TYPE) }}\n}}\n\n"
+        ).unwrap();
+    }
+
+    if needs_ast_type_import {
+        out.push_str("use oxc_ast::AstType;\n\n");
+    }
+    out.push_str(&impls_buf);
+
+    let target_path = root.join("crates/oxc_linter/src/rule_runner_impls.rs");
```
overlookmotel: Could the generated file go in a `generated` directory, to match the convention set by `oxc_ast_tools` codegen?

overlookmotel (commented):
I've gone through this PR in some depth, as obviously it's critical that rules don't get skipped when they shouldn't be. I think I've found a couple of potential bugs.

The rest of my comments are less important, but I feel that any simplifications that can be made would be a benefit as they'd make the code easier to understand, so any logic flaws stick out more.

(personally I find `syn` a bit of a nightmare - it's complicated and verbose, which tends to obscure meaning)

crates/oxc_linter/src/rule.rs
```diff
@@ -322,4 +334,83 @@ mod test {
...
+        assert!(!R::ANY_NODE_TYPE, "{} should not have ANY_NODE_TYPE set to true", R::NAME);
+        for node_type in node_types {
+            assert!(
+                types.has(*node_type),
+                "{}: missing {:?} in its NODE_TYPES (this means it will incorrectly skip nodes it needs to lint)",
+                R::NAME,
+                node_type
+            );
+        }
+    }
```
overlookmotel: It'd be safer to check in both directions. i.e.:

1. `types` has bit set for all of `node_types`.
2. `types` *doesn't* have bit set for any types *not* in `node_types`.

The 1st is more critical, but the 2nd would give extra assurance that the codegen is working correctly.

tasks/linter_codegen/src/main.rs
camchenry: Yeah thinking about this more, I realize there isn't much point to validating this here. I think using Strings instead is simpler, because the Rust compiler will check the validity anyway.

tasks/linter_codegen/src/main.rs
camchenry: Good catch! I ended up removing all of this code anyway since we can get away with just using the string name and relying on the Rust compiler to validate that all of the source code generated is correct.

overlookmotel:
I'm part-way through reviewing this. I'll finish review tomorrow.

crates/oxc_linter/src/lib.rs
```diff
@@ -170,7 +172,28 @@ impl Linter {
                 // Collect rules into a Vec so that we can iterate over the rules multiple times
                 let rules = rules.collect::<Vec<_>>();
 
+                // TODO: It seems like there is probably a more intelligent way to preallocate space here. This will
+                // likely incur quite a few unnecessary reallocs currently. We theoretically could compute this at
+                // compile-time since we know all of the rules and their AST node type information ahead of time.
```
overlookmotel, line 177: Yes. I guess the ideal would be to recycle these `Vec`-of-`Vec`s - clear them after each file, and reuse them for next file. Pretty quickly they'll stabilize at a size which is enough to accommodate the data in future iterations, and growth/reallocation will stop.

Unlike statically setting size of `Vec`s to the maximum possible, this approach would be adaptable to different rule sets.

But obviously multi-threading complicates recycling. Thread local storage?

crates/oxc_linter/src/lib.rs
```diff
@@ -170,7 +172,28 @@ impl Linter {
...
+                let mut rules_by_ast_type = vec![Vec::new(); AST_TYPE_MAX as usize + 1];
+                // TODO: Compute needed capacity. This is a slight overestimate as not 100% of rules will need to run on all
+                // node types, but it at least guarantees we won't need to realloc.
+                let mut rules_any_ast_type = Vec::with_capacity(rules.len());
+
                 for (rule, ctx) in &rules {
+                    // Collect node type information for rules. In large files, benchmarking showed it was worth
+                    // collecting rules into buckets by AST node type to avoid iterating over all rules for each node.
```
overlookmotel, line 185: Add here:

```rs
let rule = *rule;
```

Currently `rules_by_ast_type` and `rules_any_ast_type` contain `&&RuleEnum`, which is unnecessary indirection - `&RuleEnum` would be preferable.

crates/oxc_linter/src/lib.rs
```diff
@@ -170,7 +172,28 @@ impl Linter {
...
+                        let (ast_types, all_types) = rule.types_info();
+                        if all_types {
+                            rules_any_ast_type.push((rule, ctx));
+                        } else {
+                            for ty in ast_types {
+                                rules_by_ast_type[ty as usize].push((rule, ctx));
+                            }
+                        }
+                    }
+
```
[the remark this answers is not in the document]
overlookmotel: I think mr bot is kind of right here. Unless I misunderstood, `rules` was already filtered by `rule.should_run` earlier, so it's not necessary to test `rule.should_run` again here.

Where I differ with mr bot is that I don't think it will affect behavior - it's just redundant work. Therefore not a blocker to merging - we can fix it later.

crates/oxc_linter/src/rule.rs
```diff
@@ -66,6 +66,16 @@ pub trait Rule: Sized + Default + fmt::Debug {
...
+pub trait RuleRunner: Rule {
+    /// `AstType`s that rule acts on
+    const NODE_TYPES: &AstTypesBitset;
+    /// `true` if codegen can't figure out what node types rule acts on
+    const ANY_NODE_TYPE: bool;
+
+    fn types_info(&self) -> (&'static AstTypesBitset, bool) {
+        (Self::NODE_TYPES, Self::ANY_NODE_TYPE)
+    }
+}
```
overlookmotel, line 78: We could combine these 2, to make return value of `fn types_info` only 8 bytes, instead of 16.

```suggestion
pub trait RuleRunner: Rule {
    /// `AstType`s that rule acts on.
    /// `None` if codegen can't figure out what node types rule acts on.
    const NODE_TYPES: Option<&AstTypesBitset>;

    fn types_info(&self) -> Option<&'static AstTypesBitset> {
        Self::NODE_TYPES
    }
}
```

(yes I know 2 fields was the design I suggested in the first place!)

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+            let type_idents: Vec<String> =
+                detected_types.into_iter().map(|v| format!("AstType::{v}")).collect();
+            (format!("AstTypesBitset::from_types(&[{}])", type_idents.join(", ")), false)
+        } else {
+            ("AstTypesBitset::new()".to_string(), true)
+        };
+
+        write!(
+            out,
+            "impl RuleRunner for crate::rules::{plugin_module}::{rule_module}::{rule_struct} {{\n    const NODE_TYPES: &AstTypesBitset = &{node_types_init};\n    const ANY_NODE_TYPE: bool = {any_node_type};\n\n}}\n\n",
```
overlookmotel, line 54: It's unfortunate that we can't split this into multiple lines to make it more readable, and rely on `rustfmt` to indent it nicely. But `rustfmt` gives up trying to format it when the `impl` line is really long.

Probably we'd need to also run the code through `prettyplease` before `rustfmt`, similar to how `oxc_ast_tools` does:

https://github.com/oxc-project/oxc/blob/5648b31167086ea118abb8ade05cf83519dd7080/tasks/ast_tools/src/output/rust.rs#L15-L30

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+    // Sort deterministically
+    rule_entries.sort_by(|a, b| {
+        let ord = a.plugin_module_name.cmp(b.plugin_module_name);
+        if ord == std::cmp::Ordering::Equal {
+            a.rule_module_name.cmp(b.rule_module_name)
+        } else {
+            ord
+        }
+    });
```
overlookmotel, line 146: Could simplify this by adding `#[derive(PartialEq, Eq, PartialOrd, Ord)]` on `RuleEntry`, and then using `rule_entries.sort_unstable()`.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+    false
+}
+
+fn extract_variants_from_pat(pat: &Pat, out: &mut BTreeSet<String>) -> CollectionResult {
+    match pat {
+        Pat::Or(orpat) => {
+            let mut result = CollectionResult::Complete;
+            for p in &orpat.cases {
+                if extract_variants_from_pat(p, out) == CollectionResult::Incomplete {
+                    result = CollectionResult::Incomplete;
```
overlookmotel, line 276: Nit. Could `break` the loop here. No point exploring the pattern further.

Or `return CollectionResult::Incomplete` in this branch, and return `CollectionResult::Complete` unconditionally after the loop. No need for the `result` var.

crates/oxc_semantic/src/ast_types_bitset.rs
```diff
@@ -17,10 +19,12 @@ impl AstTypesBitset {
...
-    pub fn from_types(types: &[AstType]) -> Self {
+    pub const fn from_types(types: &[AstType]) -> Self {
         let mut bitset = Self::new();
-        for &ty in types {
-            bitset.set(ty);
+        let mut i = 0;
+        while i < types.len() {
+            bitset.set(types[i]);
+            i += 1;
         }
```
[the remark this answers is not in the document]
overlookmotel: Untrue. `set` *is* a `const` method. And quite evidently it *does* compile.

overlookmotel (approved):
Looks good!

This more conservative approach of only matching the simplest `if` / `else` statements obviously reduces the perf impact a lot, but I do think it's a good approach to start minimal and build on it step-by-step. It'll be easier to verify correctness as add support for more/complex patterns in later PRs.

None of my comments below are blockers. Just some thoughts for future, and I have submitted follow-on PRs for some of them.

tasks/linter_codegen/src/main.rs
```diff
@@ -0,0 +1,330 @@
...
+        if block.stmts.len() != 1 {
+            return None;
+        }
+        let stmt = &block.stmts[0];
+        let Stmt::Expr(Expr::If(ifexpr), _) = stmt else { return None };
+        let mut variants = BTreeSet::new();
+        let result = collect_if_chain_variants(ifexpr, &mut variants);
+        if result == CollectionResult::Incomplete || variants.is_empty() {
+            return None;
+        }
```
overlookmotel, line 214: The pattern of passing `out: &mut BTreeSet<String>` through various free functions is not incorrect, but it's perhaps a little un-idiomatic. I wonder if it'd be preferable to make some of these functions methods of the `IfElseKindDetector` trait instead?

Or, if you want to share some of the logic with other detector structs in future PRs, could add a `Variants` type which wraps a `BTreeSet<String>` and has methods like `extract_variants_from_pat` on it?

As I said in an earlier comment, the problem with `syn` is that it's really verbose and hard to follow (or at least I find it so). So in my view, anything we can do to make this code as structured and clear as possible will be a big help with verifying correctness.

overlookmotel:
I ran ecosystem CI: https://github.com/oxc-project/oxlint-ecosystem-ci/actions/runs/17525661651

Number of errors+warnings matches most recent run on main: https://github.com/oxc-project/oxlint-ecosystem-ci/actions/runs/17524680066

I think we're good!

@camc314 Do you want to have a look before merging? (and make sure I've correctly understood how ecosystem CI works)

overlookmotel:
### Merge activity

* **Sep 7, 7:53 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Sep 7, 7:55 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Sep 8, 3:18 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Sep 8, 3:18 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Sep 8, 3:27 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

crates/oxc_linter/src/lib.rs
overlookmotel: #13574

tasks/linter_codegen/src/main.rs
overlookmotel: #13575

tasks/linter_codegen/src/main.rs
overlookmotel: #13576

camc314 (approved):

PR #13163, opened 2025-08-17
fix(rulegen): strip comments from config

PeterCardenas:
Rule generation created invalid rust due to comments inside the configuration object. Additionally, the format was being called asynchronously which was causing errors to show up after the command finished and interrupted the shell prompt.

This PR fixes both of these issues by stripping the comments from the configuration object before converting it to a JSON literal and waiting for format to finish so that errors will pop up before the command finishes.


camc314 (approved):
thanks

PR #13229, opened 2025-08-20
fix(minifier): fix string concatenation for lone surrogates

h-a-n-a:
Fixes a simple [use case](https://playground.oxc.rs/#eNpNkUFvwyAMhf9K5dOqRdPWqVKVc9Trpu3KIYQ4ERLBkYFuVdX/PhOWbCdseHzvGW5goAZDPpDDJ0fjQ6tUak6vp3b3uFvq5nBs98r/1yj4FSkQVelEpmAPFRDUN+Dk8xKuPupvqCMnrMBZH9c6GJpxa65TR27tImsfBuIJ6kG7gPcKZs0BORO1c/T1gTGxf0sx2B7PyZtoSfzK9ZlRtBd814w+lF1BZPMFIfXmUJhR84iSDDAcnl+OMoOVUXXEvkHjNOuMF9KaxlCPIy4Dotedw09KbHDS8+Y2WW8HW/CGppzpL4o8ZWRyZ5kkn1+QOwryGIWf8y1/UkzuP5yFkg8=) in optimizing lone surrogates. Also initialized and pinpointed some places that need a fix in future PRs.

With input:
```js
'\uD83D' + '\uDD25'
`\uD83D` + `\uDD25`
```

Previously outputs:

```js
"�d838�dd25"
"�d838�dd25"
```

Now outputs:

```js
"\ud83d\udd25"
"\ud83d\udd25"
```

Ported some optimization fixes I made in https://github.com/swc-project/swc/pull/10987 and I'm really appreciate the job you all have done. Without the previous investigation made by @overlookmotel, @Boshen and Oxc team, this will not happen.

### DIFF
--- a/crates/oxc_minifier/tests/peephole/mod.rs
+++ b/crates/oxc_minifier/tests/peephole/mod.rs
@@ -26,3 +26,12 @@ fn test(source_text: &str, expected: &str) {
 fn test_same(source_text: &str) {
     test(source_text, source_text);
 }
+
+#[track_caller]
+fn test_output(source_text: &str, expected: &str) {
+    let options = default_options();
+    let actual = crate::run(source_text, oxc_span::SourceType::default(), Some(options))
+        .trim_end()
+        .to_string();
+    assert_eq!(actual, expected, "\nfor source\n{source_text}\nexpect\n{expected}\ngot\n{actual}");
+}
--- a/crates/oxc_ecmascript/src/array_join.rs
+++ b/crates/oxc_ecmascript/src/array_join.rs
@@ -11,7 +11,15 @@ pub trait ArrayJoin<'a> {
 impl<'a> ArrayJoin<'a> for ArrayExpression<'a> {
     fn array_join(&self, ctx: &impl GlobalContext<'a>, separator: Option<&str>) -> Option<String> {
         let strings = self.elements.iter().map(|e| e.to_js_string(ctx)).collect::<Option<Vec<_>>>();
-        strings
-            .map(|v| v.iter().map(AsRef::as_ref).collect::<Vec<_>>().join(separator.unwrap_or(",")))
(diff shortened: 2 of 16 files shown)


Boshen:
This is a lot of code to support lone surrogates, I'll take a deeper look later.

h-a-n-a:
> This is a lot of code to support lone surrogates, I'll take a deeper look later.

@Boshen Thanks! That would be great and I'd love to fix all the `TODO`s I listed in the PR, probably in the future PRs. If you need any help, just let me know ;-)

crates/oxc_minifier/tests/peephole/mod.rs
```diff
@@ -26,3 +26,12 @@ fn test(source_text: &str, expected: &str) {
 fn test_same(source_text: &str) {
     test(source_text, source_text);
 }
+
+#[track_caller]
+fn test_output(source_text: &str, expected: &str) {
```
h-a-n-a, line 31: Added a standalone test runner. 

The rationale behind this is that the original test runner would generate `actual: "\ud83d\udd25"` and `expected: "🔥"` if I write 

```
test("return '\\uD83D' + '\\uDD25'", "return '\\ud83d\\udd25'")
```

Although the execution result in JavaScript runtime is the same, it's not the same as it compares with each other in Rust. The `expected` half will be interpreted as a surrogate pair in Rust and the `actual` half will be an escaped surrogate pair.

h-a-n-a:
FYI: we've replaced our implementation here in swc with `Wtf8Buf`.

See:

1. https://github.com/swc-project/swc/pull/11104
2. https://github.com/swc-project/swc/pull/11144

Boshen:
We'll fix this properly in https://github.com/oxc-project/oxc/issues/16886

PR #13233, opened 2025-08-20
feat(linter/no-native): implement promise rule no-native

lambda47:
Related #4655


camc314 (commented):
wouldn't this be caught by eslint/no-undef rule?

lambda47:
> wouldn't this be caught by eslint/no-undef rule?

Yes, I read eslint/no-undef rule, it can cover this rule. we don't need this rule.

PR #13257, opened 2025-08-22
feat(oxfmt): Initialize apps/oxfmt crate with minimum formatter usage

leaysgur:
Part of #10573

- Add `apps/oxfmt` crate
- Copy minimum functionalities from `apps/oxlint`
  - Currently, these two are very similar
  - However, as following Prettier more, they may become different
  - There may still have in common, but at this point, it is unclear
- Format files with `oxc_formatter` and save
  - Usage not optimized


leaysgur:
Finally, I'm trying Graphite...! #13259

PR #12438, opened 2025-07-21
test(formatter): check performance

Dunqing:
Do some experiments in this PR to see how the performance goes

PR #13266, opened 2025-08-23
perf(formatter): implement buffer pre-allocation optimization

taearls:
## Summary

This PR implements buffer pre-allocation optimizations for the formatter to improve performance by reducing memory allocations during the formatting process.

## Motivation

The formatter previously allocated memory dynamically as needed during formatting, which caused frequent reallocations and memory copies. Profiling showed that these allocations were a significant performance bottleneck, especially for large files.

### 1. Buffer Pre-allocation in VecBuffer

Added capacity estimation and pre-allocation to the main formatting buffer:

- **Location**: `src/formatter/mod.rs`
- **Implementation**: Pre-allocates buffer with 1.2x source text size + 1KB overhead
- **Rationale**: Analysis of real-world code showed formatted output is typically 1.1-1.3x the source size

### 2. String Buffer Pre-allocation in Printer

Extended pre-allocation to the printer's string buffer:

- **Location**: `src/formatter/printer/mod.rs`
- **Implementation**: Added `with_capacity()` constructor that accepts initial capacity
- **Rationale**: Ensures the final string construction also benefits from pre-allocation

## Performance Impact

Benchmark results show consistent improvements across all test files:

| File | Before | After | Improvement |
|------|--------|-------|-------------|
| RadixUIAdoptionSection.jsx | 28.644 µs | 26.893 µs | **6.1%** |
| react.development.js | 1.2721 ms | 1.1923 ms | **6.3%** |
| cal.com.tsx | 20.866 ms | 19.562 ms | **6.3%** |
| binder.ts | 2.6692 ms | 2.5046 ms | **6.2%** |

**Average improvement: 6.2%**

### Capacity Calculation

The capacity formula `(source_text.len() * 1.2) + 1024` was chosen based on:

1. **Empirical observation**: Most formatted code expands by 10-30% due to:
   - Added whitespace and indentation
   - Line breaks for readability
   - Additional punctuation (commas, semicolons)

2. **Safety margin**: The 1KB additional buffer accounts for:
   - Edge cases with heavy formatting
   - Small files that might have higher expansion ratios

### Memory Trade-offs

- **Pros**: 
  - Eliminates reallocation overhead
  - Reduces memory fragmentation
  - Improves cache locality

- **Cons**: 
  - Slightly higher initial memory usage
  - May over-allocate for files that shrink during formatting

The trade-off favors performance over minimal memory usage, which aligns with formatter goals where speed is critical.

## Testing

All existing tests pass:
- Unit tests: ✓
- Integration tests: ✓
- Prettier conformance tests: ✓

No functional changes were made; this is purely a performance optimization.

## Future Optimizations

This PR lays groundwork for potential future improvements:
- Adaptive capacity estimation based on file type
- Reusable buffer pools for multiple formatting operations
- More granular pre-allocation for specific formatting structures

## Review Notes

- The changes are isolated to buffer allocation logic
- No changes to formatting rules or output
- Backwards compatible with existing API

### DIFF
--- a/crates/oxc_formatter/src/formatter/mod.rs
+++ b/crates/oxc_formatter/src/formatter/mod.rs
@@ -80,11 +80,16 @@ use self::{format_element::document::Document, group_id::UniqueGroupIdBuilder, p
 pub struct Formatted<'a> {
     document: Document<'a>,
     context: FormatContext<'a>,
+    estimated_capacity: usize,
 }
 
 impl<'a> Formatted<'a> {
     pub fn new(document: Document<'a>, context: FormatContext<'a>) -> Self {
-        Self { document, context }
+        Self { document, context, estimated_capacity: 0 }
+    }
+    
+    pub fn new_with_capacity(document: Document<'a>, context: FormatContext<'a>, capacity: usize) -> Self {
+        Self { document, context, estimated_capacity: capacity }
     }
 
     /// Returns the context used during formatting.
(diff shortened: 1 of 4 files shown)


Dunqing:
Thanks for working on improving performance. We had the same approach to improve performance in #12438, but we haven't looked deeply into what the balance capacity is that we should pre-reserve. In that PR, we pre-reserve the length of the source text, and that shows the performance improvement better than this, and we still think that it's too much. 

Close for now. We will start working on improving performance when everything is working fine.

PR #13327, opened 2025-08-27
fix(transformer/legacy-decorator): emit correct metadata types for enum

### LINKED ISSUE #13319: transformer: include inferred type for enums in the same module for `emitDecoratorMetadata`
For enums in the same module, TypeScript outputs the types in `__metadata` ([TS playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&isolatedModules=true&verbatimModuleSyntax=true&useDefineForClassFields=true&noEmitHelpers=true#code/KYOwrgtgBA8gDgFwJYHsQGcoG8BQUoBmKKUAvFAOToIBOSIA5hTgL447AAecKNCUAYwA2AQ3SYAYsWx4oAAQAmwAbxEJgACgCUsoig0pEqDAC5YRtOi1n4yS9jazFy1eu2yARiJo2LGVjhAA), in this case `String`) even for `isolatedModules: true`.
```ts
enum Options {
  foo = 'string'
}

export class Foo {
  @decorate()
  foo(options: Options): Options {}

  @decorate()
  bar: Options
}
```
But Oxc does not output the types and falls back to the generic `typeof (_ref = typeof Options !== "undefined" && Options) === "function" ? _ref : Object` ([reproduction using transform function exposed from rolldown](https://stackblitz.com/edit/github-xv4yss-fvqn9tsn?file=package.json,build.mjs)).

Dunqing:
* close:  #13319

### DIFF
--- a/crates/oxc_transformer/src/decorator/legacy/metadata.rs
+++ b/crates/oxc_transformer/src/decorator/legacy/metadata.rs
@@ -137,7 +137,8 @@ impl<'a, 'ctx> LegacyDecoratorMetadata<'a, 'ctx> {
 }
 
 impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
-    // `#[inline]` so compiler knows `stmt` is a `TSEnumDeclaration`
+    // `#[inline]` because this is a hot path and most `Statement`s are not `TSEnumDeclaration`s.
+    // We want to avoid overhead of a function call for the common case.
     #[inline]
     fn enter_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
         // Collect enum types here instead of in `enter_ts_enum_declaration` because the TypeScript
--- a/crates/oxc_transformer/src/decorator/legacy/metadata.rs
+++ b/crates/oxc_transformer/src/decorator/legacy/metadata.rs
@@ -92,9 +92,10 @@ use std::collections::VecDeque;
 use oxc_allocator::{Box as ArenaBox, TakeIn};
 use oxc_ast::ast::*;
 use oxc_data_structures::stack::NonEmptyStack;
-use oxc_semantic::ReferenceFlags;
+use oxc_semantic::{Reference, ReferenceFlags, SymbolId};
 use oxc_span::{ContentEq, SPAN};
 use oxc_traverse::{MaybeBoundIdentifier, Traverse};
+use rustc_hash::FxHashMap;
 
 use crate::{
     Helper,
@@ -102,6 +103,17 @@ use crate::{
     state::TransformState,
     utils::ast_builder::create_property_access,
 };
+
+/// Type of an enum inferred from its members
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum EnumType {
+    /// All members are string literals or template literals with string-only expressions
+    String,
+    /// All members are numeric, bigint, unary numeric, or auto-incremented
+    Number,
+    /// Mixed types or computed values
+    Object,
+}
 
 pub enum MethodMetadata<'a> {
     Constructor(Expression<'a>),
@@ -111,15 +123,31 @@ pub enum MethodMetadata<'a> {
 pub struct LegacyDecoratorMetadata<'a, 'ctx> {
     ctx: &'ctx TransformCtx<'a>,
     metadata_stack: NonEmptyStack<VecDeque<MethodMetadata<'a>>>,
+    enum_types: FxHashMap<SymbolId, EnumType>,
 }
 
 impl<'a, 'ctx> LegacyDecoratorMetadata<'a, 'ctx> {
     pub fn new(ctx: &'ctx TransformCtx<'a>) -> Self {
-        LegacyDecoratorMetadata { ctx, metadata_stack: NonEmptyStack::new(VecDeque::new()) }
+        LegacyDecoratorMetadata {
+            ctx,
+            metadata_stack: NonEmptyStack::new(VecDeque::new()),
+            enum_types: FxHashMap::default(),
+        }
     }
 }
 
 impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
+    // `#[inline]` so compiler knows `stmt` is a `TSEnumDeclaration`
+    #[inline]
+    fn enter_statement(&mut self, stmt: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        // Collect enum types here instead of in `enter_ts_enum_declaration` because the TypeScript
+        // plugin transforms enum declarations in `enter_statement`, and we need to collect the
+        // enum type before it gets transformed.
+        if let Statement::TSEnumDeclaration(decl) = stmt {
+            self.collect_enum_type(decl, ctx);
+        }
+    }
+
     fn enter_class(&mut self, class: &mut Class<'a>, ctx: &mut TraverseCtx<'a>) {
         if class.is_expression() || class.declare {
             return;
@@ -201,6 +229,63 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
 }
 
 impl<'a> LegacyDecoratorMetadata<'a, '_> {
+    /// Collects enum type information for decorator metadata generation.
+    fn collect_enum_type(&mut self, decl: &TSEnumDeclaration<'a>, ctx: &TraverseCtx<'a>) {
+        let symbol_id = decl.id.symbol_id();
+
+        // Optimization:
+        // If the enum doesn't have any type references, that implies that no decorators
+        // refer to this enum, so there is no need to infer its type.
+        let has_type_reference =
+            ctx.scoping().get_resolved_references(symbol_id).any(Reference::is_type);
+        if has_type_reference {
+            let enum_type = Self::infer_enum_type(&decl.body.members);
+            self.enum_types.insert(symbol_id, enum_type);
+        }
+    }
+
+    /// Check if an expression is a numeric expression (including unary expressions)
+    fn is_numeric_expression(expr: &Expression<'a>) -> bool {
+        expr.is_number_literal()
+            || matches!(
+                expr, Expression::UnaryExpression(unary) if
+                matches!(
+                    // These operators still produce numeric results.
+                    unary.operator, UnaryOperator::UnaryNegation | UnaryOperator::UnaryPlus | UnaryOperator::BitwiseNot
+                ) && unary.argument.is_number_literal()
+            )
+    }
+
+    /// Infer the type of an enum based on its members
+    fn infer_enum_type(members: &[TSEnumMember<'a>]) -> EnumType {
+        let mut enum_type = EnumType::Object;
+
+        for member in members {
+            if let Some(init) = &member.initializer {
+                match init {
+                    Expression::StringLiteral(_) | Expression::TemplateLiteral(_)
+                        if enum_type != EnumType::Number =>
+                    {
+                        enum_type = EnumType::String;
+                    }
+                    expr if Self::is_numeric_expression(expr) && enum_type != EnumType::String => {
+                        enum_type = EnumType::Number;
+                    }
+                    // For other expressions, we can't determine the type statically
+                    _ => return EnumType::Object,
+                }
+            } else {
+                // No initializer means numeric (auto-incrementing from previous member)
+                if enum_type == EnumType::String {
+                    return EnumType::Object;
+                }
+                enum_type = EnumType::Number;
+            }
+        }
+
+        enum_type
+    }
+
     pub fn pop_method_metadata(&mut self) -> Option<MethodMetadata<'a>> {
         self.metadata_stack.last_mut().pop_front()
     }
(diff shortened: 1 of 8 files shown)


Dunqing:
* **#13346** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13346?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13327** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13327?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13327?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=13327).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,74 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
 impl<'a> LegacyDecoratorMetadata<'a, '_> {
+    /// Collects enum type information for decorator metadata generation.
+    fn collect_enum_type(&mut self, decl: &TSEnumDeclaration<'a>, ctx: &TraverseCtx<'a>) {
+        let symbol_id = decl.id.symbol_id();
+
+        // Optimization:
+        // If the enum doesn't have any type references, it is said that no decorator will
+        // refer to this enum, so there is no need to infer its type.
+        let has_type_reference =
+            ctx.scoping().get_resolved_references(symbol_id).any(Reference::is_type);
```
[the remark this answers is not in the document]
Dunqing: We can't cache it, and we only look it up once for each enum.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -111,15 +123,26 @@ pub enum MethodMetadata<'a> {
...
+        }
     }
 }
 
 impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
+    fn enter_statement(&mut self, node: &mut Statement<'a>, ctx: &mut TraverseCtx<'a>) {
+        if let Statement::TSEnumDeclaration(decl) = node {
+            self.collect_enum_type(decl, ctx);
+        }
+    }
```
overlookmotel: Any reason not to use `enter_ts_enum_declaration` here?

```suggestion
    fn enter_ts_enum_declaration(&mut self, decl: &mut TSEnumDeclaration<'a>, ctx: &mut TraverseCtx<'a>) {
        self.collect_enum_type(decl, ctx);
    }
```

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,71 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
 }
 
 impl<'a> LegacyDecoratorMetadata<'a, '_> {
+    /// Collects enum type information for decorator metadata generation.
+    fn collect_enum_type(&mut self, decl: &TSEnumDeclaration<'a>, ctx: &TraverseCtx<'a>) {
+        let symbol_id = decl.id.symbol_id();
+
+        // Optimization:
+        // If the enum doesn't have any type references, it is said that no decorator will
+        // refer to this enum, so there is no need to infer its type.
```
overlookmotel:
```suggestion
        // If the enum doesn't have any type references, that implies that no decorators
        // refer to this enum, so there is no need to infer its type.
```

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,71 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
+            ctx.scoping().get_resolved_references(symbol_id).any(Reference::is_type);
+        if has_type_reference {
+            let enum_type = Self::infer_enum_type(&decl.body.members);
+            self.add_enum_type(symbol_id, enum_type);
+        }
+    }
+
+    fn add_enum_type(&mut self, symbol_id: SymbolId, enum_type: EnumType) {
+        self.enum_types.insert(symbol_id, enum_type);
+    }
```
overlookmotel: Style nit: This function is only 1 line. You could inline it into `collect_enum_type`.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,71 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
+                        has_number = true;
+                    }
+                    // For other expressions, we can't determine the type statically
+                    _ => {
+                        return EnumType::Object;
+                    }
+                }
+            } else {
+                // No initializer means numeric (auto-incrementing from previous member)
+                has_number = true;
```
overlookmotel: You could bail out early and return `EnumType::Object` here if `has_string == true`.

Then the whole function could be simplified, replacing `has_string` and `has_number` with just `enum_type`:

```rs
let mut enum_type = EnumType::Object;
for member in members {
    if let Some(init) = &member.initializer {
        match init {
            Expression::StringLiteral(_) if enum_type != EnumType::Number => {
                enum_type = EnumType::String;
            }
            // ...
            _ => return EnumType::Object,
        }
    } else {
        // No initializer means numeric (auto-incrementing from previous member)
        if enum_type == EnumType::String {
            return EnumType::Object;
        }
        enum_type = EnumType::Number;
    }
}
enum_type
```

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,71 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
+                match init {
+                    Expression::StringLiteral(_) if !has_number => {
+                        has_string = true;
+                    }
+                    Expression::TemplateLiteral(literal)
+                        if !has_number
+                            && literal
+                                .expressions
+                                .iter()
+                                .all(|expr| matches!(expr, Expression::StringLiteral(_))) =>
```
overlookmotel: Is it relevant what type the `expressions` are? All `TemplateLiteral`s evaluate to a string, no matter what the expressions are.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -350,6 +438,20 @@ impl<'a> LegacyDecoratorMetadata<'a, '_> {
         name: &TSTypeName<'a>,
         ctx: &mut TraverseCtx<'a>,
     ) -> Expression<'a> {
+        // Check if this is an enum type reference - if so, return the primitive type directly
+        if let TSTypeName::IdentifierReference(ident) = name {
+            let binding = MaybeBoundIdentifier::from_identifier_reference(ident, ctx);
+            if let Some(symbol_id) = binding.symbol_id {
```
overlookmotel: `MaybeBoundIdentifier` is overkill here.

```suggestion
            let symbol_id = ctx.scoping().get_reference(ident.reference_id()).symbol_id();
            if let Some(symbol_id) = symbol_id {
```

If `MaybeBoundIdentifier::from_identifier_reference` is inlined, it will make no difference, but I'm not sure if it will be inlined.

crates/oxc_transformer/src/decorator/mod.rs
```diff
@@ -81,6 +87,17 @@ impl<'a> Traverse<'a, TransformState<'a>> for Decorator<'a, '_> {
...
+    #[inline]
+    fn enter_ts_enum_declaration(
+        &mut self,
+        node: &mut TSEnumDeclaration<'a>,
+        ctx: &mut TraverseCtx<'a>,
+    ) {
+        if self.options.legacy {
+            self.legacy_decorator.enter_ts_enum_declaration(node, ctx);
+        }
+    }
```
overlookmotel: This code is extraneous. There is no `enter_ts_enum_declaration` method on `LegacyDecorator` (though I suggest switching `enter_statement` for `enter_ts_enum_declaration` - see comment above).

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -201,6 +224,71 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
+                        has_string = true;
+                    }
+                    Expression::NumericLiteral(_) | Expression::BigIntLiteral(_) if !has_string => {
+                        has_number = true;
+                    }
+                    Expression::UnaryExpression(expr)
+                        if expr.argument.is_number_literal() && !has_string =>
+                    {
+                        has_number = true;
+                    }
```
overlookmotel: Not all `UnaryExpression`s evaluate to a number. It depends on the operator:

* `!123` -> boolean
* `typeof 123` -> string
* `void 123` -> undefined
* `delete 123` -> boolean

The first one is not entirely stupid - `!0` and `!1` are sometimes used as shortcuts for `true` / `false`.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: The TypeScript plugin transforms enums in `enter_statement`, but we need to collect enum types before the transformation occurs.

https://github.com/oxc-project/oxc/blob/6ff27b4d31ff2520836d3c046df90b01c1f68c4f/crates/oxc_transformer/src/typescript/enum.rs#L38-L40

crates/oxc_transformer/src/decorator/mod.rs
Dunqing: See https://github.com/oxc-project/oxc/pull/13327#discussion_r2307026118

Dunqing: I forgot to remove this.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
overlookmotel: Ah ha! Makes sense. Maybe add a comment explaining that?

overlookmotel: It may be a benefit to add `#[inline]`. Most statements aren't `TSEnumDeclaration`s, so want to handle the common case without a function call. (compiler may have chosen to inline it already, but probably best to make sure)

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: Good catch, I forgot the importance of the operator. In this case, we only care about `+`, `-` and `~`. For the `typeof 123`, it isn't allowed. See [TypeScript Playground](https://www.typescriptlang.org/play/?strict=false&useDefineForClassFields=true&noEmitHelpers=true&experimentalDecorators=true&emitDecoratorMetadata=true&isolatedModules=false&verbatimModuleSyntax=false#code/KYOwrgtgBA8gDgFwJYHsQGcoG8BQUoBmKKUAvFAgJ5zAoFQCCOOAvs8AB5woBOCUAYwA2AQ3SYAYsWx4oAAQAmwAbxEJgACgCUsoig0pEqDAC5YRtOi1n4yS9jazFy1eu2yARiJ42LGVjhAA)

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: Wow, sounds good!

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: Yes, you are right, I misunderstood this.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: I added a comment and marked it as `#[inline]`.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
overlookmotel: > For the `typeof 123`, it isn't allowed. See [TypeScript Playground](https://www.typescriptlang.org/play/?strict=false&useDefineForClassFields=true&noEmitHelpers=true&experimentalDecorators=true&emitDecoratorMetadata=true&isolatedModules=false&verbatimModuleSyntax=false#code/KYOwrgtgBA8gDgFwJYHsQGcoG8BQUoBmKKUAvFAgJ5zAoFQCCOOAvs8AB5woBOCUAYwA2AQ3SYAYsWx4oAAQAmwAbxEJgACgCUsoig0pEqDAC5YRtOi1n4yS9jazFy1eu2yARiJ42LGVjhAA)

Ah ha. Does this mean we can assume any computed expression (e.g. `foo()`) evaluates to a number?

But maybe the gain of that is not worth the trouble. Some more complex expressions that evaluate to strings seem to be allowed e.g. `enum E { X = 'foo' + 'bar' + 'qux' }`. So we'd need to handle those to accurately determine whether string or number.

overlookmotel (approved):
Pushed a commit just to update the comment about `#[inline]` - hope you don't mind, I thought it'd save time vs another round of review. Will merge as soon as CI passes.

PR #13368, opened 2025-08-28
feat(parser): report more invalid modifier locations

### LINKED ISSUE #11713: Forbid invalid modifiers
Some examples:

```ts
interface Foo {
  abstract method();
}
```

```ts
type Foo = {static bar};
```

```ts
function foo(private parameter) {}
```

```ts
class Foo {
      declare get getter() {}
}
```

More cases can be found in https://github.com/prettier/prettier/blob/main/tests/format/misc/errors/typescript/modifiers/format.test.js

ulrichstark:
### Changes

- Added all 22 examples of #11713 to `tasks/coverage/misc/fail`
- Added `Default` and `Export` to modifiers
- Added three new TS diagnostics: 1029, 1031, and 1070
- Split off `FunctionKind::Constructor` from `FunctionKind::ClassMethod` to differentiate when verifying modifiers of formal parameter
- Added multiple checks to verify modifiers

### Results
- "Negative Passed" percentage of `parser_babel.snap` and `parser_typescript.snap` increased
- Better diagnostic messages in both snapshots
- All new failing examples of original issue do really fail
- Closes #11713


ulrichstark:
I'm unsure about the failing allocations CI check... I did run the command locally, committed my changes but it still failed. Just reset back and force pushed to remove the attempt of making this CI check happy.

Boshen:
Great work!

PR #13526, opened 2025-09-02
fix(parser): improve diagnostics around modifier checks

### LINKED ISSUE #11713: Forbid invalid modifiers

ulrichstark:
Follow up to #13368.
Closes #11713 by fixing five more cases mentioned in https://github.com/oxc-project/oxc/issues/11713#issuecomment-3239687597.

Other changes:
- Added TS(1024) diagnostic
- Fixed the broken oxc-11713-13.ts test file
- Made TS(1070) not report redundantly if it's actually the responsibility of TS(1071)
- Branch off the more specific TS(1028) diagnostic from the general TS(1030) diagnostic like TypeScript does
- Remove incorrect code to parse modifiers on typescript this parameter


Boshen:
Thank you for the nice work!

PR #13533, opened 2025-09-03
perf(parser): optimize parsing modifiers by avoiding redundant checks

ulrichstark:
I refactored the code inside `modifiers.rs` step by step to avoid many redundant checks. It's now also shorter and easier to read and reason about in my opinion. And while I was at it, I properly branched off the more specific TS(1028) diagnostic from the general TS(1030) diagnostic like TypeScript does. Allocations also went down a bit.


ulrichstark:
Readability, allocations and conformance to TypeScript might be better, but I failed to reach the original goal of better performance. It's actually a bit worse now... Probably because of my use of `self.lexer.peek_token` instead of previous `self.try_parse`. `peek_token` might be more lightweight, but `try_parse` avoids the `rewind` and the follow up `bump_any` on the happy path. I will close this PR to save anyone the review effort. If anyone is reading this and would still have this change merged, I can re-open if you want.

Edit: I thought about this again and I will probably revive this PR in the following days and improve it.

PR #13560, opened 2025-09-06
perf(data_structures): add optimized str trim utils

camchenry:
Not sure if this is going anywhere, mainly just trying to get some practice in with optimizations and assembly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1920,6 +1920,7 @@ dependencies = [
  "num-traits",
  "oxc_allocator",
  "oxc_ast",
+ "oxc_data_structures",
  "oxc_span",
  "oxc_syntax",
 ]
--- a/crates/oxc_codegen/Cargo.toml
+++ b/crates/oxc_codegen/Cargo.toml
@@ -22,7 +22,7 @@ doctest = true
 [dependencies]
 oxc_allocator = { workspace = true }
 oxc_ast = { workspace = true }
-oxc_data_structures = { workspace = true, features = ["code_buffer", "slice_iter", "stack"] }
+oxc_data_structures = { workspace = true, features = ["code_buffer", "slice_iter", "stack", "string_utils"] }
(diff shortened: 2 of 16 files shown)


camchenry:
* **#13560** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13560?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13560?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=13560).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Interesting! It will be interesting to see if this makes a difference.

One other possible optimization is to check for non-ASCII characters in batches of 8 or 16 bytes, by splitting the string into chunks with `[u8]::chunks_exact`. And then something like: https://godbolt.org/z/GnoqbKnrq

(conveniently, SIMD movemask op takes the top bit of each byte in a set of bytes - if the result is 0, then no bytes have their top bit set i.e. all bytes in the batch are ASCII)

So you can check a batch of bytes for non-ASCII in 1 shot and divert to cold path if *any* of them are non-ASCII. Then in the loop over the individual bytes, you can skip the `byte.is_ascii()` check.

Note: The choice of batch size (8 or 16) depends on how long you expect strings to be. You can only use the optimized batched version if there's enough bytes to form a batch, and have to fallback to byte-by-byte checking for the odd-numbered "tail".

I'm not sure if that'll help or not in this case, because probably most trim operations trim fairly few (or none at all) bytes. But since it sounds like you're interested in these kinds of optimizations, I thought I'd mention it.

PR #13604, opened 2025-09-09
fix(language_server): don't resend diagnostic on save, when `typeAware` is disabled and run is onType

yongsk0066:
Fixes an edge case introduced in #13332 where diagnostics disappear when saving a file with `oxc.lint.run: "onType"` and `oxc.typeAware: false`.

## Problem

The logic from #13332 assumes `tsgolint` always exists when `Run::OnType`:
```rust
(ServerLinterRun::OnSave, Run::OnType) => (false, true)  // always run tsgolint
```

But when `type_aware: false`, there is no `tsgo_linter`, causing the server to return empty diagnostics instead of None, which clears all error highlights in VSCode.

clear-diagnostics

## Solution

Check if `tsgo_linter` actually exists:
```rust
(ServerLinterRun::OnSave, Run::OnType) => {
    let should_run_tsgo = self.tsgo_linter.as_ref().is_some();
    (false, should_run_tsgo)
}
```

Now returns None when no linters should run, preserving existing diagnostics.

**Test:** Added `test_lint_on_run_on_type_on_save_without_type_aware` with fixture

### DIFF
--- a/crates/oxc_language_server/fixtures/linter/lint_on_run/on_type/on-save-no-type-aware.ts
+++ b/crates/oxc_language_server/fixtures/linter/lint_on_run/on_type/on-save-no-type-aware.ts
@@ -1,0 +1,7 @@
+debugger;
+
+async function returnsPromise() {
+  return "value";
+}
+
+returnsPromise().then(() => {});
--- a/crates/oxc_language_server/src/linter/server_linter.rs
+++ b/crates/oxc_language_server/src/linter/server_linter.rs
@@ -331,9 +331,12 @@ impl ServerLinter {
             (ServerLinterRun::OnType, Run::OnType) => (true, false),
             // it does not match, run nothing
             (ServerLinterRun::OnType, Run::OnSave) => (false, false),
-            // run only tsglint on save, even if the user wants it with type
-            // tsgolint only supports the OS file system.
-            (ServerLinterRun::OnSave, Run::OnType) => (false, true),
+            // In onType mode, only TypeScript type checking runs on save
+            // If type_aware is disabled (tsgo_linter is None), skip everything to preserve diagnostics
+            (ServerLinterRun::OnSave, Run::OnType) => {
+                let should_run_tsgo = self.tsgo_linter.as_ref().is_some();
+                (false, should_run_tsgo)
+            }
         };
 
         // return `None` when both tools do not want to be used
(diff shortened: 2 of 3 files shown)


Sysix:
Thank you for looking after the bug.

> But when type_aware: false, there is no tsgo_linter, causing the server to return empty diagnostics instead of None, which clears all error highlights in VSCode.

My guess is, that you are still on the latest release, and not using the `main` language server.
This should been already fixed in #13490 with an integration test in the editor.

The lines:
https://github.com/oxc-project/oxc/blob/78a1146c3385e5b4f81ecad2127809e51c4ea8a9/crates/oxc_language_server/src/linter/server_linter.rs#L356-L361

It will only insert it to `ServerLinterDiagnostics`. It will still return the `oxlint` diagnostics on save. 


If you want to try it yourself, you can follow the instructions here:
https://oxc.rs/docs/contribute/vscode.html#building-extension-and-installing-it-locally

yongsk0066:
> Thank you for looking after the bug.
> 
> > But when type_aware: false, there is no tsgo_linter, causing the server to return empty diagnostics instead of None, which clears all error highlights in VSCode.
> 
> My guess is, that you are still on the latest release, and not using the `main` language server. This should been already fixed in #13490 with an integration test in the editor.
> 
> The lines:
> 
> https://github.com/oxc-project/oxc/blob/78a1146c3385e5b4f81ecad2127809e51c4ea8a9/crates/oxc_language_server/src/linter/server_linter.rs#L356-L361
> 
> It will only insert it to `ServerLinterDiagnostics`. It will still return the `oxlint` diagnostics on save.
> 
> If you want to try it yourself, you can follow the instructions here: [oxc.rs/docs/contribute/vscode.html#building-extension-and-installing-it-locally](https://oxc.rs/docs/contribute/vscode.html#building-extension-and-installing-it-locally)

Ah, I see it was already fixed. Thanks! I'll check more carefully next time. Thanks for the review. I'll go ahead and close this PR.

yongsk0066:
Hey @Sysix, could you review this once more? I tested and my fix also seems valid.
it takes a different approach by returning None when no linting is needed, which reduces memory usage and network calls. Would this be a viable alternative?

Sysix (approved):
Yes, this is optimization for `onSave` configuration. 
If `typeAware` is not enabled, your MR will not send the latest (redundant) diagnostics to the client.

camc314 (approved):

PR #13632, opened 2025-09-10
fix(transformer): improve legacy decorator handling and fix constructor parameter decorators

### LINKED ISSUE #6079: [Bug]: Decorators/Constructor: method metadata leaks to class `beta.36`
### Reproduction link or steps

[Reproduction](https://stackblitz.com/edit/rolldown-rolldown-starter-stackblitz-vq4vutgx?file=package.json)

Run 
```
pnpm i
pnpm build 
pnpm start
```
`beta.36` version is going to produce an error

### Output Diff

Image

### What is expected?

Both should run successfully.

### What is actually happening?

`beta.36` is broken:

```bash
 rolldown36-di-repro-min@0.0.1 start /home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx
> pnpm run start:35 && pnpm run start:36


> rolldown36-di-repro-min@0.0.1 start:35 /home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx
> node dist-compare/entry-35.js

prob.run(): ok

> rolldown36-di-repro-min@0.0.1 start:36 /home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx
> node dist-compare/entry-36.js

TypeError: can't access property "getInfo", this.b is undefined
    at run (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:1077:4)
    at run (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:1113:17)
    at runProblem (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:1090:43)
    at entry.ts (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:1131:2)
    at <anonymous> (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:18:55)
    at  (file:///home/projects/rolldown-rolldown-starter-stackblitz-vq4vutgx/dist-compare/entry-36.js:1135:1)
    at _0x370799 (https://rolldownrolldownstarterstackbl-dkme.w-corp-staticblitz.com/blitz.96435430.js:31:827521)
    at run (https://rolldownrolldownstarterstackbl-dkme.w-corp-staticblitz.com/blitz.96435430.js:31:828286)
    at _0x1b2515 (https://rolldownrolldownstarterstackbl-dkme.w-corp-staticblitz.com/blitz.96435430.js:31:843394)
    at _evaluate (https://rolldownrolldownstarterstackbl-dkme.w-corp-staticblitz.com/blitz.96435430.js:31:843723)
    at evaluate (https://rolldownrolldownstarterstackbl-dkme.w-corp-staticblitz.com/blitz.96435430.js:31:840218)
    at run (node:internal/modules/esm/module_job:158:3371)

Node.js v20.19.1
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 1.
```

### System Info

```Shell
System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 378.73 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.18.0 - ~/.proto/shims/node
    Yarn: 4.6.0 - ~/.proto/shims/yarn
    npm: 10.9.2 - ~/.proto/shims/npm
    pnpm: 10.9.0 - ~/.proto/shims/pnpm
    bun: 1.2.2 - ~/.proto/shims/bun
  Browsers:
    Chrome: 140.0.7339.81
    Edge: 140.0.3485.54
    Safari: 18.6
```

### Any additional comments?

Sorry to be that guy, but it seems that experimental decorators + metadata are causing problems again 😅 

In the 36 output, the class `Problem` is decorated with class-level metadata that says its constructor has zero params -  `__decorateMetadata("design:paramtypes", [])`. This is wrong for a subclass that relies on DI via the base constructor.

Not sure if this is a Rolldown regression or an OXC one, but it might be related to [this](https://github.com/oxc-project/oxc/issues/13284). Hard to pinpoint what caused this as there was a big jump in OXC.

Dunqing:
Fixes https://github.com/rolldown/rolldown/issues/6079. Regressed introduced by https://github.com/oxc-project/oxc/pull/13227.

The root cause is that the class plugin will insert a constructor method in some cases, but metadata handling is processed before the constructor is inserted, which causes the method count not to be exactly the same as the method metadata count. Thus, decorator metadata is inserted in the incorrect place.

This PR refactors the core transformation logic to ensure that the new methods and properties inserted by other plugins won't affect the decorator transformation.

### Transformation Flow Changes

  Before: Complex upfront analysis with separate processing paths
  - Used check_class_has_decorated() to analyze entire class tree
  - Separate handling for class vs element decorators
  - Multiple disconnected stacks for state tracking

  After: Unified incremental processing
  - Process decorators as we traverse (exit_method_definition,
  exit_property_definition, etc.)
  - Accumulate decoration statements in single stack during traversal
  - Unified processing in transform_class() using consolidated state

### DIFF
--- a/crates/oxc_transformer/src/decorator/legacy/metadata.rs
+++ b/crates/oxc_transformer/src/decorator/legacy/metadata.rs
@@ -87,11 +87,9 @@
 ///
 /// ## References
 /// * TypeScript's [emitDecoratorMetadata](https://www.typescriptlang.org/tsconfig#emitDecoratorMetadata)
-use std::collections::VecDeque;
-
 use oxc_allocator::{Box as ArenaBox, TakeIn};
 use oxc_ast::ast::*;
-use oxc_data_structures::stack::NonEmptyStack;
+use oxc_data_structures::stack::SparseStack;
 use oxc_semantic::{Reference, ReferenceFlags, SymbolId};
 use oxc_span::{ContentEq, SPAN};
 use oxc_traverse::{MaybeBoundIdentifier, Traverse};
@@ -115,14 +113,22 @@ enum EnumType {
     Object,
 }
 
-pub enum MethodMetadata<'a> {
-    Constructor(Expression<'a>),
-    Normal([Expression<'a>; 3]),
-}
-
 pub struct LegacyDecoratorMetadata<'a, 'ctx> {
     ctx: &'ctx TransformCtx<'a>,
-    metadata_stack: NonEmptyStack<VecDeque<MethodMetadata<'a>>>,
+    /// Stack of method metadata arrays, each array contains 3 expressions:
+    /// `[design:type, design:paramtypes, design:returntype]`
+    ///
+    /// Only the method that needs to be pushed onto a stack is the method metadata,
+    /// which should be inserted after all real decorators. However, method parameters
+    /// will be processed before the metadata generation, so we need to temporarily store
+    /// them in a stack and pop them when in exit_method_definition.
+    method_metadata_stack: SparseStack<[Expression<'a>; 3]>,
+    /// Stack of constructor metadata expressions, each expression
+    /// is the `design:paramtypes`.
+    ///
+    /// Same as `method_metadata_stack`, but for constructors. Because the constructor is specially treated
+    /// in the class, we need to handle it in `exit_class` rather than `exit_method_definition`.
+    constructor_metadata_stack: SparseStack<Expression<'a>>,
     enum_types: FxHashMap<SymbolId, EnumType>,
 }
 
@@ -130,13 +136,26 @@ impl<'a, 'ctx> LegacyDecoratorMetadata<'a, 'ctx> {
     pub fn new(ctx: &'ctx TransformCtx<'a>) -> Self {
         LegacyDecoratorMetadata {
             ctx,
-            metadata_stack: NonEmptyStack::new(VecDeque::new()),
+            method_metadata_stack: SparseStack::new(),
+            constructor_metadata_stack: SparseStack::new(),
             enum_types: FxHashMap::default(),
         }
     }
 }
 
 impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
+    #[inline]
+    fn exit_program(&mut self, _program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
+        debug_assert!(
+            self.method_metadata_stack.is_exhausted(),
+            "All method metadata should have been popped."
+        );
+        debug_assert!(
+            self.constructor_metadata_stack.is_exhausted(),
+            "All constructor metadata should have been popped."
+        );
+    }
+
     // `#[inline]` because this is a hot path and most `Statement`s are not `TSEnumDeclaration`s.
     // We want to avoid overhead of a function call for the common case.
     #[inline]
@@ -150,28 +169,27 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
     }
 
     fn enter_class(&mut self, class: &mut Class<'a>, ctx: &mut TraverseCtx<'a>) {
-        if class.is_expression() || class.declare {
-            return;
-        }
+        let should_transform = !(class.is_expression() || class.declare);
 
         let constructor = class.body.body.iter_mut().find_map(|item| match item {
             ClassElement::MethodDefinition(method) if method.kind.is_constructor() => Some(method),
             _ => None,
         });
 
-        if let Some(constructor) = constructor
+        let metadata = if should_transform
+            && let Some(constructor) = constructor
             && !(class.decorators.is_empty()
                 && constructor.value.params.items.iter().all(|param| param.decorators.is_empty()))
         {
             let serialized_type =
                 self.serialize_parameter_types_of_node(&constructor.value.params, ctx);
 
-            self.metadata_stack.push(VecDeque::from_iter([MethodMetadata::Constructor(
-                self.create_metadata("design:paramtypes", serialized_type, ctx),
-            )]));
+            Some(self.create_metadata("design:paramtypes", serialized_type, ctx))
         } else {
-            self.metadata_stack.push(VecDeque::new());
-        }
+            None
+        };
+
+        self.constructor_metadata_stack.push(metadata);
     }
 
     fn enter_method_definition(
@@ -179,29 +197,32 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
         method: &mut MethodDefinition<'a>,
         ctx: &mut TraverseCtx<'a>,
     ) {
-        if method.kind.is_constructor() || method.value.is_typescript_syntax() {
+        if method.kind.is_constructor() {
+            // Handle constructor in `enter_class`
             return;
         }
 
-        let is_decorated = !method.decorators.is_empty()
-            || method.value.params.items.iter().any(|param| !param.decorators.is_empty());
-        if !is_decorated {
-            return;
-        }
+        let is_typescript_syntax = method.value.is_typescript_syntax();
+        let is_decorated = !is_typescript_syntax
+            && (!method.decorators.is_empty()
+                || method.value.params.items.iter().any(|param| !param.decorators.is_empty()));
 
-        let metadata = [
-            self.create_metadata("design:type", Self::global_function(ctx), ctx),
-            {
-                let serialized_type =
-                    self.serialize_parameter_types_of_node(&method.value.params, ctx);
-                self.create_metadata("design:paramtypes", serialized_type, ctx)
-            },
-            {
-                let serialized_type = self.serialize_return_type_of_node(&method.value, ctx);
-                self.create_metadata("design:returntype", serialized_type, ctx)
-            },
-        ];
-        self.metadata_stack.last_mut().push_back(MethodMetadata::Normal(metadata));
+        let metadata = is_decorated.then(|| {
+            [
+                self.create_metadata("design:type", Self::global_function(ctx), ctx),
+                {
+                    let serialized_type =
+                        self.serialize_parameter_types_of_node(&method.value.params, ctx);
+                    self.create_metadata("design:paramtypes", serialized_type, ctx)
+                },
+                {
+                    let serialized_type = self.serialize_return_type_of_node(&method.value, ctx);
+                    self.create_metadata("design:returntype", serialized_type, ctx)
+                },
+            ]
+        });
+
+        self.method_metadata_stack.push(metadata);
     }
 
     #[inline]
@@ -281,8 +302,12 @@ impl<'a> LegacyDecoratorMetadata<'a, '_> {
         enum_type
     }
 
-    pub fn pop_method_metadata(&mut self) -> Option<MethodMetadata<'a>> {
-        self.metadata_stack.last_mut().pop_front()
+    pub fn pop_method_metadata(&mut self) -> Option<[Expression<'a>; 3]> {
+        self.method_metadata_stack.pop()
+    }
+
+    pub fn pop_constructor_metadata(&mut self) -> Option<Expression<'a>> {
+        self.constructor_metadata_stack.pop()
     }
 
     fn serialize_type_annotation(
--- a/crates/oxc_transformer/src/decorator/legacy/mod.rs
+++ b/crates/oxc_transformer/src/decorator/legacy/mod.rs
@@ -50,6 +50,7 @@ use std::mem;
 use oxc_allocator::{Address, GetAddress, TakeIn, Vec as ArenaVec};
 use oxc_ast::{NONE, ast::*};
 use oxc_ast_visit::{Visit, VisitMut};
+use oxc_data_structures::stack::NonEmptyStack;
 use oxc_semantic::{ScopeFlags, SymbolFlags};
 use oxc_span::SPAN;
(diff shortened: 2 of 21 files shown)


Dunqing:
* **#13632** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/13632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=13632).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/decorator/legacy/mod.rs
```diff
@@ -50,6 +50,7 @@ use std::mem;
 use oxc_allocator::{Address, GetAddress, TakeIn, Vec as ArenaVec};
 use oxc_ast::{NONE, ast::*};
 use oxc_ast_visit::{Visit, VisitMut};
+use oxc_data_structures::stack::NonEmptyStack;
```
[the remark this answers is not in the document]
Dunqing: This is totally nonsense.

crates/oxc_transformer/src/decorator/legacy/mod.rs
```diff
@@ -104,8 +115,9 @@ impl<'a, 'ctx> LegacyDecorator<'a, 'ctx> {
             emit_decorator_metadata,
             metadata: LegacyDecoratorMetadata::new(ctx),
             class_decorated_data: None,
-            ctx,
             decorations: FxHashMap::default(),
+            class_decoration_stack: NonEmptyStack::new(ClassDecorations::default()),
```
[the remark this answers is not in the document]
Dunqing: The order is correct, please double-check your review.

crates/oxc_transformer/src/decorator/legacy/mod.rs
```diff
@@ -326,22 +450,24 @@ impl<'a> LegacyDecorator<'a, '_> {
...
-            class_or_constructor_parameter_is_decorated,
-            class_element_is_decorated,
-            has_private_in_expression_in_decorator,
-        } = Self::check_class_has_decorated(class);
+        // Legacy decorator does not allow in class expression.
+        if class.is_expression() || class.declare {
+            return;
+        }
 
+        let current_class_decorations = self.class_decoration_stack.pop();
```
overlookmotel: `class_decoration_stack` is always pushed to in `enter_class` (regardless of whether it's a class expression, or whether `declare == true`).

So I think the `pop` also needs to happen unconditionally - before the `if class.is_expression() || class.declare` check.

Otherwise, the stack will get out of sync with something like:

```js
class C {
  @dec method1() {}
  method2() {
    return class {}; // class expression
  }
  @dec method3() {}
}
```

When using stacks, I think it's generally a good idea to `debug_assert!` that the stack has been emptied again in `exit_program`. This commit adds such a check, and it panics when running conformance, which demonstrates something is wrong: https://github.com/oxc-project/oxc/commit/ae084fa932c43e13e5cae2fd36a9e4d280762ddb

I suggest adding similar checks in `exit_program` for `method_metadata_stack` and `constructor_metadata_stack` too, to make sure they're operating as intended.

crates/oxc_transformer/src/decorator/legacy/mod.rs
```diff
@@ -81,6 +72,22 @@ struct ClassDecoratedData<'a> {
     alias_binding: Option<BoundIdentifier<'a>>,
 }
 
+/// Decoration state for a single class level.
+/// This struct groups together all the state needed while processing decorators for one class,
+/// allowing clean management of nested class scenarios through a stack of these structs.
```
overlookmotel: AI wrote this comment, right? You can tell by how it's excessively long, and the last 2 lines don't really say anything! 😄

```suggestion
/// Decoration state for a single class.
```

overlookmotel (commented):
This PR changes quite a lot, and I'm not that familiar with decorators transform to start with, so it's hard to review properly without spending hours on it.

There's one bug related to stacks (see below).

Also, a couple of comments/questions:

1. Are you sure that decorators always get transformed by other transforms before they're removed from AST? (e.g. in a case like `@(x ?? y)`)

2. How `method_metadata_stack` and `constructor_metadata_stack` are pushed/popped looks a little fragile. Is it guaranteed that every `push` has a corresponding `pop`, even if some other transform e.g. adds/removes a constructor or method in between, or if some classes are class expressions?

This decorators transform is pretty damn tricky!

(I hope you don't mind I pushed 1 unimportant commit, just to reformat the test cases. Conformance CI fail is not related to that commit - it passes locally. Something must have changed on main since PR was originally submitted.)

crates/oxc_transformer/src/decorator/legacy/mod.rs
Dunqing: Ah, yes. Most of the comments are written by AI, and I tweaked them a little bit.

crates/oxc_transformer/src/decorator/legacy/mod.rs
Dunqing: Thanks, I see the problem. This is an easily overlooked point. I've pushed https://github.com/oxc-project/oxc/pull/13632/commits/c5937d20247dfd26db76116e51f1a4e8471e7386 and https://github.com/oxc-project/oxc/pull/13632/commits/54d9766c981a80cec691c1b024bd2226a8d064a4 to ensure that every `push` has a corresponding `pop`. 

The `constructor_metadata_stack` is a little tricky because its execution order is important. I was planning to refactor the `constructor` related implementation to be the same as the `method`. However, I am not sure whether it is possible, and I consider that the current changes are already significant, so I've given up on refactoring in this PR.

Dunqing:
> 1. Are you sure that decorators always get transformed by other transforms before they're removed from AST? (e.g. in a case like `@(x ?? y)`)

Ideally, yes, the decorators are transformed in the `exit_**` phase like `exit_method_definition` and `exit_property_definition`, so before entering these methods, the expressions are already transformed by other plugins in `enter_expression`  or `exit_expression`.

However, there is a weird situation that will break this. For example, if expressions are handled by the sub-visitor and it is called in the `exit_class` or something else that is even later, then this will probably be overlooked.

Dunqing:
@overlookmotel I've resolved your requested changes. Can you take another look? I am happy if you want to make a change directly in this PR or in the follow-up PRs.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -115,28 +113,49 @@ enum EnumType {
...
+    #[inline]
+    fn exit_program(&mut self, _program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
+        debug_assert!(
+            self.method_metadata_stack.last().is_none(),
+            "All method metadata should have been popped."
+        );
+        debug_assert!(
+            self.constructor_metadata_stack.last().is_none(),
+            "All constructor metadata should have been popped."
+        );
```
overlookmotel: These checks don't make sure the stack is emptied, only that the item on top of stack is `None`.

`SparseStack` is never empty (like `NonEmptyStack`), so the way to test that every `push` was followed by a `pop` is this:

```suggestion
        debug_assert!(
            self.method_metadata_stack.len() == 1,
            "All method metadata should have been popped."
        );
        debug_assert!(
            self.constructor_metadata_stack.len() == 1,
            "All constructor metadata should have been popped."
        );
```

Unfortunately, after making that change, there's panics in conformance "All method metadata should have been popped". So stack push/pop must be getting out of sync somewhere still.

overlookmotel:
> decorators are transformed in the `exit_**` phase like `exit_method_definition` and `exit_property_definition`, so before entering these methods, the expressions are already transformed by other plugins in `enter_expression` or `exit_expression`.

OK great!

> However, there is a weird situation that will break this. For example, if expressions are handled by the sub-visitor and it is called in the `exit_class` or something else that is even later, then this will probably be overlooked.

Hmm. Yes, you're right. But I'm not sure what we could do to solve this. There is a period in time between when the decorators are extracted from the AST, and when they're inserted again. Any code examining the AST during that window of time won't see them. But in practice, I don't *think* this will present a problem.

overlookmotel:
This PR made me realize that it's confusing how to check that stacks are back to initial state at end of AST traversal, because it's different depending on the stack type:

* `Stack` -> `stack.is_empty()`
* `NonEmptyStack` -> `stack.len() == 1`
* `SparseStack` -> `stack.len() == 1`

This is not obvious, and it's really error-prone (bad design on my part).

#13672 adds an `is_exhausted` method to `NonEmptyStack` and `SparseStack`, which I think makes the intent clearer.

Once that PR is merged, I suggest using `is_exhausted` for the `exit_program` stack checks in this PR.

Dunqing:
> This PR made me realize that it's confusing how to check that stacks are back to initial state at end of AST traversal, because it's different depending on the stack type:
> 
> * `Stack` -> `stack.is_empty()`
> * `NonEmptyStack` -> `stack.len() == 1`
> * `SparseStack` -> `stack.len() == 1`
> 
> This is not obvious, and it's really error-prone (bad design on my part).
> 
> #13672 adds an `is_exhausted` method to `NonEmptyStack` and `SparseStack`, which I think makes the intent clearer.
> 
> Once that PR is merged, I suggest using `is_exhausted` for the `exit_program` stack checks in this PR.

Thank you for adding this method. I had incorrectly used `is_empty` a few times 🥲.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: Fixed in https://github.com/oxc-project/oxc/pull/13632/commits/01750d6ff5f1896a265f20e3e505b296e8fea1fa

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -150,58 +169,59 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
     ) {
-        if method.kind.is_constructor() || method.value.is_typescript_syntax() {
+        if method.kind.is_constructor() {
+            // Handle constructor in `enter_class`
             return;
         }
 
-        let is_decorated = !method.decorators.is_empty()
+        let is_typescript_syntax = method.value.is_typescript_syntax();
+        let is_decorated = !is_typescript_syntax && !method.decorators.is_empty()
```
overlookmotel: Nit:

```rs
let is_decorated = !is_typescript_syntax && !method.decorators.is_empty()
    || method.value.params.items.iter().any(|param| !param.decorators.is_empty());
```

Could you please add some brackets to disambiguate whether this is:

```rs
let is_decorated = (!is_typescript_syntax && !method.decorators.is_empty())
    || method.value.params.items.iter().any(|param| !param.decorators.is_empty());
```

or:

```rs
let is_decorated = !is_typescript_syntax && (
    !method.decorators.is_empty()
    || method.value.params.items.iter().any(|param| !param.decorators.is_empty())
);
```

I actually can't remember the associativity rules, so I'm not sure which it is.

overlookmotel (approved):
I added a commit just to rename a var to be more accurate description of what it is.

One nit comment below. If you're willing to make that change, please go ahead and merge afterwards. Or if you can't be bothered, just merge it now!

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -115,28 +113,49 @@ enum EnumType {
...
+    fn exit_program(&mut self, _program: &mut Program<'a>, _ctx: &mut TraverseCtx<'a>) {
+        debug_assert!(
+            self.method_metadata_stack.is_exhausted(),
+            "All method metadata should have been popped."
+        );
+        debug_assert!(
+            self.constructor_metadata_stack.is_exhausted(),
+            "All constructor metadata should have been popped."
+        );
+    }
```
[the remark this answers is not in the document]
overlookmotel: Our conformance testing is thorough, so I think `debug_assert!` is sufficient.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: Should be the latter, but, yeah, I can't distinguish which will execute first from looking at the code

Dunqing:
Thank you for reviewing. Merging now!

Dunqing:
### Merge activity

* **Sep 11, 8:01 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Sep 11, 8:06 AM UTC**: `Dunqing` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Sep 11, 8:12 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #13806, opened 2025-09-16
fix(formatter): Add parentheses for `PrivateInExpression` in super class

nnnnoel:


Dunqing (changes requested):
I think we can use `is_class_extends` here

https://github.com/oxc-project/oxc/blob/d03999ba1b0bf499f4430831b23fee014c329c6c/crates/oxc_formatter/src/parentheses/expression.rs#L943-L948

Dunqing (approved):
Thank you for your contribution!

PR #13905, opened 2025-09-18
fix(parser): forbid readonly in parser instead of semantic

### LINKED ISSUE #13883: parser: forbid `readonly` in parser instead of semantic
> Looks like these two cases are semantic errors instead of syntax errors.
> 
> ```ts
> function foo(readonly parameter) {}
> ```
> 
> and
> 
> ```ts
> class Foo { method(readonly parameter) {} }
> ```
> 
> Can we fix? 

 _Originally posted by @fisker in [#11713](https://github.com/oxc-project/oxc/issues/11713#issuecomment-3307224195)_

ulrichstark:
Closes #13883

This also fixes duplicate essentially same diagnostics coming from parser and semantic each.


Boshen:
Very nice, thank you!

PR #13965, opened 2025-09-21
perf(parser): optimize `parse_simple_arrow_function_expression`

ulrichstark:
The function now accepts `IdentifierReference` instead of `Expression` to make its implementation simpler and avoid one match. Also all code around `self.ctx` is removed because it wasn't needed. This results in less code and a speedup in parser benchmarks of around 1% on my device.


camc314 (approved):
nice, codspeed shows 1% as well!

Boshen:
Nice!

PR #14016, opened 2025-09-23
fix(formatter): Add parentheses inside `TSOptionalType`

nnnnoel:


nnnnoel:
Correctly implemented by https://github.com/oxc-project/oxc/pull/14022

PR #14028, opened 2025-09-23
feat(data_structures): add `Stack::clear`

sapphi-red:
Added `Stack::clear` for #14029.

### DIFF
--- a/crates/oxc_data_structures/src/stack/standard.rs
+++ b/crates/oxc_data_structures/src/stack/standard.rs
@@ -334,6 +334,25 @@ impl<T> Stack<T> {
         unsafe { self.cursor.read() }
     }
 
+    /// Clear the stack, removing all values.
+    ///
+    /// Note that this method has no effect on the allocated capacity
+    /// of the stack.
+    #[inline]
+    pub fn clear(&mut self) {
+        if self.is_empty() {
+            return;
+        }
+
+        debug_assert!(self.end > self.start);
+
+        // SAFETY: Checked above that stack is not empty, so stack is allocated.
+        // Stack contains `self.len()` initialized entries, starting at `self.start`
+        unsafe { self.drop_contents() };
+        // Move cursor back to start. This is equivalent to setting len to 0
+        self.cursor = self.start;
+    }
+
     /// Get number of entries on stack.
     #[inline]
     pub fn len(&self) -> usize {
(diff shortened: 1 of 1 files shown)


sapphi-red:
* **#14028** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=14028).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel (approved):
`self.is_empty()` check may not actually be required. When the stack's content type (`T`) is non-`Drop`, `self.drop_contents()` becomes a no-op anyway. So if we could remove `self.is_empty()` check,this function becomes just a single instruction and no branches, when `T` is non-`Drop`.

`drop_contents` has a safety comment saying "Stack must be allocated", but I'm not sure it's actually required. I *think* it's fine to call `ptr::drop_in_place` with an empty slice. But let's leave it as is - maybe there was a reason I added this safety constraint which I don't remember right now.

sapphi-red:
When the stack is not allocated, `self.start` is a dangling pointer. In that case, the return value of `ptr::slice_from_raw_parts_mut(self.start().as_ptr(), self.len())` is not valid. I think that's the reason of the "Stack must be allocated" comment.
This `self.is_empty()` check indirectly checks that the stack is allocated from the fact that "non-empty stack would always have the stack allocated".

overlookmotel:
> When the stack is not allocated, `self.start` is a dangling pointer. In that case, the return value of `ptr::slice_from_raw_parts_mut(self.start().as_ptr(), self.len())` is not valid. I think that's the reason of the "Stack must be allocated" comment. This `self.is_empty()` check indirectly checks that the stack is allocated from the fact that "non-empty stack would always have the stack allocated".

Yeah, I think that's what I had in mind at the time I wrote the code. But now I think probably it *is* legal to create a slice starting at a dangling pointer, as long as it's 0 length (dangling pointers are correctly aligned).

Anyway, it doesn't really matter much. This PR just made me think about it again.

sapphi-red:
> Yeah, I think that's what I had in mind at the time I wrote the code. But now I think probably it _is_ legal to create a slice starting at a dangling pointer, as long as it's 0 length (dangling pointers are correctly aligned).

Ah, I see. I think it's legal. I've just found this:

> `data` must be non-null and aligned even for zero-length slices or slices of ZSTs. One reason for this is that enum layout optimizations may rely on references (including slices of any length) being aligned and non-null to distinguish them from other data. You can obtain a pointer that is usable as `data` for zero-length slices using `NonNull::dangling()`.
> https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html#:~:text=data%20must%20be,NonNull%3A%3Adangling().

...which is what we're doing here.

overlookmotel:
> You can obtain a pointer that is usable as `data` for zero-length slices using `NonNull::dangling()`

Nice. I hadn't spotted that in docs before.

PR #14038, opened 2025-09-23
fix(napi/parser): generate `.d.mts` extension for types

danielroe:
the latest version of `oxc-parser` (on npm) does not have resolvable types because the entry is `.mjs` and the types are `.d.ts` (these need to match - `.d.mts`). you can see this in the failing CI for [this renovate PR](https://github.com/nuxt/nuxt/actions/runs/17940186619/job/51014702361?pr=33290) but I'm happy to provide a more minimal reproduction if it helps

I also spotted that `napi/parser/test/visit.test.ts` was importing `visitorKeys` which was not exported from the corresponding declaration, so took the liberty to add that export as well.

let me know if this needs any changes 🙏 

I hope it helps


overlookmotel:
Thanks very much for spotting this and fixing. It needs a few amends to work around the oddities of NAPI-RS build and our infrastructure. I'm going to push a couple of commits - hope you don't mind.

Marking as draft for now.

danielroe:
I don't mind at all! ❤️ 

thank you for spotting this so quickly. 🙏

overlookmotel:
Well I broke it, so I should fix it!

PR #14117, opened 2025-09-25
feat(linter): add `unicorn/no-array-sort` rule

cscnk52:
Relates to #684
Relates to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v61.0.2/docs/rules/no-array-sort.md


camc314 (approved):
thank you

PR #14160, opened 2025-09-26
perf(parser): cleanup and optimize various parsing functions

ulrichstark:
- Avoid unnecessary Option wrapping and checks by splitting parse_normal_list function into a breakable and non-breakable variant
- Avoid unnecessary Option wrapping and checks in parse_variable_declaration_for_statement function
- Break up parse_break_or_continue_statement function into function for break and continue
- Remove unnecessary checks in is_using_declaration and use cheaper lexer.peek_token()
- Make parse_variable_declaration take `declare: bool` instead of `&Modifiers` avoid constructing empty Modifiers
- some other minor code improvements

I hope these changes are not too much and therefore difficult to review. It's now less code with better performance. I was in the flow of refactoring and couldn't stop :)

Edit:

I don't see the improvements on Codspeed I'm seeing locally when running `cargo benchmark --bench parser` against main... Feel free to close my PR if there are no verifiable performance improvements. My local output:

image

### DIFF
--- a/crates/oxc_parser/src/cursor.rs
+++ b/crates/oxc_parser/src/cursor.rs
@@ -158,8 +158,7 @@ impl<'a> ParserImpl<'a> {
     #[inline]
     pub(crate) fn can_insert_semicolon(&self) -> bool {
         let token = self.cur_token();
-        let kind = token.kind();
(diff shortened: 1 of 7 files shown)


Boshen:
Thank you! Less code, more performant!

PR #14203, opened 2025-09-28
fix(editor): execute `oxc.path.server` in win32 with shell

### LINKED ISSUE #14167: VSCode: allow `oxc.path.server` to be used in Linux & Windows
> I see https://github.com/oxc-project/oxc/pull/13542 was merged an hour ago. This only covers the case with the relative path, correct? I could not share `oxc.path.server` as a project wide setting in my `.vscode/settings.json`, if I have a mix of Windows and non-Windows developers as I sometimes would have to use `.exe` and sometimes not. (Referring to [this comment](https://github.com/oxc-project/oxc/issues/12849#issuecomment-3160832991).) 

 _Originally posted by @donaldpipowitch in [#12849](https://github.com/oxc-project/oxc/issues/12849#issuecomment-3278213562)_

Created an own issue for it, because it is more work than expected.
There is no `.exe`, only `.CMD` and `.ps1` (in NPM/PNPM). 
Bun has  `.exe` file in `node_modules/.bin`. 

There could be 2 solutions:

Enable `shell` option, so VSCode can use `node_modules/.bin/language_server` too.
This will enable a security hole (https://github.com/oxc-project/oxc/pull/13732#discussion_r2345327975), which should be covered by https://github.com/oxc-project/oxc/pull/13740 and https://github.com/oxc-project/oxc/pull/13734.
I would like to force somehow more checks before enabling this.

Search for `.exe` inside `node_modules`.
This could be a bit difficult because the package manager could but the file somewhere else than expected.
The other editors could have solved this already. But we should check if this works with different package managers as expected.

Sysix:
closes #14167
This allows cross-platform configuration in a project.
The project can now define `oxc.path.server` to `node_modules/.bin/oxc_language_server` and windows will pick it up too.

 With `validateSafeBinaryPath` in `Pathvalidator.ts`, we should be sure that bad actors can not harm the user's machine.

This is needed because the VSCode extension will not ship the language server in the future.

### DIFF
--- a/editors/vscode/client/extension.ts
+++ b/editors/vscode/client/extension.ts
@@ -130,7 +130,7 @@ export async function activate(context: ExtensionContext) {
   );
 
   async function findBinary(): Promise<string> {
-    let bin = configService.getUserServerBinPath();
+    const bin = configService.getUserServerBinPath();
     if (workspace.isTrusted && bin) {
       try {
         await fsPromises.access(bin);
@@ -168,6 +168,12 @@ export async function activate(context: ExtensionContext) {
           command: path!,
           args: ['--lsp'],
           options: {
+            // On Windows we need to run the binary in a shell to be able to execute the shell npm bin script.
+            // Searching for the right `.exe` file inside `node_modules/` is not reliable as it depends on
+            // the package manager used (npm, yarn, pnpm, etc) and the package version.
+            // The npm bin script is a shell script that points to the actual binary.
+            // Security: We validated the userDefinedBinary in `configService.getUserServerBinPath()`.
+            shell: process.platform === 'win32',
             env: serverEnv,
           },
         };
--- a/editors/vscode/client/PathValidator.ts
+++ b/editors/vscode/client/PathValidator.ts
@@ -8,7 +8,7 @@
  *
  * The check for malicious characters is not needed, but it's an additional layer of security.
  * When using `shell: true` in `LanguageClient.ServerOptions`, it can be vulnerable to command injection.
- * Even though we are not using `shell: true`, it's a good practice to validate the input.
+ * We are using `shell: true` only on Windows when the paths ends with `node_modules/.bin/oxc_language_server`.
  */
 export function validateSafeBinaryPath(binary: string): boolean {
   // Check for path traversal (including Windows variants)
--- a/editors/vscode/tests/ConfigService.spec.ts
+++ b/editors/vscode/tests/ConfigService.spec.ts
@@ -18,20 +18,55 @@ suite('ConfigService', () => {
     await Promise.all(keys.map(key => conf.update(key, undefined)));
   });
 
-  testSingleFolderMode('resolves relative server path with workspace folder', async () => {
-    const service = new ConfigService();
-    const nonDefinedServerPath = service.getUserServerBinPath();
+  const getWorkspaceFolderPlatformSafe = () => {
+    let workspace_path = WORKSPACE_FOLDER.uri.path;
+    if (process.platform === 'win32') {
+      workspace_path = workspace_path.replaceAll('/', '\\');
+      if (workspace_path.startsWith('\\')) {
+        workspace_path = workspace_path.slice(1);
+      }
+    }
+    return workspace_path;
+  };
 
-    strictEqual(nonDefinedServerPath, undefined);
+  suite('getUserServerBinPath', () => {
+    testSingleFolderMode('resolves relative server path with workspace folder', async () => {
+      const service = new ConfigService();
+      const nonDefinedServerPath = service.getUserServerBinPath();
 
-    await conf.update('path.server', '/absolute/oxc_language_server');
-    const absoluteServerPath = service.getUserServerBinPath();
+      strictEqual(nonDefinedServerPath, undefined);
 
-    strictEqual(absoluteServerPath, '/absolute/oxc_language_server');
+      await conf.update('path.server', '/absolute/oxc_language_server');
+      const absoluteServerPath = service.getUserServerBinPath();
 
-    await conf.update('path.server', './relative/oxc_language_server');
-    const relativeServerPath = service.getUserServerBinPath();
+      strictEqual(absoluteServerPath, '/absolute/oxc_language_server');
 
-    strictEqual(relativeServerPath, WORKSPACE_FOLDER.uri.path + '/relative/oxc_language_server');
+      await conf.update('path.server', './relative/oxc_language_server');
+      const relativeServerPath = service.getUserServerBinPath();
+
+      let workspace_path = getWorkspaceFolderPlatformSafe();
+      strictEqual(relativeServerPath, `${workspace_path}/relative/oxc_language_server`);
+    });
+
+    testSingleFolderMode('returns undefined for unsafe server path', async () => {
+      const service = new ConfigService();
+      await conf.update('path.server', '../unsafe/oxc_language_server');
+      const unsafeServerPath = service.getUserServerBinPath();
+
+      strictEqual(unsafeServerPath, undefined);
+    });
+
+   testSingleFolderMode('returns backslashes path on Windows', async () => {
+      if (process.platform !== 'win32') {
+        return;
+      }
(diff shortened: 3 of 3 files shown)


Sysix:
* **#14203** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/14203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/14203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15680** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15680?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15611** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15610** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15610?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15609** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15609?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15563** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15563?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#15608](https://github.com/oxc-project/oxc/pull/15608) <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15608?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#15505** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15568** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15568?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=14203).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

editors/vscode/client/extension.ts
```diff
@@ -155,10 +154,20 @@ export async function activate(context: ExtensionContext) {
     );
   }
 
-  const command = await findBinary();
+  const userDefinedBinary = configService.getUserServerBinPath();
+  const command = await findBinary(userDefinedBinary);
   const run: Executable = {
     command: command!,
     options: {
+      // On Windows we need to run the binary in a shell to be able to execute the shell npm bin script.
```
nrayburn-tech: Is this because windows can’t execute this script directly? I think there’s another alternative to spawn a cmd prompt with the argument to execute the script, then you don’t need shell: true.

Not sure if it matters or if there’s any reason to prefer one or the other.

editors/vscode/client/extension.ts
```diff
@@ -155,10 +154,20 @@ export async function activate(context: ExtensionContext) {
...
+      // This is only needed when the user explicitly configures the binary to point to the npm bin script.
+      // The extension is shipped with the `.exe` file, we don't need to run it in a shell.
+      // Searching for the right `.exe` file inside `node_modules/` is not reliable as it depends on
+      // the package manager used (npm, yarn, pnpm, etc) and the package version.
+      // The npm bin script is a shell script that points to the actual binary.
+      // Security: We validated the userDefinedBinary in `configService.getUserServerBinPath()`.
+      shell: process.platform === 'win32' && command === userDefinedBinary &&
+        userDefinedBinary?.endsWith('node_modules\\.bin\\oxc_language_server'),
```
nrayburn-tech: I don’t personally think this (or the other path validations) are necessary.

If the user has opted into the trusted workspace, they have consented to these settings. I don’t think the extension is responsible for validating what the path might look like or what the contents of the evaluated file might be.

It has potential to make it harder for users to use, for what I would consider little to no real security benefit.

editors/vscode/client/extension.ts
Sysix: Yes, it does not handle the sh shebang in the file. Calling the CMD file or ps1 file inside `node_modules/.bin/` will not work. This is the only solution or searching for the `.exe` file.
But searching for the `.exe` file will create more problems with the location of the .exe. 
Your examples in the issue did not check for different package versions. Calling the wrapper script in `node_modules/.bin/` "should" be the right way.

editors/vscode/client/extension.ts
Sysix: A user would trust the extension more than a freshly opened workspace. Even when a user has trusted a workspace by accident (when the workspace is inside a parent folder, which is already been trusted). Opening the extension to execute untrusted binaries / shell script should still be validated.
Maybe not all restrictions are needed and can frustrate users to set up the `oxc.path.server` correctly.
But I will take the cost for better end-user security.

nrayburn-tech: I think what you’re trying to solve is possibly a gap in the trusted workspace feature. Instead of trusting an entire workspace, you want users to trust specific values within the workspace. It doesn’t feel like the responsibility of the extension.

I think the definition of setting the oxc.path.server “correctly” is subjective. If I want to use whatever characters in my path, that shouldn’t be prevented by the extension.

I understand the desire for security though. I don’t think this is a blocker by any means, just wanted to bring it up as something to consider.

nrayburn-tech:
I won’t be able to actually review any code for a few days, but I don’t think I’ll have much to share other than the few considerations I mentioned before.

Just things for the team to consider.

For what it’s worth, the IntelliJ plugin doesn’t do any path validation and it’s not something I would personally add. The IntelliJ plugin also uses some abstraction provided by IntelliJ, but I believe by default it executes the same file that is being suggested for this PR. However, if a user specifies an exe path directly, that still works too. Specifying an exe path directly just won’t work cross platform (in typical scenarios).

---

I think I mentioned this a while ago when the IDE development was only beginning. It might be beneficial to have documentation on how we want the different IDE plugins owned by the Oxc team to function. This would provide users with a consistent experience across the different IDEs. Currently, users likely have a slightly different experience between the various IDEs (VS Code, IntelliJ, Zed, etc.)

Sysix:
moving it back to a draft.
Reason: maybe with the node integration for js-plugins, this is not needed anymore.
Idea from: https://github.com/oxc-project/oxc-intellij-plugin/issues/239#issuecomment-3393576927

EDIT: there is still a need after testing it with windows

camc314 (approved):

camc314:
### Merge activity

* **Nov 14, 1:18 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 14, 1:18 PM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 14, 1:22 PM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) removed this pull request due to **downstack failures on PR #[15680](https://app.graphite.com/github/pr/oxc-project/oxc/15680)**.
* **Nov 14, 1:22 PM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) removed this pull request due to **downstack failures on PR #[15680](https://app.graphite.com/github/pr/oxc-project/oxc/15680)**.
* **Nov 14, 1:32 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 14, 1:36 PM UTC**: `Sysix` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 14, 1:41 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

PR #14325, opened 2025-10-03
fix(parser): forbid abstract class members with implementation in parser instead of semantic

### LINKED ISSUE #14013: Forbid abstract property to have an initializer instead of semantic
```typescript
class Foo {
  abstract prop = 3;
}
```

ulrichstark:
The following four cases are now errors in parser instead of semantic errors.
[TS Playground](https://www.typescriptlang.org/play/?#code/IYIwzgLgTsDGEAJYBthjAgYge2wg3gFAIKiQzwIAOU2VCAvAgMwDchxp40ciAtgFMIAC2wATABQBKAggC+HEmR6UA5kITqIEAVGkFOJKEICuUAHYIAjOxILOyiojAaX23RIBuwZCYEAuBHMTPhBdGSISEk4FOSA) for reference.
Closes #14013 

```typescript
abstract class Foo {
  abstract prop = 3;

  abstract method() { }

  abstract get getter() {
    return 1;
  }

  abstract set setter(value: number) {
    
  }
}
```

### DIFF
--- a/tasks/coverage/misc/fail/oxc-14013.ts
+++ b/tasks/coverage/misc/fail/oxc-14013.ts
@@ -3,11 +3,11 @@ abstract class Foo {
 
   abstract method() { }
 
-  abstract get accessor() {
+  abstract get getter() {
     return 1;
   }
 
-  abstract set accessor(value: number) {
+  abstract set setter(value: number) {
     
   }
 }
--- a/tasks/coverage/misc/fail/oxc-14013.ts
+++ b/tasks/coverage/misc/fail/oxc-14013.ts
@@ -1,0 +1,13 @@
+abstract class Foo {
+  abstract prop = 3;
+
+  abstract method() { }
+
+  abstract get accessor() {
+    return 1;
+  }
+
+  abstract set accessor(value: number) {
+    
+  }
+}
--- a/tasks/coverage/snapshots/parser_misc.snap
+++ b/tasks/coverage/snapshots/parser_misc.snap
@@ -1,7 +1,7 @@
 parser_misc Summary:
 AST Parsed     : 49/49 (100.00%)
 Positive Passed: 49/49 (100.00%)
-Negative Passed: 91/91 (100.00%)
+Negative Passed: 92/92 (100.00%)
 
   × Cannot assign to 'arguments' in strict mode
    ╭─[misc/fail/arguments-eval.ts:1:10]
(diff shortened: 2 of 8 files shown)


tasks/coverage/misc/fail/oxc-14013.ts
```diff
@@ -0,0 +1,13 @@
+abstract class Foo {
+  abstract prop = 3;
+
+  abstract method() { }
+
+  abstract get accessor() {
```
fisker: Suggest use name `getter`/`setter` to avoid confusion. I thought it's auto-accessor at first when you mentioned in https://github.com/oxc-project/oxc/issues/14013#issuecomment-3339194335.

ulrichstark: Changed both names from "accessor" to "getter" and "setter" in https://github.com/oxc-project/oxc/pull/14325/commits/e659c9efb7abfb8665294277f6db8f1ee100f852

tasks/coverage/snapshots/parser_misc.snap
```diff
@@ -2851,6 +2851,38 @@ Negative Passed: 91/91 (100.00%)
...
+
+  × TS(1245): Method 'method' cannot have an implementation because it is marked abstract.
+   ╭─[misc/fail/oxc-14013.ts:4:12]
+ 3 │ 
+ 4 │   abstract method() { }
+   ·            ──────
+ 5 │ 
+   ╰────
+
+  × TS(1318): Accessor 'getter' cannot have an implementation because it is marked abstract.
```
fisker, line 2870: The message for accessor is different from typescript https://www.typescriptlang.org/play/?#code/IYIwzgLgTsDGEAJYBthjAgYge2wg3gFAIkKiQzwIDmAponRBLVABQCUBAvoV0A

fisker (commented):
Can't approve since I don't have access, but LGTM.

tasks/coverage/snapshots/parser_misc.snap
ulrichstark: Yes true. I deliberately decided for that message because it closer matches the diagnostic messages for methods and properties, and I also had the name of the accessor already available in that code reporting the diagnostic.

If anyone wants to have the message "An abstract accessor cannot have an implementation." like TypeScript instead, I can do the change.

Boshen (approved):

PR #14341, opened 2025-10-04
feat(linter): add eslint/no-param-reassign rule

matthewdavi:
## Summary
- port ESLint's `no-param-reassign` rule to `oxlint`, including full option handling (`props`, ignore lists)
- add comprehensive pass/fail coverage mirroring the upstream test suite, and update snapshots
- register the rule with the ESLint catalog so it is available via oxlint


camc314 (commented):
thank you!

camc314 (approved):

PR #14357, opened 2025-10-05
ci: should clippy share cache with `cargo check`?

Boshen:


Boshen:
Nope it's running with another profile `cargo lint --profile dev-no-debug-assertions`

PR #14379, opened 2025-10-06
fix(language_server): ignore JS plugins

overlookmotel:
Previously we produced an error when JS plugins are defined in config but no `ExternalLinter` is present. That's the desired behavior in `oxlint` CLI, because it happens when user is running Oxlint on an unsupported platform (32-bit, or big-endian).

But it's *not* what we want to happen in language server. Language server does not yet support JS plugins, but user may want to use them in their project (via `oxlint` CLI), and language server should not fail with a "JS plugins are not supported" error on any file in such a project.

This PR prevents that by adding an `is_enabled` field to `ExternalPluginStore`. By default, `is_enabled = true`, but language server sets it to `false`. When `false`, JS plugins are silently ignored, rather than producing an error.

Note: A better solution would be to pass around `Option<ExternalPluginStore>` instead of `ExternalPluginStore` having an `is_enabled` flag. But JS plugin support will be added to the language server pretty soon, at which point we'll need to revert this change. `is_enabled` flag approach will produce less churn when we do that.

### DIFF
--- a/crates/oxc_language_server/fixtures/linter/js_plugins/.oxlintrc.json
+++ b/crates/oxc_language_server/fixtures/linter/js_plugins/.oxlintrc.json
@@ -1,0 +1,7 @@
+{
+  "jsPlugins": ["./plugin.js"],
+  "rules": {
+    "js-plugin/no-debugger": "error",
+    "no-unused-vars": "off"
+  }
+}
--- a/crates/oxc_language_server/fixtures/linter/js_plugins/index.js
+++ b/crates/oxc_language_server/fixtures/linter/js_plugins/index.js
@@ -1,0 +1,3 @@
+debugger;
+
+let x;
--- a/crates/oxc_language_server/fixtures/linter/js_plugins/plugin.js
+++ b/crates/oxc_language_server/fixtures/linter/js_plugins/plugin.js
@@ -1,0 +1,21 @@
+const plugin = {
+  meta: {
+    name: 'js-plugin',
+  },
+  rules: {
+    'no-debugger': {
+      create(context) {
+        return {
+          DebuggerStatement(debuggerStatement) {
+            context.report({
+              message: 'Unexpected Debugger Statement',
+              node: debuggerStatement,
+            });
+          },
+        };
+      },
+    },
+  },
+};
+
+export default plugin;
--- a/crates/oxc_language_server/src/linter/server_linter.rs
+++ b/crates/oxc_language_server/src/linter/server_linter.rs
@@ -109,7 +109,7 @@ impl ServerLinter {
 
         let base_patterns = oxlintrc.ignore_patterns.clone();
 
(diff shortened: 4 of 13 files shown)


overlookmotel:
* **#14379** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14379?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14378** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14378?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=14379).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
Something weird going on... the test which this PR added passes, but it *also* passes on `main` branch. So either the test doesn't test what it's meant to, or this PR may be unnecessary.

@camc314 @Sysix Any ideas?

camc314:
> Something weird going on... the test which this PR added passes, but it _also_ passes on `main` branch. So either the test doesn't test what it's meant to, or this PR may be unnecessary.
> 
> @camc314 @Sysix Any ideas?

https://github.com/oxc-project/oxc/blob/d36d22767e735d9128bf3ef57567d3074e44bed6/crates/oxc_language_server/src/linter/server_linter.rs#L112-L118

we `unwrap_or_default` if we fail to build config store i.e. fail to load JS plugin, which is probably why it passes on `main`

overlookmotel:
> we `unwrap_or_default` if we fail to build config store i.e. fail to load JS plugin, which is probably why it passes on `main`

Ah ha! Thanks.

Have altered the test fixture to account for this. The config now produces different output from default, so now if the config is ignored, the test fails. This test now *does* fail on main, and passes on this PR.

So this PR is necessary to produce correct behavior in language server when project uses JS plugins.

camc314 (approved):

overlookmotel:
### Merge activity

* **Oct 7, 10:18 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Oct 7, 10:18 AM UTC**: `overlookmotel` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Oct 7, 10:24 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #14464, opened 2025-10-09
fix(linter/exhaustive-deps): ignore empty strings passed to `additionalHooks` option

holoflash:
### Issue
Passing an empty string to **additionalHooks** in the **react-hooks/exhaustive-deps** rule:
`"react-hooks/exhaustive-deps": ["error", { "additionalHooks": "" }],`
causes the linter to consider all functions to be additionalHooks.

### Behaviour in ESLint
Passing an empty string doesn't do anything.
Passing `"()"` causes the same problematic behaviour:
`"react-hooks/exhaustive-deps": ["warn", { additionalHooks: "()" }],`
The same is true for oxlint.

### Suggested fix
Check for empty string.
Since ESLint is able to take an empty string without breaking - so should oxlint.
Ideally - handle the "()" case too, or leave it as is to match original behaviour.

### How to test
Add:
`"react-hooks/exhaustive-deps": ["error", { "additionalHooks": "" }],`
to your .oxlintrc.json file
Lint the codebase.

Example:
```
function MyComponent(props) {
          const foo = bar.add();
    }
```

```
  ⚠ eslint-plugin-react-hooks(exhaustive-deps): React hook add requires an effect callback.
   ╭─[exhaustive_deps.tsx:2:23]
 1 │ function MyComponent(props) {
 2 │           const foo = bar.add();
   ·                       ─────────
 3 │         }
   ╰────
  help: Did you forget to pass a callback to the hook?

-------- rule config --------
[
  {
    "additionalHooks": ""
  }
]
```

### DIFF
--- a/crates/oxc_linter/src/rules/react/exhaustive_deps.rs
+++ b/crates/oxc_linter/src/rules/react/exhaustive_deps.rs
@@ -277,6 +277,7 @@ impl Rule for ExhaustiveDeps {
             .map(|config_json| ExhaustiveDepsConfig {
                 additional_hooks: config_json
                     .additional_hooks
+                    .filter(|pattern| !pattern.is_empty())
                     .and_then(|pattern| Regex::new(&pattern).ok()),
             })
             .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


camc314 (approved):

camc314 (approved):
Thank you!

PR #14495, opened 2025-10-11
fix(transformer/legacy-decorator): correct generating metadata for getter/setter methods

Dunqing:
* Fixes: #14493

Getter and setter methods should not have `design:returntype` metadata according to TypeScript's `emitDecoratorMetadata` behavior. They should only have `design:type` and `design:paramtypes`.

**And `design:type` is different for the getter and setter:**

Getter: `design:type` is the same as the return type of method.
Setter: `design:type` is the same as the type of the first parameter of the method

### DIFF
--- a/crates/oxc_transformer/src/decorator/legacy/metadata.rs
+++ b/crates/oxc_transformer/src/decorator/legacy/metadata.rs
@@ -113,16 +113,25 @@ enum EnumType {
     Object,
 }
 
+/// Metadata for decorated methods
+pub(super) struct MethodMetadata<'a> {
+    /// The `design:type` metadata expression
+    pub r#type: Expression<'a>,
+    /// The `design:paramtypes` metadata expression
+    pub param_types: Expression<'a>,
+    /// The `design:returntype` metadata expression (optional, omitted for getters/setters)
+    pub return_type: Option<Expression<'a>>,
+}
+
 pub struct LegacyDecoratorMetadata<'a, 'ctx> {
     ctx: &'ctx TransformCtx<'a>,
-    /// Stack of method metadata arrays, each array contains 3 expressions:
-    /// `[design:type, design:paramtypes, design:returntype]`
+    /// Stack of method metadata.
     ///
     /// Only the method that needs to be pushed onto a stack is the method metadata,
     /// which should be inserted after all real decorators. However, method parameters
     /// will be processed before the metadata generation, so we need to temporarily store
     /// them in a stack and pop them when in exit_method_definition.
-    method_metadata_stack: SparseStack<[Expression<'a>; 3]>,
+    method_metadata_stack: SparseStack<MethodMetadata<'a>>,
     /// Stack of constructor metadata expressions, each expression
     /// is the `design:paramtypes`.
     ///
@@ -182,7 +191,7 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
                 && constructor.value.params.items.iter().all(|param| param.decorators.is_empty()))
         {
             let serialized_type =
-                self.serialize_parameter_types_of_node(&constructor.value.params, ctx);
+                self.serialize_parameters_types_of_node(&constructor.value.params, ctx);
 
             Some(self.create_metadata("design:paramtypes", serialized_type, ctx))
         } else {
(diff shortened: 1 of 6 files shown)


Dunqing:
* **#14495** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=14495).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

tasks/transform_conformance/tests/legacy-decorators/test/fixtures/oxc/metadata/getter-setter-method/output.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+babelHelpers.decorate([
+    dec,
+    babelHelpers.decorateMetadata("design:type", Function),
+    babelHelpers.decorateMetadata("design:paramtypes", [])
+], C.prototype, "address", null);
+babelHelpers.decorate([
+    dec,
+    babelHelpers.decorateMetadata("design:type", Function),
```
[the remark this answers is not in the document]
overlookmotel: He right! [TS Playground](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=99#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX1DAAoNYBzEDALnilQE8AaeABxhxZBg3oGkR6NAM4YYWVGWaghYMSww4YNAArtO3egBEQMuQpgBKGgDccWYAG4AUFZAAPFoozxIUIUPgBheAG8r8eAABQn94Cmc2DiIjeFRkAFsAIy5fUIC4DGQYfABGACYAZmsAgF8bAODwUKFKVjUicRZkangRMQkDXzKy2wcnF2h3eE1UipCAmoj6xubhUXEyTp8esaqA8LqomLiklL8A9Mos3MLi+G6be0duAbcPAFFRoPGw2siWaJod5Jgnw8zsvB8kVQhdetdnK4hgAxJ6VMDVN7TVBNFptBZLC5AA)

Also worth noting that TS says having a decorator on both the getter and setter for same property is illegal. When you do that, it only produces 1 `__decorate` call for the class, and the types from the setter get priority, regardless of the order getter/setter are defined in.

Since it's illegal anyway, probably we can ignore that case. But we should adjust the test fixture so it doesn't contain this illegal pattern.

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -115,14 +115,18 @@ enum EnumType {
...
+    ///
+    /// Note: `design:returntype` is omitted for getter and setter methods.
     ///
     /// Only the method that needs to be pushed onto a stack is the method metadata,
     /// which should be inserted after all real decorators. However, method parameters
     /// will be processed before the metadata generation, so we need to temporarily store
     /// them in a stack and pop them when in exit_method_definition.
-    method_metadata_stack: SparseStack<[Expression<'a>; 3]>,
+    // method_metadata_stack: SparseStack<[Option<Expression<'a>>; 3]>,
+    method_metadata_stack: SparseStack<[Option<Expression<'a>>; 3]>,
```
overlookmotel: The first 2 are always `Some`, so how about:

```rs
struct MethodMetadata<'a> {
    type: Expression<'a>,
    param_types: Expression<'a>,
    return_type: Option<Expression<'a>>,
}

pub struct LegacyDecoratorMetadata<'a, 'ctx> {
    method_metadata_stack: SparseStack<MethodMetadata<'a>>,
    // ...
}
```

tasks/transform_conformance/tests/legacy-decorators/test/fixtures/oxc/metadata/getter-setter-method/output.js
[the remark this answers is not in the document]
Dunqing: Aha, it is right, I forgot to push my local changes.

> Also worth noting that TS says having a decorator on both the getter and setter for same property is illegal. When you do that, it only produces 1 `__decorate` call for the class, and the types from the setter get priority, regardless of the order getter/setter are defined in.

Yes, I've noticed that. Just think of this as hard; it's only left to the latter one, so doing nothing is okay.

> Since it's illegal anyway, probably we can ignore that case. But we should adjust the test fixture so it doesn't contain this illegal pattern.

Ok!

crates/oxc_transformer/src/decorator/legacy/metadata.rs
Dunqing: It looks good!

crates/oxc_transformer/src/decorator/legacy/metadata.rs
```diff
@@ -208,18 +217,33 @@ impl<'a> Traverse<'a, TransformState<'a>> for LegacyDecoratorMetadata<'a, '_> {
...
-                    self.create_metadata("design:paramtypes", serialized_type, ctx)
-                },
-                {
-                    let serialized_type = self.serialize_return_type_of_node(&method.value, ctx);
-                    self.create_metadata("design:returntype", serialized_type, ctx)
-                },
-            ]
+            // TypeScript only emits `design:returntype` for regular methods,
+            // not for getters or setters.
+            let should_add_return_type = !method.kind.is_get() && !method.kind.is_set();
```
overlookmotel: Nit:

```suggestion
            let should_add_return_type = !method.kind.is_accessor();
```

A little shorter. Also, weirdly, it's 1 less instruction - https://godbolt.org/z/afbxrf95d. Compiler *should* be able to do better and it *should* be the same either way, but for some reason it's not always good at optimizing enum discriminant matches.

Dunqing: I just noticed that `should_add_return_type` isn't necessary because the following logic already checks `is_get` and `is_set` separately. So I removed it.

Dunqing:
### Merge activity

* **Oct 22, 8:56 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Oct 22, 8:56 AM UTC**: `Dunqing` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Oct 22, 9:03 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #14546, opened 2025-10-13
chore(ci): fix windows test

lilnasy:
CI on main seems to be failing since snapshot based tests were introduced.

This PR includes two fixes:
- handling of backslashed paths when generating test code
- normalization of crlf which, were counted as two line breaks by snapshot comparison

Looks like windows test job doen't run on pull requests or feature branches, it only runs on main. You can review its status on my fork's main workflow run: https://github.com/lilnasy/oxc/actions/runs/18458898920.

### DIFF
--- a/crates/oxc_formatter/tests/fixtures/mod.rs
+++ b/crates/oxc_formatter/tests/fixtures/mod.rs
@@ -187,7 +187,7 @@ fn generate_snapshot(path: &Path, source_text: &str) -> String {
         }
 
         let options = parse_format_options(&option_json);
-        let formatted = format_source(source_text, source_type, options);
+        let formatted = normalize_newlines(format_source(source_text, source_type, options));
         snapshot.push_str(&formatted);
         snapshot.push('\n');
     }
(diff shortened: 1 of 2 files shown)


crates/oxc_formatter/tests/fixtures/mod.rs
```diff
@@ -215,5 +215,14 @@ fn test_file(path: &Path) {
...
 
+#[expect(clippy::disallowed_methods)]
+fn normalize_newlines(text: String) -> String {
+    if !text.contains('\r') {
+        return text;
+    }
+
+    text.replace("\r\n", "\n")
+}
+
```
Dunqing, line 226: Is this change necessary? I don't see any errors related to this, and the formatter should handle it correctly, regardless of whether the line break is `\r`, `\n`, or `\r\n`.

Dunqing (changes requested):
Thank you for fixing this

crates/oxc_formatter/tests/fixtures/mod.rs
lilnasy: It was. I can share the failing workflow run... let me get the link.

lilnasy: Run: https://github.com/lilnasy/oxc/actions/runs/18458140313/job/52583358663
Commit: https://github.com/lilnasy/oxc/commit/a26cec27ef237dc2cb622ec6b74a37acf875678d

Dunqing: Thank you, I see!

Dunqing (approved):

PR #14558, opened 2025-10-13
fix(linter/plugins): support plugin names with slashes in rule configuration

### LINKED ISSUE #14557: Linter plugins: Support plugins with `/` in name
See #14503 for original bug report.

`.oxlintrc.json`:

```json
{
	"jsPlugins": ["@eslint-react/eslint-plugin"],
	"rules": {
		"@eslint-react/eslint-plugin/jsx-no-comment-textnodes": ["error"],
		"@eslint-react/eslint-plugin/dom/no-dangerously-set-innerhtml": ["error"],
		"@eslint-react/eslint-plugin/naming-convention/context-name": ["error"]
	}
}
```

Result:

```
Failed to parse configuration file.
  x Plugin '@eslint-react' not found
```

### The problem

I assume the problem is that we split complete rule name into plugin name and rule name on the first `/`. But in this case, the plugin name contains a slash.

Splitting on the last slash wouldn't work either, as rule names also contain a `/`.

### Possible solutions

1. For each key in `rules`, try all possible combinations - i.e.
  * Plugin `@eslint-react`, rule `eslint-plugin/naming-convention/context-name`
  * Plugin `@eslint-react/eslint-plugin`, rule `naming-convention/context-name`
  * Plugin `@eslint-react/eslint-plugin/naming-convention`, rule `context-name`

2. Automatically remove `eslint-plugin-` prefix or `/eslint-plugin` postfix from plugin names.

Config would then look like:

```json
{
	"jsPlugins": ["@eslint-react/eslint-plugin"],
	"rules": {
		"@eslint-react/jsx-no-comment-textnodes": ["error"],
		"@eslint-react/dom/no-dangerously-set-innerhtml": ["error"],
		"@eslint-react/naming-convention/context-name": ["error"]
	}
}
```

We should probably look at what ESLint does.

magic-akari:
- Closes: #14557

### DIFF
--- a/crates/oxc_linter/src/config/rules.rs
+++ b/crates/oxc_linter/src/config/rules.rs
@@ -230,9 +230,7 @@ impl<'de> Deserialize<'de> for OxlintRules {
 fn parse_rule_key(name: &str) -> (String, String) {
     // Find the last '/' to split plugin name and rule name.
     // This handles plugin names with slashes like "@eslint-react/eslint-plugin".
-    let slash_idx = name.rfind('/');
-
-    let Some(slash_idx) = slash_idx else {
+    let Some((plugin_name, rule_name)) = name.rsplit_once('/') else {
         return (
             RULES
                 .iter()
@@ -244,9 +242,6 @@ fn parse_rule_key(name: &str) -> (String, String) {
             name.to_string(),
         );
     };
-
-    let plugin_name = &name[..slash_idx];
-    let rule_name = &name[slash_idx + 1..];
 
     let (oxlint_plugin_name, rule_name) = match plugin_name {
         "@typescript-eslint" => ("typescript", rule_name),
--- a/crates/oxc_linter/src/config/rules.rs
+++ b/crates/oxc_linter/src/config/rules.rs
@@ -228,7 +228,11 @@ impl<'de> Deserialize<'de> for OxlintRules {
 }
 
 fn parse_rule_key(name: &str) -> (String, String) {
-    let Some((plugin_name, rule_name)) = name.split_once('/') else {
+    // Find the last '/' to split plugin name and rule name.
+    // This handles plugin names with slashes like "@eslint-react/eslint-plugin".
+    let slash_idx = name.rfind('/');
+
+    let Some(slash_idx) = slash_idx else {
         return (
             RULES
                 .iter()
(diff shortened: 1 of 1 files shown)


overlookmotel (commented):
Thanks for tackling this, but unfortunately I don't think this approach works in all cases.

Some rule names also have slashes in them!

e.g. `@eslint-react/eslint-plugin/naming-convention/context-name` -> plugin `@eslint-react/eslint-plugin`, rule `naming-convention/context-name`.

magic-akari:
> Thanks for tackling this, but unfortunately I don't think this approach works in all cases.
> 
> Some rule names also have slashes in them!
> 
> e.g. `@eslint-react/eslint-plugin/naming-convention/context-name` -> plugin `@eslint-react/eslint-plugin`, rule `naming-convention/context-name`.

Thanks for the review!

You're right that with `@eslint-react/eslint-plugin/naming-convention/context-name`, the plugin name is `@eslint-react/eslint-plugin` and the rule name is `naming-convention/context-name`.

However, as @Rel1cx explained in the issue, this is actually the **legacy config** behavior. In the **flat config**, the parsing changed:
- Plugin name: `@eslint-react/naming-convention` (registered as a "sub-plugin")
- Rule name: `context-name` (no slash)

The `@eslint-react` plugin chose to register sub-plugins like `@eslint-react/naming-convention` to maintain compatibility while adapting to flat config semantics. This means in practice, rule names don't actually contain slashes in the flat config model.

Using `rsplit_once('/')` correctly handles both scenarios:
- Modern flat config: `@eslint-react/naming-convention/context-name` → `@eslint-react/naming-convention` + `context-name` ✅
- Simple plugins: `@eslint-react/jsx-no-comment-textnodes` → `@eslint-react` + `jsx-no-comment-textnodes` ✅

elliotleelewis:
@overlookmotel any qualms taking another look at this? Would love to get this resolved as it's blocking our team from fully replacing ESLint with Oxlint.

overlookmotel:
Yeah I'll be onto this shortly. But #15569 (merged today) is also a solution to the blocker. You can now do:

```json
{
  "jsPlugins": [
    {
      "name": "alias-can-be-whatever-you-like",
      "specifier": "@eslint-react/eslint-plugin/naming-convention"
    }
  ],
  "rules": {
    "alias-can-be-whatever-you-like/context-name": "error"
  }
}
```

overlookmotel:
Superceded by #17073.

PR #14602, opened 2025-10-14
fix(linter): resolve import/extensions false positives and align with ESLint behavior

### LINKED ISSUE #11881: linter: false positive for module with import/extensions rule

1.2.0

### What command did you run?

pnpm oxlint

```jsonc
{
  "categories": {
    "correctness": "off"
  },
  "plugins": ["import"],
  "rules": {
    "import/extensions": [
      "error",
      "always",
      {
        "ignorePackages": true
      }
    ]
  }
}
```

### What happened?

There seems to be a false positive on the `import/extensions` rule for node modules using an export map, e.g. the `vitest` package with its `vitest/config` export. I've prepared a minimum reproduction here: https://github.com/james-camilleri/oxlint-extensions-reproduction.

Running ESlint  on a file with the following line (correctly) indicates no error, while running oxlint returns a false positive – the import statement shouldn't have an extension as it's a module.

```js
import { defineConfig } from "vitest/config";
```

I suspect the error is caused by this line:
https://github.com/oxc-project/oxc/blob/9a2548ae9dd9d218a60673db87ff139316d0525c/crates/oxc_linter/src/rules/import/extensions.rs#L279 
It seems to assume that anything with a `/` in it is not a module. I have absolutely no experience with rust however so this might be completely wrong.

### LINKED ISSUE #12220: linter: eslint-plugin-import(extensions) false positives with missing extension

1.6.0

### What command did you run?

`oxlint -c .oxlintrc.json test.ts`

```jsonc
{
  "env": {
    "browser": true,
    "node": true
  },
  "plugins": [
    "import"
  ],
  "rules": {
    "import/extensions": [
      "error",
      "ignorePackages",
      {
        "js": "never",
        "ts": "never"
      }
    ]
  }
}
```

### What happened?

Import extensions are always required for non-package paths (relative, aliased) since #11872.
With ESLint and `eslint-plugin-import`, there are no errors when extension is missing.

Continuing from #11818.

#### Code
```typescript
import { Package } from 'vue'; // OK, ignorePackages
import { A } from './something.hooks'; // OK
import { B } from './something.ts'; // OK, File extension "ts" should not be included in the import declaration.

import { C } from './something'; // Error, Missing file extension in import declaration
import { D } from '~/common/something'; // Error, Missing file extension in import declaration
```

#### `oxlint` output
```console
❯ oxlint -c .oxlintrc.json test.ts

  × eslint-plugin-import(extensions): Missing file extension in import declaration
   ╭─[test.ts:4:1]
 3 │ 
 4 │ import { C } from './something'; // Error, Missing file extension in import declaration
   · ────────────────────────────────
 5 │ import { D } from '~/common/something'; // Error, Missing file extension in import declaration
   ╰────
  help: Add a file extension to this import.

  × eslint-plugin-import(extensions): File extension "ts" should not be included in the import declaration.
   ╭─[test.ts:2:1]
 1 │ import { A } from './something.hooks'; // OK
 2 │ import { B } from './something.ts'; // OK, File extension "ts" should not be included in the import declaration.
   · ───────────────────────────────────
 3 │ 
   ╰────
  help: Remove the file extension from this import.

  × eslint-plugin-import(extensions): Missing file extension in import declaration
   ╭─[test.ts:5:1]
 4 │ import { C } from './something'; // Error, Missing file extension in import declaration
 5 │ import { D } from '~/common/something'; // Error, Missing file extension in import declaration
   · ───────────────────────────────────────
   ╰────
  help: Add a file extension to this import.
```

### LINKED ISSUE #15509: linter: import/extensions rule should have mjs/mts/cjs/cts options
`.mjs`, `.mts`, `.cjs`, and `.cts` are valid, somewhat-common extensions for JavaScript/TypeScript code. Currently, `import/extensions` does not allow specifying anything of them. The original ESLint rule allows arbitrary values, so some projects migrating over will run into a potential incompatibility here.

Note that when implementing this, ensure that the way we check for extensions does not result in the rule for `js` overriding the rule for `mjs` (because if you just check for "does the filename end with 'js'?", mjs files would also be returned.

https://github.com/oxc-project/oxc/blob/25d577e274a680a785902eca6901d0ec38c801f0/crates/oxc_linter/src/rules/import/extensions.rs

taearls:
Closes #12220 and #16701

This PR fixes multiple false positive scenarios in the `import/extensions` rule and adds significant improvements to make it more robust, performant, and ESLint-compatible.

More thorough PR description generated by Claude Code, reviewed by me:

## Summary

This PR comprehensively fixes multiple false positive scenarios in the `import/extensions` linter rule and significantly improves ESLint compatibility, performance, and extensibility.

---

### 1. Unconfigured Extensions Now Ignored

**Before:**
```js
// Configuration: { "js": "always" }
import styles from './styles.css';  // ERROR: Unexpected use of extension ".css"
import Component from './App.vue';  // ERROR: Unexpected use of extension ".vue"
```

**After:**
```js
// Configuration: { "js": "always" }
import styles from './styles.css';  // OK - .css not configured, ignored
import Component from './App.vue';  // OK - .vue not configured, ignored
```

**Impact:** Eliminates false positives for framework-specific extensions (Vue, Svelte) and custom extensions (CSS, SCSS, etc.)

---

### 2. Root Package Name Detection Fixed

**Before:**
```js
// Configuration: ["never"]
import foo from 'pkg.js';           // ERROR: Extension ".js" should not be used
import babel from '@babel/core.js'; // ERROR: Extension ".js" should not be used
```

**After:**
```js
// Configuration: ["never"]
import foo from 'pkg.js';           // OK - 'pkg.js' is a package name, not a file
import babel from '@babel/core.js'; // OK - '@babel/core.js' is a package name

// But subpaths ARE validated:
import parser from '@babel/core/lib/parser.js'; // ERROR - this IS a file path
```

**Impact:** Correctly distinguishes between package names containing dots (part of the name) vs. file paths with extensions

---

### 3. Path Alias Detection Fixed

**Before:**
```js
// Configuration: ["always"]
import Button from '@/components/Button';  // OK (incorrectly treated as @-scoped package)
import utils from '~/utils/helpers';       // OK (incorrectly ignored)
```

**After:**
```js
// Configuration: ["always"]
import Button from '@/components/Button';     // ERROR: Missing file extension
import Button from '@/components/Button.jsx'; // OK

// Real scoped packages still work correctly:
import babel from '@babel/core';              // OK (actual package)
import types from '@types/node';              // OK (actual package)
```

**Impact:** Critical fix for Vue.js, React, Next.js, and Nuxt projects using path aliases like `@/`, `~/`, or `#/`

---

### 4. `ignorePackages` Behavior Now Matches ESLint

**Before:**
- Default: `true`
- Behavior: Unclear, affected both "always" and "never" rules inconsistently

**After:**
- Default: `false` (matches ESLint)
- Behavior: Only affects "always" rule, has no effect on "never" rule

```js
// Configuration: ["always", { "ignorePackages": true }]
import foo from 'lodash';        // OK - package exempted from "always"
import bar from './bar.js';      // OK - relative import requires extension

// Configuration: ["never", { "ignorePackages": true }]
import foo from 'lodash/fp.js';  // ERROR - ignorePackages doesn't affect "never"
```

---

### Case-Insensitive Extension Matching

**Before:**
```js
// Configuration: { "js": "always" }
import foo from './foo.JS';  // ERROR: Missing file extension (uppercase not recognized)
```

**After:**
```js
// Configuration: { "js": "always" }
import foo from './foo.JS';  // OK - extensions normalized to lowercase
import bar from './bar.Ts';  // OK - .Ts matches "ts" config
```

---

### Module Resolution Integration

**Before:** Only checked the string in the import statement

**After:** Uses `ModuleRecord::resolved_absolute_path` to validate against actual filesystem files

```js
// File: ./utils.ts exists
import utils from './utils.js';  // Now detects mismatch between written ".js" and actual ".ts"
```

**Benefits:**
- More accurate validation against actual files
- Handles path alias resolution correctly
- Better diagnostics for extension mismatches

---

### Arbitrary Extension Support

**Before:** Only supported hardcoded extensions: `js`, `jsx`, `ts`, `tsx`, `json`

**After:** Supports ANY extension

```json
// .eslintrc.json
{
  "rules": {
    "import/extensions": ["error", {
      "vue": "always",
      "svelte": "never",
      "mjs": "always",
      "cjs": "never",
      "css": "always",
      "scss": "always"
    }]
  }
}
```

---

### Path Group Overrides (Bespoke Import Specifiers)

New option for custom import protocols used by monorepo tools and custom resolvers.

```json
{
  "rules": {
    "import/extensions": ["error", "always", {
      "pathGroupOverrides": [
        { "pattern": "rootverse{*,*/**}", "action": "ignore" }
      ]
    }]
  }
}
```

**Pattern Matching:**
- `*` - Match any characters except `/`
- `**` - Match any characters including `/` (recursive)
- `{a,b}` - Match alternatives

**Actions:**
- `"enforce"` - Apply normal extension validation
- `"ignore"` - Skip all extension validation for matching imports

```js
// With above config:
import { x } from 'rootverse+debug:src';       // OK - ignored by pathGroupOverrides
import { y } from 'rootverse+bfe:src/symbols'; // OK - ignored by pathGroupOverrides
import foo from './foo';                        // ERROR - still requires extension
```

---

### Zero-Copy FxHashMap Configuration

**Before:**
```rust
struct ExtensionsConfig {
    js: FileExtensionConfig,
    jsx: FileExtensionConfig,
    ts: FileExtensionConfig,
    tsx: FileExtensionConfig,
    json: FileExtensionConfig,
}
```

**After:**
```rust
#[repr(u8)]  // 1 byte per variant
pub enum ExtensionRule {
    Always = 0,
    Never = 1,
    IgnorePackages = 2,
}

struct ExtensionsConfig {
    extensions: FxHashMap<String, &'static ExtensionRule>,
}
```

**Performance Characteristics:**
- O(1) lookups (~3-6ns for 500 entries)
- Static instances in read-only memory (zero allocations)
- Pointer equality for rule comparison
- `#[inline]` attributes on hot paths

---

### Default Behavior

**Before:** Pre-populated default rules for common extensions could cause false positives

**After:** Unconfigured extensions are ignored - no false positives by default

### ignorePackages Default

**Before:** `true`

**After:** `false` (matches ESLint)

### Configuration Formats Supported

```js
// 1. Global rule only
["always"]
["never"]
["ignorePackages"]

// 2. Per-extension only (no global rule)
[{ "js": "always", "json": "never" }]

// 3. Combined
["always", { "js": "never", "vue": "always" }]

// 4. With options
["always", {
  "ignorePackages": true,
  "checkTypeImports": true,
  "pathGroupOverrides": [...]
}]
```

---

## Testing

- **Added:** 100+ new test cases covering edge cases
- **Removed:** 21 redundant test cases
- **Result:** All tests passing with zero regressions

### Coverage Areas
- Path aliases (`@/`, `~/`, `#/`)
- Scoped packages (`@babel/core`, `@types/node`)
- Root package names with dots (`pkg.js`, `@name/pkg.js`)
- Case-insensitive extensions (`./foo.JS`, `./bar.Ts`)
- Custom extensions (`.vue`, `.mjs`, `.css`)
- Query strings and fragments
- Unicode filenames
- Monorepo protocols (bespoke import specifiers)

### Real-World Validation
- Tested against Mastodon repository
- Reduced false positive errors from 169 to 120

---

## Documentation Updates

- Expanded rule documentation with configuration examples
- Added "Known Limitations" section for package.json detection
- Clarified `ignorePackages` semantics
- Added `pathGroupOverrides` documentation
- Added per-extension configuration examples

---

### Breaking Changes

1. **`ignorePackages` default changed from `true` to `false`**

   If you relied on the old default, explicitly set it:
   ```json
   ["always", { "ignorePackages": true }]
   ```

2. **`ignorePackages` no longer affects "never" rule**

   This was undefined/inconsistent behavior before. Now it only exempts packages from the "always" rule.

### Non-Breaking Improvements

- Existing configurations continue to work
- False positives reduced significantly
- Better ESLint compatibility


camc314 (commented):
Would it be better to use use the module record for this.

code like:
```
    // Check for path aliases with single-char prefix followed by /
    // Examples: @/, ~/, #/
    if module_name.len() >= 2 {
        let chars: Vec<char> = module_name.chars().collect();
        if chars.len() >= 2 && chars[1] == '/' {
            // Single character before slash - this is a path alias
            return false;
        }
    }
```

feels flaky

taearls:
> Would it be better to use use the module record for this.
> 
> code like:
> 
> ```
>     // Check for path aliases with single-char prefix followed by /
>     // Examples: @/, ~/, #/
>     if module_name.len() >= 2 {
>         let chars: Vec<char> = module_name.chars().collect();
>         if chars.len() >= 2 && chars[1] == '/' {
>             // Single character before slash - this is a path alias
>             return false;
>         }
>     }
> ```
> 
> feels flaky

yeah that makes sense. I'll work on refactoring this to use the module record.

Update: 

after looking at it a bit more, I couldn't find any additional contextual information in the module record (aside from the module's import name that is already being used) that I could leverage to my benefit. I did update the helper function this snippet is in to be a little cleaner and more clear with its intentions.

as far as I can tell, the linter doesn't support custom resolver configurations in `tsconfig.json` yet, but I think this fixes the specific bugs that were reported in the issue. there are still some gaps though I believe. any custom aliases or typescript paths that are configured in a user's `tsconfig.json` will not be respected yet. 

however, I think adding full resolver config support deserves its own dedicated issue / PR because the filesystem lookups to resolve the `tsconfig.json` (and to safely handle if it doesn't exist) would have performance implications.

maybe I'm missing something though. let me know if you have any suggestions in terms of how I can use the module record better here

connorshea:
This MR does fix various issues in [the reproduction repo here](https://github.com/connorshea/oxlint-extensions-reproduction).

The only things it doesn't fix are [`main4.js`](https://github.com/connorshea/oxlint-extensions-reproduction/blob/03889fe319619b12d0c9cafb624f49d150ecf801/main4.js) and [`main5.js`](https://github.com/connorshea/oxlint-extensions-reproduction/blob/main/main5.js) but we can try to handle those in a different PR.

`main4.js` should fail because I have the config set as `"import/extensions": ["error", "always", { "ignorePackages": true }]`, and main4 has `import foo from "./foo";` in it. Weirdly, if it doesn't have the `console.log(foo)` line after the import, it _does_ trigger the lint error. So that's weird.

<details>
<summary>linting output on the repro repo for eslint vs main vs this branch</summary>

oxlint 1.26.0 (3 errors, and 1 error missing that is expected, based on what ESLint does):
```
connorshea@Connors-MacBook-Pro oxlint-extensions-reproduction % pnpm oxlint

  × eslint-plugin-import(extensions): File extension "JS" should not be included in the import declaration.
   ╭─[main5.js:3:1]
 2 │ // ESLint passes on this, oxlint does not currently.
 3 │ import foo from './foo.JS';
   · ───────────────────────────
 4 │ 
   ╰────
  help: Remove the file extension from this import.

  × eslint-plugin-import(extensions): Missing file extension in import declaration
   ╭─[main1.js:4:1]
 3 │ // We have `ignorePackages` set to true, so this should not error.
 4 │ import { defineConfig } from "vitest/config";
   · ─────────────────────────────────────────────
 5 │ 
   ╰────
  help: Add a file extension to this import.

  × eslint-plugin-import(extensions): Missing file extension in import declaration
   ╭─[main2.js:3:1]
 2 │ // This behavior is not correct because I have `ignorePackages` set to true, so it should be... ignored.
 3 │ import { defineConfig } from "vitest/config";
   · ─────────────────────────────────────────────
 4 │ 
   ╰────
  help: Add a file extension to this import.

Found 0 warnings and 3 errors.
Finished in 49ms on 7 files using 8 threads.
```

This PR (after being rebased on top of main, 1 error but it's different from ESLint's error):

```
connorshea@Connors-MacBook-Pro oxlint % node dist/cli.js -c ../../../oxlint-extensions-reproduction/.oxlintrc.json ~/code/oxlint-extensions-reproduction/

  × eslint-plugin-import(extensions): File extension "JS" should not be included in the import declaration.
   ╭─[/Users/connorshea/code/oxlint-extensions-reproduction/main5.js:3:1]
 2 │ // ESLint passes on this, oxlint does not currently.
 3 │ import foo from './foo.JS';
   · ───────────────────────────
 4 │ 
   ╰────
  help: Remove the file extension from this import.

Found 0 warnings and 1 error.
Finished in 324ms on 7 files with 1 rules using 8 threads.
```

ESLint (1 error):

```
connorshea@Connors-MacBook-Pro oxlint-extensions-reproduction % pnpm eslint

/Users/connorshea/code/oxlint-extensions-reproduction/main4.js
  5:17  error  Missing file extension "js" for "./foo"  import/extensions

✖ 1 problem (1 error, 0 warnings)
```

</details>

---
Aside on main5.js:

I'm not sure if we should _necessarily_ downcase the extension filename to fix `main5,js` there? And *technically* the difference may be entirely based on eslint allowing any extensions not-defined-by-the-config by default, while oxlint doesn't? So if it's because 'JS' is _technically_ a distinct extension from 'js', then maybe we would fix this behavior in a different way. :shrug:

Regardless, I don't think that edge case needs to be fixed in this MR. I do think we should consider adding a bunch of fixtures for this rule, to confirm various complex behaviors, but that can also be done in a separate PR imo.

connorshea:
This definitely seems to improve the number of discrepancies between oxlint and ESLint, based on testing it on [the Mastodon repo](https://github.com/mastodon/mastodon/blob/6337e036f3bc5ece013d5b2536d7d16b1ec658e8/eslint.config.mjs).

1.26.0 on mastodon: 169 errors
this PR (after rebase) on mastodon: 120 errors

However, I think more work will need to be done to fix further issues with this rule, as things like this are showing up as violations still after this PR, despite not being violations (config is `always` at the top-level, then `never` for importing `tsx` files):

```
x eslint-plugin-import(extensions): Missing file extension in import declaration
   ,-[app/javascript/mastodon/features/emoji/emoji_picker.tsx:7:1]
 6 | 
 7 | import { EMOJI_MODE_NATIVE } from './constants';
   : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 8 | import EmojiData from './emoji_data.json';
   `----
  help: Add a file extension to this import.
```

See [`emoji_picker.tsx`](https://github.com/mastodon/mastodon/blob/85d0cdb5f73850dc7474c5e2e64b3aa8513bd1d3/app/javascript/mastodon/features/emoji/emoji_picker.tsx). The `constants` file being imported is at `app/javascript/mastodon/features/emoji/constants.ts`, and so _should_ be enforced to never have an extension. Yet the rule still enforces it.

I'm wondering if this rule actually correctly has the ability to check the source file's type when it's being imported?

The `.oxlintrc.json` in mastodon I'm testing with is this, if anyone wants to reproduce this behavior:

<details>
<summary>oxlintrc.json</summary>

```json
{
  "plugins": [
    "import"
  ],
  "categories": {
    "correctness": "off"
  },
  "env": {
    "builtin": true,
    "es2021": true,
    "browser": true,
    "shared-node-browser": true
  },
  "rules": {
    "import/extensions": [
      "error",
      "always",
      {
        "js": "never",
        "jsx": "never",
        "ts": "never",
        "tsx": "never"
      }
    ]
  },
  "ignorePatterns": [
    "build/**/*",
    "coverage/**/*",
    "db/**/*",
    "lib/**/*",
    "log/**/*",
    "node_modules/**/*",
    "public/**/*",
    "!public/embed.js",
    "spec/**/*",
    "tmp/**/*",
    "vendor/**/*",
    "streaming/**/*",
    ".bundle/**/*",
    "storybook-static/**/*"
  ]
}
```
</details>

taearls:
> This definitely seems to improve the number of discrepancies between oxlint and ESLint, based on testing it on [the Mastodon repo](https://github.com/mastodon/mastodon/blob/6337e036f3bc5ece013d5b2536d7d16b1ec658e8/eslint.config.mjs).
> 
> 1.26.0 on mastodon: 169 errors
> this PR (after rebase) on mastodon: 120 errors
> 
> However, I think more work will need to be done to fix further issues with this rule, as things like this are showing up as violations still after this PR, despite not being violations (config is `always` at the top-level, then `never` for importing `tsx` files):
> 
> ```
> x eslint-plugin-import(extensions): Missing file extension in import declaration
>    ,-[app/javascript/mastodon/features/emoji/emoji_picker.tsx:7:1]
>  6 | 
>  7 | import { EMOJI_MODE_NATIVE } from './constants';
>    : ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>  8 | import EmojiData from './emoji_data.json';
>    `----
>   help: Add a file extension to this import.
> ```
> 
> See [`emoji_picker.tsx`](https://github.com/mastodon/mastodon/blob/85d0cdb5f73850dc7474c5e2e64b3aa8513bd1d3/app/javascript/mastodon/features/emoji/emoji_picker.tsx). The `constants` file being imported is at `app/javascript/mastodon/features/emoji/constants.ts`, and so _should_ be enforced to never have an extension. Yet the rule still enforces it.
> 
> I'm wondering if this rule actually correctly has the ability to check the source file's type when it's being imported?
> 
> The `.oxlintrc.json` in mastodon I'm testing with is this, if anyone wants to reproduce this behavior:
> 
> <details>
> <summary>oxlintrc.json</summary>
> 
> ```json
> {
>   "$schema": "./node_modules/oxlint/configuration_schema.json",
>   "plugins": [
>     "import"
>   ],
>   "categories": {
>     "correctness": "off"
>   },
>   "env": {
>     "builtin": true,
>     "es2021": true,
>     "browser": true,
>     "shared-node-browser": true
>   },
>   "rules": {
>     "import/extensions": [
>       "error",
>       "always",
>       {
>         "js": "never",
>         "jsx": "never",
>         "ts": "never",
>         "tsx": "never"
>       }
>     ]
>   },
>   "ignorePatterns": [
>     "build/**/*",
>     "coverage/**/*",
>     "db/**/*",
>     "lib/**/*",
>     "log/**/*",
>     "node_modules/**/*",
>     "public/**/*",
>     "!public/embed.js",
>     "spec/**/*",
>     "tmp/**/*",
>     "vendor/**/*",
>     "streaming/**/*",
>     ".bundle/**/*",
>     "storybook-static/**/*"
>   ]
> }
> ```
> </details>

Good to know! Thanks for the detailed report. I can take a look this week into this further, but I'll probably do it in a separate development branch because I suspect it will be a more significant refactor to get all those things working.

I don't think the file extension parsing is working how we would expect. When I initially implemented this rule, I basically just finished once I got all the tests to pass, but there have been a few problems with this rule.

I agree that we should add more testing to verify the behavior. It deserves a closer look for sure.

connorshea:
See also #15009, not sure which is the best to go with as the solution here but I definitely think we need to prioritize fixing this rule.

taearls:
I'm not sure either. From the discord conversation I'm also working on a separate branch that I was planning to point to this one when it's ready that will support custom extensions (i.e., not hard coded ones).

taearls:
I discussed and tested this rule with @connorshea on discord and pushed up some changes that will significantly improve the behavior of this rule significantly. 

the key highlights: 

- I updated the parsing logic so that if a file extension is not included in the configuration, and the global setting is NOT `never`, then this rule will not emit diagnostics about files with that extension. this makes it so users get custom file extension support without needing to manually configure potentially dozens of file extensions in their config. Users can incrementally flag what they want and have fine control over what extensions are allowed.

- the parsing logic uses the ModuleRecord struct rather than string matching against the source code. this means that we'll be able to check against the _actual_ file extension in in the filesystem rather than simply performing static analysis on source code.
- I removed the hard coded file extension configurations and am using an FxHashMap to build up a mapping based on the user's configuration.
- I used bitflags to denote `always` | `never` | `ignorePackages` so that the enum is memory efficient.
- added support for the `pathGroupOverrides` configuration option

I'm planning to review it more thoroughly over the next couple of days to make sure everything looks good and that I didn't miss anything, but meanwhile I think it's in a good enough place for review.

taearls:
> I discussed and tested this rule with @connorshea on discord and pushed up some changes that will significantly improve the behavior of this rule significantly.
> 
> the key highlights:
> 
> * I updated the parsing logic so that if a file extension is not included in the configuration, and the global setting is NOT `never`, then this rule will not emit diagnostics about files with that extension. this makes it so users get custom file extension support without needing to manually configure potentially dozens of file extensions in their config. Users can incrementally flag what they want and have fine control over what extensions are allowed.
> * the parsing logic uses the ModuleRecord struct rather than string matching against the source code. this means that we'll be able to check against the _actual_ file extension in in the filesystem rather than simply performing static analysis on source code.
> * I removed the hard coded file extension configurations and am using an FxHashMap to build up a mapping based on the user's configuration.
> * I used bitflags to denote `always` | `never` | `ignorePackages` so that the enum is memory efficient.
> * added support for the `pathGroupOverrides` configuration option
> 
> I'm planning to review it more thoroughly over the next couple of days to make sure everything looks good and that I didn't miss anything, but meanwhile I think it's in a good enough place for review.

I'm feeling good about these changes after giving them a quick review and updating the PR description.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -35,83 +37,151 @@ fn extension_missing_diagnostic(span: Span, is_import: bool) -> OxcDiagnostic {
...
+    IgnorePackages = 2,
 }
 
-impl FileExtensionConfig {
-    pub fn from(s: &str) -> FileExtensionConfig {
+/// Static instances of ExtensionRule variants for zero-copy reference sharing.
+/// These are stored in read-only memory and accessed via references throughout the application.
+static ALWAYS: ExtensionRule = ExtensionRule::Always;
+static NEVER: ExtensionRule = ExtensionRule::Never;
+static IGNORE_PACKAGES: ExtensionRule = ExtensionRule::IgnorePackages;
```
camchenry: I don't think these static variables are necessary. `ExtensionRule` is the size of a single byte, which should be nearly free to copy on any instruction set. I think we can make this type derive `Copy` (and `Clone`) instead:


```suggestion
#[repr(u8)]
#[derive(Copy, Debug, PartialEq, Eq, JsonSchema, Serialize)]
pub enum ExtensionRule {
    Always = 0,
    Never = 1,
    IgnorePackages = 2,
}
```

camchenry (commented):
Seems like this is headed in the right direction overall. I think the `ExtensionRule` not being copyable does overcomplicate some of the logic though, so I'd recommend changing that.

This does add a lot of code, but it also seems like it fixes some of the major issues with the rule. I'm inclined to say we should merge this, but could we check if there's any way to simplify this code? The original rule source is only about 100 lines of code: https://github.com/import-js/eslint-plugin-import/blob/01c9eb04331d2efa8d63f2d7f4bfec3bc44c94f3/src/rules/extensions.js#L4

It seems like maybe we're missing some ways to simplify these checks? Might as well have Claude do a comparison.

sam-goodwin:
What's the update on this PR? These changes are a must-have for my project to protect me from publishing broken ES modules.

connorshea:
mastodon w/ oxlint run result. With JS Plugins and type-aware rules.

```
Before, with oxlint 1.32.0 (--type-aware):

Found 9 warnings and 175 errors.
Finished in 7.7s on 584 files using 8 threads.

After this PR (--type-aware):

Found 9 warnings and 17 errors.
Finished in 6.7s on 584 files using 8 threads.
```

EDIT: Confirmed that I get the same results with the commit I just pushed.

crates/oxc_linter/src/rules/import/extensions.rs
connorshea: Pushed a commit to implement this change.

taearls:
> Seems like this is headed in the right direction overall. I think the `ExtensionRule` not being copyable does overcomplicate some of the logic though, so I'd recommend changing that.
> 
> This does add a lot of code, but it also seems like it fixes some of the major issues with the rule. I'm inclined to say we should merge this, but could we check if there's any way to simplify this code? The original rule source is only about 100 lines of code: https://github.com/import-js/eslint-plugin-import/blob/01c9eb04331d2efa8d63f2d7f4bfec3bc44c94f3/src/rules/extensions.js#L4
> 
> It seems like maybe we're missing some ways to simplify these checks? Might as well have Claude do a comparison.

yeah, that's a good point. I'll give that a go today.

From Claude:

Major Simplification Opportunities:
  1. process_module_record() and process_require_record() are 95% identical (~170 lines each)
  2. Repeated has_never_rules pattern (6+ occurrences, 6 lines each)
  3. Repeated has_standard_extension_rules pattern (4 occurrences)
  4. Unnecessary config clone in run_once()
  5. Parameter drilling (passing config fields as params instead of accessing from self)
  
I'll start working through these items and then test that the rule behaves the same in Mastodon to validate the changes.

taearls:
I believe I've addressed all the feedback on this PR. I was able to simplify the implementation a decent amount and verify that it still behaves the same against the rule's testing and against mastodon's codebase.

I think this is in a good place for a re-review when you get a chance @camchenry @camc314

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -130,6 +238,37 @@ declare_oxc_lint!(
     ///
     /// ESM-based file resolve algorithms (e.g., the one that Vite provides) recommend specifying the file extension to improve performance.
     ///
+    /// ### Configuration
```
connorshea: I believe you can put this comment above the config struct and it'll be included in the website right before the config options that are generated automatically. It's preferable that we not manually create a "configuration" header like this. Not a blocker, though. I can fix it myself in the future.

connorshea:
Rebuilt the latest version of this branch and ran it on the Mastodon codebase, confirmed it still works well :)

Before, with oxlint 1.32.0 (--type-aware --report-unused-disable-directives):

```
Found 33 warnings and 152 errors.
```

After, with this PR (--type-aware --report-unused-disable-directives):
```
Found 33 warnings and 12 errors.
```

(any differences in the starting #s vs the last comment I made are generally due to the `--report-unused-disable-directives` and fixes I've been working on for the `@oxlint/migrate` package)

crates/oxc_linter/src/rules/import/extensions.rs
taearls: Ahh good catch! I can fix that tomorrow.

Update: Done!

connorshea (approved):
From a docs/"does the rule work" perspective, LGTM. Will let cam or cam review and approve the actual rust code.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -116,8 +116,35 @@ impl PathGroupOverride {
...
+/// **Default behavior (no configuration)**: All imports pass. Unconfigured file extensions are ignored to avoid false positives.
+///
+/// **ignorePackages option**:
+/// - A boolean option (not per-extension) that exempts package imports from the "always" rule
+/// - Can be set in the config object: `["always", { "ignorePackages": true }]`
+/// - Legacy shorthand: `["ignorePackages"]` is equivalent to `["always", { "ignorePackages": true }]`
+/// - **Default: `false`** (matches ESLint behavior)
+/// - **With "always"**: When `true`, package imports (e.g., `lodash`, `@babel/core`) don't require extensions
+/// - **With "never"**: This option has no effect; extensions are still forbidden on package imports
+/// - Example: `["always", { "ignorePackages": true }]` allows `import foo from "lodash"` but requires `import bar from "./bar.js"`
```
camchenry: These should probably be broken out to be doc comments on the individual options themselves (`ignore_packages` and `path_group_overrides`)

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -116,8 +116,35 @@ impl PathGroupOverride {
...
+/// 2. **Per-extension rules** (object): `{ "js": "always", "jsx": "never", ... }`
+/// 3. **Combined** (array): `["always", { "js": "never" }]` or `[{ "js": "always" }]`
+///
+/// **Default behavior (no configuration)**: All imports pass. Unconfigured file extensions are ignored to avoid false positives.
+///
+/// **ignorePackages option**:
+/// - A boolean option (not per-extension) that exempts package imports from the "always" rule
+/// - Can be set in the config object: `["always", { "ignorePackages": true }]`
+/// - Legacy shorthand: `["ignorePackages"]` is equivalent to `["always", { "ignorePackages": true }]`
+/// - **Default: `false`** (matches ESLint behavior)
```
camchenry: We can omit the default value since we already serialize this automatically.


```suggestion

```

crates/oxc_linter/src/rules/import/extensions.rs
camchenry, line 141: I think if the `require_extension` is not configurable in the object, we should omit it? That is, I think we should add `#[serde(skip)]` here so it doesn't appear as if it was a supported config option in the docs.

cc @connorshea since you've been doing a lot of work in this area, correct me if I'm wrong.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -35,82 +36,216 @@ fn extension_missing_diagnostic(span: Span, is_import: bool) -> OxcDiagnostic {
...
+#[derive(Debug, Clone, JsonSchema, Serialize, Default)]
 #[serde(rename_all = "camelCase", default)]
 pub struct ExtensionsConfig {
     /// Whether to ignore package imports (e.g., 'react', 'lodash') when enforcing extension rules.
     ignore_packages: bool,
-    /// Configuration for requiring or disallowing file extensions in import/require statements.
-    require_extension: Option<FileExtensionConfig>,
-    /// Whether to check type imports when enforcing extension rules.
+    require_extension: Option<ExtensionRule>,
     check_type_imports: bool,
```
camchenry, line 143: The doc comment for this property was removed.


```suggestion
    /// Whether to check type imports when enforcing extension rules
    check_type_imports: bool,
```

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -35,82 +36,216 @@ fn extension_missing_diagnostic(span: Span, is_import: bool) -> OxcDiagnostic {
...
+    /// Check if any standard extension has a "never" rule configured.
+    /// Used for lenient behavior when module resolution is unavailable.
+    #[inline]
+    pub fn has_any_never_rules(&self) -> bool {
+        self.is_never("js")
+            || self.is_never("jsx")
+            || self.is_never("ts")
+            || self.is_never("tsx")
+            || self.is_never("mjs")
+            || self.is_never("cjs")
```
camchenry, line 213: Should `json` be in this list as well?

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
-                    return;
+                    continue;
                 };
-                let func_name = ident.name.as_str();
-                let count = call_expr.arguments.len();
-
-                if matches!(func_name, "require") && count > 0 {
-                    self.process_require_record(call_expr, ctx, config.require_extension.as_ref());
+                if ident.name.as_str() == "require" && !call_expr.arguments.is_empty() {
+                    for argument in &call_expr.arguments {
```
camchenry: The arguments length check is semi-redundant, as iterating over an empty iterator will do mostly nothing anyway:


```suggestion
                if ident.name.as_str() == "require" {
                    for argument in &call_expr.arguments {
```

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
         }
     }
 
     fn run_once(&self, ctx: &LintContext) {
-        let module_record = ctx.module_record();
-
-        let config = self.0.clone();
-
+        // Process require() calls
         for node in ctx.nodes().iter() {
```
camchenry: Note for follow-up: Could we move this to an additional `fn run` implementation instead? Rules can implement both `run` and `run_once`, and that would take the explicit node iteration out of this. That's helpful to do because we can optimize how often we call `run`.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
-            .get("json")
-            .and_then(Value::as_str)
-            .map(FileExtensionConfig::from)
-            .unwrap_or(default.cloned().unwrap_or_default()),
+/// Build configuration from JSON value with optional default rule.
+///
+/// This function dynamically parses extension configurations from JSON, supporting
+/// both individual extension fields (js, jsx, ts, tsx, json, etc.) and arbitrary
+/// custom extensions.
+fn build_config(value: &serde_json::Value, default: Option<ExtensionRule>) -> ExtensionsConfig {
```
camchenry: I think it's more idiomatic to define this as a static method of `ExtensionsConfig` so that it's associated with it.

Example:

```rs
impl ExtensionsConfig {
  fn from_json_value(value: &serde_json::Value, default: Option<ExtensionRule>) -> Self { /* ... */ }
  
  // ...
}
```

Then we instead of `let config = build_config(...)` we can write `let config = ExtensionsConfig::from_json_value(...)` so it's more explicit what we are returning here.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
-            return;
+        // Prefer resolved extension (actual file), fallback to written extension (import text)
+        let extension_to_check = resolved_extension.or(written_extension.map(CompactStr::as_str));
+
+        if let Some(ext_str) = extension_to_check {
+            // Skip validation for unconfigured extensions (prevents false positives)
+            // unless there's a global rule or it's a standard extension
+            if !config.has_rule(ext_str)
+                && require_extension.is_none()
+                && !ExtensionsConfig::is_standard_extension(ext_str)
```
camchenry: micro-optimization: put the relatively cheap operation `is_none` before the `match` and hash table lookup:


```suggestion
            if require_extension.is_none()
                && !ExtensionsConfig::is_standard_extension(ext_str)
                && !config.has_rule(ext_str) 
```

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
-                _ => config.is_never(ext_str),
+        // Get extensions
+        let resolved_extension = get_resolved_extension(ctx.module_record(), module_name);
+
+        // For ROOT packages, don't extract extensions - dots are part of package names
+        // Exception: if pathGroupOverrides explicitly enforces validation
+        let is_path_group_action_enforce =
+            matches!(path_group_action, Some(PathGroupAction::Enforce));
+        let written_extension =
+            if is_root_package_import(module_name) && !is_path_group_action_enforce {
```
camchenry: No need to compute this until we need it:
```suggestion
        let written_extension =
            if is_root_package_import(module_name) && !matches!(path_group_action, Some(PathGroupAction::Enforce)) {
```

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
+/// to lowercase for case-insensitive matching.
+///
+/// # Examples
+/// - `"./foo.js"` → `Some("js")`
+/// - `"./foo.JS"` → `Some("js")` (normalized to lowercase)
+/// - `"./foo.js?v=123"` → `Some("js")` (query params stripped)
+/// - `"./foo"` → `None`
+/// - `"./foo."` → `None` (empty extension)
+/// - `"./foo.bar/"` → `None` (directory path)
 fn get_file_extension_from_module_name(module_name: &CompactStr) -> Option<CompactStr> {
```
camchenry: Note for follow-up: can we remove `CompactStr` here? We are converting from `&str` to `CompactStr` but then calling `CompactStr::as_str` in `validate_extension`. It would be simpler to just use `&str` and `Cow` directly, without converting to `CompactStr`.

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
+    /// validation logic. This ensures require() now correctly checks pathGroupOverrides,
+    /// built-in modules, and ignorePackages (which was previously missing).
+    fn process_import(
+        &self,
+        ctx: &LintContext,
+        module_name: &str,
+        span: Span,
+        is_type_import: bool,
+        is_import: bool,
+    ) {
```
camchenry, line 533: Noting for follow-up: We could rearrange the order of these checks to do the fastest ones first, and the slowest ones last (`binary_search`, hash table lookups, etc.)

crates/oxc_linter/src/rules/import/extensions.rs
```diff
@@ -179,220 +326,407 @@ declare_oxc_lint!(
...
-                    ctx.diagnostic(extension_missing_diagnostic(span, true));
-                }
-            }
+    // Other single-char path aliases: ~/, #/
+    if module_name.len() >= 2 {
+        let bytes = module_name.as_bytes();
+        if bytes[1] == b'/' && bytes[0] != b'.' && bytes[0] != b'@' {
+            return false; // Path alias like ~/ or #/
         }
     }
```
camchenry: Note for follow-up: what's the faster way to peek the first two bytes of a str? Haven't checked the assembly, but I wonder if this incurs three bounds checks instead of just one.

camchenry (approved):
I think this looks pretty great! I'm not too familiar with the original rule, but I've read over the logic here and I think it makes sense.

I've left a number of comments and I would make some of these changes myself but I'm not currently at my laptop to do so. But hopefully this is also instructive for future PRs for the types of things I'm thinking about (and AI tends to miss on). I don't think any of the comments are particularly time consuming though.

I also left a few follow-up comments that are non-blocking and just notes for future PRs on top of this work.

Overall, I'm excited to get this merged in, as this is a big improvement to this rule. Since this is an existing rule with issues, I don't want to let perfection get in the way of good enough, so I'm in favor of these changes after feedback if @camc314 is good with it too.

taearls:
> I think this looks pretty great! I'm not too familiar with the original rule, but I've read over the logic here and I think it makes sense.
> 
> I've left a number of comments and I would make some of these changes myself but I'm not currently at my laptop to do so. But hopefully this is also instructive for future PRs for the types of things I'm thinking about (and AI tends to miss on). I don't think any of the comments are particularly time consuming though.
> 
> I also left a few follow-up comments that are non-blocking and just notes for future PRs on top of this work.
> 
> Overall, I'm excited to get this merged in, as this is a big improvement to this rule. Since this is an existing rule with issues, I don't want to let perfection get in the way of good enough, so I'm in favor of these changes after feedback if @camc314 is good with it too.

I appreciate your thoughtful feedback! It's helpful to get more insight into what the best practices are for these linting rules and what anti patterns you're checking for. I'll definitely keep that in mind for future contributions.

I pushed up commits to address each comment of feedback. Everything still works as before, but everything should be addressed now. I'm hoping that separate commits for each improvement will make it easier to review each of the changes individually.

crates/oxc_linter/src/rules/import/extensions.rs
connorshea: You are correct, and looks like it has been resolved now! :)

camchenry (approved):

taearls:
@camc314 when you get a chance can I get a review on this? I think it's good to go. Merging this will close 3 open issues.

camc314 (approved):

sebastienbarre:
The documentation for this rule is so incredibly confusing. Is there any human who has figured out the actual configuration format?  Is this below for real (from an earlier comment)? I have tried so many combinations. Where does the array start? Where does the object start? Is "always" before or after "error"? Is it the second argument to the array? Or is it in the object? What is the implication of "ignorePackages"? Does it mean "ignore package" but the rest is considered "always", or the rest is considered "never"? etc.
```
// 1. Global rule only
["always"]
["never"]
["ignorePackages"]

// 2. Per-extension only (no global rule)
[{ "js": "always", "json": "never" }]

// 3. Combined
["always", { "js": "never", "vue": "always" }]

// 4. With options
["always", {
  "ignorePackages": true,
  "checkTypeImports": true,
  "pathGroupOverrides": [...]
}]
```

connorshea:
Improved the docs a bit more here #17162, we'll need to update them again in the nearish future once we have proper tuple config support.

Thanks for the work on this @taearls!

PR #14631, opened 2025-10-15
perf(linter): detect diverging match blocks for node type skipping

camchenry:
Adds a new kind of code analysis to the linter codegen. This new detector looks for diverging statements in lint rules, like:

```rust
let something = match node.kind() {
  AstKind::CallExpression(_) => { ... }
  _ => return,
}
```

The important part is that the expression _must_ diverge (i.e., it must have something like a `return` in its wildcard block). That way, we know only the matched node types in the match arms are used in the rule.

### DIFF
--- a/tasks/linter_codegen/src/early_diverge_detector.rs
+++ b/tasks/linter_codegen/src/early_diverge_detector.rs
@@ -1,0 +1,78 @@
+use syn::{Arm, Expr, Pat, Stmt};
+
+use crate::{
+    CollectionResult, NodeTypeSet,
+    utils::{astkind_variant_from_path, is_node_kind_call},
+};
+
+/// Detects various kinds of diverging statements that narrow by more than one AST node type.
+pub struct EarlyDivergeDetector {
+    node_types: NodeTypeSet,
+}
+
+impl EarlyDivergeDetector {
+    pub fn from_run_func(run_func: &syn::ImplItemFn) -> Option<NodeTypeSet> {
+        // Only look at cases where the body has more than one top-level statement.
+        let block = &run_func.block;
+        if block.stmts.len() <= 1 {
+            return None;
+        }
+
+        // Look at the first statement in the function body.
+        let stmt = block.stmts.first()?;
+
+        // Check if it's `let something = match node.kind() { ... }` that diverges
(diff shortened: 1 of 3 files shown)


camchenry:
* **#14631** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14631?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14631?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14630** <a href="https://app.graphite.dev/github/pr/oxc-project/oxc/14630?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=14631).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camc314 (approved):
nice! this is great to see

tasks/linter_codegen/src/early_diverge_detector.rs
```diff
@@ -0,0 +1,78 @@
...
+impl EarlyDivergeDetector {
+    pub fn from_run_func(run_func: &syn::ImplItemFn) -> Option<NodeTypeSet> {
+        // Only look at cases where the body has more than one top-level statement.
+        let block = &run_func.block;
+        if block.stmts.len() <= 1 {
+            return None;
+        }
+
+        // Look at the first statement in the function body.
+        let stmt = block.stmts.first()?;
```
overlookmotel, line 22: Just a small note here:

IMO it'd likely be preferable to use `block.stmts.first().unwrap()`.

`.unwrap()` will often cause the compiler to backtrack and try to prove that the `unwrap` can't fail. In this case I'd assume it can prove this fairly easily as the `block.stmts.len() <= 1` check is right above. So it'll remove the branch. Whereas with `?` it probably won't backtrack and won't figure this out.

Someone taught me this surprising fact last year - sometimes a well-placed `assert!` can result in less code, rather than more.

Issue #14340, opened 2025-10-04
minifier: Conflate assignments with same value


overlookmotel:
Input:

```js
export let x, y;
export function setVars(value) {
  x = value;
  y = value;
}
```

Minified:

```js
export let x, y;
export function setVars(value) {
  x = value, y = value;
}
```

Could be further compressed to:

```js
export let x, y;
export function setVars(value) {
  y = x = value;
}
```

[Playground](https://playground.oxc.rs/#eNptkk2P1DAMhv9KlBNIcwAkJDSIE8tIK/ElDnuaizf1lCypU9nu7Iyq/nec0mSQ4NT4Te28fuzZB7/3eBkzq0uo7rJz1/dHOtKmnSYKGjM5QX0AlhdnSBO+dPORnLu4D26NLcO561/RciS/89nvZ88TlU+KpH5/giS486fMA6gi/6vc3zRlICkXTYmSEyh2dxgSMBRj0i6ff0ZFGSFgkwagPt3CkIeRUSxFebJYQh7tdguuw2NONQqnfktbdn60xovX2UgpktizRk3FWoSU8vMP1Inp26QSOzxsxGql8iLyGb8DW2pV17yHd/ekHK1guHlCM60xfGLOvKlmoeBbLdi5cfnjSYF7NLYe5c2r12/N1CR4h6dIeMj8MYHIIWLq2hM2WuQ4ICkkI5kNZHvKboeoTf2CCh0oNB//n8DsxfoYrRtkAoNYyYXcYY/rAlCZcOP7JF221dtgEGVda1UlYV//XavUuZX261Ct8zx+xjO2or8Qx68wYN2JNdUI53Qw2iXjjPyYpW5E6YeeMBi8+XZaTO5WfEVup6XoBXPtafkNTjYpag==)

Optimization is valid if:

1. Evaluating `value` can have no side effects.
2. Vars being assigned to must be bound, so getting/setting them can't have side effects.

---

Along the same lines:

Input:

```js
export const obj = {};
export function setProps(value) {
  obj.x = value;
  obj.y = value;
}
```

Minified:

```js
export const obj = {};
export function setProps(value) {
  obj.x = value, obj.y = value;
}
```

Could be further compressed to:

```js
export const obj = {};
export function setVars(value) {
  obj.y = obj.x = value;
}
```

Optimization is valid if:

1. Evaluating `value` can have no side effects.
[...]

PR #14637, opened 2025-10-15
perf(linter/plugins): lazy deserialize comments array

lilnasy:
Part of #14564.

Deserialize `program.comments` lazily upon first access, instead of eagerly, on assumption that most JS plugins won't access comments.

`deserializeWith` cleans up the buffer once deserializing AST is complete, so we keep a local copy of the buffer in scope of the `comments` getter, and restore it when the getter is triggered.

Buffers are reused, so the `if (localAstId !== astId)` check is required to make sure the AST in the buffer is same AST i.e. that the getter isn't accessed *after* the file has completed linting.

### DIFF
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -64,10 +64,7 @@ function deserializeProgram(pos) {
           uint8 =
           sourceText =
             void 0;
-        Object.defineProperty(this, 'comments', {
-          value: comments,
-          enumerable: true,
-        });
+        Object.defineProperty(this, 'comments', { value: comments });
         return comments;
       },
       start: 0,
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -49,9 +49,11 @@ function deserializeProgram(pos) {
       sourceType: deserializeModuleKind(pos + 125),
       hashbang: null,
       get comments() {
-        if (localAstId !== astId) {throw Error(
+        if (localAstId !== astId) {
+          throw Error(
             'The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.',
-          );}
+          );
+        }
         // Restore buffers
         uint32 = refUint32;
         uint8 = refUint8;
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -38,9 +38,9 @@ function deserializeWith(buffer, sourceTextInput, sourceByteLenInput, getLocInpu
 
 function deserializeProgram(pos) {
   let end = deserializeU32(pos + 4),
-    ref_uint32 = uint32,
-    ref_uint8 = uint8,
-    ref_sourceText = sourceText,
+    refUint32 = uint32,
+    refUint8 = uint8,
+    refSourceText = sourceText,
     localAstId = ++astId,
     program = parent = {
       __proto__: NodeProto,
@@ -52,18 +52,16 @@ function deserializeProgram(pos) {
         if (localAstId !== astId) {throw Error(
             'The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.',
           );}
-        // restore buffers
-        uint32 = ref_uint32;
-        uint8 = ref_uint8;
-        sourceText = ref_sourceText;
-        // deserialize the comments
-        let c = deserializeVecComment(pos + 24);
-        // drop the references
-        ref_uint32 = void 0;
-        ref_uint8 = void 0;
-        ref_sourceText = void 0;
-        Object.defineProperty(this, 'comments', { value: c });
-        return c;
+        // Restore buffers
+        uint32 = refUint32;
+        uint8 = refUint8;
+        sourceText = refSourceText;
+        // Deserialize the comments
+        let comments = deserializeVecComment(pos + 24);
+        // Drop the references
+        refUint32 = refUint8 = refSourceText = void 0;
+        Object.defineProperty(this, 'comments', { value: comments });
+        return comments;
       },
       start: 0,
       end,
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -49,11 +49,7 @@ function deserializeProgram(pos) {
       sourceType: deserializeModuleKind(pos + 125),
       hashbang: null,
       get comments() {
-        if (localAstId !== astId) {
-          throw Error(
-            'The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.',
-          );
-        }
+        if (localAstId !== astId) throw Error('Comments are only accessible while linting the file');
         // Restore buffers
         uint32 = refUint32;
         uint8 = refUint8;
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -1,7 +1,7 @@
 // Auto-generated code, DO NOT EDIT DIRECTLY!
 // To edit this generated file you have to edit `tasks/ast_tools/src/generators/raw_transfer.rs`.
 
-let uint8, uint32, float64, sourceText, sourceIsAscii, sourceByteLen, parent = null, getLoc;
+let uint8, uint32, float64, sourceText, sourceIsAscii, sourceByteLen, astId = 0, parent = null, getLoc;
 
 const textDecoder = new TextDecoder('utf-8', { ignoreBOM: true }),
   decodeStr = textDecoder.decode.bind(textDecoder),
@@ -38,13 +38,33 @@ function deserializeWith(buffer, sourceTextInput, sourceByteLenInput, getLocInpu
 
 function deserializeProgram(pos) {
   let end = deserializeU32(pos + 4),
+    ref_uint32 = uint32,
+    ref_uint8 = uint8,
+    ref_sourceText = sourceText,
+    localAstId = ++astId,
     program = parent = {
       __proto__: NodeProto,
       type: 'Program',
       body: null,
       sourceType: deserializeModuleKind(pos + 125),
       hashbang: null,
-      comments: deserializeVecComment(pos + 24),
+      get comments() {
+        if (localAstId !== astId) {throw Error(
+            'The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.',
+          );}
+        // restore buffers
+        uint32 = ref_uint32;
+        uint8 = ref_uint8;
+        sourceText = ref_sourceText;
+        // deserialize the comments
+        let c = deserializeVecComment(pos + 24);
+        // drop the references
+        ref_uint32 = void 0;
+        ref_uint8 = void 0;
+        ref_sourceText = void 0;
+        Object.defineProperty(this, 'comments', { value: c });
+        return c;
+      },
       start: 0,
       end,
       range: [0, end],
--- a/apps/oxlint/src-js/generated/deserialize.js
+++ b/apps/oxlint/src-js/generated/deserialize.js
@@ -57,8 +57,17 @@ function deserializeProgram(pos) {
         // Deserialize the comments
         let comments = deserializeVecComment(pos + 24);
         // Drop the references
-        refUint32 = refUint8 = refSourceText = void 0;
-        Object.defineProperty(this, 'comments', { value: comments });
+        refUint32 =
+          refUint8 =
+          refSourceText =
+          uint32 =
+          uint8 =
+          sourceText =
+            void 0;
+        Object.defineProperty(this, 'comments', {
+          value: comments,
+          enumerable: true,
+        });
         return comments;
       },
       start: 0,
--- a/crates/oxc_ast/src/serialize/mod.rs
+++ b/crates/oxc_ast/src/serialize/mod.rs
@@ -149,7 +149,7 @@ impl Program<'_> {
             // Deserialize the comments
             const comments = DESER[Vec<Comment>](POS_OFFSET.comments);
             // Drop the references
-            refUint32 = refUint8 = refSourceText = undefined;
+            refUint32 = refUint8 = refSourceText = uint32 = uint8 = sourceText = undefined;
             Object.defineProperty(this, 'comments', { value: comments, enumerable: true });
             return comments;
         },
--- a/crates/oxc_ast/src/serialize/mod.rs
+++ b/crates/oxc_ast/src/serialize/mod.rs
@@ -128,9 +128,9 @@ impl Program<'_> {
     /* IF LINTER */
     // Buffers will be cleaned up after the main program is deserialized.
     // We hold references here in case the comments need to be later accessed.
-    let ref_uint32 = uint32;
-    let ref_uint8 = uint8;
-    let ref_sourceText = sourceText;
+    let refUint32 = uint32;
+    let refUint8 = uint8;
+    let refSourceText = sourceText;
     const localAstId = ++astId;
     /* END_IF */
 
@@ -142,18 +142,16 @@ impl Program<'_> {
         /* IF LINTER */
         get comments() {
             if (localAstId !== astId) throw new Error(`The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.`);
-            // restore buffers
-            uint32 = ref_uint32;
-            uint8 = ref_uint8;
-            sourceText = ref_sourceText;
-            // deserialize the comments
-            const c = DESER[Vec<Comment>](POS_OFFSET.comments);
-            // drop the references
-            ref_uint32 = undefined;
-            ref_uint8 = undefined;
-            ref_sourceText = undefined;
-            Object.defineProperty(this, 'comments', { value: c });
-            return c;
+            // Restore buffers
+            uint32 = refUint32;
+            uint8 = refUint8;
+            sourceText = refSourceText;
+            // Deserialize the comments
+            const comments = DESER[Vec<Comment>](POS_OFFSET.comments);
+            // Drop the references
+            refUint32 = refUint8 = refSourceText = undefined;
+            Object.defineProperty(this, 'comments', { value: comments });
+            return comments;
         },
         /* END_IF */
         start,
--- a/crates/oxc_ast/src/serialize/mod.rs
+++ b/crates/oxc_ast/src/serialize/mod.rs
@@ -141,7 +141,7 @@ impl Program<'_> {
         hashbang: null,
         /* IF LINTER */
         get comments() {
-            if (localAstId !== astId) throw new Error(`The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.`);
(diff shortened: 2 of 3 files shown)


apps/oxlint/src-js/generated/deserialize.js
```diff
@@ -48,12 +48,37 @@ function deserializeProgram(pos) {
...
+        uint8 = unref_uint8;
+        sourceText = unref_sourceText;
+        // drop the references
+        ref_uint32 = void 0;
+        ref_uint8 = void 0;
+        ref_sourceText = void 0;
+        Object.defineProperty(this, 'comments', { value: c });
+        return c;
+      },
+    });
```
overlookmotel: Whatever we do, lazy deser is somewhat dangerous. Nothing to stop user storing `Program` object somewhere, and accessing `comments` property on it much later on, long after linting this file has finished.

One thing I didn't mention before is that buffers are reused over and over for different files (see [here](https://github.com/lilnasy/oxc/blob/cd62cfb7bdd2f3f27987d323ed43901e67f12634/apps/oxlint/src/js_plugins/external_linter.rs#L114-L190) and [here](https://github.com/lilnasy/oxc/blob/cd62cfb7bdd2f3f27987d323ed43901e67f12634/apps/oxlint/src-js/plugins/lint.ts#L62-L80)).

So even if you have the original buffer (cached in `ref_uint32` etc), the *contents* of the buffer may have changed in meantime - it's got a different AST in it now. So `pos + 24` may now point to the middle of a `Function`,  or an `Expression`, or whatever.

So then accessing `program.comments` would produce nonsense, or in worst case might result in trying to create an array with some huge length, resulting in out-of-memory crash.

Maybe one solution is:

* Have a global counter `astId`.
* Each time you call `deserializeProgramOnly`, it increments the counter.
* Make a local copy of the current counter value in `deserializeProgram`.
* In the getter, check that `localAstId === astId`. If not, throw a sensible error, instead of producing nonsense.

lilnasy: Regarding safety, does oxlint ensure that all rules run for one file before the second one is processed?

lilnasy: That sounds like a great solution but I want to make sure I am benchmarking this right. The improvements I see are within the margin of error when running on the vscode repo.
```
hyperfine --warmup 1 --runs 10 --ignore-failure 'node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json' 'node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json'
Benchmark 1: node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json
  Time (mean ± σ):     644.3 ms ±   2.6 ms    [User: 2477.7 ms, System: 244.7 ms]
  Range (min … max):   640.9 ms … 649.1 ms    10 runs
 
Benchmark 2: node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
  Time (mean ± σ):     657.5 ms ±  14.7 ms    [User: 2488.6 ms, System: 245.4 ms]
  Range (min … max):   644.5 ms … 695.8 ms    10 runs
 
Summary
  node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json ran
    1.02 ± 0.02 times faster than node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
```

<details>
<summary>

`oxlintrc.json`

</summary>

```json
{
	"jsPlugins": [
		"./plugin.ts"
	],
	"rules": {
		"test/noop": "error"
	}
}
```

</details>

<details>
<summary>

`plugin.ts`

</summary>

```ts
import { definePlugin } from '../oxc/apps/oxlint/dist-strict/index.js';
import type { Rule } from '../oxc/apps/oxlint/dist-strict/index.js';

const noopRule: Rule = {
  create(context) {
    context.sourceCode;
    return {};
  },
};

export default definePlugin({
  meta: {
    name: 'test',
  },
  rules: {
    'noop': noopRule,
  },
});
```

</details>

```
Found 7141 warnings and 19 errors.
Finished in 1.2s on 6220 files with 89 rules using 10 threads.
```

overlookmotel: > Regarding safety, does oxlint ensure that all rules run for one file before the second one is processed?

Yes. Files are processed strictly in series. The whole process of linting each file is entirely synchronous (on JS side at least).

overlookmotel: > I want to make sure I am benchmarking this right.

I'd suggest using the plugin shown here: https://oxc.rs/blog/2025-10-09-oxlint-js-plugins.html#quick-start

Your no-op plugin is not the best for measuring this because there's an optimization where deserializing the AST is skipped entirely if `create` returns an empty visitor (there's no need to deserialize AST, because the visitor won't touch the AST anyway).

The code is hard to follow because I've avoided nice encapsulated classes and the like in favour of nasty code that runs fast, but that optimization is here (`needsVisit` is `false` if visitor object has no properties):

https://github.com/oxc-project/oxc/blob/85a2743fbbedc6bc17a96e13fe02af303dfd5cfc/apps/oxlint/src-js/plugins/lint.ts#L128-L136

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -129,13 +129,43 @@ impl Program<'_> {
         body: null,
         sourceType: DESER[ModuleKind](POS_OFFSET.source_type.module_kind),
         hashbang: null,
-        ...(COMMENTS && { comments: DESER[Vec<Comment>](POS_OFFSET.comments) }),
         start,
         end,
         ...(RANGE && { range: [start, end] }),
         ...(PARENT && { parent: null }),
     };
```
overlookmotel, line 161: #14681 added another way to feature-gate code. So now the `comments` getter can be defined inline in the object:

```js
const program = parent = {
    type: 'Program',
    body: null,
    sourceType: DESER[ModuleKind](POS_OFFSET.source_type.module_kind),
    hashbang: null,
    /* IF COMMENTS */
    get comments() {
        // ...
    },
    /* END_IF */
    start,
    end,
    ...(RANGE && { range: [start, end] }),
    ...(PARENT && { parent: null }),
};
```

apps/oxlint/src-js/generated/deserialize.js
lilnasy: Thanks, I think I have this down. I'm not seeing an improvement e2e. `hyperfine`, at least on my mac, is biased towards whichever one it runs first.

<details>
<summary>Strict first</summary>

```bash
 arsh@mac  ~/Documents/GitHub/vscode  (main) 
> hyperfine --warmup 1 --runs 5 --ignore-failure 'node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json' 'node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json'
Benchmark 1: node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.483 s ±  0.004 s    [User: 4.494 s, System: 0.221 s]
  Range (min … max):    2.480 s …  2.490 s    5 runs
 
Benchmark 2: node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.500 s ±  0.023 s    [User: 4.510 s, System: 0.238 s]
  Range (min … max):    2.472 s …  2.528 s    5 runs
 
Summary
  node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json ran
    1.01 ± 0.01 times faster than node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json
 ```
 
 </details>
 
 <details>
 <summary>Lazy first</summary>

 ```bash
  arsh@mac  ~/Documents/GitHub/vscode  (main) 
> hyperfine --warmup 1 --runs 5 --ignore-failure 'node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json' 'node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json' 
Benchmark 1: node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.482 s ±  0.014 s    [User: 4.516 s, System: 0.215 s]
  Range (min … max):    2.472 s …  2.504 s    5 runs
 
Benchmark 2: node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.492 s ±  0.012 s    [User: 4.492 s, System: 0.211 s]
  Range (min … max):    2.481 s …  2.512 s    5 runs
 
Summary
  node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json ran
    1.00 ± 0.01 times faster than node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
```

</details>

crates/oxc_ast/src/serialize/mod.rs
lilnasy: I noticed you also added a linter variant. Since the comments variant is for the linter, are they effectively the same variant?

lilnasy:
Not sure what's up with AST check CI. The generated code is up to date. CI's expected output is just different.

crates/oxc_ast/src/serialize/mod.rs
overlookmotel: Yes. But having more flags doesn't cost anything at runtime, so I figured better to keep them separate. `COMMENTS === LINTER` always at present, but we might have a use case for separating them later on.

overlookmotel:
> Not sure what's up with AST check CI. The generated code is up to date. CI's expected output is just different.

Looks like `dprint` (which we use for formatting JS/TS code) requires 2 passes to get to a stable state with this code. That's a bug in `dprint`, but #14723 provides a workaround. I'll merge it if we decide to go forwards with this PR.

The problem on this PR was that the autofix CI job ran `dprint` and pushed a commit with the changes. But then the "AST Changes" job checks the code committed is identical to the output of `cargo run -p oxc_ast_tools`, which it's not because of that extra commit. 🔄

---

Or just sidestep the formatting problem, by shortening the error message:

```js
if (localAstId !== astId) throw new Error('Comments are only accessible while linting the file');
```

This is an edge case which shouldn't be hit in practice, so I don't think we need to provide too much explanation.

apps/oxlint/src-js/generated/deserialize.js
overlookmotel: Oh hot damn. Maybe the impact of lazy deserializing comments is too small to be measurable in the context of the whole process.

OK, let's do one more test. Please add this to `oxlintrc.json`:

```
"categories": { "correctness": "off" },
```

That'll skip all Oxlint's built-in rules.

And make the plugin:

```js
const noopRule: Rule = {
  create(context) {
    // Trigger getter to deserialize AST
    const _ast = context.ast;
    // Return empty visitor to skip walking AST
    return {};
  },
};
```

Both those should reduce the overall workload, and (hopefully) reveal the gain/lack thereof of lazy deser more clearly.

lilnasy: Had to do that when there was a panic. It did seem to show a benefit but only by favoring the lazy implementation **more often than not**. I'll run again to get the exact results.

lilnasy:
I tried that as well, dprint introduces a block in that case.

overlookmotel:
Weird. I'm not seeing that locally. Anyway, it doesn't matter if it produces a block, only that it does so *consistently*.

apps/oxlint/src-js/generated/deserialize.js
lilnasy: Ran 10 times, alternating which benchmark runs first. Lazy was faster 7/10 times.
```bash
 arsh@mac  ~/Documents/GitHub/vscode  (main) 
> hyperfine --warmup 1 --runs 5 --ignore-failure 'node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json' 'node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json'
Benchmark 1: node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.167 s ±  0.037 s    [User: 3.220 s, System: 0.203 s]
  Range (min … max):    2.139 s …  2.231 s    5 runs
 
Benchmark 2: node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json
  Time (mean ± σ):      2.125 s ±  0.010 s    [User: 3.178 s, System: 0.195 s]
  Range (min … max):    2.111 s …  2.137 s    5 runs
 
Summary
  node ../oxc/apps/oxlint/dist-lazy/cli.js -c ./oxlintrc.json ran
    1.02 ± 0.02 times faster than node ../oxc/apps/oxlint/dist-strict/cli.js -c ./oxlintrc.json
 ```

<details>
<summary>

`oxlintrc.json`

</summary>

 ```json
 {
	"jsPlugins": [
		"./plugin.ts"
	],
	"categories": {
		"correctness": "off"
	},
	"rules": {
		"best-plugin-ever/no-debugger": "error"
	}
}
 ```
 
 </details>
 
 
 <details>
<summary>

`plugin.ts`

</summary>

 ```typescript
// The simplest rule of all - no debugger
const rule = {
	create(context) {
		return {
			DebuggerStatement(node) {
				context.report({
					message: "No debugger!",
					node,
				});
			},
		};
	},
};

const plugin = {
	meta: {
		name: "best-plugin-ever",
	},
	rules: {
		"no-debugger": rule,
	},
};

export default plugin;

 ```
 
 </details>

Comment (overlookmotel) [reply apps/oxlint/src-js/generated/deserialize.js]: Thanks for testing it again. I think overall this probably *is* a gain, and the penalty when it's not is small - so let's do it. I'll make a few more comments on code style, and then let's get this merged.

Thanks very much for your patience!

File: crates/oxc_ast/src/serialize/mod.rs (hunk 2 above)
Comment (overlookmotel) [reply crates/oxc_ast/src/serialize/mod.rs:161]: Would you mind reintroducing the `COMMENTS` flag? I see your point that it's redundant at present, but IMO it's clearer and more future-proof to keep it.

File: crates/oxc_ast/src/serialize/mod.rs [hunk 3]
```diff
@@ -124,12 +124,38 @@ impl Program<'_> {
     const start = IS_TS ? 0 : DESER[u32](POS_OFFSET.span.start),
         end = DESER[u32](POS_OFFSET.span.end);
 
+
+    /* IF LINTER */
+    // Buffers will be cleaned up after the main program is deserialized.
+    // We hold references here in case the comments need to be later accessed.
+    let ref_uint32 = uint32;
+    let ref_uint8 = uint8;
+    let ref_sourceText = sourceText;
```
overlookmotel: Style nit: Our convention is snake_case in Rust code, but camelCase in JS. Maybe `uint32Ref`, `uint32Copy`, or `localUint32`?

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -124,12 +124,38 @@ impl Program<'_> {
...
     const program = parent = {
         type: 'Program',
         body: null,
         sourceType: DESER[ModuleKind](POS_OFFSET.source_type.module_kind),
         hashbang: null,
-        ...(COMMENTS && { comments: DESER[Vec<Comment>](POS_OFFSET.comments) }),
+        /* IF LINTER */
+        get comments() {
+            if (localAstId !== astId) throw new Error(`The AST being accessed has already been cleaned up. Please ensure that the plugin works synchronously.`);
+            // restore buffers
```
overlookmotel: Totally unimportant nit!

```suggestion
            // Restore buffers
```

(ditto the other comments)

crates/oxc_ast/src/serialize/mod.rs
lilnasy: I pushed it right as you sent this message: [e2c1242](https://github.com/oxc-project/oxc/pull/14637/commits/e2c12424ba8832ac1a46c70c266ab4bef0b2843c).

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -124,12 +124,38 @@ impl Program<'_> {
...
+            uint32 = ref_uint32;
+            uint8 = ref_uint8;
+            sourceText = ref_sourceText;
+            // deserialize the comments
+            const c = DESER[Vec<Comment>](POS_OFFSET.comments);
```
overlookmotel: Please can we call this var `comments`? Once we have everything stable, we'll run all the code through full minification, so var names will get shortened in the runtime code anyway. We're only not doing that at present to make debugging without sourcemaps possible while we're developing.

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -124,12 +124,38 @@ impl Program<'_> {
...
+            // drop the references
+            ref_uint32 = undefined;
+            ref_uint8 = undefined;
+            ref_sourceText = undefined;
```
overlookmotel: Minifier should be able to do this automatically (#14340), but currently it doesn't.

```suggestion
            ref_uint32 = ref_uint8 = ref_sourceText = undefined;
```

overlookmotel (commented):
All nits...

lilnasy:
Addressed the review comments in [09baebc](https://github.com/oxc-project/oxc/pull/14637/commits/09baebc0e1c69153e3295d3d574cb7fcdf502580).

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -124,12 +124,36 @@ impl Program<'_> {
...
+            if (localAstId !== astId) throw new Error('Comments are only accessible while linting the file');
+            // Restore buffers
+            uint32 = refUint32;
+            uint8 = refUint8;
+            sourceText = refSourceText;
+            // Deserialize the comments
+            const comments = DESER[Vec<Comment>](POS_OFFSET.comments);
+            // Drop the references
+            refUint32 = refUint8 = refSourceText = undefined;
+            Object.defineProperty(this, 'comments', { value: comments });
```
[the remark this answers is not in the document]
overlookmotel: Incorrect. The property already has `enumerable` flag set from it's initial definition. `Object.defineProperty` doesn't change that.

crates/oxc_ast/src/serialize/mod.rs
```diff
@@ -124,12 +124,36 @@ impl Program<'_> {
...
+        get comments() {
+            if (localAstId !== astId) throw new Error('Comments are only accessible while linting the file');
+            // Restore buffers
+            uint32 = refUint32;
+            uint8 = refUint8;
+            sourceText = refSourceText;
+            // Deserialize the comments
+            const comments = DESER[Vec<Comment>](POS_OFFSET.comments);
+            // Drop the references
+            refUint32 = refUint8 = refSourceText = undefined;
```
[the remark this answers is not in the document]
overlookmotel: Oh! He right!

overlookmotel:
Thank you. I did warn you it might be tricky, but I underestimated!

lilnasy:
Thanks for guiding me through this. I did not imagine a project of oxc's scope would neat and tidy, and I enjoyed this welcome tour!

PR #14652, opened 2025-10-16
fix(semantic): add scope tracking for `with` statements

### LINKED ISSUE #8365: semantics: scopes for `with` statements are not created (Object Environment Record)
Currently semantic does not track scopes of `with` statements.
`is_global_reference` may return `true` even if that reference refers to a property of an object passed to a `with` statement.
https://github.com/oxc-project/oxc/blob/e4d66e46360dbdec289e40319f8b2a335b25c099/crates/oxc_semantic/src/symbol.rs#L342-L346

This leads to minification errors for codes like:
```js
const foo = {
  Object: class {
    constructor() { console.log('Constructor') }
  }
}
with (foo) {
    console.log(new Object()) // should print "Constructor"
}
```
[playground](https://playground.oxc.rs/#eNpFj81qwzAQhF9F7CU2hNAWejH0VMi1PfTYi6KuHQVZa3aVuMH43bPyTwzG9mpmvlkN4KACR1GSqYnMhxl+ozFfpwu6VBkXrMh8ZMzk4qtLxEVphmmmgIdATbH73MRdacac0Jc+vU9nUyhbIxtnzUXsl7KiLCc/7IGgGoCvMX/kHpP9h0rZuIfgY4KqtkF0EEcdrorc2xOFdUpso9TE7WIe99BZFuQJSVd2+HPPYYWw75KWzqdHHzDaNisJJR0uolLHqNEbflvGKBsybzMh9f/ZOHckyw3qqoDy9vL6Dupw9IcN5kvp0Proaz97HbW5QcHz8v3Za3dnnW6xXLW1sQnrOLFiYgrHQH0m3JBPJE99HB89zqRs)

This minification error may be fine as the minifier does not support `with` statements (for now?). But in this case, I think a warning should be output if the minifier found a `with` statement.
https://github.com/oxc-project/oxc/blob/e4d66e46360dbdec289e40319f8b2a335b25c099/tasks/coverage/src/tools/minifier.rs#L44

Boshen:
Fixes #8365

## Summary

The `with` statement was not creating a scope in the semantic analysis, causing `is_global_reference` to incorrectly return `true` for references inside `with` blocks. This could lead to incorrect minification.

## Problem

According to the ECMAScript specification, `with` statements should create an Object Environment Record that extends the lexical environment. Without proper scope tracking:

- `is_global_reference()` incorrectly returns `true` for identifiers inside `with` blocks
- This can cause minification to incorrectly treat local references as global
- The scope tree doesn't accurately represent the program structure

### Changes Made

1. **AST Definition** (`crates/oxc_ast/src/ast/js.rs`):
   - Added `#[scope]` attribute to `WithStatement` struct
   - Added `scope_id: Cell<Option<ScopeId>>` field

2. **Semantic Builder** (`crates/oxc_semantic/src/builder.rs`):
   - Modified `visit_with_statement` to call `enter_scope()` before visiting the body
   - Added `leave_scope()` call after visiting the body
   - Uses `ScopeFlags::empty()` similar to block statements

3. **Test Coverage** (`crates/oxc_semantic/tests/integration/scopes.rs`):
   - Added `test_with_statement` to verify scope creation
   - Tests both simple expressions and block statement bodies
   - Verifies proper scope tree structure

### Example

```javascript
const foo = {
  Object: class {
    constructor() { console.log('Constructor') }
  }
}
with (foo) {
  console.log(new Object()) // should print "Constructor"
}
```

Before this fix, `Object` inside the `with` block would be incorrectly identified as a global reference, potentially causing minification errors.

## Implementation Notes

While this implementation doesn't fully model the complex object environment semantics (where property lookups can dynamically affect identifier resolution at runtime), it ensures:

- References inside `with` blocks are no longer incorrectly marked as global
- The scope tree properly represents the lexical structure
- Provides a foundation for future improvements to object environment handling

## Testing

All existing tests pass, plus the new test verifies:
- `with` statements create scopes
- Scopes are properly nested in the scope tree
- Works correctly with both simple and block statement bodies

### DIFF
--- a/crates/oxc_ast/src/ast/js.rs
+++ b/crates/oxc_ast/src/ast/js.rs
@@ -1384,12 +1384,15 @@ pub struct ReturnStatement<'a> {
 
 /// With Statement
 #[ast(visit)]
+#[scope]
 #[derive(Debug)]
 #[generate_derive(CloneIn, Dummy, TakeIn, GetSpan, GetSpanMut, ContentEq, ESTree)]
 pub struct WithStatement<'a> {
     pub span: Span,
     pub object: Expression<'a>,
+    #[scope(enter_before)]
     pub body: Statement<'a>,
+    pub scope_id: Cell<Option<ScopeId>>,
 }
 
 /// Switch Statement
--- a/crates/oxc_ast/src/generated/assert_layouts.rs
+++ b/crates/oxc_ast/src/generated/assert_layouts.rs
@@ -461,12 +461,13 @@ const _: () = {
     assert!(offset_of!(ReturnStatement, span) == 0);
     assert!(offset_of!(ReturnStatement, argument) == 8);
 
-    // Padding: 0 bytes
-    assert!(size_of::<WithStatement>() == 40);
+    // Padding: 4 bytes
+    assert!(size_of::<WithStatement>() == 48);
     assert!(align_of::<WithStatement>() == 8);
     assert!(offset_of!(WithStatement, span) == 0);
     assert!(offset_of!(WithStatement, object) == 8);
     assert!(offset_of!(WithStatement, body) == 24);
+    assert!(offset_of!(WithStatement, scope_id) == 40);
 
     // Padding: 4 bytes
     assert!(size_of::<SwitchStatement>() == 56);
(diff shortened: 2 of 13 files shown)


overlookmotel (commented):
I think the scope should not cover `object` - scope should be entered just before `body`. This is similar to `SwitchStatement`.

Boshen:
### Merge activity

* **Oct 16, 10:01 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Oct 16, 10:01 AM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).
* **Oct 16, 10:07 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.dev/merges?org=oxc-project&repo=oxc).

PR #14715, opened 2025-10-17
feat(linter/plugins): comment-related APIs

lilnasy:
Part of #14564. Implement the remaining `SourceCode` APIs related to comments (`getCommentsBefore`, `getCommentsAfter`, `getCommentsInside`, `commentsExistBetween`).


apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -468,9 +505,16 @@ export const SOURCE_CODE = Object.freeze({
...
    */
-  // oxlint-disable-next-line no-unused-vars
   commentsExistBetween(nodeOrToken1: NodeOrToken, nodeOrToken2: NodeOrToken): boolean {
-    throw new Error('`sourceCode.commentsExistBetween` not implemented yet'); // TODO
+    // find the first comment after nodeOrToken1 end, check if it ends before nodeOrToken2 starts
+    const nodeOrToken1End = nodeOrToken1.range[1];
+    for (let i = 0; i < ast.comments.length; i++) {
+      const c = ast.comments[i];
+      if (c.range[0] >= nodeOrToken1End) {
+        return c.range[1] <= nodeOrToken2.range[0];
```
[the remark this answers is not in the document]
lilnasy: no

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
...
 
   /**
    * Get all comment tokens directly after the given node or token.
    * @param nodeOrToken - The AST node or token to check for adjacent comment tokens.
    * @returns Array of `Comment`s in occurrence order.
    */
-  // oxlint-disable-next-line no-unused-vars
   getCommentsAfter(nodeOrToken: NodeOrToken): Comment[] {
-    throw new Error('`sourceCode.getCommentsAfter` not implemented yet'); // TODO
+    if (ast === null) initAst();
```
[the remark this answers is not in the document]
lilnasy: right, switching to draft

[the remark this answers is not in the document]
overlookmotel: Don't take AI's reviews too seriously. Sometimes he catches things, but a lot of the time he talks bollocks.

In this case, the JSDoc comments were copied from ESLint. The word "directly" is odd (unnecessary), so by all means remove it if you like, but also feel free not to worry about it.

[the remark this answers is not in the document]
lilnasy: That's absolutely true. It just made me realise that I don't actually know the expected behavior. It's a draft until I take a look at eslint code.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
...
    */
-  // oxlint-disable-next-line no-unused-vars
   getCommentsBefore(nodeOrToken: NodeOrToken): Comment[] {
-    throw new Error('`sourceCode.getCommentsBefore` not implemented yet'); // TODO
+    if (ast === null) initAst();
+    const { comments } = ast;
+    if (comments.length === 0) return [];
+    const targetStart = nodeOrToken.range[0];
+    let i = comments.length - 1;
+    // linear backwards search for the index of the last comment before the provided node or token, TODO(perf): binary
```
overlookmotel: Just a nit. Try to keep comments short where you can, and capitalize (I'm a capitalist!)

```suggestion
    // Linear backwards search for last comment before `nodeOrToken`.
    // TODO: Use binary search.
```

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
    * @param nodeOrToken - The AST node or token to check for adjacent comment tokens.
    * @returns Array of `Comment`s in occurrence order.
    */
-  // oxlint-disable-next-line no-unused-vars
   getCommentsBefore(nodeOrToken: NodeOrToken): Comment[] {
-    throw new Error('`sourceCode.getCommentsBefore` not implemented yet'); // TODO
+    if (ast === null) initAst();
+    const { comments } = ast;
+    if (comments.length === 0) return [];
```
overlookmotel: I *think* can remove the line `if (comments.length === 0) return [];`.

If `comments.length === 0` then `i = -1`, `while` loop does nothing, and end with `return comments.slice(0, 0);` (empty array).

That may sound like more work, but it's less branches. And most files have *some* comments in them.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
...
+    const targetStart = nodeOrToken.range[0];
+    let i = comments.length - 1;
+    // linear backwards search for the index of the last comment before the provided node or token, TODO(perf): binary
+    while (i >= 0 && comments[i].range[1] > targetStart) {
```
overlookmotel: Can use `comments[i].end` here. `comments` is not user-provided, so we know the object shape for sure. Accessing `comment.end` should be cheaper than `comment.range[1]`, and it's more readable.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
...
-  // oxlint-disable-next-line no-unused-vars
   getCommentsAfter(nodeOrToken: NodeOrToken): Comment[] {
-    throw new Error('`sourceCode.getCommentsAfter` not implemented yet'); // TODO
+    if (ast === null) initAst();
+    const { comments } = ast;
+    if (comments.length === 0) return [];
+    const targetEnd = nodeOrToken.range[1];
+    let i = 0;
+    // linear forwards search for the index of the first comment after the provided node or token, TODO(perf): binary
+    while (i < comments.length && comments[i].range[0] < targetEnd) {
```
overlookmotel:
```suggestion
    const len = comments.length;
    while (i < len && comments[i].range[0] < targetEnd) {
```

...to avoid `comments.length` property lookup on each turn of the loop.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -169,29 +169,66 @@ export const SOURCE_CODE = Object.freeze({
...
+    if (indexStart === null) return [];
+    // linear search for the last comment inside the node's range, TODO(perf): binary
+    for (let i = indexStart; i < comments.length; i++) {
+      const c = comments[i];
+      if (c.range[1] > rangeEnd) {
+        indexEnd = i;
+        break;
+      }
+    }
+    return comments.slice(indexStart, indexEnd ?? comments.length);
```
overlookmotel: I think can be simplified.

* 1st loop find first comment which starts after `node.range[0]` (ignore whether it's before `node.range[1]`) => `indexStart`.
* 2nd loop search from (and including) that comment for first comment which starts after `node.range[1]` => `indexEnd`.
* Return `comments.slice(indexStart, indexEnd)`. If no comments inside the node, `indexStart` will equal `indexEnd`, and it'll return an empty array.

At least I *think* so...

apps/oxlint/test/fixtures/comments/plugin.ts
```diff
@@ -0,0 +1,64 @@
+import { definePlugin } from '../../../dist/index.js';
+import type { Rule } from '../../../dist/index.js';
+
+const testCommentsRule: Rule = {
```
overlookmotel, line 16: It'd be ideal to make sure tests cover all cases:

* `getCommentsBefore` a node which has no comments before it.
* `getCommentsAfter` a node which has no comments after it.
* `getCommentsInside` a node which has no comments inside it.
* `commentsExistBetween` cases which return both `true` and `false`.

Might want to also add a file to `files` which has no comments at all, and make sure all the APIs still work.

overlookmotel (commented):
Apologies that this review is premature - I know you're still working on it. But I was interested and had a look, so I thought I'd give my thoughts.

Background: In linter JS plugins, the JS-side code is going to be the slowest code in the whole path, so I'm keen to get it as tight and performant as it possibly can be. In most projects, readability and suchlike is also important, and it's normal to be a little loose, but here I think we should prioritise perf over all else.

(e.g. you'll see in the rest of the JS plugins code we never use `if (!foo)`, we'll use the cheaper `if (foo === false)` or `if (foo !== null)` instead, depending on the type of `foo`)

That's the reason for my rather fanatical nit-picking!

Note: I think it's fine to leave binary search to a later PR, but I still feel it's helpful to tighten the algorithms as much as possible at this earlier stage.

overlookmotel:
Great! Reviewing now (rebasing on latest main first).

overlookmotel:
@lilnasy I'm really sorry, I've made a massive screw-up. While rebasing on main, it seems my local git client had cached the *previous* version (before you added the whitespace-checking logic). Now I've force-pushed and destroyed all that work.

Could you please force push the most recent version from your local copy, to restore the most recent version?

I'm really sorry. Totally stupid of me!

overlookmotel:
OK, I think I've recovered the commits I over-wrote, and force pushed them back onto the PR branch. But can you please check I didn't lose anything in the process? Really sorry I've made a mess of this.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -170,29 +171,115 @@ export const SOURCE_CODE = Object.freeze({
...
+      const comment = comments[i];
+      const commentStart = comment.start;
+
+      if (commentStart < targetEnd) {
+        continue;
+      }
+      const gap = sourceText.slice(targetEnd, commentStart);
+      const whitespaceOnlyGap = whitespacePattern.test(gap);
+      if (whitespaceOnlyGap) {
+        commentsAfter.push(comment);
```
lilnasy, line 242: I tested and found `.slice()` to be faster than iteration and `.push()` in microbenchmarks. I'll have to do more testing, but I figured I'd mention that the `.push()` is what allows this method to only have a single for-loop, instead of two. Should we measure on a react codebase with `plugin-hooks` before making these decisions?

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -170,29 +171,115 @@ export const SOURCE_CODE = Object.freeze({
...
+    // TODO: Use binary search.
+    for (let i = 0; i < commentsLength; i++) {
+      const comment = comments[i];
+      if (comment.start >= rangeStart) {
+        indexStart = i;
+        break;
+      }
+    }
+
+    // Continued linear search for last comment within `node`'s range.
```
lilnasy:
```suggestion
    // Continued linear search for first comment outside `node`'s range.
    // Its index is used as `sliceEnd`, which is exclusive of the slice.
```

apps/oxlint/test/fixtures/comments/output.snap.md
lilnasy, line 1: This output was manually compared against eslint's.

apps/oxlint/test/fixtures/comments/plugin.ts
lilnasy: - [x] `getCommentsBefore` a node which has no comments before it.
- [x] `getCommentsAfter` a node which has no comments after it.
- [x] `getCommentsInside` a node which has no comments inside it.
- [x] `commentsExistBetween` cases which return both true and false.
- [x] run the plugin against a file with no comments.

overlookmotel:
I'm still going through it. Have pushed a few commits - all just nits, except the last one.

overlookmotel (approved):
Nice one! I had completely misunderstood what these methods are meant to do, because I hadn't looked at ESLint's implementation. Thank you for figuring out the correct logic.

I've pushed a bunch of commits, but they're all just style nits, perf micro-micro-optimizations, and increasing the test coverage. The logic is unaltered, and all seems correct.

I've not checked the output against ESLint, but I have been through the snapshot and checked it's correct according to my new understanding of what these methods are meant to do.

The only one of my commits which actually changed behavior is this one: 1aadc6d78c8e1e94f692f175b0a68427977f1614

Going to merge now, but please come back if you disagree with any of my changes.

apps/oxlint/src-js/plugins/source_code.ts
overlookmotel: Thanks for checking. Interesting!

I think it's fine to leave it as is for now. The larger perf improvement will be from switching from linear to binary search, and that may alter how this needs to work anyway.

apps/oxlint/test/fixtures/comments/plugin.ts
overlookmotel: I think the tests now cover all these.

apps/oxlint/src-js/plugins/source_code.ts
overlookmotel: Thanks for making that change. I've reverted to using `range` for `nodeOrToken`, as that's user-supplied and the shape of the object they pass in is unknown. I think ESLint tends to use `range`.

PR #14738, opened 2025-10-17
feat(linter): merge vitest/prefer-to-be rule into existing jest/prefer-to-be rule implementation

taearls:
It seems that most of the functionality from vitest's version of the prefer-to-be rule were already present in the Jest implementation, except when comparing floats. I added support for this and a unit test for it. 

I updated the diagnostic messaging to be more consistent: 

1. there were some missing start brackets `[` when giving help messages for 
2. the snippet from the source code in the help text did not match what was underlined
3. `not` was omitted from the help messaging when it was in a computed expression

I added prefer-to-be to the list of `VITEST_COMPATIBLE_RULES` too.


taearls:
converting to a draft for now. I'm noticing that some of the snapshots need to be fixed

camc314 (approved):
Thank you

PR #14768, opened 2025-10-19
fix(linter/plugins): handle utf16 characters within comment spans

lilnasy:
Part of #14564. Corrects start and end offsets to accommodate two byte characters. Conversion of UTF-8 indices to UTF-16 takes place on the rust side for performance.

### DIFF
--- a/apps/oxlint/test/e2e.test.ts
+++ b/apps/oxlint/test/e2e.test.ts
@@ -220,4 +220,8 @@ describe('oxlint CLI', () => {
   it('should support comments-related APIs in `context.sourceCode`', async () => {
     await testFixture('comments');
   });
+
+  it('should support UTF16 characters in source code and comments with correct spans', async () => {
+    await testFixture('unicode-comments');
+  });
 });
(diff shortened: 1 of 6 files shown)


overlookmotel (approved):
Great stuff!

All looks correct, so merging. But I'd suggest as follow-up to expand the tests:

* Add `start` and `end` of each comment to snaphot.
* Add `range` to snapshot (or assert `comment.range[0] === comment.start && comment.range[1] === comment.end` for all comments).
* Add `loc` to snapshot for each comment, and eyeball that they look correct.

It might also be good to call `context.report()` for each comment individually, passing the `Comment` as `node` property. That'd:

1. Test the translation of offsets back to UTF-8 when reporting errors (that happens on Rust side).
2. Make the snapshot more readable and easier to eyeball for errors.

lilnasy:
@overlookmotel I left out testing `start` and `end` explicitly for brevity. The comments' value being correct implies they were correct when used to slice `sourceText`. I can add the assertions and loc right away!

Issue #14743, opened 2025-10-18
linter: various rules are missing `options` documentation on the Rules pages of the website


connorshea:
I think this was covered slightly by #6050, but may have been missed as being within scope of the problem.

`no-unused-vars`, for example, has [various options available](https://github.com/oxc-project/oxc/blob/4283fd8eb922e4ea23cf9045e57696095d31d6b0/crates/oxc_linter/src/rules/eslint/no_unused_vars/options.rs). But the [page for the rule](https://oxc.rs/docs/guide/usage/linter/rules/eslint/no-unused-vars.html) makes no mention of them.

The same is true with [`max-classes-per-file`](https://github.com/oxc-project/oxc/blob/6440cde67502bdc72020a1b8992b5445b7b490f3/crates/oxc_linter/src/rules/eslint/max_classes_per_file.rs), which has options for `ignoreExpressions` and `max`, but no mention of that [in the documentation](https://oxc.rs/docs/guide/usage/linter/rules/eslint/max-classes-per-file.html#eslint-max-classes-per-file).

It might be good to enforce this somehow via a check, but I'm not really familiar enough with rust to say how.

Here's a script to get all rules without the autogenerated config set up, but that do have configuration options:

<details>
<summary>Script for finding lint rules with options but no autogen config</summary>

```js
#!/usr/bin/env node
// Find eslint rule files that have from_configuration
// but no "config =" in declare_oxc_lint!

import fs from 'fs';
import path from 'path';
import { fileURLToPath } from 'url';

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);

const RULES_DIR = path.join(__dirname, 'crates/oxc_linter/src/rules');

function scanDirectory(dir) {
  const files = fs.readdirSync(dir, { withFileTypes: true });
  const results = [];

  for (const file of files) {
    const fullPath = path.join(dir, file.name);

    if (file.isDirectory()) {
      // Recursively scan subdirectories
      results.push(...scanDirectory(fullPath));
    } else if (file.isFile() && file.name.endsWith('.rs')) {
      const content = fs.readFileSync(fullPath, 'utf8');

      const hasDeclareLint = content.includes('declare_oxc_lint!(');
      const hasFromConfig = content.includes('fn from_configuration(');
      const hasConfigParam = (/^\s*config\s*=\s/m).test(content);

      if (hasDeclareLint && hasFromConfig && !hasConfigParam) {
[...]

PR #14789, opened 2025-10-19
docs(linter): Improve the Options documentation for no-empty-function rule

connorshea:
Previously this wasn't following the same format as most other rules with options, now it is.

Related to #14743

I tried converting this to auto-gen docs, but I have no idea how to get that working without the compiler complaining, honestly.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_empty_function.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_empty_function.rs
@@ -117,10 +117,12 @@ declare_oxc_lint!(
     ///
     /// Example:
     /// ```json
-    /// "no-empty-function": [
-    ///   "error",
-    ///   { "allow": ["functions"] }
-    /// ]
+    /// {
+    ///   "no-empty-function": [
+    ///     "error",
+    ///     { "allow": ["functions"] }
+    ///   ]
+    /// }
     /// ```
     ///
     /// `allow` accepts the following values:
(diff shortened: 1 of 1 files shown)


crates/oxc_linter/src/rules/eslint/no_empty_function.rs
```diff
@@ -106,16 +106,21 @@ declare_oxc_lint!(
...
-    /// // oxlint.json
-    /// {
-    ///     "rules": {
-    ///         "no-empty-function": ["error", { "allow": ["functions"] }]
-    ///     }
-    /// }
+    /// "no-empty-function": [
+    ///   "error",
+    ///   { "allow": ["functions"] }
+    /// ]
```
connorshea: This is what the json example looks like in a lot of files, but not sure if it'd be better to have the full rules block and/or a comment with the `.oxlintrc.json` line for clarity like it did before? 🤷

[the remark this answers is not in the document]
connorshea: I'm assuming this is fine as the docs do this all over the place.

camc314: part of the problem here, is perhaps this isn't valid json?
maybe:
`["no-empty-function": ["error": { "allow": ["functions"] }]`
would be better?

connorshea: Updated the example code, it's now valid JSON, albeit I used an object wrapper instead of an array since the array wouldn't be valid.

camc314 (approved):
Thank you!

crates/oxc_linter/src/rules/eslint/no_empty_function.rs
camc314: yep you're right

PR #14791, opened 2025-10-19
docs(linter): Add configuration docs to no-useless-escape lint rule.

connorshea:
Also add missing text for "Why is this bad?" section. I've updated this to use the autogen docs rather than an explicit Options section, since that seems to be the preferred implementation path for this kind of addition.

Part of #14743.

Results in:

```md
## Configuration

### allowRegexCharacters

type: `string[]`

default: `[]`

An array of characters that are allowed to be escaped unnecessarily in regexes.
```


camc314 (approved):
Thank yoU!

PR #14792, opened 2025-10-19
feat(rulegen): Add options section for doc block in template.

connorshea:
If there's a rule config option, add an Options section to the docs in the generated rule code.

This should help ensure new rules get this section in the future.

Part of #14743. Follows up on work in #13320.

Here is the output I get when I run `just new-rule strict` to generate an equivalent of [the ESLint strict rule](https://eslint.org/docs/latest/rules/strict):

```
declare_oxc_lint!(
    /// ### What it does
    ///
    /// Briefly describe the rule's purpose.
    ///
    /// ### Why is this bad?
    ///
    /// Explain why violating this rule is problematic.
    ///
    /// ### Examples
    ///
    /// Examples of **incorrect** code for this rule:
    /// ```js
    /// FIXME: Tests will fail if examples are missing or syntactically incorrect.
    /// ```
    ///
    /// Examples of **correct** code for this rule:
    /// ```js
    /// FIXME: Tests will fail if examples are missing or syntactically incorrect.
    /// ```
    ///
    /// ### Options
    ///
    /// #### optionName
    ///
    /// `{ type: FIXME, default: FIXME }`
    ///
    /// FIXME: Explanation of what this option does.
    ///
    /// Example:
    /// ```json
    /// {
    ///   "eslint/strict": [
    ///     "error",
    ///     { "optionName": "FIXME: example option value" }
    ///   ]
    /// }
    /// ```
```

I'm not sure if it's possible to get the option _names_ in the template, but if anyone has ideas on that I'd be happy to hear them. For now it just populates one option which is always `optionName`.

### DIFF
--- a/tasks/rulegen/template.txt
+++ b/tasks/rulegen/template.txt
@@ -52,6 +52,26 @@ declare_oxc_lint!(
     /// ```{{language}}
     /// FIXME: Tests will fail if examples are missing or syntactically incorrect.
     /// ```
+{{#if rule_config}}
+    ///
+    /// ### Options
+    ///
+    /// #### optionName
+    ///
+    /// `{ type: FIXME, default: FIXME }`
+    ///
+    /// FIXME: Explanation of what this option does.
+    ///
+    /// Example:
+    /// ```json
+    /// {
+    ///   "{{mod_name}}/{{kebab_rule_name}}": [
+    ///     "error",
+    ///     { "optionName": "FIXME: example option value" }
+    ///   ]
+    /// }
+    /// ```
+{{/if}}
     {{pascal_rule_name}},
     {{mod_name}},
     nursery, // TODO: change category to `correctness`, `suspicious`, `pedantic`, `perf`, `restriction`, or `style`


tasks/rulegen/template.txt
```diff
@@ -52,6 +52,26 @@ declare_oxc_lint!(
...
+    /// #### optionName
+    ///
+    /// `{ type: FIXME, default: FIXME }`
+    ///
+    /// FIXME: Explanation of what this option does.
+    ///
+    /// Example:
+    /// ```json
+    /// {
+    ///   "{{mod_name}}/{{kebab_rule_name}}": [
```
connorshea, line 68: I think this won't work correctly because `mod_name` is going to return with underscores here, but haven't been able to test a non-eslint rule to generate yet.

Although, this'll only be a problem for jsx-a11y since react-perf is already complete. May be a problem in the future if any other plugins are added with two word names, though.

connorshea:
I tried adding code for looping through the option keys and adding subheaders for each. It works on the rule I was testing with, but it's mostly Copilot-generated and I have no way to evaluate its quality since I don't know Rust, so didn't add it to the PR: https://github.com/connorshea/oxc/commit/9ec2ed03930ce25b3b972a9819ac104dc7bd6660

If someone else wants to look at that and contribute it/a better version to this PR, that'd be fine by me. Otherwise I think this PR is about as much as I can do without more rust.

connorshea:
I'm going to close this PR in favor of updating the template to add FIXME comments that can be picked up by the documentation autogen, I think. I didn't know about that feature before making this change.

PR #14799, opened 2025-10-20
docs(linter): Add configuration docs info for import/namespace rule

connorshea:
I can/should probably start batching these changes into larger PRs rather than opening one at a time.

Part of #14743.

Exported docs result:

```md
## Configuration

### allowComputed

type: `boolean`

default: `false`

Whether to allow computed references to an imported namespace.
```


camc314 (approved):
Thank yoU!

PR #14837, opened 2025-10-21
docs(linter): Update no-plusplus rule to expose documentation for config options

connorshea:
This results in the following configuration text in the documentation for this rule:

```md
## Configuration

### allowForLoopAfterthoughts

type: `boolean`

default: `false`

Whether to allow `++` and `--` in for loop afterthoughts.
```

Part of #14743.


camc314 (commented):
Thank you!

PR #14841, opened 2025-10-21
docs(linter): Move documentation comments for the valid-typeof rule to the struct

connorshea:
This ensures the autogenerated documentation handles the configuration section. Part of #14743.

The generated docs look like this (but without the backslashes before the code blocks):

```md
## Configuration

### requireStringLiterals

type: `boolean`

default: `false`

The `requireStringLiterals` option when set to `true`, allows the comparison of `typeof`
expressions with only string literals or other `typeof` expressions, and disallows
comparisons to any other value. Default is `false`.

With `requireStringLiterals` set to `true`, the following are examples of **incorrect** code:
\```js
typeof foo === undefined
typeof bar == Object
typeof baz === "strnig"
typeof qux === "some invalid type"
typeof baz === anotherVariable
typeof foo == 5
\```

With `requireStringLiterals` set to `true`, the following are examples of **correct** code:
\```js
typeof foo === "undefined"
typeof bar == "object"
typeof baz === "string"
typeof bar === typeof qux
\```
```

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/valid_typeof.rs
+++ b/crates/oxc_linter/src/rules/eslint/valid_typeof.rs
@@ -3,6 +3,7 @@ use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::{GetSpan, Span};
 use oxc_syntax::operator::UnaryOperator;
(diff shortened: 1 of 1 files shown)


camc314 (approved):
Thank you!

PR #14846, opened 2025-10-21
docs(linter): Add autogenerated docs for no-magic-numbers rule.

connorshea:
Part of #14743.

This is a draft because this change results in an "index[n]" header in the generated page due to the type/structure of the struct. I'm not actually 100% sure if this is a problem I should fix in this class, or a problem with the doc generator itself that needs to be fixed somehow.

Generated docs:

```md
## Configuration

### detectObjects

type: `boolean`

default: `false`

When true, numeric literals used in object properties are considered magic numbers.


### enforceConst

type: `boolean`

default: `false`

When true, enforces that number constants must be declared using `const` instead of `let` or `var`.


### ignore

type: `array`

default: `[]`

An array of numbers to ignore if used as magic numbers. Can include floats or BigInt strings.


#### ignore[n]






### ignoreArrayIndexes

type: `boolean`

default: `false`

When true, numeric literals used as array indexes are ignored.


### ignoreClassFieldInitialValues

type: `boolean`

default: `false`

When true, numeric literals used as initial values in class fields are ignored.


### ignoreDefaultValues

type: `boolean`

default: `false`

When true, numeric literals used as default values in function parameters and destructuring are ignored.


### ignoreEnums

type: `boolean`

default: `false`

When true, numeric literals in TypeScript enums are ignored.


### ignoreNumericLiteralTypes

type: `boolean`

default: `false`

When true, numeric literals used as TypeScript numeric literal types are ignored.


### ignoreReadonlyClassProperties

type: `boolean`

default: `false`

When true, numeric literals in readonly class properties are ignored.


### ignoreTypeIndexes

type: `boolean`

default: `false`

When true, numeric literals used to index TypeScript types are ignored.
```

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/no_magic_numbers.rs
+++ b/crates/oxc_linter/src/rules/eslint/no_magic_numbers.rs
@@ -7,6 +7,7 @@ use oxc_macros::declare_oxc_lint;
 use oxc_semantic::AstNodes;
 use oxc_span::{GetSpan, Span};
 use oxc_syntax::operator::UnaryOperator;
+use schemars::JsonSchema;
 
 use crate::{AstNode, context::LintContext, rule::Rule};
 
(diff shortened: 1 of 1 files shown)


connorshea:
@camc314 I think this is probably fine to merge as-is? Unless you have an idea on fixing this

camc314 (approved):

PR #14847, opened 2025-10-21
docs(linter): Add doc comments for func_style options.

connorshea:
This one is a bit weird in that there's an overrides > namedExports option that's nested? But the struct doesn't appear to quite match that expectation. Also, the struct doesn't quite match the fact that there's no key for `style`, it's just a string value at the top level.

Part of #14743.

Generated docs:

```md
## Configuration

### allowArrowFunctions

type: `boolean`

default: `false`

When true, arrow functions are allowed regardless of the style setting.


### allowTypeAnnotation

type: `boolean`

default: `false`

When true, functions with type annotations are allowed regardless of the style setting.


### style

type: `"expression" | "declaration"`

default: `"expression"`

The style to enforce. Either "expression" (default) or "declaration".
```


camc314 (approved):
thank you!

PR #14807, opened 2025-10-20
docs(linter): Add documentation on `ignoreRestSiblings` option for `no-unused-vars` rule

howenyap:
Explains the `ignoreRestSiblings` config option for the [no unused vars](https://oxc.rs/docs/guide/usage/linter/rules/eslint/no-unused-vars) rule.

Related to [this](https://github.com/oxc-project/oxc/issues/14764) issue.

PR #14881, opened 2025-10-22
fix(linter): Error on unexpected tokens when declaring lint rule

howenyap:
Related to [this](https://github.com/oxc-project/oxc/pull/14807#issuecomment-3422865650) issue where missing commas in `declare_oxc_lint!` would cause documentation to be generated wrongly without erroring.

In the image below, there is no warning about the missing comma after `dangerou_suggestion`, this causes `config` to not be generated in the documentation.

image

With the fix, users will be notified about the missing comma.

image


howenyap:
cc @connorshea

camc314 (approved):

camc314 (approved):
Thank you!

PR #14897, opened 2025-10-22
fix(editor): stop client when delete .oxlintrc.json if requireConfig

liangmiQwQ:
Closes #14896 by extending watcher detect file deleting and stop client

### DIFF
--- a/editors/vscode/client/extension.ts
+++ b/editors/vscode/client/extension.ts
@@ -331,15 +331,25 @@ function updateStatsBar(
 }
 
 function generateActivatorByConfig(config: VSCodeConfig, context: ExtensionContext): void {
-  const watcher = workspace.createFileSystemWatcher('**/.oxlintrc.json', false, true, true);
+  const watcher = workspace.createFileSystemWatcher('**/.oxlintrc.json', false, true, !config.requireConfig);
   watcher.onDidCreate(async () => {
-    watcher.dispose();
     allowedToStartServer = true;
     updateStatsBar(context, config.enable);
     if (client && !client.isRunning() && config.enable) {
       await client.start();
     }
   });
+
+  watcher.onDidDelete(async () => {
+    // only can be called when config.requireConfig
+    allowedToStartServer = (await workspace.findFiles(`**/.oxlintrc.json`, '**/node_modules/**', 1)).length > 0
+    if(!allowedToStartServer){
+      updateStatsBar(context, false)
+      if(client && client.isRunning()){
+        await client.stop()
+      }
+    }
+  })
 
   context.subscriptions.push(watcher);
 }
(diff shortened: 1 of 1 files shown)


camc314 (commented):
Thank you, is it possible to write a test for this?

editors/vscode/client/extension.ts
```diff
@@ -331,15 +331,25 @@ function updateStatsBar(
 }
 
 function generateActivatorByConfig(config: VSCodeConfig, context: ExtensionContext): void {
-  const watcher = workspace.createFileSystemWatcher('**/.oxlintrc.json', false, true, true);
+  const watcher = workspace.createFileSystemWatcher('**/.oxlintrc.json', false, true, !config.requireConfig);
```
Sysix, line 334: This function never gets called when `config.requireConfig` is `false`
```suggestion
  const watcher = workspace.createFileSystemWatcher('**/.oxlintrc.json', false, true, false);
```

Sysix (commented):
Thanks for the work ❤️  Looks good for me! Did you manually test it with the docs from: https://oxc.rs/docs/contribute/vscode.html#building-extension-and-installing-it-locally

I can test it tomorrow, if you do not have time.

>  is it possible to write a test for this?

@camc314 
Currently, not :/ Because the server is started at the beginning of the tests, and the extension does not apply the `oxc.requireConfig` changes to the `generateActivatorByConfig` logic. That why the `onDidCreate`-part had no tests too

editors/vscode/client/extension.ts
liangmiQwQ: Emm, I think this is expected? because if [requireConfig](https://github.com/oxc-project/oxc/blob/b60779353c5c9df73addfb1eaf4d8af35a7415b2/editors/vscode/README.md?plain=1#L52) is false, we actually  don't want to stop client when `.oxlintrc.json` is deleted

liangmiQwQ:
> Thanks for the work ❤️ Looks good for me! Did you manually test it with the docs from: https://oxc.rs/docs/contribute/vscode.html#building-extension-and-installing-it-locally
> 
> I can test it tomorrow, if you do not have time.
> 
> > is it possible to write a test for this?
> 
> @camc314 Currently, not :/ Because the server is started at the beginning of the tests, and the extension does not apply the `oxc.requireConfig` changes to the `generateActivatorByConfig` logic. That why the `onDidCreate`-part had no tests too

Thanks, I've actually tested it manually on local VSCode, and it seems work; but test cases seem missed and I don't know how to do it (or it is very difficult for current project struct?)

editors/vscode/client/extension.ts
Sysix: This function never gets called when `config.requireConfig` is false on extension activation.
At the end, it should be the same result 👍

Sysix (approved):

PR #14912, opened 2025-10-22
fix(linter): use aliases when parsing cli rules

lilnasy:
- Fixes #14906 
#### What went wrong
- `parse_rule_key` has the logic for aliasing, and it is called by `OxlintRules`, but only during deserialization.
- `CliRunner`, via `ConfigStoreBuilder`, instantiates `OxlintRules` directly. No post processing, aliasing doesn't take place.
#### Fix
- Aliasing logic in `parse_rule_key` is split off as `unalias_plugin_name`.
- `unalias_plugin_name` is now called by `ConfigStoreBuilder`.
- `LintFilterKind::parse` seemed like a better place at first, but lifetimes required cloning.


camc314 (approved):
Thank you!

PR #14939, opened 2025-10-24
fix(linter): reverse extends overrides priority

wagenet:
Fixes #14925

### DIFF
--- a/crates/oxc_linter/src/config/oxlintrc.rs
+++ b/crates/oxc_linter/src/config/oxlintrc.rs
@@ -242,8 +242,8 @@ impl Oxlintrc {
         let env = self.env.clone();
         let globals = self.globals.clone();
 
-        let mut overrides = self.overrides.clone();
-        overrides.extend(other.overrides);
+        let mut overrides = other.overrides.clone();
+        overrides.extend(self.overrides.clone());
 
         let plugins = match (self.plugins, other.plugins) {
             (Some(self_plugins), Some(other_plugins)) => Some(self_plugins | other_plugins),
--- a/crates/oxc_linter/src/config/oxlintrc.rs
+++ b/crates/oxc_linter/src/config/oxlintrc.rs
@@ -218,7 +218,7 @@ impl Oxlintrc {
     /// Merges two [Oxlintrc] files together
     /// [Self] takes priority over `other`
     #[must_use]
-    pub fn merge(&self, other: Oxlintrc) -> Oxlintrc {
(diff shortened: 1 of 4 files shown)


camc314 (approved):
Sorry for the delay getting to this.

I'm worred about this breaking user's configs, but technically this is a bug fix rather than a breaking change, so I guess it's ok.

crates/oxc_linter/src/config/oxlintrc.rs
```diff
@@ -218,7 +218,7 @@ impl Oxlintrc {
     /// Merges two [Oxlintrc] files together
     /// [Self] takes priority over `other`
     #[must_use]
-    pub fn merge(&self, other: Oxlintrc) -> Oxlintrc {
+    pub fn merge(&self, other: &Oxlintrc) -> Oxlintrc {
```
overlookmotel, line 221: Changing `other` from `Oxlintrc` to `&Oxlintrc` introduces the necessity for an additional `clone`. The caller has an owned `Oxlintrc`, so this is unnecessary.

I know taking an owned `Oxlintrc` is a bit weird, but given that we have one, we may as well avoid the clone.

Could you revert that change please?

crates/oxc_linter/src/config/oxlintrc.rs
```diff
@@ -218,7 +218,7 @@ impl Oxlintrc {
     /// Merges two [Oxlintrc] files together
     /// [Self] takes priority over `other`
```
overlookmotel, line 219: Is this comment still accurate? If not, please update it.

Given that this is a bit confusing (hence why implementation was wrong before this PR), it'd also be helpful to expand this comment with an example of what merging looks like - show who wins when `self` and `other` both have the same property.

overlookmotel (commented):
Sorry I'm late to the party!

Would appreciate it if you could follow-up with fixes for the 2 things below. if you don't have time, please just say so, and I'll do it.

---

Secondly, has this change also reversed the order that `extends` are applied?

If so, good! Behavior prior to this PR was not in line with what's described in the docs: https://oxc.rs/docs/guide/usage/linter/config-file-reference.html#extends (so it *was* a bug fix, not a breaking change).

But could you please add a test for this, so we don't break it again by accident?

If this PR *didn't* fix that, we should re-open #14925 for this 2nd change.

overlookmotel:
@wagenet Just a ping to check you saw my comments above. If you don't have time to make the requested changes, no problem. But please let me know either way, so I can put them on my list if you're not going to be able to get to it.

PR #15011, opened 2025-10-28
fix(linter): handle jsPlugin failures

wagenet:
Prior to this change, certain JS Plugin failures would just point to Oxlint internals instead of the actual JS error.


camc314 (changes requested):
Thanks for contributing, can we write a test for this change?

PR #15013, opened 2025-10-28
docs(linter): Add configuration option docs for eslint/sort-imports rule.

connorshea:
Part of #14743.

Generated docs:

```md
## Configuration

### allowSeparatedGroups

type: `boolean`

default: `false`

When `true`, the rule allows import groups separated by blank lines to be treated independently.


### ignoreCase

type: `boolean`

default: `false`

When `true`, the rule ignores case-sensitivity when sorting import names.


### ignoreDeclarationSort

type: `boolean`

default: `false`

When `true`, the rule ignores the sorting of import declarations (the order of `import` statements).


### ignoreMemberSort

type: `boolean`

default: `false`

When `true`, the rule ignores the sorting of import members within a single import declaration.


### memberSyntaxSortOrder

type: `array`


Specifies the sort order of different import syntaxes.


#### memberSyntaxSortOrder[n]

type: `"none" | "all" | "multiple" | "single"`
```

This is imperfect because of the handling of `memberSyntaxSortOrder[n]` by the website generation tool, but I think this is probably fine as-is.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/sort_imports.rs
+++ b/crates/oxc_linter/src/rules/eslint/sort_imports.rs
@@ -10,6 +10,7 @@ use oxc_ast::ast::{ImportDeclaration, ImportDeclarationSpecifier, Statement};
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::Span;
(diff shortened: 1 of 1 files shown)


camchenry (approved):
Thanks!

PR #15026, opened 2025-10-28
fix(linter): prevent underflow in count_comment_lines for JSX files

ityuany:
When processing JSX inline comments like `{/* comment */}` with the eslint/max-lines rule using skipComments: true, the count_comment_lines function would calculate start_line > end_line, causing integer underflow.

This resulted in:
- Debug mode: panic with "attempt to subtract with overflow"
- Release mode: wrapping to usize::MAX, incorrect line count

Fix by using saturating_sub which:
- Prevents panic in all build modes
- Returns 0 for inline comments (logically correct)
- Maintains correct behavior for multi-line comments

Added test case to prevent regression.

### DIFF
--- a/crates/oxc_linter/src/utils/comment.rs
+++ b/crates/oxc_linter/src/utils/comment.rs
@@ -36,24 +36,79 @@ mod test {
     use oxc_span::SourceType;
 
     #[test]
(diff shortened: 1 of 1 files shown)


camc314 (approved):
Thank you

PR #15032, opened 2025-10-28
fix(import/no-namespace): remove dot special case

### LINKED ISSUE #15007: linter: is `import/import-no-namespace`'s `ignore` option working properly?

latest

### What command did you run?

Through VSCode's oxc 1.24.0 extension

_No response_

### What happened?

One my TypeScript file has the following import:
```typescript
import * as z from 'zod';
```
Another one has:
```typescript
import type * as DrizzleKit from 'drizzle-kit/api';
```

Both lines are getting flagged by the `import/import-no-namespace` rule. This is a rule I'd like to use, but as far as Zod is concerned, importing the whole namespace is the recommended way.

Now instead of littering my code with `// oxlint-disable-line import/no-namespace` each time I use Zod (which seems to trigger Copilot too), I figured I would add an exception for that rule straight into my `.oxlintrc.json`.

[The documentation](https://oxc.rs/docs/guide/usage/linter/rules/import/import-no-namespace.html#options) confuses me though (besides `ignore` being spelled `ignores` in the example):

`ignore`: array of glob strings for modules that should be ignored by the rule :
```json
{
  "ignore": ["*.json"]
}
```

I can't read Rust for the life of me, so I'm not sure what's going on with the `ignore` parameter, but this:
```
"import/no-namespace": ["warn", { "ignore": ["zod"] }],
```
will disable the rule for BOTH `'zod'` and `'drizzle-kit/api'`, whereas it should only disable it for `'zod'` and still warn me about the second one.
 
What kind of globbing is really performed here?

Thank you

lilnasy:
- Closes #15007 
- The original implementation from #7229 had an unexplained behavior where presence of any ignore pattern would skip all diagnostics if the import specifier did not have a dot.
- `eslint-plugin-import` does not have this behavior: [no-namespace.js#L109](https://github.com/import-js/eslint-plugin-import/blob/01c9eb04331d2efa8d63f2d7f4bfec3bc44c94f3/src/rules/no-namespace.js#L109).

### DIFF
--- a/crates/oxc_linter/src/rules/import/no_namespace.rs
+++ b/crates/oxc_linter/src/rules/import/no_namespace.rs
@@ -107,19 +107,11 @@ impl Rule for NoNamespace {
             ImportImportName::NamespaceObject => {
                 let source = entry.module_request.name();
 
-                if self.ignore.is_empty() {
-                    ctx.diagnostic(no_namespace_diagnostic(entry.local_name.span));
-                } else {
-                    if !source.contains('.') {
-                        return;
-                    }
-
-                    if self.ignore.iter().any(|pattern| {
-                        glob_match(pattern.as_str(), source.trim_start_matches("./"))
-                    }) {
-                        return;
-                    }
-
+                if self.ignore.is_empty()
+                    || self.ignore.iter().all(|pattern| {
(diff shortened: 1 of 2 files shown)


camc314 (approved):
thank you!

PR #15052, opened 2025-10-29
docs(linter): Add configuration option docs for eslint/no-multi-assign rule.

connorshea:
Part of #14743.

Generated docs:

```md

## Configuration

### ignoreNonDeclaration

type: `boolean`

default: `false`

When set to `true`, the rule allows chains that don't include initializing a variable in a declaration or initializing a class field.

Examples of **correct** code for this option set to `false`:
```js
let a;
let b;
a = b = "baz";

const x = {};
const y = {};
x.one = y.one = 1;
```

Examples of **incorrect** code this option set to `true`:
```js
let a = b = "baz";

const foo = bar = 1;

class Foo {
a = b = 10;
}
```
```


camchenry (approved):
Thanks!

camc314 (approved):

PR #15054, opened 2025-10-29
docs(linter): Add configuration option docs for eslint/no-useless-computed-key rule.

connorshea:
These docs already existed, just needed to be moved.

Part of #14743.

Generated docs:

```md
## Configuration

### enforceForClassMembers

type: `boolean`

default: `true`

The `enforceForClassMembers` option controls whether the rule applies to
class members (methods and properties).

Examples of **correct** code for this rule with the `{ "enforceForClassMembers": false }` option:
\```js
class SomeClass {
["foo"] = "bar";
[42] = "baz";
get ['b']() {}
set ['c'](value) {}
static ["foo"] = "bar";
}
\```
```


camchenry (approved):

connorshea:
I think I've set up graphite now?

camc314 (approved):

PR #15176, opened 2025-10-31
docs(linter): Add configuration option docs for promise/no-callback-in-promise rule.

connorshea:
Part of #14743.

Generated docs:

```md
## Configuration

### callbacks

type: `string[]`

default: `["callback", "cb", "done", "next"]`

List of callback function names to check for within Promise `then` and `catch` methods.


### exceptions

type: `string[]`

default: `[]`

List of callback function names to allow within Promise `then` and `catch` methods.
```


connorshea:
Actually I think Copilot is right and I misunderstood the intent of the `callbacks` config field here, it's distinct from `exceptions` (which was not previously represented in the config struct, and `callbacks` never existed in the original ESLint rule, which is the source of my assumption/confusion).

camc314 (approved):

PR #15202, opened 2025-11-01
fix(linter): ignore script tag in code comment

liangmiQwQ:
Partial loader tries parsing the `<script>` tag in the code comment incorrectly

For example, we have a `.vue` file

```vue
<!-- <script setup lang="tsx"> -->
<script setup lang="ts">
const count = ref(0)
</script>
```

And the loader will regard these as script part

```
 -->
<script setup lang="ts">
const count = ref(0)
```

So this PR added some `Finder`s for code comment to ignore the `<script>` in code comment, also works on `.svelte` and `.astro` files

### DIFF
--- a/crates/oxc_linter/src/loader/partial_loader/astro.rs
+++ b/crates/oxc_linter/src/loader/partial_loader/astro.rs
@@ -2,12 +2,9 @@ use memchr::memmem::{Finder, FinderRev};
 
 use oxc_span::{SourceType, Span};
 
-use crate::loader::{
-    JavaScriptSource,
-    partial_loader::{COMMENT_END, COMMENT_START, SCRIPT_START, find_script_start},
-};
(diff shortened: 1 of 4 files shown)


camc314 (approved):

connorshea:
I hate myself for noticing this, but:

```rs
#[test]
fn test_script_inside_style_code_comment() {
    let source_text = r"
    <style scoped>
    /* <script>teehee</script> */
    .a { color: red; }
    </style>
    <script setup>c</script>
    ";

    let result: JavaScriptSource<'_> = parse_vue(source_text);
    assert_eq!(result.source_text, "c");
}
```

```
thread 'loader::partial_loader::vue::test::test_script_inside_style_code_comment' (25820582) panicked at crates/oxc_linter/src/loader/partial_loader/vue.rs:329:9:
  left: "teehee"
 right: "c"
```

PR #15214, opened 2025-11-01
fix(linter): patch `@typescript-eslint/scope manager`

lilnasy:
- Part of #14827
- Fixes the bug noticed in a review in https://github.com/oxc-project/oxc/pull/14890#discussion_r2452315938
- Originally reported as https://github.com/typescript-eslint/typescript-eslint/issues/11714
- Originally fixed in https://github.com/typescript-eslint/typescript-eslint/pull/11715
- Patching in oxlint because neither the bug nor the fix has been reviewed in the `typescript-eslint` org.

### DIFF
--- a/apps/oxlint/test/fixtures/sourceCode_scope_methods/output.snap.md
+++ b/apps/oxlint/test/fixtures/sourceCode_scope_methods/output.snap.md
@@ -106,7 +106,7 @@
  8 |   };
    `----
 
-  x scope-plugin(scope): isGlobalReference(Math): false
+  x scope-plugin(scope): isGlobalReference(Math): true
    ,-[files/index.js:7:28]
  6 |   return function innerFunction() {
  7 |     return localConstant + Math.PI;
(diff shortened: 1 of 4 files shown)


overlookmotel (approved):
Sorry it's taken me so long to get to this. It's a bit of a head-scratcher, what with TS-ESLint's strange behavior, and the layer of changes that ESLint lays on top.

But let's merge this, and track what happens on your PR and on https://github.com/typescript-eslint/typescript-eslint/issues/11762. We can amend our approach if necessary.

There's still a small incompatibility:

In ESLint, in a script (not module) containing `var foo;`, `isGlobalReference(foo) === true`, but in our impl it returns `false`.

But in ESLint, when `sourceType` is `"commonjs"`, `isGlobalReference(foo) === false`.

We don't support `sourceType: "commonjs"` at present. Our behavior for `"script"` behaves more like `"commonjs"` in ESLint - which is probably a good thing, because that's more likely to be the type of files people will be linting with Oxlint.

PR #15249, opened 2025-11-04
feat(linter): permit comments in `.oxlintrc.json` via json schema file

DecimalTurn:
With this small change, VS Code will stop complaining about comments in oxlint configuration files.
Similar usages: https://github.com/search?q=path%3A*schema*.json+%2F%5E%5B+%5D%7B0%2C4%7D%22allowcomments%22%3A+true%2F&type=code

### DIFF
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -571,7 +571,8 @@ mod test {
     #[test]
     fn test_schema_json() {
         let path = get_project_root().unwrap().join("npm/oxlint/configuration_schema.json");
-        let schema = schemars::schema_for!(Oxlintrc);
+        let mut schema = schemars::schema_for!(Oxlintrc);
+        schema.schema.extensions.insert("allowComments".to_string(), serde_json::Value::Bool(true));
         let json = serde_json::to_string_pretty(&schema).unwrap();
         let existing_json = fs::read_to_string(&path).unwrap_or_default();
         if existing_json.trim() != json.trim() {
--- a/crates/oxc_linter/src/lib.rs
+++ b/crates/oxc_linter/src/lib.rs
@@ -572,6 +572,8 @@ mod test {
     fn test_schema_json() {
         let path = get_project_root().unwrap().join("npm/oxlint/configuration_schema.json");
         let mut schema = schemars::schema_for!(Oxlintrc);
+        // setting `allowComments` to true to allow comments in JSON schema files
+        // https://github.com/microsoft/vscode-json-languageservice/blob/356d5dd980d49c6ac09ec8446614a6f94016dcea/src/jsonLanguageTypes.ts#L127-L131
         schema.schema.extensions.insert("allowComments".to_string(), serde_json::Value::Bool(true));
         let json = serde_json::to_string_pretty(&schema).unwrap();
         let existing_json = fs::read_to_string(&path).unwrap_or_default();
--- a/npm/oxlint/configuration_schema.json
+++ b/npm/oxlint/configuration_schema.json
@@ -121,6 +121,7 @@
       ]
     }
   },
+  "allowComments": true,
   "definitions": {
     "AllowWarnDeny": {
       "oneOf": [
--- a/crates/oxc_linter/src/snapshots/schema_json.snap
+++ b/crates/oxc_linter/src/snapshots/schema_json.snap
@@ -125,6 +125,7 @@ expression: json
       ]
     }
   },
+  "allowComments": true,
   "definitions": {
     "AllowWarnDeny": {
       "oneOf": [


npm/oxlint/configuration_schema.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Oxlintrc",
   "description": "Oxlint Configuration File\n\nThis configuration is aligned with ESLint v8's configuration schema (`eslintrc.json`).\n\nUsage: `oxlint -c oxlintrc.json --import-plugin`\n\n::: danger NOTE\n\nOnly the `.json` format is supported. You can use comments in configuration files.\n\n:::\n\nExample\n\n`.oxlintrc.json`\n\n```json\n{\n\"$schema\": \"./node_modules/oxlint/configuration_schema.json\",\n\"plugins\": [\"import\", \"typescript\", \"unicorn\"],\n\"env\": {\n\"browser\": true\n},\n\"globals\": {\n\"foo\": \"readonly\"\n},\n\"settings\": {\n},\n\"rules\": {\n\"eqeqeq\": \"warn\",\n\"import/no-cycle\": \"error\",\n\"react/self-closing-comp\": [\"error\", { \"html\": false }]\n},\n\"overrides\": [\n{\n\"files\": [\"*.test.ts\", \"*.spec.ts\"],\n\"rules\": {\n\"@typescript-eslint/no-explicit-any\": \"off\"\n}\n}\n]\n}\n```",
+  "allowComments": true,
```
connorshea: This is a generated file, we unfortunately will need to do this a different way otherwise it will get overridden next time someone regenerates it.

DecimalTurn: Ah, oops, I'll check if I can make that change in the generation process as well.

connorshea: I made a ticket on the website repo to improve the docs for this https://github.com/oxc-project/oxc-project.github.io/issues/598

I believe the simplest way to regenerate the configuration_schema.json is to run `cargo test -p oxc_linter`, if you need to test your changes. It uses the schemars crate to pull information/types from rust doc comments and decorators across the oxlint code.

DecimalTurn: Indeed, that's a bit unintuitive that the test is generating the schema, but it works I guess.
I've amended the commit with the fix including the generation step + snapshot and a commit message that better fits the repo's conventions.

crates/oxc_linter/src/lib.rs
```diff
@@ -571,7 +571,8 @@ mod test {
     #[test]
     fn test_schema_json() {
         let path = get_project_root().unwrap().join("npm/oxlint/configuration_schema.json");
-        let schema = schemars::schema_for!(Oxlintrc);
+        let mut schema = schemars::schema_for!(Oxlintrc);
+        schema.schema.extensions.insert("allowComments".to_string(), serde_json::Value::Bool(true));
```
connorshea, line 577: Should probably add a comment explaining that this is necessary for VS Code and other editors to allow comments in a `.json` file.

npm/oxlint/configuration_schema.json
connorshea: nice, thanks!

connorshea (approved):
Would ideally have the extra comment in lib.rs explaining why this is necessary (`.json` not allowing comments normally), but not a huge problem, LGTM! Thanks :) 

I'll leave it to camc to decide whether to merge, in case there are any concerns he has.

connorshea:
cc: @leaysgur should maybe make the same change in oxfmt (in a separate PR) if we're allowing .json files?

crates/oxc_linter/src/lib.rs
DecimalTurn: I hesitated to mention VS Code in the code directly since this is technically not limited to VS Code. Any editor that supports JSON schemas and `allowComments` can benefit from this change.

leaysgur:
Thanks for letting me know. 😉 

I was just wondering: is this purely for VSCode?
I couldn't find any mention of this field in the JSON Schema specification.

If that's the case, I personally don't think it's a good solution.

I believe similar settings exist in local editor configurations, and simply using the `.jsonc` extension feels more natural.

connorshea:
> Thanks for letting me know. 😉
> 
> I was just wondering: is this purely for VSCode? I couldn't find any mention of this field in the JSON Schema specification.
> 
> If that's the case, I personally don't think it's a good solution.
> 
> I believe similar settings exist in local editor configurations, and simply using the `.jsonc` extension feels more natural.

It is not part of the JSON Schema spec officially, no. I think it improves the developer experience without really harming users of other editors, we already have the oxc extension for VS Code telling the editor to treat `.oxlintrc.json` files as "JSON with Comments", this just allows that to happen without needing the extension. I assume the same is true for `.oxfmtrc.json` files with the oxc extension, though I haven't checked. But ultimately up to you for oxfmt.

camc314 (approved):

leaysgur:
(I forgot to bring this to Cam during the meeting.)

As a result, feel free to send a PR for `oxfmt` too. 👌🏻 
I'd like to make it mandatory to add a comment stating that it might only work for VSCode though.

camc314 (approved):

PR #15251, opened 2025-11-04
docs(linter): Initial attempt to update the configuration_schema with all rule names.

connorshea:
This will allow editors and tools that consume the schema to provide better autocompletion for rule names in the configuration file.

Screenshot 2025-11-03 at 8 42 48 PM

Unfortunately, it also does not allow for rules with arbitrary names (which is obviously necessary because of aliases _and_ js plugins), and the code is quite scuffed right now since I've generated it with GitHub Copilot. I'd like to figure out the right way to implement this, so just opening this as a vague Proof of Concept.


connorshea:
I think this is probably worth considering in the future, but for now it's too messy

PR #15270, opened 2025-11-04
fix(oxfmt): ignore symlink ends with `.js`

liangmiQwQ:
### Problem

if we have a directory like this and run oxfmt in it

```bash
.
├── bar.js -> foo
├── foo
│   └── file.js
└── root.js
```

It will panic like this: 

```bash
thread '<unnamed>' panicked at apps/oxfmt/src/service.rs:51:52:
Failed to read file: Os { code: 21, kind: IsADirectory, message: "Is a directory" }
```

because it try reading `bar.js` as a `.js` file but it is actually a symlink

As I know, symlink is common in some projects with complex script or in pnpm node_modules

### Fixes

Fixed by using `FileType::is_file` instead of `!FileType::is_dir`

Thanks ❤️


liangmiQwQ:
Maybe it related to #11930? but it is on oxfmt side

apps/oxfmt/src/walk.rs
```diff
@@ -227,7 +227,8 @@ impl ignore::ParallelVisitor for WalkVisitor {
                     return ignore::WalkState::Continue;
                 };
 
-                if !file_type.is_dir()
+                #[expect(clippy::filetype_is_file)]
```
leaysgur, line 231:
```suggestion
                // Use `is_file()` to detect symlinks to the directory named `.js`
                #[expect(clippy::filetype_is_file)]
```

leaysgur:
Thank you~.

> Maybe it related to https://github.com/oxc-project/oxc/issues/11930? but it is on oxfmt side

Potentially, but `oxfmt` follows Prettier's behavior for now and it ignores symlinks.

I will merge this w/ my nits commit just adding the comment.

PR #15329, opened 2025-11-05
feat(linter): implement react/state-in-constructor rule

baevm:
This PR adds react/state-in-constructor rule, issue #1022, also moved some methods to ast/react utils to reuse in this rule

[rule doc](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/state-in-constructor.md)
[rule source code](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/state-in-constructor.js)


camc314 (approved):
Thank you!

PR #15364, opened 2025-11-06
fix(linter/plugins): Return empty object for unimplemented parserServices

magic-akari:
## Summary

- Changed `sourceCode.parserServices` from throwing an error to returning an empty object `{}`
- Aligns with ESLint specification for parsers that don't implement parser services
- Prevents crashes in ESLint plugins such as `eslint-plugin-better-tailwindcss`.

## Motivation

According to the [ESLint custom parser documentation](https://eslint.org/docs/latest/extend/custom-parsers#parseforeslint-return-object):

> services can contain any parser-dependent services (such as type checkers for nodes). The value of the services property is available to rules as context.sourceCode.parserServices. **Default is an empty object.**

The current implementation throws an error when `parserServices` is accessed, which breaks plugins that use defensive checks like:

```typescript
if(typeof ctx.sourceCode.parserServices?.defineTemplateBodyVisitor === "function"){
  // ...
}
```

Plugin authors defensively handle:
- ✅ `parserServices` being undefined
- ✅ `defineTemplateBodyVisitor` not existing
- ❌ `parserServices` getter throwing an error (unexpected)

## Solution

Return an empty object `{}` for unimplemented parser services, which:
1. Matches ESLint's documented default behavior
2. Allows plugins' defensive checks to work correctly
3. Provides correct semantics for "not implemented"

## Test Plan

- Existing tests should continue to pass
- Plugins using defensive checks like the example above will no longer crash

## Related

Fixes the compatibility issue with plugins like [eslint-plugin-better-tailwindcss](https://github.com/schoero/eslint-plugin-better-tailwindcss/blob/a90cc3f547098a4293d7b0b7998bf6774f717cd2/src/utils/rule.ts#L140)

### DIFF
--- a/apps/oxlint/src-js/plugins/source_code.ts
+++ b/apps/oxlint/src-js/plugins/source_code.ts
@@ -127,7 +127,7 @@ export const SOURCE_CODE = Object.freeze({
 
   // Get parser services for the file.
   get parserServices(): { [key: string]: unknown } {
-    throw new Error('`sourceCode.parserServices` not implemented yet'); // TODO
+    return {}; // TODO
   },
 
   // Get source text as array of lines, split according to specification's definition of line breaks.
--- a/apps/oxlint/test/e2e.test.ts
+++ b/apps/oxlint/test/e2e.test.ts
@@ -254,4 +254,8 @@ describe('oxlint CLI', () => {
   it('should support UTF16 characters in source code and comments with correct spans', async () => {
     await testFixture('unicode-comments');
   });
+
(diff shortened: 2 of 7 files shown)


overlookmotel:
@copilot Define an empty object in a top level `const parserServices = {};`. Make `parserServices` getter return that const, rather then creating a new object each time.

overlookmotel (approved):
Copilot isn't answering my call. I'll make that change in a follow-up.

I pushed a couple of small changes to this PR.

PR #15513, opened 2025-11-09
perf(lexer): skip single space in `read_next_token`

overlookmotel:
It's very common for tokens to be separated by a single space. e.g. `const x = 1`, `x === y`.

Previously a single space resulted in calling the `SPS` byte handler, which consumes the space, and then going round the loop again in `Lexer::read_next_token`.

Instead, branchlessly consume a single space (if there is one) before calling the byte handler.

Gives between 2% and 7% perf improvement on parser benchmarks.

---

This also enables a further optimization (not yet implemented).

Now the handler for whitespace (`SPS`) no longer has a hot path for single spaces - it's now only called for a tab, or a 2nd space in a row. In both those cases, it's quite likely there'll be more whitespace following it, so it can now be optimized for that case, and continue consuming bytes until it finds one that *isn't* whitespace.

If handlers for whitespace, line breaks, and comments all continue consuming bytes until they find a "real" token, then we can get rid of `Kind::Skip`, and remove the loop from `read_next_token`. This would remove another unpredictable branch.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -329,25 +329,68 @@ impl<'a> Lexer<'a> {
     fn read_next_token(&mut self) -> Kind {
         self.trivia_builder.has_pure_comment = false;
         self.trivia_builder.has_no_side_effects_comment = false;
+
+        let end_pos = self.source.end();
         loop {
-            let offset = self.offset();
-            self.token.set_start(offset);
+            // Single spaces between tokens are common, so consume a space before processing the next token.
+            // Do this without a branch. This produces more instructions, but avoids an unpredictable branch.
+            // Can only do this if there are at least 2 bytes left in source.
+            // If there aren't 2 bytes left, delegate to `read_next_token_at_end` (cold branch).
+            let mut pos = self.source.position();
+            // SAFETY: `source.end()` is always equal to or after `source.position()`
+            let remaining_bytes = unsafe { end_pos.offset_from(pos) };
+            if remaining_bytes >= 2 {
+                // Read next byte.
+                // SAFETY: There are at least 2 bytes remaining in source.
+                let byte = unsafe { pos.read() };
 
-            let Some(byte) = self.peek_byte() else {
-                // Hint to compiler that this branch is rarely taken (only once at EOF)
-                #[cold]
-                fn eof() -> Kind {
-                    Kind::Eof
+                // If next byte is a space, advance by 1 byte.
+                // Do this with maths, instead of a branch.
+                let is_space = byte == b' ';
+                // SAFETY: There are at least 2 bytes remaining in source, so advancing 1 byte cannot be out of bounds
+                pos = unsafe { pos.add(usize::from(is_space)) };
+                self.source.set_position(pos);
+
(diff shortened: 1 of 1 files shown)


overlookmotel:
* **#15513** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15513?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15513?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15520** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15520?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15519** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15519?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15517** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15517?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15513).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
First version of this PR showed very weird benchmark results - massive regression on the lexer benchmarks, but improvement on all parser benchmarks (between 2% - 7%).

Turns out that this was an oddity of the benchmark code. The increase in size of `next_token` lead to it no longer getting inlined into lexer benchmark, which was a massive perf hit.

#15519 and #15520 fix that, so now we can see the effects of this change in isolation. And they're quite good!

Boshen (commented):
Claude study assembly code and discovered the jump table has a huge branch mis-prediction problem, but I couldn't figure out where the branch is.

Question: can we use simd to find the next position that is not a whitespace?

Thought: We should study v8 at some point.

Boshen:
### Merge activity

* **Nov 9, 4:45 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 9, 4:45 PM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 9, 4:55 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

overlookmotel:
> Claude study assembly code and discovered the jump table has a huge branch mis-prediction problem, but I couldn't figure out where the branch is.

Interesting! Can you post what he said?

I don't think it's surprising though. The branch *is* the jump table. CPU cannot predict which way it'll jump.

I've opened an issue in backlog about this: https://github.com/oxc-project/backlog/issues/192

> Question: can we use simd to find the next position that is not a whitespace?

Yes, and Rust stabilized `avx512` support recently. But I don't think the problem of poor branch prediction is specific to whitespace. It affects all tokens.

> Thought: We should study v8 at some point.

Yes probably!

PR #15537, opened 2025-11-10
docs(linter): Add `search: false` frontmatter to generated linter docs.

connorshea:
This prevents the docs from indexing generated pages for search. See https://vitepress.dev/reference/default-theme-search#example-excluding-pages-from-search

We could alternatively do this in the vitepress config side of things and explicitly block any results with these three names, up to the maintainers.

Searching for the CLI would bring up the `generated` pages, so there would be two results and it was a bit confusing:
Screenshot 2025-11-09 at 6 18 01 PM

### DIFF
--- a/justfile
+++ b/justfile
@@ -244,7 +244,7 @@ watch-playground:
 # When testing changes to the website documentation, you may also want to run `dprint fmt --staged`
 # in the website directory.
 website path:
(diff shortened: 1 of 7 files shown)


Boshen (commented):
🤔 why a `--frontmatter` flag? This whole task is for the website, which means those pages should have `search: false` by default.

Please let me know if I forgot how this works 😅

connorshea:
> 🤔 why a `--frontmatter` flag? This whole task is for the website, which means those pages should have `search: false` by default.
> 
> Please let me know if I forgot how this works 😅

Oh you're right, for some reason I thought it was used by the rules CLI flag to generate the rules table, I'll remove the flag.

PR #15542, opened 2025-11-10
test(coverage/typescript): Add reviewed tsc diagnostics error codes

### LINKED ISSUE #11582: parser: Improve TS error story

leaysgur:
Finally fixes #11582 

This PR adds a script that generates a list of TSC error codes.

The codes are collected based on the TypeScript version managed by `just submodules`.
And the list is then committed to Git.

When the submodule version is updated, the script is also called and the list is regenerated.

If TSC introduces new diagnostic codes, the diff will show the added code, which we need to review.
Either leave it for the parser to report or ignore it if we cannot support it.


leaysgur:
* **#15542** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15542).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Boshen (commented):
I'm not comfortable maintaining another submodule 😢 please let me study what this PR is actually doing first.

Boshen:
Opps I'm mistake, there is actually no submodule, just a task.

Boshen:
### Merge activity

* **Nov 13, 1:05 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 13, 1:06 PM UTC**: `Boshen` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 13, 1:07 PM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) couldn't merge this PR because **it was in draft mode**.

Boshen (commented):
This should be part of the conformance runner instead of separate js task, which I assume is slow.

I'll get claude to work on this at some point.

leaysgur:
Then, my Claude will do it tomorrow morning. 🫡

leaysgur:
#15686

PR #15574, opened 2025-11-11
fix(linter): Allow file extensions without a dot in react/jsx-filename-extension rule

### LINKED ISSUE #15508: linter: react/jsx-filename-extension config option does not work the same as original rule

1.26.0

### What command did you run?

`oxlint -c .oxlintrc.json`

```jsonc
{
  "plugins": [
    "react"
  ],
  "rules": {
    "react/jsx-filename-extension": [
      "error",
      {
        "extensions": [
          // baz.tsx would trigger this lint error right now even though it should not, because `tsx` does not have a `.` at the start of it.
          ".jsx",
          "tsx"
        ]
      }
    ]
  }
}
```

### What happened?

One works and the other does not.

The original ESLint rule is checking the extension like so:
`const isAllowedExtension = allowedExtensions.some((extension) => filename.slice(-extension.length) === extension);`

Which is why that worked for ESLint, `.jsx` grabs the last 4 characters of the filename, `jsx` grabs the last 3 characters, and thus either way it passes as long as the file has an acceptable extension. In oxc, it only works for extensions with `.` at the start.

```rs
let extensions = config
    .and_then(|v| v.get("extensions"))
    .and_then(Value::as_array)
    .map(|v| {
        v.iter()
            .filter_map(serde_json::Value::as_str)
            // filter to only
            // this (silently) only considers extensions that start with a dot, different behavior from eslint
            .filter(|&s| s.starts_with('.'))
            .map(|s| &s[1..])
            .map(CompactStr::from)
            .collect()
    })
    .unwrap_or(vec![CompactStr::from("jsx")]);
```

I noticed this issue because of Mastodon having `tsx` as [a config option on this lint rule in their eslint config](https://github.com/mastodon/mastodon/blob/6337e036f3bc5ece013d5b2536d7d16b1ec658e8/eslint.config.mjs#L272-L277).

Solution: Make this work the same with either `.tsx` or `tsx`. Omitting/including the dot should not matter.

connorshea:
Allow `tsx` to work fine for the rule config, previously it would be thrown out.

Also update the help diagnostic to note which extensions are allowed by the current config, and add a few more tests.

Fixes #15508

### DIFF
--- a/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
+++ b/crates/oxc_linter/src/rules/react/jsx_filename_extension.rs
@@ -277,8 +277,8 @@ fn test() {
             None,
             Some(PathBuf::from("foo.js")),
         ),
+        // Test that a commented-out JSX code snippet does not count.
         (
-            // Test that a commented-out JSX code snippet does not count.
             "// export function MyComponent() { return <><Comp /><Comp /></>;}\n",
             Some(serde_json::json!([{ "allow": "as-needed", "ignoreFilesWithoutCode": true }])),
             None,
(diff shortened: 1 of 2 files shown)


camc314 (approved):

connorshea:
Ah I couldn't figure out how to get the suggestion from copilot working so figured they weren't a problem, thanks for resolving those!

PR #15627, opened 2025-11-12
perf(formatter/sort_imports): Use `unsafe {}` for more perf

leaysgur:


leaysgur:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15627?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#15627** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15625** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15625?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#15626](https://github.com/oxc-project/oxc/pull/15626) <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15626?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15627).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

leaysgur:
- formatter[binder.ts]: +1% = total +3%
- formatter[react.development.js]: +1% = total +3%
- formatter[cal.com.tsx]: +2% = total +3%
- formatter[RadixUIAdoptionSection.jsx]: -1% = total -1%

overlookmotel (commented):
I'm afraid there's a memory leak here.

### If all entries get moved

If all the entries in `elements` get moved into `new_elements`, then you're not leaking any `FormatElement`s. But the `Vec` also owns a heap allocation in which it stores the bytes comprising those `FormatElement`s. That allocation also needs to be freed.

*If* all the entries get moved into `new_elements`, then you could free the original `Vec`'s allocation by adding this at the very end:

```rs
// All entries in `elements` have now been moved.
// Truncate the `Vec`, and drop it.
unsafe { elements.set_len(0) };
let _ = ManuallyDrop::into_inner(elements);
```

If you're going to do this, I'd suggest adding a bunch of debug assertions to make sure no bugs.

* Count the entries which are moved in the loop and `debug_assert!` that count is what was expected after the loop - to catch any bugs.
* Check that no 2 ranges overlap.

### If not all entries get moved

If it's possible that not all the entries get moved, then it's much more tricky. The above won't work, because you'll be leaking some of the `FormatElement`s too.

### Alternative approach

Assuming that all entries get moved, you may find that it's equally performant to use less unsafe.

`FormatElement` has a niche, so `Option<FormatElement>` is the same size as `FormatElement`. Compiler is smart enough to see that it can convert `FormatElement` to `Option<FormatElement>` in place, so this line has zero cost:

```rs
let mut elements = elements.into_iter().map(Some).collect::<Vec<_>>();
```

It compiles to same number of instructions as an unsafe version which does the conversion without any checks:

https://godbolt.org/z/voncYPPWs

Once this is inlined, even those instructions will disappear.

Then:

1. Wrap `elements` in `ManuallyDrop` before the loop.
2. Use `take` in the loop, but add some unchecked assertions so compiler removes checks.
3. `set_len(0)`, unpack the `ManuallyDrop` after the loop (as above).

Putting it all together:

```rs
use oxc_data_structures::assert_unchecked;

// These assertions not for safety, but just to make sure the conversion of `Vec` to `Option`s is zero-cost
const _: () = {
    assert!(size_of::<Option<FormatElement>>() == size_of::<FormatElement>());
    assert!(align_of::<Option<FormatElement>>() == align_of::<FormatElement>());
};
let elements = elements.into_iter().map(Some).collect::<Vec<_>>();
let mut elements = ManuallyDrop::new(elements);

let mut moved_count = 0;
for instruction in element_copy_plan {
    match instruction {
        ElementCopyInstruction::CopyRange(range) => {
            for idx in range {
                // Inform compiler that `idx` is in bounds, and that the element at `idx` is `Some`.
                // This removes bounds check from `elements[idx]`, and removes the `.unwrap()`.
                unsafe {
                    assert_unchecked!(idx < elements.len());
                    assert_unchecked!(elements[idx].is_some());
                }
                let element = elements[idx].take().unwrap();
                next_elements.push(element);
                moved_count += 1; // This line will be optimized out in release mode
            }
        }
        ElementCopyInstruction::InsertLine(mode) => {
            next_elements.push(FormatElement::Line(mode));
        }
    }
}

debug_assert!(moved_count == elements.len());

unsafe { elements.set_len(0) };
let _ = ManuallyDrop::into_inner(elements);
```

You *may* find that's as performant as the very unsafe version in this PR.

---

One other optimization: Prevent `new_elements` from having to reallocate by counting how many `InsertLine` instructions are added, and reserve enough space in `new_elements`.

```rs
let mut next_elements = Vec::with_capacity(elements.len() + insert_line_count);
```

If you're *really* sure you've got the count exactly right, then you can add another unchecked assertion to remove another bounds check:

```rs
let element = elements[idx].take().unwrap();
// SAFETY: There's definitely room for this without going out of bounds
unsafe { assert_unchecked!(next_elements.len() < next_elements.capacity()) };
next_elements.push(element);
```

(and ditto before `next_elements.push(FormatElement::Line(mode));`)

overlookmotel:
Actually, most `FormatElement`s presumably won't move. So what you *really* want to do is shuffle them around in place, rather than copying them all to a new `Vec`. But that's even harder to do...

Now I understand why the perf improvement is proportional to file size. The bigger the file, the more `FormatElement`s you're having to copy (regardless of how many are `import`s).

leaysgur:
Thank you for taking the time. It really helps! 🙏🏻 (Even though it's your day off)

I've updated the code for now, but the biggest problem is that I don't fully understand the content...! (I'll take time again once the benchmark files are updated)

I think some of the advice could also apply to the previous PR, so I plan to implement it there.

leaysgur:
I'll try again with resolving the conflicts in the near future.

PR #15664, opened 2025-11-13
feat(transformer): support tagged template expression with `</script` transformation

### LINKED ISSUE #15306: codegen: add a way to keep `</script>` as-is for tagged template
Because ``String.raw`</script>` === String.raw`<\/script>` `` returns `false`, I think there should be a way to keep `</script>` as-is or do a different transformation to avoid `</script>`. Note that `` `</script>` === `<\/script>` `` returns true.

If we encounter this case and the escape option is not explicitly configured,we should also emit a warning as the user has to escape on their selves.

esbuild lowers this case (https://github.com/evanw/esbuild/commit/58ac420581421b36c03649c0018286e5a8e58ffd#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL48-R50, [esbuild try](https://esbuild.github.io/try/#dAAwLjI1LjEyAABmb29gPC9zY3JpcHQ+YA)) so the user doesn't have to configure it.

See https://github.com/rolldown/rolldown/issues/6829 for context

Dunqing:
Implement a transformation plugin for tagged template expressions containing </script> tags to prevent script tag issues in browser environments.

  This plugin transforms tagged template literals containing `</script` (case-insensitive) to use a helper function call with template caching, matching esbuild's behavior.

 ### Example

  Transforms tagged templates to use cached template objects with a helper function:

  Input:
```js
foo`</script>`
bar`<script>${content}</script>`
```
  Output:
```js
var _templateObject, _templateObject2;
foo(_templateObject || (_templateObject = babelHelpers.taggedTemplateLiteral(["<\/script>"])));
bar(_templateObject2 || (_templateObject2 = babelHelpers.taggedTemplateLiteral(["<script>", "<\/script>"])), content);
```

 Closes #15306

### DIFF
--- a/crates/oxc_transformer/src/plugins/tagged_template_transform.rs
+++ b/crates/oxc_transformer/src/plugins/tagged_template_transform.rs
@@ -1,0 +1,214 @@
+//! Tagged Template Transform
+//!
+//! This plugin transforms tagged template expressions containing `</script` tags
+//! to use a helper function call instead, preventing script tag issues in browser environments.
+//!
+//! ## Background
+//!
+//! Tagged template literals containing `</script` can break when the code is embedded
+//! directly in HTML script tags, because the browser will incorrectly parse `</script` as
+//! the end of the script tag.
+//!
+//! Additionally, String.raw`</script>` !== String.raw`<\/script>` due to how tagged
+//! template literals preserve the raw string values.
+//!
+//! ## Solution
+//!
+//! This plugin transforms tagged template expressions containing `</script` to use
+//! a cached template object with a helper function, matching esbuild's behavior.
+//!
+//! ## Example
+//!
+//! Input:
+//! ```js
+//! foo`</script>`
+//! foo`<script>${content}</script>`
+//! ```
+//!
+//! Output:
+//! ```js
+//! var _foo;
+//! var _foo2;
+//! foo(_foo || (_foo = babelHelpers.taggedTemplateLiteral(["<\/script>"])));
+//! foo(_foo2 || (_foo2 = babelHelpers.taggedTemplateLiteral(["<script>", "<\/script>"])), content);
+//! ```
+//!
+//! ## References
+//!
+//! - esbuild implementation: <https://github.com/evanw/esbuild/blob/d6427c91edab734da686c4c5d29ed580b08b9fd5/internal/js_parser/js_parser.go#L13894-L13907>
+//! - Issue: <https://github.com/oxc-project/oxc/issues/15306>
+
+use std::iter;
+
+use oxc_allocator::{TakeIn, Vec as ArenaVec};
+use oxc_ast::ast::*;
+use oxc_semantic::SymbolFlags;
+use oxc_span::SPAN;
+use oxc_traverse::{BoundIdentifier, Traverse};
+
+use crate::{
+    common::helper_loader::Helper,
+    context::{TransformCtx, TraverseCtx},
+    state::TransformState,
+};
+
+pub struct TaggedTemplateTransform<'a, 'ctx> {
+    ctx: &'ctx TransformCtx<'a>,
+}
+
+impl<'a> Traverse<'a, TransformState<'a>> for TaggedTemplateTransform<'a, '_> {
+    // `#[inline]` because it is a hot path and it is directly delegated to `transform_tagged_template`
+    #[inline]
+    fn enter_expression(&mut self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
+        self.transform_tagged_template(expr, ctx);
+    }
+}
+
+impl<'a, 'ctx> TaggedTemplateTransform<'a, 'ctx> {
+    pub fn new(ctx: &'ctx TransformCtx<'a>) -> Self {
+        Self { ctx }
+    }
+
+    /// Check if the template literal contains a `</script` tag; note it is case-insensitive.
+    fn contains_closing_script_tag(quasi: &TemplateLiteral) -> bool {
+        const SCRIPT_TAG: &[u8] = b"</script";
+
+        quasi.quasis.iter().any(|quasi| {
+            let raw = &quasi.value.raw;
+
+            // The raw string must be at least as long as the script tag
+            if raw.len() < SCRIPT_TAG.len() {
+                return false;
+            }
+
+            let raw_bytes = raw.as_bytes();
+            // Get the bytes up to the last possible starting position of the script tag
+            let max_remain_len = raw_bytes.len().saturating_sub(SCRIPT_TAG.len());
+            let raw_bytes_iter = raw_bytes[..=max_remain_len].iter().copied().enumerate();
+            for (idx, byte) in raw_bytes_iter {
+                if byte == b'<'
+                    && SCRIPT_TAG
+                        .iter()
+                        .zip(raw_bytes[idx..].iter())
+                        .all(|(a, b)| *a == b.to_ascii_lowercase())
+                {
+                    return true;
+                }
+            }
+
+            false
+        })
+    }
+
+    /// Transform a tagged template expression to use the [`Helper::TaggedTemplateLiteral`] helper function
+    // `#[inline]` so that compiler can see `expr` should be a `TaggedTemplateExpression` and reduce redundant checks
+    #[inline]
+    fn transform_tagged_template(&self, expr: &mut Expression<'a>, ctx: &mut TraverseCtx<'a>) {
+        if !matches!(expr, Expression::TaggedTemplateExpression(tagged) if Self::contains_closing_script_tag(&tagged.quasi))
+        {
+            return;
+        }
+
+        let Expression::TaggedTemplateExpression(tagged) = expr.take_in(ctx.ast) else {
+            unreachable!();
+        };
+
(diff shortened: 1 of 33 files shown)


Dunqing:
* **#15664** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15664?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15664).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_transformer/src/plugins/tagged_template_transform.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    /// Check if the template literal contains a `</script` tag; note it is case-insensitive.
+    fn contains_closing_script_tag(quasi: &TemplateLiteral) -> bool {
+        const SCRIPT_TAG: &[u8] = b"</script";
+
+        quasi.quasis.iter().any(|quasi| {
+            let raw_bytes = &quasi.value.raw.as_bytes();
+            // Get the bytes up to the last possible starting position of the script tag
+            let max_remain_len = raw_bytes.len().saturating_sub(SCRIPT_TAG.len());
+            let raw_bytes_iter = raw_bytes[..=max_remain_len].iter().copied().enumerate();
```
sapphi-red, line 87: The conformance CI is panicing here with:
```
thread '<unnamed>' (7453) panicked at crates/oxc_transformer/src/plugins/tagged_template_transform.rs:80:43:
range end index 0 out of range for slice of length 0
```

npm/runtime/src/helpers/esm/taggedTemplateLiteralEscape.js
```diff
@@ -0,0 +1,6 @@
+function _taggedTemplateLiteralEscape(e, t) {
+  return t || (t = e.slice(0)), Object.freeze(Object.defineProperty(e, "raw", {
+    value: Object.freeze(t)
+  }));
```
sapphi-red: esbuild's helper function is like this
```suggestion
  return Object.freeze(Object.defineProperty(e, "raw", {
    value: Object.freeze(t || e.slice())
  }));
```
I think we can use it instead of the babel's one as this would be slightly shorter when minified.

Dunqing: I forgot to remove this; I reused the Babel one. See https://github.com/oxc-project/oxc/blob/66df8c4b1b69e5d75a5b13ca4b215751f417e30e/npm/runtime/src/helpers/taggedTemplateLiteral.js, it is still not the same though.

sapphi-red: Ah, ok. It feels a bit verbose, but it's definitely not a blocker.

Dunqing: Yes, we have a plan in the future to revamp all helper functions to make them more performant and minifier-friendly.

overlookmotel:
@Dunqing in case you weren't aware, there's a panic in conformance.

Dunqing:
> @Dunqing in case you weren't aware, there's a panic in conformance.

Sorry, I forgot to push the local changes.

crates/oxc_transformer/src/plugins/tagged_template_transform.rs
sapphi-red: @Dunqing It seems this panic is still happening.

Dunqing: Aha, fixed now

sapphi-red (approved):
LGTM on my side

overlookmotel:
I'm afraid I don't think this is right. I'll make sure I'm correct about that, and write up what I think the problem is.

In meantime, have rebased on latest main + added one formatting nit commit.

overlookmotel:
What we're missing is a 2nd array for raw strings if `cooked` and `raw` are not the same:

Input:

```js
foo`</script>`;
foo`</script>\n`;
foo`</script>\u`; // No cooked because invalid escape `\u`
```

Current output:

```js
// Correct
foo(_temp || (_temp = _taggedTemplateLiteral(["</script>"])));
// `raw` property incorrect in `foo`
foo(_temp2 || (_temp2 = _taggedTemplateLiteral(["</script>\n"])));
// Syntax error!
foo(_temp3 || (_temp3 = _taggedTemplateLiteral(["</script>\u"])));
```

Output should be:

```js
foo(_temp || (_temp = _taggedTemplateLiteral(["</script>"])));
foo(_temp2 || (_temp2 = _taggedTemplateLiteral(["</script>\n"], ["</script>\\n"])));
//                                                            ^^^^^^^^^^^^^^^^^^
foo(_temp3 || (_temp3 = _taggedTemplateLiteral([void 0], ["</script>\\u"])));
//                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[Babel REPL](https://babeljs.io/repl#?config_lz=N4IgZglgNgpgdgQwLYxALhAJxgBygOgCsBnEAGhDwFcBzCOUtAbSZAAEAjBDmKAemp04AWgAumBAzAB7TEjEwkeBKJjCoEVRKikywAL4BdQxWLSqmAMYwAKgE8cqDEmkATKrBD6gA&code_lz=GYexAMB4HoGcGMBOBLADgFwHzgNwChQIYEUNMAdAO1wLCjiTS3IFdcACaadgORHfhgA1gFMAJuwBGI-AEMWsEe2SUAbrIA2yMXiA&version=7.28.5)

[ESBuild REPL](https://esbuild.github.io/try/#dAAwLjI1LjEyAABmb29gPC9zY3JpcHQ+YDsKZm9vYDwvc2NyaXB0PlxuYDsKZm9vYDwvc2NyaXB0Plx1YDsgLy8gTm8gY29va2VkIGJlY2F1c2UgaW52YWxpZA)

---

Because this only transforms tagged template literals containing `</script>`, I think we're OK to merge this, and fix in a follow-up. It'll only be incorrect for tagged template literals which contain *both* `</script>` and escapes e.g. `\n`, and syntax error only occurs if there's an invalid escape sequence in the template.

Additionally, is this transform enabled in benchmarks? I think it may be a perf regression. We should check if `expr` is an `Expression::TaggedTemplateExpression` in `enter_expression` and NOT add `#[inline]` to `transform_tagged_template` - because most expressions are not `TaggedTemplateExpression`s.

crates/oxc_transformer/src/plugins/tagged_template_transform.rs
```diff
@@ -0,0 +1,214 @@
...
+            let raw = &quasi.value.raw;
+
+            // The raw string must be at least as long as the script tag
+            if raw.len() < SCRIPT_TAG.len() {
+                return false;
+            }
+
+            let raw_bytes = raw.as_bytes();
+            // Get the bytes up to the last possible starting position of the script tag
+            let max_remain_len = raw_bytes.len().saturating_sub(SCRIPT_TAG.len());
```
overlookmotel, line 86: No need for `saturating_sub` here. You already checked that `len` < `SCRIPT_TAG.len()` above, so just `raw_bytes.len() - SCRIPT_TAG.len()` would do (cheaper).

crates/oxc_transformer/src/plugins/tagged_template_transform.rs
```diff
@@ -0,0 +1,214 @@
...
+            for (idx, byte) in raw_bytes_iter {
+                if byte == b'<'
+                    && SCRIPT_TAG
+                        .iter()
+                        .zip(raw_bytes[idx..].iter())
+                        .all(|(a, b)| *a == b.to_ascii_lowercase())
+                {
+                    return true;
+                }
+            }
```
overlookmotel, line 97: We have a faster way to check for `</script` in `oxc_codegen` here:

https://github.com/oxc-project/oxc/blob/3b548dc7dea2fa49c50f8d17052a2f4c74c15a2a/crates/oxc_codegen/src/str.rs#L726-L743

overlookmotel:
Actually, this transform is enabled in benchmarks. Effect is only -1%.

That's acceptable for now. We can improve it in a follow-up.

overlookmotel (approved):
We've decided to merge, and fix the bug + perf in follow-ups.

overlookmotel:
### Merge activity

* **Nov 17, 11:54 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 17, 11:54 AM UTC**: `overlookmotel` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 17, 12:01 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

Issue #15688, opened 2025-11-14
linter: `oxlint --rules --format=json` broken on main


connorshea:
### What version of Oxlint are you using?

dev

### What command did you run?

`oxlint --rules --format=json`

### What does your `.oxlintrc.json` config file look like?

<details>
<summary>config</summary>

```jsonc
{
  "$schema": "./node_modules/oxlint/configuration_schema.json",
  "plugins": [
    "import",
    "jsdoc",
    "promise",
    "react",
    "jsx-a11y"
  ],
  "categories": {
    "correctness": "off"
  },
  "env": {
    "builtin": true,
    "es2021": true,
    "browser": true,
    "shared-node-browser": true
  },
  "rules": {
    "for-direction": "error",
    "no-async-promise-executor": "error",
    "no-case-declarations": "error",
    "no-class-assign": "error",
    "no-compare-neg-zero": "error",
    "no-cond-assign": "error",
    "no-const-assign": "error",
    "no-constant-binary-expression": "error",
    "no-constant-condition": "error",
    "no-control-regex": "error",
    "no-debugger": "error",
    "no-delete-var": "error",
    "no-dupe-class-members": "error",
    "no-dupe-else-if": "error",
    "no-dupe-keys": "error",
    "no-duplicate-case": "error",
    "no-empty": [
      "error",
      {
        "allowEmptyCatch": true
      }
    ],
    "no-empty-character-class": "error",
    "no-empty-pattern": "error",
    "no-empty-static-block": "error",
    "no-ex-assign": "error",
    "no-extra-boolean-cast": "error",
    "no-fallthrough": "error",
    "no-func-assign": "error",
    "no-global-assign": "error",
    "no-import-assign": "error",
    "no-invalid-regexp": "error",
    "no-irregular-whitespace": "error",
    "no-loss-of-precision": "error",
    "no-new-native-nonconstructor": "error",
    "no-nonoctal-decimal-escape": "error",
    "no-obj-calls": "error",
    "no-prototype-builtins": "error",
    "no-redeclare": "error",
    "no-regex-spaces": "error",
    "no-self-assign": "error",
    "no-setter-return": "error",
    "no-shadow-restricted-names": "error",
    "no-sparse-arrays": "error",
    "no-this-before-super": "error",
    "no-unexpected-multiline": "error",
    "no-unsafe-finally": "error",
    "no-unsafe-negation": "error",
    "no-unsafe-optional-chaining": "error",
    "no-unused-labels": "error",
[...]

PR #15689, opened 2025-11-14
fix(linter): restores `oxlint --rules -f=json` functionality.

OtherwiseJunk:
Addresses bug https://github.com/oxc-project/oxc/issues/15688

When `self.options.output_options.format` is any value other than `OutputFormat::Default` we will revert to using the previous method for reporting `--rules`.

This preserves the `enabled?` column in the CLI while ensuring the json format is correctly output.

I have also added a test for validating that when `--rules -f=json` is invoked we return a list of JSON objects with the expected structure, without creating a test that will fail as more rules are added.

To achieve this a `test_output` method was added which returns the stdout we generate, if there is a more desirable way to test the output I would be happy to remove this method and modify the test.

### DIFF
--- a/apps/oxlint/src/lint.rs
+++ b/apps/oxlint/src/lint.rs
@@ -22,7 +22,7 @@ use oxc_linter::{
 
 use crate::{
     cli::{CliRunResult, LintCommand, MiscOptions, ReportUnusedDirectives, WarningOptions},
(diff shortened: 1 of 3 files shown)


camc314 (approved):
thank you, sorry i didn't think about this case when i merged the other Pr.

PR #15690, opened 2025-11-14
feat(minify): add directives option

xin2017338:
https://terser.org/docs/options/ 

- directives (default: true) -- remove redundant or non-standard directives

When the js module is of the Module type, it may output from ejs to cjs. Therefore, we add this configuration to keep it similar to terser

It is possible to fix the related issues

https://github.com/rolldown/rolldown/issues/6880
https://github.com/rolldown/rolldown/issues/6879

### DIFF
--- a/crates/oxc_minifier/src/options.rs
+++ b/crates/oxc_minifier/src/options.rs
@@ -54,6 +54,9 @@ pub struct CompressOptions {
 
     /// Limit the maximum number of iterations for debugging purpose.
     pub max_iterations: Option<u8>,
+
+    /// Remove redundant or non-standard directives
+    pub directives: bool,
 }
 
 impl Default for CompressOptions {
@@ -75,6 +78,7 @@ impl CompressOptions {
             treeshake: TreeShakeOptions::default(),
             drop_labels: FxHashSet::default(),
             max_iterations: None,
+            directives: true,
         }
     }
 
@@ -90,6 +94,7 @@ impl CompressOptions {
             treeshake: TreeShakeOptions::default(),
             drop_labels: FxHashSet::default(),
             max_iterations: None,
+            directives: true,
         }
     }
 
@@ -105,6 +110,7 @@ impl CompressOptions {
             treeshake: TreeShakeOptions::default(),
             drop_labels: FxHashSet::default(),
             max_iterations: None,
+            directives: true,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


sapphi-red:
> When the js module is of the Module type, it may output from ejs to cjs. Therefore, we add this configuration to keep it similar to terser

I don't think so. If the code is not ESM, then the code should be parsed with Script mode, which makes `node.source_type.is_module()` to return `false`.

xin2017338:
You're right. There is also a problem with detecting module type here

https://repl.rolldown.rs/#eNptjkEKwzAMBL8idPEl+AEu/YkvwbFLiisFW24DwX+v0tDSQy5adiVmtWFCt+FMU1xtuNfdELq/YMCg3mOrEaqUOYjHiydPcV24SLWJGa5gVIwnPY/opLTYByyc88QvsoEpzTcrP/zJ5ug5oDDFNLYssPniCYCbLE3c1wIkLo9RA6MfmiPsu3wGdu1+KiyPEqtgfwNltE1r

output

{ span: Span { start: 0, end: 35 }, source_text: "\"use strict\";\n\nexports.foo = 'foo'\n", comments: Vec([]), hashbang: None, directives: Vec([Directive { span: Span { start: 0, end: 13 }, expression: StringLiteral { span: Span { start: 0, end: 12 }, value: "use strict", raw: Some("\"use strict\""), lone_surrogates: false }, directive: "use strict" }]), body: Vec([ExpressionStatement(ExpressionStatement { span: Span { start: 15, end: 34 }, expression: AssignmentExpression(AssignmentExpression { span: Span { start: 15, end: 34 }, left: StaticMemberExpression(StaticMemberExpression { span: Span { start: 15, end: 26 }, object: Identifier(IdentifierReference { span: Span { start: 15, end: 22 }, name: "exports", reference_id: Cell { value: Some(ReferenceId(0)) } }), property: IdentifierName { span: Span { start: 23, end: 26 }, name: "foo" }, optional: false }), right: StringLiteral(StringLiteral { span: Span { start: 29, end: 34 }, value: "foo", raw: Some("'foo'"), lone_surrogates: false }), operator: Assign }) })]), scope_id: Cell { value: Some(ScopeId(0)) }, source_type: SourceType { language: JavaScript, module_kind: Module, variant: Standard } }

source_type module_kind = Module

xin2017338:
At present, it seems that adding this configuration is still quite necessary

Boshen:
> 
At present, it seems that adding this configuration is still quite necessary

can you elaborate your use case? I try to avoid adding options until it's necessary for a specific modern usecase.

sapphi-red:
I think #15691 solved the same problem without a new option.

PR #15703, opened 2025-11-14
feat(linter/import): implement no-extraneous-dependencies

HaukeSchnau:
## Summary
- port ESLint’s `import/no-extraneous-dependencies` rule into `oxc_linter`, resolving the nearest `package.json`, honoring includeInternal/includeTypes toggles, glob-based allowances for dev/optional/peer/bundled dependencies, and skipping Node built-ins/internal modules by regex
- accept `packageDir` as either a string or an array to stay compatible with ESLint config, add the required `fast-glob` dependency, and register the rule in the generated runner and rule list
- add fixtures plus lint snapshots so coverage captures the new rule and document the upstream rule reference inside the implementation

## Testing
- [ ] `cargo test -p oxc_linter import::no-extraneous-dependencies`
- [ ] `just test`


crates/oxc_linter/src/rules/import/no_extraneous_dependencies.rs
```diff
@@ -0,0 +1,776 @@
...
+        Argument, CallExpression, ExportAllDeclaration, ExportNamedDeclaration, Expression,
+        ImportDeclaration, ImportDeclarationSpecifier, ImportExpression, ImportOrExportKind,
+        StringLiteral, TSImportEqualsDeclaration, TSModuleReference,
+    },
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_resolver::NODEJS_BUILTINS;
+use oxc_span::Span;
+use regex::Regex;
```
connorshea: I believe lazy regex is always preferred over regex in the oxlint codebase.

HaukeSchnau: Thanks for the heads up, I've replaced the regex usages with lazy regex :)

tongyliu:
hi @HaukeSchnau or @camc314 -- is there an ETA on getting this landed?

crates/oxc_linter/src/rules/import/no_extraneous_dependencies.rs
```diff
@@ -0,0 +1,838 @@
...
+            None => {
+                current = dir.parent();
+            }
+        }
+    }
+
+    Ok(None)
+}
+
+fn read_package_json(path: &Path) -> Result<Option<DependencySets>, PackageJsonError> {
```
KuSh, line 614: This would help implementing other rules and should be made available as an utils.

rules that could benefit from package.json parsing:
- eslint-plugin-unicorn/expiring-todo-comments
- [jest/no-deprecated-functions](https://github.com/oxc-project/oxc/blob/1aec3368fac8c557b52bfd6a4220fef961a63f13/crates/oxc_linter/src/rules/jest/no_deprecated_functions.rs#L120)

Also our own sniff_for_frameworks function [refers to](https://github.com/oxc-project/oxc/blob/1aec3368fac8c557b52bfd6a4220fef961a63f13/crates/oxc_linter/src/context/host.rs#L471) searching in package.json

HaukeSchnau: Done :)

camchenry:
Hi, thank you for the contribution. I am going to close this for now. I think we'll want to discuss how to resolve `package.json` as a team as well so we don't have a major performance regression here.

Issue #14829, opened 2025-10-20
Linter plugins: Token-related `SourceCode` APIs


overlookmotel:
As noted in JS Plugins Milestone 2 (#14826), our first priority is to implement all APIs related to "fix code"-type linter plugins.

Later on, we'll want to also support stylistic/formatting rules, which rely on `Token`-related APIs in `SourceCode`. These methods are currently stubbed out in [tokens.ts](https://github.com/oxc-project/oxc/blob/b8f8ce5ec531728f3599d824b1abcaa4fc708c8f/apps/oxlint/src-js/plugins/tokens.ts).

## JS implementation

Unlike many other APIs, I am unclear if it's feasible to implement the `Token`-related APIs by pulling in a JS dependency.

[@typescript-eslint/parser](https://typescript-eslint.io/packages/parser/) does provide an array of `Token`s from its parser, but it uses TypeScript's parser under the hood. Parsing every file again on JS side using TypeScript would be *really* slow.

[espree](https://www.npmjs.com/package/espree) (ESLint's parser) has a `tokenize()` method. That doesn't parse, only tokenizes, so its performance might be acceptable (according to Claude, anyway). However, will it work for TS? Espree can't parse TS, but maybe that doesn't matter when you're only asking it to tokenize? (Claude says yes, but I don't entirely believe him!)

If Espree *can* work for us, I think this should be our first port of call, in interests of getting a working implementation out quickly.

#### Laziness

Like we do with comments, we'd want to lazily call `tokenize()`, only when one of the token-related APIs is first used, so that this work is skipped when no JS rule requires tokens.

## Rust implementation

This will be what we'll want to do in the end, as it'll be faster. However, it's challenging.

### Altering the lexer

Oxc's [lexer](https://github.com/oxc-project/oxc/tree/main/crates/oxc_parser/src/lexer) generates a stream of `Token`s, however, they're not stored anywhere. We'd need to adapt it to also store all the `Token`s in an `oxc_allocator::Vec` in the arena. That `Vec` can then be shared with JS through the same "raw transfer" mechanism as is used for the AST.

One difficulty is that the lexer sometimes has to unwind to where syntax is ambiguous reading forwards through the file. e.g. is `(a, b)` a `SequenceExpression`, or the start of an arrow function `(a, b) => ...`? When rewinding occurs, we'd need to also rewind the stream of `Token`s so they don't get included twice in these cases.

### Avoiding perf hit

[...]

PR #15861, opened 2025-11-19
feat(linter/plugins): Token-related `SourceCode` APIs (TS ESLint implementation)

lilnasy:
- "Slow but working" implementation for https://github.com/oxc-project/oxc/issues/14829#issuecomment-3512679496
- Uses the parser from `@typescript-eslint/typescript-estree` for tokens. The source code is fully parsed and everything but the tokens and comments is discarded.
- Tests are directly adapted from `eslint`.
- Direct commits from maintainers welcome.

### Tasks
- [x] `getTokens`
- [x] Move token types to `src-js/plugins/tokens.ts`
- [x] Bundle typescript, handling `_filename` usage in ES modules.
- [x] ~[Prevent class name mangling by `tsdown`](https://github.com/oxc-project/oxc/pull/15861#discussion_r2542816947). Could be deferred.~ https://github.com/oxc-project/oxc/pull/15906#pullrequestreview-3487271352
- [x] [Test regex-divide ambiguous syntax](https://github.com/oxc-project/oxc/pull/15861#discussion_r2542919409).
- [x] Test for conformance directly against `eslint` tests.
- [ ] [Parse as JSX only if source text is JSX](https://github.com/oxc-project/oxc/pull/15861#discussion_r2542850640). Could be deferred.
- [x] Lazy load the parser on demand. #16011

### Future work
- [x] `getFirstToken` #16002 
- [x] `getFirstTokens`#15976
- [x] `getLastToken` #16003
- [x] `getLastTokens` #16000 
- [x] `getTokenBefore` #15956 
- [x] `getTokenOrCommentBefore` #16044 
- [x] `getTokensBefore` #15955 
- [x] `getTokenAfter` #15973 
- [x] `getTokenOrCommentAfter` #16045 
- [x] `getTokensAfter` #15971 
- [x] `getTokensBetween` #16034 
- [x] `getFirstTokenBetween` #16032 
- [x] `getFirstTokensBetween` #16019 
- [x] `getLastTokenBetween` #16008 
- [x] `getLastTokensBetween` #16033 
- [x] `getTokenByRangeStart` #16043 
- [x] `isSpaceBetween` #16055
- [x] `isSpaceBetweenTokens` #16055

### Decisions
- [x] `@typescript-eslint/typescript-estree` peer-depends on `typescript`. How should we package it? Currently, `typescript` is being made an optional dependency.
  - **overlookmotel (on discord): ideally bundled, direct runtime dependency otherwise.**
- [x] Deprecated methods are being removed altogether in ESLint 10: `getTokenOrCommentBefore`, `getTokenOrCommentAfter`, and `isSpaceBetweenTokens`. These are surface level deprecations: the functionality was merged with other methods (the `includeComments: true` option) and plugins can migrate with a one line change. I'm guessing we are targeting fairly modern, actively developed projects. Should we expose them?
  - **[overlookmotel and camc314](https://github.com/oxc-project/oxc/pull/15861#issuecomment-3552651295): yes.**
- [x] `Program` range is different between ESLint and TS-ESLint. Which one should we follow?
  - **[overlookmotel](https://github.com/oxc-project/oxc/pull/15861#discussion_r2547915407): they are going to both change and align soon, we should update our range calculation to match when they do.**

### DIFF
--- a/apps/oxlint/package.json
+++ b/apps/oxlint/package.json
@@ -20,9 +20,10 @@
     "node": "^20.19.0 || >=22.12.0"
   },
   "devDependencies": {
-    "@typescript-eslint/scope-manager": "8.46.2",
     "@types/esquery": "^1.5.4",
     "@types/estree": "^1.0.8",
+    "@typescript-eslint/scope-manager": "8.46.2",
+    "@typescript-eslint/typescript-estree": "^8.47.0",
     "eslint": "^9.36.0",
     "esquery": "^1.6.0",
     "execa": "^9.6.0",
@@ -30,7 +31,7 @@
     "rolldown": "catalog:",
     "tsdown": "catalog:",
     "type-fest": "^5.2.0",
-    "typescript": "catalog:",
+    "typescript": "5.9.3",
     "vitest": "catalog:"
   },
   "napi": {
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -46,7 +46,9 @@ export interface RangeOptions {
  */
 export type FilterFn = (token: Token) => boolean;
 
-// AST token type.
+/**
+ * AST token type.
+ */
 export type Token =
   | BooleanToken
   | CommentToken
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -146,7 +146,7 @@ function initTokens() {
     sourceType: 'module',
     tokens: true,
     comment: true,
-    // TODO: enable jsx only when needed.
+    // TODO: Enable JSX only when needed
     jsx: true,
   }));
 }
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -173,7 +173,6 @@ export function resetTokens() {
  * @param afterCount? - The number of tokens after the node to retrieve.
  * @returns Array of `Token`s.
  */
-/* oxlint-disable no-unused-vars */
 export function getTokens(
   node: Node,
   countOptions?: CountOptions | number | FilterFn | null,
@@ -276,7 +275,6 @@ export function getTokens(
 
   return nodeTokens;
 }
-/* oxlint-enable no-unused-vars */
 
 /**
  * Get the first token of the given node.
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -1,11 +1,12 @@
 /*
  * `SourceCode` methods related to tokens.
  */
-
+import { parse } from '@typescript-eslint/typescript-estree';
 import { sourceText, initSourceText } from './source_code.js';
 import { assertIsNonNull } from '../utils/asserts.js';
 
-import type { Comment, Node, NodeOrToken, Token } from './types.ts';
+import type { Comment, Node, NodeOrToken } from './types.ts';
+import type { Span } from './location.ts';
 
 /**
  * Options for various `SourceCode` methods e.g. `getFirstToken`.
@@ -44,6 +45,118 @@ export interface RangeOptions {
  */
 export type FilterFn = (token: Token) => boolean;
 
+// AST token type.
+export type Token =
+  | BooleanToken
+  | CommentToken
+  | IdentifierToken
+  | JSXIdentifierToken
+  | JSXTextToken
+  | KeywordToken
+  | NullToken
+  | NumericToken
+  | PrivateIdentifierToken
+  | PunctuatorToken
+  | RegularExpressionToken
+  | StringToken
+  | TemplateToken;
+
+interface BaseToken extends Omit<Span, 'start' | 'end'> {
+  type: Token['type'];
+  value: string;
+}
+
+export interface BooleanToken extends BaseToken {
+  type: 'Boolean';
+}
+
+export type CommentToken = BlockCommentToken | LineCommentToken;
+
+export interface BlockCommentToken extends BaseToken {
+  type: 'Block';
+}
+
+export interface LineCommentToken extends BaseToken {
+  type: 'Line';
+}
+
+export interface IdentifierToken extends BaseToken {
+  type: 'Identifier';
+}
+
+export interface JSXIdentifierToken extends BaseToken {
+  type: 'JSXIdentifier';
+}
+
+export interface JSXTextToken extends BaseToken {
+  type: 'JSXText';
+}
+
+export interface KeywordToken extends BaseToken {
+  type: 'Keyword';
+}
+
+export interface NullToken extends BaseToken {
+  type: 'Null';
+}
+
+export interface NumericToken extends BaseToken {
+  type: 'Numeric';
+}
+
+export interface PrivateIdentifierToken extends BaseToken {
+  type: 'PrivateIdentifier';
+}
+
+export interface PunctuatorToken extends BaseToken {
+  type: 'Punctuator';
+}
+
+export interface RegularExpressionToken extends BaseToken {
+  type: 'RegularExpression';
+  regex: {
+    flags: string;
+    pattern: string;
+  };
+}
+
+export interface StringToken extends BaseToken {
+  type: 'String';
+}
+
+export interface TemplateToken extends BaseToken {
+  type: 'Template';
+}
+
+// Tokens for the current file parsed by TS-ESLint.
+// Created lazily only when needed.
+let tokens: Token[] | null = null;
+let comments: CommentToken[] | null = null;
+let tokensWithComments: Token[] | null = null;
+
+/**
+ * Initialize TS-ESLint tokens for current file.
+ */
+function initTokens() {
+  assertIsNonNull(sourceText);
+  ({ tokens, comments } = parse(sourceText, {
+    sourceType: 'module',
+    tokens: true,
+    comment: true,
+    // TODO: enable jsx only when needed.
+    jsx: true,
+  }));
+}
+
+/**
+ * Discard TS-ESLint tokens to free memory.
+ */
+export function resetTokens() {
+  tokens = null;
+  comments = null;
+  tokensWithComments = null;
+}
+
 /**
  * Get all tokens that are related to the given node.
  * @param node - The AST node.
@@ -63,7 +176,103 @@ export function getTokens(
   countOptions?: CountOptions | number | FilterFn | null,
   afterCount?: number | null,
 ): Token[] {
-  throw new Error('`sourceCode.getTokens` not implemented yet'); // TODO
+  if (tokens === null) initTokens();
+
+  assertIsNonNull(tokens);
+  assertIsNonNull(comments);
+
+  /**
+   * Maximum number of tokens to return.
+   */
+  const count = typeof countOptions === 'object' && countOptions !== null ? countOptions.count : null;
+
+  /**
+   * Number of preceding tokens to additionally return.
+   */
+  const beforeCount = typeof countOptions === 'number' ? countOptions : 0;
+
+  /**
+   * Number of following tokens to additionally return.
+   */
+  afterCount =
+    (typeof countOptions === 'number' || typeof countOptions === 'undefined') && typeof afterCount === 'number'
+      ? afterCount
+      : 0;
+
+  /**
+   * Function to filter tokens.
+   */
+  const filter =
+    typeof countOptions === 'function'
+      ? countOptions
+      : typeof countOptions === 'object' && countOptions !== null
+        ? countOptions.filter
+        : null;
+
+  /**
+   * Whether to return comment tokens.
+   */
+  const includeComments =
+    typeof countOptions === 'object' &&
+    countOptions !== null &&
+    'includeComments' in countOptions &&
+    countOptions.includeComments;
+
+  /**
+   * Source array of tokens to search in.
+   */
+  let nodeTokens: Token[] | null = null;
+  if (includeComments) {
+    if (tokensWithComments === null) {
+      tokensWithComments = [...tokens, ...comments].sort((a, b) => a.range[0] - b.range[0]);
+    }
+    nodeTokens = tokensWithComments;
+  } else {
+    nodeTokens = tokens;
+  }
+
+  let sliceStart = nodeTokens.length;
+  let sliceEnd: number | undefined = undefined;
+
+  const { range } = node,
+    rangeStart = range[0],
+    rangeEnd = range[1];
+
+  // Binary search for first token within `node`'s range.
+  for (let lo = 0, hi = nodeTokens.length; lo < hi; ) {
+    const mid = (lo + hi) >> 1;
+    if (nodeTokens[mid].range[0] < rangeStart) {
+      lo = mid + 1;
+    } else {
+      sliceStart = hi = mid;
+    }
+  }
+
+  // Binary search for the first token outside `node`'s range.
+  for (let lo = sliceStart, hi = nodeTokens.length; lo < hi; ) {
+    const mid = (lo + hi) >> 1;
+    if (nodeTokens[mid].range[0] < rangeEnd) {
+      lo = mid + 1;
+    } else {
+      sliceEnd = hi = mid;
+    }
+  }
+
+  sliceStart = Math.max(0, sliceStart - beforeCount);
+  // `sliceEnd` would remain undefined here if the node contains the last token of the file.
+  if (sliceEnd !== undefined) sliceEnd += afterCount;
+
+  nodeTokens = nodeTokens.slice(sliceStart, sliceEnd);
+
+  // Logically, filter must remain before count.
+  if (filter) {
+    nodeTokens = nodeTokens.filter(filter);
+  }
+  if (typeof count === 'number' && count < nodeTokens.length) {
+    nodeTokens = nodeTokens.slice(0, count);
+  }
+
+  return nodeTokens;
 }
 /* oxlint-enable no-unused-vars */
 
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -9,6 +9,8 @@ import { assertIsNonNull } from '../utils/asserts.js';
 import type { Comment, Node, NodeOrToken } from './types.ts';
 import type { Span } from './location.ts';
 
+const { max } = Math;
+
 /**
  * Options for various `SourceCode` methods e.g. `getFirstToken`.
  */
@@ -247,7 +249,7 @@ export function getTokens(
     }
   }
 
-  sliceStart = Math.max(0, sliceStart - beforeCount);
+  sliceStart = max(0, sliceStart - beforeCount);
   // `sliceEnd` would remain undefined here if the node contains the last token of the file
   if (sliceEnd !== undefined) sliceEnd += afterCount;
 
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -163,7 +163,7 @@ export function resetTokens() {
 /**
  * Get all tokens that are related to the given node.
  * @param node - The AST node.
- * @param countOptions? - Options object. If this is a function then it's `options.filter`.
+ * @param countOptions? - Options object. If this is a function then it's `countOptions.filter`.
  * @returns Array of `Token`s.
  */
 /**
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -182,27 +182,19 @@ export function getTokens(
   assertIsNonNull(tokens);
   assertIsNonNull(comments);
 
-  /**
-   * Maximum number of tokens to return.
-   */
+  // Maximum number of tokens to return
   const count = typeof countOptions === 'object' && countOptions !== null ? countOptions.count : null;
 
-  /**
-   * Number of preceding tokens to additionally return.
-   */
+  // Number of preceding tokens to additionally return
   const beforeCount = typeof countOptions === 'number' ? countOptions : 0;
 
-  /**
-   * Number of following tokens to additionally return.
-   */
+  // Number of following tokens to additionally return
   afterCount =
     (typeof countOptions === 'number' || typeof countOptions === 'undefined') && typeof afterCount === 'number'
       ? afterCount
       : 0;
 
-  /**
-   * Function to filter tokens.
-   */
+  // Function to filter tokens
   const filter =
     typeof countOptions === 'function'
       ? countOptions
@@ -210,18 +202,14 @@ export function getTokens(
         ? countOptions.filter
         : null;
 
-  /**
-   * Whether to return comment tokens.
-   */
+  // Whether to return comment tokens
   const includeComments =
     typeof countOptions === 'object' &&
     countOptions !== null &&
     'includeComments' in countOptions &&
     countOptions.includeComments;
 
-  /**
-   * Source array of tokens to search in.
-   */
+  // Source array of tokens to search in
   let nodeTokens: Token[] | null = null;
   if (includeComments) {
     if (tokensWithComments === null) {
@@ -239,7 +227,7 @@ export function getTokens(
     rangeStart = range[0],
     rangeEnd = range[1];
 
-  // Binary search for first token within `node`'s range.
+  // Binary search for first token within `node`'s range
   for (let lo = 0, hi = nodeTokens.length; lo < hi; ) {
     const mid = (lo + hi) >> 1;
     if (nodeTokens[mid].range[0] < rangeStart) {
@@ -249,7 +237,7 @@ export function getTokens(
     }
   }
 
-  // Binary search for the first token outside `node`'s range.
+  // Binary search for the first token outside `node`'s range
   for (let lo = sliceStart, hi = nodeTokens.length; lo < hi; ) {
     const mid = (lo + hi) >> 1;
     if (nodeTokens[mid].range[0] < rangeEnd) {
@@ -260,12 +248,12 @@ export function getTokens(
   }
 
   sliceStart = Math.max(0, sliceStart - beforeCount);
-  // `sliceEnd` would remain undefined here if the node contains the last token of the file.
+  // `sliceEnd` would remain undefined here if the node contains the last token of the file
   if (sliceEnd !== undefined) sliceEnd += afterCount;
 
   nodeTokens = nodeTokens.slice(sliceStart, sliceEnd);
 
-  // Logically, filter must remain before count.
+  // Filter before limiting by `count`
   if (filter) {
     nodeTokens = nodeTokens.filter(filter);
   }
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -1,6 +1,7 @@
 /*
  * `SourceCode` methods related to tokens.
  */
+
 import { parse } from '@typescript-eslint/typescript-estree';
 import { sourceText, initSourceText } from './source_code.js';
 import { assertIsNonNull } from '../utils/asserts.js';
@@ -177,7 +178,6 @@ export function getTokens(
   afterCount?: number | null,
 ): Token[] {
   if (tokens === null) initTokens();
-
   assertIsNonNull(tokens);
   assertIsNonNull(comments);
 
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -254,12 +254,8 @@ export function getTokens(
   nodeTokens = nodeTokens.slice(sliceStart, sliceEnd);
 
   // Filter before limiting by `count`
-  if (filter) {
-    nodeTokens = nodeTokens.filter(filter);
-  }
-  if (typeof count === 'number' && count < nodeTokens.length) {
-    nodeTokens = nodeTokens.slice(0, count);
-  }
+  if (filter) nodeTokens = nodeTokens.filter(filter);
+  if (typeof count === 'number' && count < nodeTokens.length) nodeTokens = nodeTokens.slice(0, count);
 
   return nodeTokens;
 }
--- a/apps/oxlint/src-js/plugins/types.ts
+++ b/apps/oxlint/src-js/plugins/types.ts
@@ -7,6 +7,7 @@ export interface Visitor {
 */
 
 import type { Span } from './location.ts';
+import type { Token } from './tokens.ts';
 
 import type { VisitorObject as Visitor } from '../generated/visitor.d.ts';
 export type { Visitor };
@@ -29,12 +30,6 @@ export type VisitFn = (node: Node) => void;
 
 // AST node type.
 export interface Node extends Span {}
-
-// AST token type.
-export interface Token extends Span {
-  type: string;
-  value: string;
-}
 
 // Currently we only support `Node`s, but will add support for `Token`s later.
 export type NodeOrToken = Node | Token;
--- a/apps/oxlint/test/fixtures/sourceCode_token_methods/files/eslint-test-case.js
+++ b/apps/oxlint/test/fixtures/sourceCode_token_methods/files/eslint-test-case.js
@@ -1,0 +1,3 @@
+/*A*/ var answer /*B*/ = /*C*/ a /*D*/ * b; /*E*/ //F
+call();
+/*Z*/
--- a/apps/oxlint/test/fixtures/sourceCode_token_methods/files/index.tsx
+++ b/apps/oxlint/test/fixtures/sourceCode_token_methods/files/index.tsx
@@ -1,0 +1,61 @@
+// CommentToken (Line)
+/* CommentToken (Block) */
+
+// BooleanToken
+true;
+false;
+
+// NullToken
+null;
+let undefined!: void;
+
+// NumericToken
+123;
+3.14;
+
+// StringToken
+('string');
+
+// TemplateToken
+tagged`template ${'literal'}`;
+
+// RegularExpressionToken
+/pattern/g;
+// prettier-ignore
+// Not a RegularExpressionToken
+1 /not_a_regex/gu;
+
+// IdentifierToken
+let identifier = 'value';
+
+if (identifier) {
+  for (let i = 0; i < 10; i++) {
+    ((NaN + '') as any) ** 5;
+  }
+}
+
+class MyClass extends Error {
+  // PrivateIdentifierToken
+  #private = 'field';
+  constructor() {
+    super();
+    this.#private;
+    #private in this;
+  }
+}
+
+// PunctuatorToken (operators and punctuation: =, +, -, (), {}, [], etc.)
+(false, Infinity, eval)?.('use strict') satisfies MyClass;
+[1, 2, 3];
+{
+  key: ('value');
+}
+
+type T = { [key: string]: number };
+interface I extends T {
+  x: T;
+}
+
+function JSX() {
+  return <div>Hello World</div>;
+}
--- a/apps/oxlint/test/fixtures/sourceCode_token_methods/plugin.ts
+++ b/apps/oxlint/test/fixtures/sourceCode_token_methods/plugin.ts
@@ -1,0 +1,53 @@
+import type { Plugin, Rule } from '../../../dist/index.js';
+
+const rule: Rule = {
+  create(context) {
+    const { sourceCode } = context;
+
+    return {
+      Program(program) {
+        for (const node of program.body) {
+          const tokens = sourceCode.getTokens(node);
+          context.report({
+            message:
+              `Tokens for ${node.type}:\n` +
+              tokens
+                .map(
+                  ({ type, loc, range, value }) =>
+                    `${type.padEnd(17)} ` +
+                    `loc=${loc.start.line}:${loc.start.column}-${loc.end.line}:${loc.end.column} `.padEnd(16) +
+                    `range=${range[0]}-${range[1]} `.padEnd(10) +
+                    `"${value}"`,
+                )
+                .join('\n'),
+            node,
+          });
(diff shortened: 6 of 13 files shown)


overlookmotel:
> Deprecated methods are being removed altogether in ESLint 10: `getTokenOrCommentBefore`, `getTokenOrCommentAfter`, and `isSpaceBetweenTokens`. These are surface level deprecations: the functionality was merged with other methods (the `includeComments: true` option) and plugins can migrate with a one line change.

Cameron and I had a bit of an argument about exactly this question!

We concluded in the end to keep all the deprecated methods, to maximize compatibility with older plugins, which may take some time to get updated (or in some cases, will *never* get updated). Our rationale is that most of these methods are just aliases, so no maintenance burden to keep them. The only one which is slightly different from its non-deprecated "brother" is `isSpaceBetweenTokens`, and that's pretty simple to implement - it just treats `JSXText` differently (if I remember right).

> I'm guessing we are targeting fairly modern, actively developed projects. 

That's true, but even actively developed projects may use old unmaintained plugins.

Note: I added the stubs in https://github.com/oxc-project/oxc/pull/15645. That PR also added tests which illustrate the difference in behavior between `isSpaceBetween` and `isSpaceBetweenTokens`.

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -62,7 +85,9 @@ export function getTokens(
   countOptions?: CountOptions | number | FilterFn | null | undefined,
   afterCount?: number | null | undefined,
 ): Token[] {
-  throw new Error('`sourceCode.getTokens` not implemented yet'); // TODO
+  if (tsTokens === null) initTsTokens();
+  const tokens = tsTokens.filter((token) => token.range[0] >= node.range[0] && token.range[1] <= node.range[1]);
```
overlookmotel: Please add TODO comments saying:

1. We need to handle `beforeCount` / `afterCount` / `countOptions`.
2. We should use binary search here.

apps/oxlint/src-js/plugins/types.ts
```diff
@@ -31,11 +31,84 @@ export type VisitFn = (node: Node) => void;
...
+  type: 'PrivateIdentifier';
+}
+
+export interface PunctuatorToken extends BaseToken {
+  type: 'Punctuator';
+}
+
+export interface RegularExpressionToken extends BaseToken {
+  type: 'RegularExpression';
+}
```
overlookmotel:
```suggestion
export interface RegularExpressionToken extends BaseToken {
  type: 'RegularExpression';
  regex: { flags: string, pattern: string };
}
```

I happened to notice the extra `regex` field in Espree's tokens when I was scoping out how to implement tokens.

Confirmed TS-ESLint parser has it too: [AST explorer](https://ast-explorer.dev/#eNolyEEKgCAQBdCryF8X7j1Iq9lMMoRhKipBiHdvouV7AxEOJ9/cfA2lY0HR6E+RP1ZpMaTvvb7l3dtDkRWDkjEE8RdvUlvIieA0IndpnUBpYr4FUx+P)

It looks like `RegularExpression` tokens are the only ones with any extra properties:

https://github.com/dbarabashh/typescript-eslint/blob/28cf8032c2492bb3c55dd7dd145249f2246034ad/packages/typescript-estree/src/node-utils.ts#L567-L611

apps/oxlint/tsdown.config.ts
```diff
@@ -14,9 +14,10 @@ const commonConfig: UserConfig = {
     // Files copied from `oxc-parser`.
     // Not bundled, to avoid needing sourcemaps when debugging.
     /^\.\.?\/.*\/dist\//,
+    'typescript',
   ],
   fixedExtension: false,
-  noExternal: ['@typescript-eslint/scope-manager'],
+  noExternal: ['@typescript-eslint/typescript-estree', '@typescript-eslint/scope-manager'],
```
overlookmotel: This line isn't required. See #15872.

oxfmtrc.jsonc
```diff
@@ -6,9 +6,11 @@
     "**/dist/**",
     "**/tests/**",
     "**/generated/**",
-    // Ignore `fixtures` directories except for `apps/oxlint/test/fixtures`
     "**/fixtures/**",
+    // Ignore only the linted files in `oxlint`'s fixtures.
+    // Test plugins should be formatted.
     "!apps/oxlint/test/fixtures/**",
+    "apps/oxlint/test/fixtures/**/files",
```
overlookmotel: If possible, it'd be ideal to continue to format the files in `apps/oxlint/test/fixtures/**/files`.

If you need to stop the formatter altering the fixture file, you can add a `// prettier-ignore` comment above statements you want to "freeze", like in [this test case](https://github.com/oxc-project/oxc/blob/61c21d69e55eb7e4675f123f2c95ba992d94a85b/apps/oxlint/test/fixtures/estree/files/index.ts#L6-L15).

npm/oxlint/package.json
```diff
@@ -25,6 +25,9 @@
   "engines": {
     "node": "^20.19.0 || >=22.12.0"
   },
+  "optionalDependencies": {
+    "typescript": ">=4.8.4 <6.0.0"
+  },
```
overlookmotel: As discussed, unfortunately I think we need to promote `typescript` to a full `dependency`. We can also then tighten the version range to whatever is latest version.

apps/oxlint/tsdown.config.ts
```diff
@@ -14,9 +14,10 @@ const commonConfig: UserConfig = {
     // Files copied from `oxc-parser`.
     // Not bundled, to avoid needing sourcemaps when debugging.
     /^\.\.?\/.*\/dist\//,
+    'typescript',
```
overlookmotel: I think this line not required either, once `typescript` is in `dependencies` in `package.json`.

apps/oxlint/test/fixtures/scope_manager/output.snap.md
```diff
@@ -3,8 +3,8 @@
 
 # stdout
 ```
-  x scope-manager-plugin(scope): File has 12 scopes: <GlobalScope>, <ModuleScope>, topLevelFunction, <BlockScope>, TopLevelModule, GenericInterface, TestClass, <ClassStaticBlockScope>,
-  | <FunctionScope>, <FunctionScope>, <BlockScope>, <BlockScope>
+  x scope-manager-plugin(scope): File has 12 scopes: <GlobalScope>, <ModuleScope>, topLevelFunction, <BlockScope$1>, TopLevelModule, GenericInterface, TestClass, <ClassStaticBlockScope>,
+  | <FunctionScope$1>, <FunctionScope$1>, <BlockScope$1>, <BlockScope$1>
```
overlookmotel: Weird that this apparently unrelated change occurs. I tried bumping `@typescript-eslint/scope-manager` dep to v8.47.0 in isolation, and it didn't change this snapshot.

Any idea what's going on?

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -43,6 +43,29 @@ export interface RangeOptions {
...
+let tsTokens: Token[] | null = null;
+
+/**
+ * Initialize TS-ESLint tokens for current file.
+ */
+function initTsTokens() {
+  const ast = parse(sourceText, {
+    sourceType: 'module',
+    tokens: true,
+    jsx: true,
```
overlookmotel, line 152: `jsx` property should be set depending on whether file is JSX/TSX or plain JS/TS.

There are some edge cases where you get a different AST depending on whether file is TS or TSX (it alters the interpretation of `<` in some places). I don't actually know if it'd alter the *tokens* either way, but probably better to be safe.

Right now we don't have a way to get whether the source is JSX or not on JS side (the flag is there in the buffer, but it doesn't appear in ESTree spec, so it's not deserialized). But we can add that ability later. Please just add a TODO comment for now.

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -43,6 +43,29 @@ export interface RangeOptions {
  */
 export type FilterFn = (token: Token) => boolean;
 
+// Tokens for the current file parsed by TS-ESLint.
+// Created lazily only when needed.
+let tsTokens: Token[] | null = null;
```
overlookmotel: Nit: I suggest renaming this to `tokens` (and `initTokens`, `resetTokens`). The fact that we're using TS parser at present is an implementation detail. And hopefully we won't for long!

apps/oxlint/src-js/plugins/types.ts
```diff
@@ -31,11 +31,84 @@ export type VisitFn = (node: Node) => void;
...
+
+export interface StringToken extends BaseToken {
+  type: 'String';
+}
+
+export interface TemplateToken extends BaseToken {
+  type: 'Template';
+}
```
overlookmotel: I suggest moving all these types to `src-js/plugins/tokens.ts`. It was my mistake originally to put loads of type defs in one "everything and the kitchen sink" `types.ts` file, and I've been slowly moving them all out to other files. Please forgive me - I'd never used TypeScript before last month, so I'm still learning...

If we want to expose these types publicly (I assume we do?), need to also re-export them from `src-js/index.ts`.

apps/oxlint/test/fixtures/sourceCode_token_methods/files/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+3.14;
+
+// StringToken
+"string";
+
+// TemplateToken
+tagged`template ${'literal'}`;
+
+// RegularExpressionToken
+/pattern/g;
```
overlookmotel, line 23: I suggest also adding this weirdo:

```js
1 /not_a_regex/gu;
```

That should be tokenized as `1`, `/`, `not_a_regex`, `/`, `gu`, `;` (not `1`, `/not_a_regex/gu`, `;`).

[AST explorer](https://ast-explorer.dev/#eNo9yUEKwjAQheGrhFkXitt6DXEVKCE+SjXNhMxElNK7O1Lt7r3/WynRQPfwDBLrXJQ6KhYgpQJ2op2T6zPrGMaKCa9+amcDNlh9ds6TcKsRl3eBp8H+wreWbHc7Iy7hiioz591TUIgenjh+u9aGX1F+IMs/+rzR9gGTVTYR)

apps/oxlint/test/fixtures/sourceCode_token_methods/plugin.ts
```diff
@@ -0,0 +1,24 @@
...
+  create(context) {
+    const { sourceCode } = context;
+
+    return {
+      Program(node) {
+        const tokens = sourceCode.getTokens(node);
+        context.report({
+          message: JSON.stringify(tokens, null, 2),
+          node,
+        });
```
overlookmotel: To make the snapshot easier to verify by eye, I suggest looping through all tokens, and making a separate `context.report` call for each. Then each token will be underlined in the code, alongside the token's details.

overlookmotel (commented):
Possibly premature review, as it's still marked as draft, but I'm keen to get it merged so I thought I'd go through it now.

Apart from the comments below, this looks good to me. Sorry for the volume of comments - most are pretty small details.

Once we're happy with `getTokens` impl, I think we should merge this, and we can add more methods in separate PRs. No need to do the whole API in a single PR.

A couple of points which we can also leave to follow-ups:

1. We should ideally lazy-load `@typescript-eslint/typescript-estree` package only when `getTokens` is first called.

2. I assume TS-ESLint's parser also generates a `ScopeManager`. If it does, we may as well cache it, to avoid running scope analysis again if plugin does `sourceCode.getTokens()` followed by getting `sourceCode.scopeManager`.

apps/oxlint/src-js/plugins/tokens.ts
lilnasy: What I had in mind was that we would soon include an `espree`-powered pipeline alongside it, and `tsTokens` would distinguish it from `espreeTokens`.

apps/oxlint/test/fixtures/sourceCode_token_methods/plugin.ts
lilnasy: The reason it is the way it is now is that I am manually copying the same JSON output from a `typescript-eslint` plugin with the same source text. It's like a makeshift conformance test until I can figure out how to create a solid testing foundation.

apps/oxlint/test/fixtures/sourceCode_token_methods/files/index.tsx
lilnasy: https://github.com/oxc-project/oxc/pull/15861/commits/176a1843ecb88fcc01b07f655d089a4a29206273
The ambiguity is being undone by autofix CI. I'll revisit it at a later point.

apps/oxlint/test/fixtures/scope_manager/output.snap.md
lilnasy: I narrowed it down to `rolldown` but I don't know the conditions that bring this change. The relevant options exposed in `tsdown` don't help. I'll review the `rolldown` code sometime soon.

lilnasy: A dependency of `@typescript-eslint/typescript-estree`, `ts-api-utils` also has classes by these names: https://github.com/JoshuaKGoldberg/ts-api-utils/blob/4b3d274bc2dcc73b14b286e1dd1ad120453a4c14/src/usage/scopes.ts#L295. `rolldown` is disambiguating them to be safe.

apps/oxlint/src-js/plugins/tokens.ts
overlookmotel: Ah ha I see. I was assuming we'll *either* generate tokens with TS-ESLint *or* Espree, so either would go into the same var. But maybe I misunderstood.

Anyway, it was just a naming suggestion, not important.

apps/oxlint/test/fixtures/scope_manager/output.snap.md
lilnasy: This will probably sort itself out when the parser is lazy loaded from a separate chunk.

overlookmotel: Ah I see. Ideally we don't want it to be possible for e.g. updating TSDown to break our tests because e.g. it outputs code in a different order, so which class gets called `BlockScope` and which gets called `BlockScope$1` randomly changes.

Please see #15906 for a potential solution (if it's workable).

apps/oxlint/test/fixtures/sourceCode_token_methods/plugin.ts
overlookmotel: Ah ha. Makes sense. Maybe do both? Or verify it's correct using the method you're using, and then switch to what I suggested before we merge this PR?

(maybe that's what you had in mind anyway)

lilnasy:
> 2. I assume TS-ESLint's parser also generates a `ScopeManager`. If it does, we may as well cache it, to avoid running scope analysis again if plugin does `sourceCode.getTokens()` followed by getting `sourceCode.scopeManager`.

`@typescript-eslint/parser` does. `@typescript-eslint/typescript-estree` does not.

The former delegates to the latter for parsing and lexing. I decided to use `@typescript-eslint/typescript-estree` directly to instantiate scope managers and parsers granularly.

apps/oxlint/test/fixtures/scope_manager/output.snap.md
overlookmotel: Great. Merging #15906. Hopefully this should disappear from the diff now.

overlookmotel:
> The former delegates to the latter for parsing and lexing. I decided to use `@typescript-eslint/typescript-estree` directly to instantiate scope managers and parsers granularly.

Ah ha that makes perfect sense. I had misunderstood.

apps/oxlint/test/fixtures/sourceCode_token_methods/files/eslint-test-case.js
```diff
@@ -0,0 +1 @@
+/*A*/ var answer /*B*/ = /*C*/ a /*D*/ * b; /*E*/ //F\n    call();\n/*Z*/
```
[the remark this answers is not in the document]
overlookmotel: It's a fixture!

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -63,7 +174,101 @@ export function getTokens(
...
+    countOptions.includeComments;
+
+  /**
+   * Source array of tokens to search in.
+   */
+  let nodeTokens: Token[] | null = null;
+  if (includeComments) {
+    if (tokensWithComments === null) {
+      assertIsNonNull(tokens);
+      assertIsNonNull(comments);
```
overlookmotel: I suggest moving these 2 up to directly after `if (tokens === null) initTokens();`, so it's clear why these assertions are safe. `assertIsNonNull` has no runtime cost - it's just a compile-time TS assertion, and minifier strips it out entirely - so it costs nothing to evaluate it early.

I think that should also remove the need for `assertIsNonNull(nodeTokens);` below.

Note: It's going to change to be an actual runtime assertion in tests, but will remain a no-op in release builds (#15921).

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -63,7 +174,101 @@ export function getTokens(
...
+  // `sliceEnd` would remain undefined here if the node contains the last token of the file.
+  sliceEnd = sliceEnd === undefined ? undefined : sliceEnd + afterCount;
+
+  nodeTokens = nodeTokens.slice(sliceStart, sliceEnd);
+
+  // Logically, filter must remain before count.
+  if (filter) {
+    nodeTokens = nodeTokens.filter(filter);
+  }
+  if (count && count < nodeTokens.length) {
```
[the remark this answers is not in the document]
overlookmotel: I think he right on this one.

apps/oxlint/package.json
```diff
@@ -19,17 +19,20 @@
...
   },
+  "dependencies": {
+    "typescript": "5.9.3"
+  },
   "devDependencies": {
-    "@typescript-eslint/scope-manager": "8.46.2",
     "@types/esquery": "^1.5.4",
     "@types/estree": "^1.0.8",
+    "@typescript-eslint/scope-manager": "8.46.2",
+    "@typescript-eslint/typescript-estree": "^8.47.0",
```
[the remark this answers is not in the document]
overlookmotel: Nope. `@typescript-eslint/typescript-estree` is bundled. So `devDependencies` is where it should be.

npm/oxlint/package.json
```diff
@@ -25,6 +25,9 @@
   "engines": {
     "node": "^20.19.0 || >=22.12.0"
   },
+  "dependencies": {
+    "typescript": "5.9.3"
+  },
```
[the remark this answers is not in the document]
overlookmotel: Yes, obviously not ideal to pull in such a massive dependency. But that's what we've decided to do, and it's only a stopgap until we have our own Rust version.

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -44,6 +45,116 @@ export interface RangeOptions {
...
+    // TODO: enable jsx only when needed.
+    jsx: true,
+  }));
+}
+
+/**
+ * Discard TS-ESLint tokens to free memory.
+ */
+export function resetTokens() {
+  tokens = null;
```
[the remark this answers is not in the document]
overlookmotel: The damn robot is right!

apps/oxlint/test/tokens.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    check(
+      store.getTokens(Program, {
+        filter: (t) => t.type === 'Identifier',
+      }),
+      ['answer', 'a', 'b', 'call'],
+    );
+  });
+
+  // TODO: this test is failing because of a difference in program range between `espree` and `@typescript-eslint/typescript-estree`.
+  // Unskip after integrating `espree`.
```
overlookmotel: This is a really weird one. ESLint is changing it's behavior in ESLint 10 to make the range of `Program` be `[0, <length of source text>]` (no matter whether there's leading/trailing comments/whitespace or not). TS-ESLint plans to follow suit.

https://eslint.org/blog/2025/10/whats-coming-in-eslint-10.0.0/#updates-to-program-ast-node-range-coverage
https://github.com/typescript-eslint/typescript-eslint/issues/11026#issuecomment-3421887632

Oxc will make the change too, and everyone will be aligned (at last).

So I suggest modifying these tests and unskipping them by either:

1. `const Program = { range: [5, 36] } as Node;` (to match current Espree). or
2. Manually add the extra tokens to the array below, so the tests pass.

Either is fine to do, I think. Just need to make a comment explaining why.

overlookmotel: PS Thanks for putting in the unglamorous work of upping the test coverage. Appreciate it.

overlookmotel (commented):
Apart from the small comments below, looks good!

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -63,7 +174,101 @@ export function getTokens(
...
+    if (nodeTokens[mid].range[0] < rangeEnd) {
+      lo = mid + 1;
+    } else {
+      sliceEnd = hi = mid;
+    }
+  }
+
+  sliceStart = sliceStart - beforeCount;
+  // `sliceEnd` would remain undefined here if the node contains the last token of the file.
+  sliceEnd = sliceEnd === undefined ? undefined : sliceEnd + afterCount;
```
overlookmotel: Nit:

```suggestion
  sliceStart -= beforeCount;
  // `sliceEnd` would remain undefined here if the node contains the last token of the file.
  if (sliceEnd !== undefined) sliceEnd += afterCount;
```

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -63,7 +174,101 @@ export function getTokens(
...
+
+  /**
+   * Number of preceding tokens to additionally return.
+   */
+  const beforeCount = typeof countOptions === 'number' ? countOptions : 0;
+
+  /**
+   * Number of following tokens to additionally return.
+   */
+  afterCount = typeof afterCount === 'number' ? afterCount : 0;
```
overlookmotel: According to the JSDoc comment on this function (which is lifted from ESLint), params are *either*:

1. `(countOptions: CountOptions | FilterFn)` or
2. `(beforeCount?: number, afterCount?: number)`

That would mean that `afterCount` param should be ignored if `countOptions` is an object or function.

Does ESLint's implementation follow the JSDoc comment, or not actually do what it says it does? I guess we should do whatever ESLint does, but if behavior is contrary to the doc comment, it'd be good to add a comment explaining why.

Or we could open a bug report on ESLint.

Sorry, it really is insanely anal of me to question this!

lilnasy: > Sorry, it really is insanely anal of me to question this!

YOU WORK FULL TIME ON A LINTER! It's part of the job description 😭.

Edit: `afterCount` is ignored if `countOptions` is an object or function. https://github.com/eslint/eslint/blob/v9.39.1/lib/languages/js/source-code/token-store/index.js#L237-L245

lilnasy: Corrected the behavior in e0f42635b.

apps/oxlint/test/tokens.test.ts
lilnasy: Went with 1. f75f3891b.

FTR, the actual ranges were different from what I had written in the comments earlier because I wasn't replacing `\n` in the playground.

Here are the actual ranges: [`eslint`](https://explorer.eslint.org/#eslint-explorer=IntcInN0YXRlXCI6e1widG9vbFwiOlwiYXN0XCIsXCJjb2RlXCI6e1wiamF2YXNjcmlwdFwiOlwiLypBKi92YXIgYW5zd2VyLypCKi89LypDKi9hLypEKi8qIGIvKkUqLy8vRlxcbiAgICBjYWxsKCk7XFxuLypaKi9cIixcImpzb25cIjpcIi8qKlxcbiAqIFR5cGUgb3IgcGFzdGUgc29tZSBKU09OIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgSlNPTiBjb2RlIGlzIGludGVycHJldGVkXFxuICogYnkgY2xpY2tpbmcgXFxcIkpTT05cXFwiIGluIHRoZSBoZWFkZXIgYW5kIHNlbGVjdGluZyBkaWZmZXJlbnRcXG4gKiBvcHRpb25zLlxcbiAqXFxuICogVGhpcyBleGFtcGxlIGlzIGluIEpTT05DIG1vZGUsIHdoaWNoIGFsbG93cyBjb21tZW50cy5cXG4gKi9cXG5cXG57XFxuICAgIFxcXCJrZXkxXFxcIjogW3RydWUsIGZhbHNlLCBudWxsXSxcXG4gICAgXFxcImtleTJcXFwiOiB7XFxuICAgICAgICBcXFwia2V5M1xcXCI6IFsxLCAyLCBcXFwiM1xcXCIsIDFlMTAsIDFlLTNdXFxuICAgIH1cXG59XCIsXCJtYXJrZG93blwiOlwiPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBNYXJrZG93biBoZXJlIHRvIGxlYXJuIG1vcmUgYWJvdXRcXG50aGUgc3RhdGljIGFuYWx5c2lzIHRoYXQgRVNMaW50IGNhbiBkbyBmb3IgeW91LlxcblxcblRoZSB0YWJzIGFyZTpcXG5cXG4tIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuYmUgdXNlZnVsIHRvIHVuZGVyc3RhbmQgdGhlIHN0cnVjdHVyZSBvZiB0aGUgY29kZS4gWW91XFxuY2FuIHZpZXcgdGhpcyBzdHJ1Y3R1cmUgYXMgSlNPTiBvciBpbiBhIHRyZWUgZm9ybWF0LlxcblxcbllvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgTWFya2Rvd24gY29kZSBpcyBpbnRlcnByZXRlZFxcbmJ5IGNsaWNraW5nIFxcXCJNYXJrZG93blxcXCIgaW4gdGhlIGhlYWRlciBhbmQgc2VsZWN0aW5nIGRpZmZlcmVudFxcbm9wdGlvbnMuXFxuXFxuVGhpcyBleGFtcGxlIGlzIGluIENvbW1vbk1hcmsgbW9kZS5cXG4tLT5cXG5cXG4jIEVTTGludCBNYXJrZG93biBFeGFtcGxlXFxuXFxuVGhpcyBpcyBhbiBleGFtcGxlIG9mIGEgTWFya2Rvd24gZmlsZSB0aGF0IGNhbiBiZSBwYXJzZWRcXG5ieSBFU0xpbnQuIE1hcmtkb3duIGlzIGEgc2ltcGxlIG1hcmt1cCBsYW5ndWFnZSB0aGF0IGlzXFxub2Z0ZW4gdXNlZCBmb3IgZG9jdW1lbnRhdGlvbi5cXG5cXG4jIyBGZWF0dXJlc1xcblxcbi0gTWFrZSB0aGluZ3MgKml0YWxpYyosICoqYm9sZCoqLCBvciBgY29kZWBcXG4tIENyZWF0ZSBbbGlua3NdKGh0dHBzOi8vZXNsaW50Lm9yZylcXG4tIFN1cHBvcnRzIEhUTUwgPHNwYW4gc3R5bGU9XFxcImNvbG9yOiByZWQ7XFxcIj5lbGVtZW50czwvc3Bhbj5cXG4tIExpc3RzXFxuICAtIE5lc3RlZCBsaXN0c1wiLFwiY3NzXCI6XCIvKipcXG4gKiBUeXBlIG9yIHBhc3RlIHNvbWUgQ1NTIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbiAqIHRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuICpcXG4gKiBUaGUgdGFicyBhcmU6XFxuICpcXG4gKiAtIEFTVCAtIFRoZSBBYnN0cmFjdCBTeW50YXggVHJlZSBvZiB0aGUgY29kZSwgd2hpY2ggY2FuXFxuICogICBiZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG4gKiAgIGNhbiB2aWV3IHRoaXMgc3RydWN0dXJlIGFzIEpTT04gb3IgaW4gYSB0cmVlIGZvcm1hdC5cXG4gKlxcbiAqIFlvdSBjYW4gY2hhbmdlIHRoZSB3YXkgdGhhdCB0aGUgQ1NTIGNvZGUgaXMgaW50ZXJwcmV0ZWRcXG4gKiBieSBjbGlja2luZyBcXFwiQ1NTXFxcIiBpbiB0aGUgaGVhZGVyIGFuZCBzZWxlY3RpbmcgZGlmZmVyZW50XFxuICogb3B0aW9ucy5cXG4gKi9cXG5cXG5AaW1wb3J0IHVybCgnaHR0cHM6Ly9mb250cy5nb29nbGVhcGlzLmNvbS9jc3MyP2ZhbWlseT1Sb2JvdG86d2dodEA0MDA7NzAwJmRpc3BsYXk9c3dhcCcpO1xcblxcbmJvZHkge1xcblxcdGZvbnQtZmFtaWx5OiBzYW5zLXNlcmlmO1xcbn1cXG5cXG5oMSB7XFxuXFx0Y29sb3I6ICMzMzM7XFxufVxcblxcbnAge1xcblxcdG1hcmdpbjogMWVtIDA7XFxufVwiLFwiaHRtbFwiOlwiPCFET0NUWVBFIGh0bWw%2BXFxuPCEtLVxcblR5cGUgb3IgcGFzdGUgc29tZSBIVE1MIGhlcmUgdG8gbGVhcm4gbW9yZSBhYm91dFxcbnRoZSBzdGF0aWMgYW5hbHlzaXMgdGhhdCBFU0xpbnQgY2FuIGRvIGZvciB5b3UuXFxuXFxuVGhlIHRhYnMgYXJlOlxcblxcbi0gQVNUIC0gVGhlIEFic3RyYWN0IFN5bnRheCBUcmVlIG9mIHRoZSBjb2RlLCB3aGljaCBjYW5cXG5iZSB1c2VmdWwgdG8gdW5kZXJzdGFuZCB0aGUgc3RydWN0dXJlIG9mIHRoZSBjb2RlLiBZb3VcXG5jYW4gdmlldyB0aGlzIHN0cnVjdHVyZSBhcyBKU09OIG9yIGluIGEgdHJlZSBmb3JtYXQuXFxuLS0%2BXFxuXFxuPGh0bWwgbGFuZz1cXFwiZW5cXFwiPlxcbiAgICA8aGVhZD5cXG4gICAgICAgIDxtZXRhIGNoYXJzZXQ9XFxcIlVURi04XFxcIj5cXG4gICAgICAgIDx0aXRsZT5IVE1MPC90aXRsZT5cXG4gICAgPC9oZWFkPlxcbiAgICA8Ym9keT5cXG4gICAgICAgIDxwPlRleHQ8L3A%2BXFxuICAgIDwvYm9keT5cXG48L2h0bWw%2BXCJ9LFwibGFuZ3VhZ2VcIjpcImphdmFzY3JpcHRcIixcImpzT3B0aW9uc1wiOntcInBhcnNlclwiOlwiZXNwcmVlXCIsXCJzb3VyY2VUeXBlXCI6XCJtb2R1bGVcIixcImVzVmVyc2lvblwiOlwibGF) (5-55), [`oxc`](https://playground.oxc.rs/#eNptUk2P0zAQ/SuVT2B1ZUBCQkUcYLuVVuJLHDhAOLjOJHhx7Ghm0nYV5b8zTht3D9tL4zfz5uO9GZVTG2X0R20OFlc20hHQ6E/afDD6Vhtr9FYbvdobfaeNMbsqruTnbAgvXr6votG/tFFrldRmVDjE/Bd8ZLVhHGCtmoSdZQZUm8YGeorcXzFGaZwDBfGUgmWot+CCRcs+RSrB41/PQL11UKDOxjZcny51PQJdKeRSL+HzUPTY7VNYXq5pL2nTWvUWKQ87KjgxRJK+og/TSXaUndPxB/CA8dvA5GvYDdHl0ZZSuSfgAb5bFO6Czryf7+4jo5eKrgQIZGz27g4x4QWVGbJ+5xlcio2X6dRYqT4MrdArtfldKX7sgRz6niv1Z10pJ1q1CT3kuCS7hAiOo0ggQKVS01RqkkQcwjVHhiFZkm9yuZsaGh/9rPTMOVqMQpqEdzXoPBdbbEE8VkBvXr1+K9oMBNvMh13C22CJdh5CXTaFUw/oO+llg1iaxNGysUQ7zwX9Amxry7bI8fwpjIpEzl5EBYxWzFwcdKmGFuZDjPnUis8PVCc59osnMSaeay1IgHbJnassBzTKa7ku2Tz1n+EApeg/gP6r7WC5tJkqRqewE9Mz4wC4T7ScZt4nPog1OVS+JoFn+ece5WvKeJbZEqvpP3yNSp0=) (5-61), and [`typescript-estree`](https://typescript-eslint.io/play/#ts=5.9.2&showAST=es&fileType=.tsx&code=PQKggiwG4IYE4AIYDsDOB3ApnUAhSAvKAMKQygAikICARqAKKTDABiAUAlwgMYwA2-ABQBKANztQALUhA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false) (5-61).

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -44,6 +45,118 @@ export interface RangeOptions {
...
+ */
+function initTokens() {
+  assertIsNonNull(sourceText);
+  ({ tokens, comments } = parse(sourceText, {
+    sourceType: 'module',
+    tokens: true,
+    comment: true,
+    // TODO: enable jsx only when needed.
+    jsx: true,
+  }));
```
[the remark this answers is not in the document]
lilnasy: Nope. The types are checked and they do match.

pnpm-lock.yaml
```diff
@@ -117,7 +120,7 @@ importers:
         specifier: ^5.2.0
         version: 5.2.0
       typescript:
-        specifier: 'catalog:'
+        specifier: 5.9.3
```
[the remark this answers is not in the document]
lilnasy: This is intentional. I don't think a change in `typescript`'s internal parsing behavior should prevent the project from type checking using the new version.

 It would be unexpected if tests start failing after a routine tooling update. It's two different concerns.

apps/oxlint/test/fixtures/sourceCode_token_methods/files/eslint-test-case.js
```diff
@@ -0,0 +1,3 @@
+/*A*/ var answer /*B*/ = /*C*/ a /*D*/ * b; /*E*/ //F
+call();
```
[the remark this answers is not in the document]
lilnasy: Leave the fixtures alone!

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -63,7 +176,103 @@ export function getTokens(
...
+  for (let lo = sliceStart, hi = nodeTokens.length; lo < hi; ) {
+    const mid = (lo + hi) >> 1;
+    if (nodeTokens[mid].range[0] < rangeEnd) {
+      lo = mid + 1;
+    } else {
+      sliceEnd = hi = mid;
+    }
+  }
+
+  sliceStart -= beforeCount;
```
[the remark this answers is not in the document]
lilnasy: Good catch!

apps/oxlint/src-js/plugins/tokens.ts
overlookmotel: Ha ha ha. Yes I guess linting is the ultimate anal arena!

pnpm-lock.yaml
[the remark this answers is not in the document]
overlookmotel: I'm not sure what the right thing to do here is. I see your logic, but also you've chosen not to pin `@typescript-eslint/typescript-estree` dependency, which seems maybe inconsistent.

But anyway, it's a small point. Let's merge and ponder it later!

overlookmotel (approved):
🍾 Champagne for everyone!

I've squashed and rebased on latest main, and pushed a few commits with style nits - hope you don't mind. I haven't made any substantive changes. Will merge as soon as CI passes.

pnpm-lock.yaml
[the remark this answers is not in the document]
overlookmotel: @camc314 Any opinion on this?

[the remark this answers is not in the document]
camc314: I don't fully understand what's going on here. But I think we should just use the catalog version? From my understanding, these deps are all being bundled, and (hopefuully) our CI shuold fail if we update typescript and there's a change that breaks us?

camc314 (commented):
Amazing work! Thank you!

PR #15601, opened 2025-11-11
style(linter/plugins): update `oxfmt` config to format `oxlint` fixtures

overlookmotel:
Update `oxfmt` config so it formats `apps/oxlint/test/fixtures`.

Actual formatting changes were in PRs #15600 and #15674. This PR alters the config so all these files will continue to be formatted in future.

PR #15862, opened 2025-11-19
style(oxlint/plugins): update `oxfmt` config to only format plugins in `oxlint` fixtures

lilnasy:
- Follow up to #15600 and #15601.
- Makes the inclusion of the fixtures more specific to only target the plugin files.
- The intention is to prevent autofix CI formatting lintable files and autobreaking assertions in the process.

### DIFF
--- a/oxfmtrc.jsonc
+++ b/oxfmtrc.jsonc
@@ -8,7 +8,7 @@
     "**/generated/**",
     // Ignore `fixtures` directories except for `apps/oxlint/test/fixtures`
     "**/fixtures/**",
-    "!apps/oxlint/test/fixtures/**",
+    "!apps/oxlint/test/fixtures/**/plugin.ts",
     "tasks/coverage/node-compat-table",
     "tasks/coverage/misc",
     "tasks/coverage/src/runtime/babelHelpers.js",
--- a/oxfmtrc.jsonc
+++ b/oxfmtrc.jsonc
@@ -6,9 +6,11 @@
     "**/dist/**",
     "**/tests/**",
     "**/generated/**",
-    // Ignore `fixtures` directories except for `apps/oxlint/test/fixtures`
     "**/fixtures/**",
-    "!apps/oxlint/test/fixtures/**/plugin.ts",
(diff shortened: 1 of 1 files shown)


oxfmtrc.jsonc
```diff
@@ -8,7 +8,7 @@
     "**/generated/**",
     // Ignore `fixtures` directories except for `apps/oxlint/test/fixtures`
     "**/fixtures/**",
-    "!apps/oxlint/test/fixtures/**",
+    "!apps/oxlint/test/fixtures/**/plugin.ts",
```
lilnasy: This doesn't work. Linted files are ignored as intended, but `plugin.ts` are not negated, they are also ignored. Not sure why. Single asterisk has the same issue.

lilnasy: It didn't work because gitignore style globs can't select files that are in an unselected directory.

overlookmotel:
Personally, I like formatting everything that can be formatted. We can use `// prettier-ignore` comments on parts of files where whitespace etc matters.

Unless this is going to be a major ball-ache for token APIs?

lilnasy:
This doesn't uniquely affect token APIs. Edit: maybe I spoke too soon. I imagine there will be more cases like this: https://github.com/oxc-project/oxc/pull/15861#discussion_r2542919409.

I was thrown off by the autofix CI autobreaking already written tests. I'll add `// prettier-ignore` for now and refactor the assertions at a later point.

overlookmotel:
I don't know what editor you use. I use VS Code, and have installed the Oxc VS Code extension, with "experimental format" and "format on save" options enabled. That's quite useful for this, as when editing fixture files, you see immediately if the file is going to get formatted in a way that you don't want, and it's obvious where to put the `// prettier-ignore` comments.

But Oxfmt is alpha, so completely understand if you don't want to use it.

Prettier doesn't support a `// prettier-ignore-file` comment to skip formatting on whole file, so currently Oxfmt doesn't support it either. But we can request that feature (seems useful to me), and that'd make all this rather easier.

PR #15886, opened 2025-11-19
refactor(linter/plugins): enable TS `strictNullChecks`

overlookmotel:
Adapted from #15868. Enable `strictNullChecks` in TSConfig for `apps/oxlint`.

Rather than using `!` all over the place, I've added a function `assertIsNonNull`, which performs the same function. Idea is that we can later on produce a debug build where `assertIsNonNull` does an actual runtime assertion (rather than just a TS `asserts`) that the value is not `null` or `undefined`. We can use the debug build in tests, and catch any logic errors.

Unfortunately, TypeScript doesn't seem able to track side effects, so there seems to be no way to tell it "after calling `initAst()`, `ast` is always non-null". Don't want to switch to a having a `getAst` function which returns a non-null value, as that'd introduce unnecessary temp vars, and function calls, on hot paths. So this seems to be the best we can do without incurring runtime cost.

### DIFF
--- a/apps/oxlint/src-js/cli.ts
+++ b/apps/oxlint/src-js/cli.ts
@@ -24,7 +24,7 @@ function loadPluginWrapper(path: string, packageName: string | null): Promise<st
       return loadPlugin(path, packageName);
     });
   }
-  return loadPlugin(path, packageName);
+  return loadPlugin!(path, packageName);
 }
 
 /**
--- a/apps/oxlint/src-js/plugins/load.ts
+++ b/apps/oxlint/src-js/plugins/load.ts
@@ -8,6 +8,7 @@ import type { Context } from './context.ts';
 import type { JsonValue } from './json.ts';
 import type { RuleMeta } from './rule_meta.ts';
 import type { AfterHook, BeforeHook, Visitor, VisitorWithHooks } from './types.ts';
+import type { SetNullable } from './utils.ts';
 
 const ObjectKeys = Object.keys;
 
@@ -181,7 +182,7 @@ async function loadPluginImpl(path: string, packageName: string | null): Promise
     // Create `RuleDetails` object for rule.
     const ruleDetails: RuleDetails = {
       rule: rule as CreateRule, // Could also be `CreateOnceRule`, but just to satisfy type checker
-      context: null as Readonly<Context>, // Filled in below
+      context: null!, // Filled in below
       isFixable,
       messages,
       ruleIndex: 0,
@@ -197,7 +198,7 @@ async function loadPluginImpl(path: string, packageName: string | null): Promise
 
     if ('createOnce' in rule) {
       // TODO: Compile visitor object to array here, instead of repeating compilation on each file
-      let visitorWithHooks = rule.createOnce(context);
+      let visitorWithHooks = rule.createOnce(context) as SetNullable<VisitorWithHooks, 'before' | 'after'>;
       if (typeof visitorWithHooks !== 'object' || visitorWithHooks === null) {
         throw new TypeError('`createOnce` must return an object');
       }
@@ -219,9 +220,9 @@ async function loadPluginImpl(path: string, packageName: string | null): Promise
         afterHook = null;
       }
 
-      (ruleDetails as Writable<CreateOnceRuleDetails>).visitor = visitor;
-      (ruleDetails as Writable<CreateOnceRuleDetails>).beforeHook = beforeHook;
-      (ruleDetails as Writable<CreateOnceRuleDetails>).afterHook = afterHook;
+      (ruleDetails as unknown as Writable<CreateOnceRuleDetails>).visitor = visitor;
+      (ruleDetails as unknown as Writable<CreateOnceRuleDetails>).beforeHook = beforeHook;
+      (ruleDetails as unknown as Writable<CreateOnceRuleDetails>).afterHook = afterHook;
     }
 
     registeredRules.push(ruleDetails);
--- a/apps/oxlint/src-js/plugins/scope.ts
+++ b/apps/oxlint/src-js/plugins/scope.ts
@@ -8,7 +8,7 @@ import {
   type ScopeManager as TSESLintScopeManager,
 } from '@typescript-eslint/scope-manager';
 import { ast, initAst } from './source_code.js';
-import { assertIs } from './utils.js';
+import { assertIs, assertIsNonNull } from './utils.js';
 
 import type * as ESTree from '../generated/types.d.ts';
 import type { SetNullable } from './utils.ts';
@@ -109,6 +109,7 @@ const analyzeOptions: SetNullable<AnalyzeOptions, 'sourceType'> = {
  */
 function initTsScopeManager() {
   if (ast === null) initAst();
+  assertIsNonNull(ast);
 
   analyzeOptions.sourceType = ast.sourceType;
   assertIs<AnalyzeOptions>(analyzeOptions);
@@ -199,6 +200,7 @@ export function isGlobalReference(node: ESTree.Node): boolean {
   if (node.type !== 'Identifier') return false;
 
   if (tsScopeManager === null) initTsScopeManager();
+  assertIsNonNull(tsScopeManager);
 
   const { scopes } = tsScopeManager;
   if (scopes.length === 0) return false;
@@ -229,6 +231,8 @@ export function isGlobalReference(node: ESTree.Node): boolean {
 export function getDeclaredVariables(node: ESTree.Node): Variable[] {
   // ref: https://github.com/eslint/eslint/blob/e7cda3bdf1bdd664e6033503a3315ad81736b200/lib/languages/js/source-code/source-code.js#L904
   if (tsScopeManager === null) initTsScopeManager();
+  assertIsNonNull(tsScopeManager);
+
   // @ts-expect-error // TODO: Our types don't quite align yet
   return tsScopeManager.getDeclaredVariables(node);
 }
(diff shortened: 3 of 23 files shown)


overlookmotel:
* **#15922** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15921** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15921?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15920** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15911** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15911?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15887** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15888** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15886** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15886?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15886?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#15885** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15884** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15886).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

apps/oxlint/src-js/cli.ts
```diff
@@ -24,7 +24,7 @@ function loadPluginWrapper(path: string, packageName: string | null): Promise<st
       return loadPlugin(path, packageName);
     });
   }
-  return loadPlugin(path, packageName);
+  return loadPlugin!(path, packageName);
```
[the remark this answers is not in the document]
overlookmotel: Not using `assertIsNonNull` here because that creates a shared chunk between `cli.js` and `plugin.js`, containing `assertIsNonNull`, and `assertIsNonNull` then doesn't get shaken out by minifier anywhere in the package.

apps/oxlint/src-js/plugins/scope.ts
```diff
@@ -254,6 +259,7 @@ export function getScope(node: ESTree.Node): Scope {
       return scope.type === 'function-expression-name' ? scope.childScopes[0] : scope;
     }
 
+    // @ts-expect-error - Don't want to create a new variable just to make it nullable
     node = node.parent;
```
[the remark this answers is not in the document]
overlookmotel: It's fine!

apps/oxlint/src-js/plugins/visitor.ts
```diff
@@ -224,7 +225,7 @@ export function addVisitorToCompiled(visitor: Visitor): void {
     // TODO: Combine the two hashmaps `NODE_TYPE_IDS_MAP` and selectors cache into one `Map`
     // to avoid 2 hashmap lookups for selectors?
     let typeId = NODE_TYPE_IDS_MAP.get(name);
-    if (typeId !== void 0) {
+    if (typeId !== undefined) {
```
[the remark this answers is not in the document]
overlookmotel: Nonsense. This PR replaces *all* `void 0` with `undefined`.

apps/oxlint/src-js/plugins/load.ts
```diff
@@ -181,7 +182,7 @@ async function loadPluginImpl(path: string, packageName: string | null): Promise
     // Create `RuleDetails` object for rule.
     const ruleDetails: RuleDetails = {
       rule: rule as CreateRule, // Could also be `CreateOnceRule`, but just to satisfy type checker
-      context: null as Readonly<Context>, // Filled in below
```
camc314, line 184:
```suggestion
      context: null!, // Filled in below
```

I think `null!` is the idomatic way to say to TS that i'm assigning to this thing later, rather than having a bunch of type casts

apps/oxlint/src-js/plugins/scope.ts
```diff
@@ -254,6 +259,7 @@ export function getScope(node: ESTree.Node): Scope {
       return scope.type === 'function-expression-name' ? scope.childScopes[0] : scope;
     }
 
+    // @ts-expect-error - Don't want to create a new variable just to make it nullable
```
camc314, line 262: I made an issue for a minifier enhancement for this case:

https://github.com/oxc-project/oxc/issues/15908

apps/oxlint/src-js/plugins/utils.ts
```diff
@@ -42,6 +42,17 @@ export function getErrorMessage(err: unknown): string {
...
+ * Assert a value is not `null` or `undefined`.
+ *
+ * Has no runtime effect - only for guiding the type-checker.
+ * Minification removes this function and all calls to it, so it has zero runtime cost.
+ *
+ * @param value - Value
+ */
+// oxlint-disable-next-line no-unused-vars
+export function assertIsNonNull<T>(value: T | null | undefined): asserts value is T {}
+
```
camc314, line 55:
```suggestion
export function assertIsNonNull<T>(value: T | null | undefined): asserts value is T {
  if (process.env.NODE_ENV !== 'production') {
    if (value === null || value === undefined) {
      throw new TypeError('Value is null or undefined');
    }
  }
}
```

We might need to configure this to make sure we do dev builds/prod builds ect, but I think it's a good idea to have assertions here

apps/oxlint/src-js/plugins/visitor.ts
```diff
@@ -371,13 +372,17 @@ export function finalizeCompiledVisitor() {
   for (let i = mergedEnterVisitorTypeIds.length - 1; i >= 0; i--) {
     const typeId = mergedEnterVisitorTypeIds[i];
     const enterExit = compiledVisitor[typeId] as CompilingNonLeafVisitorEntry;
-    enterExit.enter = mergeVisitFns(enterExit.enter as VisitFn[]);
+    assertIsNonNull(enterExit);
+    assertIs<VisitFn[]>(enterExit.enter);
```
camc314, line 376: Is this type assertion save? `enterExit.enter` is of type `VisitFn | VisitFn[] | null`, should the type of `compiledVisior` be changed ?

apps/oxlint/src-js/plugins/scope.ts
overlookmotel: That'll probably save a lot of bytes, given how common it is in TS code.

Can we leave this as is for now? I am a bit of fanatic about not incurring perf cost just to satisfy TSC, and it wouldn't surprise me if Sapphi comes back saying "unfortunately this optmization is not possible because of getters/proxies/some other obscurity".

apps/oxlint/src-js/plugins/utils.ts
overlookmotel: We'd need to set up TSDown to clean out that code in production builds (and keep it in debug builds). The point of `assertIsNonNull` is an alternative to `!` which allows us to make it a runtime check in debug build later on. But I'd prefer to leave that to a follow-on PR, so as not to incur that perf cost in the meantime. That OK with you?

apps/oxlint/src-js/plugins/visitor.ts
overlookmotel: This is really tricky. `compiledVisitor` changes type throughout the compilation process.

Any ID in `mergedEnterVisitorTypeIds` is guaranteed to point to an entry which has `enter` being `VisitFn[]`. But I couldn't find any way to express this in the type system. And refactoring so it *does* work with TS would hurt perf.

The convoluted code in this file is, as you commented before, "crazy". But it did speed up compiling visitors by 30%, so I think it's worth it.

I've inserted `assertIs` calls here (rather than `as` type cast) to remind me to replace these later on with different assertion functions which do a runtime check in debug mode.

apps/oxlint/src-js/plugins/load.ts
overlookmotel: Ah ha! Much better. Have made that change.

apps/oxlint/src-js/cli.ts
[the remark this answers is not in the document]
overlookmotel: Your "colleague" Mr Copilot already brought this up.

apps/oxlint/src-js/plugins/scope.ts
camc314: sounds good to me

apps/oxlint/src-js/plugins/visitor.ts
camc314: eh, personally i dislike `assertIs` more, because it reminds me of `asserts!` in rust, which actually asserts stuff, where as `assertIs` is just type magic, built on hopes and dreams

camc314:
### Merge activity

* **Nov 20, 3:49 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Nov 20, 3:49 PM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Nov 20, 3:58 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

Issue #15797, opened 2025-11-17
linter: `eslint/sort-keys/allowLineSeparatedGroups` is not working


ryansuhartanto:
### What version of Oxlint are you using?

1.29.0

### What command did you run?

`oxlint --config linter/eslint/sort-keys/allowLineSeparatedGroups/.oxlintrc.json linter/eslint/sort-keys/allowLineSeparatedGroups`

### What does your `.oxlintrc.json` config file look like?

```jsonc
{
  "plugins": [
    "eslint"
  ],
  "rules": {
    "eslint/sort-keys": [
      "error",
      {
        "allowLineSeparatedGroups": true
      }
    ]
  }
}
```

### What happened?

Checkout this file or visit this repo <https://github.com/ryansuhartanto/oxc-issues/tree/main/linter/eslint/sort-keys/allowLineSeparatedGroups>:
```ts
interface Meow {
  name: string;
  age: number;
  color: string;
}

export const meow: Meow = {
  name: "Whiskers", // Empty line below 👇️

  age: 3, // Empty line above 👆️
  color: "Tabby",
};
```

It disregard the `allowSeparateTypeImports` option with this error:
```
> oxc-issues@ lint [omitted]/oxc-issues
> oxlint --type-aware --config linter/eslint/sort-keys/allowLineSeparatedGroups/.oxlintrc.json linter/eslint/sort-keys/allowLineSeparatedGroups


  × eslint(sort-keys): Object keys should be sorted
    ╭─[linter/eslint/sort-keys/allowLineSeparatedGroups/index.ts:7:27]
  6 │     
  7 │ ╭─▶ export const meow: Meow = {
  8 │ │     name: "Whiskers",
  9 │ │   
 10 │ │     age: 3,
 11 │ │     color: "Tabby",
 12 │ ╰─▶ };
    ╰────
  help: Replace `name: "Whiskers",
        
          age: 3,
          color: "Tabby"` with `age: 3, color: "Tabby", name: "Whiskers"`.

Found 0 warnings and 1 error.
Finished in 104ms on 1 file with 53 rules using 8 threads.
 ELIFECYCLE  Command failed with exit code 1.
```

xuanduc987:
The document for this rule is incorrect, the first option is sort order so the correct rule should be
```
{
    "sort-keys": ["error", "asc", {"allowLineSeparatedGroups": true}]
}
```

connorshea:
[...]

PR #15904, opened 2025-11-20
feat(linter): Support auto generate config document for tuple lint option

xuanduc987:
Related to #15797. This allows configuration for rule like `eslint/sort-keys` to render correctly.

The generated result is bellow

```markdown
## Configuration

### The 1st option

type: `"desc" | "asc"`

Sorting order for keys. Accepts "asc" for ascending or "desc" for descending.


### The 2nd option

This option is a object with the following properties:

#### allowLineSeparatedGroups

type: `boolean`

default: `false`

When true, groups of properties separated by a blank line are sorted independently.


#### caseSensitive

type: `boolean`

default: `true`

Whether the sort comparison is case-sensitive (A < a when true).


#### minKeys

type: `integer`

default: `2`

Minimum number of properties required in an object before sorting is enforced.


#### natural

type: `boolean`

default: `false`

Use natural sort order so that, for example, "a2" comes before "a10".
```

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/sort_keys.rs
+++ b/crates/oxc_linter/src/rules/eslint/sort_keys.rs
@@ -13,10 +13,11 @@ use serde::{Deserialize, Serialize};
 use crate::{AstNode, context::LintContext, rule::Rule};
 
 #[derive(Debug, Default, Clone)]
-pub struct SortKeys(Box<SortKeysOptions>);
+pub struct SortKeys(Box<SortKeysConfig>);
 
 #[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "lowercase")]
+/// Sorting order for keys. Accepts "asc" for ascending or "desc" for descending.
 pub enum SortOrder {
     Desc,
     #[default]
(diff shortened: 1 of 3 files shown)


connorshea:
I'm glad to see someone working on improving this, thank you 🙏

connorshea:
A few other rules that will need this in order to get documentation generation working for them (we can do these in later PRs, just making a note):

- [func-names](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/func_names.rs)
- [yoda](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/yoda.rs)
- [arrow-body-style](https://github.com/oxc-project/oxc/blob/main/crates/oxc_linter/src/rules/eslint/arrow_body_style.rs)
- [max-dependencies](https://github.com/oxc-project/oxc/blob/9b564250fe35aaee3dacc1798381c7e746eb6f8d/crates/oxc_linter/src/rules/import/max_dependencies.rs) I'm not sure if this one is really fixed by this PR, we may need to add some additional handling, but this PR at least gets us closer to accurately representing it.

ryansuhartanto:
Another one that I use and notice is [`import/consistent-type-specifier-style`](https://github.com/oxc-project/oxc/blob/oxlint_v1.29.0/crates/oxc_linter/src/rules/import/consistent_type_specifier_style.rs). It accepts a literal string and not an object.

connorshea (approved):
Tested this locally with the `just website` command to make sure it came out fine and didn't cause changes on any other pages, and this all looks good to me, thanks!

Only concern is that it doesn't seem to pull the default value for the first option, but we can look into fixing that later.

connorshea:
> Another one that I use and notice is [`import/consistent-type-specifier-style`](https://github.com/oxc-project/oxc/blob/oxlint_v1.29.0/crates/oxc_linter/src/rules/import/consistent_type_specifier_style.rs). It accepts a literal string and not an object.

Thanks! I've added my examples and that one to #16023.

PR #15926, opened 2025-11-21
feat(formatter)!: Default lineWidth: 100

### LINKED ISSUE #15900: oxfmt: default print width from 80 to 100
According to https://github.com/oxc-project/oxc/discussions/15851

leaysgur:
Fixes #15900 

- [x] oxc_formatter: Update default value
- [x] oxc_formatter: Fix affected unit tests
- [x] oxc_formatter: Accept snapshot tests
- [x] oxc_formatter: Review and fix snapshots which actually tests for width
- [x] oxfmt: Confirmed as unaffected
- [x] prettier_conformance: Make it work


leaysgur:
* **#15926** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/15926?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=15926).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Dunqing:
I think we should have run both 80 and 100 in the fixtures. You can add an `options.json` to the JS and TS folders, which is like https://github.com/oxc-project/oxc/blob/d9bd61702651a3b5a2afc7f2cae72c68184c5cda/crates/oxc_formatter/tests/fixtures/js/arguments/long-curried-call/options.json . 

But we lack the ability to resolve the options file deeply upward, so we need to support it first.

leaysgur:
> I think we should have run both 80 and 100 in the fixtures.

For now, I've updated existing tests that check length to maintain the same behavior.

Do you still think it's better to run two patterns? 👀 

> But we lack the ability to resolve the options file deeply upward, so we need to support it first.

This means merging the `options.json` near the test file with the `options.json` file in a higher directory, correct?
(Currently, there's a feature to search by tracing back to the parent, but once it finds one nearby, it stops searching there.)

Personally, I'm not comfortable to add logic within the test domain. Prefer declarative as possible.

In that case, how about adding both the 80 and 100 options as infrastructure for all cases?

Dunqing:
> > I think we should have run both 80 and 100 in the fixtures.
> 
> For now, I've updated existing tests that check length to maintain the same behavior.

Most of the fixture tests found in the real codebases use the 80 print width, so we should keep them as they are to match the original intention.

> Do you still think it's better to run two patterns? 👀

I am fine with either way: keep using 80 for the fixtures or run both 80 and 100. Either way can maintain the original intention. Run in 100 to verify that the default behavior works or not.

> > But we lack the ability to resolve the options file deeply upward, so we need to support it first.
> 
> This means merging the `options.json` near the test file with the `options.json` file in a higher directory, correct? (Currently, there's a feature to search by tracing back to the parent, but once it finds one nearby, it stops searching there.)
> 
Yes.
> Personally, I'm not comfortable to add logic within the test domain. Prefer declarative as possible.
> 
> In that case, how about adding both the 80 and 100 options as infrastructure for all cases?

That sounds good

leaysgur:
Thanks~! 🙏🏻

Then I will close this PR and recreate with:

- Change default
- Specify 80 for all fixture tests

And then following PR:

- Add 100(or default) for all fixture tests

PR #15955, opened 2025-11-21
feat(linter/plugins): implement `SourceCode#getTokensBefore()`

lilnasy:
- Follow up to #15861.

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -357,7 +357,69 @@ export function getTokensBefore(
   nodeOrToken: NodeOrToken | Comment,
   countOptions?: CountOptions | number | FilterFn | null,
 ): Token[] {
-  throw new Error('`sourceCode.getTokensBefore` not implemented yet'); // TODO
+  if (tokens === null) initTokens();
+  debugAssertIsNonNull(tokens);
+  debugAssertIsNonNull(comments);
+
+  // Maximum number of tokens to return
+  const count =
+    typeof countOptions === 'number'
+      ? max(0, countOptions)
+      : typeof countOptions === 'object' && countOptions !== null
+        ? countOptions.count
+        : null;
+
+  // Function to filter tokens
+  const filter =
+    typeof countOptions === 'function'
+      ? countOptions
+      : typeof countOptions === 'object' && countOptions !== null
+        ? countOptions.filter
+        : null;
+
+  // Whether to return comment tokens
+  const includeComments =
+    typeof countOptions === 'object' &&
+    countOptions !== null &&
+    'includeComments' in countOptions &&
+    countOptions.includeComments;
+
+  // Source array of tokens to search in
+  let nodeTokens: Token[] | null = null;
+  if (includeComments) {
+    if (tokensWithComments === null) {
+      tokensWithComments = [...tokens, ...comments].sort((a, b) => a.range[0] - b.range[0]);
+    }
+    nodeTokens = tokensWithComments;
+  } else {
+    nodeTokens = tokens;
+  }
+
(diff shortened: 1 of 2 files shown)


apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -357,7 +357,93 @@ export function getTokensBefore(
...
+      }
+    }
+    // unreachable
+  } else {
+    if (DEBUG) {
+      throw new Error('Unexpected case');
+    }
+    // Also unreachable, but having this line quells the type checker
+    tokensBefore = [];
+  }
```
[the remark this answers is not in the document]
lilnasy: > This else block should be removed

It's needed because I want the common case to be the first case for performance, and if `typeof` checks are removed from the 4th `else` branch, `typescript` doesn't infer that the only case left is where `filter` is a function and `count` is a number.

> the comment on line 437 is also incorrect

It's not.

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -357,7 +357,93 @@ export function getTokensBefore(
...
+  // Source array of tokens to search in
+  let nodeTokens: Token[] | null = null;
+  if (includeComments) {
+    if (tokensWithComments === null) {
+      tokensWithComments = [...tokens, ...comments].sort((a, b) => a.range[0] - b.range[0]);
+    }
+    nodeTokens = tokensWithComments;
+  } else {
+    nodeTokens = tokens;
+  }
```
lilnasy, line 395: This is repetitive, but I don't want a function dispatch here for performance.

I wonder if we could make the bundler inline functions with a macro or a directive, like what `terser` does for function calls marked with `/* #__INLINE__ */`.

apps/oxlint/src-js/plugins/tokens.ts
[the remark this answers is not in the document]
lilnasy: Refactored in a way that typescript can infer the type narrowing.

overlookmotel (approved):
I need to go out now, so don't have time to review in full. But don't want to block you, so merging now and I'll take another look another time.

apps/oxlint/src-js/plugins/tokens.ts
overlookmotel: > This is repetitive, but I don't want a function dispatch here for performance.

Good point about function dispatch. I think it'd be OK only to conditionally incur that cost. Something like:

```js
if (tokensWithComments === null) initTokensWithComments();
debugAssertIsNonNull(tokensWithComments);
```

```js
/**
  * Initialize `tokensWithComments`.
  * Caller must ensure that `tokens` and `comments` are initialized before calling this function.
  */
function initTokensWithComments() {
  // TODO: Use `start` instead of `range[0]` once we have our own tokens
  tokensWithComments = [...tokens, ...comments].sort((a, b) => a.range[0] - b.range[0]);
}
```

> I wonder if we could make the bundler inline functions with a macro or a directive, like what `terser` does for function calls marked with `/* #__INLINE__ */`.

Yes, that'd be ideal. At present Oxc minifier doesn't do *any* function inlining, but once it does, we should do that.

overlookmotel: Note: #15968 renamed `assertIsNonNull` to `debugAssertIsNonNull`.

overlookmotel:
Oh damn it, actually I can't merge this because I've screwed it up with #15968. Sorry, I should have merged this before I did my PRs. I'll be back at keyboard tomorrow morning and will try to get both PRs merged then (I'll handle any merge conflicts between them).

lilnasy:
> I'll handle any merge conflicts between them

I plan to implement a few more methods by tomorrow, so I did some git fu to prevent conflicts. You could merge them in any sequence.

PR #15971, opened 2025-11-22
feat(linter/plugins): implement `SourceCode#getTokensAfter()`

lilnasy:
- Follow up to #15861.

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -565,7 +565,84 @@ export function getTokensAfter(
   nodeOrToken: NodeOrToken | Comment,
   countOptions?: CountOptions | number | FilterFn | null,
 ): Token[] {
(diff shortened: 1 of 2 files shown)


overlookmotel:
I've squashed and rebased on main (after merging the 2 preceding PRs), but I've obviously screwed something up in the process. I have to go out now, but I'll figure it out and fix when I get back later / tomorrow morn. @lilnasy No need for you to do anything - I made this mess, so I'll fix it! I'll get this and the other 2 PRs merged once I fix it.

overlookmotel:
Oh. I think maybe problem is an interaction with tests for `getTokenBefore` which set `sourceText` to a different value.

lilnasy:
Yes, you are right. I'm implementing the fix.

Edit: passing now.

overlookmotel (approved):

overlookmotel (approved):

PR #15997, opened 2025-11-23
feat(linter): add ignores option to overrides

### LINKED ISSUE #15932: linter: Support negations in overrides
This syntax would be very useful in an override:

```
      "files": ["packages/services/**", "!packages/services/**/*.test.ts"],
      "rules": {
        "no-restricted-imports": [
          "error",
          {
            "patterns": [
              {
                "group": [
...snip...
```

Currently it doesn't seem to work.

xuanduc987:
This will add a new `ignores` option to oxlint override config, allowing excluding some files matched by `files` from applying the override.

Resolves #15932

### DIFF
--- a/apps/oxlint/src-js/package/config.generated.ts
+++ b/apps/oxlint/src-js/package/config.generated.ts
@@ -419,6 +419,10 @@ export interface OxlintOverride {
    */
   globals?: OxlintGlobals;
   /**
+   * If a file matches any of these glob patterns, the overrides configuration won't apply.
+   */
+  ignores?: GlobSet;
+  /**
    * JS plugins for this override, allows usage of ESLint plugins with Oxlint.
    *
    * Read more about JS plugins in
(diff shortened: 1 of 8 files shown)


connorshea:
I think it should probably be called `ignorePatterns` to match the root-level config option name? Although I guess it depends on how we want it to work with regards to merging with the ignore values with the root ignores

xuanduc987:
I chose this name to be similar to `eslint`

---

`eslint` also used to have `ignorePatterns` config option at the root level, and in the override, it had `excludedFiles` options.

But now, with flat config, `eslint` uses `ignores` name in its [configuration objects](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects)

todor-a:
Hey, any plans to merge this? It's in the docs but apparently it's not supported 🤔 https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#overrides-n-excludefiles

xuanduc987:
I had just rebased this PR onto current main branch

xuanduc987:
@camc314 Sorry for the ping but could you take a look at this PR?

JasonKaz:
> Hey, any plans to merge this? It's in the docs but apparently it's not supported 🤔 https://oxc.rs/docs/guide/usage/formatter/config-file-reference.html#overrides-n-excludefiles

That is for oxfmt, not oxlint.

This is an essential feature used by our eslint configuration. Without it we cannot migrate to oxlint.

avcharovBetsson:
HI @camc314, we are in need to this feature to migrate to oxlint
 is there any issues with approving this PR?

PR #16055, opened 2025-11-24
refactor(linter/plugins): reimplement `isSpaceBetween()` `isSpaceBetweenTokens()` using tokens

lilnasy:
- Follow up to https://github.com/oxc-project/oxc/pull/15498 and https://github.com/oxc-project/oxc/pull/15861
- Note: JSX special case (as part of the `isSpaceBetweenTokens()` method) is being removed in ESLint v10 (see [eslint/eslint#20137](https://github.com/eslint/eslint/pull/20137/files#diff-e85eaebdcf6af45c06572019cbd900cff91518c789049c3bd0e8724c18645ab0))

### Tasks:
- [x] Token-based reimplementation of `isSpaceBetween()`
- [x] Token-based reimplementation of `isSpaceBetweenTokens()`
- [x] Pass existing tests from #15498
- [x] New tests for JS syntax
- [x] New tests for JSX syntax

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -1727,8 +1727,22 @@ export function isSpaceBetween(first: NodeOrToken, second: NodeOrToken): boolean
   // The "between" range
   let rangeStart: number, rangeEnd: number;
 
+  // Find the range between the two nodes/tokens.
+  //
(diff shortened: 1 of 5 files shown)


overlookmotel (approved):
Excellent! Thanks very much for adding so many tests.

I've pushed a few commits - all nits. I've only split it into so many tiny commits so you can see what I'm trying to do clearly (if you're interested).

PR #16071, opened 2025-11-24
perf(linter/plugins): optimize merging of `tokens` and `comments`

lilnasy:
- Follow up to #15861.
- Implements the performance optimization pointed out in #15952.

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -186,9 +186,12 @@ function initTokensWithComments() {
   // TODO: Replace `range[0]` with `start` once we have our own tokens which have `start` property.
   tokensWithComments = [];
 
+  const tokensLength = tokens.length,
+    commentsLength = comments.length;
+
   let tokensIndex = 0,
     commentsIndex = 0;
-  while (tokensIndex < tokens.length && commentsIndex < comments.length) {
+  while (tokensIndex < tokensLength && commentsIndex < commentsLength) {
     const token = tokens[tokensIndex],
       comment = comments[commentsIndex];
 
@@ -203,10 +206,11 @@ function initTokensWithComments() {
   }
 
   // After one of `tokens` or `comments` is exhausted, directly push the other's elements
-  while (commentsIndex < comments.length) tokensWithComments.push(comments[commentsIndex++]);
-  while (tokensIndex < tokens.length) tokensWithComments.push(tokens[tokensIndex++]);
+  while (commentsIndex < commentsLength) tokensWithComments.push(comments[commentsIndex++]);
+  while (tokensIndex < tokensLength) tokensWithComments.push(tokens[tokensIndex++]);
 
   if (DEBUG) {
+    if (tokensWithComments.length !== tokensLength + commentsLength) {
       throw new Error("Not all tokens and comments were merged");
     }
   }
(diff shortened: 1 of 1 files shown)


apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -181,10 +181,34 @@ export function initTokens() {
 function initTokensWithComments() {
   debugAssertIsNonNull(tokens);
   debugAssertIsNonNull(comments);
-  // TODO: `tokens` and `comments` are already sorted, so there's a more efficient algorithm to merge them.
-  // That'd certainly be faster in Rust, but maybe here it's faster to leave it to JS engine to sort them?
   // TODO: Replace `range[0]` with `start` once we have our own tokens which have `start` property.
-  tokensWithComments = [...tokens, ...comments].sort((a, b) => a.range[0] - b.range[0]);
+  tokensWithComments = [];
+
+  let tokensIndex = 0,
```
overlookmotel, line 193: Add before loop:

```ts
const tokensLength = tokens.length
  commentsLength = comments.length;
```

and then use those vars in the rest of the function, to avoid repeated `.length` property lookups.

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -181,10 +181,34 @@ export function initTokens() {
...
+      tokensIndex++;
+    } else {
+      tokensWithComments.push(comment);
+      commentsIndex++;
+    }
+  }
+
+  // After one of `tokens` or `comments` is exhausted, directly push the other's elements
+  while (commentsIndex < comments.length) tokensWithComments.push(comments[commentsIndex++]);
+  while (tokensIndex < tokens.length) tokensWithComments.push(tokens[tokensIndex++]);
```
overlookmotel: One of the arrays is exhausted at this point.

```suggestion
  if (tokensIndex < tokensLength) {
    tokensWithComments.push(...tokens.slice(tokensIndex));
  } else {
    tokensWithComments.push(...comments.slice(commentsIndex));
  }
```

apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -181,10 +181,34 @@ export function initTokens() {
...
+
+    // TODO: Replace `range[0]` with `start` once we have our own tokens which have `start` property.
+    if (token.range[0] <= comment.range[0]) {
+      tokensWithComments.push(token);
+      tokensIndex++;
+    } else {
+      tokensWithComments.push(comment);
+      commentsIndex++;
+    }
+  }
```
overlookmotel: I think can possibly make this even faster with nested loops, though I can't quite put my finger on how to write it.

Something like:

* Outer loop:
  * If no comments remaining, push all remaining tokens and exit.
  * If no tokens remaining, push all remaining comments and exit.
  * Get next token and comment.
  * If token starts before comment:
    * Inner loop: Repeatedly push tokens until hit one that's after the comment.
    * Push comment.
    * Continue outer loop.
  * Otherwise:
    * Inner loop: Repeatedly push comments until hit one that's after the token.
    * Push token.
    * Continue outer loop.

i.e. push them in "runs". I imagine that'd be faster for the common case where you'll have a load of tokens in a row, then occasionally a comment or two, then another load of tokens, and so on...

overlookmotel (approved):
Since we're optimizing, could try and speed it up even more - suggestion below. But this PR is likely faster than what we had before, so we could merge it as is if you don't have the time/inclination to try to shave off a few more cycles.

overlookmotel (approved):
Great!

PR #16091, opened 2025-11-25
style(all): apply oxfmt default options

Boshen:
Embrace the defaults!


leaysgur:
Will investigate && merge after fixing CI.

PR #16132, opened 2025-11-26
feat(parser): add diagnostic for expected ident after optional chain

camchenry:
Currently we just emit "unexpected token", but we actually have a good amount of context here to provide better help + error message.

PR #16135, opened 2025-11-26
perf(parser): remove bounds check for getting strings from lexer

camchenry:
One profile of the parser showed that this function took up 1% of the time on `binder.ts`. It seemed like this is a case where we could possibly remove the bounds checks, since our tokens are guaranteed to be contained within the source string. Since this is called really often, it's worth optimizing.

+1% on all parser benchmarks: 

image

### DIFF
--- a/crates/oxc_parser/src/lexer/string.rs
+++ b/crates/oxc_parser/src/lexer/string.rs
@@ -270,6 +270,12 @@ impl<'a> Lexer<'a> {
             }
             _ => {}
         }
-        &source_text[start..end]
+        debug_assert!(start <= source_text.len());
+        debug_assert!(end <= source_text.len());
+        debug_assert!(start <= end);
+        debug_assert!(source_text.is_char_boundary(start));
+        debug_assert!(source_text.is_char_boundary(end));
+        // SAFETY: `token` is guaranteed to be within the source and on UTF-8 boundaries.
+        unsafe { source_text.get_unchecked(start..end) }
     }
 }


camchenry:
* **#16135** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16135?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16133** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16132** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16132?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16131** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16131?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=16135).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
I've been um-ing and ah-ing about this. Yes, `token.start()` and `token.end()` *should* make a valid slice, but can we *prove* it? And can we prove that the `Token`'s `Kind` corresponds to its span, so we're offsetting `start` / `end` correctly?

When using unsafe code, IMO you should have more confidence than "it *should* be right".

I'd like to see how far we can get without resorting to unsafe. Here's a start: #16317.

I also wonder if we can remove the logic for altering `start` / `end` based on `Kind` from `get_string`, by moving that logic into the call sites. Each call site I think knows what the `Kind` is, so knows what to do. The logic doesn't need to be centralized.

We could also see if putting `#[inline]` on this function makes a difference. The gain in this PR might be down to reducing the size of the function, so it gets inlined, rather than the change in logic itself. Maybe...

overlookmotel:
Note: #16283 would also change this function.

overlookmotel:
> We could also see if putting `#[inline]` on this function makes a difference. The gain in this PR might be down to reducing the size of the function, so it gets inlined, rather than the change in logic itself. Maybe...

Nope. Tried this in #16329 and it had no effect.

camchenry:
> I've been um-ing and ah-ing about this. Yes, `token.start()` and `token.end()` _should_ make a valid slice, but can we _prove_ it? And can we prove that the `Token`'s `Kind` corresponds to its span, so we're offsetting `start` / `end` correctly?
> 
> When using unsafe code, IMO you should have more confidence than "it _should_ be right".
> 
> I'd like to see how far we can get without resorting to unsafe. Here's a start: #16317.
> 
> I also wonder if we can remove the logic for altering `start` / `end` based on `Kind` from `get_string`, by moving that logic into the call sites. Each call site I think knows what the `Kind` is, so knows what to do. The logic doesn't need to be centralized.
> 
> We could also see if putting `#[inline]` on this function makes a difference. The gain in this PR might be down to reducing the size of the function, so it gets inlined, rather than the change in logic itself. Maybe...

Yeah, I'm not sure I can think of a proof of the top of my head that would definitely _prove_ that it always is correct. In the same sense that spans on tokens should always be correct since they should refer to offsets within the program, but there's nothing technically stopping us from creating an invalid span accidentally (or on purpose). I think the reasoning is sound, but it's only a 1% performance improvement or so according to our benchmarks, so I don't feel like it's absolutely critical to change.

I've also been thinking about this issue recently: https://github.com/oxc-project/backlog/issues/46. I feel like some of the ideas in that thread have the potential to be much more impactful than just a meager 1% in exchange for adding an `unsafe` block. I'd be more inclined to merge this if we were seeing 5% improvements or something.

overlookmotel:
1% is definitely worth having. It's really tempting.

But my personal view about unsafe code is that's it's about using constraints in the type system to *guarantee* that the code is sound. You should be able to put a comment above the `unsafe { }` block explaining how it's *impossible* that your assumptions can be violated. In this case, as you say, it's trivial to break the rules - we haven't locked down the types to prevent you doing that.

So, on balance, I'm afraid that I don't think we should do this.

However, I do think there's probably space to get at least some of the perf improvement without unsafe. This idea, for example:

> I also wonder if we can remove the logic for altering `start` / `end` based on `Kind` from `get_string`, by moving that logic into the call sites. Each call site I think knows what the `Kind` is, so knows what to do. The logic doesn't need to be centralized.

There's probably other ways too that I've not thought of.

---

> I've also been thinking about this issue recently: https://github.com/oxc-project/backlog/issues/46

Yes, I agree, I think there's a lot of potential there. In particular, an `Ident` type which stores the hash inline would make all the hashmap lookups in `oxc_semantic` a good bit faster.  I've also been slowly laying the groundwork for https://github.com/oxc-project/backlog/issues/46#issuecomment-2603153807 in work on allocator and raw transfer. When we *finally* replace Bumpalo with our own allocator (I hope early next year), we'll store all strings in a single contiguous block at start of the allocator, and then we can try this out.

If you're interested in working on our string types, please shout on that issue, and let's discuss.

Boshen:
> So, on balance, I'm afraid that I don't think we should do this.

PR #16184, opened 2025-11-27
fix(linter/plugins): handle zero-token files in `SourceCode#getLastToken()`

lilnasy:
- The root cause of the "iffy" line of code in #16003 was an incorrectly implemented test.
- The corrected test revealed that the code attempted to index the `0` index of an empty array.
- This PR fixes both issues.
- Also took the opportunity to remove a variable declaration in the binary search.

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -655,19 +655,17 @@ export function getLastToken(
 
   // Binary search for the last token within `node`'s range
   const nodeTokensLength = nodeTokens.length;
-  let lastTokenIndex = nodeTokensLength;
-  for (let lo = 0, hi = nodeTokensLength; lo < hi; ) {
-    const mid = (lo + hi) >> 1;
+  let lastTokenIndex = 0;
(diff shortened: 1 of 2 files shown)


apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -655,19 +655,17 @@ export function getLastToken(
...
-      lo = mid + 1;
+      lastTokenIndex = mid + 1;
     } else {
       hi = mid;
     }
   }
 
-  // TODO: this early return feels iffy
-  if (lastTokenIndex === nodeTokensLength) return null;
+  lastTokenIndex--;
```
lilnasy, line 668: In the zero-token file case, the binary search loop doesn't run any iterations, and `lastTokenIndex` gets negated to `-1`, precluding the filter loops below as well.

overlookmotel (approved):

apps/oxlint/src-js/plugins/tokens.ts
overlookmotel: Smart!

PR #16188, opened 2025-11-27
perf(linter/plugins): optimize `getTokens()` and other methods

lilnasy:
- Removes branching in some cases where it was used to calculate slices.
- Avoids indexing into arrays past their length.
- Renames `nodeTokens` to `tokenList`. In the first few methods implemented, they contained tokens of the given node. Not the case for later methods, but the misnomer stayed.
- Improves consistency in comments. Clarifies that binary searches yield tokens before or after a certain point, they are not necessarily _within_ the range we care about.

### DIFF
--- a/apps/oxlint/src-js/plugins/tokens.ts
+++ b/apps/oxlint/src-js/plugins/tokens.ts
@@ -399,7 +399,7 @@ export function getTokens(
   sliceEnd = min(sliceEnd + afterCount, tokensLength);
 
   if (typeof filter !== "function") {
-    return tokenList.slice(sliceStart, sliceEnd);
+    return tokenList.slice(sliceStart, min(sliceStart + (count ?? sliceEnd), sliceEnd));
   }
 
   const allTokens: Token[] = [];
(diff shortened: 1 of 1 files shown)


apps/oxlint/src-js/plugins/tokens.ts
```diff
@@ -399,7 +399,7 @@ export function getTokens(
   sliceEnd = min(sliceEnd + afterCount, tokensLength);
 
   if (typeof filter !== "function") {
-    return tokenList.slice(sliceStart, sliceEnd);
+    return tokenList.slice(sliceStart, min(sliceStart + (count ?? sliceEnd), sliceEnd));
```
lilnasy, line 402: Sorry for sneaking in a fix in this PR. It would be too inconvenient to extract this out without graphite.

overlookmotel: No problem. Thanks for pointing it out.

overlookmotel (approved):

PR #15822, opened 2025-11-18
feat(linter): Add support for options for jsPlugins

wagenet:
Fixes #14825

PR #16215, opened 2025-11-27
feat(linter/plugins): rule options

lilnasy:
- Implements #14825
- Rebased #15822. Refactored. Pruned the slop. Tests passing.

### Big changes from the original approach
- [x] Options IDs are stored in their own arrays. This was laid out in #16158. The original approach was to store `ruleId`-`optionId` pairs as tuples in an array - several heap allocated objects and excessive pointer indirection.
- [x] The rust code populates the options by calling a new JS callback named `setupConfigs()`. I expect we would use this callback for more plugin related data, including settings which are currently implemented in a hacky way.
- [ ] Investigate a refactor to avoid requiring a mutable reference to `ExternalPluginStore` everywhere.

<details>
<summary>

#### Smaller Changes

</summary>

- Removed the extra fixture ("working version"). There were no meaningful changes compared to the first fixture. Probably slop.

</details>


overlookmotel:
Just a note: As I mentioned on Discord, I don't think we should worry about merging options with default options in this first PR - can leave that to a follow-on. But FYI I've implemented the merging logic in #16217 because I needed it for the rule tester.

overlookmotel:
> Refactor to avoid requiring a mutable reference to `ExternalPluginStore` everywhere.

I know it was me that said this should be possible, but I could very well be wrong. If it's not possible, don't worry about it.

apps/oxlint/src/lint.rs
```diff
@@ -291,6 +291,18 @@ impl CliRunner {
...
+        // This code is in the oxlint app, not in oxc_linter crate
+        if let Some(ref external_linter) = external_linter {
+            if let Err(err) = external_plugin_store.setup_configs(external_linter) {
+                print_and_flush_stdout(
+                    stdout,
+                    &format!("Failed to setup external plugin options: {err}\n"),
+                );
+                return CliRunResult::InvalidOptionConfig;
+            }
+        }
```
lilnasy: I could use help here. I exposed `setup_configs()` for the `oxlint` crate by putting it on `ExternalPluginStore` only because it's the first place I found. I could do with a few pointers on separation of concerns regarding these structs.

apps/oxlint/src-js/plugins/lint.ts
```diff
@@ -156,6 +155,9 @@ function lintFileImpl(
     // Set `options` for rule
     const optionsId = optionsIds[i];
     debugAssert(optionsId < allOptions.length, "Options ID out of bounds");
+
+    // If the rule has no user-provided options, use the plugin-provided default
+    // options (which falls back to `DEFAULT_OPTIONS`)
     ruleDetails.options =
       optionsId === DEFAULT_OPTIONS_ID ? ruleDetails.defaultOptions : allOptions[optionsId];
```
lilnasy, line 170: @overlookmotel You mentioned this could be simplified to `allOptions[optionsId]`. How would plugin-provided default options get used in that case?

apps/oxlint/src-js/plugins/options.ts
```diff
@@ -128,3 +130,23 @@ function mergeValues(configValue: JsonValue, defaultValue: JsonValue): JsonValue
...
+  const parsed = JSON.parse(optionsJson);
+  if (DEBUG) {
+    if (!isArray(parsed))
+      throw new TypeError("Expected optionsJson to decode to an array", { cause: parsed });
+    // Basic shape validation: each element must be an array (options tuple array)
+    for (let i = 0; i < parsed.length; i++) {
+      const el = parsed[i];
+      if (!isArray(el)) throw new TypeError("Each options entry must be an array", { cause: el });
+    }
+  }
```
lilnasy, line 152: This is a debug check because I this validation is done probably done on the rust side as well. I need to confirm this.

apps/oxlint/src/lint.rs
overlookmotel: Yeah, the types are a bit of a convoluted mess here. I'm tied up today. Will take a look tomorrow.

lilnasy: - Guessing this is related to https://github.com/oxc-project/oxc/issues/12437?

apps/oxlint/src/js_plugins/external_linter.rs
```diff
@@ -59,6 +61,45 @@ pub enum LintFileReturnValue {
...
+fn wrap_setup_configs(
+    cb: JsSetupConfigsCb,
+) -> Box<dyn Fn(String) -> Result<(), String> + Send + Sync> {
+    Box::new(move |options_json: String| {
+        let (tx, rx) = channel();
+
+        // Send data to JS
+        let status = cb.call_with_return_value(
+            FnArgs::from((options_json,)),
+            ThreadsafeFunctionCallMode::Blocking,
```
overlookmotel:
```suggestion
            ThreadsafeFunctionCallMode::NonBlocking,
```

That follows `wrap_lint_file`, which is also a synchronous call. I can't remember the details exactly, but I do remember that `NonBlocking` doesn't do quite what you'd expect! I think it relates to whether *putting the function in the execution queue* is blocking or not, not whether the function call itself is blocking.

apps/oxlint/src-js/plugins/config.ts
```diff
@@ -0,0 +1,26 @@
...
+  try {
+    setOptions(optionsJSON);
+    // TODO: flesh out error handling.
+    // Currently, the only procedure that may fail is `JSON.parse()`
+    // in `setOptions()`, but it won't because the JSON string from
+    // the rust side is serde serialized.
+    return "ok";
+  } catch (err) {
+    return err instanceof Error ? err.message : String(err);
+  }
```
overlookmotel: We control the string passed to this function, so `optionsJSON` should always be valid JSON. So I don't think any need for error handling here. It can just directly call `setOptions` and return `void`.

apps/oxlint/src-js/plugins/options.ts
```diff
@@ -22,8 +22,10 @@ export type Options = JsonValue[];
 // Default rule options
 export const DEFAULT_OPTIONS: Readonly<Options> = Object.freeze([]);
 
-// All rule options
-export const allOptions: Readonly<Options>[] = [DEFAULT_OPTIONS];
+// All rule options.
+// Indexed by options ID sent alongside ruleId for each file.
+// Element 0 is always the default options (empty array).
+export let allOptions: Readonly<Options>[] = [DEFAULT_OPTIONS];
```
overlookmotel, line 28: Since it's always over-written before linting starts, we can initialize `allOptions` as `null` - 1 less object creation.

This does mean we need to call `debugAssertIsNonNull(allOptions)` before using it, but that's probably no bad thing, as it checks that things are happening in the right order (and `debugAssertIsNonNull` is zero cost in release build).

apps/oxlint/src-js/plugins/options.ts
```diff
@@ -128,3 +130,23 @@ function mergeValues(configValue: JsonValue, defaultValue: JsonValue): JsonValue
...
   return freeze(merged);
 }
+
+/**
+ * Set all external rule options.
+ * Called once from Rust after config building, before any linting occurs.
+ * @param optionsJson - JSON string of outer array of per-options arrays.
+ */
+export function setOptions(optionsJson: string): void {
+  const parsed = JSON.parse(optionsJson);
```
overlookmotel: Nit: Write directly to `allOptions` here. If TS complains, just tell him it's fine!

apps/oxlint/src-js/plugins/options.ts
```diff
@@ -128,3 +130,23 @@ function mergeValues(configValue: JsonValue, defaultValue: JsonValue): JsonValue
...
+  if (DEBUG) {
+    if (!isArray(parsed))
+      throw new TypeError("Expected optionsJson to decode to an array", { cause: parsed });
+    // Basic shape validation: each element must be an array (options tuple array)
+    for (let i = 0; i < parsed.length; i++) {
+      const el = parsed[i];
+      if (!isArray(el)) throw new TypeError("Each options entry must be an array", { cause: el });
```
overlookmotel: Maybe I don't know what I'm talking about, but isn't `cause` meant to be another `Error`? I think we can lose that.

apps/oxlint/src/lint.rs
```diff
@@ -291,6 +291,18 @@ impl CliRunner {
             }
         };
 
+        // TODO: refactor this elsewhere.
+        // This code is in the oxlint app, not in oxc_linter crate
+        if let Some(ref external_linter) = external_linter
```
overlookmotel: This is totally valid Rust, but our "house style" is to avoid `ref` where possible. This is preferable:

```suggestion
        if let Some(external_linter) = &external_linter
```

apps/oxlint/src/run.rs
```diff
@@ -40,12 +40,28 @@ pub type JsLintFileCb = ThreadsafeFunction<
...
+
+/// JS callback to setup configs.
+#[napi]
+pub type JsSetupConfigsCb = ThreadsafeFunction<
+    // Arguments
+    FnArgs<(String,)>, // Stringified options array
+    // Return value
+    String, // Result ("ok" or error message)
+    // Arguments (repeated)
+    FnArgs<(String,)>,
```
overlookmotel: Because there's only 1 argument here, can skip the `FnArgs` abstraction.

```suggestion
pub type JsSetupConfigsCb = ThreadsafeFunction<
    // Arguments
    String, // Options array as JSON
    // Return value
    String, // Result ("ok" or error message)
    // Arguments (repeated)
    String,
```

If that gives you problems (weird errors elsewhere due to the change), don't worry about it - we can change it in a follow-up. NAPI-RS's API is pretty hairy.

apps/oxlint/test/fixtures/custom_plugin_with_options/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+            const shouldReport = options[0];
+
+            if (shouldReport) {
+              context.report({
+                message: JSON.stringify(context.options, null, 2),
+                node,
+              });
+            }
+          },
+        };
```
overlookmotel: Would be simpler just to dump `context.options` to the snapshot as JSON, and return an empty visitor `{}`.

apps/oxlint/test/fixtures/custom_plugin_with_options/plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+    name: "test-plugin-options",
+  },
+  rules: {
+    "check-options": {
+      meta: {
+        messages: {
+          wrongValue: "Expected value to be {{expected}}, got {{actual}}",
+          noOptions: "No options provided",
+        },
+      },
```
overlookmotel: This is extraneous. It can go. Leftovers from the AI?

crates/oxc_linter/src/config/config_store.rs
```diff
@@ -174,8 +174,13 @@ impl Config {
...
 
-        let mut external_rules =
-            self.base.external_rules.iter().copied().collect::<FxHashMap<_, _>>();
+        // Build a hashmap of existing external rules keyed by rule id with value (options_id, severity)
+        let mut external_rules: FxHashMap<ExternalRuleId, (ExternalOptionsId, AllowWarnDeny)> =
+            self.base
+                .external_rules
+                .iter()
+                .map(|(rule_id, options_id, severity)| (*rule_id, (*options_id, *severity)))
+                .collect();
```
overlookmotel: You don't need to type annotation here. This would be shorter:

```suggestion
        let mut external_rules = self.base
            .external_rules
            .iter()
            .map(|(rule_id, options_id, severity)| (*rule_id, (*options_id, *severity)))
            .collect::<FxHashMap<_, _>>();
```

(`_` means "compiler you figure it out")

crates/oxc_linter/src/config/config_store.rs
```diff
@@ -253,7 +258,8 @@ impl Config {
 
         let external_rules = external_rules
             .into_iter()
-            .filter(|(_, severity)| severity.is_warn_deny())
+            .filter(|(_, (.., severity))| severity.is_warn_deny())
```
overlookmotel: This would be more idiomatic:

```suggestion
            .filter(|(_, (_, severity))| severity.is_warn_deny())
```

crates/oxc_linter/src/config/rules.rs
```diff
@@ -106,10 +109,22 @@ impl OxlintRules {
                     if external_plugin_store.is_enabled() {
                         let external_rule_id =
                             external_plugin_store.lookup_rule_id(plugin_name, rule_name)?;
+
+                        // Add options to store and get options ID
+                        let options_id = if let Some(config) = &rule_config.config {
+                            external_plugin_store.add_options(config.clone())
+                        } else {
+                            // No options - use reserved index 0
+                            ExternalOptionsId::from_usize(0)
```
overlookmotel: `ExternalOptionsId::from_usize(0)` appears in a few places. It'd be preferable to define a const for it.

Add where `ExternalOptionsId` is defined (`external_plugin_store.rs`):

```rs
impl ExternalOptionsId {
    const NONE: Self = Self::from_usize(0);
}
```

Then use `ExternalOptionsId::NONE` everywhere else.

crates/oxc_linter/src/external_plugin_store.rs
```diff
@@ -15,13 +17,18 @@ define_index_type! {
...
+}
+
 #[derive(Debug)]
 pub struct ExternalPluginStore {
     registered_plugin_paths: FxHashSet<PathBuf>,
 
     plugins: IndexVec<ExternalPluginId, ExternalPlugin>,
     plugin_names: FxHashMap<String, ExternalPluginId>,
     rules: IndexVec<ExternalRuleId, ExternalRule>,
+    options: IndexVec<ExternalOptionsId, serde_json::Value>,
```
overlookmotel, line 37: This would be a bit more type-safe - ensure that each entry is a `Vec` (so an array on JS side):

```suggestion
    options: IndexVec<ExternalOptionsId, Vec<serde_json::Value>>,
```

crates/oxc_linter/src/external_plugin_store.rs
```diff
@@ -119,6 +129,27 @@ impl ExternalPluginStore {
...
+    /// Add options to the store and return its ID.
+    /// Returns index 0 for empty arrays or null values (no options).
+    pub fn add_options(&mut self, options: serde_json::Value) -> ExternalOptionsId {
+        // If it's null or an empty array, return reserved index 0
+        if options.is_null() || options.as_array().is_some_and(Vec::is_empty) {
+            return ExternalOptionsId::from_usize(0);
+        }
+
+        self.options.push(options)
+    }
```
overlookmotel, line 152: As far as I can see, `ESLintRule::config` field is always `None` or it's `Some` containing `serde_json::Array` with non-zero length. It's never `Some` containing any other type.

https://github.com/oxc-project/oxc/blob/43a6c3209e7c03e3b83d37879ddf7ff919f44d2b/crates/oxc_linter/src/config/rules.rs#L269-L306

This is poor design. `ESLintRule::config` should be `Option<Vec<serde_json::Value>>`, not `Option<serde_json::Value>`. But that's another story...

For now, here could just do:

```rs
let options = options.as_array().unwrap();
debug_assert!(!options.is_empty(), "`options` should never be an empty `Vec`");
self.options.push(options)
```

overlookmotel (commented):
I'll need to go through this again in depth. But looking good!

Some small points below. You seem to be picking up Rust much faster than I did!

apps/oxlint/src-js/plugins/options.ts
overlookmotel: Yes, keeping it as a debug check is good.

apps/oxlint/src-js/plugins/lint.ts
overlookmotel: Ah yes, that was my original thought. But that was before I brought default options into the mix. So as you have it here is correct. We'll change it once we implement merging options with default options.

apps/oxlint/src-js/plugins/config.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Called from Rust side after all configuration options have been resolved.
+ *
+ * Note: the name `setupConfigs` is currently incorrect, as we only populate rule options.
+ * The intention is for this function to transfer all configurations in a multi-config workspace.
+ * The configuration relevant to each file would then be resolved on the JS side.
+ *
+ * @param optionsJSON - JSON serialization of an array containing all rule options across all configurations.
+ * @returns "ok" on success, or error message on failure
+ */
+export function setupConfigs(optionsJSON: string): string {
```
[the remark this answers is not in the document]
overlookmotel: Your comment is not helpful, Mr Robot!

apps/oxlint/src/lint.rs
overlookmotel: I think the way you've done it is good!

Personally I find all the logic around config in linter labyrinthine and confusing/confused. You're asking about separation of concerns. The problem IMO is that there isn't any! Or at least if there is, it's not clear what it is (to me anyway).

apps/oxlint/src/run.rs
```diff
@@ -16,6 +16,30 @@ use crate::{
...
+pub fn set_external_options_json(plugin_store: &ExternalPluginStore) {
+    let _ = EXTERNAL_OPTIONS_JSON.set(plugin_store.serialize_all_options());
+}
+
+/// JS callable function to retrieve the serialized external rule options.
+/// Returns a JSON string of options arrays. Called once from JS after creating the external linter.
+#[cfg(all(feature = "napi", target_pointer_width = "64", target_endian = "little"))]
+#[napi]
+pub fn get_external_rule_options() -> Option<String> {
+    EXTERNAL_OPTIONS_JSON.get().cloned()
```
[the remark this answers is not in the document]
overlookmotel: Outdated.

apps/oxlint/src/run.rs
overlookmotel: It'd also be ideal to remove the `String` return value, since I'm suggesting there's no need for error handling.

I'm not entirely sure how to do that. Maybe `()` as the return type? If that doesn't work, maybe use `bool`, and always return `true` from `setupConfigs` on JS side (and ignore the return value on Rust side).

apps/oxlint/src/js_plugins/external_linter.rs
lilnasy: Yeah, Node.js docs say `NonBlocking` blocks "until space becomes available in the queue."
https://nodejs.org/api/n-api.html#:~:text=If%20set%20to%20napi_tsfn_blocking%2C%20the%20API%20blocks%20until%20space%20becomes%20available%20in%20the%20queue.

apps/oxlint/src-js/plugins/options.ts
lilnasy: It's not overwritten by `RuleTester`. I'll come back to this and review how `RuleTester` works.

apps/oxlint/src-js/plugins/options.ts
lilnasy: `cause` doesn't have to be an error. One common pattern is to have structured error details as `cause`.

apps/oxlint/src/run.rs
lilnasy: Yes, `()` works.

apps/oxlint/test/fixtures/custom_plugin_with_options/plugin.ts
lilnasy: Yeah, good catch.

apps/oxlint/src-js/plugins/options.ts
overlookmotel: Ah OK, true. Let's leave it as is for now then, and handle this in a follow-on PR.

lilnasy:
Something came up and there are a few comments I haven't addressed yet.

overlookmotel:
Thanks for addressing my feedback. We're doing a release in a few hours, and I'm really keen for this to be included in the release. I hope you don't mind, but I'm going to take over this PR and finish it up so can merge it ASAP. It doesn't look like there's much left to do.

I'll start by squashing commits and rebasing on latest main. Please don't push any more changes in the meantime.

apps/oxlint/src/lint.rs
overlookmotel: I've moved it down a bit.

crates/oxc_linter/src/config/config_store.rs
overlookmotel: I've made this change.

crates/oxc_linter/src/external_plugin_store.rs
overlookmotel: Let's leave this for a follow-on.

crates/oxc_linter/src/external_plugin_store.rs
overlookmotel: I've made a change to this effect.

We can leave altering the type of `ESLintRule::config` to `Option<Vec<serde_json::Value>>` to a follow-on.

overlookmotel (commented):
I've pushed some changes. All nits and code style apart from these ones which are more substantive:

* 330604e980ea3135875dbeb26ae12fa20839fd9d
* 63881d71c052a3a3ed96382166f1d1376c5287f7
* 8ecc0714244792d50ae00f782f967c7579e89bd3
* 8f2a6e282b2cb2b29170493ae44f972840b616e5
* 736841bf77f28ce4bd6da2b146083cd9c7ad0410

Even those ones are all still minor though.

Great work! This will be in today's release.

crates/oxc_linter/src/external_plugin_store.rs
overlookmotel: #16339

crates/oxc_linter/src/external_plugin_store.rs
overlookmotel: #16336

apps/oxlint/src-js/plugins/options.ts
overlookmotel: #16358

PR #16265, opened 2025-11-29
fix(traverse): fix panic when truncating non-ASCII variable names

tbvjaos510:
## Summary

Fix panic when `get_var_name_from_node` processes non-ASCII variable names (Korean, CJK, Greek, etc.).

## Problem

When transpiling code with non-ASCII variable names like Korean:

```typescript
const [서비스이용약관_query, 개인정보수집동의_query] = useQueries({...});
```
The following panic occurs:

`thread 'tokio-runtime-worker' panicked at oxc_traverse/src/ast_operations/gather_node_parts.rs:26:14:
assertion failed: self.is_char_boundary(new_len)`

## Cause

get_var_name_from_node was using byte-based truncate(20):

name.truncate(20);  // Cuts at byte 20, not character 20

Korean characters are 3 bytes in UTF-8, so "서비스이용약관" (7 chars = 21 bytes) gets cut in the middle of '관', causing the panic.

## Solution

Changed to character-based truncation to match https://github.com/babel/babel/blob/419644f27c5c59deb19e71aaabd417a3bc5483ca/packages/babel-traverse/src/scope/index.ts#L210:
```typescript
if name.len() > 20 {
    name = name.chars().take(20).collect();
}
```
The len() > 20 check avoids unnecessary allocation when string is already short enough (20 bytes guarantees ≤20 chars).

## Test Plan

Added tests for:
- 2-byte UTF-8 (Greek letters)
- 3-byte UTF-8 (Korean characters)
- 4-byte UTF-8
- Mixed ASCII + multi-byte

### DIFF
--- a/crates/oxc_traverse/src/ast_operations/gather_node_parts.rs
+++ b/crates/oxc_traverse/src/ast_operations/gather_node_parts.rs
@@ -22,8 +22,23 @@ pub fn get_var_name_from_node<'a, N: GatherNodeParts<'a>>(node: &N) -> String {
 
     if name.is_empty() {
         name = "ref".to_string();
-    } else {
-        name.truncate(20);
+    } else if name.len() > 20 {
+        // Truncate to ~20 bytes, respecting UTF-8 char boundaries.
+        // This diverges slightly from Babel (which limits to 20 chars),
+        // but the goal is just to avoid overly long variable names.
+        let bytes = name.as_bytes();
+        if bytes[19] < 0x80 {
+            // 20th byte is ASCII, safe to truncate here
+            name.truncate(20);
+        } else {
+            // Byte 19 may be in the middle of a multi-byte char.
+            // Walk back to find the start byte (continuation bytes are 10xxxxxx).
+            let mut truncate_at = 19;
+            while truncate_at > 0 && (bytes[truncate_at] & 0xC0) == 0x80 {
+                truncate_at -= 1;
+            }
+            name.truncate(truncate_at);
+        }
     }
 
     to_identifier(name)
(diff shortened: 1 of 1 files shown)


crates/oxc_traverse/src/ast_operations/gather_node_parts.rs
```diff
@@ -22,8 +22,8 @@ pub fn get_var_name_from_node<'a, N: GatherNodeParts<'a>>(node: &N) -> String {
 
     if name.is_empty() {
         name = "ref".to_string();
-    } else {
-        name.truncate(20);
+    } else if name.len() > 20 {
```
connorshea, line 25: Should probably have a comment here explaining why this needs to be done this way, if possible :)

tbvjaos510: Added a comment as you suggested! Thanks!

overlookmotel (commented):
Thanks for discovering this and for making a fix (and for adding lots of tests).

`Chars` iterator is pretty slow. I would suggest this instead:

* Check if 20th byte is ASCII.
* If it is, truncate to 20 bytes.
* If it isn't, find start byte of the character which byte 20 may be in middle of, and cut before that byte instead (i.e. maybe truncate to 19, 18, or 17 instead).

This is a slight divergence from Babel, but I think that's fine. The point is not to replicate the exact truncation strategy, but just not to produce massively long variable names.

---

Also, please add a test for 4-byte UTF-8 characters. That'll complete the set.

tbvjaos510:
@overlookmotel 
Thanks for the suggestion! Updated to use byte-based truncation instead of Chars iterator. Also added a test for 4-byte UTF-8 characters (CJK Extension B).

overlookmotel (approved):
Thank you!

Issue #16207, opened 2025-11-27
Linter plugins: Rustify token methods


overlookmotel:
Continuation of #14829.

All tokens APIs are now implemented (#15861, and many later PRs - thanks @lilnasy!).

We are currently using TS-ESLint to re-parse source and generate tokens. This works, but is very slow, and bloats the size of `oxlint` package. We should Rustify it!

Am copying content from #14829 to this new issue, to avoid needing to wade through the whole thread there.

---

It's going to be challenging.

## Altering the lexer

Oxc's [lexer](https://github.com/oxc-project/oxc/tree/main/crates/oxc_parser/src/lexer) generates a stream of `Token`s, however, they're not stored anywhere. We'd need to adapt it to also store all the `Token`s in an `oxc_allocator::Vec` in the arena. That `Vec` can then be shared with JS through the same "raw transfer" mechanism as is used for the AST.

One difficulty is that the lexer sometimes has to unwind to where syntax is ambiguous reading forwards through the file. e.g. is `(a, b)` a `SequenceExpression`, or the start of an arrow function `(a, b) => ...`? When rewinding occurs, we'd need to also rewind the stream of `Token`s so they don't get included twice in these cases.

## Avoiding perf hit

Oxc's parser is not only used in the linter - it's the first step of the parse-transform-minify-print and parse-format pipelines. Its performance in these pipelines is **critical**, and hurting perf while implementing what we need for linter would be unacceptable.

We'll need some kind of conditional compilation, so linter uses a different version of the parser from other pipelines.

Cargo features are problematic, due to feature unification in tests, so I would propose an approach using generics:

```rs
// ----- Traits -----

trait ParserConfig {
    type Tokens<'a>: TokenStore<'a>;
}

trait TokenStore<'a> {
    type Checkpoint;

    fn new(allocator: &'a Allocator) -> Self;
    fn push(token: Token);
    fn checkpoint(&self) -> Self::Checkpoint;
    fn rewind(&mut self, checkpoint: Self::Checkpoint);
}

// ----- parse-transform-minify-print pipeline -----

struct StandardParserConfig;

impl ParserConfig for StandardParserConfig {
    type Tokens<'a> = NoopTokenStore;
}

struct NoopTokenStore;

impl<'a> TokenStore<'a> for NoopTokenStore {
    type Checkpoint = ();

    fn new(_allocator: &'a Allocator) -> Self { Self }
    fn push(_token: Token) {}
    fn checkpoint(&self) {}
[...]

PR #16283, opened 2025-11-29
perf(lexer): store escaped identifiers in a `Vec`

lilnasy:
Closes #16208.


crates/oxc_parser/src/lexer/string.rs
```diff
@@ -246,13 +246,18 @@ impl<'a> Lexer<'a> {
 
     #[cold]
     fn save_escaped_string(&mut self, s: &'a str) {
-        self.escaped_strings.insert(self.token.start(), s);
-        self.token.set_escaped(true);
+        self.escaped_strings.push(s);
+        // We are _probably_ not going to have to deal with more than 4.3 billion escaped
+        // identifiers in a single file
```
overlookmotel: It's impossible to overflow `u32` here. Source text is bounded to `u32::MAX` bytes maximum. So it's impossible for source to contain more than `u32::MAX / 5` escaped strings (the most you could get is `"\n","\n","\n","\n","\n",...`).

crates/oxc_parser/src/lexer/string.rs
```diff
@@ -246,13 +246,18 @@ impl<'a> Lexer<'a> {
...
+        #[expect(clippy::cast_possible_truncation)]
+        let index = (self.escaped_strings.len() - 1) as u32;
```
overlookmotel: Oh actually we could avoid initializing the `Vec`s with a single dummy entry and just store `self.escaped_strings.len()` here - i.e. `index + 1`. Then when retrieving the string from the `Vec`, deduct 1 from index there instead.

crates/oxc_parser/src/lexer/token.rs
```diff
@@ -149,18 +149,21 @@ impl Token {
...
-        self.0 |= u128::from(escaped) << ESCAPED_SHIFT;
+    pub fn escape_index(&self) -> u32 {
+        ((self.0 >> ESCAPE_INDEX_SHIFT) & ESCAPE_INDEX_MASK) as u32
+    }
+
+    #[inline]
+    pub(crate) fn set_escape_index(&mut self, index: u32) {
+        self.0 &= !(ESCAPE_INDEX_MASK << ESCAPE_INDEX_SHIFT); // Clear current `escape_index` bits
+        self.0 |= u128::from(index) << ESCAPE_INDEX_SHIFT;
     }
```
overlookmotel, line 169: It'd be better to put the `u32` for escaped index last (`ESCAPED_SHIFT = 96`). Then compiler will be able to optimize these 2 functions better, as the `u32` has proper alignment (aligned on a 4-byte boundary).

overlookmotel (commented):
Nice work. A few comments below.

Run `just allocs` to fix the CI fail - this alters the number of allocations made in parser, I guess.

crates/oxc_parser/src/lexer/string.rs
lilnasy: That makes sense, thanks

crates/oxc_parser/src/lexer/token.rs
lilnasy: Oh wow, 3% speedup from just this change!

crates/oxc_parser/src/lexer/token.rs
```diff
@@ -36,7 +36,7 @@ const _: () = {
...
 
     assert!(is_valid_shift(IS_ON_NEW_LINE_SHIFT));
-    assert!(is_valid_shift(ESCAPED_SHIFT));
+    assert!(is_valid_shift(ESCAPE_INDEX_SHIFT));
```
overlookmotel: You can delete this line entirely. It's only relevant for `bool` fields.

crates/oxc_parser/src/lexer/token.rs
```diff
@@ -53,7 +53,7 @@ impl Default for Token {
         // end: 0,
         // kind: Kind::default(),
         // is_on_new_line: false,
-        // escaped: false,
+        // escape_index: 0,
```
overlookmotel: Nit: Please move this down to bottom of the list, so the list has same order as the fields in `Token`.

crates/oxc_parser/src/lexer/token.rs
```diff
@@ -68,7 +68,7 @@ impl fmt::Debug for Token {
             .field("start", &self.start())
             .field("end", &self.end())
             .field("is_on_new_line", &self.is_on_new_line())
-            .field("escaped", &self.escaped())
+            .field("escape_index", &self.escape_index())
```
overlookmotel: Ditto. Please move to bottom of the list.

overlookmotel (commented):
Thanks for doing this.

The odd thing is that while there's a large improvement on 1 of the lexer benchmarks, the parser benchmarks (which is what matters) actually regress a tiny bit (sub-1% change). It's not the boost I'd hoped for. The mysteries of the compiler 🤷...

But we're within the noise threshold, so it could not indicate very much. It'll be interesting to see if the benchmarks stay the same when you push again, and they re-run.

overlookmotel:
Just to let you know, reason I've stalled on this is (a) unfortunately it's not showing the perf improvement in parser that I hoped for, and (b) we may need those 4 spare bytes in `Token` for something else - https://github.com/oxc-project/backlog/issues/193#issuecomment-3606557203 - but that's not clear yet.

In any case, the original motivation for this was to enable unescaping identifiers in tokens in linter (#16207), but it turns out TS-ESLint doesn't do that, so maybe we don't need to either.

lilnasy:
I thought that might be the case. My original motivation was to find a good onboard to lexer internals, and I've gotten it!

PR #16291, opened 2025-11-30
feat(semantic): ScopeFlags::With

aapoalas:
Simple addition but unfortunately untested: I locally fail tsgo related tests, probably for reasons related to not having tsgo installed?


crates/oxc_syntax/src/scope.rs
```diff
@@ -115,6 +116,7 @@ impl ScopeFlags {
         self.contains(Self::CatchClause)
     }
 
+    #[inline]
```
aapoalas, line 119: **note:** Unrelated to the PR itself but all the other `is_X` methods had this, only this one was missing it.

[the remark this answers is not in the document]
aapoalas: Mentioned in separate comment.

crates/oxc_semantic/src/builder.rs
```diff
@@ -1715,7 +1715,7 @@ impl<'a> Visit<'a> for SemanticBuilder<'a> {
         /* cfg */
 
         self.visit_expression(&stmt.object);
-        self.enter_scope(ScopeFlags::empty(), &stmt.scope_id);
+        self.enter_scope(ScopeFlags::With, &stmt.scope_id);
```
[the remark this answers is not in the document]
aapoalas: Added the flag into the scope attribute. CatchClause does both the scope attribute and `enter_scope(ScopeFlags::CatchClause, ..)` so did not remove the "manual setting" in the builder.

crates/oxc_syntax/src/scope.rs
lilnasy: Good improvement! This was added in [#9567](https://github.com/oxc-project/oxc/pull/9567/files#diff-61df19c7f8d9efbddd80f401c5c04d6111bafba5fa7fba5335718f13cf6a695d). I don't see a reason why it shouldn't be consistent.

crates/oxc_semantic/src/builder.rs
lilnasy, line 1718: Following references of `ScopeFlags.is_empty()`: it is referenced only by `ScopeFlags.is_block()`, which is referenced only by this JSX transform: 
https://github.com/oxc-project/oxc/blob/6a641f937c7e725c6938765d52060643ed37edff/crates/oxc_transformer/src/jsx/jsx_self.rs#L72

The expected behavior of the transform is explained here: https://github.com/babel/babel/blob/v7.26.2/packages/babel-plugin-transform-react-jsx-self/src/index.ts#L2-L5.

Now that with-blocks are no longer considered "blocks", the transform works incorrectly in a (admittedly highly specific) case.

For JSX elements inside with-blocks inside constructors, the loop over scope ancestors stops at the with-block and `is_inside_constructor()` incorrectly returns `false`. `__self={this}` attribute shouldn't be added in this case because the `this` value is the same inside a with-block and outside it, but it is.

I don't know if this matters, certainly not to nova, but I figured it's worth pointing out.

crates/oxc_syntax/src/scope.rs
lilnasy: Good improvement! This was added in [#9567](https://github.com/oxc-project/oxc/pull/9567/files#diff-61df19c7f8d9efbddd80f401c5c04d6111bafba5fa7fba5335718f13cf6a695d). I don't see a reason why it shouldn't be consistent.

lilnasy (commented):
Looks good to me! I need to review `oxc_semantic` for an upcoming feature, and I figured this would be a good opportunity. I am not too familiar with it yet so take my comments with a grain of salt.

crates/oxc_semantic/src/builder.rs
overlookmotel: Good find!

However, in this specific case, I don't think it's possible for a bug to occur because `with` is illegal in strict mode, and the body of classes is always strict mode. So the behavior is correct - if you hit a `with` scope, you can stop searching up the scope tree, as it's guaranteed you can't be in a class constructor.

But... I'd be surprised if there's nowhere else in the entire codebase which checks `ScopeFlags` and will behave wrongly after this change. But they'll be hard to find, as I doubt there are many usages of `with` in conformance tests.

crates/oxc_syntax/src/scope.rs
overlookmotel: `with` is complicated, and it's possible this PR may stall. So I've made broken out this change into #16309.

overlookmotel (commented):
`with` complicates symbol resolution a lot. Is there anything else we need to do to make `Semantic` make sense of it?

For example, I guess minifier should not mangle var names where they're accessed "through" a `with`, in the same way that it I think takes account of whether a var can be "seen" by an `eval()`.

e.g. `let foo = 1; with ({foo: 2}) foo = 3;` would change behavior if `foo` var is renamed. Though maybe that's such an edge case, we shouldn't worry about it.

Out of interest, what's your use case for this in Nova? My sincere condolences if you actually need to make `with` behave correctly. Nightmare!

aapoalas:
> `with` complicates symbol resolution a lot. Is there anything else we need to do to make `Semantic` make sense of it?
> 
> For example, I guess minifier should not mangle var names where they're accessed "through" a `with`, in the same way that it I think takes account of whether a var can be "seen" by an `eval()`.
> 
> e.g. `let foo = 1; with ({foo: 2}) foo = 3;` would change behavior if `foo` var is renamed. Though maybe that's such an edge case, we shouldn't worry about it.
> 
> Out of interest, what's your use case for this in Nova? My sincere condolences if you actually need to make `with` behave correctly. Nightmare!

`with` is actually really simple! That is, from an engine point of view _if_ the engine is not interested in doing any optimisations. I have like 90% of `with` done and am only missing the bytecode instruction for it because I'm lazy and don't like it as a feature :D

The use-case I have is actually the same that you describe, basically: I am currently writing a sorts of "runtime minification" (or just an optimisation) probably better known as register allocation, wherein I'd like to rewrite eg. the `foo` variable in your example into a stack slot if the transformation is sound. In a `with` scope, such a transformation is not sound as your example illustrates, so _if_ I had proper `with` support then I'd need to be able to detect them for this optimisation to be soundly applied.

Currently, I am applying this optimisation based on the Scoping information mainly, hence my interest in the flag :)

Boshen:
Didn't find any problems from PR review, merge so we can get a release out next week.

PR #16403, opened 2025-12-02
fix(napi/parser): reorder fields of `ExportEntry` on JS side

overlookmotel:
NAPI-RS 3.6.0 makes a change to the ordering of objects. It now puts optional fields last.

This made tests in `napi/parser` for module record fail when we tried to update (#16383) because the `module_request` field of `ExportEntry` moves to last which breaks the snapshots.

The change in NAPI-RS is unlikely to be reverted, because it's a sizeable perf optimization: https://github.com/napi-rs/napi-rs/pull/2990

To work around this problem:

1. Move the field in the `ExportEntry` intermediate struct in `napi/parser` to last, so NAPI's output matches what you'd expect from the struct definition.
2. Alter the `#[estree]` attr on `ExportEntry` struct in `oxc_syntax` crate to match.

Note: The `ExportEntry` struct in `napi/parser` is just an intermediate structure used for serialization. So I think it's fine to fiddle with its field order. The actual `ExportEntry` struct used in the module record is in `oxc_syntax` crate, and it remains unaltered.

PR #16411, opened 2025-12-02
feat(napi/parser)!: represent empty optional fields on JS side as `null`

overlookmotel:
Follow-on after #16383 and #16403.

While working on those I discovered an option NAPI-RS has to represent `Option::None` as `null`, instead of omitting the field entirely.

Use that option on the structs used for transferring errors and module record over to JS.

I think this is likely more performant because:

1. NAPI-RS can create all properties of the objects using only the faster `node_api_create_object_with_properties` API.
2. It produces consistent object shapes, so JS engine can better optimize code using these objects.

It also brings the shape of the data perfectly into line between standard transfer and raw transfer, and is consistent with how empty fields in the AST are represented as `null`.

I would have used this option before if I'd known it existed.

### Breaking change

I've marked this as a breaking change because code consuming these objects would now need to check for empty fields with `value === null` instead of `value === undefined`.

However in practice, most people probably use `!value` or `value ?? ...`, so it's unlikely to affect many users. This only affects module record and errors anyway, not the AST itself, as we transfer that as JSON, not via NAPI.

### DIFF
--- a/crates/oxc_napi/src/error.rs
+++ b/crates/oxc_napi/src/error.rs
@@ -4,7 +4,7 @@ use napi_derive::napi;
 
 use oxc_diagnostics::{LabeledSpan, NamedSource, OxcDiagnostic};
 
-#[napi(object)]
+#[napi(object, use_nullable = true)]
 pub struct OxcError {
     pub severity: Severity,
     pub message: String,
@@ -64,7 +64,7 @@ impl From<&OxcDiagnostic> for OxcError {
     }
 }
 
-#[napi(object)]
+#[napi(object, use_nullable = true)]
 pub struct ErrorLabel {
     pub message: Option<String>,
     pub start: u32,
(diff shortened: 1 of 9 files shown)


overlookmotel:
* **#16412** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16412?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16411** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16411?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16411?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=16411).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #16416, opened 2025-12-03
perf(all): experiment: add dedicated identifier string type

camchenry:
- Related to https://github.com/oxc-project/backlog/issues/193

This is a proof of concept. Mainly to see how bad the damage is if we add a new `Ident` type. Will help guide the API and also how we approach implementing this.


camchenry:
* **#16416** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=16416).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camchenry:
Like 1-3% regression in performance so far. Not too bad actually for a completely naive first implementation. I'm going to try and make some smaller refactors to the codebase first though.
The real improvement should come when we take advantage of the hash function for hash maps.

image

crates/oxc_span/src/ident.rs
```diff
@@ -0,0 +1,323 @@
...
+#[derive(Clone, Copy, Eq)]
+pub struct Ident<'a> {
+    ptr: NonNull<u8>,
+    // Length in bottom 32 bits, hash in top 32 bits
+    len_and_hash: u64,
+    _marker: PhantomData<&'a str>,
+}
+
+impl<'a> Ident<'a> {
+    pub fn new(s: &str) -> Self {
```
overlookmotel: This is unsound at present, as it silently extends the lifetime.

Currently I think it allows this:

```rs
let ident = {
  let s: String = "hello".to_string();
  Ident::new(s.as_str())
};
// Use after free!
println!("{}", ident.as_str());
```

```suggestion
    pub fn new(s: &'a str) -> Self {
```

I mention this now because the lifetime constraint may require further refactoring of calling code (and `AsRef` / `Defer` impls).

camchenry: Good catch! This actually seemingly only required a small change to some code in the transformer, where we were passing in a reference to an Atom, instead of the Atom itself. Apart from that, it seemed to pass the borrow checker just fine.

camchenry:
@overlookmotel I ended up making a lot of changes I didn't mean to, so please excuse some of the unnecessary changes. I think I've committed a crime on our codebase. However, the main point is that I converted some of the usage of `FxHashMap<Atom, _>` to use `IdentHashMap<_>` instead. I did not include the arena-allocated hash maps as that presents quite a challenge in implementation, though I think that's where some of the most valuable work might be.

Regardless, we are starting to at least some positive benefits that we would expect. It's essentially breakeven once you take into account the additional parsing time, but it's 2-3% faster in semantic already: 
image

crates/oxc_span/src/ident.rs
```diff
@@ -0,0 +1,384 @@
...
+        let ptr = NonNull::from(s).cast::<u8>();
+
+        // Produce a hash of the string
+        // NOTE: This is creating a fresh hasher for each identifier, which is probably bad for performance?
+        // But, I want to see how terrible it is and keep the API simple for testing.
+        let hash = {
+            let mut hasher = FxHasher::default();
+            s.hash(&mut hasher);
+            hasher.finish()
+        };
```
overlookmotel: Creating a hasher costs nothing. Do "jump to definition" on `FxHasher::default()` and you'll see.

overlookmotel: I just realized something. `hasher.finish()` is rotating the bits (again, do "jump to definition"). That's not what we want to happen as we want the bits with highest entropy in top 32 bits - we do the rotation ourselves in `Hash::hash` instead.

So I'd suggest undoing the rotation again here.

```rs
let hash = {
    let mut hasher = FxHasher::default();
    s.hash(&mut hasher);
    hasher.finish()
};

// `FxHasher::finish` performs a rotation.
// That's not what we want as we want the highest entropy bits in top 32 bits.
// Undo that rotation here by rotating back in the opposite direction.
// This code is the exact reverse of the code in `FxHasher::finish`,
// so compiler should see that together they make a no-op, and remove both rotations.
#[cfg(target_pointer_width = "64")]
const ROTATE: u32 = 26;
#[cfg(target_pointer_width = "32")]
const ROTATE: u32 = 15;

let hash = hash.rotate_right(ROTATE);
```

If we have a higher-entropy hash, that should mean less collisions. If so, it might move the semantic benchmarks a bit.

crates/oxc_span/src/ident.rs
```diff
@@ -0,0 +1,384 @@
...
+// TODO: Add rest of `FromIn` impls for `Ident`
+
+impl<'new_alloc> CloneIn<'new_alloc> for Ident<'_> {
+    type Cloned = Ident<'new_alloc>;
+
+    #[inline]
+    fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
+        Ident::from_in(self.as_str(), allocator)
+    }
+}
```
overlookmotel, line 249: This is the place where re-hashing will be happening unnecessarily in semantic when copying from 1 arena to another.

I think this should do the trick - copy string, but don't recalculate hash:

```rs
impl<'new_alloc> CloneIn<'new_alloc> for Ident<'_> {
    type Cloned = Ident<'new_alloc>;

    #[inline]
    fn clone_in(&self, allocator: &'new_alloc Allocator) -> Self::Cloned {
        let s = allocator.alloc_str(self.as_str());
        let ptr = NonNull::from(s).cast::<u8>();
        Ident { ptr, len_and_hash: self.len_and_hash, _marker: PhantomData }
    }
}
```

If there are any other methods which create an `Ident` from another `Ident`, they should do the same - but I think this is maybe the only one?

crates/oxc_span/src/ident.rs
```diff
@@ -0,0 +1,384 @@
...
+        // where N is the exponent of number of buckets in the hash table.
+        // So just rotate by 25 bits, so 32-bit hash occupies top 7 bits
+        // and bottom 25 bits. This gives good entropy as long as HashMap
+        // contains less than 58 million entries ((1 << 25) * 7 / 8).
+        // The bits in the middle have low entropy, but that doesn't matter
+        // because `hashbrown` will ignore them anyway.
+        let hash = self.len_and_hash.rotate_left(32 - 7);
+        hasher.write_u64(hash);
+    }
+}
```
overlookmotel, line 275: All these small methods should have `#[inline]` on them. Probably compiler will inline them anyway, but just to make sure.

crates/oxc_span/src/ident.rs
```diff
@@ -0,0 +1,384 @@
...
+            hasher.finish()
+        };
+
+        // With FxHasher, highest entropy is in top 32 bits. Clear bottom 32 bits.
+        let hash = hash & !(u32::MAX as u64);
+        // We know `s.len()` is <= u32::MAX so don't bother masking it
+        let len = s.len() as u64;
+
+        // Identifiers cannot have zero size.
+        assert!(len > 0, "identifiers cannot have zero length");
```
overlookmotel: For purposes of benchmarking, I'd suggest commenting out this assertion, or making it `debug_assert!`. We'll find ways to guarantee `len > 0` (and `len <= u32::MAX`) at call sites, so eventually we won't need this assertion and it'll still be safe.

camchenry: Good point. I did notice that running `just coverage` did trip this assertion, which is interesting, but I'll move it to `debug_assert!` for now since it's not critical for this current work.

crates/oxc_span/src/ident.rs
camchenry: Ah, good to know that doing `FxHasher::default()` is actually not doing anything. If we end up porting the fxhash code into `oxc`, hopefully we can make the rotation simpler (by just not doing it), and not need to rely on this hopefully turning into a no-op.

crates/oxc_span/src/ident.rs
overlookmotel: Odd. We should investigate why this causes a fail on conformance. But that's for another time...

crates/oxc_span/src/ident.rs
overlookmotel: Yes this is a bit of a hack. We will want to pull `FxHasher`'s code into Oxc as you say. But I'd be very confident compiler will remove it, as long as `FxHasher::finish` gets inlined (which it really should). Compiler is really good at simple optimizations like this.

https://godbolt.org/z/334EEeKvf

Anyway, it's only 2 ops if it doesn't. The thing that *might* have an effect on perf is if we get less hash collisions.

Boshen:
I'll move this forward.

Boshen:
closed by the stack https://github.com/oxc-project/oxc/pull/18400

Issue #13651, opened 2025-09-10
minifier: inline functions that return a const value


IWANABETHATGUY:
for code: 
```js
{
  let AOT = true;
  const getAOT = /* @__NO_SIDE_EFFECTS__ */ () => AOT;
  // const setAOT = () => {
  //   AOT = false;
  // };
  // setAOT();
  const fn = getAOT() ? () => console.log("Hi") : () => {};
  fn();
}
```
### Swc

https://playground.oxc.rs/#eNptUm1r2zAQ/itCn5ISlm0wGBndC03MAlsz1rJPAaPaZ0+bLBndOc0w+e87SZbbD/1k6U7Pyz3nUVZyI8ejFcIAiS+He3EtyA/wIZQqZ5FEC5Tq6yvxuSxvD+Xdfrsrd0Wxu7m/K0txtRaLpbj+GOARt15PUMzQ1I863BSTUKMMJiUuXvIhgRbLZxYay6+TD2b6NNGFnjPwyrh2cZRf9VEuxSZLJbrGRp6LXEknN6P0gw0foy3JTZhzJRvnO0UEXm6in2eV/VONvLIYGnNFs7YiqLdQGeUVaXYzNx9/awLsVQVZplO2NfOtcl3vARmQ7li5fm7iv+7BmXS7rGSvPAZ3o4QzgUUW4p0RnnkoZYx7/Ak0eHsYCHUNxWCr4CWTBRnwJ/ihPGNzNeJ+vd9b8poZq6nBaiGaqMbneegkT8rzDlgb8O3rN+9YfkDYQqMtFM7fGIVYaDD1rALnHrzuwJIyHJPjlJx/6naa5up3IFUrUrOPl+MdJbLjnn2Dt4ozinnz88rV0EJcrg3rm+JbyT9YO/7Hp7GtdRS5csVAm99GlryWMH5eGU/u+m9wgpn0L0B/qzrIC49QztKZgnMNiBP4B4cMTv0QJgeXXV7+A+YfLck=

**output**:
```js
console.log("Hi");
```
### Oxc

[...]

PR #16477, opened 2025-12-04
feat(minifier): fold iife arrow functions in call expressions

armano2:
Improve folding of iife 

## details

- check all `CallExpression` if they can be collapsed
  eg. passed as arguments or in assignment
- functions without arguments and body can be collapsed to to `void 0`
 `(() => {})()` or `(function () { })()` to `void 0`
- with one statement and without return will be collapsed to `(c, void 0)`
  `(() => { fn() })()` to `(fn(), void 0)`
- `FunctionExpression` is not supported
  `(function () { ... })()`

## notes

- function with arguments could be handled by first in-lining them
- empty function with arguments could be also if all arguments do not have defaults
- we should consider optimizing function definitions by removing not used arguments
- initially i added support for `FunctionExpression` but there is no way to track `this` usage, and thus we are unable to minify this

## change example

code
```js
{
  let AOT = true;
  const getAOT = /* @__NO_SIDE_EFFECTS__ */ () => AOT;
  const fn = getAOT() ? () => console.log("Hi") : () => {};
  fn();
}
```

currently is collapsed to 

```js
(/* @__PURE__ */ /* @__NO_SIDE_EFFECTS__ */ (() => !0)() ? () => console.log("Hi") : () => {})();
```

and with this change

```js
console.log('Hi');
```

------------------

from what i can see @sapphi-red worked on this file


armano2:
i had to split logic a little as i was unable to make additional tests pass for pure functions

```rs
(if let Expression::Identifier(id) = &call_expr.callee
                    ctx.state.pure_functions.contains_key(&symbol_id)
```

i do not think we should use pure information from functions in place when we use them

crates/oxc_minifier/tests/peephole/inline_single_use_variable.rs
```diff
@@ -226,20 +226,20 @@ fn test_inline_single_use_variable() {
...
     // cannot be compressed to `function wrapper() { let foo; return foo.bar = (() => { foo = {} })(), foo }`
     test(
         "function wrapper() { let foo; const bar = (() => { foo = {} })(); foo.bar = bar; return foo }",
-        "function wrapper() { let foo, bar = (() => { foo = {} })(); return foo.bar = bar, foo }",
+        "function wrapper() { let foo, bar = (foo = {}, void 0); return foo.bar = void 0, foo }",
     );
     test(
         "function wrapper() { let foo = {}; const bar = (() => { console.log() })(); foo.bar = bar; return foo }",
-        "function wrapper() { let foo = {}; return foo.bar = (() => { console.log() })(), foo }",
+        "function wrapper() { let foo = {}, bar = (console.log(), void 0); return foo.bar = void 0, foo; }",
```
armano2, line 238: this result is surprisingly longer, but i think this is correct, `bar` should be removed here

armano2:
i noticed a bug in this implementation
`this` is not handled inside `FunctionExpression`, i was hoping to used references, but it looks like they `this` is not defined as such

armano2:
@sapphi-red thank you for running tests issues reported by it should be fixed

is there any guide what tests we have to run localy?

sapphi-red:
Running the tests ran by the CI should be fine.

armano2:
Conformance tests should be ok, its interesting that 
- `Object.getPrototypeOf(async function () {}())` = Promise
- `Object.getPrototypeOf(async function* () {}())` = AsyncGenerator
- `Object.getPrototypeOf(async function* () {}())` = Generator

i'm not sure tho if this is this is the best way to check if we can collapse pife/iife by evaluating ancesotrs and checking if they are `blocks`

i did some cleanup and sorry for keeping it open while it was still WIP

armano2:
if you find some time, review would be appreciated

crates/oxc_minifier/tests/peephole/inline_single_use_variable.rs
armano2: if run with all optimizations this actually gets simplified nicely.
```js
function wrapper() {let foo = {};return console.log(), foo.bar = void 0, foo;}
```

instead of
```js
function wrapper() {let foo = {};return foo.bar = (() => {console.log();})(), foo;}
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
             return false;
         }
 
+        !e.may_have_side_effects(ctx)
+    }
+
+    pub fn fold_iife_expression(e: &mut Expression<'a>, ctx: &mut Ctx<'a, '_>) {
```
sapphi-red: Would you move this function to `substitute_alternate_syntax.rs` and name it `substitute_iife_call` since it is now also used in cases where the return value is being used?

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
+                        && f.body.is_empty()
+                        && (!f.r#async
+                            || ctx.ancestors().nth(1).is_some_and(Ancestor::is_parent_of_statement))
                 }
-                Expression::ArrowFunctionExpression(f) => f.params.is_empty() && f.body.is_empty(),
                 _ => false,
             };
             if is_empty_iife {
-                return true;
+                *e = ctx.value_to_expr(call_expr.span, ConstantValue::Undefined);
```
sapphi-red:
```suggestion
                *e = ctx.ast.void_0(call_expr.span);
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
+        let Expression::CallExpression(call_expr) = e else { return };
+
         if call_expr.arguments.is_empty() {
             let is_empty_iife = match &call_expr.callee {
                 Expression::FunctionExpression(f) => {
-                    f.params.is_empty() && f.body.as_ref().is_some_and(|body| body.is_empty())
+                    f.params.is_empty()
+                        && f.body.as_ref().is_some_and(|body| body.is_empty())
+                        && (!f.r#async && !f.generator
+                            || ctx.ancestors().nth(1).is_some_and(Ancestor::is_parent_of_statement))
```
sapphi-red:
```suggestion
                            || ctx.ancestors().nth(1).is_some_and(Ancestor::is_parent_of_statement)) // return value is not used
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
+                }
+                Expression::ArrowFunctionExpression(f) => {
+                    f.params.is_empty()
+                        && f.body.is_empty()
+                        && (!f.r#async
+                            || ctx.ancestors().nth(1).is_some_and(Ancestor::is_parent_of_statement))
```
sapphi-red:
```suggestion
                            || ctx.ancestors().nth(1).is_some_and(Ancestor::is_parent_of_statement)) // return value is not used
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
-                    return Self::remove_unused_expression(e, ctx);
+                    ctx.state.changed = true;
+                    return;
                 }
                 match &mut f.body.statements[0] {
                     Statement::ExpressionStatement(expr_stmt) => {
-                        // Replace "(() => { foo() })" with "foo()"
-                        *e = expr_stmt.expression.take_in(ctx.ast);
-                        return Self::remove_unused_expression(e, ctx);
+                        // Replace "(() => { foo() })" with "(foo(), undefined)"
```
sapphi-red:
```suggestion
                        // Replace "(() => { foo() })()" with "(foo(), undefined)"
```
(was a typo before this PR)

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
+                        *e = ctx.ast.expression_sequence(expr_stmt.span, {
+                            let mut sequence = ctx.ast.vec();
+                            sequence.push(expr_stmt.expression.take_in(ctx.ast));
+                            sequence
+                                .push(ctx.value_to_expr(call_expr.span, ConstantValue::Undefined));
```
sapphi-red:
```suggestion
                                .push(ctx.ast.void_0(call_expr.span));
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -562,50 +563,71 @@ impl<'a> PeepholeOptimizations {
...
+                            sequence
+                        });
+                        ctx.state.changed = true;
                     }
                     Statement::ReturnStatement(ret_stmt) => {
                         if let Some(argument) = &mut ret_stmt.argument {
                             // Replace "(() => { return foo() })" with "foo()"
```
sapphi-red:
```suggestion
                            // Replace "(() => { return foo() })()" with "foo()"
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -1008,20 +1031,35 @@ mod test {
...
-        // test("(async () => { let b = a; b() })()", "(async () => a())();");
-        // test("var a = (function() {})()", "var a = /* @__PURE__ */ function() {}();");
+        test("(async () => { let b = a; b() })()", "(async () => { a() })();");
+        test("var a = (function() {})()", "var a = void 0;");
+        test("a((() => b())());", "a(b())");
+        test("a((() => true)());", "a(!0)");
+        test("a((() => { return true })());", "a(!0)");
+
+        test_same("var a = (function () { b() })()");
+        test_same("var a = (function () { return b() })()");
```
sapphi-red, line 983:
```suggestion
        test_same("var a = (function () { return b() })()");
        test_same("var a = (function () { return this })()");
        test_same("var a = (function () { return arguments })()");
        test_same("var a = (function () { return new.target })()");
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
```diff
@@ -1008,20 +1031,35 @@ mod test {
...
+
         test("(() => x)()", "x;");
-        test("/* @__PURE__ */ (() => x)()", "");
+        test("(() => { return x })()", "x;");
+        test_same("(function () { return x })()");
+
+        test_same("var a = /* @__PURE__ */ (() => x)(y, z)");
+        test("(/* @__PURE__ */ (() => !0)() ? () => x() : () => {})();", "x();");
+        test("/* @__PURE__ */ (() => x)()", "x;");
+        test("/* @__PURE__ */ (() => { return x })()", "x;");
```
sapphi-red:
```suggestion
        test("(/* @__PURE__ */ (() => !0)() ? () => x() : () => {})();", "");
        test("/* @__PURE__ */ (() => x)()", "");
        test("/* @__PURE__ */ (() => { return x })()", "");
```
We should fix this as the output will be bigger in cases like:
```js
var a = /* @__PURE__ */ (() => {
  x;
  y;
  z;
})()

// a was not used
```

sapphi-red (changes requested):

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
armano2: thx, didn't know about those helpers, i'm still wrapping my head around all of them :)

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
armano2: do you mean to add comment? as value is actually used `is_parent_of_statement` returns `bool` and we are using it for condition check

armano2: if result of call is used in any way by checking if parent is a block like statement

eg, this is needed to get rid of 
```js
(async () => {})
// ....
{
  (async () => {})
}
```

i could remove this as this case should be really rare

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
armano2: all except this one seem ok,

```
test("(/* @__PURE__ */ (() => !0)() ? () => x() : () => {})();", "");
```
only `(() => !0)()` is pure

armano2: and i'm going to add one more test case
```js
var a = /* @__PURE__ */ (() => x)()
var a = x
```

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
armano2: as for tests, should i move them to?

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
sapphi-red: Yes, I meant to add comment to clarify that this function call is to check whether the return value of the IIFE call is not used. Is my understanding correct?
For example, I guess `(async () => {})()` without assignment or not passed to the argument are the cases that matches this condition.

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
sapphi-red: Hmm, I guess it's ok to keep it in the current file.

crates/oxc_minifier/src/peephole/remove_unused_expression.rs
armano2: correct, comment applied

armano2:
i was thinking abot expanding this logic to functions with arguments, but that is going to come as separate pr

sapphi-red (approved):
Thank you!

sapphi-red:
We cannot close #13651 as it requires more optimizations.
For example, I think this PR won't fix cases like: [Oxc](https://playground.oxc.rs/#eNptVGFP2zAQ/SuRPwHqgCFNQp3YhqDVkBhlKxpfkConuQQXx87sS1sU9b/vOY3TfuBLa7/LvXv37pJWZGIs2heTJJo4uZ49JVcJu4a+BiizxnNSEu/ws5Pkx2LxMFvM724ni8l0Orl5mi8WyclZcnScXH0L6V1e0ZiMlTVJYS0iHX1kKwyYdpQIfe8zQ8xqOtW2PHoRP9WLOE7GfazddqSgNUfH3XEbftbK5HZ9ihJdyVAJQYTESFgxboVrTPjTyrAYh55GorCukszkxLiQ2gNhJ40P8IAoCJFM+S1lWjoZ+vBDcP2qmHwtMxqgSppS76+ZrWpHHim7kj6zNaL95b1KrY63rCj7tO1I1NL5oKsVtGEyHmUxGfYbtCO1tus/xI0zs4a9ymnaOxypQklyK3qUDrkR7fL+Xt4ZdgqM2V4UQTWrbOKcdT0KDcGqTgPOB1a1ovH0JNO9DSzTZ5XzqxhfjASZfFbcK4MuhS4gt0Y17uOX56imgkO/G8t7m5Z+M/8A/hduj87WqCWk/2SIcsrBiTkpyCtvbFXJENMaKPpQQ7PoZR37R3wt3z0eSZ3M3ojnmBnS48MRlRXtlPf1bbqkjJ+drMERTQ2FOq3XDCPTBgrJHaTBrWGNdn6xdFhxUJC/OP/8BQRw8JYK5Eytu9HS+6kinQ8DoU1NTlVkWGosnsXeDYNBtFI8oL+IZS4ZJvRT+3hhW+EhtsbsyRmJnYtSM5tTSd2rYcKMh3Vc+tzia9C7aYzljisimsr4bMcS1zwsS3wH0Lmt72lFA+kbUf0Al+PudKnYR6un2M2QsSKXWn/gpTJhBiE0nLaA886+AA+nbcCDzbGn7X8Rwaq9), [SWC](https://play.swc.rs/?version=1.13.3&code=H4sIAAAAAAAAAzWNPQuDMBRF9%2FyKh1PioHuD%2FaCNtEsddM9gEwmEPGgjDpL%2F3pdqhwePezj3rgzAmwiXboAG4ns2kpIRwyfCZOIW1yWctX52un%2FclFZtq65DrzWUNXABzTHbWbNzGKPDABaRwErRv8sG6tkKiZx2LzP0pvI48eLuCgGHnaxJ%2FmwbuMhfoltceOFSUXneyhOSJfYFrxIZZ8IAAAA%3D&config=H4sIAAAAAAAAA32VyW4bMQxA7%2F4KQ%2BceiqDooR8Q5NJvEGSJGsvVMhApx4PA%2F17O5jgxx7cRHxeJ23zs9nt1Qqv%2B7D%2F4kw%2B9qQj1dmYJDpnMhSUKbDJoa%2BhJ%2FVjpCUfkTUSYRNeZqFgKwkoWWQo5%2BOHety2pr4B4J2OpqV1LkAm%2F2i%2BslvcRUG338kMpEUx%2BQrRBHTJBB1VybEuMpkfQZ1MFL%2BNNTQ1YpBAjbARO97X0Is8uUCiZYz5SB8ZpWxwIKFSwFM4gmXEsNsvIzxPeM2EHh9Z1Uz2%2FWcPZxGZIiAmXqSR8W8HrsQQk7VuWUjjDjRzMcEnud8vgdQVqNT%2FanUrIGzX5B8AZiAYxmwSS30nDcz9tWfunliF7blkaBM79Lb0yQ8dJ1SF4IbNjZqBSkKpZwTULY2atdJ0Fb6QPgwMN3nOvCK7xPZA9SkFp6KF4AXB9jZe6agb6NoUbfByIJ%2FiVX0lygy0aydBxm%2BKQDiU%2BCZCAjsU9UeBSUNnGlbfEpd%2FmLTvg1gAnqjScwOMS4AGgoiOQ1Bs8HuxRd7EcPtfEonC97dtkcvc5719WLlwIKq%2BYN4j9w45TxFMwBVaALz9ffqndYqlScW1yufwCxq6Y9X7fKa27e723Cvh3NVxku%2Bt%2FCxPAVFAGAAA%3D)

armano2:
> We cannot close #13651 as it requires more optimizations

this is related to what i discovered in #16964, cross scope variables that are not static are never in-lined

PR #16479, opened 2025-12-04
feat(ast, parser, codegen): add `CommentKind::MultilineBlock`

Dunqing:
Add `CommentKind::MultilineBlock` to indicate a block comment that has at least one linebreak, which would be useful in the Codegen and Formatter

Examples:

CommentKind::Block:
```js
/* single line block comment */
```

CommentKind::MultilineBlock

```js
/* multiline 
block
 comment */
```

Maybe it would be better to rename `CommentKind::Block` to `CommentKind::SinglelineBlock.` as well

### DIFF
--- a/crates/oxc_ast/src/ast/comment.rs
+++ b/crates/oxc_ast/src/ast/comment.rs
@@ -17,6 +17,9 @@ pub enum CommentKind {
     Line = 0,
     /// Block comment
     Block = 1,
+    /// Multiline block comment (contains line breaks)
+    #[estree(rename = "Block")]
+    MultilineBlock = 2,
 }
 
 /// Information about a comment's position relative to a token.
@@ -172,7 +175,9 @@ impl Comment {
     pub fn content_span(&self) -> Span {
         match self.kind {
             CommentKind::Line => Span::new(self.span.start + 2, self.span.end),
-            CommentKind::Block => Span::new(self.span.start + 2, self.span.end - 2),
+            CommentKind::Block | CommentKind::MultilineBlock => {
+                Span::new(self.span.start + 2, self.span.end - 2)
+            }
         }
     }
 
@@ -185,7 +190,13 @@ impl Comment {
     /// Returns `true` if this is a block comment.
     #[inline]
     pub fn is_block(self) -> bool {
-        self.kind == CommentKind::Block
+        matches!(self.kind, CommentKind::Block | CommentKind::MultilineBlock)
+    }
+
+    /// Returns `true` if this is a multi-line block comment.
+    #[inline]
+    pub fn is_multiline_block(self) -> bool {
+        self.kind == CommentKind::MultilineBlock
     }
 
     /// Returns `true` if this comment is before a token.
--- a/crates/oxc_parser/src/lexer/comment.rs
+++ b/crates/oxc_parser/src/lexer/comment.rs
@@ -1,5 +1,6 @@
 use memchr::memmem::Finder;
 
+use oxc_ast::CommentKind;
 use oxc_syntax::line_terminator::is_line_terminator;
 
 use crate::diagnostics;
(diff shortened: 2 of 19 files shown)


Dunqing:
* **#16640** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16640?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16501** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16481** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16479** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16479?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16479?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=16479).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

crates/oxc_parser/src/lexer/comment.rs
```diff
@@ -82,9 +83,10 @@ impl<'a> Lexer<'a> {
     /// Section 12.4 Multi Line Comment
     pub(super) fn skip_multi_line_comment(&mut self) -> Kind {
         // If `is_on_new_line` is already set, go directly to faster search which only looks for `*/`
-        if self.token.is_on_new_line() {
-            return self.skip_multi_line_comment_after_line_break(self.source.position());
-        }
+        // TODO(FIXME): commented out since it doesn't means there is a line break in the comment, while it means the line break is before the comment
+        // if self.token.is_on_new_line() {
+        //     return self.skip_multi_line_comment_after_line_break(self.source.position());
+        // }
```
Dunqing: We need to remove this fast branch, as now `skip_multi_line_comment_after_line_break` only handles multiline line block comments. But I am not sure if it is correct.

 @overlookmotel cc

crates/oxc_ast/src/ast/comment.rs
```diff
@@ -17,6 +17,8 @@ pub enum CommentKind {
     Line = 0,
     /// Block comment
     Block = 1,
+    /// Multiline block comment (contains line breaks)
+    MultilineBlock = 2,
```
overlookmotel, line 22: To fix ESTree:

```suggestion
    /// Multiline block comment (contains line breaks)
    #[estree(rename = "Block")]
    MultilineBlock = 2,
```

crates/oxc_parser/src/lexer/comment.rs
overlookmotel: I think this change is fine.

Only difference is that when there was already a line break seen before the comment, it doesn't go directly to the fast path which uses `memchr` to find end of the comment. It uses the slower loop instead. But still, once the loop finds a line break, it goes to the `memchr` fast path for the rest of the comment.

It should not affect perf much. Especially for the common pattern where block comment starts with `/*\n` or `/**\n` - so it only searches 3 or 4 bytes with the slow loop in common case.

I suggest just changing the comment to explain:

```suggestion
        // We need to identify if comment contains line breaks or not
        // (`CommentKind::Block` or `CommentKind::MultilineBlock`).
        // So we have to use the loop below for the first line of the comment even if
        // `Token`'s `is_on_new_line` flag is already set.
        // If the loop finds a line break before end of the comment, we then switch to
        // the faster `skip_multi_line_comment_after_line_break` which searches 
        // for the end of the comment using `memchr`.
```

overlookmotel: See my comment below.

Dunqing: Thank you for the detailed explanation! I've applied your two suggestions.

overlookmotel (approved):

overlookmotel:
I pushed 1 commit to just maintain "correct" import order.

2nd PR in this stack is awaiting review by @leaysgur, so I've not merged this PR - I assume better we merge the whole stack in one go, rather than just this commit getting included in today's release.

PR #16495, opened 2025-12-04
chore(linter): replaces ✨ with ✅

r2abreu:
> Replace ✨ with ✅
> A check mark is more readily associated with something being finished or done
> The yellow is harder to see against a bright background

See: https://github.com/oxc-project/oxc/issues/15927


camc314 (approved):
Thank you!

PR #16496, opened 2025-12-04
chore(linter): replaces docs column with hyperlinked name

r2abreu:
> Make the name of each rule a hyperlink to that rule's docs
> The full links in the "Docs" column take up a lot of space

See: https://github.com/oxc-project/oxc/issues/15927


camc314 (approved):
Thanks!

connorshea:
This has a minor problem, in that some rules have no docs (like a lot of the rules near the bottom of the list on [the react plugin](https://github.com/oxc-project/oxc/issues/1022)). If anyone wants to open a PR to only apply the hyperlink when there's a docs link for the rule, that'd be good :) But not a big problem, so no worries.

Issue #16359, opened 2025-12-01
Linter plugins: Pass globals to JS plugins


overlookmotel:
Globals can be specified in Oxlint config files, and are already parsed on Rust side.

We need to send them over to JS side, and make them available to JS plugins as `context.languageOptions.globals`.

I think we can use the same technique as #16215 used for passing options from Rust to JS - send them over once after parsing all configs, and reference them with IDs thereafter.

lilnasy:
For background, #16215 introduced `setup_configs()`, which runs once per process and gives oxlint an opportunity to serialize data that does **not** change between files to lint. Rule options worked easily for this approach because they are not subject to merging and overriding so they can be marshaled once then used as-is.

I think globals, and settings for that matter, will not be as simple -- at least, it will not be simple if we want to support merging multiple configurations (actually, I am not sure if merging is supported even for internal plugins yet, but it's a consideration regardless) and overrides, supported since #8915.
https://github.com/oxc-project/oxc/blob/8c85e087f0e9f087988e145ad1e9bd4afbdbf066/crates/oxc_linter/src/config/config_store.rs#L235-L237

E2E test for this current behavior: [config](https://github.com/oxc-project/oxc/pull/16512/files#diff-a56a021c38ec859aa879b884a91e60d68386dc2383139ac6881d2fc2aa42cb2a), [snapshot](https://github.com/oxc-project/oxc/pull/16512/files#diff-041648a519d0a66cdb7eeff68ff64f9c5135c11a8ef89e2e9557e93ba24fe1a6). The test is passing with straightforward (but inefficient) serialization of globals for every file: [oxc_linter/src/lib.rs](https://github.com/oxc-project/oxc/pull/16512/files#diff-207225884c5e031ffd802bb99e4fbacbd8364b1343a1cec5485bf50f29186300).

## Heavyweight approach
My first idea was to serialize the base globals and override globals as-is, and reimplement some of the resolve, glob-match, override logic on the JS side. I have gotten it working with the tests passing, but I am not happy with it. I don't like the fact that it required an additional dependency (analog to fast-glob on the rust side), and that it would likely snowball into more rust logic being rewritten in JS, which is is the opposite of what we generally want. We need bit of a smarter solution.

## Smart approach
[...]

PR #16512, opened 2025-12-04
feat(linter/plugins): allow JS plugins to access `globals`

lilnasy:
- This PR implements a naive, inefficient solution to #16359.
- To establish current and expected behavior, this PR serializes rust-resolved and -overridden globals once for every file.

## Future work
- [ ] Add test for globals from the `env` field. Implement support if not currently supported.
- [ ] Optimization: we want to serialize `globals` for all files only once at the start of the lint run.

## AI Usage
- This PR was drafted by Claude Sonnet 4.5 in Plan + Agent mode, then manually reviewed and iterated on.


crates/oxc_linter/src/lib.rs
```diff
@@ -463,12 +463,21 @@ impl Linter {
             None => "{}".to_string(),
         };
 
+        let globals_json = serde_json::to_string(ctx_host.globals()).unwrap_or_else(|e| {
+            let path = path.to_string();
```
overlookmotel: You can remove this `let path = path.to_string();` line. `format!` only requires a `&str` which `path` already is. No need to convert it to a `String`.

apps/oxlint/src-js/plugins/context.ts
```diff
@@ -185,10 +186,24 @@ const LANGUAGE_OPTIONS = freeze({
...
+    // We support both on Rust side, but convert to ESLint's spelling for JS side.
+    const result: Record<string, "readonly" | "writable" | "off"> = {};
+    for (const [key, value] of Object.entries(globals)) {
+      if (value === "writeable") {
+        result[key] = "writable";
+      } else {
+        result[key] = value;
+      }
+    }
+    return Object.keys(result).length === 0 ? null : result;
```
overlookmotel: This logic I think should to move into `initGlobals`. Then it only happens once per file, instead of each time you access `languageOptions.globals`.

Ideally, we'd just rename `GlobalValue::Writeable` enum member to `Writable` in Rust code. Then it'd be serialized as `"writable"` in the JSON, and we could remove this code from JS side. But we can do that in a later PR. Please just add a TODO comment for now.

Also, the `Object.keys(result).length === 0` check is more expensive than we'd like. The `null` value is already taken (meaning "not deserialized yet"). But what we could do is add an empty object as a top-level const, and check for equality.

```ts
// Empty globals object.
// No need to freeze this object, as it's never passed to user.
export const EMPTY_GLOBALS: Globals = {};

export function initGlobals(): undefined {
  debugAssertIsNonNull(globalsJSON);

  if (globalsJSON === "{}") {
    globals = EMPTY_GLOBALS;
    return;
  }

  globals = JSON.parse(globalsJSON);

  // `globals` was deserialized from JSON, so we can use a simple `for..in` loop here
  for (const key in globals) {
    if (globals[key] === "writeable") globals[key] === "writable";
  }

  // Freeze the globals object, to prevent any mutation of `globals` by plugins.
  // No need to deep freeze since all keys are just strings.
  return Object.freeze(globals);
}
```

Then in `get globals()` can do `return globals === EMPTY_GLOBALS ? null : globals;

apps/oxlint/src-js/plugins/lint.ts
```diff
@@ -142,6 +148,9 @@ export function lintFileImpl(
   // Pass settings JSON to context module
   setSettingsForFile(settingsJSON);
 
+  // Pass globals JSON to context module
```
overlookmotel:
```suggestion
  // Pass globals JSON to globals module
```

The comment above for settings was also inaccurate. Fixed that in #16559.

crates/oxc_linter/src/context/host.rs
```diff
@@ -259,6 +259,11 @@ impl<'a> ContextHost<'a> {
         &self.config.settings
     }
 
+    #[inline]
+    pub fn globals(&self) -> &crate::OxlintGlobals {
```
overlookmotel: Nit: Could you please import this at top-level, and then:


```suggestion
    pub fn globals(&self) -> &OxlintGlobals {
```

overlookmotel (commented):
I know you've said this is just an experiment, but I think we should merge it!

Yes, it's inefficient to serialize globals for every file (ditto settings). But at least it works! Much better than just not supporting it at all, which is what we have now.

A few comments below - none major.

---

We also support an `env` option in config: https://oxc.rs/docs/guide/usage/linter/config.html#configuration-file-format

I'm not sure globals from `env` already get merged into `OxlintGlobals` on Rust side, in which case it'll work already after this PR.

If so, we should add a test for that. If not, we should add support for `env` (simplest way would be to just apply `env` to `OxlintGlobals` on Rust side, before serializing to JSON). Either way, I suggest leaving that to a follow-up PR.

lilnasy:
@overlookmotel Makes sense. I'll address the review comments tonight and add todo comments for performance.

apps/oxlint/test/fixtures/globals/files/nested/index.js
```diff
@@ -0,0 +1 @@
+let x;
```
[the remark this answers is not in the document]
lilnasy: It's a test. Leave it alone.

lilnasy:
Addressed all review comments. CI is failing because spell check flaked.

crates/oxc_linter/src/config/globals.rs
```diff
@@ -61,6 +61,7 @@ impl OxlintGlobals {
 #[serde(rename_all = "lowercase")]
 pub enum GlobalValue {
     Readonly,
+    // TODO: #[serde(rename = "writable")] for ESLint compatibility
```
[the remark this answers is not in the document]
lilnasy: This can be done in a separate, smaller PR focused on making sure there are no breaking changes when we add the alias to the deserializer.

overlookmotel (approved):
Excellent!

Pushed a few commits with my usual style nits, but made no substantive changes.

By the way, the `as` casts I've added are ugly and probably make it less readable, but I think on balance preferable to avoid `// @ts-expect-error` comments where possible. So annoying that TS doesn't allow type-asserting side effects of functions. It seems designed to guide you *away* from performant patterns, and make you introduce extra temp variables everywhere if you want to type cast.

PR #16553, opened 2025-12-06
chore(linter): Update whitespace trim in a few rule files

connorshea:
Otherwise the next person that edits these files will end up with goofy problems if they use an editor that actually follows the editorconfig (like mine does).

I found the problematic files by just doing a find+replace on "What it does", since that should be in every rule file, which caused my editor to save and applied to the editorconfig settings accordingly. The only one I left with this problem was `curly`, as it seemed to be reliant on the weird whitespace.

And then also replaces some usage of tabs in some rule tests with equivalent spaces. None of these rules are stylistic, so the difference is not relevant to the behavior of the rule.

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/arrow_body_style.rs
+++ b/crates/oxc_linter/src/rules/eslint/arrow_body_style.rs
@@ -515,15 +515,15 @@ fn test() {
         ("var foo = () => { /* do nothing */ };", None),
         (
             "var foo = () => {
-			 /* do nothing */
-			};",
+              /* do nothing */
+            };",
             None,
         ),
         (
             "var foo = (retv, name) => {
-			retv[name] = true;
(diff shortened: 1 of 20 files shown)


connorshea:
### Merge activity

* **Dec 6, 2:33 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 6, 2:33 AM UTC**: `connorshea` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Dec 6, 2:38 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

PR #16556, opened 2025-12-06
docs(linter): Fix the rule config docs for `eslint/no-inner-declarations` rule.

connorshea:
This fixes the `no-inner-declarations` docs by treating them as a tuple. Although, it also breaks the tests for no-inner-declarations. 🙃 

I _think_ there may be a bug in the tests/logic for our `no-inner-declarations` rule (I think oxc assumes strict mode by default maybe?). We previously returned `None` for `block_scoped_expressions` instead of `allow`, which [the original rule](https://github.com/eslint/eslint/blob/main/lib/rules/no-inner-declarations.js) does not do. It always just sets `allow` as the default value as far as I can tell.

I'm unsure if I should add an `impl Default for NoInnerDeclarationsConfigObject` and set it to return None, and retain that behavior, or if we should fix this behavior to match the original ESLint rule better? Up to cam, probably.

Generated docs:

```md
## Configuration

### The 1st option

type: `"functions" | "both"`

#### `"functions"`

Disallows function declarations in nested blocks.

#### `"both"`

Disallows function and var declarations in nested blocks.

### The 2nd option

#### blockScopedFunctions

type: `"allow" | "disallow"`

##### `"allow"`

Allow function declarations in nested blocks in strict mode (ES6+ behavior).

##### `"disallow"`

Disallow function declarations in nested blocks regardless of strict mode.
```

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/sort_keys.rs
+++ b/crates/oxc_linter/src/rules/eslint/sort_keys.rs
@@ -8,14 +8,18 @@ use oxc_diagnostics::OxcDiagnostic;
 use oxc_macros::declare_oxc_lint;
 use oxc_span::{GetSpan, Span};
 use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize};
 
-use crate::{AstNode, context::LintContext, rule::Rule};
+use crate::{
+    AstNode,
+    context::LintContext,
+    rule::{DefaultRuleConfig, Rule},
+};
 
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, Deserialize)]
 pub struct SortKeys(Box<SortKeysConfig>);
 
-#[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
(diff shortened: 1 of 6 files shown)


connorshea:
Marking as a draft until we have a discussion on whether this current behavior is the desired one.

crates/oxc_linter/src/rules/eslint/sort_keys.rs
```diff
@@ -102,50 +97,18 @@ declare_oxc_lint!(
...
-        let allow_line_separated_groups = config
-            .get("allowLineSeparatedGroups")
-            .and_then(serde_json::Value::as_bool)
-            .unwrap_or(false);
-
-        Self(Box::new(SortKeysConfig(
-            sort_order,
-            SortKeysOptions { case_sensitive, natural, min_keys, allow_line_separated_groups },
-        )))
+        serde_json::from_value::<DefaultRuleConfig<SortKeys>>(value)
```
xuanduc987, line 100: I believe we can use `serde_json::from_value::<SortKeys>(value)` here, so the change to `DefaultRuleConfig` is unnecessary?

connorshea:
I am going to close this for now until we get to a solution/decision on the tuple configs, so we don't end up with a bunch of PRs open that depend on each other. I'll reuse this PR's code changes in the future.

PR #689, opened 2023-08-04
feat(semantic): add `node_id` to `Reference`

makotot:
Closes #685 

Intend to use this in the following ways in #672.
```rs
let node = ctx.nodes().get_node(reference.node_id());
if !self.type_of && has_typeof_operator(node, ctx) {
    return;
}
```

PR #16580, opened 2025-12-08
feat(minifier): simplify destructuring array assigment

armano2:
simplify destructuring array assignment pattern

```js
var [a] = [1] -> var a = 1
var [, [c]] = [2, [3]] -> var [] = 2, c = 3
```

use cases in a wild:
- https://github.com/oracle/bots-node-sdk/blob/ce1aa5ffff53ae9e988a799bd90a3e6ddbbf1611/bin/lib/spawn.js#L20
- https://github.com/Leask/Starlink-Signal-Status/blob/59c01b306ed028da0d8986e737e84146202159fc/lib/slss.mjs#L29
- https://github.com/search?q=%22%2C+err%5D+%3D+%5B%22+language%3AJavaScript+&type=code


crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +447,169 @@ impl<'a> PeepholeOptimizations {
...
+            decl.id.kind.is_array_pattern()
+                && matches!(decl.init, Some(Expression::ArrayExpression(_)))
+        }) {
+            return false;
+        }
+
+        let mut changed = false;
+        let mut new_var_decl: Vec<'a, VariableDeclarator<'a>> = ctx.ast.vec();
+
+        for mut decl in declarations.drain(..) {
```
armano2: TODO: we do not want to drain/rebuild all declarations

armano2:
i have been playing around and trying to figure out some of cases for object assignments

initially i tough that i can just compress this

```js
var { a, a: c } = { a: { f: 1 } }
```
to
```js
var a = { f: 1}, c = a
```

this kinda gets annoying as syntax is really fluid eg.

```js
var { a: { f }, a } = { a: { f: 1 } }
```
i think this should be collapsed to 
```js
var { f } = a = { f: 1 }
```

note:
we have to keep references correct,

eg.
```js
var { a: { c }, a } = { a: { c: { d: {} } } }
console.log(c === a.c)

/// 

var { c } = a = { c: { d: {} } }
console.log(c === a.c)
```

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -1869,4 +2069,52 @@ mod test {
...
+        test("var [a, b, c = 2] = [1, 2, 3, 4]", "var a = 1, b = 2, c = 3, [] = [4]");
+        test("var [a, b, c = 3] = [1, 2]", "var a = 1, b = 2, c = 3");
+        test("var [a, b] = [1, 2, 3]", "var a = 1, b = 2, [] = [3]");
+        test("var [a] = [123, 2222, 2222]", "var a = 123, [] = [2222, 2222]");
+        test("const [a] = []", "const a = void 0");
+        // spread
+        test("var [...a] = [...b]", "var [...a] = [...b]");
+        test("var [a, a, ...d] = []", "var a, a, [...d] = []");
+        test("var [a, ...d] = []", "var a, [...d] = []");
+        test("var [a, ...d] = [1, ...f]", "var a = 1, [...d] = [...f]");
```
armano2: this could be simplified even more `[...d] = [...f] -> d = [...f]`, double clone is not needed and generaly spreads could be always removed from initializer if there is no elements in init

```suggestion
        test("var [a, ...d] = [1, ...f]", "var a = 1, d = [...f]");
```

we could also implement this for strings

```js
[a, b, c] = "123"
```

crates/oxc_ecmascript/src/constant_evaluation/value_type.rs
```diff
@@ -111,6 +111,16 @@ impl<'a> DetermineValueType<'a> for Expression<'a> {
...
 
+impl<'a> DetermineValueType<'a> for ArrayExpressionElement<'a> {
+    fn value_type(&self, ctx: &impl GlobalContext<'a>) -> ValueType {
+        match self {
+            ArrayExpressionElement::SpreadElement(_) => ValueType::Undetermined,
+            ArrayExpressionElement::Elision(_) => ValueType::Undefined,
+            _ => self.as_expression().map_or(ValueType::Undetermined, |expr| expr.value_type(ctx)),
+        }
+    }
+}
```
armano2: is it ok, if we add impl for `ArrayExpressionElement` here, or should i do this inline?

armano2:
i think this is not fully correct, why var's have to be so annoying
```js
var [a, b] = [1, 2]
console.log(a, b) // 1, 2
var [b, a] = [a, b]
console.log(a, b) // 2, 1
```

```js
var [a] = [a]; // should not change
var [a, b] = [b, a]; // should not change
var [a, ...b] = [b, a]; // should not change
var [a, ...b] = [...b, a]; // should not change
var [a, b] = [b, a], [a, b] = [b, a]; // should not change
var [a, b] = [1, 2], [a, b] = [b, a] // var a = 2, b = 1"
```

armano2:
scoping and variables seem to have an issue that is going to block this, outside of checking right side of assigment for identifier references,

i would either expect this to be present in `symbol_redeclarations`, in root_unresolved_references, or somwhere.

i did small attempt on getting all identifiers_reference's from expression, tho this seem so hacky, 

@sapphi-red do you have any suggestions? maybe I'm missing something

bdw, also looks that we are not modifying scope/ references correctly when we remove/inline stuff 
[playground](https://playground.oxc.rs/#eNptU8Fu2zAM/ZVAZw9oAwwoMuwwtAswoFu7dVgPXQ+0zbhKZcmj6CRFkH/fk2M5OfRk6dEkHx+f9qYyC7MhmdHs8+yymJX4zIvZE+H4jPNTWczo+dNfbwoTzGJvpPfp46xXs1DpuTCrIC2pspjFilwEokI+JnhCbAyOlOsbrhwJqQ0+TsHti1WOHVU8QS35xp2uVWg74YiUY8tYhQ7R8fLWlsHlW7VqxrRDYTqSmHjtDe+UfURbzKtxh3HIubD9xdqLv+s12pqXva8Ss1wqtWTZ8D0JcjM65P25+uZVLCpWJ1IM1mqrryJBRhQcklQDB5zPpNqbPvJvKk8yKJWPttYXs5gXhn19t7q1HlMatwLdDt10jF9doJtNCv3sg55kWsfdwzvwv3S7l9Chl6H4wTPXXKMm9mRBr7kObUsp5hxQzGGnYTHLNs+P+JbeIn4phapX1gfsDOn554xSy0fmY/9QrrnSR6EONbKoqdHA9YtCyLIHQ5azNKg12eiol5I0DNcZjvOLy48oAAVveIWcZZBrRzEuLbt6WgjvOhbbsldyMF6A76bFINpandDvrFSTQoRxa+8bdm8iyHbYPYsneC5TrULNDQ9Pw6cdT3ZcxzrgjY1qeh90qJURx03+d6iSbZ7Mkt8AJg/dLW94KvrK3P2Aytk7Qyr8GNwS3kwZG5YyxDMtrU87SKHpdABcD/IleDodEp5kHt/V4T/t5XcS)

-----------------------

if we could get `node_id` for symbol initialization, that could also work #689
or have a `init` (var decl) create its reference

sapphi-red:
Maybe we should think about it by introducing a temporary variable instead?
```js
// input
var [a, b] = [1, 2]
console.log(a, b) // 1, 2
var [b, a] = [a, b]
console.log(a, b) // 2, 1

// output
var [a, b] = [1, 2]
console.log(a, b) // 1, 2
var _b = a, _a = b
b = _b, a = _a
console.log(a, b) // 2, 1
```
Because we have to handle cases like
```js
var [a, b] = [1, 2]
console.log(a, b) // 1, 2
var [b, a] = [f_a(), f_b()]
console.log(a, b) // 2, 1

function f_a() { return a }
function f_b() { return b }
```
which cannot be detected by traversing the right side of the declaration.

Although, it might be difficult to figure out the condition that is safe to optimize the output:
```js
// output
var _b = a, _a = b
b = _b, a = _a
console.log(a, b)

// output (optimized): what is the condition we can safely do this?
console.log(b, a)
```

armano2:
that's kinda work, but for
```js
var [a, b] = [b, a] // in - 14
var _ = b, _2 = a, a = _, b = _2 // out - 19
var _2 = a, a = b, b = _2 // ideal - 11
```
```js
var [a, b] = [b, a], [a, b] = [b, a] // in - 26
var _3 = b, _4 = a, a = _3, b = _4, _ = b, _2 = a, a = _, b = _2 // out - 35
```
this is unexpected, tho its not related to this code
```js
var [a, b] = [1, 2], [a, b] = [b, a] // in - 26 
var a = 1, _ = 2, _2 = a, a = 2, b = _2 // out - 19
var a = 2, b = 1 // ideal - 11
```
----------------

```js
const [[x, y, z] = [4, 5, 6]] = [] // in - 25
const x = 4, y = 5, z = 6 // out - 17 [ideal]
```


if we can improve in-lining of `var` this could be significantly optimized

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,254 @@ impl<'a> PeepholeOptimizations {
...
+            init_expr.elements.iter().position(ArrayExpressionElement::is_spread)
+        {
+            min(id_pattern.elements.len(), spread_index)
+        } else {
+            id_pattern.elements.len()
+        };
+
+        if decl.kind.is_var() && index > 1 {
+            for init_val in &mut init_expr.elements {
+                if init_val.value_type(ctx).is_undetermined() {
```
sapphi-red: What is the purpose of this `if`? Is it to avoid unnecessary temporary vars? In that case, I think we should use `.is_literal_value(false, ctx)` instead. For example, `.value_type(ctx)` returns `ValueType::boolean` for `!a`.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,254 @@ impl<'a> PeepholeOptimizations {
...
+        }
+
+        let mut init_iter = init_expr.elements.drain(..);
+
+        for id_item in id_pattern.elements.drain(0..index) {
+            let init_val = match init_iter.next() {
+                None | Some(ArrayExpressionElement::Elision(_)) => {
+                    ArrayExpressionElement::from(ctx.ast.void_0(SPAN))
+                }
+                Some(init_val) => init_val,
```
sapphi-red:
```suggestion
                None | Some(ArrayExpressionElement::Elision(_)) => {
                    ctx.ast.void_0(SPAN)
                }
                Some(ArrayExpressionElement::SpreadElement(_)) => {
                    unreachable!("spread element does not exist until `index`")
                }
                Some(init_val) => init_val.to_expression(),
```
I think it'd be nice to explicitly handle SpreadElement here. I guess this would remove the need of `impl<'a> DetermineValueType<'a> for ArrayExpressionElement<'a>`.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,254 @@ impl<'a> PeepholeOptimizations {
...
+            };
+
+            match id_item {
+                // `[a = b] = [??]`
+                Some(BindingPattern::AssignmentPattern(mut pattern)) => {
+                    let init_value_type = init_val.value_type(ctx);
+                    if init_value_type.is_undefined() {
```
sapphi-red: We need to keep the RHS when `init_val` has sideeffects. (e.g. `let [a = 1] = [void foo()]`)

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,254 @@ impl<'a> PeepholeOptimizations {
...
+                        id,
+                        NONE,
+                        Some(init_val.into_expression()),
+                        decl.definite,
+                    ));
+                }
+                // `[,] = [??]` => `void [] = [??]`
+                None => {
+                    // `[,] = [,]` => `void [] = [,]`
+                    if !init_val.value_type(ctx).is_undefined() {
```
sapphi-red: We need to keep the RHS when `init_val` has sideeffects here as well.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,254 @@ impl<'a> PeepholeOptimizations {
...
+            true
+        } else {
+            init_expr.elements = ctx.ast.vec_from_iter(init_iter);
+            false
+        }
+    }
+
+    /// Simplifies destructuring assignments by transforming array patterns into a sequence of
+    /// variable declarations, whenever possible. This function modifies the input declarations
+    /// and returns whether any changes were made.
```
sapphi-red:
```suggestion
    /// Simplifies destructuring assignments by transforming array patterns into a sequence of
    /// variable declarations, whenever possible. This function modifies the input declarations
    /// and returns whether any changes were made.
    ///
    /// For some inputs, this transformation may increase the code size as this transformation
    /// injects additional temporary variables. But we assume those cases are rare.
```

crates/oxc_minifier/src/peephole/minimize_statements.rs
armano2: correct, there is no need to create variables for literals

thank you

crates/oxc_minifier/src/peephole/minimize_statements.rs
armano2: like always thx, i didn't think about that, i'm going to update it

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +446,262 @@ impl<'a> PeepholeOptimizations {
...
+        if decl.kind.is_var() {
+            for init_expr in &mut init_expr.elements {
+                if !init_expr.is_literal_value(false, ctx) {
+                    // create intermediate declarations to avoid issues with hoisting
+                    // `var [a] = [b]` => `var _a = b, a = _a`
+                    let ident =
+                        ctx.generate_uid_in_current_scope("_", SymbolFlags::FunctionScopedVariable);
+                    result.push(ctx.ast.variable_declarator(
+                        init_expr.span(),
+                        VariableDeclarationKind::Let,
```
armano2: i'm not sure about this, we are in scope of `Var`, but we are creating `Let` here

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -2161,13 +2161,13 @@ mod test {
...
-        test("let [a, , c, d] = [, 3, , 4]", "let a, [] = [3], c, d = 4");
+        test("let [, , ] = [1, 2]", "");
+        test("let [a, , c, d] = [, 3, , 4]", "let a, c, d = 4");
         test("let [a, , c, d] = [void 0, e, null, f]", "let a, [] = [e], c = null, d = f");
-        test("let [a, , c, d] = [1, 2, 3, 4]", "let a = 1, [] = [2], c = 3, d = 4");
-        test("let [ , , a] = [1, 2, 3, 4]", "let [] = [1], [] = [2], a = 3, [] = [4]");
-        test("let [ , , ...t] = [1, 2, 3, 4]", "let [] = [1], [] = [2], t = [3, 4]");
-        test("let [ , , ...t] = [1, ...a, 2, , 4]", "let [] = [1], [, ...t] = [...a, 2, , 4]");
+        test("let [a, , c, d] = [1, 2, 3, 4]", "let a = 1, c = 3, d = 4");
+        test("let [ , , a] = [1, 2, 3, 4]", "let a = 3, [] = [4]");
```
armano2: remainder of inlining eg `[] = [4]` will be removed by #16855 #17074

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -2161,13 +2161,13 @@ mod test {
         test("let [a = foo] = [,]", "let a = foo");
         // holes
         test("let [, , , ] = [, , , ]", "");
-        test("let [, , ] = [1, 2]", "let [] = [1], [] = [2]");
-        test("let [a, , c, d] = [, 3, , 4]", "let a, [] = [3], c, d = 4");
+        test("let [, , ] = [1, 2]", "");
+        test("let [a, , c, d] = [, 3, , 4]", "let a, c, d = 4");
         test("let [a, , c, d] = [void 0, e, null, f]", "let a, [] = [e], c = null, d = f");
```
armano2: with full optimization this is going to be reduced to `e,f;`

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -1869,4 +2136,83 @@ mod test {
...
+        test("let [a, b] = [1, 2], [a, b] = [b, a]", "let a = 1, b = 2, a = 2, b = 2");
+        test("let [a, b] = [b, a], [a, b] = [b, a]", "let a = b, b = a, a = b, b = a");
+        // const
+        test("const [[x, y, z] = [4, 5, 6]] = []", "const x = 4, y = 5, z = 6;");
+        test("const [a, ...d] = []", "const a = void 0, d = [];");
+        test("const [a] = []", "const a = void 0");
+        // vars
+        test("var [a] = [a]", "var a = a");
+        test("let [...a] = [b, c]", "let a = [b, c]");
+        test("var [a, b] = [1, ...[2, 3]]", "var _ = [2, 3], a = 1, [b] = [..._]");
```
armano2: surprising that we are already not in-lining spreads if possible #17358

armano2:
ok, all changes applied, and it seem to be working fine

armano2:
q: what do you think about not processing vars if there is more than one non literal value?
```rs
// in this case only one variable will be created and we could create it only if its not first
decl.kind.is_var() &&
init_expr.elements.iter().filter(|elem| !elem.is_literal_value(false, ctx)).count() > 1
```

or even do not do that if there is any non literal

```rs
// in this case we never create variables
decl.kind.is_var() &&
init_expr.elements.iter().any(|elem| !elem.is_literal_value(false, ctx))
```


that would let us handle all cases where we extensively create additional variables for vars

--------------

this solves issues like:
```js
var [a, b] = [1, 2], [a, b] = [b, a] // in
var a = 2, b = 1 // new
var a = 1, _ = 2, _2 = a, a = _, b = _2 // old
```
```js
var [a, b] = [!d, !a] // in - out
var _ = !d, _2 = !a, a = _, b = _2 // old
```
```js
var [a, b] = [b, a], [a, b] = [b, a] // in - out
var _3 = b, _4 = a, a = _3, b = _4, _ = b, _2 = a, a = _, b = _2 // old
```

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -443,6 +445,266 @@ impl<'a> PeepholeOptimizations {
...
+                            };
+                            // check whathever id is present in init [a] = [b]
+                            binding_identifiers.iter().any(|id| id.symbol_id().eq(ref_symbol))
+                        }
+                        None => e.is_spread(), // do not allow spread
+                        _ => !e.is_literal_value(false, ctx),
+                    }
+                });
+            }
+        }
```
armano2: maybe we should extract this to something more generic

ideally we would want to resolve it based on nodeId by comparing sub-tree, but that's currently not supported

this produces better output than creating intermediate variables that would otherwise could never be in-lined

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -1869,4 +2127,82 @@ mod test {
...
+        test("let [a, b] = [b, a], [a, b] = [b, a]", "let a = b, b = a, a = b, b = a");
+        // const
+        test("const [[x, y, z] = [4, 5, 6]] = []", "const x = 4, y = 5, z = 6;");
+        test("const [a, ...d] = []", "const a = void 0, d = [];");
+        test("const [a] = []", "const a = void 0");
+        // vars
+        test("var [a] = [a]", "var a = a");
+        test("var [...a] = [b, c]", "var a = [b, c]");
+        test_same("var [a, b] = [1, ...[2, 3]]");
+        test_same("var [a, b] = [c, ...[d, e]]");
```
armano2: spread is currently not supported, see #17358

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -455,6 +459,262 @@ impl<'a> PeepholeOptimizations {
...
+            false
+        }
+    }
+
+    /// Simplifies destructuring assignments by transforming array patterns into a sequence of
+    /// variable declarations, whenever possible. This function modifies the input declarations
+    /// and returns whether any changes were made.
+    ///
+    /// For some inputs, this transformation may increase the code size as this transformation
+    /// injects additional temporary variables. But we assume those cases are rare.
```
sapphi-red:
```suggestion
```
I guess this is no longer relevant.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -481,6 +485,263 @@ impl<'a> PeepholeOptimizations {
...
+                .first()
+                .is_some_and(|e| e.as_ref().is_none_or(BindingPattern::is_assignment_pattern))
+        {
+            return false;
+        }
+
+        if decl.kind.is_var() && init_len > 1 {
+            let binding_identifiers = decl.id.get_binding_identifiers();
+            if !binding_identifiers.is_empty() {
+                let binding_identifier_symbol_ids = HashSet::from_iter_in(binding_identifiers.iter().map(|id| id.symbol_id()), ctx.ast.allocator);
```
sapphi-red: I think it'd be nice to have a comment why we only allow these values.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -481,6 +485,266 @@ impl<'a> PeepholeOptimizations {
...
+                Some(ArrayExpressionElement::SpreadElement(_)) => {
+                    unreachable!("spread element does not exist until `index`")
+                }
+                Some(init_item) => init_item.into_expression(),
+            };
+
+            match id_item {
+                // `[a = b] = [??]`
+                Some(BindingPattern::AssignmentPattern(mut pattern)) => {
+                    if init_item.is_literal_value(false, ctx) {
```
sapphi-red: I think it is easier to understand the code if we use `init_item.value_type(ctx)` here.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -481,6 +485,266 @@ impl<'a> PeepholeOptimizations {
...
+                        id,
+                        NONE,
+                        Some(init_item),
+                        decl.definite,
+                    ));
+                }
+                // `[] = [??]` => `[] = [??]`
+                None => {
+                    // unused literals can be removed `[] = [1]`, `[] = [void 0]`
+                    if !init_item.is_literal_value(false, ctx) {
```
sapphi-red: I think we can use `init_item.may_have_side_effects(ctx)` as whether it's a sideeffect is what matters here.

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -481,6 +485,266 @@ impl<'a> PeepholeOptimizations {
...
+    }
+
+    /// Simplifies destructuring assignments by transforming array patterns into a sequence of
+    /// variable declarations, whenever possible. This function modifies the input declarations
+    /// and returns whether any changes were made.
+    ///
+    /// For some inputs, this transformation may increase the code size as this transformation
+    /// injects additional temporary variables. But we assume those cases are rare.
+    fn simplify_destructuring_assignment(
+        _kind: VariableDeclarationKind,
```
sapphi-red: we can now remove this parameter.

crates/oxc_minifier/tests/peephole/minimize_statements.rs
```diff
@@ -80,3 +79,81 @@ fn test_max_conditional_depth_caps_return_ternary_chain() {
...
+    test("let [a, , b] = [, , , ]", "let a, b");
+    test("const [a, , b] = [, , , ]", "const a = void 0, b = void 0;");
+    // nested
+    test("let [a, [b, c]] = [1, [2, 3]]", "let a = 1, b = 2, c = 3");
+    test("let [a, [b, [c, d]]] = [1, ...[2, 3]]", "let a = 1, [b, [c, d]] = 2, [] = [3]");
+    test("let [a, [b, [c, ]]] = [1, [...2, 3]]", "let a = 1, [b, [c]] = [...2, 3]");
+    test("let [a, [b, [c, ]]] = [1, [2, [...3]]]", "let a = 1, b = 2, [c] = [...3];");
+    // self reference
+    test("let [a] = [a]", "let a = a");
+    test("let [a, b] = [b, a]", "let b = b");
```
sapphi-red: We probably should add cases like below and explain that we ignore TDZ errors.
```js
let [a, b] = [1, a]
```

armano2: case added, this is actually preety good case that was producing wrong result, it should work now correctly

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -523,6 +526,281 @@ impl<'a> PeepholeOptimizations {
...
+                    new_id.span(),
+                    last.kind,
+                    new_id,
+                    NONE,
+                    new_init,
+                    last.definite,
+                    ctx,
+                );
+                if Self::is_empty_array_destruction_assignment(last) {
+                    ctx.replace_variable_declarator(last, new_decl);
```
armano2: i do understand that those helpers are used to drop references, but i'm not 100 sure if we want to run it here, as there should no be reference changes here, outside removal of void/undefined, but that is handled inside `simplify_array_destruction_assignment`, some guidance would be appreciated

its also not fully clear to me, how to check it / run tests if we remove everything correctly now 

```suggestion
                    *last = new_decl;
```

Dunqing:
@codex review

crates/oxc_minifier/src/peephole/minimize_statements.rs
```diff
@@ -568,6 +571,281 @@ impl<'a> PeepholeOptimizations {
...
+            // check if any symbol used in init is defined in binding identifiers of id
+            // `id = init`
+            // `[a, b] = [b, a]`
+            let mut binding_identifier_symbol_ids = HashSet::new_in(GetAllocator::allocator(ctx));
+            id_kind.bound_names(&mut |ident| {
+                binding_identifier_symbol_ids.insert(ident.symbol_id());
+            });
+
+            if !binding_identifier_symbol_ids.is_empty() {
+                return !binding_identifier_symbol_ids.iter().any(|symbol_id| {
```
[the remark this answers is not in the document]
armano2: change applied i did big refactor and simplification

new version only allow literals to be in-lined

crates/oxc_minifier/src/peephole/minimize_variable_declaration.rs
```diff
@@ -0,0 +1,208 @@
...
+            } else if Self::is_empty_array_destruction_assignment(last) {
+                ctx.drop_variable_declarator(&declarations.remove(i));
+            }
+        }
+    }
+
+    /// Determines whether an array destruction assignment can be simplified.
+    /// - `let [x, y] = [1, 2];` -> true
+    /// - `let [x, y] = [...arr];` -> false
+    fn can_simplify_array_to_array_destruction_assignment(
```
armano2, line 57: this can be simplified now

Issue #15972, opened 2025-11-22
linter: Support more CLI flags via config file options


connorshea:
For (almost?) any CLI flags, we should also be able to configure behavior via the configuration file.

For comparison, ESLint allows configuring `reportUnusedDisableDirectives` via the config file, [as described here](https://eslint.org/docs/latest/use/configure/configuration-files#reporting-unused-disable-directives). And it's enabled by default in ESLint v9.

In order to ensure that everyone is using the same settings on a software project, configuring everything via the config file is preferable. It makes it easier for all LSP-based editors to get consistent linting, and ensures that all developers on a project can be using the same set of rules when running `npm run oxlint` as their coworkers, their editor, and CI.

Flags that should be added to the config file:

- [x] `"typeAware": true/false`, equivalent to the `--type-aware` flag. Defaults to false if unset. https://github.com/oxc-project/oxc/pull/19614
- [x] `"reportUnusedDisableDirectives": off/warn/error`, equivalent to `--report-unused-disable-directives` and `--report-unused-disable-directives-severity`. Defaults to off if unset. (#19799)
- [ ] `"noIgnore: true/false"`, equivalent to `--no-ignore`, defaults to false if unset.
- [ ] `"disableNestedConfig": true/false`, equivalent to `--disable-nested-config`. I'm a bit unsure on this one, as it'd potentially be a bit weird behavior-wise.
- [x] `max-warnings` #19777
- [x] `type-check` #19764 
- [ ] ~`"tsconfig": path/to/file`, equivalent to `--tsconfig=`. Only necessary if the tsconfig file is not at the root of the cwd.~ I (@camc314) am going to deliberatly skip supporting this. We have auto tsconfig discovery that should be used instead of this - if someone want this, they can make an issue and we can re-evaluate.

Output for these from the `--help` command:
```
        --tsconfig=<./tsconfig.json>  TypeScript `tsconfig.json` path for reading path alias and
                              project references for import plugin. If not provided, will look for
                              `tsconfig.json` in the current working directory.
        --no-ignore           Disable excluding files from `.eslintignore` files, --ignore-path
                              flags and --ignore-pattern flags
        --report-unused-disable-directives  Report directive comments like `// oxlint-disable-line`,
                              when no errors would have been reported on that line anyway
        --report-unused-disable-directives-severity=SEVERITY  Same as
                              `--report-unused-disable-directives`, but allows you to specify the
                              severity level of the reported errors. Only one of these two options
                              can be used at a time.
[...]

PR #16583, opened 2025-12-08
feat(linter): Add `linterOptions.typeAware` 

connorshea:
Built with help from GitHub Copilot + Raptor mini (GitHub's preview model).

Part of #15972.

This allows setting the `typeAware` option in the config file. It is nested under the new top-level `linterOptions` field. The name was chosen as it's the one used by ESLint and will also hold things like the `reportUnusedDisableDirectives` setting in the future.

The goal of this to allow teams to set this in the config file, and that'll help ensure that everyone on a project is using the same linter rules. Otherwise, when I adopt this at work (hopefully in the near-ish future) I will need to have every dev on the team do unnecessary manual configuration.

Without this, everyone would need to configure their editor settings to enable type-aware lint rules and also ensure that they're running oxlint with the right command if they use it via the CLI.

My hope is that more linter options will be added under `linterOptions` in the near-ish future to enable more consistency like this, but this seemed like one of the more important options to add directly to the config file.

Examples:

```jsonc
{
  "linterOptions": {
    "typeAware": false
  },
  "rules": {
    /// Even though it's set to `error,` this is NOT run because it is a type-aware rule and we have `typeAware: false`.
    "typescript/no-floating-promises": "error"
  }
}
```

```jsonc
{
  "linterOptions": {
    "typeAware": true
  },
  "rules": {
    /// This is a type-aware rule and will be run because `typeAware` is true in linterOptions.
    "typescript/no-floating-promises": "error"
  }
}
```

TODO:

- [x] This should inform whether type-aware rules are enabled in the LSP/VS Code extension. If the user has not set a setting to specifically override the config file.
- [ ] Maybe mention this config option in the `--type-aware` flag docs.
- [x] Get this working in nested configs
- [x] Should `LintConfig` have a `linter_options` field at all, is that what determines the available options for a nested config file? Or should it be excluded like categories are?
- [x] Should we add a `--disable-type-aware` flag to disable the behavior, for the case where the config has `typeAware: true`? Decision: Let's defer this for now.
- [ ] Add more tests? (LSP test added, need to add a test for nested configs as well)


apps/oxlint/src/lint.rs
```diff
@@ -301,6 +301,15 @@ impl CliRunner {
 
         let config_store = ConfigStore::new(lint_config, nested_configs, external_plugin_store);
 
+        // Determine whether type-aware rules should be enabled.
+        // CLI flag has precedence; otherwise fall back to configuration file.
+        let effective_type_aware = if self.options.type_aware {
+            true
+        } else {
+            // ConfigStore::type_aware_enabled returns whether the base config requested it
+            config_store.type_aware_enabled()
```
Sysix: > Note that I gave type_aware the type Option<bool> because I intended to make it not effect nested config files, we can undo that if we want it to work for nested config files.

Nested Configs are the default. For me, it would be strange to only support it in the root config. 
Example: Team X owns `packages/package-X` and they want to test type aware.
Team Y does not want type aware. 
The `extends` should cover project wide configurations.

The Language server part should be here:

https://github.com/oxc-project/oxc/blob/913b1b348c47eff8fa0d96e31930674b5d68eb0f/crates/oxc_language_server/src/linter/server_linter.rs#L126

theoludwig:
> My hope is that more linter options will be added under linterOptions in the near-ish future

Good idea. :+1: In addition to `typeAware` and `reportUnusedDisableDirectives`, we could also add `typeCheck` and `denyWarnings`.

apps/oxlint/src/lint.rs
connorshea: I have updated this accordingly, and I _believe_ it should now support nested configs and work with the LSP? Not 100% sure on that, need to test. I also need to figure out the best way to implement automated tests for the expected behavior of this...

camc314:
will review tomorrow

crates/oxc_language_server/src/linter/snapshots/fixtures_linter_tsgolint_nested_type_aware@nested__test-with-no-lsp-config.ts.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/oxc_language_server/src/linter/tester.rs
+---
+########## 
+Linted file: fixtures/linter/tsgolint/nested_type_aware/nested/test-with-no-lsp-config.ts
+----------
+########## Diagnostic Reports
+File URI: file://<variable>/fixtures/linter/tsgolint/nested_type_aware/nested/test-with-no-lsp-config.ts
+
+########### Code Actions/Commands
```
connorshea, line 10: This should have diagnostics but currently does not. See `test_nested_config_nested_file_without_lsp_config_is_type_aware` in the `server_linter.rs` for the test case definition. Very plausible that my test is just wrong or I've messed up the implementation itself.

connorshea: I think this isn't working because I determine whether type-aware is enabled or not in [config_store](https://github.com/oxc-project/oxc/blob/8928b58ed6a7df8f6d77ce3f78b246db2cb64133/crates/oxc_linter/src/config/config_store.rs) via the base config.

And so this line in `lint.rs` always returns false if the root config is unset/false, no matter what the nested configs do: 

```rs
let effective_type_aware = if self.options.type_aware { true } else { config_store.type_aware_enabled() };` 
```

Which means tsgolint never gets activated, and there's no way to activate it _after_ the base config has been evaluated. I guess I could try to enable it if _any_ of the configs have it set as true, but I'm not sure that's really the desired behavior?

crates/oxc_language_server/src/linter/server_linter.rs
```diff
@@ -619,6 +630,7 @@ impl ServerLinter {
             || old_options.fix_kind != new_options.fix_kind
             || old_options.unused_disable_directives != new_options.unused_disable_directives
             // TODO: only the TsgoLinter needs to be dropped or created
+            // TODO: Does this need to be updated to also support the linterOptions from the oxlintrc?
```
Sysix, line 644: Changing an `.oxlintrc` config path will be handled by `handle_watched_file_change` and currently always reforces a rebuild.

crates/oxc_language_server/src/linter/server_linter.rs
```diff
@@ -386,7 +396,8 @@ impl Tool for ServerLinter {
             watchers.push(normalize_path(pattern).to_string_lossy().to_string());
         }
 
-        if options.type_aware {
+        // TODO: May need to check for the type-aware setting from the config file as well?
```
Sysix, line 413: see above comment, with `handle_watched_file_change`

Sysix (commented):
great work so far 👍 
I think we need to create a filter for `oxlint` command which files should be passed to `TsGoLintState`. The same should somehow work with the language server.
I build some month ago `LintIgnoreMatcher`, which resolved the `ignorePatterns` for the linted file, maybe you could find/extend some useful code for it :)

connorshea:
Closing in favor of #18047 because that rebase scared me

PR #16616, opened 2025-12-08
feat(editor): adjust oxlint oxfmt config file editor supporting

liangmiQwQ:
Added default editor schema for `.oxfmtrc.json(c)` files

Added nested oxlint config support (`**/.oxlintrc.json`)

And since `oxlintrc.json`, `oxlint.json` are not the config files which can be selected by default, this PR removed them from default schema, user should add `$schema` field manualy if they want to use their own configs


editors/vscode/package.json
```diff
@@ -227,24 +227,25 @@
...
     ],
     "languages": [
       {
         "id": "jsonc",
         "filenames": [
-          "oxlintrc.json",
-          "oxlint.json",
-          ".oxlintrc.json",
-          ".oxlint.json"
+          "**/.oxlintrc.json",
```
Sysix: From reading the [docs](https://code.visualstudio.com/api/references/contribution-points#contributes.languages), glob pattern is only allowed for `filenamePatterns`?

I guess this is the same for `fileMatch` in `jsonValidation`?

liangmiQwQ: I just have tested it locally, what your said is true, I have also found that we can actually use glob pattern (`fileMatch`) but it is not necessary because vscode will searches for all files with `.oxlintrc.json` if we don't use glob pattern, so there is no different between `.oxlintrc.json` and `**/.oxlintrc.json`, so for clearing code, I will remove it

Sysix (approved):

camc314 (approved):

PR #16619, opened 2025-12-09
feat(minifier): substitute function.call(this) for iife

armano2:
there is a need for an asusmption here that user has not modified `Function.prototype.call` to do something else

```js
Function.prototype.call = (arg) => { console.log(arg); }
(function () {}).call("???") 
```


crates/oxc_minifier/src/peephole/minimize_functions.rs
```diff
@@ -0,0 +1,91 @@
+use crate::ctx::Ctx;
+use crate::peephole::PeepholeOptimizations;
+use oxc_allocator::TakeIn;
+use oxc_ast::ast::{Argument, Expression, Function};
+
+impl<'a> PeepholeOptimizations {
+    fn convert_function_expression_to_arrow_function_expression(
```
armano2, line 8: i'm bad at naming stuff

crates/oxc_minifier/src/peephole/minimize_functions.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        test_same("(function () {}).call(foo)");
+        test_same("(function () {}).call(test())");
+        test_same("(function () { foo() }).call(test)");
+        test_same("(function* () {foo()}).call(this)");
+        test_same("(async function () {foo()}).call(this)");
+        test("(function* test () {foo()}).call(this)", "(function* () {foo()}).call(this)");
+        test("(function* test () {}).call(this)", "(function* () {}).call(this)");
+        test("(async function test(){foo()}).call(this)", "(async function (){foo()}).call(this)");
+        test_same("(function test() {console.log(test.name)}).call(this)");
```
sapphi-red:
```suggestion
        test_same("(function test() {console.log(test.name)}).call(this)");

        test_same("(function () {console.log(arguments)}).call(this)");
        test_same("(function () {console.log(new.target)}).call(this)");
```
We need to handle these two cases.

There's also this behavior difference. But I didn't find a way to trigger it with IIFEs.

> The _thisArg_ value is passed without modification as the **this** value. This is a change from Edition 3, where an **undefined** or **null** _thisArg_ is replaced with the [global object](https://tc39.es/ecma262/2025/multipage/global-object.html#sec-global-object) and [ToObject](https://tc39.es/ecma262/2025/multipage/abstract-operations.html#sec-toobject) is applied to all other values and that result is passed as the **this** value. Even though the _thisArg_ is passed without modification, [non-strict functions](https://tc39.es/ecma262/2025/multipage/ordinary-and-exotic-objects-behaviours.html#non-strict-function) still perform these transformations upon entry to the function.
> https://tc39.es/ecma262/2025/multipage/fundamental-objects.html#sec-function.prototype.call

sapphi-red (changes requested):
See the comment above

crates/oxc_minifier/src/peephole/minimize_functions.rs
armano2: this is actually going to be hard, i have hard time finding a way to check all nodes outside of adding impl for `Visit`,

i'm actually surprised that scope don't contain more information's

either by setting a bitmask flag or using references like state, 

- `ThisExpression` -> `this`
- `Identifier` -> `arguments` ()
- `Super` -> `super`
- `MetaProperty` -> `new.target`

q, and toughts?

iterating over entire tree is not optimal, eg, loadsh wraps entire codebase in iff like call, and if user imports any iffe package, this can degrade performance just to remove few bytes

sapphi-red: I don't have an idea here. @Boshen What do you think?

camchenry:
Closing as stale. Feel free to update and re-open.

PR #16636, opened 2025-12-09
fix(transformer): `import.meta.foo.bar` matches and `replaces import.meta.bar`

### LINKED ISSUE #16623: transformer(define): `import.meta.foo.bar` matches and replaces `import.meta.bar`
```js
console.log(import.meta.bar)
```
is transformed to
```js
console.log(34)
```
when `define: { 'import.meta.foo.bar': '34' }` is set. But this should not happen.

Reproduction: https://stackblitz.com/edit/github-xv4yss-hmr4oabe?file=build.mjs

qinyuhang:
This commit fix issue https://github.com/oxc-project/oxc/issues/16623

Fixes https://github.com/oxc-project/oxc/issues/16623

### DIFF
--- a/crates/oxc_transformer_plugins/src/replace_global_defines.rs
+++ b/crates/oxc_transformer_plugins/src/replace_global_defines.rs
@@ -545,7 +545,7 @@ impl<'a> ReplaceGlobalDefines<'a> {
                             // `import.meta.env` should not match `import.meta.env.*`
                             return has_matched_part && !is_full_match;
                         }
-                        return true;
+                        return i == 0;
                     }
                     Expression::Identifier(_) => {
                         return false;
--- a/crates/oxc_transformer_plugins/tests/integrations/replace_global_defines.rs
+++ b/crates/oxc_transformer_plugins/tests/integrations/replace_global_defines.rs
@@ -81,6 +81,14 @@ fn dot() {
 }
 
 #[test]
+fn dot_with_partial_overlap() {
+    let config = config(&[("import.meta.foo.bar", "1")]);
+    test("console.log(import.meta.bar)", "console.log(import.meta.bar)", &config);
+    test("console.log(import.meta.foo)", "console.log(import.meta.foo)", &config);
+    test("console.log(import.meta.foo.bar)", "console.log(1)", &config);
+}
+
+#[test]
 fn dot_with_overlap() {
     let config =
         config(&[("import.meta.env.FOO", "import.meta.env.BAR"), ("import.meta.env", "__foo__")]);


Dunqing:
@IWANABETHATGUY  @sapphi-red Would you mind reviewing this? I don't know how this plugin works. However, if you don't have time, I am happy to review it.

napi/transform/test/import.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, it } from "vitest";
```
IWANABETHATGUY: An integration test case should be enough.

qinyuhang: Do you mean only keep the first part?
```js
    const input = `
console.log(import.meta.bar);
`;
```

qinyuhang: Or Move the test to `transformer.test.ts` file?

qinyuhang: Moved to `napi/transform/test/transform.test.ts`

qinyuhang: @IWANABETHATGUY Hi, the test case is now moved to `napi/transform/test/transform.test.ts`. I'm resolving this comment.

IWANABETHATGUY: I think you misunderstood me, so let me clarify: please just revert the modification in https://github.com/oxc-project/oxc/pull/16636/changes#diff-f621003c9ce8f84b6144f9806dfa7c533da41acf8572bfdf43940736db9f0a3c. Both tests cover the same case, as shown here: https://github.com/oxc-project/oxc/pull/16636/changes#diff-5853f1afdcb039633b0cdd087877c6402c03b8cbe96cbb0e2eb593b738ca90b8

qinyuhang: > I think you misunderstood me, so let me clarify: please just revert the modification in #16636. Both tests cover the same case, as shown here: #16636

@IWANABETHATGUY 
I'm confused, do you mean to remove the test case added in `napi/transform/test/import.test.ts`(now moved to `napi/transform/test/transform.test.ts`)?

qinyuhang: > I think you misunderstood me, so let me clarify: please just revert the modification in #16636. Both tests cover the same case, as shown here: #16636

Hi, the test case in `napi/transform/test/transform.test.ts` is reverted. Would you like to review again? Thanks.

crates/oxc_transformer_plugins/src/replace_global_defines.rs
```diff
@@ -545,7 +545,7 @@ impl<'a> ReplaceGlobalDefines<'a> {
                             // `import.meta.env` should not match `import.meta.env.*`
                             return has_matched_part && !is_full_match;
                         }
-                        return true;
```
qinyuhang, line 548: The current implementation matches token from backwards. returns true on search to the meta expression.
while the iteration of meta expression did not reach meta.
code: `import.meta.bar` -> reached meta.
define: `import.meta.foo.bar` -> index did not reach 0.
the old implementation returns true, because it believes match is over. But it did not check wether the definition reaches the meta.

PR #16679, opened 2025-12-10
docs(linter): Added in the missing jest rules that are compatible with vitest.

Afsoon:
After the review @connorshea did in https://github.com/oxc-project/oxc/pull/16540#pullrequestreview-3548500490, suggested to me add a missing part of documentation in the linter rule. I hadn't added it because the reference rule I used lack of it.

So this PR aims to add it in all jest rules stated as compatible.

## Before this PR

The following Jest linter only had this documentation part:

- consistent-test-it
- expect-expect
- no-alias-methods
- no-commented-out-tests
- no-disabled-tests
- no-focused-tests
- no-identical-title
- no-test-prefixes
- prefer-hooks-in-order
- valid-describe-callback
- valid-expect

## Adding in this PR

- max-expects
- max-nested-describe
- no-conditional-expect
- no-conditional-in-test
- no-duplicate-hooks
- no-hooks
- no-interpolation-in-snapshots
- no-restricted-jest-methods: I have doubts here see my comment in the [vitest-eslint-plugin issue](https://github.com/oxc-project/oxc/issues/4656#issuecomment-3616755697)
- no-restricted-matchers
- no-standalone-expect
- no-test-return-statement
- prefer-comparison-matcher
- prefer-each
- prefer-equality-matcher
- prefer-expect-resolves
- prefer-hooks-on-top
- prefer-lowercase-title
- prefer-mock-promise-shorthand
- prefer-strict-equal
- prefer-to-be
- prefer-to-have-length
- prefer-todo
- require-to-throw-message
- require-top-level-describe

### DIFF
--- a/crates/oxc_linter/src/rules/jest/max_expects.rs
+++ b/crates/oxc_linter/src/rules/jest/max_expects.rs
@@ -66,6 +66,17 @@ declare_oxc_lint!(
     ///     expect(true).toBeDefined();
     /// });
     /// ```
+    ///
+    /// This rule is compatible with [eslint-plugin-vitest](https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/max-expects.md),
+    /// to use it, add the following configuration to your `.oxlintrc.json`:
+    ///
+    /// ```json
+    /// {
+    ///   "rules": {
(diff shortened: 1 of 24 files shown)


camc314 (approved):
Thank you!

PR #16680, opened 2025-12-10
feat(oxfmt): .editorconfig support

### LINKED ISSUE #15850: oxfmt: respect `.editorconfig`
```
    /// Set to tab or space to use tabs or spaces for indentation, respectively.
    /// Option tab implies that an indentation is to be filled by as many hard tabs as possible, with the rest of the indentation filled by spaces.
    /// A non-normative explanation can be found in the indentation_ section.
    /// The values are case-insensitive.
    pub indent_style: EditorConfigProperty<IndentStyle>,

    /// Set to a whole number defining the number of columns used for each indentation level and the width of soft tabs (when supported).
    /// If this equals tab, the indent_size shall be set to the tab size, which should be tab_width (if specified); else, the tab size set by the editor.
    /// The values are case-insensitive.
    pub indent_size: EditorConfigProperty<usize>,

    /// Set to a whole number defining the number of columns used to represent a tab character.
    /// This defaults to the value of indent_size and should not usually need to be specified.
    pub tab_width: EditorConfigProperty<usize>,

    /// Set to lf, cr, or crlf to control how line breaks are represented.
    /// The values are case-insensitive.
    pub end_of_line: EditorConfigProperty<EndOfLine>,

    /// Set to true to remove all whitespace characters preceding newline characters in the file and false to ensure it doesn’t.
    pub trim_trailing_whitespace: EditorConfigProperty<bool>,

    /// Set to true ensure file ends with a newline when saving and false to ensure it doesn’t.
    /// Editors must not insert newlines in empty files when saving those files, even if insert_final_newline = true.
    pub insert_final_newline: EditorConfigProperty<bool>,

    /// Prettier print width.
    /// Not part of spec <https://github.com/editorconfig/editorconfig-vscode/issues/53#issuecomment-462432616>
    /// But documented in <https://prettier.io/docs/next/configuration#editorconfig>
    pub max_line_length: EditorConfigProperty<MaxLineLength>,
```

Same as https://prettier.io/docs/next/configuration#editorconfig
Relate to #15066

Boshen:
closes #15850

@leaysgur handing this over to you, because we need a bigger change to merge the options.

https://github.com/prettier/prettier/blob/1cc636785661d9c0b53a31845348ed0d5456211b/src/config/resolve-config.js#L74-L86

If you look at the TODO line, we need to create default options, then override it by editorconfig, then override by oxfmtrc.json

`.editorconfig` defines overrides already

```
[*]
charset = utf-8
trim_trailing_whitespace = true
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 4

[*.{js,mjs,cjs,jsx,ts,tsx,json,yml}]
indent_size = 2

```

So each path receives a different config. I think you may want to plan ahead for oxfmtrc override, which should behave exactly the same as editorconfig.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,6 @@ indent_size = 4
 
 [*.{js,mjs,cjs,jsx,ts,tsx,json,yml}]
 indent_size = 2
+
+[*.{md}]
+indent_size = 2
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -629,6 +629,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 
 [[package]]
+name = "editorconfig-parser"
+version = "0.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ebf5c801bcfdbcf7a706b19b314daec914ef1f6054047adfd195fff9c38984f9"
+dependencies = [
+ "globset",
+]
+
+[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


leaysgur:
#17043

PR #16736, opened 2025-12-11
feat(oxfmt): `oxfmt --migrate prettier`

### LINKED ISSUE #15849: oxfmt: `--migrate prettier`
Call `prettier` and get its config, then pick out the options we use or differ, then save to `.oxfmtrc.jsonc`

### Migration

- If `printWidth` specified, use it as-is.
  - Otherwise, set `printWidth: 80` for prettier's default
- Warn or comment out not supported format options like `experimentalXxx`
- Warn *meta-transform* plugins may not be applied?
- Anything else?

### Consideration

- First, we need to understand `prettier` API for config resolution
  - IIRC, `prettier.resolveConfig()` requires file path, because their config is resolved by each target file
  - In that case, which file path to be targeted?
  - We do not support nested config, always target root(cwd) config?

Boshen:
For simplicity this PR only reads prettier config and .prettierignore from cwd.

closes #15849

Created by Claude, @leaysgur this over to you.

### DIFF
--- a/apps/oxfmt/src-js/bindings.d.ts
+++ b/apps/oxfmt/src-js/bindings.d.ts
@@ -8,7 +8,8 @@
  * 2. `setup_config_cb`: Callback to setup Prettier config
  * 3. `format_embedded_cb`: Callback to format embedded code in templates
  * 4. `format_file_cb`: Callback to format files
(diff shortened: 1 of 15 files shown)


leaysgur:
#16771 and #16773

PR #16792, opened 2025-12-13
feat(minifier): inline variable while folding for boolean expression

liangmiQwQ:
I noticed the current minifier won't fold the condition which has included a varible even if the value is known and no side effects, it also influences dce

<details>

<summary>👈🏻 Here is an example (from `rolldown test.js -m`)</summary>

### Input
```js
export const a = "hello"

if (a === "world") {
  console.log("will never be run")
}
```

### Output (after oxfmt)
```js
const e = `hello`;
e === `world` && console.log(`will never be run`);
export { e as a };
```

### Expected
```js
const e = `hello`;
export { e as a };
```

</details>

This PR has enhanced it by adding constant getting. It only works on boolean expression to ensure it won't lengthen other expressions unexpectedly


Boshen (commented):
This is a tricky one, I didn't do this yet because code after change may become bigger. For example if the const value is a long string, and the operation is `+`.

I'll have to hold off this for now.

liangmiQwQ:
> This is a tricky one, I didn't do this yet because code after change may become bigger. For example if the const value is a long string, and the operation is `+`.
> 
> I'll have to hold off this for now.

Thanks, I also understand it. so I only inline it for the boolean expression (which uses operator like `=`, `<=`, `>` e.g.) However, when I have time, I may also consider other ways to implement this and make it more universal!

PR #16820, opened 2025-12-14
feat(linter): implement eslint/accessor-pairs rule

tt-a1i:
## Summary

Implements the ESLint `accessor-pairs` rule that enforces getter/setter pairs in objects, classes, and TypeScript interfaces.

### Features
- Checks for setters without getters (default: enabled)
- Checks for getters without setters (default: disabled)
- Supports object literals, class bodies, and Object.defineProperty/defineProperties/create
- TypeScript support for interfaces and type literals (`enforceForTSTypes` option)
- Handles computed property keys using structural comparison via `ContentEq`
- Separates static and instance members in classes

### Configuration options
- `setWithoutGet` (default: true) - Report setters without getters
- `getWithoutSet` (default: false) - Report getters without setters
- `enforceForClassMembers` (default: true) - Apply to class members
- `enforceForTSTypes` (default: false) - Apply to TypeScript interfaces/types

## Test plan
- [x] All 125 pass test cases from ESLint's test suite
- [x] All 177 fail test cases from ESLint's test suite (adapted for TypeScript parser compatibility)
- [x] Snapshot tests generated and verified
- [x] Clippy passes
- [x] Formatting passes

Related to #479 (ESLint core rules umbrella issue)

### DIFF
--- a/crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
+++ b/crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
@@ -1,0 +1,2070 @@
+use oxc_ast::{
+    AstKind,
+    ast::{
+        Argument, CallExpression, ClassBody, Expression, MethodDefinitionKind, ObjectExpression,
+        ObjectPropertyKind, PropertyKey, PropertyKind, TSMethodSignatureKind, TSSignature,
+        TSTypeLiteral,
+    },
+};
+use oxc_diagnostics::OxcDiagnostic;
+use oxc_macros::declare_oxc_lint;
+use oxc_span::{ContentEq, GetSpan, Span};
+use rustc_hash::FxHashMap;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::{AstNode, context::LintContext, rule::Rule};
+
+fn setter_without_getter_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Setter is defined without a getter")
+        .with_help("Define a getter for this property")
+        .with_label(span)
+}
+
+fn getter_without_setter_diagnostic(span: Span) -> OxcDiagnostic {
+    OxcDiagnostic::warn("Getter is defined without a setter")
+        .with_help("Define a setter for this property")
+        .with_label(span)
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AccessorPairsConfig {
+    #[serde(default = "default_true")]
+    set_without_get: bool,
+    #[serde(default)]
+    get_without_set: bool,
+    #[serde(default = "default_true")]
+    enforce_for_class_members: bool,
+    #[serde(default, rename = "enforceForTSTypes")]
+    enforce_for_ts_types: bool,
+}
+
+fn default_true() -> bool {
+    true
+}
+
+impl Default for AccessorPairsConfig {
+    fn default() -> Self {
+        Self {
+            set_without_get: true,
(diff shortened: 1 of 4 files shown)


crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
```diff
@@ -0,0 +1,2074 @@
...
+pub struct AccessorPairsConfig {
+    #[serde(default = "default_true")]
+    set_without_get: bool,
+    #[serde(default)]
+    get_without_set: bool,
+    #[serde(default = "default_true")]
+    enforce_for_class_members: bool,
+    #[serde(default, rename = "enforceForTSTypes")]
+    enforce_for_ts_types: bool,
+}
```
connorshea, line 46: No need for default_true, can just set `default` in the top-level `serde()` and it will use the values from the `impl Default` below.

```suggestion
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
#[serde(rename_all = "camelCase", default)]
pub struct AccessorPairsConfig {
    set_without_get: bool,
    get_without_set: bool,
    enforce_for_class_members: bool,
    #[serde(rename = "enforceForTSTypes")]
    enforce_for_ts_types: bool,
}
```

Should also probably include doc comments for each of these config options, explaining what their purpose is.

crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
```diff
@@ -0,0 +1,2074 @@
...
+);
+
+impl Rule for AccessorPairs {
+    fn from_configuration(value: serde_json::Value) -> Self {
+        let config = value
+            .as_array()
+            .and_then(|arr| arr.first())
+            .map(|v| serde_json::from_value::<AccessorPairsConfig>(v.clone()).unwrap_or_default())
+            .unwrap_or_default();
+        Self(Box::new(config))
```
connorshea: Will also need to add the DefaultRuleConfig import, but this should work:

```suggestion
        serde_json::from_value::<DefaultRuleConfig<AccessorPairs>>(value)
            .unwrap_or_default()
            .into_inner()
```

crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
```diff
@@ -0,0 +1,2074 @@
...
+    ///         this.val = value;
+    ///     }
+    ///     get a() {
+    ///         return this.val;
+    ///     }
+    /// }
+    /// ```
+    AccessorPairs,
+    eslint,
+    pedantic,
```
connorshea, line 119: This will fix the test that's failing:

```suggestion
    pedantic,
    config = AccessorPairsConfig,
```

connorshea (dismissed):
Just a few suggestions

crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
tt-a1i: Done! Applied the top-level `#[serde(default)]` and added doc comments for each config option.

crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
tt-a1i: Done! Used `DefaultRuleConfig<AccessorPairsConfig>` pattern.

crates/oxc_linter/src/rules/eslint/accessor_pairs.rs
tt-a1i: Done! Added `config = AccessorPairsConfig` to the macro.

camc314 (approved):
Thank you!

PR #16880, opened 2025-12-15
feat(linter/eslint-plugin-vitest): Add require-hook as vitest compatible jest rule

Afsoon:

# Sumarry
Add `require hook` as a vitest compatible rule.

Oxlint-migrate PR: https://github.com/oxc-project/oxlint-migrate/pull/288

# Changes made
I had to sligthy modify the implementation to be compatible with both frameworks. The modifications were for 2 specific tests:

```javascripts
("describe.for([])('%s', (value) => {})", None),
(
  "describe('scoped', () => {
      test.scoped({ example: 'value' });
   });",
  None,
 ),
```

Both test cases must pass. The first test case failed because `describe.for` [only exists in vitest](https://vitest.dev/api/#describe-for) and this exact part of `parse_jest_fn.rs` failed in the jest branch.

https://github.com/oxc-project/oxc/blob/eba07e6fd945069e7694e3880b36e4f0f8f78d6d/crates/oxc_linter/src/utils/jest/parse_jest_fn.rs#L109C1-L115C10

Second test failed because `test.scoped` doesn't exist neither frameworks, `scoped` is a custom property made by a user. Reviewing the Jest and Vitest eslint code they are only interested in the beginning, not in the full chain.

### DIFF
--- a/crates/oxc_linter/src/rules/jest/require_hook.rs
+++ b/crates/oxc_linter/src/rules/jest/require_hook.rs
@@ -9,13 +9,14 @@ use oxc_semantic::AstNode;
 use oxc_span::{CompactStr, Span};
 use schemars::JsonSchema;
 use serde::Deserialize;
+use std::borrow::Cow;
 
 use crate::{
     context::LintContext,
     rule::{DefaultRuleConfig, Rule},
     utils::{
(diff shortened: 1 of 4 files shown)


camc314 (approved):
Thank you!

PR #16882, opened 2025-12-15
fix(linter/unicorn/prefer-dom-node-dataset): ignore awaited getAttribute call

Thiry1:
I've followed up on the changes in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2334


camc314 (approved):
Thank you!

PR #16885, opened 2025-12-15
feat(linter): implement eslint/no-inline-comments

tt-a1i:
## Summary

Implement the ESLint `no-inline-comments` rule which disallows comments on the same line as code.

### Features
- Detects inline comments (both line and block comments)
- Supports `ignorePattern` option for regex-based exceptions
- Exempts ESLint/OxLint directive comments (`eslint-disable`, `eslint `, etc.)
- Exempts `global`, `globals`, and `exported` directive comments (block comments only, matching ESLint behavior)
- Handles JSX empty expression comments (`{/* comment */}`)

### Performance optimizations
- Pre-collects `JSXEmptyExpression` spans to avoid repeated AST traversal
- Returns `&str` slices instead of `String` allocations in comment context detection

Closes #2987


camc314 (approved):
Thank you!

PR #16936, opened 2025-12-16
test(oxfmt): Test `ignorePatterns` is inserted at the bottom

leaysgur:
Follow up #16926

PR #16939, opened 2025-12-16
feat(oxfmt): Expose Node.js API: `format(fileName, sourceText, options?)`

### LINKED ISSUE #15913: oxfmt: node API
> https://prettier.io/docs/api

- `format(text: string, options: TODO): string`
  - If possible, calls `prettier.format()` internally as like CLI does

leaysgur:
Fixes #15913


leaysgur:
* **#16939** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16939?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16939?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#16936** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16935** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16934** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#16933** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/16933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=16939).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

apps/oxfmt/src/main_napi.rs
```diff
@@ -91,3 +97,96 @@ pub async fn run_cli(
...
+    // TODO: Plugins support
+    // Use `block_in_place()` to avoid nested async runtime access
+    if let Err(err) = tokio::task::block_in_place(|| {
+        external_formatter.setup_config(&external_config.to_string(), num_of_threads)
+    }) {
+        return FormatResult {
+            code: source_text,
+            errors: vec![OxcError::new(format!("Failed to setup external formatter: {err}"))],
+        };
+    }
```
Dunqing, line 178: Can we set up the config on the JS side? Similar to https://github.com/oxc-project/oxc/pull/16939/changes#r2622086879

apps/oxfmt/src-js/index.ts
```diff
@@ -1,2 +1,113 @@
...
+  if (typeof fileName !== "string") throw new TypeError("`fileName` must be a string");
+  if (typeof sourceText !== "string") throw new TypeError("`sourceText` must be a string");
+
+  return napiFormat(
+    fileName,
+    sourceText,
+    options ?? {},
+    setupConfig,
+    formatEmbeddedCode,
+    formatFile,
```
Dunqing, line 14: Why do we need to pass the `formatFile` function to the Rust side? Can we determine if the file should delegate to `Prettier` on the JS side? If so, that will be more performant and avoid some mutual calling overhead.

apps/oxfmt/src-js/bindings.d.ts
```diff
@@ -1,7 +1,48 @@
 /* auto-generated by NAPI-RS */
 /* eslint-disable */
+export interface Comment {
+  type: 'Line' | 'Block'
+  value: string
+  start: number
+  end: number
+}
```
Dunqing, line 8: Do you know why this type was generated?

Dunqing (changes requested):
I noticed that the current `format` API is directly delegated to the Rust side's `format.` I think we should change it to be used only when necessary; otherwise, we should delegate to `formatFile` on the JS side, which can avoid calling `setupConfig` on the Rust side, as it already owns the configuration. This way, https://github.com/oxc-project/oxc/pull/16934 is no longer needed.

apps/oxfmt/src-js/bindings.d.ts
leaysgur: It seems this comes from https://github.com/oxc-project/oxc/blob/main/crates/oxc_napi/src/comment.rs

We only need `OxcError` from `oxc_napi`, but no way to exclude...?

leaysgur:
@Dunqing As the current design, these are intentional.

> Can we set up the config on the JS side?

To handle this entirely on the JS side, validation and applying default values would have to be done in JS.

However, if `printWidth` is absent in the options, Rust logic is required to determine that it should be `100`.

https://github.com/oxc-project/oxc/blob/main/apps/oxfmt/src/core/config.rs#L37

Incidentally, the upcoming implementation for `editorconfig` will make this approach of caching settings on the first run impossible anyway (since it varies per path...), so coincidentally, it will be resolved.

> Can we determine if the file should delegate to Prettier on the JS side?

Similarly, the logic to determine whether to call Prettier based on the filename exists only on the Rust side.

https://github.com/oxc-project/oxc/blob/main/apps/oxfmt/src/core/support.rs

---

Ultimately, it might be possible to make all necessary parts publicly available via napi and usable on the JS side, but I think a slightly later timing would be better.

apps/oxfmt/src-js/bindings.d.ts
Dunqing: It's fine to leave it

Dunqing:
> To handle this entirely on the JS side, validation and applying default values would have to be done in JS.
> 
> However, if `printWidth` is absent in the options, Rust logic is required to determine that it should be `100`.
> 
> https://github.com/oxc-project/oxc/blob/main/apps/oxfmt/src/core/config.rs#L37
> 

I think this is fine if you pass `{ printWidth: 100, ...userOptions }` to Prettier

> Incidentally, the upcoming implementation for `editorconfig` will make this approach of caching settings on the first run impossible anyway (since it varies per path...), so coincidentally, it will be resolved.

I might remember incorrectly; didn't Boshen say that we only support root-level editorconfig?

> Similarly, the logic to determine whether to call Prettier based on the filename exists only on the Rust side.
> 
> https://github.com/oxc-project/oxc/blob/main/apps/oxfmt/src/core/support.rs


I know that, but on the JS side, shouldn't we only need to check which files can be formatted by the Rust side? That would be much simpler than checking for an external formatter

> 
> Ultimately, it might be possible to make all necessary parts publicly available via napi and usable on the JS side, but I think a slightly later timing would be better.

Anyway, I am not keen to get this change immediately, just brought this up to see what we can improve here. So let's get this merge and iterate on it.

Dunqing (approved):

Dunqing:
### Merge activity

* **Dec 16, 1:42 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 16, 1:42 PM UTC**: `Dunqing` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Dec 16, 1:49 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

leaysgur:
Thank you. I think your point is still valid.

I also just realized that `worker_threads` isn't necessary for API usage, so I'll improve it later. 👍🏻

PR #16946, opened 2025-12-16
feat(linter): suggestion for no-named-as-default

gaurav-rk9:
implements fix suggestion for import/no-named-as-default
closes part of https://github.com/oxc-project/oxc/issues/10477

### DIFF
--- a/crates/oxc_linter/src/rules/import/no_named_as_default.rs
+++ b/crates/oxc_linter/src/rules/import/no_named_as_default.rs
@@ -54,7 +54,8 @@ declare_oxc_lint!(
     /// ```
     NoNamedAsDefault,
     import,
-    suspicious
+    suspicious,
+    suggestion
 );
 
 impl Rule for NoNamedAsDefault {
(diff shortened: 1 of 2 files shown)


crates/oxc_linter/src/rules/import/no_named_as_default.rs
```diff
@@ -72,11 +73,47 @@ impl Rule for NoNamedAsDefault {
...
+                                Span::new(import_span.start, open_brace_pos + 1),
+                                format!("{{ {import_name},"),
+                            )
+                        },
+                    );
+                } else if ctx
+                    .find_next_token_within(import_span.end, import_entry.statement_span.end, "*")
+                    .is_some()
+                {
+                    // `import foo, * as ns` - no fix
```
gaurav-rk9, line 100: or maybe, we could suggest
```
import { foo } from "something";
import * as ns from "something";
```

camc314 (changes requested):
Hey, thanks for working on this. unfortunately, I'm going to close this as won't fix. This is because I think there's too big a chance of incorrect fixes when trying to fix this rule. For example, the default export won't necessary have same type as the named export causing type errors ect.

The original rule, nor eslint-plugin-import-x implementations don't have fixers/suggestions for this rule either.

PR #17024, opened 2025-12-17
feat(parser): add token collection option

lilnasy:
Part of #16207. This is just to measure the performance impact of unconditionally always storing tokens in an `oxc_allocator::Vec`.

### DIFF
--- a/crates/oxc_parser/src/lexer/mod.rs
+++ b/crates/oxc_parser/src/lexer/mod.rs
@@ -6,8 +6,9 @@
 //!     * [v8](https://v8.dev/blog/scanner)
 
 use rustc_hash::FxHashMap;
+use std::fmt::Debug;
 
-use oxc_allocator::Allocator;
+use oxc_allocator::{Allocator, Vec as ArenaVec};
 use oxc_ast::ast::RegExpFlags;
 use oxc_diagnostics::OxcDiagnostic;
 use oxc_span::{SourceType, Span};
@@ -45,6 +46,7 @@ pub struct LexerCheckpoint<'a> {
     source_position: SourcePosition<'a>,
     token: Token,
     errors_snapshot: ErrorSnapshot,
+    tokens_len: usize,
 }
 
 #[derive(Debug, Clone)]
@@ -86,6 +88,11 @@ pub struct Lexer<'a> {
 
     /// `memchr` Finder for end of multi-line comments. Created lazily when first used.
     multi_line_comment_end_finder: Option<memchr::memmem::Finder<'static>>,
+
+    /// Tokens collected from the lexer.
+    ///
+    /// If `collect_tokens` is `false`, this will be `None`.
+    tokens: Option<ArenaVec<'a, Token>>,
 }
 
 impl<'a> Lexer<'a> {
@@ -97,6 +104,7 @@ impl<'a> Lexer<'a> {
         allocator: &'a Allocator,
         source_text: &'a str,
         source_type: SourceType,
+        collect_tokens: bool,
         unique: UniquePromise,
     ) -> Self {
         let source = Source::new(source_text, unique);
@@ -114,6 +122,7 @@ impl<'a> Lexer<'a> {
             escaped_strings: FxHashMap::default(),
             escaped_templates: FxHashMap::default(),
             multi_line_comment_end_finder: None,
+            tokens: if collect_tokens { Some(ArenaVec::new_in(allocator)) } else { None },
         }
     }
 
@@ -126,7 +135,7 @@ impl<'a> Lexer<'a> {
         source_type: SourceType,
     ) -> Self {
         let unique = UniquePromise::new_for_tests_and_benchmarks();
-        Self::new(allocator, source_text, source_type, unique)
+        Self::new(allocator, source_text, source_type, false, unique)
     }
 
     /// Get errors.
@@ -149,10 +158,15 @@ impl<'a> Lexer<'a> {
         } else {
             ErrorSnapshot::Count(self.errors.len())
         };
+        let tokens_len = match &self.tokens {
+            Some(tokens) => tokens.len(),
+            None => 0,
+        };
         LexerCheckpoint {
             source_position: self.source.position(),
             token: self.token,
             errors_snapshot,
+            tokens_len,
         }
     }
 
@@ -168,6 +182,10 @@ impl<'a> Lexer<'a> {
             source_position: self.source.position(),
             token: self.token,
             errors_snapshot,
+            tokens_len: match &self.tokens {
+                Some(tokens) => tokens.len(),
+                None => 0,
+            },
         }
     }
 
@@ -180,6 +198,9 @@ impl<'a> Lexer<'a> {
         }
         self.source.set_position(checkpoint.source_position);
         self.token = checkpoint.token;
+        if let Some(tokens) = self.tokens.as_mut() {
+            tokens.truncate(checkpoint.tokens_len);
+        }
     }
 
     pub fn peek_token(&mut self) -> Token {
@@ -229,6 +250,9 @@ impl<'a> Lexer<'a> {
         self.token.set_end(self.offset());
         let token = self.token;
         self.trivia_builder.handle_token(token);
+        if let Some(tokens) = self.tokens.as_mut() {
+            tokens.push(token);
+        }
         self.token = Token::default();
         token
     }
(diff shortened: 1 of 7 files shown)


lilnasy:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17024?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#17026** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17026?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17025** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17025?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17095** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17024** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17024?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17024?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17047** <a href="https://app.graphite.com/github/pr/oxc-project/oxc/17047?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=17024).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

overlookmotel:
I discussed with Boshen today. He wants to abandon the `ParserConfig` trait approach for now. Plan is to configure whether to store tokens or not with a runtime option. Boshen feels it's OK to take a small slowdown in compiler pipeline in return for less complicated code. If we find the perf hit is too much, we can look at bringing in `ParserConfig` trait later on.

Just to be transparent: Personally I am not entirely on board with this decision - Boshen and I have different priorities. I prioritize perf over all else (pretty much), whereas Boshen puts more weight on avoiding complex generics (readable code) and keeping compile times down. But we follow our fearless captain! So the decision is made.

I'd like to apologise to you. I thought that the approach was agreed already, but it seems not. So I've sent you down a pointless path of getting into the `ParserConfig` stuff (which was not trivial) and then reversed direction. Doing that is not cool - we should have reached agreement on the design *first*. I'm sorry.

So... here's what I suggest:

* Add `tokens: bool` field to `ParseOptions` to enable/disable collecting tokens (default false).
* Add `tokens: ArenaVec<'a, Token>` to `ParserReturn` (which will be empty if collecting tokens is disabled).
* Stack this PR on top of #17047.
* Alter the `parser_tokens` benchmark to pass `tokens: true` in options.

I've added the `parser_tokens` benchmark in a separate PR, so we'll see the costs clearly in CodSpeed on this PR - it'll show cost to both compiler pipeline, and to linter.

---

I think we can improve perf when collecting tokens. Simplest thing would be to pre-allocate capacity in tokens `Vec` for `source_text.len()` tokens. There can never be more tokens than there are bytes in source text, so this will guarantee that the `Vec` doesn't have to grow during the parsing process.

Growing a `Vec` is really expensive (especially if it's large) as all the contents of the `Vec` (pre-growth) have to be copied to the new allocation (post-growth). You can see this cost in the CodSpeed flame graphs for lexer benchmarks (`RawVec::finish_grow::grow`).

Pre-allocating a lot of space is a speed vs memory usage trade-off. I imagine it'll be worth it.

But... since you have Graphite now, please make that optimization in a separate PR on top of this one, so we can see the effect it has on benchmarks in isolation.

lilnasy:
> Simplest thing would be to pre-allocate capacity in tokens Vec for source_text.len() tokens.

#17095 shows it improves performance by ~23% after this change degrades it by 28%, but that is misleading because of how CodSpeed does its maths. When I had the preallocation change in this PR, the degradation dropped down only to ~27%.

tasks/benchmark/benches/parser.rs
```diff
@@ -19,6 +19,7 @@ fn bench_parser(criterion: &mut Criterion) {
                 Parser::new(&allocator, source_text, source_type)
                     .with_options(ParseOptions {
                         parse_regular_expression: true,
+                        collect_tokens: true,
```
overlookmotel: This should be `false` here, and `true` on the *other* benchmark - `bench_parser_with_tokens`.

overlookmotel: But ergh! So the cost to parser with runtime option in parse-transform-minify-print pipeline is 6%-8%. That's a lot. We might have to bring back `ParserConfig`! :)

overlookmotel: I've pushed a commit to switch round which benchmark gets `collect_tokens: true`. I just want the bench results to be clear so I can discuss with Boshen.

I've not restacked rest of the stack - didn't want to touch any of the rest.

crates/oxc_parser/src/lexer/mod.rs
```diff
@@ -239,6 +263,19 @@ impl<'a> Lexer<'a> {
...
+    /// Retrieve collected tokens.
+    /// This should only be called once per source text, as it consumes the tokens vec.
+    /// Panics if the lexer was not configured to collect tokens.
+    #[inline]
+    pub fn tokens(&mut self) -> ArenaVec<'a, Token> {
+        if let Some(tokens) = self.tokens.take() {
+            tokens
+        } else {
+            let backtrace = std::backtrace::Backtrace::capture();
+            panic!("Can't retrieve tokens because they were not collected\n{backtrace}");
```
overlookmotel, line 275: I've never seen `Backtrace` used before. I think `panic!` automatically produces a backtrace, so it's not required. Any reason why you added this?

crates/oxc_parser/src/lib.rs
```diff
@@ -467,13 +482,14 @@ impl<'a> ParserImpl<'a> {
         let (module_record, module_record_errors) = self.module_record_builder.build();
         if errors.len() != 1 {
             errors.reserve(self.lexer.errors.len() + self.errors.len());
-            errors.extend(self.lexer.errors);
-            errors.extend(self.errors);
+            errors.append(&mut self.lexer.errors);
+            errors.append(&mut self.errors);
```
overlookmotel: This change is likely good, but it's incidental to this PR. Want to make a separate PR for this?

crates/oxc_parser/src/lexer/mod.rs
lilnasy: I wasn't seeing the names of the methods in the call stack until I added this. It's possible I missed something.

We don't have to keep this, I needed it just for debugging.

crates/oxc_parser/src/lib.rs
lilnasy: It maybe incidental for `self.errors`, but `extend()` moves `self.lexer` while consuming `self.lexer.errors`. That prevented the `self.lexer.tokens()` call below.

overlookmotel: Actually, maybe it's not good. `extend` is I think slightly cheaper.

But this code could be optimized in other way.

1. `module_record_errors.len()` is not included in reservation.
2. Could extend one of the existing `Vec`s instead of creating a new one.

Something like:

```rs
if !self.source_type.is_typescript() {
  module_record_errors.truncate();
}

errors = self.lexer.errors;
errors.reserve(self.errors.len() + module_record_errors.len());
errors.extend(self.errors);
errors.extend(module_record_errors);
```

overlookmotel: > It maybe incidental for `self.errors`, but `extend()` moves `self.lexer` while consuming `self.lexer.errors`. That prevented the `self.lexer.tokens()` call below.

Ah ha! Sorry I was wrong. It's not incidental.

overlookmotel: But could you just move the `self.lexer.tokens()` call to earlier, before the error-handling code, and then leave it as using `extends()`?

tasks/benchmark/benches/parser.rs
overlookmotel: Let's not worry about performance for now. If you're able to get it working and conformance tests passing, then we'll loop back and fix the perf. If necessary, we may have to go back to `ParserConfig` trait.

crates/oxc_parser/src/lexer/mod.rs
camc314: fyi probably:

https://github.com/Boshen/oxc/blob/main/Cargo.toml#L252-L255

overlookmotel: Ah I see. Usually running with `RUST_BACKTRACE=1` gives you stack traces.

Turning on `debug` temporarily can also help sometimes:

https://github.com/oxc-project/oxc/blob/5a7fcd11e00f4eb60c14152e9bd623c134633b51/Cargo.toml#L247-L255

Let me know if neither of those works.

PR #17074, opened 2025-12-18
fix(minifier): preserve sideeffectful destructing delcaration for array and object destruction

armano2:
fixes #17038

expands logic used to determine if we can remove empty array and object patterns by checking if assigned value is valid:

### DIFF
--- a/crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
+++ b/crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
@@ -11,6 +11,25 @@ impl<'a> PeepholeOptimizations {
             && !ctx.scoping().root_scope_flags().contains_direct_eval()
     }
 
+    fn can_explode_expression_as_array_pattern(expr: &Expression<'a>, ctx: &Ctx<'a, '_>) -> bool {
+        match expr {
+            Expression::ArrayExpression(_)
+            | Expression::StringLiteral(_)
+            | Expression::TemplateLiteral(_) => true,
+            Expression::Identifier(ident) => {
+                ctx.scoping.current_scope_flags().is_function()
+                    && ident.name == "arguments"
+                    && ctx.is_global_reference(ident)
+            }
+            _ => false,
+        }
+    }
+
+    fn can_explode_expression_as_object_pattern(expr: &Expression<'a>, ctx: &Ctx<'a, '_>) -> bool {
+        matches!(expr, Expression::ObjectExpression(_))
+            || Self::can_explode_expression_as_array_pattern(expr, ctx)
+    }
+
     pub fn should_remove_unused_declarator(
         decl: &VariableDeclarator<'a>,
         ctx: &Ctx<'a, '_>,
@@ -29,8 +48,18 @@ impl<'a> PeepholeOptimizations {
                 }
                 false
             }
-            BindingPatternKind::ArrayPattern(ident) => ident.is_empty(),
-            BindingPatternKind::ObjectPattern(ident) => ident.is_empty(),
+            BindingPatternKind::ArrayPattern(ident) => {
+                ident.is_empty()
+                    && decl.init.as_ref().is_some_and(|expr| {
+                        Self::can_explode_expression_as_array_pattern(expr, ctx)
+                    })
+            }
+            BindingPatternKind::ObjectPattern(ident) => {
+                ident.is_empty()
+                    && decl.init.as_ref().is_some_and(|expr| {
+                        Self::can_explode_expression_as_object_pattern(expr, ctx)
+                    })
+            }
             BindingPatternKind::AssignmentPattern(_) => false,
         }
     }
@@ -189,13 +218,28 @@ mod test {
         test_options("var x", "", &options);
         test_options("var x = 1", "", &options);
         test_options("var x = foo", "foo", &options);
+
         test_options("var [] = []", "", &options);
         test_options("var [] = [1]", "", &options);
         test_options("var [] = [foo]", "foo", &options);
+        test_options("var [] = 'foo'", "", &options);
+        test_options(
+            "export var f = () => { var [] = arguments }",
+            "export var f = () => { arguments; }",
+            &options,
+        );
+        test_same_options("var [] = arguments", &options);
+        test_same_options("var [] = null", &options);
+        test_same_options("var [] = void 0", &options);
+        test_same_options("var [] = 1", &options);
+        test_same_options("var [] = a", &options);
+
         test_options("var {} = {}", "", &options);
         test_options("var {} = { a: 1 }", "", &options);
         test_options("var {} = { foo }", "foo", &options);
-        test_options("var {} = { foo: { a } }", "a", &options);
+        test_same_options("var {} = null", &options);
+        test_same_options("var {} = a", &options);
+
         test_same_options("var x; foo(x)", &options);
         test_same_options("export var x", &options);
         test_same_options("using x = foo", &options);
--- a/crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
+++ b/crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
@@ -253,7 +253,6 @@ mod test {
         test_same_options("var {} = null", &options);
         test_same_options("var {} = a", &options);
         test_same_options("var {} = null", &options);
-        test_same_options("var {} = null", &options);
         test_same_options("var {} = void 0", &options);
 
         test_same_options("var x; foo(x)", &options);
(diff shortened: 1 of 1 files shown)


crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -11,6 +11,25 @@ impl<'a> PeepholeOptimizations {
...
+                    && ident.name == "arguments"
+                    && ctx.is_global_reference(ident)
+            }
+            _ => false,
+        }
+    }
+
+    fn can_explode_expression_as_object_pattern(expr: &Expression<'a>, ctx: &Ctx<'a, '_>) -> bool {
+        matches!(expr, Expression::ObjectExpression(_))
+            || Self::can_explode_expression_as_array_pattern(expr, ctx)
```
armano2: this could actually be more extended, any value that is not undefined/null should be allowed

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -11,6 +11,25 @@ impl<'a> PeepholeOptimizations {
             && !ctx.scoping().root_scope_flags().contains_direct_eval()
     }
 
+    fn can_explode_expression_as_array_pattern(expr: &Expression<'a>, ctx: &Ctx<'a, '_>) -> bool {
```
armano2: this name is bad, idk how to actually call it

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -11,6 +11,25 @@ impl<'a> PeepholeOptimizations {
...
+        match expr {
+            Expression::ArrayExpression(_)
+            | Expression::StringLiteral(_)
+            | Expression::TemplateLiteral(_) => true,
+            Expression::Identifier(ident) => {
+                ctx.scoping.current_scope_flags().is_function()
```
armano2: compared to https://github.com/oxc-project/oxc/blob/3d2b492def5ca0d99eacb0145b22734468460d47/crates/oxc_ecmascript/src/side_effects/expressions.rs#L338-L344
we are checking if `arguments` is correctly used within function scope

armano2: @sapphi-red q: what do you think about moving this to be in more centralized place or at-least making helper, like?

```rs
    /// Checks if the given identifier reference corresponds to the "arguments" object 
    /// in the context of a function.
    pub fn is_function_arguments(&self, ident: &IdentifierReference<'a>) -> bool {
        ident.name == "arguments"
            && self.is_global_reference(ident)
            && self.current_scope_flags().is_function()
    }
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
sapphi-red: I think `value_type` can handle them.
```rust
!matches!(
    expr.value_type(ctx),
    ValueType::Null | ValueType::Undefined | ValueType::Undetermined
)
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
sapphi-red: I think we can call it `is_sync_iterator_expr` as that's what we are checking.

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -11,6 +11,25 @@ impl<'a> PeepholeOptimizations {
...
 
+    fn can_explode_expression_as_array_pattern(expr: &Expression<'a>, ctx: &Ctx<'a, '_>) -> bool {
+        match expr {
+            Expression::ArrayExpression(_)
+            | Expression::StringLiteral(_)
+            | Expression::TemplateLiteral(_) => true,
+            Expression::Identifier(ident) => {
+                ident.name == "arguments"
+                    && ctx.is_global_reference(ident)
+                    && ctx.current_scope_flags().is_function()
```
sapphi-red: I guess we need to crawl up until we find a function scope or the top level scope to handle cases like
```js
function foo() {
  console.log((() => arguments)())
}
foo()
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
sapphi-red: I think we can leave it inlined for now. We have to add a method to `crates/oxc_ecmascript/src/global_context.rs` to share the function, but I'm not sure what method should be added.

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -189,13 +218,28 @@ mod test {
...
+
         test_options("var [] = []", "", &options);
         test_options("var [] = [1]", "", &options);
         test_options("var [] = [foo]", "foo", &options);
+        test_options("var [] = 'foo'", "", &options);
+        test_options(
+            "export var f = () => { var [] = arguments }",
+            "export var f = () => { arguments; }",
+            &options,
+        );
```
sapphi-red, line 243:
```suggestion
        test_options(
            "export function f() { var [] = arguments }",
            "export function f() {}",
            &options,
        );
        test_same_options(
            "export var f = () => { var [] = arguments }",
            &options,
        );
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -189,13 +218,28 @@ mod test {
...
+        test_same_options("var [] = null", &options);
+        test_same_options("var [] = void 0", &options);
+        test_same_options("var [] = 1", &options);
+        test_same_options("var [] = a", &options);
+
         test_options("var {} = {}", "", &options);
         test_options("var {} = { a: 1 }", "", &options);
         test_options("var {} = { foo }", "foo", &options);
-        test_options("var {} = { foo: { a } }", "a", &options);
+        test_same_options("var {} = null", &options);
```
sapphi-red, line 253:
```suggestion
        test_same_options("var {} = null", &options);
        test_same_options("var {} = void 0", &options);
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
armano2: i completely missed that one, this is really useful

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
armano2: as for removal of `arguments` stamements, should that be done in this PR?

this is related to `may_have_side_effects` for `IdentifierReference` missing check for `arguments`

https://github.com/oxc-project/oxc/blob/b86a134d3faf822d977acc45eaec8f54a6e34e99/crates/oxc_ecmascript/src/side_effects/expressions.rs#L65-L75

`arguments` is a technically global variable that is not bound to a scope

and called from
https://github.com/oxc-project/oxc/blob/5b5f2fda369b5a06470290eb6063ef557bce5759/crates/oxc_minifier/src/peephole/remove_unused_expression.rs#L17-L33

[fun fact](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Arguments_not_allowed#examples) for `arguments` 

```
function f() {
  class C {
    args = { ...arguments }; // SyntaxError: arguments is not valid in fields
  }
  return new C();
}
```

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -1,8 +1,7 @@
-use oxc_ast::ast::*;
-
-use crate::{CompressOptionsUnused, ctx::Ctx};
-
 use super::PeepholeOptimizations;
+use crate::{CompressOptionsUnused, ctx::Ctx};
+use oxc_ast::ast::*;
+use oxc_ecmascript::constant_evaluation::{DetermineValueType, ValueType};
```
armano2: something is wrong with my formater, idk why it keep changing imports

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
sapphi-red: Ah, I assumed that it was implemented. Yeah, let's leave that in this PR.

crates/oxc_minifier/src/peephole/remove_unused_declaration.rs
```diff
@@ -197,13 +225,37 @@ mod test {
...
+        test_same_options("var [] = a", &options);
+
         test_options("var {} = {}", "", &options);
         test_options("var {} = { a: 1 }", "", &options);
         test_options("var {} = { foo }", "foo", &options);
-        test_options("var {} = { foo: { a } }", "a", &options);
+        test_same_options("var {} = null", &options);
+        test_same_options("var {} = a", &options);
+        test_same_options("var {} = null", &options);
+        test_same_options("var {} = null", &options);
```
sapphi-red:
```suggestion
        test_same_options("var {} = null", &options);
```

sapphi-red (approved):
Thank you!

PR #17118, opened 2025-12-19
feat(linter): implement `unicorn/no-immediate-mutation` rule

tt-a1i:
## Summary

Implements the `unicorn/no-immediate-mutation` rule that disallows mutating a variable immediately after initialization.

Closes https://github.com/oxc-project/oxc/issues/9534

### What it detects

- Array mutations: `array.push()`, `array.unshift()`
- Object property assignments: `obj.foo = bar`
- `Object.assign()` calls
- Set mutations: `set.add()`
- Map mutations: `map.set()`

### Features

- Handles both variable declarations and assignment expressions
- Correctly skips self-referencing mutations (e.g., `array.push(array[0])`)
- Supports all statement containers including SwitchCase
- Validates global references for Object/Set/Map to avoid false positives when these identifiers are shadowed

### Examples

```js
// Bad
const array = [1, 2];
array.push(3);

const obj = { foo: 1 };
obj.bar = 2;

const set = new Set([1]);
set.add(2);

// Good
const array = [1, 2, 3];

const obj = { foo: 1, bar: 2 };

const set = new Set([1, 2]);
```

## Test plan

- All 94 pass cases and 118 fail cases from eslint-plugin-unicorn pass
- Clippy passes with no warnings

## AI Assistance Disclosure

### DIFF
--- a/crates/oxc_linter/src/generated/rule_runner_impls.rs
+++ b/crates/oxc_linter/src/generated/rule_runner_impls.rs
@@ -3301,6 +3301,12 @@ impl RuleRunner for crate::rules::unicorn::no_hex_escape::NoHexEscape {
     const RUN_FUNCTIONS: RuleRunFunctionsImplemented = RuleRunFunctionsImplemented::Run;
 }
 
+impl RuleRunner for crate::rules::unicorn::no_immediate_mutation::NoImmediateMutation {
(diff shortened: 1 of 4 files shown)


camc314 (approved):
Thank you!

PR #17119, opened 2025-12-19
feat(linter): implement vue/no-deprecated-destroyed-lifecycle

tt-a1i:
## Summary

Implements `vue/no-deprecated-destroyed-lifecycle` rule that detects deprecated Vue 2 lifecycle hooks and suggests Vue 3 alternatives.

- Detects `beforeDestroy` → suggests `beforeUnmount`
- Detects `destroyed` → suggests `unmounted`
- Works with `export default { ... }` and `defineComponent({ ... })`
- Handles all key types: identifiers, string literals, template literals
- Handles shorthand properties by converting to `newName: oldName` syntax

Part of #11440

## Test plan

- [x] Unit tests for pass/fail cases
- [x] Fix tests for autofix functionality
- [x] Snapshot tests generated

## AI Assistance Disclosure

### DIFF
--- a/crates/oxc_linter/src/generated/rule_runner_impls.rs
+++ b/crates/oxc_linter/src/generated/rule_runner_impls.rs
@@ -4009,6 +4009,13 @@ impl RuleRunner for crate::rules::vue::max_props::MaxProps {
     const RUN_FUNCTIONS: RuleRunFunctionsImplemented = RuleRunFunctionsImplemented::Run;
 }
 
+impl RuleRunner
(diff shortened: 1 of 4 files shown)


tt-a1i:
Pushed a small follow-up to fix the Lint failure (clippy::single_match) by refactoring the run_once loop to use if let instead of a single-arm match. Also kept the earlier fix-safety guard (skip autofix when beforeUnmount/unmounted already exist in the same object).

PR #17158, opened 2025-12-20
fix(linter): fix false positive in `react-hooks/exhaustive-deps` for method calls

michaelgmcd:
## Summary

  Fixes a false positive in the `exhaustive-deps` rule where calling a method like `form.reset()` would warn about missing `form` dependency even when `form.reset` was already listed in the dependency array. This is how biome functions (pretty large blocker in us migrating).

  Before this fix:
  ```jsx
  useEffect(() => {
    form.reset();
  }, [form.reset]); // ⚠️ incorrectly warned: missing dependency 'form'
```

  After this fix:
```jsx
  useEffect(() => {
    form.reset();
  }, [form.reset]); // ✓ no warning - form.reset covers the call
```

  Both [form.reset] and [form] are now accepted as valid dependency arrays for form.reset() calls.

  Test plan

  - Added test case for method call with method reference in deps (props.toggleEditMode() with [props.toggleEditMode])
  - Added test case for method call with parent object in deps (form.reset() with [form])
  - Updated snapshots to reflect more accurate diagnostic messages (now reports prop.hello instead of prop)

### DIFF
--- a/crates/oxc_linter/src/rules/react/exhaustive_deps.rs
+++ b/crates/oxc_linter/src/rules/react/exhaustive_deps.rs
@@ -1351,7 +1351,19 @@ impl<'a> Visit<'a> for ExhaustiveDepsVisitor<'a, '_> {
         if let Ok(source) = analyze_property_chain(&it.object, self.semantic) {
             if let Some(source) = source {
                 if is_parent_call_expr {
-                    self.found_dependencies.insert(source);
+                    // When calling a method like `form.reset()`, the dependency should be
+                    // `form.reset` (the method reference), not just `form`. This allows
+                    // users to list `form.reset` in their dependency array instead of `form`.
+                    let new_chain = Vec::from([it.property.name]);
+                    let symbol_id =
+                        self.semantic.scoping().get_reference(source.reference_id).symbol_id();
+                    self.found_dependencies.insert(Dependency {
+                        name: source.name,
+                        reference_id: source.reference_id,
+                        span: source.span,
+                        chain: [source.chain, new_chain].concat(),
+                        symbol_id,
+                    });
                 } else {
                     let new_chain = Vec::from([it.property.name]);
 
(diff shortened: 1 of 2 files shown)


camc314 (changes requested):
Requesting changes as a note to myself to investigate this more.

But we (oxlint) reports the same as eslint here.

Screenshot 2025-12-20 at 12 36 49

Edit: although the originlal plugin only reports an issue when it's a call expression (member expression is OK)

Screenshot 2025-12-20 at 12 38 22

camc314:
Thanks for the PR! I had another look at this, and unfortunately, this PR deviates from ESLint's actual behaviour rather than fixing a false potitive.

When using eslint with the following code:

```
// Case 1: form.reset() with [form.reset] - ESLint warns
function Case1({ form }) {
  useEffect(() => {
    form.reset();
  }, [form.reset]);
}
```

```
// Case 2: form.reset() with [form] - ESLint passes
function Case2({ form }) {
  useEffect(() => {
    form.reset();
  }, [form]);
}
```

```
// Case 3: form.reset (member access, NOT a call) with [form.reset] - ESLint passes
function Case3({ form }) {
  useEffect(() => {
    console.log(form.reset);
  }, [form.reset]);
}
```

From these examples, the plugins' logic it that; when  you call a method like form.reset(), you need the object itself (form) as a dependency, not the method reference (form.reset). This is probably because of this bindings - the behavior of a method call depends on what object it's called on:
```ts
const form1 = { data: 'a', reset() { console.log(this.data); } };
const form2 = { data: 'b', reset: form1.reset }; // same function reference!
form1.reset(); // logs 'a'
form2.reset(); // logs 'b'`
```
If form changes from form1 to form2, form.reset could be referentially equal (same function), but calling form.reset() produces different results because this is different. That's why ESLint requires form (the object) as the dependency, not form.reset (the method reference).produces different results because this is different. That's why ESLint requires form (the object) as the dependency, not form.reset (the method reference).

oxlint's fixer (`--fix` or `--fix-dangerously`) should support automatically fixing these cases to help you migrate - if not that sounds like a bug, and we can create an issue to get the missing parts of the fixer fully implemented.

PR #17160, opened 2025-12-20
feat(linter): implement jest/valid-expect-in-promise rule

tt-a1i:
## Summary

Implements `jest/valid-expect-in-promise` rule - the last unimplemented recommended Jest rule.

This rule ensures that `expect()` calls inside promise callbacks (`.then()`, `.catch()`, `.finally()`) are properly awaited or returned to prevent false positive test results.

**Handles:**
- Variable declarations with promise chains
- `return` and `await` for handling promises
- `Promise.all/race/any/resolve` patterns
- `expect(promise).resolves/rejects` patterns
- Arrow function implicit returns
- `.each` data row parameters vs done callbacks

## Known Limitations

Aligned with eslint-plugin-jest behavior, with the following known limitations:

1. **Scope-insensitive variable tracking**: Promise variables are tracked by name only, without full scope analysis. Shadowing in nested blocks may cause false negatives.

2. **`expect(promise).resolves/rejects` treated as handled**: When a promise variable is passed to `expect().resolves/rejects`, it's considered "handled" for tracking purposes. The missing `return`/`await` on the resolves chain itself is the domain of `valid-expect` rule, not this rule.

3. **Variable reassignment edge cases**: Reassigning promise variables before await/return may not be correctly tracked in all cases.

4. **Conservative `.each` done callback detection**: For `.each` tests, if the callback has ≥1 parameter and the last parameter is a binding identifier, it's conservatively treated as having a done callback (since we can't statically determine how many data columns exist). For regular `it`/`test`, exactly 1 binding identifier parameter is required. This may skip checking in edge cases.

5. **`globalAliases` not supported**: oxc doesn't support the `globalAliases` setting from eslint-plugin-jest, so the corresponding test case is omitted.

## Test plan

- [x] 120+ pass cases
- [x] 70+ fail cases  
- [x] `cargo test -p oxc_linter -- valid_expect_in_promise` passes
- [x] `cargo clippy -p oxc_linter` passes with no warnings

## AI Assistance Disclosure

### DIFF
--- a/apps/oxlint/src/snapshots/_--vitest-plugin -c fixtures__eslintrc_vitest_replace__eslintrc.json fixtures__eslintrc_vitest_replace__foo.test.js@oxlint.snap
+++ b/apps/oxlint/src/snapshots/_--vitest-plugin -c fixtures__eslintrc_vitest_replace__eslintrc.json fixtures__eslintrc_vitest_replace__foo.test.js@oxlint.snap
@@ -23,7 +23,7 @@ working directory:
   help: Remove the appending `.skip`
 
 Found 1 warning and 1 error.
-Finished in <variable>ms on 1 file with 102 rules using 1 threads.
+Finished in <variable>ms on 1 file with 103 rules using 1 threads.
 ----------
 CLI result: LintFoundErrors
 ----------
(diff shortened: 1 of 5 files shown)


camc314 (changes requested):
Thanks for working on this, but not all of the test cases have been ported.

I have ported them for you, please fix the failures, and I will re-review

I'm going to mark this as draft - please mark as ready for review once all tests aare passing, and i'll take another look.

tt-a1i:
All the test cases you added have been addressed and CI is green now. Ready for another review when you have time. Thanks! @camc314

tt-a1i:
gentle ping @camc314 - this should be ready for re-review now

camchenry:
It looks like this was merged already in https://github.com/oxc-project/oxc/pull/21170. Thank you for your contribution!

PR #17185, opened 2025-12-20
fix(linter/plugins): use correct ScriptKind for tokens

wagenet:
## Summary

When parsing tokens for JS plugins, the TypeScript parser was always using `ScriptKind.TSX` regardless of file extension. This caused TypeScript to incorrectly parse generic arrow functions like `<T>() => {}` in `.ts` files, creating bogus `JsxText` tokens that overlap with comments.

The fix reads the JSX and TypeScript flags from the buffer metadata:
- `isJsx()` reads the `is_jsx` flag from `RawTransferMetadata`
- `isTypescript()` reads the `is_ts` flag from `RawTransferMetadata`

This ensures the `ScriptKind` reflects what Oxc actually parsed, not just what the file extension suggests.

## The Bug

Using the wrong ScriptKind (e.g., TSX for a .ts file) causes TypeScript to parse generic arrow functions like `<T>() => {}` incorrectly, creating bogus `JsxText` tokens that overlap with comments. This manifests as errors like:

```
Error running JS plugin.
Error: Overlapping token/comments: last end: 2715, next start: 2689
```

## Changes

- Added `is_jsx` field to `RawTransferMetadata` struct (in both `napi/parser` and `crates/oxc_linter`)
- Added `IS_JSX_FLAG_POS` constant to generated constants (via `ast_tools`)
- Added `isJsx()` and `isTypescript()` helper functions in `source_code.ts`
- Updated `getScriptKind()` in `tokens_parse.ts` to use buffer flags instead of file extension matching
- Added snapshot test with `generic_arrow.ts` test file

## Test plan

- [x] Added `generic_arrow.ts` test file that triggers the bug when parsed as TSX
- [x] Snapshot test verifies correct tokenization (no overlapping tokens error)
- [x] All oxlint tests pass


wagenet:
I'm not sure this is the very best solution, but it does address a real issue with incorrect parsing.

apps/oxlint/src-js/plugins/tokens_parse.ts
```diff
@@ -266,3 +268,31 @@ function hasJSXAncestor(node: ts.Node | undefined): boolean {
...
+
+/**
+ * Determine TypeScript ScriptKind based on file extension.
+ *
+ * @param path - File path
+ * @param tsModule - TypeScript module
+ * @returns Appropriate ScriptKind for the file
+ */
+function getScriptKind(path: string, tsModule: typeof import("typescript")): ts.ScriptKind {
+  const ext = path.slice(path.lastIndexOf(".")).toLowerCase();
```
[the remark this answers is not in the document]
camc314: I think this would be fine as we'd fallthrough to the default case

apps/oxlint/test/tokens_script_kind.test.ts
camc314, line 1: these tests are probably overkill - I think we can just add a snapshot test asserting the tokens -just add another file in here

https://github.com/oxc-project/oxc/blob/0c1f82ba043f0916a9826b63257cc9d03871390a/apps/oxlint/test/fixtures/tokens/files/index.js#L0-L1

camc314: fixed

camc314 (approved):

overlookmotel (changes requested):
I'm afraid this is not the right way to solve this. We need to get the "is it JSX" flag from the `Program`. It's not completely trivial, hence why I left it as a TODO.

overlookmotel:
I've marked it as "requested changes" not necessarily to request that you make the changes, but just to stop the PR getting merged in current state by accident.

If you want to tackle it, need to create a `const` with the offset of the byte for the JSX flag in `SourceType` in codegen, and then look up that byte in the buffer, to determine whether file is JSX or not. There are some other `const`s for similar things e.g. byte offset of source text `length` field.

I don't know if that's enough to go on...

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -66,6 +79,36 @@ export function initSourceText(): void {
   sourceText = textDecoder.decode(buffer.subarray(0, sourceByteLen));
 }
 
+// Byte offset of `variant` field within `SourceType` struct.
+// SourceType layout: language (1 byte), module_kind (1 byte), variant (1 byte).
+// `variant` is `LanguageVariant` enum: 0 = Standard, 1 = Jsx.
+const SOURCE_TYPE_VARIANT_OFFSET = 2;
```
overlookmotel: This should not be hard-coded here. It should be generated by the codegen (`raw_transfer.rs`).

apps/oxlint/test/fixtures/tokens/files/generic_arrow.ts
```diff
@@ -0,0 +1,8 @@
+const obj = {
+  fn: <T>(arg: T): T => {
+    return arg;
+  },
+};
+
+// A comment after the object
+export { obj };
```
overlookmotel, line 8: Is it possible to also add a `.tsx` fixture? (which would fail to parse if parsed as plain TS)

overlookmotel (commented):
Thank you *very* much for working on this.

Actually - and sorry to change my mind - there's an easier way to do this which I hadn't spotted before.

`RawTransferMetadata` already contains an `is_ts` field. We should add an `is_jsx` field too.

For convoluted reasons, there are 2 copies of `RawTransferMetadata` in `apps/oxlint` and `napi/parser`. Both must be identical, so you'd need to add the field to both.

Then get `ast_tools` codegen to generate an `IS_JSX_FLAG_POS` constant, same as for `IS_TS_FLAG_POS`.

---

Please also run the following after making the change, and commit any change to the conformance snapshot:

```sh
cd apps/oxlint
pnpm run init-conformance
pnpm run build-conformance
pnpm run conformance
```

`init-conformance` only needs to be run once - it initializes the ESLint submodule. If you make changes after that, just run the last 2 commands again.

I don't know if this is covered by ESLint's conformance tests, but it'd be good to find out!

apps/oxlint/src-js/generated/constants.ts
```diff
@@ -5,5 +5,7 @@ export const BUFFER_SIZE = 2147483616;
 export const BUFFER_ALIGN = 4294967296;
 export const DATA_POINTER_POS_32 = 536870902;
 export const IS_TS_FLAG_POS = 2147483612;
+export const IS_JSX_FLAG_POS = 2147483613;
 export const PROGRAM_OFFSET = 0;
 export const SOURCE_LEN_OFFSET = 16;
+export const SOURCE_TYPE_OFFSET = 124;
```
overlookmotel: I think this addition is extraneous? Can't see anywhere using `SOURCE_TYPE_OFFSET`.

apps/oxlint/src-js/plugins/source_code.ts
```diff
@@ -66,6 +71,30 @@ export function initSourceText(): void {
...
+ * Check if the source is TypeScript (as opposed to JavaScript).
+ * This reads the `is_ts` flag from the buffer metadata.
+ *
+ * @returns `true` if the source is TypeScript
+ */
+export function isTypescript(): boolean {
+  debugAssertIsNonNull(buffer);
+  // IS_TS_FLAG_POS contains a bool: 0 = JS, 1 = TS
+  return buffer[IS_TS_FLAG_POS] === 1;
+}
```
overlookmotel: It seems a bit overkill to have separate functions for these, since they're essentially 1 line, and only used in `getScriptKind`. Could you inline them into `getScriptKind` please?

You'll need to export `buffer` from this module and import it in `tokens_parse.ts`. I think that's fine to do.

overlookmotel (approved):
Thanks for addressing my feedback. And sorry for the delay - only got back to my desk today.

(oops I didn't mean to approve the changes)

overlookmotel:
I think CI fail is caused by an oddity in snapshot encoding. If you build `oxlint` with `pnpm run build-test`, and then rerun the tests with snapshot update, it should solve it.

overlookmotel:
You've waited a long time for review, so I've taken the liberty of making the changes I requested myself. Will merge once CI passes.

overlookmotel (approved):

PR #17198, opened 2025-12-21
test(linter): Error when passed non-array config objects

connorshea:
This has been a problem in various rules' tests, so I think we just fix it here and be done with it. You are only allowed to pass an array value for rule configs in the Tester, nothing else.

As an aside, I think the anchor-is-valid rule is basically broken, in that the config object it takes is _completely_ different from the one in [the original rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-is-valid.md). We should maybe fix that.


camc314 (approved):

camc314:
### Merge activity

* **Dec 21, 1:30 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc) once it meets the requirements.
* **Dec 21, 1:31 PM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).
* **Dec 21, 1:52 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=oxc).

PR #17199, opened 2025-12-21
feat(linter): PoC to Enforce config option validity for many rules

connorshea:
Basically, a proof-of-concept to add `deny_unknown_fields` on config structs across the linter and replace `unwrap_or_default()` with `unwrap()` in `from_configuration`, so it raises an error if the unwrapping fails. As part of this, I added a check for `null` to ensure that an _unset_ value for config (e.g. `"plugin/rule-name: "error"`) does not cause a failure.

I've only done the `deny_unknown_fields` part for rules where the DefaultRuleConfig was used.

Part of the goal here is to see if we can use this to help users find invalid configs/unsupported config options that get migrated from ESLint more easily. Ideally we would also ship a fuller JSON Schema that outputs schema definitions for _all_ the lint rules' config options based on the config structs we've built out, but that is a bit more difficult than this change.

I think it may be worth seeing what effect this change has on the oxc-ecosystem-ci job. I definitely would not recommend merging this right now, though, as it'd be a bad idea to potentially break things right as JS plugins are going into alpha.

### DIFF
--- a/crates/oxc_linter/src/rule.rs
+++ b/crates/oxc_linter/src/rule.rs
@@ -15,8 +15,8 @@ use crate::{
 
 pub trait Rule: Sized + Default + fmt::Debug {
     /// Initialize from eslint json configuration
-    fn from_configuration(_value: serde_json::Value) -> Self {
-        Self::default()
+    fn from_configuration(_value: serde_json::Value) -> Result<Self, serde_json::error::Error> {
+        Ok(Self::default())
     }
 
     /// Serialize rule configuration to JSON. Only used for sending rule configurations
@@ -82,7 +82,7 @@ pub trait Rule: Sized + Default + fmt::Debug {
 ///
 /// ```ignore
 /// impl Rule for MyRule {
-///     fn from_configuration(value: serde_json::Value) -> Self {
+///     fn from_configuration(value: serde_json::Value) -> Result<Self, serde_json::error::Error> {
 ///         serde_json::from_value::<DefaultRuleConfig<MyRule>>(value)
 ///             .unwrap_or_default()
 ///             .into_inner()
--- a/crates/oxc_linter/src/rule.rs
+++ b/crates/oxc_linter/src/rule.rs
@@ -125,7 +125,7 @@ where
                 .unwrap_or_else(T::default);
             Ok(DefaultRuleConfig(config))
         } else if value == serde_json::Value::Null {
-            return Ok(DefaultRuleConfig(T::default()));
+            Ok(DefaultRuleConfig(T::default()))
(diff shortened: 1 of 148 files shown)


connorshea:
Surprisingly, this didn't cause any new problems in the ecosystem ci...? https://github.com/oxc-project/oxc-ecosystem-ci/actions/runs/20404230629

I guess not many of them are configuring rules, admittedly, since they don't have a config file at all.

camc314:
This is a great start - but lets change our approach of how we implement this slighly?

I think the plan should look something like this:

1. convert `from_configuration` to return a result
```diff
 impl Rule for AccessorPairs {
-    fn from_configuration(value: serde_json::Value) -> Self {
-        serde_json::from_value::<DefaultRuleConfig<AccessorPairs>>(value)
+    fn from_configuration(value: serde_json::Value) -> Result<Self, serde_json::error::Error> {
+        Ok(serde_json::from_value::<DefaultRuleConfig<AccessorPairs>>(value)
             .unwrap_or_default()
-            .into_inner()
+            .into_inner())
     }
```
(in this first version, we can just always return OK, and unwrap in the caller (here)):

https://github.com/Boshen/oxc/blob/main/crates/oxc_linter/src/config/rules.rs#L107

2. convert a single rule to start returning an actual result, and implement the error handling in the config builder (probably want to store all errors, then propogate them up via the CLI)
3. once we've tested 2, and all looks good, do it for all the other rules

connorshea:
> This is a great start - but lets change our approach of how we implement this slighly?
> 
> I think the plan should look something like this:
> 
>     1. convert `from_configuration` to return a result
> 
> 
> ```diff
>  impl Rule for AccessorPairs {
> -    fn from_configuration(value: serde_json::Value) -> Self {
> -        serde_json::from_value::<DefaultRuleConfig<AccessorPairs>>(value)
> +    fn from_configuration(value: serde_json::Value) -> Result<Self, serde_json::error::Error> {
> +        Ok(serde_json::from_value::<DefaultRuleConfig<AccessorPairs>>(value)
>              .unwrap_or_default()
> -            .into_inner()
> +            .into_inner())
>      }
> ```
> 
> (in this first version, we can just always return OK, and unwrap in the caller (here)):
> 
> https://github.com/Boshen/oxc/blob/main/crates/oxc_linter/src/config/rules.rs#L107
> 
>     2. convert a single rule to start returning an actual result, and implement the error handling in the config builder (probably want to store all errors, then propogate them up via the CLI)
> 
>     3. once we've tested 2, and all looks good, do it for all the other rules

This makes sense to me, I've opened #17479 to implement the first step.

PR #17224, opened 2025-12-21
feat(linter/eslint-plugin-vitest): Implement valid-title eslint rule

Afsoon:

Set `valid-title` as a compatible vitest rule

https://github.com/oxc-project/oxlint-migrate/pull/301

### DIFF
--- a/crates/oxc_linter/src/config/settings/vitest.rs
+++ b/crates/oxc_linter/src/config/settings/vitest.rs
@@ -1,3 +1,4 @@
+use oxc_span::CompactStr;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -12,4 +13,6 @@ pub struct VitestPluginSettings {
     /// to accommodate TypeScript type checking scenarios.
     #[serde(default)]
     pub typecheck: bool,
+    #[serde(rename = "vitestImports", default)]
+    pub vitest_imports: Vec<CompactStr>,
 }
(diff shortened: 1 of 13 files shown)


crates/oxc_linter/src/config/settings/vitest.rs
```diff
@@ -12,4 +13,7 @@ pub struct VitestPluginSettings {
     /// to accommodate TypeScript type checking scenarios.
     #[serde(default)]
     pub typecheck: bool,
+    #[serde(default)]
+    #[serde(rename = "vitestImports")]
```
connorshea:
```suggestion
    #[serde(rename = "vitestImports", default)]
```

connorshea (changes requested):
Started a review and then realized that... _technically_ valid-title is already implemented for vitest partially via the jest/valid-title rule? It just wasn't done properly: https://github.com/oxc-project/oxc/commit/bd746034e04e2b001f556465932db94207fcfe9a

Do we need to do anything specific for that?

Afsoon:
@connorshea Looks that not, we have to extend the current `valid-title`, apart of adding as a compatible vitest rule. Vitest have some test that depends on the `vitestImports` pattern. I'm going to set this PR as draft and update it when I have time for it. 

Thanks!

Afsoon:
Doing from zero `valid-title` and `warn-todo` I have found some issues that are probably better talked before to do anything. The current PR doesn't handle the `vitestImports` test. The other issue is how vitest and jest are detected.

Afsoon:
@camc314 I have recovered the changes I made to handle the test with `vitestImports` settings, and added thoses test in the rule

Afsoon:
Closing, as it makes more sense re do it once https://github.com/oxc-project/oxc/pull/18603 and https://github.com/oxc-project/oxc/pull/18467 are merged

PR #17238, opened 2025-12-22
fix(diagnostics): handle WouldBlock errors when writing to stdout

wagenet:
## Summary

- Fix panic when oxlint stdout pipe buffer fills up
- Add retry logic for `WouldBlock` errors with 1ms backoff
- Handle `Interrupted` and `BrokenPipe` gracefully

When oxlint is spawned as a child process (e.g., via Node.js execa/child_process), stdout is connected to a pipe. If the parent process doesn't consume output fast enough, the pipe buffer fills (8KB on macOS, 64KB on Linux) and writes return `WouldBlock` (EAGAIN, error code 35 on macOS). Previously this caused a panic:

```
Os { code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" }
```

This PR replaces `write_all` with `write_all_retry` that properly handles transient I/O errors by retrying.

## Test plan

- [x] Added unit tests for `write_all_retry` and `flush_retry`
- [x] Tests cover: success, WouldBlock retry, BrokenPipe, other errors, partial writes, empty buffer

### DIFF
--- a/crates/oxc_diagnostics/src/service.rs
+++ b/crates/oxc_diagnostics/src/service.rs
@@ -227,14 +227,13 @@ impl DiagnosticService {
     }
 
     /// Write all bytes to the writer, retrying on `WouldBlock` and `Interrupted`.
-    /// Returns `Ok(true)` on success, `Ok(false)` if pipe is broken (consumer gone),
-    /// or `Err` on other I/O errors.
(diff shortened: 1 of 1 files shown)


camc314:
fixed  by #19292

PR #17279, opened 2025-12-22
feat(linter): added rule for `react/jsx-max-depth`

Kenzo-Wada:
related: https://github.com/oxc-project/oxc/issues/1022

added react/jsx-max-depth rule to oxlinter!


camc314 (changes requested):
I re-ported the tests, and not all are passing. 

Please try to avoin deleting test cases when porting these rules.


Please mark this as ready for review once the tests are passing, and I will take another look

Kenzo-Wada:
@camc314 thanks always! i think i've done all tests, now ready to review.

camc314 (changes requested):
The performance regression here is massive - we'll need to get it down to a much more reasonable level before it can be merged

crates/oxc_linter/src/rules/react/jsx_max_depth.rs
```diff
@@ -0,0 +1,519 @@
...
+    ctx: &LintContext<'_>,
+    visited_symbols: &mut FxHashSet<SymbolId>,
+) -> usize {
+    let mut max_child_depth = 0;
+
+    for child_node in ctx.nodes().iter() {
+        let parent = ctx.nodes().parent_node(child_node.id());
+        if parent.id() != node.id() {
+            continue;
+        }
```
camc314: this is going to be ridiculously slow. use ctx.nodes.parent or ancestors

crates/oxc_linter/src/rules/react/jsx_max_depth.rs
```diff
@@ -0,0 +1,519 @@
...
+/// Check if a JSX node is a leaf (has no JSX children)
+fn is_leaf_jsx_node(node: &AstNode<'_>, ctx: &LintContext<'_>) -> bool {
+    for child_node in ctx.nodes().iter() {
+        let parent = ctx.nodes().parent_node(child_node.id());
+        if parent.id() == node.id()
+            && matches!(child_node.kind(), AstKind::JSXElement(_) | AstKind::JSXFragment(_))
+        {
+            return false;
+        }
+    }
```
camc314: same here. this, iterating over all nodes to find the parent is going to be really really slow

crates/oxc_linter/src/rules/react/jsx_max_depth.rs
Kenzo-Wada: fixed: https://github.com/oxc-project/oxc/pull/17279/changes/24d1667e5ceebab50fa68cbf68a4c7d2c2ee0e47

crates/oxc_linter/src/rules/react/jsx_max_depth.rs
Kenzo-Wada: fixed: https://github.com/oxc-project/oxc/pull/17279/changes/24d1667e5ceebab50fa68cbf68a4c7d2c2ee0e47

camc314 (approved):
Thank you!

